PR #31890, opened 2022-09-01
[CIGolang]add longduration(NonPreRelease) case

kuiwang02:
/cc @Xia-Zhao-rh

### DIFF
--- a/ci-operator/step-registry/openshift/e2e/test/qe/long-duration/openshift-e2e-test-qe-long-duration-chain.yaml
+++ b/ci-operator/step-registry/openshift/e2e/test/qe/long-duration/openshift-e2e-test-qe-long-duration-chain.yaml
@@ -1,6 +1,7 @@
 chain:
   as: openshift-e2e-test-qe-long-duration
   steps:
+  - ref: openshift-extended-test-longduration
   - ref: cucushift-pre
   - ref: cucushift-e2e-long-duration
   - ref: cucushift-e2e-report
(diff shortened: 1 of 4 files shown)


ci-operator/step-registry/openshift/e2e/test/qe/long-duration/openshift-e2e-test-qe-long-duration-chain.yaml
```diff
@@ -1,6 +1,7 @@
 chain:
   as: openshift-e2e-test-qe-long-duration
   steps:
+  - ref: openshift-extended-test-longduration
```
liangxia, line 4: For this special use case, perhaps we should run `openshift-extended-test-longduration` after `cucushift-e2e-long-duration`, given cucushift runs for 30m40s, and ginkgo runs for 7h57m30s, so that we do not need to wait for 8 hours before we can see cucushift results.

kuiwang02: @liangxia Thanks, I understand your concern. 
The golang testing is becoming focus, so I would like to run golang case on refresh cluster firstly to get clean result. I guess it should be key concern from release manager
About if we could get cucushift case result quickly before test is done, it should not be key concern from release manager's view.

So, I would like to keep golang before cucushifit here.

@chengzhang1016 @jhou1 could you please share your opinion here? thanks

PR #31841, opened 2022-08-31
add workflow vsphere-upi-disconnected-sdn-workers-rhcos-rhel8

lihongan:

PR #31793, opened 2022-08-30
add 4.12 e2e tests for aws-ipi-sdn-workers-rhel8 related

lihongan:
follow up https://github.com/openshift/release/pull/31307

PR #31891, opened 2022-09-01
add 4.12 e2e test for vsphere-upi-disconnected-sdn-workers-rhcos-rhel8

lihongan:
follow up https://github.com/openshift/release/pull/31841

and fix the tag issue in https://github.com/openshift/release/pull/31793


lihongan:
/assign @liangxia 
Looks the test on vsphere is executed, but other two jobs on aws were failed in installation.

PR #31074, opened 2022-08-04
MGMT-11730: Run selected ztp jobs on c3.medium

adriengentil:
Keep `m3.large` for `3masters` and `capi` job flavors, use `c3.medium` for all the rest.
I tested the `m3.small` but the jobs were a bit flaky.

PR #31895, opened 2022-09-01
MGMT-11730: allow users to inject user_data in packet-setup step

adriengentil:
This changes makes possible to inject custom user data in the
packet-setup step in order to customize the machine after it gets
provisionned.

This is particulary useful when using the baremetalds-ipi-pre chain
because there is no way to inject user scripts once the machine is
provisionned and before dev-script starts.

You can see here an example on how we plan to use it (we want to 
customize the partitioning scheme):
https://github.com/openshift/release/pull/31074/files#diff-61d53b66022e5ad9455cbc2a71f7881db29651af5031c957ac918b46a05f31e6

### DIFF
--- a/ci-operator/step-registry/baremetalds/packet/setup/baremetalds-packet-setup-commands.sh
+++ b/ci-operator/step-registry/baremetalds/packet/setup/baremetalds-packet-setup-commands.sh
@@ -199,12 +199,17 @@ cat > packet-setup.yaml <<-EOF
     - cluster_type: "{{ lookup('env', 'CLUSTER_TYPE') }}"
     - packet_project_id: "{{ lookup('file', lookup('env', 'CLUSTER_PROFILE_DIR') + '/packet-project-id') }}"
     - packet_auth_token: "{{ lookup('file', lookup('env', 'CLUSTER_PROFILE_DIR') + '/packet-auth-token') }}"
-
+    - user_data_filename: "{{ lookup('env', 'USER_DATA_FILENAME') }}"
   tasks:
   - name: check cluster type
     fail:
       msg: "Unsupported CLUSTER_TYPE '{{ cluster_type }}'"
     when: "cluster_type is not regex('^packet.*$|^equinix.*$')"
+
+  - name: load user-data file content
+    set_fact:
+      user_data: "{{ lookup('file', lookup('env', 'SHARED_DIR') + '/' + user_data_filename) }}"
+    when: user_data_filename != ""
 
   - name: create Packet host with error handling
     block:
@@ -217,6 +222,7 @@ cat > packet-setup.yaml <<-EOF
         plan: ${PACKET_PLAN}
         facility: any
         tags: "{{ 'PR:', lookup('env', 'PULL_NUMBER'), 'Job name:', lookup('env', 'JOB_NAME'), 'Job id:', lookup('env', 'PROW_JOB_ID') }}"
+        user_data: "{{ user_data | default(omit) }}"
       register: hosts
       no_log: true
     - name: write device info to file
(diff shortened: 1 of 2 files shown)


ci-operator/step-registry/baremetalds/packet/setup/baremetalds-packet-setup-commands.sh
```diff
@@ -199,7 +199,7 @@ cat > packet-setup.yaml <<-EOF
     - cluster_type: "{{ lookup('env', 'CLUSTER_TYPE') }}"
     - packet_project_id: "{{ lookup('file', lookup('env', 'CLUSTER_PROFILE_DIR') + '/packet-project-id') }}"
     - packet_auth_token: "{{ lookup('file', lookup('env', 'CLUSTER_PROFILE_DIR') + '/packet-auth-token') }}"
-
+    - user_data: "{{ lookup('file', lookup('env', 'SHARED_DIR') + '/user-data', errors='ignore') }}"
```
andfasano, line 202: Setting the `user-data` via `SHARED_DIR` seems a little bit tricky, since it assumes that a previous step exists and filled it. Wouldn't be easier to fetch it directly from the step env vars (defaulted to empty)?

adriengentil: I did not want to have a variable defined in a workflow with a script in it.
Also, having a step gives us more freedom is we want to generate it from a template.

What I can do is to define a new variable that specifies the user-data filename in `SHARED_DIR`. This variable would be empty by default. What do you think?

osherdp: What's the benefit with that last suggestion? It makes it possible for steps to choose the location for such files?
Because I think it's mostly a downside, as it means there's no real convention to where it resides (and either way ci-operator does not support calling steps twice with different parameters, so generalizing it doesn't provide much)

adriengentil: It's a way to signal that a file is available to be consumed, and ensure a failure will be risen if it is not there.

The file must reside in `SHARED_DIR` and comply with the limitation of this folder, i.e: [flat hierarchy](https://docs.ci.openshift.org/docs/architecture/step-registry/#sharing-data-between-steps). So at the end, it would be a variable `USER_DATA_FILENAME` set at workflow level with the filename that is wanted (e.g. "user-data"), the step that generates the user data would generate the script in `${SHARED_DIR}/${USER_DATA_FILENAME}`. and the ansible script would look for it in `${SHARED_DIR}/${USER_DATA_FILENAME}`.

This is what I had in mind.

osherdp: ack. so it does make sense to do that and make things more explicit rather than implicit 
sounds good to me

osherdp (approved):

ci-operator/step-registry/baremetalds/packet/setup/baremetalds-packet-setup-commands.sh
andfasano: I understand that in some cases having it as a step parameter could not be too handy, even though I wouldn't expect in general to see very big user-data contents (if that's the case, and if there's the need to heavily customize the instance, my preference would be to have an Ansible customization step somewhere).
At least, and agree with @osherdp  on that, the definition of the new variable makes more clear the step documentation part for a generic user, so it also sounds reasonable to me.

adriengentil:
/unhold
The job failures are unrelated to this change.

PR #31896, opened 2022-09-01
remove OWNERS file as no members are active in openshift

smg247:
Sally, if you would still like to be an owner here let me know. If you no longer will be maintaining the `step-registry/fips-check` we should remove the file so that proper approvers can be added by blunderbuss.

/cc @sallyom


smg247:
I guess we can't just remove the OWNERS file here, so we need to find new owners...

PR #31897, opened 2022-09-01
Hypershift: Make node counts for mgmt cluster more appropriate

alvaroaleman:
Right now, the nested mgmt cluster nodecount is hardcodedn to 3*3=9
which seems to be too little for the e2e tests as they show failed
scheduling warnings all the time and is definitely too much for the
non-ha controlplane we use in the conformance tests.


enxebre:
/lgtm
/hold
periodic related fail?

enxebre:
fwiw I'm also planning to merge hypershift-aws-setup-nested-management-cluster-chain with hypershift-aws-create-cluster, they are just the same action, then the nested workflow can set envs and use the kubeconfigs appropriately.

alvaroaleman:
/hold cancel

> periodic related fail?

It seems this was caused by the tests not being in sync with the latest promoted release

PR #31903, opened 2022-09-01
[CVP-3094] Create prow jobs & configs for OCP v4.12

sonam1412:


smg247:
@sonam1412, `blunderbuss` has failed here and assigned droslean and I to this PR, but we are not the proper reviewers for this. You should look for reviewers from https://github.com/openshift/release/blob/master/ci-operator/config/redhat-openshift-ecosystem/cvp/OWNERS to check this PR.

/uncc smg247 droslean

PR #31909, opened 2022-09-01
don't run expensive etcd tests on doc changes

tjungblu:
as recommended at https://github.com/openshift/cluster-etcd-operator/pull/920#issuecomment-1234537413


kikisdeliveryservice:
@wking@wking iirc you don't need to do `skip_if_only_changed` if the job is optional right?

tjungblu:
I believe it would still run the optional jobs and not skip them - they would surely not block the merge when they fail.

tjungblu:
/assign @Elbehery 

Mustafa when you have some time tomorrow, could you PTAL for a lgtm?

PR #31915, opened 2022-09-01
Update observability builds to omit REMOTE_SRC vars to avoid double m…

mark-nc:
…irror

Signed-off-by: Mark Anderson <markande@redhat.com>


mark-nc:
For https://github.com/stolostron/backlog/issues/24969

PR #31936, opened 2022-09-02
Set up jobs for CPaaS support-packages

Molter73:
This PR adds 2 test jobs related to downstream collector driver builds:
- A check for finding any drivers that might have failed to compile, similar to the one we have for upstream builds.
- A job that pushes the built drivers into a GCP bucket to create a cache, builds the support packages for the downstream drivers and pushes that to another GCP bucket to make it available.

*Reviewer notes*
- I tried using separate steps for pushing the drivers, however, because the support package requires the drivers to be already available on GCP and there is no easy way to sync test steps, I decided to simply merge them together.
- Before merging the PR, the tests will be converted to be periodics, by adding cron 0 0 * * * to their config.

*Follow up steps*
- Create a collector image using these drivers and have it run integration tests on the platforms using kernels we have drivers built for (RHEL 7 and 8).
- Right now the workflow builds drivers for a single version (the master branch), the downstream build will be modified so it can build drivers for all supported collector versions and the drivers will be dumped into the image that is already being mirrored into OSCI.

### DIFF
--- a/ci-operator/config/stackrox/collector/stackrox-collector-master.yaml
+++ b/ci-operator/config/stackrox/collector/stackrox-collector-master.yaml
@@ -895,7 +895,9 @@ tests:
     FAILURES_DIR=/FAILURES/ /scripts/drivers-build-failures.sh
   container:
     from: cpaas-drivers
+  cron: 0 0 * * *
 - as: cpaas-push-drivers-and-support-package-post
+  cron: 0 0 * * *
   steps:
     test:
     - as: push-drivers
@@ -913,6 +915,6 @@ tests:
           cpu: 100m
           memory: 2000Mi
 zz_generated_metadata:
-  branch: mauro-setup-cpaas-support-packages
+  branch: master
   org: stackrox
   repo: collector
--- a/ci-operator/jobs/stackrox/collector/stackrox-collector-master-periodics.yaml
+++ b/ci-operator/jobs/stackrox/collector/stackrox-collector-master-periodics.yaml
@@ -1,0 +1,95 @@
+periodics:
+- agent: kubernetes
+  cluster: build01
+  cron: 0 0 * * *
+  decorate: true
+  decoration_config:
+    skip_cloning: true
+  extra_refs:
+  - base_ref: master
+    org: stackrox
+    repo: collector
+  labels:
+    ci.openshift.io/generator: prowgen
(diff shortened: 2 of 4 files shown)


Molter73:
Before merging the PR, the tests will be converted to be periodics, by adding `cron 0 0 * * *` to their config.

ci-operator/config/stackrox/collector/stackrox-collector-mauro-setup-cpaas-support-packages.yaml
```diff
@@ -866,7 +881,34 @@ tests:
...
+        requests:
+          cpu: 100m
+          memory: 2000Mi
+- as: cpaas-push-drivers-post
+  steps:
+    test:
+    - as: push-drivers-and-support-package
+      commands: |
+        /scripts/push-drivers.sh gs://collector-modules-osci/cpaas/
+        /scripts/cpaas-support-package.sh gs://collector-modules-osci/cpaas/
```
JoukoVirtanen, line 901: Use a variable to specify the bucket here

erthalion:
> Before merging the PR, the tests will be converted to be periodics, by adding cron 0 0 * * * to their config.

From what I recall one can use periodics right from the start, OSCI will still run pipelines in a similar way per change in the PR.

ci-operator/config/stackrox/collector/stackrox-collector-mauro-setup-cpaas-support-packages.yaml
```diff
@@ -849,7 +864,7 @@ tests:
     test:
     - as: push-drivers
       commands: |
-        /scripts/push-drivers.sh
+        /scripts/push-drivers.sh gs://collector-modules-osci
```
erthalion, line 867: It's probably not related to the change, but I've missed this part. Would you remind me pls, why is it needed to call `push-drivers.sh` with a test and a post-submit?

Molter73: This needs to run in a test so that we can provide credentials, thus preventing just anyone to push garbage onto the GCP bucket.

Molter73: As for why is it needed on both pre and post submits, it's because sometimes we need to rebuild the drivers on PRs and this helps with testing them.

PR #31663, opened 2022-08-25
OKD-51: mco & machine-os config updates for 4.12+

LorbusChris:
/cc @vrutkovs 
/cc @cgwalters
/cc @yuqi-zhang

PR #31963, opened 2022-09-02
OKD: Fix machine-os-content builds

LorbusChris:
PR https://github.com/openshift/release/pull/31663 introduced a change
that was supposed to first build and tag `fedora-coreos`, and then tag
that image as `machine-os-content`, too.
However it seems to not work as expected.
This change should fix that.


LorbusChris:
/cc @vrutkovs 
/cc @aleskandro
master/images has passed so this seems to work

PR #31965, opened 2022-09-02
MIXEDARCH-37: Arm worker nodes for heterogeneous workflow

deepsm007:
Workflow changes for heterogeneous deployment with ARM worker nodes. Updating the interval to specific periodic cron and balance the CI load. These tests are being triggered by release-controller based on nightlies.

https://issues.redhat.com/browse/MIXEDARCH-37 
/cc @jeffdyoung @Prashanth684 @aleskandro

### DIFF
--- a/ci-operator/config/openshift/multiarch/openshift-multiarch-master__nightly-4.11.yaml
+++ b/ci-operator/config/openshift/multiarch/openshift-multiarch-master__nightly-4.11.yaml
@@ -130,23 +130,24 @@ resources:
       memory: 200Mi
 tests:
 - as: ocp-e2e-aws-heterogeneous
-  interval: 72h
+  cron: 33 7 * * 3
   steps:
     cluster_profile: aws-2
     dependencies:
       OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE: release:multi-latest
     env:
       BASE_DOMAIN: aws-2.ci.openshift.org
-    workflow: openshift-e2e-aws
+    workflow: openshift-e2e-aws-heterogeneous
 - as: ocp-e2e-serial-aws-heterogeneous
-  interval: 72h
+  cron: 51 15 * * 4
   steps:
     cluster_profile: aws-2
     dependencies:
       OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE: release:multi-latest
     env:
       BASE_DOMAIN: aws-2.ci.openshift.org
-    workflow: openshift-e2e-aws-serial
+      TEST_SUITE: openshift/conformance/serial
+    workflow: openshift-e2e-aws-heterogeneous
 - as: ocp-e2e-aws-arm64
   cron: 58 16 */1 * 1-5
   steps:
--- a/ci-operator/step-registry/ipi/install/heterogeneous/ipi-install-heterogeneous-commands.sh
+++ b/ci-operator/step-registry/ipi/install/heterogeneous/ipi-install-heterogeneous-commands.sh
@@ -1,0 +1,63 @@
+#!/bin/bash
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+export KUBECONFIG=${SHARED_DIR}/kubeconfig
+
+trap 'CHILDREN=$(jobs -p); if test -n "${CHILDREN}"; then kill ${CHILDREN} && wait; fi' TERM
+
+# Make sure yq is installed
+if [ ! -f /tmp/yq-v4 ]; then
+  # TODO move to image
+  curl -L "https://github.com/mikefarah/yq/releases/download/v4.25.3/yq_linux_$(uname -m | sed s/aarch64/arm64/ | sed s/x86_64/amd64/)" -o /tmp/yq-v4 && chmod +x /tmp/yq-v4
+  PATH=${PATH}:/tmp
+fi
+
+# Make sure jq is installed
+if ! command -v jq; then
+  # TODO move to image
+  curl -sL https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64 > /tmp/jq
+  chmod +x /tmp/jq
+  PATH=${PATH}:/tmp
+fi
+
+dir=/tmp/installer
+mkdir "${dir}/"
+
+REGION=${LEASED_RESOURCE}
+
+echo "Fetching Worker MachineSet..."
+oc -n openshift-machine-api get -o json machinesets | jq '[.items[] | select(.spec.template.metadata.labels["machine.openshift.io/cluster-api-machine-role"] == "worker")][0]' > ${dir}/99_openshift-cluster-api_worker-machineset-0.yaml
+
+# AMI for AWS ARM
+echo "Extracting AMI..."
+oc -n openshift-machine-config-operator get configmap/coreos-bootimages -oyaml > ${dir}/coreos-bootimages.yaml
+yq-v4 eval ".data.stream" ${dir}/coreos-bootimages.yaml > ${dir}/machineset.yaml
+amiid_workers_additional=$(yq-v4 ".architectures.${ADDITIONAL_WORKER_ARCHITECTURE}.images.aws.regions.\"${REGION}\".image" ${dir}/machineset.yaml)
+echo "Updating the machineset with ${ADDITIONAL_WORKER_VM_TYPE} and ami ${amiid_workers_additional} ..."
+yq-v4 eval ".metadata.name += \"-additional\"
+| .spec.replicas = ${ADDITIONAL_WORKERS}
+| .spec.selector.matchLabels.\"machine.openshift.io/cluster-api-machineset\" = .metadata.name
+| .spec.template.metadata.labels.\"machine.openshift.io/cluster-api-machineset\" = .metadata.name" \
+${dir}/99_openshift-cluster-api_worker-machineset-0.yaml > ${dir}/99_openshift-cluster-api_worker-machineset-additional.yaml
+
+yq-v4 eval ".spec.template.spec.providerSpec.value.ami.id = \"${amiid_workers_additional}\"
+            | .spec.template.spec.providerSpec.value.instanceType = \"${ADDITIONAL_WORKER_VM_TYPE}\"
+            " -i ${dir}/99_openshift-cluster-api_worker-machineset-additional.yaml
+
+echo "Creating ${ADDITIONAL_WORKER_ARCHITECTURE} worker MachineSet"
+exec oc create -f ${dir}/99_openshift-cluster-api_worker-machineset-additional.yaml &
+
+wait "$!"
+ret="$?"
+
+if test "${ret}" -eq 0 ; then
+  touch  "${SHARED_DIR}/success"
+  # Save console URL in `console.url` file so that ci-chat-bot could report success
+  echo "https://$(env KUBECONFIG=${dir}/auth/kubeconfig oc -n openshift-console get routes console -o=jsonpath='{.spec.host}')" > "${SHARED_DIR}/console.url"
+fi
+
+echo "Exiting with ret=${ret}"
+exit "${ret}"
--- a/ci-operator/config/openshift/multiarch/openshift-multiarch-master__nightly-4.12.yaml
+++ b/ci-operator/config/openshift/multiarch/openshift-multiarch-master__nightly-4.12.yaml
@@ -130,23 +130,24 @@ resources:
       memory: 200Mi
 tests:
 - as: ocp-e2e-aws-ovn-heterogeneous
-  interval: 72h
+  cron: 08 11 * * 5
   steps:
     cluster_profile: aws-2
     dependencies:
       OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE: release:multi-latest
     env:
       BASE_DOMAIN: aws-2.ci.openshift.org
-    workflow: openshift-e2e-aws
+    workflow: openshift-e2e-aws-heterogeneous
 - as: ocp-e2e-serial-aws-ovn-heterogeneous
-  interval: 72h
+  cron: 19 23 * * 2
   steps:
     cluster_profile: aws-2
     dependencies:
       OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE: release:multi-latest
     env:
       BASE_DOMAIN: aws-2.ci.openshift.org
-    workflow: openshift-e2e-aws-serial
+      TEST_SUITE: openshift/conformance/serial
+    workflow: openshift-e2e-aws-heterogeneous
 - as: ocp-e2e-aws-sdn-arm64
   interval: 72h
   steps:
--- a/ci-operator/jobs/openshift/multiarch/openshift-multiarch-master-periodics.yaml
+++ b/ci-operator/jobs/openshift/multiarch/openshift-multiarch-master-periodics.yaml
@@ -1888,6 +1888,7 @@ periodics:
         secretName: result-aggregator
 - agent: kubernetes
   cluster: build05
+  cron: 33 7 * * 3
   decorate: true
   decoration_config:
     skip_cloning: true
@@ -1895,7 +1896,6 @@ periodics:
   - base_ref: master
     org: openshift
     repo: multiarch
-  interval: 72h
   labels:
     ci-operator.openshift.io/cloud: aws
     ci-operator.openshift.io/cloud-cluster-profile: aws-2
@@ -2642,6 +2642,7 @@ periodics:
         secretName: result-aggregator
 - agent: kubernetes
   cluster: build05
+  cron: 51 15 * * 4
   decorate: true
   decoration_config:
     skip_cloning: true
@@ -2649,7 +2650,6 @@ periodics:
   - base_ref: master
     org: openshift
     repo: multiarch
-  interval: 72h
   labels:
     ci-operator.openshift.io/cloud: aws
     ci-operator.openshift.io/cloud-cluster-profile: aws-2
@@ -4075,6 +4075,7 @@ periodics:
         secretName: result-aggregator
 - agent: kubernetes
   cluster: build05
+  cron: 08 11 * * 5
   decorate: true
   decoration_config:
     skip_cloning: true
@@ -4082,7 +4083,6 @@ periodics:
   - base_ref: master
     org: openshift
     repo: multiarch
-  interval: 72h
   labels:
     ci-operator.openshift.io/cloud: aws
     ci-operator.openshift.io/cloud-cluster-profile: aws-2
@@ -4830,6 +4830,7 @@ periodics:
         secretName: result-aggregator
 - agent: kubernetes
   cluster: build05
+  cron: 19 23 * * 2
   decorate: true
   decoration_config:
     skip_cloning: true
@@ -4837,7 +4838,6 @@ periodics:
   - base_ref: master
     org: openshift
     repo: multiarch
-  interval: 72h
   labels:
     ci-operator.openshift.io/cloud: aws
     ci-operator.openshift.io/cloud-cluster-profile: aws-2
--- a/ci-operator/step-registry/ipi/install/heterogeneous/OWNERS
+++ b/ci-operator/step-registry/ipi/install/heterogeneous/OWNERS
@@ -1,0 +1,6 @@
+approvers:
+- aleskandro
+- deepsm007
+- jaypoulz
+- jeffdyoung
+- Prashanth684
--- a/ci-operator/step-registry/ipi/install/heterogeneous/ipi-install-heterogeneous-ref.metadata.json
+++ b/ci-operator/step-registry/ipi/install/heterogeneous/ipi-install-heterogeneous-ref.metadata.json
@@ -1,0 +1,12 @@
+{
+	"path": "ipi/install/heterogeneous/ipi-install-heterogeneous-ref.yaml",
+	"owners": {
+		"approvers": [
+			"aleskandro",
+			"deepsm007",
+			"jaypoulz",
+			"jeffdyoung",
+			"Prashanth684"
+		]
+	}
+}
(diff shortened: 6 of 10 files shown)


ci-operator/config/openshift/multiarch/openshift-multiarch-master__nightly-4.11.yaml
```diff
@@ -136,7 +136,11 @@ tests:
     dependencies:
       OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE: release:multi-latest
     env:
+      ADDITIONAL_WORKER: "2"
       BASE_DOMAIN: aws-2.ci.openshift.org
+      HETEROGENEOUS: "true"
+      WORKER_ARCHITECTURE: aarch64
```
aleskandro, line 142: Hi @deepsm007 , can we rename WORKER_ARCHITECTURE to `ADDITIONAL_WORKERS_ARCHITECTURE` and the same for `ADDITIONAL_WORKER_VM_TYPE`? It would make clear that these values only affect the additional/patched workers, not the "normal" ones.

ci-operator/step-registry/ipi/install/install/ipi-install-install-commands.sh
```diff
@@ -300,6 +300,26 @@ echo "$(date +%s)" > "${SHARED_DIR}/TEST_TIME_INSTALL_START"
...
+  fi
+  ls ${dir}/openshift
+  /tmp/yq-go eval ".metadata.name += \"-additional\"
+   | .spec.replicas = ${ADDITIONAL_WORKER}
+   | .spec.selector.matchLabels.\"machine.openshift.io/cluster-api-machineset\" = .metadata.name
+   | .spec.template.metadata.labels.\"machine.openshift.io/cluster-api-machineset\" = .metadata.name" \
+   ${dir}/openshift/99_openshift-cluster-api_worker-machineset-0.yaml > ${dir}/openshift/99_openshift-cluster-api_worker-machineset-additional.yaml
+   amiid_workers_additional=$(openshift-install coreos print-stream-json | /tmp/yq-go ".architectures.${ADDITIONAL_WORKER_ARCHITECTURE}.images.aws.regions.\"${LEASED_RESOURCE}\".image")
+   /tmp/yq-go eval ".spec.template.spec.providerSpec.value.ami.id = \"${amiid_workers_additional}\"
+   | .spec.template.spec.providerSpec.value.instanceType = \"${ADDITIONAL_WORKER_VM_TYPE}\"
```
aleskandro, line 319:
```
/tmp/yq-go eval ".spec.template.spec.providerSpec.value.ami.id = \"${amiid_workers_additional}\"
               | .spec.template.spec.providerSpec.value.instanceType = \"${ADDITIONAL_WORKER_VM_TYPE}\"
                " -i ${dir}/openshift/99_openshift-cluster-api_worker-machineset-additional.yaml
```

ci-operator/step-registry/ipi/install/install/ipi-install-install-commands.sh
```diff
@@ -300,6 +300,25 @@ echo "$(date +%s)" > "${SHARED_DIR}/TEST_TIME_INSTALL_START"
 openshift-install --dir="${dir}" create manifests &
 wait "$!"
 
+# Adds arm64 workers for heterogeneous deployment
+if [[ "${HETEROGENEOUS:-}"  == 'true' ]]; then
```
aleskandro, line 304:
```
if [ "${HETEROGENEOUS:-}"  = 'true' ]; then
```

to be posix-compliant

ci-operator/step-registry/ipi/install/install/ipi-install-install-commands.sh
```diff
@@ -300,6 +300,25 @@ echo "$(date +%s)" > "${SHARED_DIR}/TEST_TIME_INSTALL_START"
...
+  if [ ! -f /tmp/yq ]; then
```
aleskandro, line 305: should we add a condition like

```
if [ ! -f /tmp/yq ] && [ $(/tmp/yq -V | sed 's/^.*version \([0-9]*\)\..*$/\1/') -ge 4 ]; then
```

to be sure that we are using the proper yq version and there is no need to download it?

deepsm007: yq needs to be added as part of the image. Its been utilized for other scenarios, there are different version being used by different teams and we want to confirm before changing and not breaking stuff

Prashanth684:
the test failures for the e2e are because a lot of the tests pull images from the image registry. we would need to denylist these tests which pull images from the image registry (like tools, debug etc..) until imagestreams and imageregistry support pulling manifestlists.

ci-operator/step-registry/openshift/e2e/aws/heterogeneous/conf/OWNERS
```diff
@@ -0,0 +1,6 @@
+approvers:
```
wking, line 1: If your approver-set for this step is expected to continue to match the parent directory, you can use a symlink here instead of the copy/paste (e.g. see [`hack/dedup-owners.sh`](https://github.com/openshift/release/blob/master/hack/dedup-owners.sh)).

ci-operator/step-registry/openshift/e2e/aws/heterogeneous/conf/openshift-e2e-aws-heterogeneous-conf-commands.sh
```diff
@@ -0,0 +1,85 @@
...
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+# List of exclude tests from conformance/serial suite
+if [ "${HETEROGENEOUS:-}" = 'true' ]; then
+    cat > "${SHARED_DIR}/excluded_tests" << EOF
+"[sig-apps][Feature:DeploymentConfig] deploymentconfigs  should adhere to Three Laws of Controllers [Skipped:Disconnected] [Suite:openshift/conformance/parallel]"
+"[sig-apps][Feature:DeploymentConfig] deploymentconfigs adoption will orphan all RCs and adopt them back when recreated [Skipped:Disconnected] [Suite:openshift/conformance/parallel]"
```
wking, line 11: This feels like a brittle approach.  Is there something tracking teaching the origin suite to handle this, either by making the test-cases robust enough to work on heterogeneous clusters, or by teaching them to skip automatically on heterogeneous clusters?

wking: Ah, your later "This is temporary until we support manifest list for internal registry." suggests that will make the cluster robust enough to run these tests, and my ask is just extending that `documentation` string to include a tracker ref for that effort.

ci-operator/step-registry/openshift/e2e/test/openshift-e2e-test-commands.sh
```diff
@@ -269,7 +269,10 @@ function suite() {
         echo "${TESTS}" | grep "${TEST_SKIPS}" || { exit_code=$?; echo 'Error: no tests were found matching the TEST_SKIPS regex:'; echo "$TEST_SKIPS"; return $exit_code; } &&
         TEST_ARGS="${TEST_ARGS:-} --file /tmp/tests"
     fi &&
-
+    # TO-DO remove heterogeneous block after manifest list support in internal registry
+    if [ -f "${SHARED_DIR}/heterogeneous_tests" ]; then
+        TEST_ARGS="${TEST_ARGS:-} --file ${SHARED_DIR}/heterogeneous_tests"
```
wking, line 274: I'm a bit confused about how this will interact with `TEST_SKIPS`.  Can we use `TEST_SKIPS` for the stopgap heterogeneous exclusions?  Or is it too hard to build the large skip list into a regexp or something?

deepsm007: I tried handling the list into a regex but it was more complicated for a temporary effort. There aren't much common patterns between the list (70 tests)

ci-operator/step-registry/ipi/install/install/ipi-install-install-commands.sh
```diff
@@ -300,6 +300,32 @@ echo "$(date +%s)" > "${SHARED_DIR}/TEST_TIME_INSTALL_START"
 openshift-install --dir="${dir}" create manifests &
 wait "$!"
 
+# Adds arm64 workers for heterogeneous deployment
+if [ "${HETEROGENEOUS:-}" = 'true' ]; then
+  if [ ! -f /tmp/yq-v4 ]; then
+    curl -L "https://github.com/mikefarah/yq/releases/download/v4.25.3/yq_linux_$(uname -m | sed s/aarch64/arm64/ | sed s/x86_64/amd64/)" -o /tmp/yq-v4 && chmod +x /tmp/yq-v4
```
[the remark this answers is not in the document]
deepsm007: The install command already handles multiple yq internal to the pod. We plan to have a larger clean-up for individual worflow/chains. It is using yqv3 for prometheus as well 😓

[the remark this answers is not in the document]
deepsm007: The container is using `installer` image as suppose to a `upi-installer` https://github.com/openshift/release/pull/31965/files#diff-b45b888d80972af63e3a533b34212a24876af63d4551951d03f205ca78be6cb4R3

ci-operator/step-registry/ipi/install/install/ipi-install-install-ref.yaml
```diff
@@ -28,6 +28,18 @@ ref:
   - name: SPOT_INSTANCES
     default: "false"
     documentation:  "Use AWS Spot Instances for worker nodes.  Set to 'true' to opt into spot instances.  Explicitly set to 'false' to opt out.  Leave unset for the default, which may change."
+  - name: HETEROGENEOUS
+    default: "false"
+    documentation: "Adds additional workers of different architecture."
```
wking, line 33: This feels a bit redundant.  Or at least, there are multiple levels of knobs:

* `HETEROGENEOUS` is a big, generic knob.  I'd expect setting this true to give you a heterogeneous cluster without having to care about details like "how many heterogeneous compute?" or "which architecture should they run?".
* `ADDITIONAL_WORKER_ARCHITECTURE` is a mid-level knob, which can express more details than `HETEROGENEOUS`, but doesn't have to dip into anything cloud-specific (and also cannot control things like memory available to the heterogeneous nodes).
* `ADDITIONAL_WORKER_VM_TYPE` is a low-level knob, which gives folks full control, but needs cloud-specific knowledge (and possibly CI-account-quota-specific-knowledge) to be able to set it.

I'm not clear on whether we need all three layers of knobs, or how we avoid confusing people who set impossible combinations (e.g. `HETEROGENEOUS=false` while asking for ARM `ADDITIONAL_WORKER_ARCHITECTURE` and setting a s390x `ADDITIONAL_WORKER_VM_TYPE`).

ci-operator/step-registry/ipi/install/install/ipi-install-install-commands.sh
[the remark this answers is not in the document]
wking: can it move to using `upi-installer`?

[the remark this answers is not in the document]
deepsm007: I need to verify the contents of both the images and make sure its not missing any specific libs used by others. AFAIK `installer` does not have `yq-go`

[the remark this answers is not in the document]
wking: alternatively, can this move to an earlier step, and feed `${SHARED_DIR}/manifest_(manifest_name).yaml` as documented [here][1], with an example [here][2]?

[1]: https://github.com/openshift/release/blob/afcfe4d27a24c2a0133b0d6a2331b03e3fc9bef6/ci-operator/step-registry/ipi/install/install/ipi-install-install-ref.yaml#L37
[2]: https://github.com/openshift/release/blob/afcfe4d27a24c2a0133b0d6a2331b03e3fc9bef6/ci-operator/step-registry/ipi/conf/telemetry/ipi-conf-telemetry-commands.sh#L13

ci-operator/step-registry/ipi/install/install/ipi-install-install-ref.yaml
Prashanth684: and to add to this how would the user add multiple architectures in the future if they wanted to? should heterogeneous have its own configuration object?

Prashanth684: we could also do away with `ADDITIONAL_WORKER_VM_TYPE` by choosing a sane default ourselves

ci-operator/step-registry/openshift/e2e/test/openshift-e2e-test-commands.sh
aleskandro: Another alternative Deep and I were discussing is like:
```bash 
TESTS="$(openshift-tests run --dry-run --provider "${TEST_PROVIDER}" "${TEST_SUITE}")" &&
if [[ -n "${TEST_SKIPS}" ]]; then
    TESTS=$(echo "${TESTS}" | grep -v "${TEST_SKIPS}") &&
    EXCLUDED_TESTS=$(echo "${TESTS}" | grep "${TEST_SKIPS}") &&
    [[ -n "${EXCLUDED_TESTS}" ]] || { exit_code=$?; echo 'Error: no tests were found matching the TEST_SKIPS regex:'; echo "$TEST_SKIPS"; return $exit_code; } &&
    echo "Skipping tests:\n${EXCLUDED_TESTS}"
fi &&
if [ -f "${SHARED_DIR}/excluded_tests" ]; then
    TESTS=$(echo $TESTS | grep -v -F -f ${SHARED_DIR}/excluded_tests) &&
    EXCLUDED_TESTS="${EXCLUDED_TESTS}\n$(<${SHARED_DIR}/excluded_tests)" &&
    echo "Skipping tests:\n${EXCLUDED_TESTS}"
fi &&
if [[ -n "${EXCLUDED_TESTS}" ]]; then
    echo ${TESTS} > /tmp/tests &&
    TEST_ARGS="${TEST_ARGS:-} --file /tmp/tests"
fi &&
```

Regardless of the temporary work related to the heterogeneous clusters in this PR, it would honor both TEST_SKIPS and the excluded_tests file and would work for any user needing this behavior.

ci-operator/step-registry/ipi/install/install/ipi-install-install-ref.yaml
aleskandro: > I'm not clear on whether we need all three layers of knobs
> HETEROGENEOUS=false while asking for ARM ADDITIONAL_WORKER_ARCHITECTURE

`HETEROGENEOUS=false` currently leads to any `ADDITIONAL_*` var being ignored. 

> while asking for ARM ADDITIONAL_WORKER_ARCHITECTURE and setting a s390x 
 ADDITIONAL_WORKER_VM_TYPE

This would fail even for a user launching the installer with a wrong install-config. For instance, the following install-config (for an arm64 installation when the user sets an amd64 instance type) would not produce a validation error, leading the installation to fail later. 

```yaml
compute:
- architecture: arm64
  hyperthreading: Enabled
  name: worker
  platform:
    aws:
      type: m6a.xlarge
  replicas: 3
```

> and to add to this how would the user add multiple architectures in the future if they wanted to? should heterogeneous have its own configuration object?

Could we address this whenever we get the `compute` array in the install-config supporting heterogeneous architectures and multiple values? 

doing this now in this repo may mean that the additional workers are provisioned at a different time than the other original workers provided by the installer via the install-config.yaml. It leads to >15mins increased time before running the tests. By "hacking" the manifests created by the installer, we are actually doing an installation with heterogeneous workers at day-0 and this should be considered temporary IMO, until we get the installer supporting this conf.

> we could also do away with ADDITIONAL_WORKER_VM_TYPE by choosing a sane default ourselves based on arch

I was thinking at a solution like the SIZE_VARIANT we leverage, and also got this recent change that should/could be better addressed for arm64 jobs (currently, we set COMPUTE_NODE_TYPE at job config level rather than having a per-arch default): https://github.com/openshift/release/blob/f5463172571a2f32cda2dd51df3d4569d7f2415b/ci-operator/step-registry/ipi/conf/aws/ipi-conf-aws-commands.sh#L62-L72

IMO, beyond a possible default, we should consider a `sizing` capability for specific testing cases like the performance and scalability ones.

Prashanth684: we will get a validation error when we specify differing arches for compute and control plane. some of these hacks will go away with IPI support once we grow that organically in the installer but a separate step for heterogeneous still looks like the way to go.

ci-operator/step-registry/openshift/e2e/test/openshift-e2e-test-commands.sh
deepsm007: It will go away with IR manifest support for imagestream tags

ci-operator/step-registry/ipi/install/install/ipi-install-install-commands.sh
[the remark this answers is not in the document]
deepsm007: The image change will be handled in separate PR by confirming contents of both `installer` and `upi-installer` to not break CI

ci-operator/step-registry/ipi/install/install/ipi-install-install-commands.sh
```diff
@@ -300,6 +300,39 @@ echo "$(date +%s)" > "${SHARED_DIR}/TEST_TIME_INSTALL_START"
 openshift-install --dir="${dir}" create manifests &
 wait "$!"
 
+# Adds arm64 workers for heterogeneous deployment 
+# TO-DO Remove once we have IPI support for heterogeneous in 4.13 https://issues.redhat.com/browse/MIXEDARCH-143 
+if [ ! -z "$ADDITIONAL_WORKERS" ]; then
```
vrutkovs, line 305: This needs to be a new step. See https://github.com/openshift/release/blob/48c1af152e2ddee2e0b698212ef94ff337030fe8/ci-operator/step-registry/ingress/azure/custom-ingresscontroller/ingress-azure-custom-ingresscontroller-commands.sh#L44-L105 as an example of custom MachineSet and https://github.com/openshift/release/blob/826a02f24699747244b64524f93a9861f7e4ba60/ci-operator/step-registry/ipi/install/monitoringpvc/ipi-install-monitoringpvc-commands.sh for an example of mixing in additional manifest for installer

ci-operator/step-registry/ipi/install/heterogeneous/ipi-install-heterogeneous-commands.sh
```diff
@@ -0,0 +1,47 @@
...
+if ! command -v jq; then
+  # TODO move to image
+  curl -sL https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64 > /tmp/jq
+  chmod +x /tmp/jq
+  PATH=${PATH}:/tmp
+fi
+
+REGION=${LEASED_RESOURCE}
+
+echo "Creating arm worker machineset"
```
wking, line 29: Should this be:

```sh
echo "Creating ${ADDITIONAL_WORKER_ARCHITECTURE} worker MachineSet"
```

or some such?

ci-operator/step-registry/ipi/install/heterogeneous/ipi-install-heterogeneous-commands.sh
```diff
@@ -0,0 +1,47 @@
...
+| .spec.replicas = ${ADDITIONAL_WORKERS}
+| .spec.selector.matchLabels.\"machine.openshift.io/cluster-api-machineset\" = .metadata.name
+| .spec.template.metadata.labels.\"machine.openshift.io/cluster-api-machineset\" = .metadata.name" \
+${dir}/openshift/99_openshift-cluster-api_worker-machineset-0.yaml > ${dir}/openshift/99_openshift-cluster-api_worker-machineset-additional.yaml
+
+yq-v4 eval ".spec.template.spec.providerSpec.value.ami.id = \"${amiid_workers_additional}\"
+            | .spec.template.spec.providerSpec.value.instanceType = \"${ADDITIONAL_WORKER_VM_TYPE}\"
+            " -i ${dir}/openshift/99_openshift-cluster-api_worker-machineset-additional.yaml
+
+oc create -f ${dir}/openshift/99_openshift-cluster-api_worker-machineset-additional.yaml &
```
wking, line 44: for debugging purposes, you may want a:

```sh
cat "${dir}/openshift/99_openshift-cluster-api_worker-machineset-additional.yaml"
```

here.

Or use [`${ARTIFACT_DIR}`][1] as your scratch space, instead of `${dir}/openshift`?

[1]: https://docs.ci.openshift.org/docs/architecture/step-registry/#available-environment-variables

ci-operator/step-registry/ipi/install/heterogeneous/ipi-install-heterogeneous-commands.sh
```diff
@@ -0,0 +1,47 @@
...
+
+wait "$!"
+ret="$?"
```
wking, line 47: [Shellcheck][1]:

```
In .//ci-operator/step-registry/ipi/install/heterogeneous/ipi-install-heterogeneous-commands.sh line 47:
ret="$?"
^-^ SC2034: ret appears unused. Verify use (or export if used externally).
```

Makes sense to me.  Maybe just:

```sh
exec oc create -f ...
```

To get the shell out of the way for TERM signals?

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_release/31965/pull-ci-openshift-release-master-step-registry-shellcheck/1579982966887550976

ci-operator/step-registry/ipi/install/heterogeneous/ipi-install-heterogeneous-commands.sh
deepsm007: done

ci-operator/step-registry/ipi/install/heterogeneous/ipi-install-heterogeneous-commands.sh
deepsm007: Updated

ci-operator/step-registry/ipi/install/install/ipi-install-install-commands.sh
deepsm007: Updated

wking (approved):
/lgtm

PR #31971, opened 2022-09-03
Fixes the input images replacement for the initial layers of the fedora-coreos image

aleskandro:
Hi @LorbusChris, the `fedora-coreos` [image](https://github.com/openshift/release/pull/31663/files#diff-e0402ef5a86a7b685295c54436b6fa8565f460b069ad81c76d601c230132aeb9R28) is based on the processed [Dockerfile.template](https://github.com/openshift/okd-machine-os/blob/master/Dockerfile.template) from the `layering-build` image (Dockerfile.ci)... 

I see that `Dockerfile.template` essentially rebuilds the image `okd-rpms` in the third stage based on two layers (the artifacts and machine-config-operator tags) that are not replaced by  the prow config... 

IIUC, we may want to either change the lines 1-9 of the okd-machine-os' [Dockerfile.template](https://github.com/openshift/okd-machine-os/blob/master/Dockerfile.template) to be `FROM  registry.ci.openshift.org/origin/4.12:okd-rpms AS rpms` or the prow config for the `fedora-coreos` image as in this PR.  Or is it a wanted behavior?

Refers #31663

### DIFF
--- a/ci-operator/config/openshift/okd-machine-os/openshift-okd-machine-os-master.yaml
+++ b/ci-operator/config/openshift/okd-machine-os/openshift-okd-machine-os-master.yaml
@@ -25,13 +25,16 @@ images:
       - registry.ci.openshift.org/origin/4.12:machine-config-operator
   to: okd-rpms
 - inputs:
+    artifacts:
+      as:
+      - registry.ci.openshift.org/origin/4.12:artifacts
     layering-build:
       paths:
       - destination_dir: .
         source_path: /tmp/Dockerfile
-    okd-rpms:
+    mco:
       as:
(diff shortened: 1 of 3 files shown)


vrutkovs:
>I see that Dockerfile.template essentially rebuilds the image okd-rpms in the third stage based on two layers (the artifacts and machine-config-operator tags) that are not replaced by the prow config...

Right, we'll want to get rid of that

aleskandro:
Ok, so should I close this one?

vrutkovs:
Yeah, I think so. We want to extract MCO binary from `okd-rpms`, so that it would the single source of RPMs and binaries

aleskandro:
Ok, thanks @vrutkovs ... Closing

PR #31973, opened 2022-09-04
openshift-os: promote `openshift-os-src` image

vrutkovs:
This image contains just the source code of openshift/os, so that
okd-machine-os could reuse some overlays, instead of importing
openshift-os as a submodule.

Other images are not referring to this image, so it would not be included
in the payload


ci-operator/config/openshift/os/openshift-os-release-4.12.yaml
```diff
@@ -34,6 +38,17 @@ images:
       as:
       - registry.ci.openshift.org/coreos/coreos-assembler:latest
   to: build-image
+- dockerfile_literal: |
+    # just the source code
+    FROM registry.access.redhat.com/ubi9/ubi-minimal
+    WORKDIR /src
+    # Copy the source code
+    COPY . .
```
LorbusChris, line 46: hm I'm not sure I like the complexity of introducing another image for this, on the other hand having the the `os` git submodule in `okd-machine-os` isn't great either. If we go this way, local rebuilds of `okd-machine-os` will be trickier, as it might not be obvious how the `openshift-os-src` image is created.

Currently, all of this would be for just the `06gcp-routes` overlay. Alternatively we could carry a copy of the gcp-routes overlay in the 99okd overlay that would have to be updated manually (it's 3 files). Maybe there's even a github action to keep dirs/files in sync across repos and create a PR automatically.

vrutkovs: >Alternatively we could carry a copy of the gcp-routes overlay that would have to be updated manually (it's 3 files)

Right, that's an option, however more overlays may need to be carried (i.e. `dhcp-chrony`) and we won't get fixes for those automatically, so I'd prefer to get a copy of these overlays automatically. Submodules is a solution, but a fairly ugly one

cgwalters: This discussion overlaps some with bits of https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/OPSZ7I4JLFTDC65FPEOUN4CZJNEGKGDS/ - basically, perhaps we should be factoring these systems units into RPMs.  

Something like `openshift-node-common.rpm` or something?  Actually if we did this, it would also be a good place to take a stab at some of https://github.com/openshift/machine-config-operator/issues/3137 too.

PR #31986, opened 2022-09-05
HCO: explicitly set COMPUTE_NODE_TYPE on GCP

tiraboschi:
With https://github.com/openshift/release/pull/31754 ,
the default machine type on GCP moved from
n2-standard-4 to e2-standard-4 which does not
support nested virtualization.
Explicitly ask for n2-standard-4 in all the GCP
lanes to enable back the virtualization support.


nunnatsa (approved):
/approve

PR #31992, opened 2022-09-05
Adding missing Metal team leads for restricted labels

pamoedom:
We need Metal Dev & QE team members to also be part of `backport-risk-assessed` and `cherry-pick-approved` restricted labels for the origin repo.


core-services/prow/02_config/openshift/origin/_pluginconfig.yaml
```diff
@@ -12,6 +12,7 @@ label:
       - sttts
       - s-urbaniak
       - tkashem
+      - dtantsur
```
dtantsur, line 15: nit: ideally keep the list sorted

pamoedom: ok, good point, it seems that only "backport-risk-assessed" is sorted and not the "cherry-pick-approved", but let me fix it at least for the "backport-risk-assessed" one to keep it in order.

pamoedom: done

bparees:
I know there may be some concern that we'd really only want to give people permission to BPRA/CPA code that they are specifically involved in, but since we don't have that level of configurability, and given that the PRs still need approval from the more tightly constrained set of people who do have those permissions on specific code areas, i think this is ok.

I would hope that people who aren't actively involved in an area of code, would not be BPRA/CPA'ing those PRs.

There is also somewhat reduced risk here since BPRA/CPA was mostly created to more carefully constrain what is being backported into z-stream payloads that go out to customers.  While technically origin goes into the payload (in the form of the openshift-tests binary that customers can use), it is not going to directly impact a customer's cluster health/function.

/approve

PR #32023, opened 2022-09-06
Move baremetal 4.12 nightly jobs to build05

tdomnesc:
build05 can connect to the new internal hardware

https://issues.redhat.com/browse/METAL-331


derekhiggins:
/lgtm
Looks good to me, pass rates look similar to the equinix version

mandre (commented):
You may consider adding the BM team lead to https://github.com/openshift/release/blob/master/ci-operator/config/openshift/release/OWNERS.

/approve

dtantsur:
hi @mandre, is this a standard practice? I don't think other team leads are there, the list is quite light.

mandre:
> hi @mandre, is this a standard practice? I don't think other team leads are there, the list is quite light.

Good question. I believe that if your team manages a CI infra and you need to make frequent adjustment where jobs run, you should have someone from your team able to approve the changes. That's definitely the case for the OpenStack team.

PR #32027, opened 2022-09-06
Add config change trigger to telco CI image build

sshnaidm:


danilo-gemoli:
@jmguzik @droslean PR looks good but I'm not sure it will work together with our `applyconfig` tool.
According to [BuildConfig trigger doc](https://docs.openshift.com/container-platform/4.11/cicd/builds/triggering-builds-build-hooks.html#builds-configuration-change-triggers_triggering-builds-build-hooks) the `ConfigChange` trigger only works when the resource is created rather than when it is patched, and it is likely that `applyconfig` would patch it.

PR #32028, opened 2022-09-06
[telco5g] Run nightly info lane job on exact build tag

sshnaidm:


ci-operator/step-registry/telco-bastion/setup/telco-bastion-setup-commands.sh
```diff
@@ -25,6 +25,8 @@ set -x
 KCLI_PARAM=""
 if [ ! -z $OO_CHANNEL ] ; then
     KCLI_PARAM="-P openshift_image=registry.ci.openshift.org/ocp/release:$OO_CHANNEL"
+elif [[ "$PROW_JOB_ID" =~ "nightly" ]]; then
```
yuvalk, line 28: maybe add a comment here to document that this is how we distinguish a nightly from a periodic run

yuvalk:
/lgtm
but let's talk with testinfra to see if theres a better way to distinguish periodics from nightlies

PR #32029, opened 2022-09-06
supplemental-ci-images: Stop buildiong fcos-buildroot

cgwalters:
See https://github.com/coreos/fedora-coreos-config/issues/1950
and https://github.com/coreos/fedora-coreos-pipeline/pull/622

It's now being built by the FCOS pipeline.


dustymabe (approved):
/lgtm

PR #32034, opened 2022-09-06
Pin OCP version for HyperShift/Kubevirt presubmits to 4.11

orenc1:
OCP 4.12 have started to enforce Pod Security Profiles policies, which are not yet implemented in the last production release of CNV.
Pinning OCP to v4.11 to let CNV to complete installation successfully.

### DIFF
--- a/ci-operator/config/openshift/hypershift/openshift-hypershift-main.yaml
+++ b/ci-operator/config/openshift/hypershift/openshift-hypershift-main.yaml
@@ -75,6 +75,10 @@ releases:
       include_built_images: true
       name: "4.12"
       namespace: ocp
+  previous:
+    release:
+      channel: stable
+      version: "4.11"
 resources:
   '*':
     limits:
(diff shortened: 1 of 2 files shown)


ci-operator/config/openshift/hypershift/openshift-hypershift-main.yaml
```diff
@@ -165,7 +169,7 @@ tests:
       - env: HYPERSHIFT_RELEASE_LATEST
         name: hypershift-operator
       - env: OCP_IMAGE_LATEST
-        name: release:latest
+        name: release:previous
```
alvaroaleman, line 172: This means that the PR isn't tested anymore, previous doesn't `nclude_built_images: true`

orenc1: Thanks, fixed.
The goal is to have the infra cluster in openshift-ci deployed with ocp 4.11.
/cc @davidvossel

davidvossel (approved):
/lgtm
/approve

@orenc1 thanks for jumping on this.

For whoeven in hypershift that will review this, the context here is that we're having to revert the infra cluster back to 4.11 because CNV fails to install in 4.12 due to the recent pod security change. CNV is working on making their install work in 4.12, but until then we're using 4.11 to unblock hypershift/kubevirt development

alvaroaleman (approved):

PR #32039, opened 2022-09-06
NO-ISSUE: Add mainteners from assisted for packet-setup step

adriengentil:
Since metal team is focusing to transistion on OFCIR, we (assisted)
would like to become mainteners of the packet-setup step in order to
take care of the ansible packet flow.

### DIFF
--- a/ci-operator/step-registry/baremetalds/packet/OWNERS
+++ b/ci-operator/step-registry/baremetalds/packet/OWNERS
@@ -1,1 +1,9 @@
-../OWNERS
+approvers:
+- andfasano
+- ardaguclu
+- derekhiggins
+- elfosardo
+- honza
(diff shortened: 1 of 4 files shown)


ci-operator/step-registry/baremetalds/packet/setup/OWNERS
```diff
@@ -1 +0,0 @@
-../OWNERS
```
derekhiggins, line 1: I wonder if you should add the maintainers one level up? 
i.e. in ci-operator/step-registry/baremetalds/packet

adriengentil: it makes sense, I'll update

osherdp (approved):

PR #32045, opened 2022-09-06
[kubevirt] update ssp operator to use ocp 4.12

ksimon1:
update ssp operator to use ocp 4.12
Signed-off-by: Karel Šimon <ksimon@redhat.com>


ksimon1:
Kubevirt is not prepared for 4.12 yet

PR #32051, opened 2022-09-06
OSDe2e export report files using ci-operator

MrSantamaria:
After moving pipelines to use CI-Operator the REPORT_DIR needed to be set as $ARTIFACT_DIR in order to write OSDe2e's outputs. 
https://docs.ci.openshift.org/docs/how-tos/artifacts/#ci-operator-artifacts
https://issues.redhat.com/browse/SDCICD-797 

This change should also update Testgrid to display individual test inside OSDe2e.


MrSantamaria:
/tide/merge-method-squash

PR #32057, opened 2022-09-07
fix ingress LB type in workflow gcp-upi-private-xpn-ingress-glb

lihongan:
Found that the ingress LB `clientAccess: Global` was not set correctly in workflow gcp-upi-private-xpn-ingress-glb.
This PR is to fix the issue.

### DIFF
--- a/ci-operator/config/openshift/verification-tests/openshift-verification-tests-master__installer-rehearse-4.12.yaml
+++ b/ci-operator/config/openshift/verification-tests/openshift-verification-tests-master__installer-rehearse-4.12.yaml
@@ -70,7 +70,7 @@ tests:
   interval: 960h
   steps:
     cluster_profile: gcp-qe
-    workflow: cucushift-installer-rehearse-gcp-upi-private-xpn
+    workflow: cucushift-installer-rehearse-gcp-upi-private-xpn-ingress-glb
 - as: installer-rehearse-vsphere
   interval: 960h
   steps:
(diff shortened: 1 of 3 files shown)


lihongan:
@jianli-wei @jianlinliu this one also keeps failing from yesterday.

lihongan:
Let me test with installer-rehearse-4.12.

lihongan:
/hold
LB is still incorrect

lihongan:
@jianlinliu I checked the `ref: upi-conf-gcp` and seems it creats manifest then creats Ignition configs, no chance to update the manifest, any suggestion?
Or should we use ENV and add my manifest in the ref ?

jianlinliu:
> @jianlinliu I checked the `ref: upi-conf-gcp` and seems it creats manifest then creats Ignition configs, no chance to update the manifest, any suggestion? Or should we use ENV and add my manifest in the ref ?

Do you mean there is no place to inject your create manifest before creating ignition configs, right? 

Compared with ipi-install-install ref, it has such logic in https://github.com/openshift/release/blob/master/ci-operator/step-registry/ipi/install/install/ipi-install-install-commands.sh#L315-L331.

So my suggestion is:
1. If your feature does not care bout IPI or UPI, cover it in IPI.
2. If have to cover it in UPI, suggest to update upi-conf-{gcp,azure,aws}-ref using similar logic as ipi-install-install, instead of introduce a new ENV.

lihongan:
@jianlinliu The logic in ipi-install-install ref looks like good solution. Thanks for the info.
Since this LB type is meaningful only with Private XPN cluster, I'll try the 2nd suggestion later.

lihongan:
The LB is configured correctly this time, but from the upi-conf-gcp build-log seems it also injected another `manifest_cluster-monitoring-config.yaml`, see below:
```
Will include manifests:
/tmp/secret/manifest_cluster-monitoring-config.yaml
/tmp/secret/manifest_cluster-ingress-default-ingresscontroller.yaml
Creating Ignition configs...
```
Not sure it that's expected? The manifest is created in `ipi-install-monitoringpvc`.

jianlinliu:
> seems it also injected another manifest_cluster-monitoring-config.yaml

Yeah, it is expected, because `ipi-conf-gcp` chain includes `ipi-install-monitoringpvc` ref, the manifest_cluster-monitoring-config.yaml was created from that step. Your PR are fixing another similar issues together.

PR #32060, opened 2022-09-07
add 4.12 e2e test for azure-ipi-private-sdn and gcp-upi-private-xpn-sdn

lihongan:


lihongan:
/assign @liangxia 
looks most cases are passed in both tests

PR #32065, opened 2022-09-07
Fix incorrect exit value

yunjiang29:
* Added `set -o errexit`

Allow the script to handle errors automatically, except `openshift-install create cluster`
 
* Added `set -o pipefail`
```
$ grep some-string /non/existent/file | sort
grep: /non/existent/file: No such file or directory % echo $?
0
```
```
$ set -o pipefail
$ grep some-string /non/existent/file | sort
grep: /non/existent/file: No such file or directory $ echo $?
2
```


yunjiang29:
@jianlinliu please help review, thanks

jianlinliu:
looks good to me, I am waiting for the verify job get completed.

PR #32078, opened 2022-09-07
Removes cos-85-lts integration tests

Stringy:
As per this page: https://cloud.google.com/container-optimized-os/docs/release-notes , cos-85-lts is no longer supported, and the images have been removed from the GCP project. This PR removes our integration tests that run on that platform


Molter73 (approved):
/lgtm

PR #32081, opened 2022-09-07
NO-ISSUE: test e2e-assisted-metal with m3.small

adriengentil:
Let's test this setting in test-infra for some time to see if it works well.

`m3.small` has less CPU than `c3.medium` (8 vs. 24) and the same amount of memory (64GB).
It costs 1.05$ vs. 1.50$ for `c3.medium`.

/cc @osherdp @eliorerz

### DIFF
--- a/ci-operator/step-registry/assisted/common/setup/prepare/assisted-common-setup-prepare-commands.sh
+++ b/ci-operator/step-registry/assisted/common/setup/prepare/assisted-common-setup-prepare-commands.sh
@@ -253,21 +253,24 @@ cat > run_test_playbook.yaml <<-"EOF"
           mkfs.xfs -f /dev/nvme0n1
           mount /dev/nvme0n1 {{ REPO_DIR }}
         when: "equinix_metadata.json.plan == 'm3.large.x86'"
-      - name: Setup working directory and swap for machine type c3.medium.x86
+      - name: Setup extra swap for machine type {{ equinix_metadata.json.plan }}
         ansible.builtin.shell: |
-          # c3.medium.x86 has 64GB of RAM which is not enough for most of assisted jobs.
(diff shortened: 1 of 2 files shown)


ci-operator/step-registry/assisted/common/setup/prepare/assisted-common-setup-prepare-commands.sh
```diff
@@ -253,21 +253,24 @@ cat > run_test_playbook.yaml <<-"EOF"
...
           # Get disk where / is mounted
           ROOT_DISK=$(lsblk -o pkname --noheadings --path | grep -E "^\S+" | sort | uniq)
 
-          # Setup the smallest disk available (240GB) as swap
+          # Setup the smallest disk available as swap
           SWAP_DISK=$(lsblk -o name --noheadings --sort size --path | grep -v "${ROOT_DISK}" | head -n1)
           mkswap "${SWAP_DISK}"
           swapon "${SWAP_DISK}"
-        when: "equinix_metadata.json.plan == 'c3.medium.x86'"
+        when: "equinix_metadata.json.plan == 'c3.medium.x86' or equinix_metadata.json.plan == 'm3.small.x86'"
```
osherdp, line 273: not critical for now, but maybe we should use the equinix metadata spec entry (which is kind of a host inventory) instead of doing those conditions for certain plans

osherdp (approved):

ci-operator/step-registry/assisted/common/setup/prepare/assisted-common-setup-prepare-commands.sh
adriengentil: well, I think we now cover all the machine types :D ... but yeah it would be more dynamic

PR #32087, opened 2022-09-07
medik8s/nmo: use a ns != default for operator deployment

slintes:
Pod security admission related settings don't work in default which causes deployment to fail.


slintes:
/hold

wait for rehearsal

slintes:
unrelated flaky test, the deployment issue is resolved by using another namespace

beekhof:
/lgtm
I'm assuming this PR is needed to fix the failing jobs?

slintes:
> I'm assuming this PR is needed to fix the failing jobs?

yes, OLM does not work in the default namespace anymore because of pod security admission, so operator deployment failed. Thx for the lgtm

PR #32090, opened 2022-09-07
add the sdn version of the azure-techpreview workflow

jluhrsen:
this slipped through the cracks here [0] and without calling the specific sdn workflow, the job was using ovn but having -sdn- in it's name. This causes a
hard failure in the test case checking on that. fix it.

[0] https://github.com/openshift/release/commit/ee439d96f0ccb0e4cf5449f21544a57916f06f69


kikisdeliveryservice:
ahhh thanks for this @jluhrsen !

vrutkovs (approved):
/lgtm

PR #32107, opened 2022-09-08
correct installer version in upi-conf-gcp

jianlinliu:
1. correct installer version in upi-conf-gcp to avoid mismatched behavior between installer and payload
2. enable qe-catalogsource in some workflows


jianlinliu:
@gpei @jianli-wei pls help review

PR #32108, opened 2022-09-08
Fix single-node periodics reporter configs

omertuc:
Adds reporter config for some SNO jobs that didn't have one,
and removed unused Release Status link from the message on all of them.


eranco74 (commented):
/lgtm

vrutkovs:
/approve
/hold

We might want to bump SNO version as it was devpreview in 4.8

PR #32115, opened 2022-09-08
Bumping the consumed nightly to 4.12.

ybettan:
DTK is running some e2e test on every PR but the DTK image used is the image from nightly builds, therefore, we need to make sure we consume the up-to-date nightly builds.


ybettan:
/hold
Waiting for all jobs to pass

PR #32117, opened 2022-09-08
Add HyperShift jobs to storage components

jsafrane:
AWS EBS CSI driver operator + cluster-storage-operator + snapshot operators are moving to HyperShift, so let's make sure we don't break anything there.

@openshift/storage


bertinatto (commented):
/lgtm

PR #32120, opened 2022-09-08
Fixing missing yq libs from installer images

deepsm007:
Recent change of removing yq from libvirt-installer and installer image is breaking s390x/ppc64le CI jobs


r4f4:
Anyway I guess a better approach is to add `yq` to the proper images instead of downloading it for every job run and getting possible rate limiting issues.

PR #32121, opened 2022-09-08
Updated acs-fleet-manager e2e test resources

johannes94:
Update resources because rhacs-operator runs into OOM error.

### DIFF
--- a/ci-operator/config/stackrox/acs-fleet-manager/stackrox-acs-fleet-manager-main.yaml
+++ b/ci-operator/config/stackrox/acs-fleet-manager/stackrox-acs-fleet-manager-main.yaml
@@ -39,9 +39,6 @@ resources:
   '*':
     requests:
       cpu: 2000m
-      # TODO(09/08/2022): The memory was set to 8000Mi because the rhacs-operator had OOM issues.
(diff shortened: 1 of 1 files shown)


ci-operator/config/stackrox/acs-fleet-manager/stackrox-acs-fleet-manager-main.yaml
```diff
@@ -39,7 +39,7 @@ resources:
   '*':
     requests:
       cpu: 2000m
-      memory: 6000Mi
+      memory: 8000Mi
```
ebensh, line 42: Add a TODO with date applied, date you think it will no longer apply (or a release version that should go out when it can be reverted/revisited), and a tiny bit of context. Otherwise we'll never come back and request 8 GiB forever.

ci-operator/config/stackrox/acs-fleet-manager/stackrox-acs-fleet-manager-main.yaml
```diff
@@ -71,7 +71,7 @@ tests:
       resources:
         requests:
           cpu: 2000m
-          memory: 6000Mi
+          memory: 8000Mi
```
ebensh, line 74: Same here.

ebensh (approved):

PR #32043, opened 2022-09-06
stackrox: avoid image copies for reqd jobs

gavin-stackrox:
@janisz, FYI, I _think_ this change will make some of the required CI jobs faster on stackrox PRs as it will skip some of the image copying that seems to bog OpenShift CI down. But I'm a little concerned that this will introduce 429s as it adds more source cloning. So I'll need to dig into the logs/code to see how this functions before committing to it. And run this 10x to see if there is _any_ performance boost. 

/hold
/cc @janisz 

open the gate: go-unit-tests go-unit-tests-release go-postgres-tests

PR #32126, opened 2022-09-08
stackrox: retry #32043

gavin-stackrox:
This PR uses the `ci-perf-hacks` branch to retry #32043 in a controlled environment where the real PR lifecycle can be tested. (And thus avoid the breakage introduced by #32043).
/cc @house-d @sbostick @janisz


sbostick (approved):
LGTM

PR #32128, opened 2022-09-08
Microshift generic-informing jobs

copejon:
[USHIFT-404](https://issues.redhat.com//browse/USHIFT-404)

### DIFF
--- a/core-services/testgrid-config-generator/_allow-list.yaml
+++ b/core-services/testgrid-config-generator/_allow-list.yaml
@@ -36,5 +36,5 @@ periodic-ci-openshift-release-master-nightly-4.10-e2e-azure-csi-migration: broke
 periodic-ci-openshift-release-master-nightly-4.10-e2e-gcp-csi-migration: broken
 periodic-ci-openshift-release-master-nightly-4.11-e2e-aws-ovn-serial: informing
 release-openshift-origin-installer-e2e-azure-compact-4.4: broken
-periodic-ci-openshift-microshift-main-periodics-e2e-openshift-conformance-parallel: informing
-periodic-ci-openshift-microshift-main-periodics-e2e-openshift-conformance-serial: informing
+periodic-ci-openshift-microshift-main-periodics-e2e-openshift-conformance-parallel: generic-informing
+periodic-ci-openshift-microshift-main-periodics-e2e-openshift-conformance-serial: generic-informing


core-services/testgrid-config-generator/_allow-list.yaml
```diff
@@ -36,20 +36,20 @@ periodic-ci-openshift-release-master-nightly-4.10-e2e-azure-csi-migration: broke
...
-periodic-ci-openshift-microshift-main-periodics-e2e-openshift-conformance-sig-node: informing
-periodic-ci-openshift-microshift-main-periodics-e2e-openshift-conformance-sig-scheduling: informing
-periodic-ci-openshift-microshift-main-periodics-e2e-openshift-conformance-sig-storage: informing
-periodic-ci-openshift-microshift-main-periodics-e2e-reboot: informing
-periodic-ci-openshift-microshift-main-periodics-rebase-on-nightlies: informing
-periodic-ci-openshift-microshift-main-periodics-test-srpm: informing
-periodic-ci-openshift-microshift-main-periodics-test-unit: informing
-periodic-ci-openshift-microshift-main-periodics-e2e-openshift-conformance-parallel: broken
-periodic-ci-openshift-microshift-main-periodics-e2e-openshift-conformance-serial: broken
+periodic-ci-openshift-microshift-main-periodics-copr-rpm-health-check: generic-informing
```
deads2k, line 39: why generic-informing instead of informing?

deads2k, line 39: this file doesn't have many entries in it.  How are other jobs included?

copejon: The microshift test grid entries hadn't appeared yet and a look at the logs hinted at [why](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-prow-auto-testgrid-generator/1567664205102845952#1:build-log.txt%3A10-26).  The `unrecognized release type in job` message is generated by the testgrid-config-generator [L235-L261](https://github.com/openshift/ci-tools/blob/6fb323030b93edd8979c4863b697bba637760a0b/cmd/testgrid-config-generator/main.go#L235-L261).  If we want to rely on openshift-ci automation to generate the testgrid tabs for microshift, it'll need to be `generic-informining`.   Generic informing jobs can be found at https://testgrid.k8s.io/redhat-openshift-informing.

An alternative would be to alter the generator to create a separate tab for microshift.

Else, we can do what it looks like SNO and other projects have done: write/generate a config manually and open a PR directly into upstream testgrid, [here](https://github.com/kubernetes/test-infra/tree/master/config/testgrids/openshift).

I've not dealt with the testgrid before, so I'm open to whichever approach you'd recommend.

copejon: No idea.  The approach here is what's recommended by CI [docs](https://docs.ci.openshift.org/docs/how-tos/add-jobs-to-testgrid/#adding-a-non-release-gating-job-to-testgrid).

deads2k: We need to appear in https://testgrid.k8s.io/redhat-openshift-ocp-release-4.12-informing.  If that is gong to require a manual config, then that will have to be the way.

copejon: A better answer: openshift-ci runs a periodic testgrid config generator job that scans jobs, produces configs, and opens PRs into test-infra.  The tool is geared to place CI configs with ocp, origin, or release in the name are considered informing.  CI configs without those names default to generic-informing.

The bright side is that test-grid dashboard tabs can exist on more than one dashboard.  Once we merge this PR, we can open a PR upstream to copy the generated dashboard into the ocp-release-4.12-informing dashboard.

dhellmann: Would it make sense to either add 'microshift' to the list of strings that tool looks for or add one of the existing strings to our job names? Maybe 'origin', since that's the more general "upstream" indicator, IIRC.

If we follow the current approach, would we have to have similar PRs to add our jobs to the 4.13 dashboard and others in the future?

copejon: Unfortunately no.  So the tool looks for those substrings in test names and generates dashboard names from those.  The catch is in naming. There's a tight coupling, via this naming convention, that is present throughout the automation.  The test names must be in the pattern of `org-repo-branch-test-name`, and this pattern is passed through each layer of the config generators, from CI config -> Prow config -> TestGrid config). It makes it easy to ID at a glance what code base each job is testing.

But there is hope. First, tests can be represented on more than one dashboard at once.  It's just a matter of adding the dashboard-tab config to whichever dashboard _by hand._  So, if we tag our tests as generic-informing, they'll have a dashboard generated at https://testgrid.k8s.io/redhat-openshift-informing.  Then we open a PR in test-infra that just copies the generated dashboard-tabs into the release-ocp-4.12-informing dashboard.

I've gone around and around with the CI maintainers, and AFAICT, this looks like the least hacky workaround.

hongkailiu (approved):

dhellmann:
I see an approval from several hours ago, but no label. What is likely to have caused that?

PR #32130, opened 2022-09-08
Reorg ofcir calls

derekhiggins:
Now that the internal ofcir has cihosts, re-organise the ofcir calls to mke it more clear whats happening.


derekhiggins:
cc @tdomnesc

PR #32131, opened 2022-09-08
Fix env variables for hypershift cluster-bot

enxebre:
This was an attempt to set the env variables for cluster-bot hypershift launch https://github.com/openshift/release/pull/32106, however they are still not set.

This PR tries to set them in the right place (openshift-release-infra-periodics.yaml?) and also rename to prefix with HYPERSHIFT_ to avoid potential clashes


enxebre:
cc @bradmwilliams ptal

PR #32134, opened 2022-09-08
Revert "don't run expensive etcd tests on doc changes"

tjungblu:
Reverts openshift/release#31909 to see whether this made all our jobs optional somehow


tjungblu:
fixed with #32150

PR #32141, opened 2022-09-08
Add prow jobs for branches of integreatly-operator and cloud-resource-operator

carlkyrillos:
Adding prow checks for [this branch](https://github.com/integr8ly/integreatly-operator/tree/mgdapi-3425-gcp) in integreatly-operator and [this branch](https://github.com/integr8ly/cloud-resource-operator/tree/mgdapi-3425-gcp) in cloud-resource-operator

### DIFF
--- a/core-services/prow/02_config/_plugins.yaml
+++ b/core-services/prow/02_config/_plugins.yaml
@@ -824,6 +824,12 @@ config_updater:
         - ci
       gzip: true
       name: ci-operator-misc-configs
+    ci-operator/config/integr8ly/cloud-resource-operator/integr8ly-cloud-resource-operator-mgdapi-*.yaml:
+      clusters:
+        app.ci:
+        - ci
+      gzip: true
+      name: ci-operator-misc-configs
     ci-operator/config/integr8ly/integreatly-operator/integr8ly-integreatly-operator-*-next-*.yaml:
       clusters:
         app.ci:
@@ -1623,6 +1629,12 @@ config_updater:
       gzip: true
       name: job-config-misc
     ci-operator/jobs/integr8ly/cloud-resource-operator/integr8ly-cloud-resource-operator-MGDAPI-*.yaml:
+      clusters:
+        app.ci:
+        - ci
+      gzip: true
+      name: job-config-misc
+    ci-operator/jobs/integr8ly/cloud-resource-operator/integr8ly-cloud-resource-operator-mgdapi-*.yaml:
       clusters:
         app.ci:
         - ci


ci-operator/jobs/integr8ly/cloud-resource-operator/integr8ly-cloud-resource-operator-MGDAPI-3425-GCP-presubmits.yaml
```diff
@@ -0,0 +1,317 @@
...
+          secretName: registry-pull-credentials
+      - name: result-aggregator
+        secret:
+          secretName: result-aggregator
+    trigger: (?m)^/test( | .* )coverage,?($|\s.*)
+  - agent: kubernetes
+    always_run: true
+    branches:
+    - ^MGDAPI-3425-GCP$
+    - ^MGDAPI-3425-GCP-
```
KevFan, line 63:
```suggestion
    - ^mgdapi-3425-gcp$
    - ^mgdapi-3425-gcp-
```
Kinda minor but I think using lower case branch would be better if possible :thinking: 

I think there will be issues with the `images` prow check when the final feature branch tries to merge to master, which required it to be overriden. (Happened for STS feature branch before - https://github.com/integr8ly/cloud-resource-operator/pull/315#issuecomment-1135925720 and guide also uses a lower case feature branch name - https://github.com/RHCloudServices/integreatly-help/blob/master/guides/branching-operator.md)

carlkyrillos: Sounds good I'll make the change

carlkyrillos: @KevFan Actually will I be able to do that because the branch names are uppercase `MGDAPI`

KevFan: @carlkyrillos No, I think lowercased branch names will need to be created by the maintainers before making this change :thinking:

carlkyrillos: @laurafitzgerald CCing you about the branch name change. Does that sounds good to you? If so I can get started on renaming(recreating) the [GCP branch](https://github.com/integr8ly/integreatly-operator/tree/MGDAPI-3425-GCP) to be all lowercase.

KevFan:
@carlkyrillos Looking at the job failure and documentation, it looks like it is failing due a missing entry to the plugin configuration for the lowerd cased CRO feature branches 🤔 There is only an entry for the upper case feature branches for CRO:
* https://github.com/openshift/release/blob/master/core-services/prow/02_config/_plugins.yaml#L816-L821

But for integreatly-operator it has the config for both upper and lower feature branches, which is why is it these files are not mentioned in the prow error output:
* https://github.com/openshift/release/blob/master/core-services/prow/02_config/_plugins.yaml#L833-L844

carlkyrillos:
@wking Could you take a look at this PR when you get a chance?

laurafitzgerald:
@carlkyrillos you can also try slack channel `forum-testplatform` and use the handle `dptp-helpdesk` to try to get this one mereged.

PR #32143, opened 2022-09-08
Stackrox release 3.72

parametalol:
CI configuration files to support Stackrox release 3.72.

### DIFF
--- a/ci-operator/config/stackrox/stackrox/stackrox-stackrox-release-3.72.yaml
+++ b/ci-operator/config/stackrox/stackrox/stackrox-stackrox-release-3.72.yaml
@@ -1,16 +1,24 @@
 base_images:
+  aro:
+    name: automation-flavors
+    namespace: stackrox
+    tag: aro-0.3.8
   eks:
     name: automation-flavors
     namespace: stackrox
     tag: eks-0.3.1
-  golang-1.17.2:
+  golang-1.17.12:
     name: golang
     namespace: stackrox
-    tag: 1.17.2
+    tag: 1.17.12
   osd:
     name: automation-flavors
     namespace: stackrox
     tag: osd-0.3.6
+  rosa:
+    name: automation-flavors
+    namespace: stackrox
+    tag: rosa-0.3.8
   ubi:
     name: ubi
     namespace: ocp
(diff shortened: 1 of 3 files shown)


ci-operator/config/stackrox/stackrox/stackrox-stackrox-release-3.72.yaml
```diff
@@ -0,0 +1,367 @@
+base_images:
+  eks:
+    name: automation-flavors
+    namespace: stackrox
+    tag: eks-0.3.1
+  golang-1.17.2:
+    name: golang
+    namespace: stackrox
+    tag: 1.17.2
```
vikin91, line 9: Isn't this a bit too old? We run on 1.17.12 or 13 now, right?

0x656b694d: That's the version used in nightlies. Do we want to bump it now for the release?

vikin91: Just spotted this, so not sure how much it affects the release. Nightlies on the release branch sound strange. Isn't it the builder that would build the upstream released version?

vikin91: Let's maybe ask @gavin-stackrox what this version mens to learn for the future.

The PR is good to be merged for me 👍

ci-operator/config/stackrox/stackrox/stackrox-stackrox-release-3.72.yaml
```diff
@@ -0,0 +1,367 @@
+base_images:
+  eks:
+    name: automation-flavors
+    namespace: stackrox
+    tag: eks-0.3.1
+  golang-1.17.2:
```
gavin-stackrox, line 6: This needs to be `1.17.12` to match operator/Dockerfile on the target branch. (and in the 3 other places in this file). 
Marcin recently updated it, so you may have used a slightly stale openshift/release...master.yaml as the template.

gavin-stackrox (changes requested):

PR #32158, opened 2022-09-09
Remove and mv nmstate jobs

derekhiggins:
Shrinking the CI usage footprint by keeping the most important jobs.

Then moving the remaining jobs to run on the internal HW.


creydr (approved):
/lgtm

PR #32163, opened 2022-09-09
Move baremetal 4.11 nightly jobs to build05

tdomnesc:
so they run on internal lab hardware

https://issues.redhat.com/browse/METAL-331


dgoodwin:
What is your take on all the failed metal job rehersals? There's a variety of failures in there, no consistent pattern but it does look like these may not pass on the new hardware.

tdomnesc:
> What is your take on all the failed metal job rehersals? There's a variety of failures in there, no consistent pattern but it does look like these may not pass on the new hardware.

I have retested this over and over to compare to the current success rate, and it is identical. The failures are not because of the new hardware.

PR #32168, opened 2022-09-09
Assisted-service: add vsphere UMN optional job

vrutkovs:
This sets up vsphere Assisted cluster with user managed networking enabled


vrutkovs:
/cc @eliorerz 

Seems this parameter is not being passed correctly to -test-infra. Removed periodics so that we could test -test-infra PR to fix it and add periodics later on

eliorerz (approved):
LGTM

PR #31193, opened 2022-08-09
Fix issue DPTP-3048 Extend periodic intervals

bear-redhat:
`find ../release/ci-operator/config/ ../release/ci-operator/jobs/ -name "*.yaml" -exec python ./is-version-lower-than/main.py {} \;`
`make update`

Ref: https://github.com/bear-redhat/tools/blob/4054d781c5b949bd894c6abeda9b30412793ade3/is-version-lower-than/main.py

PR #32190, opened 2022-09-12
Fix up  ipi-packet-check frequency

derekhiggins:
This was recently changed as part of an effort
to blanket reduce the frequency of various jobs.
But only running it twice a month means leaked hosts don't get cleaned up for weeks.

Set LEAK_THRESHOLD_SECONDS to 5 hours as the default (4 hours) isn't long enough for upgrade jobs that are set to have a 5 hour deadline.

Also remove the cleanup job for the old profile that no longer exists.


derekhiggins:
cc @bear-redhat @jmguzik 
Undoing changes done in https://github.com/openshift/release/pull/31193 and https://github.com/openshift/release/pull/32083

PR #32125, opened 2022-09-08
MGMT-11925: Align common workflow with the old baremetal step

adriengentil:
Few variables were not set properly, also with the common workflow, the old `MAKEFILE_TARGET` was split in 2:
- `MAKEFILE_SETUP_TARGET` is run in `assisted-common-setup-prepare`
  step
- `MAKEFILE_TARGET` is run in `assisted-common-test` step

PR #32207, opened 2022-09-12
Restore assisted-image-service in openshift-installer

adriengentil:
`assisted-image-service` is required by the baremetal-workflow.

x-ref: https://github.com/openshift/release/pull/32125


r4f4 (commented):
/lgtm

PR #32222, opened 2022-09-12
add crun and cgroupv2 variants to cri-o CI

sohankunkerkar:
So the integration tests are still failing. I will add those in a separate PR once we figure out the actual issue.

### DIFF
--- a/ci-operator/step-registry/cri-o/crun/e2e/test/cri-o-crun-e2e-test-commands.sh
+++ b/ci-operator/step-registry/cri-o/crun/e2e/test/cri-o-crun-e2e-test-commands.sh
@@ -1,0 +1,63 @@
+#!/bin/bash
+set -o nounset
+set -o errexit
+set -o pipefail
+
+# shellcheck source=/dev/null
+source "${SHARED_DIR}/env"
+
+#####################################
+###############Log In################
+#####################################
+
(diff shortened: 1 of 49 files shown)


smg247:
@sohankunkerkar, I know the bot added me, but I am not the right person to review this one. Try adding some folks from https://github.com/openshift/release/blob/master/ci-operator/config/cri-o/cri-o/OWNERS :)

/uncc

ci-operator/step-registry/cri-o/crun/e2e/test/cri-o-crun-e2e-test-commands.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
```
haircommander, line 1: this can be done as a follow up, but is there anyway to share this bash code and replace variables? there's a decent amount of duplication

haircommander:
/approve

as a follow up, I think we should look into deduplicating the code as much as possible

ci-operator/step-registry/cri-o/crun/e2e/test/cri-o-crun-e2e-test-commands.sh
sohankunkerkar: Agree, I was thinking of handling that as part of the code optimization once we include all variants.

rphillips:
/lgtm

+1 to refactoring to share code

PR #32226, opened 2022-09-13
launch a podman container to support podman command

xinredhat:
To add support for Podman commands, this PR launches a Podman container to provide podman commands.
1. launch a Podman container with a yaml file
2. install all required softwares in the Podman container
3.  ckcp setup is executed in the Podman container

### DIFF
--- a/ci-operator/config/openshift-pipelines/pipeline-service/openshift-pipelines-pipeline-service-main.yaml
+++ b/ci-operator/config/openshift-pipelines/pipeline-service/openshift-pipelines-pipeline-service-main.yaml
@@ -12,6 +12,17 @@ build_root:
     name: release
     namespace: openshift
     tag: golang-1.18
+images:
+- dockerfile_literal: |-
+    FROM src
+    RUN pip3 install --upgrade setuptools pip && pip3 install yamllint && yamllint --version && \
+        SHELLCHECK_VERSION=0.7.1 && \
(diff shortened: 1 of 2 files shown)


ci-operator/config/openshift-pipelines/pipeline-service/openshift-pipelines-pipeline-service-main.yaml
```diff
@@ -56,22 +47,87 @@ tests:
...
+
+        export PATH=/usr/local/go/bin:$PATH
+        export PATH=$(go env GOPATH)/bin:$PATH
         HOME_DIR="$(pwd)"
         echo "Start building kcp plugin ..."
-        kcp_version="$(yq '.images[] | select(.name == "kcp") | .newTag' "ckcp/openshift/overlays/dev/kustomization.yaml")"
-        echo "kcp version:" $kcp_version
-        git clone https://github.com/kcp-dev/kcp.git -b $kcp_version
-        cd $HOME_DIR/kcp && go mod vendor && make install
+        KCP_TAG="$(yq '.images[] | select(.name == "kcp") | .newTag' "ckcp/openshift/overlays/dev/kustomization.yaml")"
```
bnallapeta, line 108: Although this will work, we recently made the change to move config [here](https://github.com/openshift-pipelines/pipeline-service/blob/main/config/config.yaml). This should be the source when we are fetching the version. See [this](https://github.com/openshift-pipelines/pipeline-service/blob/main/ckcp/openshift_dev_setup.sh#L143) for example.
I am okay if you make this change in a future PR too.

bnallapeta (approved):
/lgtm

PR #31237, opened 2022-08-10
ci-operator/config/openshift: Decouple Cincinnati ecosystem from OpenShift branching

wking:
graph-data only has a development branch.  Cincinnati and its operator recently pivoted to a similar model, [releasing an OpenShift Update Service 5.0 on 2022-08-03][0] based on the master branches.  This commit rips out all the old release-* stuff, and also:

* Drops `releases` in favor of cluster claims [1], because we are decoupled from the OpenShift core and do not need to install a fresh cluster for our testing.

* Drops `promotion`.  We'd already had most of our images in `exclude_images`, so not much change in terms of actually-promoted images.  But dropping the promotion property [opts us out of OpenShift branching][2].

[0]: https://access.redhat.com/errata/RHEA-2022:5871
[1]: https://docs.ci.openshift.org/docs/architecture/ci-operator/#testing-with-a-cluster-from-a-cluster-pool
[2]: https://docs.ci.openshift.org/docs/architecture/branching/#how-do-i-opt-my-repository-into-automated-branching

PR #32230, opened 2022-09-13
Remove cincinnati-operator configs from o-priv

jmguzik:
Since merging: https://github.com/openshift/release/pull/31237 our automation: https://github.com/openshift/release/pull/31573 fails on trying to partly delete cincinnati-operator from prow config.

My proposition is to delete fully this configuration from openshift-priv (which is the reason of failure) as files that caused cincinnati-operator configs to be reconciled were removed in https://github.com/openshift/release/pull/31237 

/cc @hongkailiu @wking 
@wking I am cc-ing also you as you are the author of discussed PR


jmguzik:
@hongkailiu @wking I have asked Danilo to LGTM this to enable our automation. If there are serious concerns against doing that we will revert.

PR #32236, opened 2022-09-13
Add secrets to enable pull-through cache to quay.io

jmguzik:

### DIFF
--- a/core-services/ci-secret-bootstrap/_config.yaml
+++ b/core-services/ci-secret-bootstrap/_config.yaml
@@ -3788,6 +3788,23 @@ secret_configs:
     - non_app_ci
     name: cloudfront-buildfarm-internal-registry-private-key
     namespace: openshift-image-registry
+- from:
+    AWS_ACCESS_KEY_ID:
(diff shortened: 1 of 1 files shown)


clusters/build-clusters/build03/quayio-pull-through-cache/quayio-pull-through-cache.yaml
```diff
@@ -0,0 +1,68 @@
...
+      matchLabels:
+        app: quayio-pull-through-cache
+    template:
+      metadata:
+        labels:
+          app: quayio-pull-through-cache
+      spec:
+        containers:
+        - name: quayio-cache
+          image: docker.io/registry:2.8.1
```
droslean, line 35: This will create a problem since docker.io has a `pull` limit request. We could push it in our quay.io repo and use that one instead.

droslean (commented):
It would be better to introduce this service in two different PRs. 
- A PR that adds the secret and wait for the bootstrapper job to create it
- A PR that introduces the service and the deployment plus the required alerts

Also, do the deployment labels satisfy the Grafana dashboards we have?

PR #32238, opened 2022-09-13
OCPBUGS-1235: openstack: remove workaround to deploy sriov network operator

EmilienM:
We previously had a workaround which would change the
`deploy/configmap.yaml` to contain the one in `manifests/stable` so
supported nics could be tested in our CI.

However `deploy/configmap.yaml` has everything we need now, so we don't
need that workaround.

### DIFF
--- a/ci-operator/step-registry/openshift/e2e/openstack/sriov/openshift-e2e-openstack-sriov-workflow.yaml
+++ b/ci-operator/step-registry/openshift/e2e/openstack/sriov/openshift-e2e-openstack-sriov-workflow.yaml
@@ -11,6 +11,7 @@ workflow:
       - ref: optional-operators-subscribe
       - ref: openstack-provision-sriov-networknodepolicy
     test:
+      - ref: openstack-test-dpdk
       - ref: openstack-test-sriov
     post:
       - chain: ipi-openstack-post
@@ -27,4 +28,4 @@ workflow:
       WORKER_REPLICAS: 0
   documentation: |-
     The openshift-e2e-openstack-sriov workflow executes an installation of
-    OpenShift on OpenStack in NFV environment and run SR-IOV tests.
+    OpenShift on OpenStack in NFV environment and run DPDK and SR-IOV tests.
--- a/ci-operator/config/openshift/sriov-network-operator/openshift-sriov-network-operator-master.yaml
+++ b/ci-operator/config/openshift/sriov-network-operator/openshift-sriov-network-operator-master.yaml
@@ -129,6 +129,8 @@ tests:
     cluster_profile: openstack-nfv
     dependencies:
       OO_INDEX: ci-index
+    env:
+      OPENSTACK_DPDK_NETWORK: intel-dpdk
     workflow: openshift-e2e-openstack-sriov
 - as: e2e-openstack-nfv-hwoffload
   optional: true
--- a/ci-operator/config/openshift/sriov-network-operator/openshift-sriov-network-operator-release-4.10.yaml
+++ b/ci-operator/config/openshift/sriov-network-operator/openshift-sriov-network-operator-release-4.10.yaml
@@ -129,6 +129,8 @@ tests:
     cluster_profile: openstack-nfv
     dependencies:
       OO_INDEX: ci-index
+    env:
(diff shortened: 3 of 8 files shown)


ci-operator/step-registry/openshift/e2e/openstack/sriov/openshift-e2e-openstack-sriov-workflow.yaml
```diff
@@ -11,6 +11,7 @@ workflow:
       - ref: optional-operators-subscribe
       - ref: openstack-provision-sriov-networknodepolicy
     test:
+      - ref: openstack-test-dpdk
```
pierreprinetti, line 14: Having multiple steps in the `test` stanza means that if one fails, the next ones are not run. Are you OK with that?

EmilienM: yes

ci-operator/step-registry/openstack/provision/sriov-operator/openstack-provision-sriov-operator-commands.sh
```diff
@@ -62,8 +62,6 @@ if [ -n "${is_dev_version:-}" ]; then
     echo "The SR-IOV will be installed from Github using release-${oc_version} branch."
     git clone --branch release-${oc_version} https://github.com/openshift/sriov-network-operator /tmp/sriov-network-operator
     pushd /tmp/sriov-network-operator
-    # Until https://github.com/openshift/sriov-network-operator/pull/613 merges
-    cp manifests/stable/supported-nic-ids_v1_configmap.yaml deploy/configmap.yaml
```
mandre, line 66: That seems fine, we have the `Red_Hat_Virtio_network_device: "1af4 1000 1000"` NIC we need in https://github.com/openshift/sriov-network-operator/blob/release-4.12/deploy/configmap.yaml

ci-operator/step-registry/openshift/e2e/openstack/sriov/openshift-e2e-openstack-sriov-workflow.yaml
```diff
@@ -11,6 +11,7 @@ workflow:
       - ref: optional-operators-subscribe
       - ref: openstack-provision-sriov-networknodepolicy
```
mandre, line 12: Is it normal that it's missing the `openstack-provision-sriov-operator` step?

ci-operator/step-registry/openshift/e2e/openstack/sriov/openshift-e2e-openstack-sriov-workflow.yaml
mandre: I guess we need to remove the `openstack-test-dpdk` step from the [`openshift-e2e-openstack-nfv` workflow](https://github.com/openshift/release/blob/41439f3/ci-operator/step-registry/openshift/e2e/openstack/nfv/openshift-e2e-openstack-nfv-workflow.yaml#L15) otherwise it would run twice, right?

mandre (commented):
Per your second commit:
> Our periodic job failed because a regression was introduced in the repo.
>
> We need to skip the DPDK tests if the network name is not defined, this
> is for the HW offload job that can't run DPDK on the same host.

Could you be more specific what regression and what repo you're talking about? I'm unclear what issue we're trying to solve here.

ci-operator/step-registry/openshift/e2e/openstack/sriov/openshift-e2e-openstack-sriov-workflow.yaml
EmilienM: yes because the operator is not deployed from OLM or source, but from the CI step registry in `optional-operators-subscribe`. So we don't need `openstack-provision-sriov-operator`.

EmilienM:
/pj-rehearse pull-ci-openshift-sriov-network-operator-release-4.10-e2e-openstack-nfv periodic-ci-shiftstack-shiftstack-ci-main-periodic-4.13-e2e-openstack-nfv-intel

PR #32246, opened 2022-09-13
provider-cert-tool: Run more often

bostrt:
We'd like to run the openshift-provider-certification-tool periodic CI job more often so we can generate test results to make better decisions on tests to exclude from certification and to get a better signal on the performance/reliability of tool.

/cc @mtulio @rvanderp3


bostrt:
Dedicated cert machine failed to scale-up in time. 

/test pj-rehearse

PR #32248, opened 2022-09-13
Aws public subnets only workflow

r4f4:
Depends on https://github.com/openshift/installer/pull/6342

This change introduces a new workflow to deploy AWS clusters in CI with public subnets only. Even though this is not officially supported by the Openshift Installer, it can be enabled if the `OPENSHIFT_INSTALL_AWS_PUBLIC_ONLY` env var is present.
The goal of doing so is to reduce CI expenditure since NAT gateways are quite expensive.

Note:  `[sig-network] Services should be able to update service type to NodePort listening on same port number but different protocols [Suite:openshift/conformance/parallel] [Suite:k8s]` test, which was recently introduced by the Kube rebase, is expected to fail at the moment. That's because it tries to reach the nodes on their External IPs, which is blocked by the OCP firewall rules. The test needs to be changed upstream [1]; in the meantime, jobs using this workflow will have to skip the test.

[1] https://github.com/kubernetes/kubernetes/blob/master/test/e2e/network/service.go#L1347


bear-redhat:
Is this `OPENSHIFT_INSTALL_AWS_PUBLIC_ONLY` switch available on existing AWS workflow or its a new workflow?

r4f4:
> Is this `OPENSHIFT_INSTALL_AWS_PUBLIC_ONLY` switch available on existing AWS workflow or its a new workflow?

It's a new workflow for now since it also depends on a properly configured VPC. The end goal is to use public subnets so we can remove NAT gateways from CI jobs and reduce costs from NAT charges. My idea was to test this without breaking existing workflows and once it's validated, then we can decide how we are going to transition existing jobs to NAT-less setups.

bear-redhat:
I think we need to fix the test first...

r4f4:
`Networking Granular Checks` tests are failing because of the way these kubernetes tests are picking the node IP to use. As we can see here [1], external IPs are preferred over internal IPs. However Openshift SG setup blocks 0/0, since nodes in the public subnets should not expose ports to the internet, only using LB (security reasons) - mainly in integrated platforms. Changing the tests to pick the internal IP in `openshift/origin`:

```
diff --git a/vendor/k8s.io/kubernetes/test/e2e/framework/network/utils.go b/vendor/k8s.io/kubernetes/test/e2e/framework/network/utils.go
index a99f289bc3..afc058e813 100644
--- a/vendor/k8s.io/kubernetes/test/e2e/framework/network/utils.go
+++ b/vendor/k8s.io/kubernetes/test/e2e/framework/network/utils.go
@@ -815,9 +814,11 @@ func (config *NetworkingTestConfig) setup(selector map[string]string) {
                secondaryFamily = v1.IPv4Protocol
        }
        // Get Node IPs from the cluster, ExternalIPs take precedence
-       config.NodeIP = e2enode.FirstAddressByTypeAndFamily(nodeList, v1.NodeExternalIP, family)
+       config.NodeIP = e2enode.FirstAddressByTypeAndFamily(nodeList, v1.NodeInternalIP, family)
+       framework.Logf("Internal NodeIP: %s", config.NodeIP)
        if config.NodeIP == "" {
-               config.NodeIP = e2enode.FirstAddressByTypeAndFamily(nodeList, v1.NodeInternalIP, family)
+               config.NodeIP = e2enode.FirstAddressByTypeAndFamily(nodeList, v1.NodeExternalIP, family)
+               framework.Logf("External NodeIP: %s", config.NodeIP)
        }
        if config.DualStackEnabled {
                config.SecondaryNodeIP = e2enode.FirstAddressByTypeAndFamily(nodeList, v1.NodeExternalIP, secondaryFamily)

```

results in the granular checks passing.

[1] https://github.com/kubernetes/kubernetes/blob/master/test/e2e/framework/network/utils.go#L818-L822

ci-operator/step-registry/ipi/conf/aws/publicsubnets/ipi-conf-aws-publicsubnets-commands.sh
```diff
@@ -0,0 +1,238 @@
...
+
+aws --region "${REGION}" cloudformation wait stack-create-complete --stack-name "${STACK_NAME}" &
+wait "$!"
+echo "Waited for stack"
+
+subnets="$(aws --region "${REGION}" cloudformation describe-stacks --stack-name "${STACK_NAME}" | jq -c '[.Stacks[].Outputs[] | select(.OutputKey | endswith("SubnetIds")).OutputValue | split(",")[]]' | sed "s/\"/'/g")"
+echo "Subnets : ${subnets}"
+
+# save stack information to ${SHARED_DIR} for deprovision step
+echo "${STACK_NAME}" >> "${SHARED_DIR}/publicsubnetsstackname"
```
r4f4, line 216: By using `${SHARED_DIR}/sharednetworkstackname` here I could re-use the shared-vpc deprovision step.

r4f4:
> Is this `OPENSHIFT_INSTALL_AWS_PUBLIC_ONLY` switch available on existing AWS workflow or its a new workflow?

It is now available in the `ipi-install-install` and `ipi-install-install-aws` steps.

bear-redhat:
```Command curl -g -q -s 'http://10.129.3.38:9080/dial?request=hostname&protocol=udp&host=18.116.53.75&port=30045&tries=1'
retrieved map[]
expected map[netserver-0:{} netserver-1:{} netserver-2:{}]
```
Shall we fix the tests as well?

ci-operator/step-registry/ipi/conf/aws/publicsubnets/ipi-conf-aws-publicsubnets-commands.sh
```diff
@@ -0,0 +1,238 @@
...
+TAGS="Key=expirationDate,Value=${EXPIRATION_DATE}"
+
+CONFIG="${SHARED_DIR}/install-config.yaml"
+PATCH=/tmp/install-config-publicsubnets.yaml.patch
+
+REGION="${LEASED_RESOURCE}"
+
+CLUSTER_NAME="$(yq-go r "${CONFIG}" 'metadata.name')"
+
+# FIXME: Parameterize installer/upi/aws/cloudformation/01_vpc.yaml to allow
```
patrickdillon, line 21: /shrug

I think the heredoc is actually fine.

patrickdillon (commented):
I thought we needed to edit the machine manifests to set 'PublicIP' on the compute nodes. Doing that would be a bit of a challenge, because manifest creation and editing is not broken into its own step (which it should be imho). 

I see the e2e tests are getting to about where we expect. Do we need to edit those manifests or have we determined that is unnecessary?

r4f4:
> I thought we needed to edit the machine manifests to set 'PublicIP' on the compute nodes. Doing that would be a bit of a challenge, because manifest creation and editing is not broken into its own step (which it should be imho).
> 
> I see the e2e tests are getting to about where we expect. Do we need to edit those manifests or have we determined that is unnecessary?

In my tests they were basically redundant. `MapPublicIP = true` in the VPC definition is enough.

r4f4:
/remove-lifecycle stale 
waiting on https://github.com/openshift/kubernetes/pull/1446 to fix the tests

r4f4:
Rebased to fix merge conflicts.

r4f4:
/pj-rehearse pull-ci-openshift-installer-release-4.12-e2e-aws-ovn-public-subnets

r4f4:
/pj-rehearse ci/rehearse/openshift/installer/master/e2e-aws-public-subnets

r4f4:
/pj-rehearse pull-ci-openshift-installer-master-e2e-aws-ovn-public-subnets

r4f4:
Tests are still not working because, even though https://github.com/openshift/kubernetes/pull/1446 has merged, the changes are not yet available in https://github.com/openshift/origin/.

r4f4:
/pj-rehearse pull-ci-openshift-installer-master-e2e-aws-ovn-public-subnets

jupierce:
/pj-rehearse pull-ci-openshift-installer-master-e2e-aws-ovn-public-subnets

bear-redhat:
@r4f4 I might miss something but it seems like `OPENSHIFT_INSTALL_AWS_PUBLIC_ONLY` is just print a line without doing anything in aws-ipi-install?

r4f4:
> @r4f4 I might miss something but it seems like `OPENSHIFT_INSTALL_AWS_PUBLIC_ONLY` is just print a line without doing anything in aws-ipi-install?

That env var is needed by the Openshift Installer since public subnets is a CI-only feature and not officially supported.

bear-redhat:
> > @r4f4 I might miss something but it seems like `OPENSHIFT_INSTALL_AWS_PUBLIC_ONLY` is just print a line without doing anything in aws-ipi-install?
> 
> That env var is needed by the Openshift Installer since public subnets is a CI-only feature and not officially supported.
Make sense. I was curious why it is defined by never been used in workflow :-)

bear-redhat:
`e2e-aws-ovn-public-subnets` is failing. I think we had discussion over this test as well. Any plans to fix it or we shall ignore?

bear-redhat:
/lgtm
Based on discussion, the failed test `e2e-aws-ovn-public-subnets` is introduced in this PR and no existing tests will be effected. Upstream fix is needed for the failed test.

r4f4:
I have updated the PR description to include the reasoning for this change and to point out that the failing e2e test is expected at the moment.

2uasimojo:
I think this may be causing [failures](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cloud-credential-operator/521/pull-ci-openshift-cloud-credential-operator-master-e2e-upgrade/1633913145904861184#1:build-log.txt%3A181) due to the variable being unset in some paths. We should default it.

https://github.com/openshift/release/pull/37135

PR #32250, opened 2022-09-13
VolSync: Don't configure default StorageClass

JohnStrunk:
The default SC has now changed to be a CSI driver, so it's no longer necessary for us to adjust it.


tesshuflower (approved):
/lgtm

PR #32257, opened 2022-09-13
Add `centos:stream8` and `centos:stream9` into the `openshift` ns

cgwalters:
This is related to https://issues.redhat.com/browse/OCPBUGS-1165 where we're using a long dead Fedora release in CI and only noticed now.

We should generally be using CentOS stream for this stuff unless there's a strong reason to use Fedora.

Let's start by mirroring it into the `openshift` namespace.  This will make it easier to consume as part of `base_images`.

(Other CI steps here directly reference the `quay.io` image)


travier (commented):
/lgtm

travier (approved):

PR #32267, opened 2022-09-14
add fedora variant to cri-o CI

sohankunkerkar:
We will skip the e2e tests until we figure out the actual issue.


sohankunkerkar:
@skyler is planning to handle the deduplication of the test and gather scripts.

haircommander:
/approve

LGTM

skyler:
@sohankunkerkar though an exciting challenge to be sure, I will not be doing that 😛 (I think you've tagged the wrong person).

PR #32272, opened 2022-09-14
add sdn migraiton ovn job

zhaozhanqi:


ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.11__amd64-nightly.yaml
```diff
@@ -87,6 +87,17 @@ tests:
...
     - chain: openshift-e2e-test-qe
     workflow: cucushift-installer-rehearse-aws-ipi-proxy-techpreview
+- as: e2e-aws-ipi-sdn-migration-ovn
+  cron: 0 6 */3 * *
+  steps:
+    cluster_profile: aws-qe
+    env:
+      BASE_DOMAIN: qe.devcluster.openshift.com
+      UPGRADE_CHECK_RUN_TAGS: '@4.11 and @aws-ipi and @upgrade-check and @network-ovnkubernetes'
+      UPGRADE_PRE_RUN_TAGS: '@4.11 and @aws-ipi and @upgrade-prepare and @network-openshiftsdn'
```
liangxia, line 97: Are we supposed to add to 4.11, rather than 4.12 ?

zhaozhanqi: 4.11 also supported this feature. so I think we will add this testing on 4.11 and 4.12 and following version

liangxia: We mainly focus on 4.12 ( in Prow ).  4.11 and previous versions are covered in Jenkins ATM.

liangxia: For 4.12, we mainly focus on Prow CI. And for older versions, we mainly focus on Jenkins.
So if you would like to add this test to 4.11, then it's better to add it to 4.12 first, then backport to 4.11 (or add 4.12 and 4.11 at the same time). This is to avoid a test is covered in both Jenkins and Prow, while another test is missed in both Prow and Jenkins.

liangxia:
BTW, there are conflicts.

PR #32276, opened 2022-09-14
Add CRI-O e2e conmonrs job

saschagrunert:
Adding a conmonrs variant job which injects the environment variable `USE_CONMONRS` as `true`. The rest will be done in the ansible playbook within the CRI-O repository.

Refers to https://github.com/cri-o/cri-o/pull/6225

PTAL @wgahnagl @haircommander @rphillips

### DIFF
--- a/ci-operator/step-registry/cri-o/e2e/test/cri-o-e2e-test-commands.sh
+++ b/ci-operator/step-registry/cri-o/e2e/test/cri-o-e2e-test-commands.sh
@@ -38,6 +38,7 @@ fi
 #####################################
 
 instance_name=$(<"${SHARED_DIR}/gcp-instance-ids.txt")
+USE_CONMONRS=${USE_CONMONRS:-false}
 
 tar -czf - . | gcloud compute ssh --zone="${ZONE}" ${instance_name} -- "cat > \${HOME}/cri-o.tar.gz"
 timeout --kill-after 10m 400m gcloud compute ssh --zone="${ZONE}" ${instance_name} -- bash - << EOF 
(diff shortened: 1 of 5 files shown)


ci-operator/step-registry/cri-o/e2e/test/cri-o-e2e-test-commands.sh
```diff
@@ -58,6 +59,5 @@ timeout --kill-after 10m 400m gcloud compute ssh --zone="${ZONE}" ${instance_nam
     tar -xzvf cri-o.tar.gz -C "\${REPO_DIR}"
     cd "\${REPO_DIR}/contrib/test/ci"
     echo "localhost" >> hosts
-    ansible-playbook e2e-main.yml -i hosts -e "TEST_AGENT=prow" --connection=local -vvv --tags setup,e2e
+    ansible-playbook e2e-main.yml -i hosts -e "TEST_AGENT=prow USE_CONMONRS=$USE_CONMONRS" --connection=local -vvv --tags setup,e2e
```
haircommander, line 62: in the other variants, we use an ansible variable, not a env variable. Part of me thinks they should be consistent (the other part doesn't really care :) )

saschagrunert: I'm wondering how it would fit into the current structure. Would that mean that we have to create a copy of the whole workflow to avoid the nev variable? (My goal was to inject as less code as possible since we can switch to conmonrs at a certain point in time)

haircommander: @wgahnagl @sohankunkerkar do y'all have any preferences

sohankunkerkar: > I think you can default the variable to false in cri-o's main.yml, then run certain installation steps if the variable `use_conmonrs` is true

Yeah, this sounds good. We can also use `--extra-vars` to pass that variable

sohankunkerkar (approved):

ci-operator/step-registry/cri-o/e2e/test/cri-o-e2e-test-commands.sh
sohankunkerkar: Yeah, looks reasonable to me.

PR #32285, opened 2022-09-14
Add CRI-O release-1.25 tests

saschagrunert:
Adding the tests for the release-1.25 branch. PTAL @wgahnagl @haircommander

### DIFF
--- a/ci-operator/config/cri-o/cri-o/cri-o-cri-o-release-1.25.yaml
+++ b/ci-operator/config/cri-o/cri-o/cri-o-cri-o-release-1.25.yaml
@@ -1,0 +1,64 @@
+binary_build_commands: "echo \"no build commands\" \n"
+build_root:
+  image_stream_tag:
+    name: release
+    namespace: openshift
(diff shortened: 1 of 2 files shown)


ci-operator/config/cri-o/cri-o/cri-o-cri-o-release-1.25.yaml
```diff
@@ -0,0 +1,62 @@
+base_images:
+  base:
+    name: "4.11"
```
haircommander, line 3: hm , I would expect this to look more like https://github.com/openshift/release/blob/master/ci-operator/config/cri-o/cri-o/cri-o-cri-o-main__ci.yaml than https://github.com/openshift/release/blob/master/ci-operator/config/cri-o/cri-o/cri-o-cri-o-main.yaml

saschagrunert: Yeah, this makes more sense to me.

PR #32294, opened 2022-09-14
Stackrox: use prow triggers

gavin-stackrox:
This PR changes stackrox/stackrox CI job triggering to remove some jobs on merge and trigger some others by file change. As outlined in https://docs.google.com/spreadsheets/d/1oAixMWYxKsyZhhOb3pBrKltzkmWeRy54qYUSY4kbPvY.

This is the kind of CI change that is hard to test against the target without merging first and testing later. So that is the plan.

/cc @RTann @connorgorman @vjwilson @house-d


vjwilson (approved):
/lgtm

PR #32299, opened 2022-09-14
no need to modify default driver for openshift-fips env

tesshuflower:
Signed-off-by: Tesshu Flower <tflower@redhat.com>

We missed updating the openshift-fips e2e tests when doing this change: https://github.com/openshift/release/pull/32250


JohnStrunk:
Thanks!
/lgtm

PR #32303, opened 2022-09-15
Add more job frequency levels

liangxia:
Previously we only have `daily`, `weekly`, and `monthly`. The frequency varies too much, and sometimes does not met our needs.

/cc @wsun1 @jhou1 @chengzhang1016 @xltian


chengzhang1016:
LGTM

PR #32320, opened 2022-09-15
Add mirroring for stackrox collector ci image

Molter73:
The image being mirrored is currently used as the `root` for the stackrox/collector build, it is quite big and it gets fully copied into `pipeline:src`: https://github.com/stackrox/collector/blob/03a2c942ae835d8c441ded7b29d19e71ed875d4e/.openshift-ci/build/Dockerfile.build_root#L1

I'm looking into refactoring how the `stackrox/collector` image is built and one of the steps I'd like to take is move this image to only be used by the tests that actually need it.


gavin-stackrox (approved):
Good luck!

PR #32329, opened 2022-09-15
mco: Drop layering branch config

cgwalters:
All the stuff is inbound for master, we're not using a layering git branch anymore.


jmarrero (approved):
lgtm

PR #32336, opened 2022-09-15
remove skip_if.. config for jobs w/ always_run: false

jluhrsen:
For jobs that are manually configured to never run with the 'always_run: false' flag set, adding the skip_if_only_changed config actually makes those jobs always run as long as that skip_if flag is not matched. This is not what was intended when the jobs were originally marked to never run. fix that.


tjungblu:
/lgtm

thanks a lot for your help @jluhrsen

PR #32338, opened 2022-09-15
clusters/build-clusters/common/assets/pod-network-connectivity-check-pruner: Mitigate PodNetworkConnectivityCheck leak

wking:
Mitigate [OCPBUGS-1341](https://issues.redhat.com/browse/OCPBUGS-1341) by pruning leaked PodNetworkConnectivityChecks.  It looks for checks where one or both nodes no longer exist in-cluster, and removes them.

There was an off-the-shelf `system:node-reader` to give node-listing access, but I needed to create a new Role to give PodNetworkConnectivityCheck list/delete access, because I couldn't find an existing role with delete.  I guess we could have used a system-admin-y role, but that seems like too much power for a shell script job.

There may also be a cleaner way to get access to the `openshift/tools:latest` ImageStreamTag, but I'm just going with the in-cluster registry URI, because that works and didn't take me too long to look up.

### DIFF
--- a/clusters/build-clusters/common/assets/pod-network-connectivity-check-pruner.yaml
+++ b/clusters/build-clusters/common/assets/pod-network-connectivity-check-pruner.yaml
@@ -1,0 +1,101 @@
+---
+kind: ServiceAccount
+apiVersion: v1
+metadata:
+  name: pod-network-connectivity-check-pruner
+  namespace: openshift-network-diagnostics
+  annotations:
+    kubernetes.io/description: Mitigate https://issues.redhat.com/browse/OCPBUGS-1341 by pruning leaked PodNetworkConnectivityChecks.
+---
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: pod-network-connectivity-check-pruner
+  namespace: openshift-network-diagnostics
+  annotations:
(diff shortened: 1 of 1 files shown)


bear-redhat:
It looks like `allowPrivilegeEscalation` is not allowed

wking:
> It looks like `allowPrivilegeEscalation` is not allowed

Yeah, [e.g.][1]:

    .../go/src/github.com/openshift/ci-tools/cmd/applyconfig/applyconfig.go...error validating data: ValidationError(CronJob.spec.jobTemplate.spec.template.spec.securityContext): unknown field \"allowPrivilegeEscalation\" in io.k8s.api.core.v1.PodSecurityContext...

Hmm, I guess I confused `PodSecurityContext` with (container) `SecurityContext` when trying to add the bits that the Kube API server was warning me it wanted.  Dropped the `PodSecurityContext` stuff with 065f56de899 -> 4133e0f840a.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_release/32338/pull-ci-openshift-release-master-vsphere-dry/1570520443754909696#1:build-log.txt%3A18

droslean (approved):

hongkailiu:
Screen Shot 2022-09-16 at 10 15 12 AM

The manual triggered job is done on b01.

PR #32339, opened 2022-09-15
vSphere: remove segment 93

jcpowermac:
Strange issue with this segment only in IBM.
DHCP client does not function correctly in
windows only.


aravindhp:
This should hopefully resolve WINC-887 🤞🏽

aravindhp (approved):
/lgtm

Thanks for doing this, @jcpowermac

PR #32344, opened 2022-09-16
Assisted Service: Enable agent-installer e2e tests

zaneb:
Create an e2e-agent-compact job and run it automatically against changes specific to the ephemeral agent-based installer (i.e. the OCP Dockerfile and the agent-based installer client.

For now this job is optional because it is known to be broken by a regression somewhere between the release-4.11 branch and master.

### DIFF
--- a/ci-operator/config/openshift/assisted-service/openshift-assisted-service-master.yaml
+++ b/ci-operator/config/openshift/assisted-service/openshift-assisted-service-master.yaml
@@ -3,6 +3,10 @@ base_images:
     name: "4.12"
     namespace: ocp
     tag: base
+  dev-scripts:
+    name: test
+    namespace: ocp-kni
+    tag: dev-scripts
   ocp_4.11_cli:
     name: "4.12"
     namespace: ocp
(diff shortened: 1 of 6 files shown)


ci-operator/config/openshift/assisted-service/openshift-assisted-service-master.yaml
```diff
@@ -46,6 +50,12 @@ tests:
       SOURCE_IMAGE_REF: agent-installer-api-server
     test:
     - ref: assisted-baremetal-images-publish
+- as: e2e-agent-compact
+  optional: true
+  run_if_changed: ^(Dockerfile\.assisted-service\.ocp)|(cmd/agentbasedinstaller)
+  steps:
+    cluster_profile: equinix-ocp-metal
```
derekhiggins, line 57: There are 2  possible accounts this could go on, other jobs on the repository are using "packet-assisted" while ipi jobs use the "equinix-ocp-metal" profile, I guess it depend which group this job is benefiting. If this job will be using equinix-ocp-metal can you add "cluster: build05" so that this job will have access to our internal hardware and use it if any is available.

zaneb: > There are 2 possible accounts this could go on, other jobs on the repository are using "packet-assisted" while ipi jobs use the "equinix-ocp-metal" profile, I guess it depend which group this job is benefiting.

This is really for the agent-based installer team, so I think equinix-ocp-metal is right, @celebdor can confirm.

> If this job will be using equinix-ocp-metal can you add "cluster: build05" so that this job will have access to our internal hardware and use it if any is available.

Can do. Presumably we want this for https://github.com/openshift/release/blob/master/ci-operator/config/openshift/installer/openshift-installer-master.yaml#L506-L511 as well? I can follow up with a separate PR.

zaneb:
OK, it's now failing in the way we expect due to a known bug.

ci-operator/config/openshift/assisted-service/openshift-assisted-service-master.yaml
zaneb: #32368

PR #32348, opened 2022-09-16
[prow-plugins] Add openshift loki branch patterns to shared config

periklis:


bear-redhat:
/hold
We previously ran into issues for adding new items to `misc` section of the secret configs (exceeds the size limits of k8s). I remember @bbguimaraes has done something to further break down those `misc` items, but I am not sure if that work is finished and deployed. Feel free to cancel the hold if anyone can confirm.

PR #32349, opened 2022-09-16
Upgrade used nightlies and stables for k-nmstate

creydr:
Upgrade nightlies and stables.

### DIFF
--- a/ci-operator/config/openshift/kubernetes-nmstate/openshift-kubernetes-nmstate-master.yaml
+++ b/ci-operator/config/openshift/kubernetes-nmstate/openshift-kubernetes-nmstate-master.yaml
@@ -68,11 +68,11 @@ releases:
     candidate:
       product: ocp
       stream: nightly
-      version: "4.11"
+      version: "4.12"
   stable:
     release:
       channel: stable
-      version: "4.10"
+      version: "4.11"
 resources:
   '*':
     limits:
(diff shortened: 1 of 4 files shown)


ci-operator/config/openshift/kubernetes-nmstate/openshift-kubernetes-nmstate-master.yaml
```diff
@@ -68,11 +68,11 @@ releases:
     candidate:
       product: ocp
       stream: nightly
-      version: "4.11"
+      version: "4.12"
   stable:
     release:
       channel: stable
-      version: "4.10"
+      version: "4.11"
```
cybertron, line 75: For 4.12 and 4.13 these match. Should this be 4.12?

cybertron, line 75: For 4.12 and 4.13 these match. Should this be 4.12?

creydr: The thing is, we don't have a stable build for 4.12 yet. So I updated it for the master branch to reference the last available stable.
In the 4.12/4.13 files I set it to 4.12/4.13, because there it is obvious when theses builds fail (because the stable build is not available yet)

cybertron: Okay, that makes sense. Thanks for the explanation.

cybertron:
/lgtm

We know the handler-ovn job has a flake, and that shouldn't be affected by this change anyway.

PR #32351, opened 2022-09-16
Refactor stackrox/collector configuration

Molter73:


ci-operator/config/stackrox/collector/stackrox-collector-mauro-refactor-stackrox-collector-config.yaml
```diff
@@ -266,32 +270,11 @@ resources:
...
-      commands: make -C collector check
-      from: src
-      resources:
-        requests:
-          cpu: 2000m
-          memory: 4000Mi
-    - as: shfmt-check
-      commands: make shfmt-check
-      from: src
-      resources:
```
Stringy, line 281: Why do we remove the resources? is it just that it's not so important for these tests?

Stringy: (same question for the other tests )

ci-operator/config/stackrox/collector/stackrox-collector-mauro-refactor-stackrox-collector-config.yaml
```diff
@@ -266,32 +270,11 @@ resources:
...
-  steps:
-    test:
-    - as: shellcheck-all
-      commands: make shellcheck-all
-      from: src
-      resources:
-        requests:
-          cpu: 2000m
-          memory: 4000Mi
+  commands: |
```
Stringy, line 273: [nit] somewhat outside the scope of this PR but it might be worth making this a required step

Stringy (approved):
Looks good! Lots of nice simplification in there

ci-operator/config/stackrox/collector/stackrox-collector-mauro-refactor-stackrox-collector-config.yaml
Molter73: When using the short, non-steps version of tests, there is no `resources` entry that can be added to the element, I believe you need to specify it in the global `resources` element instead, but to be honest I think it makes no difference.

ci-operator/config/stackrox/collector/stackrox-collector-mauro-refactor-stackrox-collector-config.yaml
Molter73: I agree, but the original lints were not defined as required, so we might want to do it in a follow up

Molter73:
/unhold

Rehearsal will fail on master because there are a bunch changes missing in that branch.

Stringy (approved):

PR #32162, opened 2022-09-09
Remove HACBS builder image builds

stuartwdouglas:
These have been moved to a seperate repo and no longer need to be built for every pull request.

PR #32352, opened 2022-09-16
remove jdk build images from jvm-build-service

gabemontero:
@stuartwdouglas has not had a chance to rebase https://github.com/openshift/release/pull/32162 and I had a moment today


gabemontero:
the jvm-build-service-e2e failure here is expected ... that no longer works off of main branch, but only works in repo PRs where we override the images and bundles while the kcp transition is still in flight

PR #32354, opened 2022-09-16
Restore deleted inventory from osp-operators pool

viroel:
This patch restore deleted content removed at [1]. Since [2] merged, we should not see this issue happening again.

[1] https://github.com/openshift/release/commit/4b71ac5 [2] https://github.com/openshift/ci-tools/commit/d24afb9f2726685e95b5dcee97ac6889a5a2343f


abays (approved):
/lgtm

PR #32361, opened 2022-09-16
[WIP] Update microshift test image to latest rhel 8.6

copejon:
I've cloned the latest publicly available rhel 8.6 from the rhel-cloud image family into the CI project.  This is to protect against future outages resulting from current images being removed.  This created a situation where GCP would quietly use the latest rhel-8 image.

[USHIFT-411](https://issues.redhat.com//browse/USHIFT-411)


copejon:
failures: kuttl is complaining that it can't find the kubevirt-hostpath pod, which is odd as the kuttl configs in main:HEAD do not contain any reference to it.

PR #32366, opened 2022-09-16
Use `centos:stream9` and not `fedora:29`

cgwalters:
Aiming to fix https://github.com/openshift/kubernetes/pull/1363


cgwalters:
We could probably just blow past the rehearsals for this - everything using fedora:29 is going to basically be broken anyways.

PR #32371, opened 2022-09-16
fixes etcd version

wgahnagl:
kubernetes bumped the etcd version, so we have to as well!
will be working on a better solution for this issue on Monday! :ok_hand:


mrunalp (approved):
/lgtm

PR #32375, opened 2022-09-18
Tide: Prioritize pull-request marked as critical/urgent

eliorerz:
Sometimes on Assisted-Installer we need to prioritize some pull-requests over the others regardless jira/BZ.
I added this label according to https://github.com/kubernetes/test-infra/blob/78036aad7fe6c11d2724b5390ca0ee11445c4308/prow/config/prow-config-documented.yaml#L1289-L1294

### DIFF
--- a/core-services/prow/02_config/_config.yaml
+++ b/core-services/prow/02_config/_config.yaml
@@ -379,6 +379,8 @@ tide:
     openshift-priv: https://deck-internal-ci.apps.ci.l2s4.p1.openshiftapps.com/pr
   priority:
   - labels:
+    - priority/critical-urgent
+  - labels:
     - jira/severity-critical
   - labels:
     - bugzilla/severity-urgent


core-services/prow/02_config/_config.yaml
```diff
@@ -378,6 +378,8 @@ tide:
     '*': https://prow.ci.openshift.org/pr
     openshift-priv: https://deck-internal-ci.apps.ci.l2s4.p1.openshiftapps.com/pr
   priority:
+  - labels:
```
jmguzik, line 381: I don't think we want to change this for everyone, and by editing the main file we do.

eliorerz: I'm not sure we can change it only for our team

eliorerz: This is tag that is added manually and can help the developers override the priorities of the repository pull-requests

PR #32376, opened 2022-09-18
add aws outpost support

AsherShoshan:


jianlinliu:
Several questions:
1. Pls always ensure the validation jobs get passed.
2. To be honest, I did not get the points of `OUTPOST` meaning? Can you explain more?

ci-operator/step-registry/ipi/conf/aws/outpost/ipi-conf-aws-outpost-chain.yaml
```diff
@@ -0,0 +1,6 @@
+chain:
+  as: ipi-conf-aws-outpost
```
jianlinliu, line 2: If this chain only includes one ref, I guess we do not need to introduce the chain.

AsherShoshan: so just remove *-chain from folder?

to ur questions above:

jianlinliu: > 'outpost' is an on-premise hardware of Amazon
Sounds like need a specific aws account, right? If right, I think you need start from the very beginning of CLUSTER_PROFILE creation.

AsherShoshan: Why create a new cluster_profile, and not just inject my domain + credentials?  (it is going to be temp solution until 'regular' ci aws account will have it's own 'outpost'.  right?

PR #32377, opened 2022-09-18
source /root/config.sh when collecting logs in order to collect logs with the right env vars

eranco74:
This should fix the missing CRs output (agent, infraenv, clusterdeployment, etc...) in the assisted-common-gather artifacts

### DIFF
--- a/ci-operator/step-registry/assisted/common/gather/assisted-common-gather-ref.yaml
+++ b/ci-operator/step-registry/assisted/common/gather/assisted-common-gather-ref.yaml
@@ -2,7 +2,7 @@ ref:
   as: assisted-common-gather
   best_effort: true
   timeout: 120m
-  optional_on_success: true
+  optional_on_success: false
   from: dev-scripts
   grace_period: 10m
   commands: assisted-common-gather-commands.sh


eranco74:
cc: @osherdp @adriengentil

eranco74:
In hope that some job will fail and the gather step will get executed...
/test pj-rehearse

ci-operator/step-registry/assisted/common/gather/assisted-common-gather-commands.sh
```diff
@@ -55,6 +57,8 @@ cat > gather_logs.yaml <<-EOF
           LOGS_DEST: "{{ LOGS_DIR }}"
         when: kubeconfig.stat.exists
       - name: Extract capi logs
+        ansible.builtin.shell: |
+          source /root/config.sh
         make:
           chdir: /home/assisted
           target: download_capi_logs
```
adriengentil, line 64: I'm not sure you can combine 2 actions in one task, something like that should work:
```suggestion
        ansible.builtin.shell: |
          source /root/config.sh
          cd /home/assisted
          make download_capi_logs
```

adriengentil:
> In hope that some job will fail and the gather step will get executed... /test pj-rehearse

you can turn off this setting in order to test if it works well: https://github.com/openshift/release/blob/master/ci-operator/step-registry/assisted/common/gather/assisted-common-gather-ref.yaml#L5

ci-operator/step-registry/assisted/common/gather/assisted-common-gather-commands.sh
```diff
@@ -47,20 +47,22 @@ cat > gather_logs.yaml <<-EOF
...
-        make:
-          chdir: /home/assisted
-          target: download_service_logs
+        ansible.builtin.shell: |
+          source /root/config.sh
+          make download_service_logs
         environment:
           KUBECONFIG: "/root/.kube/config"
           LOGS_DEST: "{{ LOGS_DIR }}"
         when: kubeconfig.stat.exists
```
adriengentil, line 56:
```suggestion
        args:
          chdir: /home/assisted
        when: kubeconfig.stat.exists
```

adriengentil (approved):
@eliorerz @osherdp 
FYI this PR will re-enable the gather step for all the jobs

PR #32379, opened 2022-09-19
vsphere: fix tag and dns deleting issue for rhel worker

jinyunma:
Fix issue in failed job: https://qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/qe-private-deck/logs/periodic-ci-openshift-openshift-tests-private-release-4.12-amd64-nightly-e2e-vsphere-upi-disconnected-sdn-workers-rhcos-rhel8-p2/1571288148917358592

For UPI installation, tag is created by CSI driver, in some reason, tag seems deleted, but this should not block rhel worker provision, add check to enhance it, only attach tag to vm when it is existing, as we have deprovision step of rhel work on upi.


jinyunma:
cc @jianlinliu @sgaoshang for review.

Failed job looks like worker vms are unable to get IPs when provisioning, unrelated with the change in this PR.

PR #32384, opened 2022-09-19
Fix migration jobs for release-controller configs for 4.12

pliurh:
Since 4.12, the openshift-e2e-aws workflow now uses OVN-K as the default CNI, for migration jobs we need the cluster start with OpenShiftSDN.

This patch also syncs the release config with the CNO config.

Signed-off-by: Peng Liu <pliu@redhat.com>


vrutkovs (approved):
/lgtm

PR #32393, opened 2022-09-19
Installer: add periodics to test OKD using FCOS stable/testing for firstboot

vrutkovs:
This PR adds a step to patch installer with FCOS AMI from selected stream and adds two periodic jobs to check OKD 4.12 install on FCOS testing stream and OKD 4.11 on stable stream

### DIFF
--- a/ci-operator/step-registry/ipi/conf/aws/fcos-ami/ipi-conf-aws-fcos-ami-commands.sh
+++ b/ci-operator/step-registry/ipi/conf/aws/fcos-ami/ipi-conf-aws-fcos-ami-commands.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+CONFIG="${SHARED_DIR}/install-config.yaml"
+
+URL="https://builds.coreos.fedoraproject.org/streams/${STREAM}.json"
+FCOS_AMI="$(curl -s "${URL}" | jq ".architectures.\"${ARCHITECTURE}\".images.aws.regions.\"${LEASED_RESOURCE}\".image")"
+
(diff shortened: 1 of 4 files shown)


vrutkovs:
okd-e2e-aws-ovn-stable-stream passed, testing stuck on bootstrap
/retest

cgwalters:
Seems sane to me, though it's worth noting that this will only test the firstboot flow - the running OS will still be replaced with the target OS content.  I assume that's intentional here, but I think it's worth spelling out for anyone later who might be confused at why e.g. the `kernel` in the target stream isn't used when kubelet is running.

vrutkovs:
>this will only test the firstboot flow - the running OS will still be replaced with the target OS content

Correct. For latter we're going to need a more elaborate promotion periodic

ci-operator/step-registry/ipi/conf/aws/fcos-ami/ipi-conf-aws-fcos-ami-commands.sh
```diff
@@ -0,0 +1,28 @@
...
+CONFIG="${SHARED_DIR}/install-config.yaml"
+
+# Install tools
+echo "## Install yq"
+curl -L https://github.com/mikefarah/yq/releases/download/3.3.0/yq_linux_amd64 -o /tmp/yq && chmod +x /tmp/yq
+echo "   yq installed"
+
+echo "## Install jq"
+curl -L https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64 -o /tmp/jq && chmod +x /tmp/jq
+echo "   jq installed"
```
patrickdillon, line 16: These are now baked into the upi image... can you use those instead?

PR #32340, opened 2022-09-15
Switch dynamic-plugin-sdk job to use custom CI build image

florkbr:
`openshift-dynamic-plugin-sdk-buildroot:v1` should now be published from https://github.com/openshift/release/pull/32167

PR #32401, opened 2022-09-19
Update git config safe directories for dynamic-plugin-sdk-buildroot

florkbr:
We noticed an error checking our the dynamic-plugin-sdk project when using the new buildroot image (git unsafe directory). We could theoretically only add the SDK repo - but we may eventually use this image in other git projects - so we've opted for the wildcard.

See https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_release/32340/rehearse-32340-pull-ci-openshift-dynamic-plugin-sdk-main-all-packages/1570526305030508544/build-log.txt (https://github.com/openshift/release/pull/32340)


vojtechszocs (approved):

florkbr:
@bear-redhat would you mind having a look at this fix? Thanks

PR #32408, opened 2022-09-19
volsync: Add back e2e w/ helm install

JohnStrunk:
- Add back an e2e variant that installs via Helm. This is because we use a different (custom) SCC when we install via Helm, and we need to make sure that continues to grant proper permissions.
- Make the e2e tests mandatory for merge.
- ~Update the operator catalog base to be 4.11~ Doesn't seem to work :confused:


tesshuflower (approved):
/lgtm

PR #32412, opened 2022-09-19
Add skip_if_only_changed option into opendatahub-io-notebooks-main.yaml

atheo89:
Added `skip_if_only_changed` option into opendatahub-io-notebooks-main.yaml in order for the job to skip  pull requests that only change documentation files


atheo89:
cc @samuelvl

PR #32439, opened 2022-09-20
start managing the arm01 cluster

droslean:
/hold

/cc @hongkailiu @openshift/test-platform 


droslean:
@hongkailiu You mentioned in the past that arm01 is a special case, and I see that after I added it to the managed clusters, many changes occurred and it seems that some of the hacks have been removed. Can you check if this is the best way to move forward?

PR #32440, opened 2022-09-20
Add e2e-aws-ipi-disconnected-private for ARM

lwan-wanglin:
refer [ARMOCP-451](https://issues.redhat.com/browse/ARMOCP-451)

/hold

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__arm64-nightly.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__arm64-nightly.yaml
@@ -203,6 +203,23 @@ tests:
     test:
     - chain: openshift-e2e-test-qe
     workflow: cucushift-installer-rehearse-baremetalds-ipi-ovn-ipv4
+- as: e2e-aws-ipi-disconnected-private-p2
+  cron: 0 16 */7 * *
+  steps:
+    cluster_profile: aws-qe
+    dependencies:
+      OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE: release:arm64-latest
+    env:
+      BASE_DOMAIN: qe.devcluster.openshift.com
+      COMPUTE_NODE_TYPE: m6g.large
+      E2E_RUN_TAGS: '@aws-ipi and @arm64 and @disconnected and @network-ovnkubernetes
+        and not @fips'
+      OCP_ARCH: arm64
+      TAG_VERSION: '@4.12'
+      TEST_FILTERS: ~ChkUpgrade&;~NonPreRelease&;~Serial&;~Disruptive&;~ConnectedOnly&
+    test:
+    - chain: openshift-e2e-test-qe
+    workflow: cucushift-installer-rehearse-aws-ipi-disconnected-private
 zz_generated_metadata:
   branch: release-4.12
   org: openshift
--- a/ci-operator/step-registry/mirror-images/by-oc-adm/mirror-images-by-oc-adm-commands.sh
+++ b/ci-operator/step-registry/mirror-images/by-oc-adm/mirror-images-by-oc-adm-commands.sh
@@ -21,14 +21,14 @@ fi
 echo "MIRROR_REGISTRY_HOST: $MIRROR_REGISTRY_HOST"
 
 # target release
-target_release_image="${MIRROR_REGISTRY_HOST}/${RELEASE_IMAGE_LATEST#*/}"
+target_release_image="${MIRROR_REGISTRY_HOST}/${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE#*/}"
 target_release_image_repo="${target_release_image%:*}"
 target_release_image_repo="${target_release_image_repo%@sha256*}"
 
 echo "target_release_image: $target_release_image"
 echo "target_release_image_repo: $target_release_image_repo"
 
-readable_version=$(oc adm release info "${RELEASE_IMAGE_LATEST}" --output=json | jq .metadata.version)
+readable_version=$(oc adm release info "${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE}" --output=json | jq .metadata.version)
 echo "readable_version: $readable_version"
 
 # since ci-operator gives steps KUBECONFIG pointing to cluster under test under some circumstances,
@@ -42,7 +42,7 @@ jq --argjson a "{\"${MIRROR_REGISTRY_HOST}\": {\"auth\": \"$registry_cred\"}}" '
 
 # MIRROR IMAGES
 oc adm release -a "${new_pull_secret}" mirror --insecure=true \
- --from=${RELEASE_IMAGE_LATEST} \
+ --from=${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE} \
  --to=${target_release_image_repo} \
  --to-release-image=${target_release_image} | tee "${mirror_output}"
 
--- a/ci-operator/step-registry/mirror-images/by-oc-adm/mirror-images-by-oc-adm-ref.yaml
+++ b/ci-operator/step-registry/mirror-images/by-oc-adm/mirror-images-by-oc-adm-ref.yaml
@@ -12,5 +12,8 @@ ref:
   - namespace: test-credentials
     name: openshift-custom-mirror-registry
     mount_path: /var/run/vault/mirror-registry
+  dependencies:
+  - name: "release:latest"
+    env: OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE
   documentation: |-
     Mirror release image to local image registry.
--- a/ci-operator/jobs/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12-periodics.yaml
+++ b/ci-operator/jobs/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12-periodics.yaml
@@ -5987,6 +5987,88 @@ periodics:
         secretName: result-aggregator
 - agent: kubernetes
   cluster: build03
+  cron: 0 16 */7 * *
+  decorate: true
+  decoration_config:
+    skip_cloning: true
+  extra_refs:
+  - base_ref: release-4.12
+    org: openshift
+    repo: openshift-tests-private
+  labels:
+    ci-operator.openshift.io/cloud: aws
+    ci-operator.openshift.io/cloud-cluster-profile: aws-qe
+    ci-operator.openshift.io/variant: arm64-nightly
+    ci.openshift.io/generator: prowgen
+    job-release: "4.12"
+    pj-rehearse.openshift.io/can-be-rehearsed: "true"
+  name: periodic-ci-openshift-openshift-tests-private-release-4.12-arm64-nightly-e2e-aws-ipi-disconnected-private-p2
+  spec:
+    containers:
+    - args:
+      - --gcs-upload-secret=/secrets/gcs/service-account.json
+      - --image-import-pull-secret=/etc/pull-secret/.dockerconfigjson
+      - --lease-server-credentials-file=/etc/boskos/credentials
+      - --oauth-token-path=/usr/local/github-credentials/oauth
+      - --report-credentials-file=/etc/report/credentials
+      - --secret-dir=/secrets/ci-pull-credentials
+      - --secret-dir=/usr/local/e2e-aws-ipi-disconnected-private-p2-cluster-profile
+      - --target=e2e-aws-ipi-disconnected-private-p2
+      - --variant=arm64-nightly
+      command:
+      - ci-operator
+      image: ci-operator:latest
+      imagePullPolicy: Always
+      name: ""
+      resources:
+        requests:
+          cpu: 10m
+      volumeMounts:
+      - mountPath: /etc/boskos
+        name: boskos
+        readOnly: true
+      - mountPath: /secrets/ci-pull-credentials
+        name: ci-pull-credentials
+        readOnly: true
+      - mountPath: /usr/local/e2e-aws-ipi-disconnected-private-p2-cluster-profile
+        name: cluster-profile
+      - mountPath: /secrets/gcs
+        name: gcs-credentials
+        readOnly: true
+      - mountPath: /usr/local/github-credentials
+        name: github-credentials-openshift-ci-robot-private-git-cloner
+        readOnly: true
+      - mountPath: /etc/pull-secret
+        name: pull-secret
+        readOnly: true
+      - mountPath: /etc/report
+        name: result-aggregator
+        readOnly: true
+    serviceAccountName: ci-operator
+    volumes:
+    - name: boskos
+      secret:
+        items:
+        - key: credentials
+          path: credentials
+        secretName: boskos-credentials
+    - name: ci-pull-credentials
+      secret:
+        secretName: ci-pull-credentials
+    - name: cluster-profile
+      secret:
+        secretName: cluster-secrets-aws-qe
+    - name: github-credentials-openshift-ci-robot-private-git-cloner
+      secret:
+        secretName: github-credentials-openshift-ci-robot-private-git-cloner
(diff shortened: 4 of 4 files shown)


lwan-wanglin:
@aleskandro @yunjiang29 @liangxia @kuiwang02 could you help review the PR, thanks

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__arm64-nightly.yaml
```diff
@@ -183,6 +183,23 @@ tests:
     test:
     - chain: openshift-e2e-test-qe
     workflow: cucushift-installer-rehearse-aws-ipi-sdn
+- as: e2e-aws-ipi-disconnected-private-p2
+  cron: 0 16 */2 * *
```
liangxia, line 187: p2 with every 2 days ?
Refer to https://github.com/openshift/release/blob/master/ci-operator/config/openshift/openshift-tests-private/README.md#job-frequency

lwan-wanglin: thanks, will change to cron: `0 16 */7 * *`

ci-operator/step-registry/mirror-images/by-oc-adm/mirror-images-by-oc-adm-commands.sh
```diff
@@ -20,15 +20,32 @@ if [ ! -f "${SHARED_DIR}/mirror_registry_url" ]; then
 fi
 echo "MIRROR_REGISTRY_HOST: $MIRROR_REGISTRY_HOST"
 
+# ARM clusters need to mirror image from ${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE}
```
aleskandro, line 23: I'd use the value of OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE, avoiding specific code for the arm64 cluster architecture.

That is changing all the lines from 23 to 39 to:
```
if [[ -n "$OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE" ]]; then
  source_release_image="${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE}"
else
  source_release_image="${RELEASE_IMAGE_LATEST}"
fi
```

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__arm64-nightly.yaml
```diff
@@ -183,6 +183,23 @@ tests:
     test:
     - chain: openshift-e2e-test-qe
     workflow: cucushift-installer-rehearse-aws-ipi-sdn
+- as: e2e-aws-ipi-disconnected-private-p2
```
jianlinliu, line 186: Should we add some "arm" keyword here?

lwan-wanglin: yes, a good point. @aleskandro  is it ok that I change all the profiles name with arm keyword in file openshift-openshift-tests-private-release-4.12__arm64-nightly.yaml in this PR

lwan-wanglin: @jianlinliu Checked with Alessandro, the run will add arm64 keywords in job name automatically and arm64 profiles are in a separate file, it's ok to keep it as it is

ci-operator/step-registry/mirror-images/by-oc-adm/mirror-images-by-oc-adm-ref.yaml
```diff
@@ -12,5 +12,15 @@ ref:
   - namespace: test-credentials
     name: openshift-custom-mirror-registry
     mount_path: /var/run/vault/mirror-registry
+  env:
+  - name: OCP_ARCH
```
aleskandro, line 16: This is not needed anymore right?

lwan-wanglin: thanks, forgot to remove it

lwan-wanglin:
the above failed 4.12 jobs for amd64 and arm64 all because there are failed test cases in cucushift-e2e, the installation part is successful

failed upgrade jobs seem flaky

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__arm64-nightly.yaml
liangxia: Since the file name is *_arm64-*, so the job name is "periodic-ci-openshift-openshift-tests-private-release-4.12-arm64-nightly-e2e-aws-ipi-disconnected-private-p2", which combines specific prefix `periodic-ci` + file name + job name. It already indicates it's for arm

ci-operator/step-registry/mirror-images/by-oc-adm/mirror-images-by-oc-adm-commands.sh
```diff
@@ -20,15 +20,21 @@ if [ ! -f "${SHARED_DIR}/mirror_registry_url" ]; then
 fi
 echo "MIRROR_REGISTRY_HOST: $MIRROR_REGISTRY_HOST"
 
+if [[ -n "$OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE" ]] && [[ X"$RELEASE_IMAGE_LATEST" != X"$OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE" ]]; then
```
liangxia, line 23: The `[[ X"$RELEASE_IMAGE_LATEST" != X"$OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE" ]]` is a redundant check, which is not needed.
You can check with "OVERRIDE=aaa, LATEST=bbb", and "OVERRIDE=ccc, LATEST=ccc"

lwan-wanglin: thanks, in this case, if OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE is non null, then will use OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE as source,  so removing [[ X"$RELEASE_IMAGE_LATEST" != X"$OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE" ]]

lwan-wanglin:
@jianlinliu could you help merge it? thanks

liangxia:
> @jianlinliu could you help merge it? thanks

The PR changed steps in `ci-operator/step-registry/mirror-images`, so need extra approval from `ci-operator/step-registry/mirror-images/OWNERS`

yunjiang29:
PR looks good to me, let's wait for the test complete.

ci-operator/step-registry/mirror-images/by-oc-adm/mirror-images-by-oc-adm-commands.sh
aleskandro: Hello, 

We were discussing of adding that further check (the inequality) to stay safe and because the first one (non empty string) seems always satisfied: e.g., the install step will fail if not https://github.com/openshift/release/blob/master/ci-operator/step-registry/ipi/install/install/ipi-install-install-commands.sh#L248-L252

@liangxia if you are sure they are always the same, shall we remove the if at all and only use `"$OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE"`?

ci-operator/step-registry/mirror-images/by-oc-adm/mirror-images-by-oc-adm-commands.sh
```diff
@@ -20,15 +20,21 @@ if [ ! -f "${SHARED_DIR}/mirror_registry_url" ]; then
 fi
 echo "MIRROR_REGISTRY_HOST: $MIRROR_REGISTRY_HOST"
 
+if [[ -n "$OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE" ]]; then
```
aleskandro, line 23: > The [[ X"$RELEASE_IMAGE_LATEST" != X"$OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE" ]] is a redundant check, which is not needed.
You can check with "OVERRIDE=aaa, LATEST=bbb", and "OVERRIDE=ccc, LATEST=ccc"

Hello,

We were discussing of adding that further check (the inequality) to stay safe and because the first one (non empty string) seems always satisfied: e.g., the install step will fail if OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE is empty https://github.com/openshift/release/blob/master/ci-operator/step-registry/ipi/install/install/ipi-install-install-commands.sh#L248-L252

@liangxia if you are sure they are always the same, shall we remove the if at all and only use "$OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE"?

jianlinliu: I prefer to keep checking `[[ -n "$OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE" ]]`, the logic looks good to me.

aleskandro: Hi @jianlinliu, ok, but...

If we understand/agree on this to also be a redundancy, some other changes that will be needed only to enable testing on arm64 could be avoided. 

The changes from this PR would only be to fix the use of "$OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE" instead of "$RELEASE_IMAGE_LATEST". If the above is true, the installer team, as well as any other guy doing these PRs could use "$OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE" in place of "$RELEASE_IMAGE_LATEST" and there is no need for further debugging by the arm64 team when installations don't work due to this reason. The result would be a more straightforward multi-arch support for the jobs.

What I'm saying is just that is helpful to understand where we should use one or the other variable.

liangxia: Some of the installation only provides RELEASE_IMAGE_LATEST, that's why `[[ -n "$OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE" ]]` is still needed.

aleskandro: RELEASE_IMAGE_LATEST is provided by the ci-operator and is set to be the pullspec corresponding to "release:latest". Nobody overrides it in the current /config.

OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE, in this job, is set to the same value by default (in the yaml spec of the step below). It will never be empty unless someone forces it (and this is a user error that should cause a bunch of other errors down the road).

The ci-operator will fill both the variables when executing the mirroring to the same value.

Is there a case where this variable will be filled with an empty value in this step?

https://github.com/openshift/release/blob/ae170644a22bc6ccf21de1b1a0ab05b148c0d86d/ci-operator/step-registry/mirror-images/by-oc-adm/mirror-images-by-oc-adm-ref.yaml#L15-L19

lwan-wanglin: yes, the OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE will never be empty if set dependencies in steps, got an example [/ipi-install-install-aws-ref.yaml](https://github.com/openshift/release/blob/master/ci-operator/step-registry/ipi/install/install/aws/ipi-install-install-aws-ref.yaml#L18-L22) and [ipi-install-install-aws-commands.sh](https://github.com/openshift/release/blob/master/ci-operator/step-registry/ipi/install/install/aws/ipi-install-install-aws-commands.sh#L65-L68)

yunjiang29: @aleskandro I know what you mean here, yes, we do not need to check `OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE` in this case.

yunjiang29: One interesting thing is,  `OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE` is more straightforward for multi-arch team, `RELEASE_IMAGE_LATEST` is more straightforward for installer (x86_64)

ci-operator/step-registry/mirror-images/by-oc-adm/mirror-images-by-oc-adm-ref.yaml
```diff
@@ -12,5 +12,10 @@ ref:
   - namespace: test-credentials
     name: openshift-custom-mirror-registry
     mount_path: /var/run/vault/mirror-registry
+  dependencies:
+  - name: "release:latest"
+    env: OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE
+  - name: "release:latest"
+    env: RELEASE_IMAGE_LATEST
```
yunjiang29, line 19: @lwan-wanglin I think this env can be removed

lwan-wanglin: @yunjiang29  Checked with @aleskandro in private before, we think can keep it compliant with the rest of the codebase. there are many place set the both, do you recommend to remove it?

ci-operator/step-registry/mirror-images/by-oc-adm/mirror-images-by-oc-adm-commands.sh
aleskandro: @yunjiang29, the issue is that RELEASE_IMAGE_LATEST should be the right one to use (similar to the  installer_payload_image that we have in flexy), and I agree. It is automatically provided by the ci-operator based on the values in the `releases` stanza of the config yaml.

However, this dependency is used by many other steps throughout this repo, and until we (and MA eng.) will fix all the steps to be multi-arched and/or run on clusters like the [arm01](https://github.com/openshift/release/tree/master/clusters/build-clusters/arm01), it seems hard to use "RELEASE_IMAGE_LATEST".

Also note that the installer team also uses OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE in the main installer steps, both [ipi](https://github.com/openshift/release/tree/master/ci-operator/step-registry/ipi) and [upi](https://github.com/openshift/release/tree/master/ci-operator/step-registry/upi).

ci-operator/step-registry/mirror-images/by-oc-adm/mirror-images-by-oc-adm-ref.yaml
aleskandro: I'm just not sure it is ok, there is to be a reason for which the installer jobs use it (at least a historical reason).

We can try removing it though.

lwan-wanglin:
@yunjiang29 @liangxia @jianlinliu @aleskandro Fixed all the comments issue, checked most of  jobs installation part has been success, please help review again.

aleskandro:
Thanks @lwan-wanglin

/lgtm

jianlinliu:
/lgtm

But seem like this PR need rebase

jianlinliu:
/hold 

wait for jobs get completed.

jianlinliu:
@lwan-wanglin btw, do you mind to squash multiple commits into one after the test jobs get completed.

lwan-wanglin:
the jobs has completed, and multiple commits has squashed into one.

most of jobs installation part is successful, azure jobs always failed with error `Can not perform requested operation on nested resource when Linking VNet to private DNS zone`, but it's not introduced by this pr.

@aleskandro need your lgtm

PR #32444, opened 2022-09-20
Restore base-machine-with-rpms stanza

soltysh:


ingvagabund:
@soltysh will not this change get undone by the bot again? As in https://github.com/openshift/release/pull/32435?

PR #32452, opened 2022-09-20
Remove pr-reminder for efried

2uasimojo:
I tried it. Neat idea, but I wasn't getting anything out of it over my existing methods.


smg247:
> I tried it. Neat idea, but I wasn't getting anything out of it over my existing methods.

Is there anything else it could have done that would make it more useful for you?

/lgtm

2uasimojo:
> Is there anything else it could have done that would make it more useful for you?

I think a win would be (at least the option to) include only PRs `/assign`ed to me. Another might be (again, optionally) excluding draft PRs or those with a `/hold`. Though even with those tweaks, I'm not sure it would add anything to my existing methods (see below). Also, the things described in this paragraph could be done via a GH URI with a carefully-constructed querystring, which would have the added advantage of being realtime/on-demand rather than a once-a-day snapshot.

(FYI: My current technique is via a folder in my gmail into which a filter sends all GH-related messages. I use "unread" to indicate something needs my attention: either a new message came in on that PR that I at least need to look at; or if the message indicates I need to do a review and I can't get to it right now I'll explicitly mark it unread. So at any given time when I look at that folder, bold indicates my current "queue".)

/test core-valid

smg247:
Those are very good ideas. Some of those ideas are in the backlog already, and I will add the others. Thanks for the input. Maybe in the future you can use it again.

/lgtm

PR #32460, opened 2022-09-20
CVP-3248: Update the apply kubeobjects when namespace is found.

samvarankashyap:
$subject

### DIFF
--- a/ci-operator/step-registry/optional-operators/cvp-common/apply-secrets/optional-operators-cvp-common-apply-secrets-commands.sh
+++ b/ci-operator/step-registry/optional-operators/cvp-common/apply-secrets/optional-operators-cvp-common-apply-secrets-commands.sh
@@ -94,9 +94,16 @@ LIST_OF_KUBEOBJECTS=$(find "${CUSTOM_KUBEOBJECTS_PATH}"* -type f -name kube_obje
 if [ -n "${LIST_OF_KUBEOBJECTS[0]}" ] ; then
     for i in "${LIST_OF_KUBEOBJECTS[@]}"
     do
-        echo "The following custom kubeobject has been found ! $i"
-        echo "Applying kube_objects on to the Namespace $OO_INSTALL_NAMESPACE"
-        oc apply -f "$i" -n "$OO_INSTALL_NAMESPACE"
+	echo "The following custom kubeobject has been found ! $i"
+	echo "Check if the namespace is mentioned inside the kubeobjects"
+        if oc apply -f "$i" --dry-run=client | grep  "^namespace/"
+        then
+            echo "Namespace found in kubeobjects. Applying directly"
+	    oc apply -f "$i"
+        else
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/optional-operators/cvp-common/apply-secrets/optional-operators-cvp-common-apply-secrets-commands.sh
```diff
@@ -94,9 +94,16 @@ LIST_OF_KUBEOBJECTS=$(find "${CUSTOM_KUBEOBJECTS_PATH}"* -type f -name kube_obje
 if [ -n "${LIST_OF_KUBEOBJECTS[0]}" ] ; then
     for i in "${LIST_OF_KUBEOBJECTS[@]}"
     do
-        echo "The following custom kubeobject has been found ! $i"
-        echo "Applying kube_objects on to the Namespace $OO_INSTALL_NAMESPACE"
-        oc apply -f "$i" -n "$OO_INSTALL_NAMESPACE"
+	echo "The following custom kubeobject has been found ! $i"
+	echo "Check if the namespace is mentioned inside the kubeobjects"
+        if grep -inq "namespace" $i
```
lslebodn, line 99: what is a benefit of using parameter `-n` in quiet mode `-q` ?
```
       -n, --line-number
              Prefix  each  line of output with the 1-based line number within
              its input file.
```

ci-operator/step-registry/optional-operators/cvp-common/apply-secrets/optional-operators-cvp-common-apply-secrets-commands.sh
```diff
@@ -94,9 +94,16 @@ LIST_OF_KUBEOBJECTS=$(find "${CUSTOM_KUBEOBJECTS_PATH}"* -type f -name kube_obje
...
+        then
+            echo "Namespace found in kubeobjects. Applying directly"
+	    oc apply -f "$i"
```
lslebodn, line 102: What if namespace was mentioned in comment or in some text field?
Should we really skip `-n "$OO_INSTALL_NAMESPACE"` in such case?

nmars: What @lslebodn means is, the case-insensitive string `namespace` could appear in a comment or a text field and so not necessarily indicate there is a `Namespace` object contained within the yaml document. 

I suggest we grep for a pattern like `kind:\s+Namespace`  or something like that so we can be more certain there is a `Namespace` object defined there.

ci-operator/step-registry/optional-operators/cvp-common/apply-secrets/optional-operators-cvp-common-apply-secrets-commands.sh
nmars: It should be `grep -iq` since the `-n` prints the line number, but `-q` suppresses the output.

PR #32462, opened 2022-09-20
Add aws-ovn-serial as optional presubmit to CNO

jluhrsen:
- it's also 'always_run: false' so it will need to be triggered via GH comment


jluhrsen:
https://issues.redhat.com/browse/OCPBUGS-1083

PR #32464, opened 2022-09-20
Failover from build02

bear-redhat:
Announcement: https://coreos.slack.com/archives/CFUGK0K9R/p1663695093645129

```
./hack/failover.sh --disable-cluster build02
make update
```


jupierce:
build02 has stabilized. No need to failover.

PR #32466, opened 2022-09-20
SPLAT-776: enable e2e-vsphere-zones for vsphere problem detector

bostrt:


bostrt:
The jobs failed but installation went okay!

gnufied:
How close are we from making these jobs mandatory? Will they be perma-failing for now?

jcpowermac:
> How close are we from making these jobs mandatory? Will they be perma-failing for now?

Not any time soon, we don't have the capacity to do that.

gnufied:
Does it make sense to have periodic job then? I am trying to understand - why have a job which will always fail as pre-submit hook. I know we have had jobs like these before (*cough* *cough* - migration jobs), but it will likely get ignored right?

jcpowermac:
> Does it make sense to have periodic job then? I am trying to understand - why have a job which will always fail as pre-submit hook. I know we have had jobs like these before (_cough_ _cough_ - migration jobs), but it will likely get ignored right?

Its set `optional`, should only run when you request.

rvanderp3:
> Does it make sense to have periodic job then? I am trying to understand - why have a job which will always fail as pre-submit hook. I know we have had jobs like these before (_cough_ _cough_ - migration jobs), but it will likely get ignored right?

we would like a way to test changes to the vsphere-problem-detector from within the PR.  we are working on improving the signal of the test, but this gives us some assurance that our change is behaving as we expect.

gnufied:
okay lets see how it goes. We can always revert if this turns out to be a problem.

/lgtm
/approve

PR #32480, opened 2022-09-21
telco5g: cnf-tests: TEMP: w/a to deploy sriov on 4.12 OCP

sabinaaledort:
sriov-operator can't be deployed on OCP 4.12 due to security changes. This should be fixed in OLM and land in 4.12 nightly. Doing this w/a so we can still test sriov 4.12.


sabinaaledort:
We decided to keep this lane failing, to see the OLM fix.

PR #32485, opened 2022-09-21
Bump AWS EBS CSI driver go to 1.19

jsafrane:
It is required by the new CSI driver version. https://github.com/openshift/aws-ebs-csi-driver/pull/206 fails with 1.18


bertinatto (commented):
/lgtm

PR #32486, opened 2022-09-21
add aws-outpost support

AsherShoshan:


ci-operator/step-registry/ipi/conf/aws/outpost/ipi-conf-aws-outpost-commands.sh
```diff
@@ -0,0 +1,329 @@
...
+
+CONFIG="${SHARED_DIR}/install-config.yaml"
+PATCH=/tmp/install-config-outpost.yaml.patch
+
+REGION="${REGION:-$LEASED_RESOURCE}"
+
+CLUSTER_NAME="$(yq-go r "${CONFIG}" 'metadata.name')"
+
+
+OUTPOST_ID=$(aws --region "${REGION}" outposts list-outposts | jq -r .Outposts[0].OutpostId)
```
yunjiang29, line 22: It seems there is too much to do in this step, according to the [principle of Multi-Stage](https://docs.ci.openshift.org/docs/architecture/step-registry/), we need to split them into separate functions. 
* get Outpost resources/info, (e.g. create a new step `aws-outpost-get-info`)
* generate your VPC template and create VPC , if it is similar to [official VPC template](https://github.com/openshift/installer/blob/master/upi/aws/cloudformation/01_vpc.yaml), you can update existing step `aws-provision-vpc-shared` to meet your needs, or if you think it's hard to merge into existing one, you can create a new step, like `aws-provision-vpc-outpost` (the `aws` folder contains the steps for creating/getting AWS resources that are required by creating a cluster, like preparing a VPC for cluster.)
* Prepare install config and manifests for AWS Outpost, `ipi-conf-aws-outpost` is the right place.
* Create manifests and cluster, I saw you are using `  - chain: ipi-install`, but why you created manifests in this step before invoke `ipi-install` chain.

After that, your workflow will be like
```
  steps:
  - ref: aws-outpost-get-info
  - ref: aws-provision-vpc-outpost
  - ref: ipi-conf
  - ref: ipi-conf-telemetry
  - ref: ipi-conf-aws
  - ref: ipi-conf-aws-outpost
  - chain: ipi-install
  ...
```

AsherShoshan: aws-provision-vpc-shared --> I have different template, and different parameters for the vpc
"Create manifests and cluster, I saw you are using - chain: ipi-install, but why you created manifests in this step before invoke ipi-install chain"  --> I need to change subnets in manifests.  what other chain/step should I use instead of 'ipi-install'?

yunjiang29: > I need to change subnets in manifests. what other chain/step should I use instead of 'ipi-install'?

Once you created VPC, you [write](https://github.com/openshift/release/blob/master/ci-operator/step-registry/aws/provision/vpc/shared/aws-provision-vpc-shared-commands.sh#L52-L72) subnets infos into `SHARED_DIR`,  and then `ipi-conf-aws-custom-vpc` step will [read](https://github.com/openshift/release/blob/master/ci-operator/step-registry/ipi/conf/aws/custom-vpc/ipi-conf-aws-custom-vpc-commands.sh#L12) them and patch the install-config.

In an IPI install, `create cluster` commands must be executed by `ipi-install-install`

yunjiang29: for manifests files, `ipi-install-install` will [scan](https://github.com/openshift/release/blob/master/ci-operator/step-registry/ipi/install/install/ipi-install-install-commands.sh#L315-L322) `SHARED_DIR` directory and copy manifests to install dir.

AsherShoshan: "Once you created VPC, you [write](https://github.com/openshift/release/blob/master/ci-operator/step-registry/aws/provision/vpc/shared/aws-provision-vpc-shared-commands.sh#L52-L72) subnets infos into SHARED_DIR, and then ipi-conf-aws-custom-vpc step will [read](https://github.com/openshift/release/blob/master/ci-operator/step-registry/ipi/conf/aws/custom-vpc/ipi-conf-aws-custom-vpc-commands.sh#L12) them and patch the install-config."
This is exactly what I'm doing.  getting the subnets and patching the install-config.yaml.  why to write it into SHARED_DIR. I don't understand ur comment.

AsherShoshan: "In an IPI install, create cluster commands must be executed by ipi-install-install"  - Not understood.  I'm not changing ipi-install chain.   what u mean here?

yigald18:
Hi @yunjiang29 
What is the status of this PR please ?

yunjiang29:
> Hi @yunjiang29
> What is the status of this PR please ?

@AsherShoshan as last offline discussion, there is a cluster profile issue, have you fixed it?

@yigald18 the cluster profile must be ready before testing this PR. @AsherShoshan If your code is ready for review, just let me know.

PR #32493, opened 2022-09-21
openshift/kubernetes: Also override kubelet on rhel-coreos-8

cgwalters:
https://issues.redhat.com/browse/MCO-377

As part of https://github.com/openshift/enhancements/blob/master/enhancements/ocp-coreos-layering/ocp-coreos-layering.md we are introducing a new base image, and the CI here needs to learn to also inject the new kubelet binary on top of it.

But notice how much saner and nicer this new flow is!

Once we drop usage of the old machine-os-content from the MCO, then we can drop the old flow here too.

### DIFF
--- a/ci-operator/config/openshift/kubernetes/openshift-kubernetes-master.yaml
+++ b/ci-operator/config/openshift/kubernetes/openshift-kubernetes-master.yaml
@@ -27,6 +27,10 @@ base_images:
     name: builder
     namespace: ocp
     tag: rhel-8-golang-1.19-openshift-4.12
+  rhel-coreos-8-base:
+    name: "4.12"
+    namespace: ocp
+    tag: rhel-coreos-8
   tools:
     name: "4.12"
     namespace: ocp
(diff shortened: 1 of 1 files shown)


cgwalters:
OK, I think this is good to go!

sinnykumari:
@deads2k @mfojtik PTAL as this will be good to have when we switch to newer os image content with MCO PR https://github.com/openshift/machine-config-operator/pull/3317

ashcrow:
This is important for our ability to push layering forward. Could we get a review?

haircommander:
/lgtm

(from node team)

rphillips:
lgtm

cgwalters:
This needs an /approve too

cgwalters:
OK listen folks.  There's been a WHOLE LOT OF WORK that went into https://github.com/openshift/machine-config-operator/pull/3317 and stuff before it and having this PR open for *an entire week* when it's not going to break anything is a bit frustrating.

Please.  Pretty please.  Someone drop an `/approve`.

haircommander:
am I powerful enough

/approve

ashcrow:
> This branch has conflicts that must be resolved

Do we need a rebase before the @soltysh merges?

cgwalters:
rebased :surfer:

deads2k:
looks like rehearses have worked
/approve

cgwalters:
I know this can be confusing so I want to be clear: This PR is going to basically do nothing except modify an image that's *not used by default*, so it really, really should be 100% safe to just merge right now.

But, after https://github.com/openshift/machine-config-operator/pull/3317 lands the situation will flip - the `machine-os-content` will go unused, and `rhel-coreos-8` will be used by default, and that's when the changes in this PR will actually be used.

PR #32508, opened 2022-09-21
installer: Make e2e-agent-compact job required

zaneb:
This job is now passing, so make it required for changes to the agent
installer.

Update the file skip patterns to avoid unnecessary job runs.

### DIFF
--- a/ci-operator/config/openshift/installer/openshift-installer-master.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-master.yaml
@@ -432,7 +432,7 @@ tests:
     workflow: openshift-e2e-gcp-libvirt
 - as: e2e-metal-assisted
   optional: true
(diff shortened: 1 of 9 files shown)


barbacbd (commented):
/lgtm 

Had to go and find the answers to some of the questions I had. 
Do we definitely want to `skip_if_only_changed` on the openshift install command?

barbacbd:
It does appear that some of the changed tests are failing

zaneb:
> Do we definitely want to `skip_if_only_changed` on the openshift install command?

It's to skip changes on the directory `cmd/openshift-install/agent`.

> It does appear that some of the changed tests are failing

Looks like one hit a timeout. We want to get the time down, but can always bump the timeout again if that's just how long it takes.

PR #32519, opened 2022-09-22
MGMT-11638: assisted-installer: Update master release configurations file to point to 4.12

eliorerz:
/cc @osherdp @adriengentil

### DIFF
--- a/ci-operator/config/openshift/assisted-service/openshift-assisted-service-master__edge.yaml
+++ b/ci-operator/config/openshift/assisted-service/openshift-assisted-service-master__edge.yaml
@@ -178,13 +178,13 @@ releases:
       include_built_images: true
       name: "4.12"
       namespace: ocp
-  stable-4-9:
-    integration:
-      name: "4.9"
-      namespace: ocp
   stable-4-10:
     integration:
       name: "4.10"
+      namespace: ocp
+  stable-4-11:
+    integration:
+      name: "4.11"
       namespace: ocp
 resources:
   '*':
@@ -449,7 +449,7 @@ tests:
   steps:
     cluster_profile: packet-assisted
     dependencies:
-      ASSISTED_OPENSHIFT_INSTALL_RELEASE_IMAGE: release:latest
+      ASSISTED_OPENSHIFT_INSTALL_RELEASE_IMAGE: release:stable-4-10
       OPENSHIFT_INSTALL_RELEASE_IMAGE: release:latest
     env:
       DISCONNECTED: "true"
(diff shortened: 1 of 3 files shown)


ci-operator/config/openshift/assisted-service/openshift-assisted-service-master__edge.yaml
```diff
@@ -217,7 +217,7 @@ releases:
   latest:
     integration:
       include_built_images: true
-      name: "4.10"
```
osherdp, line 220: it might mean for disconnected job to start failing on master
would we want to handle it? (e.g., keep it using 4.10 for now)

osherdp (changes requested):

ci-operator/config/openshift/assisted-service/openshift-assisted-service-master__edge.yaml
mkowalski: Why? Is disconnected installation not supported in 4.12 anymore? Or is there a bug that we should be aware of?

osherdp: because of a bug we have in 4.11 (and I think should exist in 4.12)
I think the bug is already filed, but not handled iirc

eliorerz: Maybe we can specify that disconnected job will run in 4.10 instead of latest?

osherdp: sounds good to me (as a temporal workaround)

eliorerz: Do you remember if both `e2e-ai-operator-ztp-disconnected-periodic` and `e2e-ai-operator-ztp-disconnected` need to be changed to 4.10?

osherdp: yes both of them need a <4.11 based version

ci-operator/jobs/openshift/assisted-service/openshift-assisted-service-master-periodics.yaml
```diff
@@ -337,7 +337,160 @@ periodics:
     ci-operator.openshift.io/variant: edge
     ci.openshift.io/generator: prowgen
     pj-rehearse.openshift.io/can-be-rehearsed: "true"
-  name: periodic-ci-openshift-assisted-service-master-edge-e2e-ai-operator-ztp-ipv4v6-sno-ocp-410-periodic
+  name: periodic-ci-openshift-assisted-service-master-edge-e2e-ai-operator-ztp-ipv4v6-3masters-periodic
+  spec:
```
osherdp, line 341: we're missing here the notification

osherdp (changes requested):
just one little thing

osherdp (approved):

PR #32521, opened 2022-09-22
Use the index-image with its tagged image instead of its digest form

osherdp:


eranco74:
```
[2022-09-28 12:17:56] + virsh attach-disk ostest_worker_1 /tmp/ostest_worker_1-sdb.img sdb --wwn 0x02564d22a39983ba
[2022-09-28 12:17:56] error: Failed to attach disk
[2022-09-28 12:17:56] error: Requested operation is not valid: Domain already contains a disk with that address
```
/retest

osherdp:
@eranco74 this is approximately a dupe of https://github.com/openshift/release/pull/32685
let's hope that other PR works, if so then it wouldn't be different than fixing capi-ztp flow

PR #32527, opened 2022-09-22
Fix sprint-automation kubeconfig generation

danilo-gemoli:
Periodic `periodic-sprint-automation` is failing, I suspect it was just a copy 'n paste mistake from the stanza above.

/cc @hongkailiu


hongkailiu (approved):
Thanks.

PR #32528, opened 2022-09-22
Point release:latest to AMD payload for ARM tests

liangxia:
It is tricky that we setup a arm64 cluster in a amd64 based pod. But this is how we make it work at the moment. Explicitly point `release:latest` to amd64.

/cc @aleskandro @jhou1 @pruan-rht @dis016 @JianLi-RH

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.11__arm64-nightly.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.11__arm64-nightly.yaml
@@ -33,6 +33,7 @@ releases:
       version: "4.11"
   latest:
     candidate:
+      architecture: amd64
       product: ocp
       stream: nightly
       version: "4.11"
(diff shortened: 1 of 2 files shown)


aleskandro:
Hello @liangxia just noticing this now. Please note that changing the "release:latest" image to the arm64 one will lead to many binaries not running due to the wrong architecture. The jobs run in amd64 clusters. We cannot do that switch until we get other steps around the release repo working and the arm01 build cluster ready. This is how ARM64 CI eng. is running testing today, having both the images imported for each job and letting the tests use different ways to gather the payload to use.

cc @deepsm007

liangxia:
> Hello @liangxia just noticing this now. Please note that changing the "release:latest" image to the arm64 one will lead to many binaries not running due to the wrong architecture. The jobs run in amd64 clusters. We cannot do that switch until we get other steps around the release repo working and the arm01 build cluster ready. This is how ARM64 CI eng. is running testing today, having both the images imported for each job and letting the tests use different ways to gather the payload to use.
> 
> cc @deepsm007

@aleskandro Make sense.
Update the PR to explicitly point `release:latest` to amd64

aleskandro (approved):

aleskandro:
Thanks @liangxia , definitely better for readability.

/lgtm

PR #32537, opened 2022-09-22
Fix messages and comments in telco5g job

sshnaidm:


ci-operator/step-registry/telco5g/cluster-setup/telco5g-cluster-setup-commands.sh
```diff
@@ -5,7 +5,7 @@ set -o errexit
 set -o pipefail
 
 echo "************ telco cluster setup command ************"
-# TODO: Remove once OpenShift CI will be upgraded to 4.2 (see https://access.redhat.com/articles/4859371)
+# Fix user IDs in a container
```
yuvalk, line 8: cant we get rid of it now?

sshnaidm: Nope. Still doesn't work w/o this.

yuvalk:
But you removed the todo ..

PR #32541, opened 2022-09-22
mtu workflows

anuragthehatter:
cc @openshift/team-sdn-qe @lihongan @liangxia 
I think we need to manually define ```cron``` values?

@zhaozhanqi Seems like p3 for SDN and p2 for OVN should be okay

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-nightly.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-nightly.yaml
@@ -678,6 +678,15 @@ tests:
     test:
     - chain: openshift-e2e-test-qe
     workflow: cucushift-installer-rehearse-gcp-upi-proxy
+- as: e2e-gcp-ipi-ovn-mtu-migrate-p2
+  steps:
+    cluster_profile: gcp-qe
+    env:
+      E2E_RUN_TAGS: '@gcp-ipi and @network-ovnkubernetes and not @fips'
+      TAG_VERSION: '@4.12'
+    test:
+    - chain: openshift-e2e-test-qe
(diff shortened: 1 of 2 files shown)


zhaozhanqi:
>@zhaozhanqi Seems like p3 for SDN and p2 for OVN should be okay
yes, it's ok for me

anuragthehatter:
@liangxia We probably have to override cucushift e2e failures?

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-nightly.yaml
```diff
@@ -678,6 +678,15 @@ tests:
     test:
     - chain: openshift-e2e-test-qe
     workflow: cucushift-installer-rehearse-gcp-upi-proxy
+- as: e2e-gcp-ipi-ovn-mtu-migrate-p2
```
lihongan, line 681: `cron` is required here I think. For p2 wen can set it as `cron: 0 22 */7 * *`

liangxia:
> @liangxia We probably have to override cucushift e2e failures?

The e2e jobs are optional (non-blocking) for PR merging.

BTW, there are conflicts.

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-nightly.yaml
```diff
@@ -791,6 +800,16 @@ tests:
     test:
     - ref: openshift-extended-test-longduration
     workflow: cucushift-installer-rehearse-aws-ipi-ovn
+- as: e2e-aws-ipi-sdn-mtu-migrate-p3
```
lihongan, line 803: We can set cron for this test as ` cron: 0 18 */28 * *` since it is p3.
BTW, better move the test forward (to before line 348) and place it with other aws tests together.

PR #32549, opened 2022-09-22
Added feature migration test for multicast

nicklesimba:
Signed-off-by: nicklesimba <simha.nikhil@gmail.com>

### DIFF
--- a/ci-operator/step-registry/ovn/sdn-migration/feature-migration/config-multicast/ovn-sdn-migration-feature-migration-config-multicast-commands.sh
+++ b/ci-operator/step-registry/ovn/sdn-migration/feature-migration/config-multicast/ovn-sdn-migration-feature-migration-config-multicast-commands.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+set -x
+set -o errexit
+set -o nounset
+set -o pipefail
+
+
+# Namespace may or may not be created already, creating just in case.
+oc create ns test-migration || true
+
+# Patch the netnamespace to use multicast annotation
+oc annotate netnamespace test-migration netnamespace.network.openshift.io/multicast-enabled=true
--- a/ci-operator/step-registry/ovn/sdn-migration/feature-migration/config-multicast/ovn-sdn-migration-feature-migration-config-multicast-ref.yaml
+++ b/ci-operator/step-registry/ovn/sdn-migration/feature-migration/config-multicast/ovn-sdn-migration-feature-migration-config-multicast-ref.yaml
@@ -1,0 +1,13 @@
+ref:
+  as: ovn-sdn-migration-feature-migration-config-multicast
+  from_image:
+    name: cli-jq
+    namespace: ocp
+    tag: latest
+  commands: ovn-sdn-migration-feature-migration-config-multicast-commands.sh
+  resources:
+    requests:
+      cpu: 10m
+      memory: 100Mi
+  documentation: |-
+    The step configures the multicast feature on a openshift-sdn cluster by setting multicast-enabled annotation to equal true.
--- a/ci-operator/step-registry/ovn/sdn-migration/feature-migration/validate-multicast/ovn-sdn-migration-feature-migration-validate-multicast-commands.sh
+++ b/ci-operator/step-registry/ovn/sdn-migration/feature-migration/validate-multicast/ovn-sdn-migration-feature-migration-validate-multicast-commands.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+set -x
+set -o errexit
+set -o nounset
+set -o pipefail
+
+validate_namespace_multicast_annotation () {
+  current_config_unformatted=$(kubectl get namespace test-migration -o json | jq .metadata.annotations | grep multicast-enabled)
+  current_config="$(echo -e "${current_config_unformatted}" | sed -e 's/^[[:space:]]*//')"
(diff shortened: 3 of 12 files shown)


nicklesimba:
@jluhrsen can I get a review on this one as well? It's targeted for this sprint.

pliurh:
You need to also create the `*.metadata.json` file for the new steps you add.

ci-operator/step-registry/ovn/sdn-migration/feature-migration/config-multicast/ovn-sdn-migration-feature-migration-config-multicast-commands.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+set -x
+set -o errexit
+set -o nounset
+set -o pipefail
+
+
+# I'm unsure if netnamespace will already exist, maybe i only need to patch it.
+oc create ns test-migration
```
jluhrsen: wanna add your "|| true" here as well? and remove the "maybe" part of the comment. just note that you
aren't sure if it's there or not so just doing it anyway, since it can't hurt

ci-operator/step-registry/ovn/sdn-migration/feature-migration/validate-multicast/ovn-sdn-migration-feature-migration-validate-multicast-commands.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+set -x
+set -o errexit
+set -o nounset
+set -o pipefail
+
+validate_namespace_multicast_annotation () {
+  kubectl get namespace test-migration -o json | jq .metadata.annotations | grep multicast-enabled | tee current_config
+  echo "$EXPECT_NAMESPACE_MULTICAST" | tee expected_config
+  if diff <(jq -S . current_config) <(jq -S . expected_config); then
```
jluhrsen, line 10: I think this is giving an error:

++ jq -S . expected_config
parse error: Expected string key before ':' at line 1, column 32
parse error: Expected string key before ':' at line 1, column 34

even if the error is meaningless and doesn't affect anything, it's scary to look at. is there a way
to not have it?

ci-operator/step-registry/ovn/sdn-migration/feature-migration/config-multicast/ovn-sdn-migration-feature-migration-config-multicast-commands.sh
```diff
@@ -6,7 +6,7 @@ set -o pipefail
 
 
 # I'm unsure if netnamespace will already exist, maybe i only need to patch it.
-oc create ns test-migration
+oc create ns test-migration || true
```
jluhrsen, line 9: oh, ha! it's here. should have scrolled through all the commits first

ci-operator/step-registry/ovn/sdn-migration/feature-migration/config-multicast/ovn-sdn-migration-feature-migration-config-multicast-commands.sh
nicklesimba: Leaving this unresolved for now so I don't forget to remove that "maybe" part of the comment later...

ci-operator/step-registry/ovn/sdn-migration/feature-migration/validate-multicast/ovn-sdn-migration-feature-migration-validate-multicast-commands.sh
nicklesimba: Honestly, I'm not sure. It _does_ seem to be a meaningless error, but I agree it is scary to look at. Leaving this unresolved because it's a lower priority thing to fix, but I still want to come back and fix this.

nicklesimba: I fixed this in my most recent commit. I just formatted the string to follow the style used in the rollback/egressFirewall e2e's.

nicklesimba: @jluhrsen I think this is actually causing failures/issues that are larger than just a scary error message. I'm going to revert to the previous approach for now. I think it should be okay because the tests themselves were passing fine then.

ci-operator/step-registry/ovn/sdn-migration/feature-migration/config-multicast/ovn-sdn-migration-feature-migration-config-multicast-ref.yaml
```diff
@@ -0,0 +1,13 @@
...
+    name: cli-jq
+    namespace: ocp
+    tag: latest
+  commands: ovn-sdn-migration-feature-migration-config-multicast-commands.sh
+  resources:
+    requests:
+      cpu: 10m
+      memory: 100Mi
+  documentation: |-
+    The step configures the egress firewall feature on a openshift-sdn cluster by creating a egressNetworkPolicy CR.
```
nicklesimba, line 13: Note to self: this documentation is incorrect for multicast.

PR #32564, opened 2022-09-23
WIP MGMT-12115: Assisted installer: add optional ZTP test

vrutkovs:
Add ad-hoc ztp test to make sure assisted-installer changes don't break ZTP flow.

The second commit adds support to re-use ci-index built by latest assisted-service commit. In order to use it we have to introduce a new env var:
* the index cannot be imported as "ci-index" (ci-operator disallows that), so we have to import it with a different name
* we can't use ci-operator substituted pullspec for disconnected jobs, so we have to construct image name using IMAGE_INDEX_TAG env var

### DIFF
--- a/ci-operator/config/openshift/assisted-installer/openshift-assisted-installer-master__edge.yaml
+++ b/ci-operator/config/openshift/assisted-installer/openshift-assisted-installer-master__edge.yaml
@@ -217,6 +217,8 @@ tests:
       ASSISTED_OPENSHIFT_INSTALL_RELEASE_IMAGE: release:latest
       INDEX_IMAGE: assisted-service-index
       OPENSHIFT_INSTALL_RELEASE_IMAGE: release:latest
+    env:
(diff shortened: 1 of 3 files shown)


vrutkovs:
/hold

for rehearsal (which is expected to fail)

vrutkovs:
```
'unpack job not completed: Unpack pod(openshift-marketplace/9a488701bdf4369d70e385f86c89b4dd34f8128d5f0ca96bdb5144910f9rhrh)
      container(pull) is pending. Reason: ErrImagePull, Message: rpc error: code =
      Unknown desc = reading manifest sha256:1a9af3216636d6de7ecd53ffe84cafb46f9996199fb807c7626a2e5dd345ab40
      in registry.build02.ci.openshift.org/ci-op-wrzhdvk0/pipeline: unauthorized:
      authentication required.'
```

Subscription cannot be installed - seems cluster is provisioned with a pull secret, which doesn't have a record for CI cluster

omertuc:
/uncc @omertuc 

re-cc when ready

PR #32566, opened 2022-09-23
assisted-service: create day2 master ZTP jobs

mkowalski:
Contributes-to: [MGMT-8578](https://issues.redhat.com//browse/MGMT-8578)
Contributes-to: [MGMT-12143](https://issues.redhat.com//browse/MGMT-12143)

### DIFF
--- a/ci-operator/config/openshift/assisted-service/openshift-assisted-service-master__edge.yaml
+++ b/ci-operator/config/openshift/assisted-service/openshift-assisted-service-master__edge.yaml
@@ -476,6 +476,20 @@ tests:
         NUMBER_OF_DAY2_HOSTS=2
       NUM_EXTRA_WORKERS: "5"
     workflow: assisted-baremetal-operator-ztp
+- as: e2e-ai-operator-ztp-compact-day2-masters
+  optional: true
+  steps:
+    cluster_profile: packet-assisted
+    dependencies:
+      ASSISTED_OPENSHIFT_INSTALL_RELEASE_IMAGE: release:latest
+      OPENSHIFT_INSTALL_RELEASE_IMAGE: release:latest
+    env:
+      ASSISTED_ZTP_CONFIG: |
+        SPOKE_CONTROLPLANE_AGENTS=3
+        NUMBER_OF_DAY2_HOSTS=1
+        DAY2_MASTERS=True
+      NUM_EXTRA_WORKERS: "4"
+    workflow: assisted-baremetal-operator-ztp
 - as: e2e-ai-operator-ztp-sno-day2-workers
   steps:
     cluster_profile: packet-assisted
--- a/ci-operator/jobs/openshift/assisted-service/openshift-assisted-service-master-presubmits.yaml
+++ b/ci-operator/jobs/openshift/assisted-service/openshift-assisted-service-master-presubmits.yaml
@@ -356,6 +356,81 @@ presubmits:
     - ^master$
     - ^master-
     cluster: build03
+    context: ci/prow/edge-e2e-ai-operator-ztp-compact-day2-masters
+    decorate: true
+    labels:
+      ci-operator.openshift.io/cloud: packet-edge
+      ci-operator.openshift.io/cloud-cluster-profile: packet-assisted
+      ci-operator.openshift.io/variant: edge
+      ci.openshift.io/generator: prowgen
+      pj-rehearse.openshift.io/can-be-rehearsed: "true"
+    name: pull-ci-openshift-assisted-service-master-edge-e2e-ai-operator-ztp-compact-day2-masters
+    optional: true
+    rerun_command: /test edge-e2e-ai-operator-ztp-compact-day2-masters
+    spec:
+      containers:
+      - args:
+        - --gcs-upload-secret=/secrets/gcs/service-account.json
+        - --image-import-pull-secret=/etc/pull-secret/.dockerconfigjson
+        - --lease-server-credentials-file=/etc/boskos/credentials
+        - --report-credentials-file=/etc/report/credentials
+        - --secret-dir=/secrets/ci-pull-credentials
+        - --secret-dir=/usr/local/e2e-ai-operator-ztp-compact-day2-masters-cluster-profile
+        - --target=e2e-ai-operator-ztp-compact-day2-masters
+        - --variant=edge
+        command:
+        - ci-operator
+        image: ci-operator:latest
+        imagePullPolicy: Always
+        name: ""
+        resources:
+          requests:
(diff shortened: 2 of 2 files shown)


ci-operator/config/openshift/assisted-service/openshift-assisted-service-master__edge.yaml
```diff
@@ -489,6 +503,21 @@ tests:
...
+    dependencies:
+      ASSISTED_OPENSHIFT_INSTALL_RELEASE_IMAGE: release:latest
+      OPENSHIFT_INSTALL_RELEASE_IMAGE: release:latest
+    env:
+      ASSISTED_ZTP_CONFIG: |
+        SPOKE_CONTROLPLANE_AGENTS=1
+        NUMBER_OF_DAY2_HOSTS=1
+        DAY2_MASTERS=True
+      NUM_EXTRA_WORKERS: "2"
+      PACKET_PLAN: m3.large.x86
```
adriengentil, line 519: how many VM does the job spawn? Try with the default machine I think

ci-operator/config/openshift/assisted-service/openshift-assisted-service-master__edge.yaml
```diff
@@ -476,6 +476,20 @@ tests:
...
+  steps:
+    cluster_profile: packet-assisted
+    dependencies:
+      ASSISTED_OPENSHIFT_INSTALL_RELEASE_IMAGE: release:latest
+      OPENSHIFT_INSTALL_RELEASE_IMAGE: release:latest
+    env:
+      ASSISTED_ZTP_CONFIG: |
+        SPOKE_CONTROLPLANE_AGENTS=3
+        NUMBER_OF_DAY2_HOSTS=1
+        DAY2_MASTERS=True
```
omertuc, line 490: Where is this used?

ci-operator/config/openshift/assisted-service/openshift-assisted-service-master__edge.yaml
mkowalski: I don't understand your question, it's defined just in the line you commented - `NUM_EXTRA_WORKERS: "2"`

ci-operator/config/openshift/assisted-service/openshift-assisted-service-master__edge.yaml
mkowalski: In https://github.com/openshift/assisted-service/pull/4437

ci-operator/config/openshift/assisted-service/openshift-assisted-service-master__edge.yaml
adriengentil: You set `PACKET_PLAN` to `m3.large` which is a big and more expensive machine, is there a specific reason? because I see the other job runs more VMs and you did not set `PACKET_PLAN` for it

mkowalski: This job is a dumb copy of `e2e-ai-operator-ztp-sno-day2-workers`. Whatever is set there, I set here. Unless we want to discuss reconfiguration of the other job too, I don't see a reason why one should be different from the other

mkowalski: To be clear

* `e2e-ai-operator-ztp-compact-day2-masters` is blind copy of `e2e-ai-operator-ztp-compact-day2-workers`
* `e2e-ai-operator-ztp-sno-day2-masters` is blind copy of `e2e-ai-operator-ztp-sno-day2-workers`

omertuc: The job spawns 7 VMs - 5 for the hub (3 + 2 dev-scripts default), 2 for spoke (NUM_EXTRA_WORKERS)

mkowalski: I'm not sure the discussion about those numbers really should be happening here. We have current jobs that take **some cluster** and add **some workers** to it. The new job will take **the same cluster** and **instead** of adding some workers it will add some masters.

We are not adding more nodes right now. It's just that previously we were adding workers, now we are adding masters.

>How does it even work? Are you allowed to add a master to a cluster with already 3 masters?

Yes, that's a new feature that we implement as MGMT-8578

mkowalski: Number-wise everything is good. In the compact cluster we have

* `SPOKE_CONTROLPLANE_AGENTS=3`
* `NUMBER_OF_DAY2_HOSTS=1`
* `NUM_EXTRA_WORKERS: "4"`

In the SNO we have

* `SPOKE_CONTROLPLANE_AGENTS=1`
* `NUMBER_OF_DAY2_HOSTS=1`
* `NUM_EXTRA_WORKERS: "2"`

All the numbers add up correctly. We have enough VMs to address all the needs

omertuc: yeah sorry I was mixing up the compact and SNO jobs

omertuc: >How does it even work? Are you allowed to add a master to a cluster with already 3 masters?

However, this question was directed at OCP - not at Assisted

mkowalski: >However, this question was directed at OCP - not at Assisted

Long story short, yes - in the process of replacing a master node in the OCP cluster there is a period of time when you have a cluster with 4 masters. It only an intermediate step that's supposed to be only for a short while, but nevertheless there will be a notion of cluster with 4 masters

omertuc: OK. I'll be surprised if it works with a SNO - guess we'll see

mkowalski: You are right, probably "replacing master" in case of SNO is a nonsense. Let me remove this test because I don't see any benefit

omertuc: It's not nonsense, it's definitely not officially supported though. SNO+Workers is supported since 4.11 and having temporarily 2 SNOs while the older one is being removed is not out of the question, even though it's not something we should be wasting our time on until it's supported

tl;dr not entirely nonsense but unsupported

PR #32574, opened 2022-09-23
Perform ACM GRC SonarCloud analysis in a GitHub action

mprahl:
The ACM GRC repos often have integration tests that make up the bulk of the test coverage. Since the unit and integration tests are run in a GitHub workflow, the easiest path forward is to use the SonarCloud GitHub action to post the coverage.

Relates:
https://github.com/stolostron/backlog/issues/25949


dhaiducek (changes requested):
We won't want to take it out of `2.3` and `2.4` since SonarCloud wasn't implemented in GitHub Actions until `2.5`. And it probably isn't worth backporting the GitHub Action changes to implement SonarCloud to `2.3` and `2.4`?

I will note that `go-log-utils`, `go-template-utils`, and `policy-generator-plugin` also don't have SonarCloud implemented in-repo at all currently, so is that going to change to be consistent with other repos as part of this?

mprahl:
> We won't want to take it out of `2.3` and `2.4` since SonarCloud wasn't implemented in GitHub Actions until `2.5`. And it probably isn't worth backporting the GitHub Action changes to implement SonarCloud to `2.3` and `2.4`?

You're right, it's probably not worth backporting that far since there is no code coverage requirement on older release branches.

> I will note that `go-log-utils`, `go-template-utils`, and `policy-generator-plugin` also don't have SonarCloud implemented in-repo at all currently, so is that going to change to be consistent with other repos as part of this?

Right. I want things to be consistent. I'll implement those changes there once this PR is merged.

dhaiducek:
Looks like we'll need to at least remove the `2.4` configuration for `go-template-utils` also.

I'm curious since it's not clear in the PR: did you remove the `go-log-utils`, `go-template-utils`, and `policy-generator-plugin` directories entirely? That should be safe to do if we're removing all of the files there.

mprahl:
> Looks like we'll need to at least remove the `2.4` configuration for `go-template-utils` also.
> 
> I'm curious since it's not clear in the PR: did you remove the `go-log-utils`, `go-template-utils`, and `policy-generator-plugin` directories entirely? That should be safe to do if we're removing all of the files there.

Thanks for catching that! That was a mistake when I was reverting the 2.3 and 2.4 changes in the PR.

mprahl:
> Looks like we'll need to at least remove the `2.4` configuration for `go-template-utils` also.
> 
> I'm curious since it's not clear in the PR: did you remove the `go-log-utils`, `go-template-utils`, and `policy-generator-plugin` directories entirely? That should be safe to do if we're removing all of the files there.

Thanks for catching that! That was a mistake when I was reverting the 2.3 and 2.4 changes in the PR.

dhaiducek (approved):

PR #32579, opened 2022-09-23
Tentatively add initial IDP before QE go-lang test for OKR

xingxingxia:
For QE OKR we intend to integrate IDPs users in QE's go-lang cases. Two steps: 1. set up IDP; 2. use the IDP users in cases. Current discussion shows the "2" is not supported. But discussion wants to do the "1" first then re-consider how to do "2". So submitting this PR for "1".
**Because we are currently in an exploring stage, I'd like to be as simple as possible** to reuse @liangxia's existing `cucushift-pre`, instead of add a duplicate step before `openshift-extended-test`, so I just adjust the `cucushift-pre` order.
@jianzhangbjz @liangxia please help review/approve, thanks!

### DIFF
--- a/ci-operator/step-registry/openshift/e2e/test/qe/openshift-e2e-test-qe-chain.yaml
+++ b/ci-operator/step-registry/openshift/e2e/test/qe/openshift-e2e-test-qe-chain.yaml
@@ -2,8 +2,8 @@ chain:
   as: openshift-e2e-test-qe
   steps:
   - ref: cucushift-helper-service-build-info
-  - ref: openshift-extended-test
   - ref: cucushift-pre
+  - ref: openshift-extended-test
   - ref: cucushift-e2e
   - ref: openshift-extended-web-tests
   - ref: cucushift-e2e-report


ci-operator/step-registry/openshift/e2e/test/qe/openshift-e2e-test-qe-chain.yaml
```diff
@@ -2,8 +2,8 @@ chain:
   as: openshift-e2e-test-qe
   steps:
   - ref: cucushift-helper-service-build-info
```
xingxingxia, line 4: We can see there is already a "cucushift-" step before `openshift-extended-test`, so the step `cucushift-pre` is fine to be before too.

xingxingxia: And there are many "cucushift-" steps referenced by `openshift-tests-private-` stuff. So this time, I want to be as simple as possible to just adjust the order of cucushift-pre. We can enhance if needed once https://issues.redhat.com/browse/OCPQE-10622 using the IDP users in go-lang cases is finally possible in future @jianzhangbjz.

jianzhangbjz: ok, yes, splitting the `user setting` from this `cucushift-pre` step will follow `Prow` rules, please don't forget do update this in the future.

xingxingxia: Sure. Could you share some link about "Prow rules" for me to more clear refer to?
Another question, `cucushift-helper-service-build-info` is a `cucushift-` step before `openshift-extended-test` step, does this follow or not follow "Prow rules"?
A third question, when I run "grep **cucushift** ci-operator/config/openshift/**openshift-tests-private**/openshift-openshift-tests-private-release-4.12__amd64-nightly.yaml", the result has many "cucushift" occurrences, "cucushift" stuff is referenced by "openshift-tests-private" stuff, does this follow or not follow "Prow rules"?

jianzhangbjz: > Could you share some link about "Prow rules" for me to more clear refer to?

For the step, the simple the better. Generally, the user can know what this step does via reading the step name.

jianzhangbjz: > Another question, cucushift-helper-service-build-info is a cucushift- step before openshift-extended-test step, does this follow or not follow "Prow rules"?

Nothing with the Prow rule, but with the agreement with the team automation gogal: replace cucushift with Ginkgo.

> A third question, when I run "grep cucushift ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-nightly.yaml", the result has many "cucushift" occurrences, "cucushift" stuff is referenced by "openshift-tests-private" stuff, does this follow or not follow "Prow rules"?

Yes, it's a test run file that contains `cucshift` and `ginkgo` tests. The Prow rule I mean before is about the `step wring`, you can refer to my PR: https://github.com/openshift/release/pull/32626

xingxingxia: Thanks for clarifying `for the step, the simple the better` for "Prow rules". Got.

sumansuhag:
This method changes the OpenShift QE testing setup into a platform that can handle authentication testing at scale. It also reproduces the variety of identities found in production, which betters the accuracy of continuous integration and cuts down on authentication problems after deployment.

sumansuhag:
Automated QE Pipeline with IDP Integration and Dynamic User Federation:

Dynamic IDP Setup: Automatically create and remove many IDP instances for each test. This allows running several tests at once using different authentication systems like OpenID Connect, LDAP, and SAML.

Federated User Management: Keep user information in sync across different IDPs to mimic real-world environments that use many IDPs. This makes it easier to test authentication processes that span multiple providers.

Intelligent Test Management: Improve the QE framework to choose and authenticate test users from specific IDPs based on the test situation. This ensures thorough testing of complicated workflows involving multiple tenants and IDPs.

Security and Compliance Checks: Add automated audit logs to tests to confirm that IDP settings meet security standards. This helps catch mistakes or possible weaknesses early.

Self-Repairing IDP Provisioning: Use health checks and automatic fixes for IDP test services to make the pipeline more stable and reliable.

Performance Data and Metrics: Gather detailed data on performance and authentication to improve IDP integration and find problem areas in simulated real-world settings.

PR #32584, opened 2022-09-23
Enable SPO e2e-flaky-test

mrogers950:
@jhrozek PTAL, I enabled `E2E_TEST_SELINUX` given the comment in the other e2e definition but I wasn't sure if it's necessary, given that the jobs are separated.

### DIFF
--- a/ci-operator/config/openshift/security-profiles-operator/openshift-security-profiles-operator-main.yaml
+++ b/ci-operator/config/openshift/security-profiles-operator/openshift-security-profiles-operator-main.yaml
@@ -59,6 +59,33 @@ tests:
           cpu: 100m
       timeout: 4h0m0s
     workflow: ipi-aws
+- as: e2e-flaky
+  steps:
+    cluster_profile: aws
+    test:
+    - as: e2e-flaky
+      cli: latest
+      commands: |
+        export E2E_CLUSTER_TYPE=openshift
+        export E2E_SPO_IMAGE=$IMAGE_FROM_CI
+        export E2E_SKIP_BUILD_IMAGES=true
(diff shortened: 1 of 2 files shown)


jhrozek:
> @jhrozek PTAL, I enabled `E2E_TEST_SELINUX` given the comment in the other e2e definition but I wasn't sure if it's necessary, given that the jobs are separated.

I think that comment is just wrong, I'm /almost/ sure the tests do run, let's see when the e2e on this PR finishes.

ci-operator/config/openshift/security-profiles-operator/openshift-security-profiles-operator-main.yaml
```diff
@@ -59,6 +59,33 @@ tests:
...
+        export E2E_CLUSTER_TYPE=openshift
+        export E2E_SPO_IMAGE=$IMAGE_FROM_CI
+        export E2E_SKIP_BUILD_IMAGES=true
+        export E2E_SKIP_NAMESPACED_TESTS=true
+        export E2E_TEST_SECCOMP=true
+        export E2E_TEST_PROFILE_RECORDING=false
+        export E2E_TEST_BPF_RECORDER=false
+        export E2E_TEST_LOG_ENRICHER=true
+        export E2E_TEST_SELINUX=true
+        export E2E_TEST_LABEL_POD_DENIALS=true
```
jhrozek, line 78: This can and probably should be a separate commit (to remove the same variable from the other e2e stanza) but since the first e2e patch was merged, upstream completely removed the `label_pod_denials` feature and with it went its tests and this env variable, too.

jhrozek (approved):
/lgtm
with a comment, let's see how the tests fare

PR #32590, opened 2022-09-23
Make sure slack message is complete

derekhiggins:
It was only sending one word of the message to slack


derekhiggins:
@tdomnesc mind taking a look, see #forum-metal-alerts on slack for see an example of incorrect messages

PR #32622, opened 2022-09-26
Utilizing manifest listed payload

deepsm007:
Adding aws-ovn-multi job utilizing heterogeneous payload to deploy homogeneous amd64 cluster

/cc @stbenjam  @dgoodwin

### DIFF
--- a/ci-operator/config/openshift/release/openshift-release-master__nightly-4.12.yaml
+++ b/ci-operator/config/openshift/release/openshift-release-master__nightly-4.12.yaml
@@ -95,6 +95,19 @@ releases:
       product: ocp
       stream: nightly
       version: "4.12"
+  multi-initial:
+    candidate:
+      architecture: multi
+      product: ocp
+      relative: 1
+      stream: nightly
+      version: "4.12"
+  multi-latest:
+    candidate:
+      architecture: multi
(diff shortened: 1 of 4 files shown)


ci-operator/config/openshift/release/openshift-release-master__nightly-4.12.yaml
```diff
@@ -456,6 +469,15 @@ tests:
     env:
       BASE_DOMAIN: aws-2.ci.openshift.org
     workflow: openshift-e2e-aws-sdn
+- as: e2e-aws-ovn-multi
```
dgoodwin, line 472: Sippy's expecting multiRegex = regexp.MustCompile(`(?i)-heterogeneous`)

Can we rename so it picks up the right variant there?

Or is this a multi-arch payload, but regular amd64 job?

deepsm007: It is still an amd64 job with multiarch payload

deepsm007: multiRegex are used for true heterogeneous deployments only (cp=amd64, nodes=amd64+ arm64)

PR #30994, opened 2022-08-02
wip: openshift/microshift: single workflow for all openshift/conformance e2e tests

ingvagabund:
Run openshift/conformance tests in a single job

**TODO**
- [ ] - e2e junit and e2e.log are uploaded to the GCS bucket (e.g. https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-kube-scheduler-operator/433/pull-ci-openshift-cluster-kube-scheduler-operator-master-e2e-aws/1554100065386106880/artifacts/e2e-aws/openshift-e2e-test/artifacts/)
- [ ] run `must-gather` step
- [ ] TBD

/hold

PR #32659, opened 2022-09-27
MicroShift: refactor 4.12 test suites

copejon:


pacevedom:
I see there is some collision with this one: https://github.com/openshift/release/pull/30994
Which one should we keep?

pacevedom (approved):
/lgtm

bradmwilliams:
@copejon Just so you are aware...
You changed the names of jobs, that were also part of the release validation: 
https://github.com/openshift/release/pull/32763

PR #32668, opened 2022-09-28
OCPQE-11842: Add e2e for azure-upi-disconnected

liangxia:


liangxia:
@jhou1 @pruan-rht @dis016 @JianLi-RH

PR #32680, opened 2022-09-28
WIP NO-ISSUE: Enable KSM

adriengentil:
Since the change to use c3.medium machines, the ZTP jobs are randomly failing with the following error:
```
Domain already contains a disk with that address
```
this is due to a VM that is not started, this may be because of the lack of memory with c3.medium (64GB).

This change enables KSM which is a kernel process that tries optimize the memory by merging the common pages used by the VMs.


adriengentil:
Retesting as devscripts-setup step took ~20min more than usual (77min vs ~60min usually)

/test pj-rehearse-more

adriengentil:
Closing this PR, KSM looks to work but the jobs are now slower by 10-20min.

PR #32702, opened 2022-09-28
Fix openshift-sdn repo to run the openshift-sdn jobs

danwinship:
The openshift-sdn repo is currently running the ovn-kubernetes jobs, which are "not useful" for testing openshift-sdn changes...

/assign @jluhrsen

### DIFF
--- a/ci-operator/config/openshift/sdn/openshift-sdn-master.yaml
+++ b/ci-operator/config/openshift/sdn/openshift-sdn-master.yaml
@@ -68,26 +68,26 @@ tests:
     make test-unit JUNITFILE=${ARTIFACT_DIR}/junit_report.xml --warn-undefined-variables
   container:
     from: test-bin
-- as: e2e-aws
+- as: e2e-aws-sdn
   steps:
     cluster_profile: aws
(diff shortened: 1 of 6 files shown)


danwinship:
Hm... I renamed all of the tests across all releases to include "sdn" in the name, but maybe that's not consistent with what was done elsewhere?

jluhrsen:
> Hm... I renamed all of the tests across all releases to include "sdn" in the name, but maybe that's not consistent with what was done elsewhere?

it should be fine. we've only been adding the CNI names to jobs in 4.12 and higher. to be consistent, you could just revert any changes you have in 4.11 and lower.

danwinship:
> revert any changes you have in 4.11 and lower.

They still need to be changed to use the workflows that have "sdn" in the name though, right?

jluhrsen:
> > revert any changes you have in 4.11 and lower.
> 
> They still need to be changed to use the workflows that have "sdn" in the name though, right?

no. they will run with the default CNI which is sdn in 4.11 and below, so you don't have to change.

danwinship:
@jluhrsen updated

PR #32646, opened 2022-09-27
Test e2e-gcp-ipi-sdn-migration-ovn-p2

weliang1:
For testing purpose, please DO NOT merge!!

PR #32711, opened 2022-09-29
Output cucumber commands executed for easier debug

liangxia:
Related to https://github.com/openshift/release/pull/32646#issuecomment-1261739602

/cc @zhaozhanqi @weliang1 @jhou1

### DIFF
--- a/ci-operator/step-registry/cucushift/upgrade/check/cucushift-upgrade-check-commands.sh
+++ b/ci-operator/step-registry/cucushift/upgrade/check/cucushift-upgrade-check-commands.sh
@@ -34,7 +34,9 @@ else
 fi
 
 cd verification-tests
+set -x
 cucumber --tags "${UPGRADE_CHECK_RUN_TAGS} and ${UPGRADE_SKIP_TAGS}" -p junit || true
+set +x
 
 echo "Summarizing test result..."
 failures=$(grep '<testsuite failures="[1-9].*"' "${ARTIFACT_DIR}" -r | wc -l || true)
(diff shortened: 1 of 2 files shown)


ci-operator/step-registry/cucushift/upgrade/check/cucushift-upgrade-check-commands.sh
```diff
@@ -34,6 +34,7 @@ else
 fi
 
 cd verification-tests
+echo "tags: ${UPGRADE_PRE_RUN_TAGS} and ${UPGRADE_SKIP_TAGS}"
```
zhaozhanqi, line 37: s/${UPGRADE_PRE_RUN_TAGS}/${UPGRADE_CHECK_RUN_TAGS}/

ci-operator/step-registry/cucushift/upgrade/check/cucushift-upgrade-check-commands.sh
```diff
@@ -34,6 +34,7 @@ else
 fi
 
 cd verification-tests
+echo "tags: ${UPGRADE_CHECK_RUN_TAGS} and ${UPGRADE_SKIP_TAGS}"
```
jhou1, line 37: Can we print the raw cucumber command here if we want them for debugging? e.g.

```
set -x
cucumber xxx
set +x
```

ci-operator/step-registry/cucushift/upgrade/prepare/cucushift-upgrade-prepare-commands.sh
```diff
@@ -29,6 +29,7 @@ export OPENSHIFT_ENV_OCP4_USER_MANAGER_USERS=${USERS}
 export BUSHSLICER_REPORT_DIR="${ARTIFACT_DIR}/upgrade-prepare"
 
 cd verification-tests
+echo "tags: ${UPGRADE_PRE_RUN_TAGS} and ${UPGRADE_SKIP_TAGS}"
```
jhou1, line 32: Same here

ci-operator/step-registry/cucushift/upgrade/check/cucushift-upgrade-check-commands.sh
jhou1: Or print the cucumber string command

PR #32716, opened 2022-09-29
chore: build-service: temporarily disable e2e

psturc:
This change should make e2e CI jobs for build-service optional, i.e. not being triggered automatically when a new PR is created

### DIFF
--- a/ci-operator/config/redhat-appstudio/build-service/redhat-appstudio-build-service-main.yaml
+++ b/ci-operator/config/redhat-appstudio/build-service/redhat-appstudio-build-service-main.yaml
@@ -28,7 +28,7 @@ tests:
     product: ocp
     timeout: 1h0m0s
     version: "4.10"
-  skip_if_only_changed: ^docs/|^\.github/|^\.tekton/|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|PROJECT|LICENSE)$
+  optional: true
   steps:
     dependencies:
       COMPONENT_IMAGE: redhat-appstudio-build-service-image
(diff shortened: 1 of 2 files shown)


gabemontero:
> This change should make e2e CI jobs for build-service optional, i.e. not being triggered automatically when a new PR is created

with what you have here it will still be triggered @psturc 
it just will not be required to pass for merge

if you desire for it to only be launched if some says `/test ...` you need to add something like https://github.com/openshift/release/blob/master/ci-operator/config/redhat-appstudio/jvm-build-service/redhat-appstudio-jvm-build-service-main.yaml#L38

and yes, I'm very certain on how this works :-) 

now, what you want to do here, I'll defer to you and @mmorhun on what exact behavior you all want

psturc:
> > This change should make e2e CI jobs for build-service optional, i.e. not being triggered automatically when a new PR is created
> 
> with what you have here it will still be triggered @psturc it just will not be required to pass for merge
> 
> if you desire for it to only be launched if some says `/test ...` you need to add something like https://github.com/openshift/release/blob/master/ci-operator/config/redhat-appstudio/jvm-build-service/redhat-appstudio-jvm-build-service-main.yaml#L38
> 
> and yes, I'm very certain on how this works :-)

I know you have lot more experience with openshift-ci than me, but heck I'm still not convinced about the need of `run_if_changed` field in the job definition :P

I did this change based on the [job I was maintaining in my previous team](https://github.com/openshift/release/blob/5b786ff19b610e3f36566c274f7871ddc178efb5/ci-operator/jobs/integr8ly/integreatly-operator/integr8ly-integreatly-operator-master-presubmits.yaml#L428) 

See the `always_run: false` field - isn't it sufficient to prevent this job to run on every PR? 

See the docs: https://docs.ci.openshift.org/docs/how-tos/contributing-openshift-release/#tolerated-changes-to-generated-jobs

Of course I'm happy to add it if it's really needed, but according to docs and the job I referenced above it seems to me that it's not. Just need to clarify that. Thanks

gabemontero:
maybe I'm wrong ... or maybe the openshift repos and jvm-build-service had another setting that muddled things 

let's try it as is and see what happens .... you can always follow up with another PR if need be

/lgtm

PR #32726, opened 2022-09-29
Update coredns metal-ipi ci jobs

cybertron:
These seem to be out of date compared to some of the other repos where we run these jobs. This will hopefully get them working again.


gcs278:
Seems pretty straightforward.
/lgtm

PR #32727, opened 2022-09-29
cluster-bot: add signing-secret `env` variable

hoxhaeris:
The cluster-bot is being changed to use the slack Event-API over HTTP. 
This PR adds an additional env. variable (the app signing secret).


bradmwilliams:
/hold
Waiting for secret to propagate onto the cluster

bradmwilliams (approved):

PR #32744, opened 2022-09-30
Adding workflow-config in chat-bot for various providers

miyadav:
For debugging prow failures adding one more workflow-config .
@sunzhaohua2 @jhou1 @huali9 @liangxia @yunjiang29 , PTAL if not repeated and it is good to add to help debug with exact cluster config .

### DIFF
--- a/core-services/ci-chat-bot/workflows-config.yaml
+++ b/core-services/ci-chat-bot/workflows-config.yaml
@@ -48,6 +48,10 @@ workflows:
     platform: aws
   cucushift-installer-rehearse-aws-ipi-byo-route53:
     platform: aws
+  cucushift-installer-rehearse-aws-ipi-disconnected-private:
+    platform: aws
+  cucushift-installer-rehearse-aws-ipi-disconnected-private-sno:
+    platform: aws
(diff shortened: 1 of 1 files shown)


liangxia:
The change itself looks good.

Since the change also need approval from dptp. And to avoid bothering dptp too much, we should add this kind of changes in batch (rather than one by one).

miyadav:
/hold 
Will make bulk changes for merging later.

miyadav:
> The change itself looks good.
> 
> Since the change also need approval from dptp. And to avoid bothering dptp too much, we should add this kind of changes in batch (rather than one by one).

@liangxia PTAL once you are back and let me know if we can add more or this is ok to get merged for now .
The added worflow list contains profiles whose jenkins jobs are removed as per our [CI-Unifying-doc](https://docs.google.com/spreadsheets/d/1okJGAWsIj6RV91ChgsWQNwXzmHjvd-MkPntHr-UNNI8/edit?usp=sharing)
cc @sunzhaohua2 @jhou1

bradmwilliams (approved):

PR #32752, opened 2022-09-30
add customer cidr workflow

zhaozhanqi:


ci-operator/step-registry/cucushift/installer/custom-cidr/cucushift-installer-custom-cidr-ref.metadata.json
```diff
@@ -0,0 +1,17 @@
+{
+	"path": "cucushift/installer/custom-cidr/cucushift-installer-custom-cidr-ref.yaml",
```
jianlinliu, line 2: I think this step can be shared by others, not only for QE, so suggest to move the step folder to `ci-operator/step-registry/ovn/conf/custom-cidr`.

ci-operator/step-registry/cucushift/installer/custom-cidr/cucushift-installer-custom-cidr-ref.yaml
```diff
@@ -0,0 +1,17 @@
+ref:
+  as: cucushift-installer-custom-cidr
+  from_image:
```
jianlinliu, line 3: I think this step use the same image as `ovn-conf` ref.

ci-operator/step-registry/cucushift/installer/custom-cidr/cucushift-installer-custom-cidr-ref.metadata.json
zhaozhanqi: ok, good. I will move this folder.

ci-operator/step-registry/cucushift/installer/custom-cidr/cucushift-installer-custom-cidr-ref.yaml
zhaozhanqi: updated, thanks

zhaozhanqi:
@abhat Hi, Could you help take a look this PR, which is creating one custom CIDR subnet. for this bug https://bugzilla.redhat.com/show_bug.cgi?id=1980357 , base on this workflow. we will add one test case to check the alert

PR #32754, opened 2022-09-30
SPO: Enable more tests

jhrozek:
We only enabled a subset of tests. Let's enable more to increase SPO's
test coverage.

btw it is possible the tests might fail and I'll have to debug them a
bit. Sorry in advance if this PR requires a bunch of rehearsals.


jhrozek:
/remove-lifecycle rotten
I want to get back to this, just haven't found the time. The issue is tracked in our Jira, though.

PR #32755, opened 2022-09-30
Fix image used to build and test api-provider-baremetal

elfosardo:
Use native ART image with correct GO version instead of
an outdated one.


honza:
I wish the commit message was a bit more specific: fix how? The diff isn't super clear to me either: how does removing those lines make it work again?

elfosardo:
@honza I've updated the commit message

PR #32757, opened 2022-09-30
opendatahub-io/notebooks: Upgrade build_root image

samuelvl:
https://github.com/opendatahub-io/notebooks/pull/10 is failing because the current curl version in the `build_root` image doesn't support `--retry-connrefused` flag.

Upgrade the build_root image with an image that includes a most recent curl version:

```shell
$ podman run --rm --entrypoint="" registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.17-openshift-4.10 curl --help | grep '\-\-retry'               
     --retry <num>   Retry request if transient problems occur
     --retry-connrefused Retry on connection refused (use with --retry)
     --retry-delay <seconds> Wait time between retries
     --retry-max-time <seconds> Retry only within this period
```


atheo89 (approved):
/lgtm

PR #32760, opened 2022-09-30
Update patch managers for cluster-api-provider-baremetal

elfosardo:


elfosardo:
/lgtm
adding it myself as the bot didn't take the comment from Honza, probably a race condition?

PR #32780, opened 2022-09-30
app.ci: Add CentOS Stream CoreOS 9 ImageStream

LorbusChris:
This image will be part of future OKD/SCOS releases, for which we'll have to create a new release controller config.
This also creates the `origin/scos-4.12` IS, which will be used to build OKD/SCOS release payloads

/cc @aleskandro

### DIFF
--- a/clusters/app.ci/supplemental-ci-images/okd/OWNERS
+++ b/clusters/app.ci/supplemental-ci-images/okd/OWNERS
@@ -1,0 +1,2 @@
+approvers:
+- okd-approvers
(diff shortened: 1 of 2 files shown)


aleskandro (commented):
Hi @LorbusChris I'm just wondering whether we can check in the equivalent (not yet available) arm64 one. It should land on the origin-arm64 namespace with amd64 and x86_64 raplaced to arm64 and aarch64, respectively.

bradmwilliams:
/lgtm
You will need something from DPTP to `/approve` this however.  A request can be made in #forum-testplatform

LorbusChris:
I'll have to split this PR into two, because the second imagestream added here depends on existence of the first.

PR #32788, opened 2022-10-03
Upgrade qe jobs for arm64 on aws clusters

aleskandro:
This PRs enables the cucushift- steps for upgrade testing to work with ARM64 and adds two baselines upgrade tests for AWS IPI

/cc @JianLi-RH @liangxia @jianlinliu

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__arm64-nightly-4.12-upgrade-from-stable-4.11.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__arm64-nightly-4.12-upgrade-from-stable-4.11.yaml
@@ -1,0 +1,113 @@
+base_images:
+  ansible:
+    name: "4.11"
+    namespace: ocp
+    tag: ansible
+  cli:
+    name: "4.11"
+    namespace: ocp
+    tag: cli
+  dev-scripts:
+    name: test
+    namespace: ocp-kni
+    tag: dev-scripts
+  openstack-installer:
+    name: "4.11"
+    namespace: ocp
+    tag: openstack-installer
+  tests-private-postupg:
+    name: tests-private
+    namespace: ci
+    tag: "4.12"
+  tests-private-preupg:
+    name: tests-private
+    namespace: ci
+    tag: "4.11"
+  tools:
+    name: "4.11"
+    namespace: ocp
+    tag: tools
+  upi-installer:
+    name: "4.11"
+    namespace: ocp
+    tag: upi-installer
+build_root:
+  image_stream_tag:
+    name: release
+    namespace: openshift
+    tag: golang-1.18
+releases:
+  arm64-latest:
+    prerelease:
+      architecture: arm64
+      product: ocp
+      version_bounds:
+        lower: 4.11.0-0
+        upper: 4.12.0-0
+  arm64-target:
+    candidate:
+      architecture: arm64
+      product: ocp
+      stream: nightly
+      version: "4.12"
+  latest:
+    prerelease:
+      product: ocp
+      version_bounds:
+        lower: 4.11.0-0
+        upper: 4.12.0-0
+  target:
+    candidate:
+      architecture: amd64
+      product: ocp
+      stream: nightly
+      version: "4.12"
+resources:
+  '*':
+    requests:
+      cpu: 100m
+      memory: 200Mi
+tests:
+- as: aws-ipi-p1
+  cron: 0 6 * * *
+  steps:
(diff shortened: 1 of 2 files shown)


aleskandro:
Refers ARMOCP-466

aleskandro:
Refers ARMOCP-454

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__arm64-nightly-4.12-upgrade-from-stable-4.11.yaml
```diff
@@ -0,0 +1,113 @@
...
+  cron: 0 6 * * *
+  steps:
+    cluster_profile: aws-qe
+    dependencies:
+      OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE: release:arm64-latest
+      OPENSHIFT_UPGRADE_RELEASE_IMAGE_OVERRIDE: release:arm64-target
+    env:
+      BASE_DOMAIN: qe.devcluster.openshift.com
+      OCP_ARCH: arm64
+      TEST_FILTERS_PREUPG: ~40667&
```
aleskandro, line 81: This test is currently disabled due to the investigation in OCPQE-11950

aleskandro:
/cc @JianLi-RH @liangxia @jianlinliu
/hold cancel
the ugprade jobs seem ok with the latest 4.12 builds and with the 40667 test case disabled (see [OCPQE-11950](https://issues.redhat.com/browse/OCPQE-11950))

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-ref.yaml
```diff
@@ -18,8 +18,9 @@
   - name: PLATFORM_VERSION
     default: "8.6"
     documentation: RHEL worker OS version
+  - name: OCP_ARCH
+    default: amd64
+    documentation: The architecture of the OCP cluster
   dependencies:
-  - name: "release:latest"
```
shellyyang1989, line 22: Thanks for clearing the useless variable.

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -147,7 +147,21 @@ function rhel_upgrade(){
...
 # Extract oc binary which is supposed to be identical with target release
 function extract_oc(){
-    local url; url="openshift-release-artifacts.apps.ci.l2s4.p1.openshiftapps.com"
+    local route="openshift-release-artifacts"
+    local url
+    local target_arch=""
+    if [ "${OCP_ARCH}" != "amd64" ]; then
+      # Change the route to use for non-amd64 clusters
+      route="${route}-${OCP_ARCH}"
+      if [ "$(uname -m)" = "x86_64" ]; then
```
shellyyang1989, line 156: I'm trying to understand the code block. We have both [openshift-client-linux-4.12.0-0.nightly-arm64-2022-10-20-100222.tar.gz](https://openshift-release-artifacts-arm64.apps.ci.l2s4.p1.openshiftapps.com/4.12.0-0.nightly-arm64-2022-10-20-100222/openshift-client-linux-4.12.0-0.nightly-arm64-2022-10-20-100222.tar.gz) and [openshift-client-linux-amd64-4.12.0-0.nightly-arm64-2022-10-20-100222.tar.gz](https://openshift-release-artifacts-arm64.apps.ci.l2s4.p1.openshiftapps.com/4.12.0-0.nightly-arm64-2022-10-20-100222/openshift-client-linux-amd64-4.12.0-0.nightly-arm64-2022-10-20-100222.tar.gz). What's the difference between them? Why we need amd64 client in an arm cluster?

liangxia: Some binaries are not compatible. E.g, cp, mv
So actually we are running the installation in a amd64 pod/cluster to set up the ARM cluster.

aleskandro: Currently we run oc in amd64 pods against arm64 clusters. However, we are also working at having another prow cluster in the build farm, arm01, to enable testing from arm pods. This code is future proof to allow running some of the jobs in the arm01 cluster when it will be ready.

shellyyang1989: ack. thank you both!

PR #32791, opened 2022-10-03
Change upgrade jobs config to use arch prefix (amd64-only)

aleskandro:
The aim of this PR is just to rename the upgrade qe jobs to follow the same naming as for the nightly jobs, by including the arch as a `variant`. This only fixes the already present amd64 jobs naming.

Refers [ARMOCP-466](https://issues.redhat.com//browse/ARMOCP-466)


aleskandro:
Refers ARMOCP-466

aleskandro:
All the failures here are not related to the "renaming" proposed in this pr.

PR #32795, opened 2022-10-03
Disable always_run: true

kpouget:
/cc @fcami 

I didn't see it regenerated the files ... and lost `always_run: false` :/


fcami (approved):
/lgtm

PR #32799, opened 2022-10-03
feat: update reference-addon to use 'mage' and add integration tests

Ajpantuso:
### Summary

- Migrates reference-addon jobs to 'mage'
- Adds 'integration' tests to pre-submits
- Builds reference-addon-manager using improved Dockerfile
- Pre-caches test dependencies to reduce pr_check duration.


sandhyadalavi (commented):
lgtm

PR #32803, opened 2022-10-03
Hypershift: publish test-bin image

csrwng:
Promotes the test binary image from HyperShift so that it can be used by other repositories to run the HyperShift e2e.


bparees:
/approve

name lgtm

PR #32808, opened 2022-10-03
MGMT-11926: assisted-baremetal-conf: set kubeconfig when fetching a list of conformance tests

vrutkovs:
This kubeconfig would let us skip some tests depending on
cluster config: network plugin, test provider etc.

Hopefully this should prevent some tests from running on vsphere assisted job, which should be filtered out as they are known to fail on OpenshiftSDN

### DIFF
--- a/ci-operator/step-registry/assisted/common/verify/assisted-common-verify-ref.yaml
+++ b/ci-operator/step-registry/assisted/common/verify/assisted-common-verify-ref.yaml
@@ -4,6 +4,7 @@ ref:
   cli: latest
   grace_period: 10m
   commands: assisted-common-verify-commands.sh
+  timeout: 8h
   resources:
     requests:
       cpu: "3"


vrutkovs:
/hold

This seems to skip all tests now :/

vrutkovs:
/hold cancel

This now runs correct tests, but
* new tests are failing, probably flakes
* new test step timeout would be applied only when merged

vrutkovs:
Another timeout
/retest

ci-operator/step-registry/assisted/baremetal/conf/assisted-baremetal-conf-commands.sh
```diff
@@ -33,7 +33,11 @@ EOF
         ;;
 
     conformance)
-        INCL=$(openshift-tests run "openshift/conformance/parallel" --dry-run)
+        # shellcheck disable=SC2044
+        for kubeconfig in $(find ${SHARED_DIR} -name "kubeconfig*" -exec echo {} \;); do
```
omertuc, line 37: Why the loop? `INCL` always gets overwritten so only the last iteration matters

vrutkovs: Since we download it via API it gets a random name, so we have to use a loop to find it afaik

omertuc: so the loop should break when it finds one, to make that more clear. Also a comment could be nice

vrutkovs: Done

vrutkovs:
/hold cancel
Lets run this a few times in periodics

PR #32822, opened 2022-10-04
set PAS labels for reboot test

fzdarsky:
Adds labels to the default namespace for the nginx test workload to temporarily work around a bug in OCP that prevents the test workload from coming up otherwise.


mangelajo (approved):
/lgtm

PR #32838, opened 2022-10-04
[kn-plugin-event] Fix binary build with mage

dsimansk:
Reset default GOFLAGS in builder image.

/cc @cardil

### DIFF
--- a/ci-operator/config/openshift-knative/kn-plugin-event/openshift-knative-kn-plugin-event-release-next.yaml
+++ b/ci-operator/config/openshift-knative/kn-plugin-event/openshift-knative-kn-plugin-event-release-next.yaml
@@ -43,7 +43,7 @@ releases:
   latest:
     release:
       channel: stable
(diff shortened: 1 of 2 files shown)


cardil:
Why is this needed?

cardil:
Hmm, I think the better would be to update `openshift/e2e-test.sh` in the project instead. That would help avoid possible confusion in the future.

WDYT @dsimansk?

dsimansk:
> Hmm, I think the better would be to update `openshift/e2e-test.sh` in the project instead. That would help avoid possible confusion in the future.
> 
> WDYT @dsimansk?

I don't think that would address the flag `GOFLAGS` values when `./mage` is called in build phase. Or do you suggest to wrap `./mage` script into an openshift CI specific script?

cardil:
Right. I'll integrate that change in https://github.com/openshift/release/pull/34573

PR #32848, opened 2022-10-04
ROX-10613: Use ubi-minimal for scanner-db

janisz:
Refs: https://github.com/stackrox/scanner/pull/956
Tests: https://github.com/openshift/release/pull/32858

### DIFF
--- a/ci-operator/config/stackrox/scanner/stackrox-scanner-master.yaml
+++ b/ci-operator/config/stackrox/scanner/stackrox-scanner-master.yaml
@@ -1,8 +1,4 @@
 base_images:
-  ubi:
-    name: ubi
-    namespace: ocp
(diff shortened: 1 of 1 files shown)


RTann:
I don't recall the merge order. I think we want to merge the StackRox change first and then this? Going to hold off on lgtm because I think the PR may be automatically merged once I give it

janisz:
> Going to hold off on lgtm because I think the PR may be automatically merged once I give it
Correct lets put it on hold then.

I think the merge order does not matter that much. If we merge this PR first then we can rerun scanner test and they should pass. If we merge scanner first then we can run `pj-rehearse` and it should pass. I don't know how to merge them both in a safe way so two PRs will be green.

PR #32858, opened 2022-10-05
WIP: ROX-10613: Use ubi-minimal for scanner-db

janisz:
This is a test run PR for https://github.com/stackrox/scanner/pull/956 as it requires changes in config


janisz:
\cc @RTann

PR #32864, opened 2022-10-05
OKD/SCOS release-controller configuration

aleskandro:
This PR is to:

- mirror from the latest OCP ci release to the SCOS scos-4.x image stream
- ignore the OCP-specific images
- create the new stable and nightly (scos) releases, binding them to the `origin` release-controller
- Adds the prow configs for the okd-scos variant and enable building of the okd-specific payload images for okd-scos

**Note**: The rehearse jobs are expected to fail still

Blocked by openshift/release#32780 (merged)

Related to [OKD-10](https://issues.redhat.com//browse/OKD-10)

/cc @LorbusChris @vrutkovs @lmzuccarelli @sherine-k @bradmwilliams

### DIFF
--- a/ci-operator/config/openshift/cluster-samples-operator/openshift-cluster-samples-operator-master__okd-scos.yaml
+++ b/ci-operator/config/openshift/cluster-samples-operator/openshift-cluster-samples-operator-master__okd-scos.yaml
@@ -1,0 +1,90 @@
+base_images:
+  base:
+    name: scos-4.12
+    namespace: origin
+    tag: base
+  openshift_release_golang-1.17:
+    name: release
+    namespace: openshift
+    tag: golang-1.17
+build_root:
+  from_repository: true
+images:
+- dockerfile_path: Dockerfile.okd
+  from: base
+  inputs:
+    openshift_release_golang-1.17:
+      as:
+      - registry.ci.openshift.org/openshift/release:golang-1.17
+  to: cluster-samples-operator
+promotion:
+  name: scos-4.12
+  namespace: origin
+releases:
+  initial:
+    integration:
+      name: scos-4.12
+      namespace: origin
+  latest:
+    integration:
+      include_built_images: true
+      name: scos-4.12
+      namespace: origin
+resources:
+  '*':
+    requests:
+      cpu: 100m
+      memory: 200Mi
+tests:
+- as: verify
+  commands: make verify
+  container:
+    from: src
+- as: unit
+  commands: make test-unit
+  container:
+    from: src
+  optional: true
+- as: e2e-aws-ovn
+  optional: true
+  steps:
+    cluster_profile: aws
+    workflow: openshift-e2e-aws
+- as: e2e-aws-operator
+  optional: true
+  steps:
+    cluster_profile: aws
+    test:
+    - as: test
+      cli: latest
+      commands: make test-e2e-okd
+      from: src
+      resources:
+        requests:
+          cpu: 100m
+    workflow: ipi-aws
+- as: e2e-aws-builds
+  optional: true
+  steps:
+    cluster_profile: aws
+    workflow: openshift-e2e-aws-builds
+- as: e2e-aws-jenkins
+  optional: true
+  steps:
+    cluster_profile: aws
+    workflow: openshift-e2e-aws-jenkins
+- as: e2e-aws-image-ecosystem
+  optional: true
+  steps:
+    cluster_profile: aws
+    workflow: openshift-e2e-aws-image-ecosystem
+- as: e2e-aws-upgrade
+  optional: true
+  steps:
+    cluster_profile: aws
+    workflow: openshift-upgrade-aws
+zz_generated_metadata:
+  branch: master
+  org: openshift
+  repo: cluster-samples-operator
+  variant: okd-scos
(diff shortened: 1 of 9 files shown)


core-services/release-controller/_releases/release-okd-scos-4.y-stable.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name":"4-scos-stable",
+  "as": "Stable",
+  "message": "All amd64 OKD/SCOS versions are tracked here",
+  "overrideCLIImage": "image-registry.openshift-image-registry.svc:5000/ocp/4.6:cli",
```
LorbusChris, line 5: I know this is used in all the configs, but it seems wrong to me

LorbusChris (commented):
@aleskandro thank you for working on this!

One thing that probably could be done as part of this PR is to add MCO and Installer builds with build arg TAGS=scos, and to promote them to the origin/scos-4.12 IS

core-services/release-controller/_releases/release-okd-scos-4.y-stable.json
bradmwilliams: This is a release-controller specific setting and should remain.  It simply means that the release-controller will use this version of the `cli` image to run it's various `oc adm release ...` commands.

bradmwilliams:
In order for the release-controller changes to be applied to the system, you'll also need to make the appropriate changes here as well: https://github.com/openshift/release/blob/master/hack/annotate.sh

core-services/release-controller/_releases/release-okd-scos-4.y-stable.json
LorbusChris: I wonder though why we are using an EOL'd release for that? Even 4.6 EUS support ends this month. I'm not saying this has to be changed here now, but it'd be good to have the reasoning behind this documented somewhere.

bradmwilliams: If it aint broke, dont fix it! ;)
In all honesty, the `oc adm release` functionality, we rely on, hasn't changed.  I'll test out bumping up to something more modern and update accordingly.

aleskandro: some changes happened for non-single-arch builds that may eventually impact the jobs using an old cli

LorbusChris (approved):
/lgtm

core-services/release-controller/_releases/release-okd-4.12.json
```diff
@@ -11,6 +11,18 @@
...
+      "imageStreamRef": {
+        "namespace": "origin",
+        "name": "scos-4.12",
+        "excludeTags": [
+          "machine-config-operator",
+          "installer",
+          "installer-artifacts",
+          "fedora-coreos"
+        ]
+      }
```
LorbusChris, line 25: @aleskandro @bradmwilliams thinking about this again - this would effectively make OKD/SCOS a downstream of OKD/FCOS, which might have implications when OKD/FCOS fails to publish for any reason.

Couldn't we instead add this next to `mirror-to-origin` in [release-ocp-4.12-ci.json](https://github.com/openshift/release/blob/afdb7a9db76233fc270aae5ad8a351a73896ee6a/core-services/release-controller/_releases/release-ocp-4.12-ci.json#L10)? (with an adapted list of `excludeTags` for mirroring out of `ocp`).
That'd put OKD/SCOS on the same downstream level from OCP CI as OKD/FCOS.

bradmwilliams: @LorbusChris That is how I interperated how this was implemented:
CI (OCP) -> OKD (FCOS) -> OKD (SCOS)

It sounds like what you really wanted was OCP (CI) feeding **both**, which would be done by adding the `mirror-to-okd-scos` as a peer to:  `mirror-to-origin` in `release-ocp-4.12-ci.json`

bradmwilliams: Where (what imagestream) are you planning on injecting your OKD(SCOS) CI builds?

bradmwilliams: The reason I ask is that any change to the OKD(SCOS) imagestream will cause a new release to be generated, so your promotions plus the OCP(CI) mirroring could potentially cause quite a bit of churning of releases

aleskandro: I did the mirroring from okd/fcos to okd/scos to avoid doing new prow configs for all the other payload components...  the difference between scos and fcos is a subset of the diff between fcos/scos and ocp. But yes, this means that the mirroring of the images would only happen after an FCOS build is [accepted](https://github.com/openshift/release-controller/blob/master/cmd/release-controller/sync.go#L556) and it may be better not to depend on the fcos builds acceptance. Should I change?

LorbusChris: The churn should be the same as what we already have in OKD/FCOS.
We'd directly inject/promote the following images to the SCOS imagestream:
- cluster-update-keys
- cluster-samples-operator
- installer
- installer-artifacts
- machine-config-operator
- operator-marketplace
- ironic-* images

PR #32876, opened 2022-10-05
ci: Delete Prow periodic

cgwalters:
Closes: https://github.com/openshift/os/issues/1017

It turns out that having a Prow-generated `rhel-coreos-8` distinct from the ART version is causing massive confusion because it creates a divergence between CI and nightlies.

Delete this Prow periodic so we can have just exactly one `rhel-coreos-8` built by ART.


cgwalters:
Now with a `make jobs` run

PR #32877, opened 2022-10-05
LOG-2969: Add non-blocking e2e jobs to clusterlogging

jcantrill:
ref https://issues.redhat.com/browse/LOG-2969

@periklis does this match what we discussed?


cahartma (commented):
/lgtm

PR #32878, opened 2022-10-05
jobs/4.12: s/machine-os-content/rhel-coreos-8/

cgwalters:
xref  https://github.com/openshift/os/issues/1017

This is part of ensuring we only have one `rhel-coreos-8` image that is copied from ART into CI.

### DIFF
--- a/ci-operator/jobs/openshift/release/openshift-release-release-4.12-periodics.yaml
+++ b/ci-operator/jobs/openshift/release/openshift-release-release-4.12-periodics.yaml
@@ -44,37 +44,42 @@ periodics:
 
         # if the source and destination are identical, do nothing
         export KUBECONFIG=/etc/app.ci/kubeconfig
-        from=$( oc get istag -n "${TAG_NS}" "${TAG_SOURCE}" -o template='{{ .image.metadata.name }}' )
+        from=$( oc get istag -n "${TAG_NS}" "${TAG}:${BRANCH}" -o template='{{ .image.metadata.name }}' )
         to=$( oc get istag --ignore-not-found -n ocp "${BRANCH}:${TAG}" -o template='{{ .image.metadata.name }}' )
+        from_extensions=$( oc get istag -n "${TAG_NS}" "${TAG}-extensions:${BRANCH}" -o template='{{ .image.metadata.name }}' )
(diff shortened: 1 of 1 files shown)


cgwalters:
OK yeah this is a mess...I only tried changing `promote-release-openshift-machine-os-content-e2e-aws-ovn-4.12` for now, but this needs to be copy-pasted to the other jobs

cgwalters:
Hmm `make jobs` seems to want to turn the shell script into an illegible mess...perhaps we should try to factor this out into a runnable container image and not inline shell.

ci-operator/jobs/openshift/release/openshift-release-release-4.12-periodics.yaml
```diff
@@ -44,37 +44,42 @@ periodics:
...
+        echo "Will promote ${TAG}-extensions ${from_extensions}, current is ${to_extensions}"
 
         # error out if the image isn't on quay
         to_quay="quay.io/openshift-release-dev/ocp-v4.0-art-dev@${from}"
-        if ! oc image info -a "/usr/local/pull-secret/.dockerconfigjson" "${to_quay}"; then
-          echo "error: The source image has not been pushed to quay ${from}"
-          exit 1
-        fi
+        to_quay_extensions="quay.io/openshift-release-dev/ocp-v4.0-art-dev@${from_extensions}"
+        for img in ${to_quay} ${to_quay_extensions}; do
```
bradmwilliams, line 62: @cgwalters There is an extra space at the end of this line that is causing the issue with `make jobs`.  Removing it and re-running the command should resolve the problem.

bradmwilliams (changes requested):

ci-operator/jobs/openshift/release/openshift-release-release-4.12-periodics.yaml
cgwalters: Oh wow.  The chance I would have figured that out on my own is basically 0%.  I assume you got burned by this in the past or something?

cgwalters:
OK so I believe the status here is that since https://github.com/openshift/aos-cd-jobs/pull/3253 merged, at this time we don't need these promotion jobs for 4.12+.  So I will instead open a new PR to delete them.  cc @jupierce right?

PR #32905, opened 2022-10-06
Remove hypershift-tests mirror mapping

csrwng:
Remove hypershift-tests mirror mapping


bparees:
this file is generated, so if you don't want a mirror rule, i think you have to not promote the image.

PR #32915, opened 2022-10-06
OADP-1.0: Add 4.11  e2e jobs

hhpatel14:


kaovilai:
Will have to update https://github.com/openshift/oadp-operator/pull/830 to exclude CSI for 4.11

PR #32918, opened 2022-10-07
feat: add periodic job for synthetic test data update

eggfoobar:
/hold for PR openshift/ci-tools#3052 to avoid un-needed large commit message
/cc @dgoodwin 

Auto update synthetic tests for openshift/origin, currently self approve is commented off to validate behavior.

[TRT-568](https://issues.redhat.com//browse/TRT-568)

Signed-off-by: ehila <ehila@redhat.com>

### DIFF
--- a/ci-operator/jobs/infra-periodics.yaml
+++ b/ci-operator/jobs/infra-periodics.yaml
@@ -2542,3 +2542,85 @@ periodics:
     - name: token
       secret:
         secretName: github-credentials-openshift-bot
+- agent: kubernetes
+  cluster: app.ci
+  cron: 0 0 * * */5
+  decorate: true
+  extra_refs:
+  - base_ref: master
+    org: openshift
+    repo: origin
+  labels:
+    ci.openshift.io/role: infra
+  name: periodic-update-origin-disruption-alert-data
+  spec:
+    containers:
+    - args:
+      - -c
+      - |
+        set -o errexit
+        set -o nounset
+        set -o pipefail
+
+        leeway_percent=30
+
+        echo "Updating alerts data"
+        /usr/bin/job-run-aggregator analyze-historical-data \
+        --current ./origin/pkg/synthetictests/allowedalerts/query_results.json \
+        --output-file ./origin/pkg/synthetictests/allowedalerts/query_results.json \
+        --data-type alerts \
+        --leeway $leeway_percent \
+        --google-service-account-credential-file /etc/bigquery/token
+
+        echo "Updating disruptions data"
+        /usr/bin/job-run-aggregator analyze-historical-data \
+        --current ./origin/pkg/synthetictests/allowedbackenddisruption/query_results.json \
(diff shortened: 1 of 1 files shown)


ci-operator/jobs/infra-periodics.yaml
```diff
@@ -2542,3 +2542,85 @@ periodics:
     - name: token
       secret:
         secretName: github-credentials-openshift-bot
+- agent: kubernetes
+  cluster: app.ci
+  cron: 0 * */14 * *
```
dgoodwin, line 2547: This looks like an update every hour on the 14th/28th of each month, or am I reading this wrong? Something like 0 0 */14 * * should do the trick. Actually how about every week for now so we can see how this flows.

ci-operator/jobs/infra-periodics.yaml
```diff
@@ -2542,3 +2542,85 @@ periodics:
...
+  decorate: true
+  extra_refs:
+  - base_ref: master
+    org: openshift
+    repo: origin
+  labels:
+    ci.openshift.io/role: infra
+  name: periodic-prow-auto-aggregator-synthetic-data
```
dgoodwin, line 2555:
```suggestion
  name: periodic-update-origin-disruption-alert-data
```
Seem ok?

ci-operator/jobs/infra-periodics.yaml
```diff
@@ -2542,3 +2542,85 @@ periodics:
...
+        -github-token-path=/etc/github/oauth \
+        -organization=openshift \
+        -repo=origin \
+        -git-message="Updated synthetic tests"
+        -pr-message "$(cat /var/tmp/pr_message.md)" \
+        -pr-title="Automated - Update synthetic test data" \
+        $extra_flag
+      command:
+      - /bin/bash
+      image: registry.ci.openshift.org/ci/auto-aggregator-job-names:latest
```
dgoodwin, line 2607: I assume this is piggybacking an image you setup for the job name updater and will always be latest code.

dgoodwin (commented):
As a heads up Egli, we should see the Ps drop soon now that we started recording 0's :(

And then after a few weeks I'm going to change this to filter out everything that doesn't have 100 results, so the file is going to shrink dramatically.

ci-operator/jobs/infra-periodics.yaml
eggfoobar: Yeah that's right, maybe it would be best to rename that image?

ci-operator/jobs/infra-periodics.yaml
eggfoobar: Sounds good!

ci-operator/jobs/infra-periodics.yaml
dgoodwin: Probably not worth the effort, but in a perfect world yes.

PR #32921, opened 2022-10-07
Postsubmit for ACM stolostron/image-builder on Arm build cluster

deepsm007:
It enables postsubmits for https://github.com/stolostron/image-builder on `arm01` build cluster
The required work to make all libraries ARM compatible and remove hard coded dependency for x86 is tracked in this ticket -->  https://issues.redhat.com/browse/ACM-1757 

/cc @gurnben


gurnben (approved):
/approve
/lgtm

PR #32924, opened 2022-10-07
openstack: Remove NFV gates

pierreprinetti:
The test is permafailing independently of the production code, and will need time for fixing.


mandre (commented):
It's only failing in master, isn't it?

ci-operator/jobs/openshift/machine-api-provider-openstack/openshift-machine-api-provider-openstack-master-presubmits.yaml
```diff
@@ -74,7 +74,7 @@ presubmits:
           secretName: result-aggregator
     trigger: (?m)^/test( | .* )e2e-openstack,?($|\s.*)
   - agent: kubernetes
-    always_run: true
+    always_run: false
```
mandre, line 77: we probably want to leave it running by default, so that we know when it's fixed. If we make it only available on demand, we're also very likely to forget to restore the job once it's fixed. We want it to run by default in MAPO since it has the potential to break the NFV job.

pierreprinetti: ok

mandre (commented):
/lgtm

PR #32939, opened 2022-10-07
OSDe2e PR Check patch

MrSantamaria:
Changing the PR Check to match CLI execution. 
The eventual goal is to move all our jobs to run from the binary instead of using scripts.

### DIFF
--- a/ci-operator/config/openshift/osde2e/openshift-osde2e-main.yaml
+++ b/ci-operator/config/openshift/osde2e/openshift-osde2e-main.yaml
@@ -35,7 +35,7 @@ tests:
   commands: |
     export GOFLAGS=""
     export REPORT_DIR="$ARTIFACT_DIR"
-    export CONFIGS="aws,prod,e2e-suite"
+    export CONFIGS="aws,prod,pr-check"
     export GINKGO_SKIP="Must Gather Operator"
     export SECRET_LOCATIONS="/usr/local/osde2e-common,/usr/local/osde2e-credentials,/usr/local/sd-cicd-aws-prod"
 


MrSantamaria:
Oh man that took an embarrassing amount of commits. 

The export GOFLAGS=""
Disabled the hard check of the go vendors from being used in prow. 
More on: https://go.dev/ref/mod#vendoring

To be merged afterwards:
https://github.com/openshift/osde2e/pull/1322

ci-operator/config/openshift/osde2e/openshift-osde2e-main.yaml
```diff
@@ -18,21 +22,34 @@ resources:
...
 
-    ci/prow_pr_check.sh
+    make check
+  container:
+    from: src
+- as: verify
+  commands: |
+    export GOFLAGS=""
+    export REPORT_DIR="$ARTIFACT_DIR"
+    export CONFIGS="aws,prod,e2e-suite"
```
dustman9000, line 38: Is there a reason we need to run the full e2e-suite? Is there a config with a smaller set of core tests we can use? This seems a little excessive for a PR check IMO.

MrSantamaria: We don't have anything smaller but we could make one:
The e2e-suite today:
tests:
  testsToRun:
  - '\[Suite\: e2e\]'
  - '\[Suite\: operators\]'
  - '\[Suite\: service-definition\]'
  - '\[Suite\: app-builds\]'

MrSantamaria: Created a pr-check config in the osde2e repo that only includes the service-definition tests.

PR #32944, opened 2022-10-08
OCPQE-11842: Add e2e for azure-ipi-workers-marketimage

liangxia:


jianlinliu:
The install job failed because of a rpm-ostree issue, we can reproduce it on 4.12.0-0.nightly-2022-10-05-053337, and verify it on 4.12.0-0.ci-2022-10-14-053746, get passed. So the PR looks good to me.

PR #32946, opened 2022-10-08
[vsphere] add workflow for multi zone cluster

jinyunma:
[SPLAT-773](https://issues.redhat.com/browse/SPLAT-773) add workflow for multi zone cluster

### DIFF
--- a/ci-operator/config/openshift/verification-tests/openshift-verification-tests-master__installer-rehearse-4.12.yaml
+++ b/ci-operator/config/openshift/verification-tests/openshift-verification-tests-master__installer-rehearse-4.12.yaml
@@ -81,6 +81,11 @@ tests:
   steps:
     cluster_profile: vsphere-discon
     workflow: cucushift-installer-rehearse-vsphere-upi-disconnected-sdn-workers-rhcos-rhel8
+- as: installer-rehearse-vsphere-multizone
+  interval: 960h
+  steps:
+    cluster_profile: vsphere-multizone
+    workflow: cucushift-installer-rehearse-vsphere-ipi-zones-techpreview
 zz_generated_metadata:
   branch: master
   org: openshift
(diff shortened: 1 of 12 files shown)


ci-operator/step-registry/cucushift/installer/rehearse/vsphere/ipi/zones/provision/cucushift-installer-rehearse-vsphere-ipi-zones-provision-chain.yaml
```diff
@@ -0,0 +1,8 @@
+chain:
+  as: cucushift-installer-rehearse-vsphere-ipi-zones-provision
+  steps:
+  - chain: ipi-conf-vsphere-zones
```
jianlinliu, line 4: From `ipi-conf-vsphere-zones.ref`, I saw it enables `FeatureGate`, so vsphere multip-zones is a TP feature?

jinyunma: yes, it is TP in 4.12.

jinyunma: If we don't enable featureGate in install-config.yaml with zone configuration, will get following error.
```
time="2022-10-08T04:00:19Z" level=error msg="failed to fetch Metadata: failed to load asset \"Install Config\": failed to create install config: invalid \"install-config.yaml\" file: [platform.vsphere.failureDomains: Forbidden: the TechPreviewNoUpgrade feature set must be enabled to use this field, platform.vsphere.vcenters: Forbidden: the TechPreviewNoUpgrade feature set must be enabled to use this field]"
```

jianlinliu: Thanks for the info. If that, we can follow other techpreview workflows rules, place the steps in `cucushift/installer/rehearse/vsphere/ipi/zones/techpreview`, and document the workflow get supported from 4.12.

PR #32949, opened 2022-10-09
[AWS] Custom mirror-registry pull secret was overridden in ipi-conf-aws step

yunjiang29:
The issue was introduced by https://github.com/openshift/release/pull/32851

[OCPQE-12380](https://issues.redhat.com/browse/OCPQE-12380) [AWS] Custom mirror-registry pull secret was overridden in ipi-conf-aws step

### DIFF
--- a/ci-operator/step-registry/ipi/install/install/aws/ipi-install-install-aws-commands.sh
+++ b/ci-operator/step-registry/ipi/install/install/aws/ipi-install-install-aws-commands.sh
@@ -49,7 +49,7 @@ function wait_router_lb_provision() {
       echo "waiting on router-default service load balancer ingress..."
       sleep 30
       SERVICE="$(oc -n openshift-ingress get service router-default -o json)"
-      try="((try + 1))"
+      let try+=1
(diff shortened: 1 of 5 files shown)


ci-operator/step-registry/ipi/install/install/aws/ipi-install-install-aws-commands.sh
```diff
@@ -49,7 +49,7 @@ function wait_router_lb_provision() {
       echo "waiting on router-default service load balancer ingress..."
       sleep 30
       SERVICE="$(oc -n openshift-ingress get service router-default -o json)"
-      try="((try + 1))"
+      try=$(expr $try + 1)
```
liangxia, line 52: Perhaps use `let try+=1` which is more simple and readable.

yunjiang29:
cc @jianlinliu

yunjiang29:
all installs get passed, the job failure is because some of e2e tests failed, it's not related to this PR.

/lgtm

PR #27788, opened 2022-04-13
Add unit-test cluster-proxy

xuezhaojun:
Signed-off-by: xuezhaojun <zxue@redhat.com>

The env var `XDG_CACHE_HOME` is added to avoid issue like: 
```
panic: mkdir /.cache: permission denied
```
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/stolostron_hypershift-deployment-controller/76/pull-ci-stolostron-hypershift-deployment-controller-main-unit-tests/1514094625344720896

PR #32960, opened 2022-10-09
Fix: add cache home for cluster-proxy.

xuezhaojun:
Signed-off-by: xuezhaojun <zxue@redhat.com>


xuezhaojun:
This is fixed before: https://github.com/openshift/release/pull/27788
But at that time, there is no integration-test, so this part is missed.

PR #32969, opened 2022-10-10
Default worker type for aws on arm64

aleskandro:
https://github.com/openshift/release/pull/31684 introduced a logic that aims to set the compute node types in a more uniform way with cost-saving optimization as the target.

This PR injects in the newly introduced code, a default COMPUTE_NODE_TYPE for arm64 machines so that jobs configuration don't need anymore to set it (unless needed).

For the QE jobs, this also means changing from the m6g.large to the m6g.xlarge instance types for arm64 so that the testing is performed against the same amount of resources with respect to the amd64 counterpart. The PR also fixes some flakiness in cases results and issues due to the scarce resources to execute the test suites.

/cc @yunjiang29 @jiajliu


aleskandro:
/cc @lwan-wanglin

for ARM QE peer review. can you ack this PR?

Note: the cco-manual-security-token installation is failing, but it is beyond the scope of this PR. The rest looks good.

PR #32997, opened 2022-10-10
ci-operator/step-registry: Trim wking from various OWNERS

wking:
I was on the installer team back when I was added to a number of these directories.  But I've now been on the updates team for long enough that I'm no longer current with install/deprovision mechanics.  This commit removes me from a number of locations where I don't consider myself a subject-matter expert, hopefully making it easier for folks to actually find a knowledgeable approver for their changes.

I'm leaving myself as an approver in a few places (e.g. proxy steps) where I don't consider myself a subject-matter expert, to avoid completely emptying the approver step.  I'll come back in follow-up work to try to address those, either by finding additional approvers, or by removing those steps altogether.

### DIFF
--- a/ci-operator/step-registry/aws/deprovision/cco-manual-users/OWNERS
+++ b/ci-operator/step-registry/aws/deprovision/cco-manual-users/OWNERS
@@ -1,4 +1,3 @@
 approvers:
 - patrickdillon
-- wking
 - yunjiang29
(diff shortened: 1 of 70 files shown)


wking:
Proximal trigger is me trying to back out of ipi-install-install maintenance/approval for #31965.  I don't want to trigger another reroll there, so holding this metadata shift until that one lands:

/hold

wking:
#31965 landed.

/hold cancel

PR #33025, opened 2022-10-11
scheduler-plugins: prepare for release 4.12

swatisehgal:
Signed-off-by: Swati Sehgal <swsehgal@redhat.com>

### DIFF
--- a/ci-operator/jobs/openshift-kni/scheduler-plugins/openshift-kni-scheduler-plugins-release-4.12-postsubmits.yaml
+++ b/ci-operator/jobs/openshift-kni/scheduler-plugins/openshift-kni-scheduler-plugins-release-4.12-postsubmits.yaml
@@ -4,7 +4,7 @@ postsubmits:
     always_run: true
     branches:
     - ^release-4\.12$
-    cluster: build05
+    cluster: build01
     decorate: true
     decoration_config:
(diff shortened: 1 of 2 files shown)


fromanirh (commented):
looks good, but we also need to update https://github.com/openshift/release/blob/master/ci-operator/config/openshift-kni/scheduler-plugins/openshift-kni-scheduler-plugins-master.yaml

ci-operator/config/openshift-kni/scheduler-plugins/openshift-kni-scheduler-plugins-master.yaml
```diff
@@ -22,12 +22,12 @@ promotion:
 releases:
   initial:
     integration:
-      name: "4.11"
+      name: "4.12"
```
fromanirh, line 25: if memory serves, back in time we used 4.11 here instead of the more correct 4.12 (so we used version N instead of the more correct version N+1) because version 4.12 (version N+1) was not available.
Let's merge this PR as-is, but let's post a followup to use 4.13 here. in the near future.

swatisehgal: Ack.

PR #33030, opened 2022-10-11
AGENT-371: add agent job for testing dualstack

bfournie:
Add a new agent job to test dualstack (IPV4/IPV6) in an HA cluster.

### DIFF
--- a/ci-operator/config/openshift/installer/openshift-installer-master.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-master.yaml
@@ -509,7 +509,7 @@ tests:
   steps:
     cluster_profile: nutanix
     workflow: openshift-e2e-nutanix
-- as: e2e-agent-compact
+- as: e2e-agent-compact-ipv4
   cluster: build05
   skip_if_only_changed: (^(docs|upi|hack)/)|((^|/)OWNERS(_ALIASES)?$)|((^|/)[A-Z]+\.md$)|(^(data/.*|images|pkg/.*|terraform/providers)/(alibabacloud|aws|azure|azurestack|baremetal|gcp|ibmcloud|libvirt|nutanix|openstack|ovirt|powervs|vsphere)/)
   steps:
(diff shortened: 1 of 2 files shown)


ci-operator/config/openshift/installer/openshift-installer-master.yaml
```diff
@@ -525,6 +525,24 @@ tests:
         AGENT_E2E_TEST_SCENARIO=COMPACT_IPV4
         AGENT_DEPLOY_MCE=true
     workflow: agent-e2e-generic
+- as: e2e-agent-ipv6
```
andfasano, line 528:
```suggestion
- as: e2e-agent-compact-ipv6
```

ci-operator/config/openshift/installer/openshift-installer-master.yaml
```diff
@@ -525,6 +525,24 @@ tests:
...
+  optional: true
+  skip_if_only_changed: (^(docs|upi|hack)/)|((^|/)OWNERS(_ALIASES)?$)|((^|/)[A-Z]+\.md$)|(^(data/.*|images|pkg/.*|terraform/providers)/(alibabacloud|aws|azure|azurestack|baremetal|gcp|ibmcloud|libvirt|nutanix|openstack|ovirt|powervs|vsphere)/)
+  steps:
+    cluster_profile: equinix-ocp-metal
+    env:
+      DEVSCRIPTS_CONFIG: |
+        AGENT_E2E_TEST_SCENARIO=COMPACT_IPV6
+    workflow: agent-e2e-generic
+- as: e2e-agent-dualstack
```
andfasano, line 537:
```suggestion
- as: e2e-agent-compact-dualstack
```

bfournie: Changed

zaneb (commented):
/hold
We cannot afford to add 3 jobs testing compact clusters that run on every installer patch. We need to come up with a strategy where we have max 3 jobs that give us the range of coverage we want, including across all 3 topologies. 2 of those jobs need to be triggered by something like `run_if_changed: ^(data|pkg)/(.+)?/agent/` instead of `skip_if_changed`.

ci-operator/config/openshift/installer/openshift-installer-master.yaml
```diff
@@ -509,7 +509,7 @@ tests:
   steps:
     cluster_profile: nutanix
     workflow: openshift-e2e-nutanix
-- as: e2e-agent-compact
+- as: e2e-agent-compact-ipv4
```
andfasano, line 512: Based on our discussed testing matrix, shouldn't the mce deployment be moved into a periodic one?

bfournie: Yes but I think we should address that in a separate PR as this just adds the dualstack job (the "ipv4" was added to the agent-compact job based on the earlier comment)

bfournie:
> /hold We cannot afford to add 3 jobs testing compact clusters that run on every installer patch. We need to come up with a strategy where we have max 3 jobs that give us the range of coverage we want, including across all 3 topologies. 2 of those jobs need to be triggered by something like `run_if_changed: ^(data|pkg)/(.+)?/agent/` instead of `skip_if_only_changed`.

Noting that we met to discuss the agent jobs and the results are in https://docs.google.com/document/d/1NC1j2zBz19wwFe_bBokHcpFSK99EE9CqHz0a1ny68Xg/edit#heading=h.43stdz49bjp3. It was agreed that this would be 1 of the 3 presubmit jobs, the job was changed to set DHCP and Disconnnected options to increase coverage.

PR #33033, opened 2022-10-11
Skip ci jobs for non-code changes

elfosardo:


derekhiggins:
/lgtm
Also sets some 4.8 jobs to optional, inline with other releases

PR #33055, opened 2022-10-12
OCPQE-11842: Add e2e for nutanix-ipi

liangxia:


ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-nightly.yaml
```diff
@@ -989,6 +989,17 @@ tests:
     test:
     - chain: openshift-e2e-test-qe
     workflow: cucushift-installer-rehearse-gcp-upi-xpn
+- as: e2e-nutanix-ipi-p1
+  cluster: build01
```
sgaoshang, line 993: We can only run nutanix ci on build05 cluster

PR #33062, opened 2022-10-12
Updating build farm cluster-groups

bradmwilliams:


DennisPeriquet:
Just a note that these two clusters are added back because they are [known to be functional](https://coreos.slack.com/archives/C01CQA76KMX/p1665584565884459?thread_ts=1665583749.534479&cid=C01CQA76KMX); see [build05](https://prow.ci.openshift.org/?cluster=build05) and [build01](https://prow.ci.openshift.org/?cluster=build01)

PR #33075, opened 2022-10-12
Move cluster-bot gcp launches away from build02 to build04

DennisPeriquet:
build02 has been removed from rotation for prow jobs.  This removes it from cluster-bot gcp jobs too.

I chose build04 because it seems to be the other gcp cluster according to [this doc](https://docs.ci.openshift.org/docs/getting-started/useful-links/#clusters).

### DIFF
--- a/ci-operator/jobs/openshift/release/openshift-release-infra-periodics.yaml
+++ b/ci-operator/jobs/openshift/release/openshift-release-infra-periodics.yaml
@@ -910,7 +910,7 @@ periodics:
       secret:
         secretName: ci-pull-credentials
 - agent: kubernetes
-  cluster: build03
+  cluster: build04
   cron: '@yearly'
   decorate: true
   labels:


bradmwilliams (approved):

DennisPeriquet:
CI and my PR don't agree.  I make it build04 because build04 is on gcp.  But after `make jobs`, it becomes build03 -- and so CI will fail on my use of build04.

build03 is on aws according to [this doc](https://docs.ci.openshift.org/docs/getting-started/useful-links/#clusters).

DennisPeriquet:
we may not need this as I saw cluster-bot came up and started an installation; I deleted it before it was finished because because I figure installation actually starting was enough to prove it was working.

DennisPeriquet:
build02 has been put back into rotation therefore, this PR is no longer relevant.

PR #33076, opened 2022-10-12
Bump framework cluster ACM version

dhaiducek:
Bumping the version rather than removing it and letting it default to the latest makes sense so that we seamlessly get through the churn of moving to the next release.


JustinKuli (approved):

dhaiducek:
(We'll continue to hold this PR even after my Framework PR merges until a new snapshot is promoted to `2.7-dev` because otherwise my tests will block PRs to Framework.)

dhaiducek:
Still waiting for a new snapshot to `dev`. I've followed up in Slack: https://coreos.slack.com/archives/CSZLMKPS5/p1666359253924009

PR #33078, opened 2022-10-12
Azure Stack: update feature set annotation

patrickdillon:
Annotation was changed from feature-gate to feature-set. Add a check for feature-set to make sure the bad credentials are destroyed.


r4f4:
```
INFO[2022-10-12T21:44:18Z] Running step e2e-azurestack-ipi-conf-azurestack-creds. 
INFO[2022-10-12T21:45:18Z] Step e2e-azurestack-ipi-conf-azurestack-creds succeeded after 1m0s. 
```
\lgtm

PR #32818, opened 2022-10-03
Use cluster_claim for sippy e2e test

DennisPeriquet:
Related to [TRT-246](https://issues.redhat.com//browse/TRT-246)

Inspired by https://github.com/openshift/release/pull/32456/files (built off of branch dgoodwin:sippy-e2e).

Add presubmit job for sippy that does this on a cluster-pool cluster:

* starts a postgres DB
* loads the DB with test data
* starts the sippy api-server
* setups up an Openshift route to expose the sippy api-server endpoint
* runs the e2e test

PR #33079, opened 2022-10-12
Add Service Account for access to CI Registry for sippy CI

DennisPeriquet:
Related to [TRT-264](https://issues.redhat.com//browse/TRT-264)

Related to [PR32818](https://github.com/openshift/release/pull/32818) because we need to push the sippy container image built in CI to a container registry where the cluster-pool cluster can access it when we create sippy pods to run the sippy e2e test.

ref:

* [programmatic access to the central CI registry](https://docs.ci.openshift.org/docs/how-tos/use-registries-in-build-farm/#how-do-i-get-a-token-for-programmatic-access-to-the-central-ci-registry)
* [trt-admins Rover group](https://rover.redhat.com/groups/search?q=trt-admins)


DennisPeriquet:
this is actually not what I want.

PR #33080, opened 2022-10-12
[vsphere] update metrics collection for mdcnc

rvanderp3:
updating metric collection to:
- check a static set of metrics
- support multiple datacenters and clusters
- produce results in both text and json formats
- collect metrics for hosts


rvanderp3:
cc: @jcpowermac

rvanderp3:
IBM Cloud, VMC, and multi-zone are all collecting metrics again :tada:

PR #33087, opened 2022-10-13
Reduce vexxhost quota slice

mandre:
It appears we don't have enough storage quota on vexxhost to run 18 jobs in parallel. With a 1TB storage quota, and each cluster provisioning a 100GB volume for the image registry, this only leaves room for 10 clusters. And because the e2e tests also need to provision smaller volumes, let's set the limit to 9 until we get more quota. This also should help reduce the stress on vexxhost and see if we manage to get the latency between nodes (and VIP flapping) under control.

### DIFF
--- a/core-services/prow/02_config/_boskos.yaml
+++ b/core-services/prow/02_config/_boskos.yaml
@@ -1033,8 +1033,8 @@ resources:
   min-count: 7
   state: free
   type: openstack-quota-slice
-- max-count: 18
-  min-count: 18
+- max-count: 9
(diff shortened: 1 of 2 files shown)


pierreprinetti:
ouch.
/lgtm

mandre:
As the bot suggests,
/assign @jmguzik

droslean (approved):

EmilienM:
We are reverting it via https://github.com/openshift/release/pull/55579

PR #33098, opened 2022-10-13
Deprovision: Trap logs

barbacbd:
** Add a trap to save the logs even on failure.


ci-operator/step-registry/ipi/deprovision/deprovision/ipi-deprovision-deprovision-commands.sh
```diff
@@ -60,7 +61,7 @@ set -e
 echo "Copying the Installer logs to the artifacts directory..."
 cp /tmp/installer/.openshift_install.log "${ARTIFACT_DIR}"
```
patrickdillon, line 62: Can we include the `echo` within the trap and remove this `cp`? I believe we should only need the one within the trap.

PR #33100, opened 2022-10-13
Use repo as build root for cloud provider vSphere build image

lobziik:
For use image which specified in https://github.com/openshift/cloud-provider-vsphere/blob/main/.ci-operator.yaml instead of hardcoded golang-1.17


elmiko (commented):
makes sense to me
/lgtm

PR #33104, opened 2022-10-13
DNM: Try int tests for istio. IGNORE ME

jwendell:


yxun:
I was trying to run upstream integ tests in an OCP cluster as well. 
Before I am going to try in CI, I think we need update upstream tests.
I was testing it in https://github.com/yxun/sandbox/tree/main/istio/istio by manually create a pod and then run a integ test suite

PR #33149, opened 2022-10-17
Add Arm64 ipi on Azure profiles for QE

lwan-wanglin:
Add Arm64 ipi on Azure profiles for QE. refer [ARMOCP-221](https://issues.redhat.com/browse/ARMOCP-221) [ARMOCP-222](https://issues.redhat.com/browse/ARMOCP-222)

- Add avaliablity zones obtainment becasue not all the zones have arm64 type of machines.

### DIFF
--- a/ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh
+++ b/ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh
@@ -30,13 +30,24 @@ if [[ "${SIZE_VARIANT}" == "compact" ]]; then
   workers=0
 fi
 master_type=null
+master_type_prefix=""
 if [[ "${SIZE_VARIANT}" == "xlarge" ]]; then
-  master_type=Standard_D32s_v3
+  master_type_prefix=Standard_D32
 elif [[ "${SIZE_VARIANT}" == "large" ]]; then
-  master_type=Standard_D16s_v3
+  master_type_prefix=Standard_D16
 elif [[ "${SIZE_VARIANT}" == "compact" ]]; then
-  master_type=Standard_D8s_v3
+  master_type_prefix=Standard_D8
 fi
+if [ -n "${master_type_prefix}" ]; then
+  if [ "${OCP_ARCH}" = "amd64" ]; then
+    master_type=${master_type_prefix}s_v3
+  elif [ "${OCP_ARCH}" = "arm64" ]; then
+    master_type=${master_type_prefix}ps_v5
+  fi
+fi
+
+echo "Using control plane instance type: ${master_type}"
+echo "Using compute instance type: ${COMPUTE_NODE_TYPE}"
 
 cat >> "${CONFIG}" << EOF
 baseDomain: ${BASE_DOMAIN}
@@ -44,12 +55,14 @@ platform:
   azure:
     region: ${REGION}
 controlPlane:
+  architecture: ${OCP_ARCH}
   name: master
   platform:
     azure:
       type: ${master_type}
 compute:
-- name: worker
+- architecture: ${OCP_ARCH}
+  name: worker
   replicas: ${workers}
   platform:
     azure:
--- a/ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-ref.yaml
+++ b/ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-ref.yaml
@@ -30,6 +30,10 @@ ref:
     default: 'Standard_D4s_v3'
     documentation: |-
       The instance type to use for compute nodes
+  - name: OCP_ARCH
+    default: "amd64"
+    documentation: |-
+      The architecture of the control plane nodes (e.g., amd64, arm64).
   - name: BASE_DOMAIN
     default: ci.azure.devcluster.openshift.com
     documentation: |-
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__arm64-nightly.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__arm64-nightly.yaml
@@ -226,6 +226,36 @@ tests:
     test:
     - chain: openshift-e2e-test-qe
     workflow: cucushift-installer-rehearse-aws-ipi-proxy-cco-manual-security-token-service
+- as: e2e-azure-ipi-ovn-p1
+  cron: 0 15 */2 * *
+  steps:
+    cluster_profile: azure-arm64-qe
+    dependencies:
+      OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE: release:arm64-latest
+    env:
+      BASE_DOMAIN: qe.azure.devcluster.openshift.com
+      COMPUTE_NODE_TYPE: Standard_D4ps_v5
+      E2E_RUN_TAGS: '@azure-ipi and @arm64 and @network-ovnkubernetes and not @fips'
+      OCP_ARCH: arm64
+      TAG_VERSION: '@4.12'
+    test:
+    - chain: openshift-e2e-test-qe
+    workflow: cucushift-installer-rehearse-azure-ipi-ovn
+- as: e2e-azure-ipi-sdn-p3
+  cron: 0 16 */28 * *
+  steps:
+    cluster_profile: azure-arm64-qe
+    dependencies:
+      OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE: release:arm64-latest
+    env:
+      BASE_DOMAIN: qe.azure.devcluster.openshift.com
+      COMPUTE_NODE_TYPE: Standard_D4ps_v5
+      E2E_RUN_TAGS: '@azure-ipi and @arm64 and @network-openshiftsdn and not @fips'
+      OCP_ARCH: arm64
+      TAG_VERSION: '@4.12'
+    test:
+    - chain: openshift-e2e-test-qe
+    workflow: cucushift-installer-rehearse-azure-ipi-sdn
 zz_generated_metadata:
   branch: release-4.12
   org: openshift
--- a/ci-operator/jobs/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12-periodics.yaml
+++ b/ci-operator/jobs/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12-periodics.yaml
@@ -13079,6 +13079,170 @@ periodics:
         secretName: result-aggregator
 - agent: kubernetes
   cluster: build05
+  cron: 0 15 */2 * *
+  decorate: true
+  decoration_config:
+    skip_cloning: true
+  extra_refs:
+  - base_ref: release-4.12
+    org: openshift
+    repo: openshift-tests-private
+  labels:
+    ci-operator.openshift.io/cloud: azure-arm64
+    ci-operator.openshift.io/cloud-cluster-profile: azure-arm64-qe
+    ci-operator.openshift.io/variant: arm64-nightly
+    ci.openshift.io/generator: prowgen
+    job-release: "4.12"
+    pj-rehearse.openshift.io/can-be-rehearsed: "true"
+  name: periodic-ci-openshift-openshift-tests-private-release-4.12-arm64-nightly-e2e-azure-ipi-ovn-p1
+  spec:
+    containers:
+    - args:
+      - --gcs-upload-secret=/secrets/gcs/service-account.json
+      - --image-import-pull-secret=/etc/pull-secret/.dockerconfigjson
+      - --lease-server-credentials-file=/etc/boskos/credentials
+      - --oauth-token-path=/usr/local/github-credentials/oauth
+      - --report-credentials-file=/etc/report/credentials
+      - --secret-dir=/secrets/ci-pull-credentials
+      - --secret-dir=/usr/local/e2e-azure-ipi-ovn-p1-cluster-profile
+      - --target=e2e-azure-ipi-ovn-p1
+      - --variant=arm64-nightly
+      command:
+      - ci-operator
+      image: ci-operator:latest
+      imagePullPolicy: Always
+      name: ""
+      resources:
+        requests:
+          cpu: 10m
+      volumeMounts:
+      - mountPath: /etc/boskos
+        name: boskos
+        readOnly: true
+      - mountPath: /secrets/ci-pull-credentials
+        name: ci-pull-credentials
+        readOnly: true
+      - mountPath: /usr/local/e2e-azure-ipi-ovn-p1-cluster-profile
+        name: cluster-profile
+      - mountPath: /secrets/gcs
+        name: gcs-credentials
+        readOnly: true
+      - mountPath: /usr/local/github-credentials
+        name: github-credentials-openshift-ci-robot-private-git-cloner
+        readOnly: true
+      - mountPath: /etc/pull-secret
+        name: pull-secret
+        readOnly: true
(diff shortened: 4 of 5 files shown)


ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh
```diff
@@ -20,41 +20,131 @@ function getVersion() {
...
 master_type=null
-if [[ "${SIZE_VARIANT}" == "xlarge" ]]; then
-  master_type=Standard_D32s_v3
-elif [[ "${SIZE_VARIANT}" == "large" ]]; then
-  master_type=Standard_D16s_v3
-elif [[ "${SIZE_VARIANT}" == "compact" ]]; then
-  master_type=Standard_D8s_v3
+if [ "${architecture}" = "arm64" ]; then
+  BOOTSTRAP_NODE_TYPE="Standard_D8ps_v5"
+  if [[ "${SIZE_VARIANT}" == "xlarge" ]]; then
```
aleskandro, line 50: I think we can remove the double if.

We can:
- change the suffix (ps_v5 or s_v3) based on the arch.
- change the prefix (Standard_D(8|16|32)) based on the SIZE_VARIANT.

In this way, we don't repeat the if statement for the size variant two times.

Example:
```
if arch == arm64:
  BOOTSTRAP_NODE...
  master_type_suffix=ps_v5
else:
  BOOTSTRAP...
  master_type_suffix=s_v3

if SIZE_VARIANT...
  master_type_prefix=Standard_D32
.....

master_type=${master_type_prefix}${master_type_suffix}
```

ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh
```diff
@@ -20,41 +20,131 @@ function getVersion() {
...
+  mapfile -t INSTANCE_ZONES < <(az vm list-skus --location="${REGION}" | jq -r '.[] | select(.name == "'${BOOTSTRAP_NODE_TYPE}'" or .name == "'${master_type}'" or .name == "'${COMPUTE_NODE_TYPE}'") | .locationInfo[].zones[]' | sort | uniq -c | grep ' 2 ' | awk '{print $2}')
+elif [[ "${BOOTSTRAP_NODE_TYPE}" != "${COMPUTE_NODE_TYPE}" && "${COMPUTE_NODE_TYPE}" != "${master_type}" ]]; then   # three different regions
+  mapfile -t INSTANCE_ZONES < <(az vm list-skus --location="${REGION}" | jq -r '.[] | select(.name == "'${BOOTSTRAP_NODE_TYPE}'" or .name == "'${master_type}'" or .name == "'${COMPUTE_NODE_TYPE}'") | .locationInfo[].zones[]' | sort | uniq -c | grep ' 3 ' | awk '{print $2}')
+fi
+# Calculate the maximum number of availability zones from the region
+MAX_ZONES_COUNT="${#INSTANCE_ZONES[@]}"
+# Save max zones count information to ${SHARED_DIR} for use in other scenarios
+echo "${MAX_ZONES_COUNT}" >> "${SHARED_DIR}/maxzonescount"
+
+existing_zones_setting=$(yq-go r "${CONFIG}" 'controlPlane.platform.azure.zones')
```
aleskandro, line 106: We can do this check earlier and skip some instructions when not needed.
```
existing_zone_setting=....
if [[ ${existing_zones_setting} == "" ]];
... from row 77 to 104 ...
... from row 109 to 122 ...
else
  echo "zones already set in install-config.yaml, skipped"
fi

File: ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh [hunk 3]
```diff
@@ -20,41 +20,131 @@ function getVersion() {
...
+
+# log in with az
+if [[ "${CLUSTER_TYPE}" == "azuremag" ]]; then
+    az cloud set --name AzureUSGovernment
+else
+    az cloud set --name AzureCloud
+fi
+az login --service-principal -u "${AZURE_AUTH_CLIENT_ID}" -p "${AZURE_AUTH_CLIENT_SECRET}" --tenant "${AZURE_AUTH_TENANT_ID}" --output none
+
+# Generate availability zones with OpenShift Installer required instance types
```
aleskandro, line 85: Do you think is possible to minimize the branches in the following if conditionals?

ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh
lwan-wanglin: thanks for the suggests, will fix it

ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh
lwan-wanglin: for this part, I refer to [ipi-conf-aws-commands.sh](https://github.com/openshift/release/blob/master/ci-operator/step-registry/ipi/conf/aws/ipi-conf-aws-commands.sh#L116-L130) , I think it will consider all the situations.

ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh
```diff
@@ -20,41 +20,131 @@ function getVersion() {
...
+    master_type="Standard_D16s_v3"
+  elif [[ "${SIZE_VARIANT}" == "compact" ]]; then
+    master_type="Standard_D8s_v3"
+  fi
+fi
+
+echo "Using control plane instance type: ${master_type}"
+echo "Using compute instance type: ${COMPUTE_NODE_TYPE}"
+
+# set the parameters we'll need as env vars
```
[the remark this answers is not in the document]
lwan-wanglin: The purpose to add "az login" is that we need to run `az vm list-skus ...`  command， discussed with @MayXuQQ ，amd64 doesn't hit the zone limit issues for amd64, so we can keep the origin code without manually changing zone setting for amd64 and doesn't do existing_zone_setting check and keep it as it is.

We can only do it when arch is arm64 and supported zone are not 1,2,3(such as only avaliable in zone 1 or zone 1,2),  is it feasible for you?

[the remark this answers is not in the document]
aleskandro: Given that we are doing it even for aws, I think we should have the same behavior for Azure and regardless of the architecture.  And I think what you've done is ok, it can be just re-sorted as suggested (or better).

However, I'm still wondering whether the installer can implement this feature. Users aiming to install a cluster in a region that has not the wanted machines in the AZs, don't need to check themselves and set the zones in the install-config.yaml. The installer would perform this kind of work to understand which zones can be used. Maybe @jianlinliu could help answer.

aleskandro:
cc @deepsm007

ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-ref.yaml
```diff
@@ -27,9 +27,13 @@ ref:
       in some regions or accounts due to quota issues.
 
   - name: COMPUTE_NODE_TYPE
-    default: 'Standard_D4s_v3'
```
vrutkovs, line 30: Why not set `Standard_D4ps_v5` in azure-arm64 test?

aleskandro: Hi @vrutkovs, I'd say this is compatible with the previous work done on aws: for arm64 (#32969) and in general (#31684) to let the CI use default instance types for cost-saving purposes, unless a specific test needs to override the COMPUTE_NODE_TYPE.

ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh
```diff
@@ -20,36 +20,62 @@ function getVersion() {
   echo "${version}"
 }
 
+architecture=${OCP_ARCH:-"amd64"}
```
aleskandro, line 23: We can remove lines 23-27. OCP_ARCH will be enough

ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh
```diff
@@ -20,36 +20,62 @@ function getVersion() {
...
+  architecture="arm64"
+fi
+
 CONFIG="${SHARED_DIR}/install-config.yaml"
 
 REGION="${LEASED_RESOURCE}"
 echo "Azure region: ${REGION}"
 
+if [[ "${COMPUTE_NODE_TYPE}" == "" ]]; then
+  if [ "${architecture}" = "arm64" ]; then
```
aleskandro, line 34: Remove 33-39 as suggested by @vrutkovs and let's set the COMPUTE_NODE_TYPE var from the config

ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh
```diff
@@ -20,36 +20,62 @@ function getVersion() {
...
+  else
+    COMPUTE_NODE_TYPE=Standard_D4s_v3
+  fi
+fi
 workers=3
 if [[ "${SIZE_VARIANT}" == "compact" ]]; then
   workers=0
 fi
 master_type=null
+if [ "${architecture}" = "arm64" ]; then
```
aleskandro, line 45: Use OCP_ARCH

ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh
```diff
@@ -20,36 +20,62 @@ function getVersion() {
...
+  master_type_suffix=ps_v5
+else
+  master_type_suffix=s_v3
+fi
+master_type_prefix=""
 if [[ "${SIZE_VARIANT}" == "xlarge" ]]; then
```
aleskandro, line 51: Move 51-57 after line 44 and call it "master_type".

ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh
```diff
@@ -20,36 +20,62 @@ function getVersion() {
...
+    COMPUTE_NODE_TYPE=Standard_D4s_v3
+  fi
+fi
 workers=3
 if [[ "${SIZE_VARIANT}" == "compact" ]]; then
   workers=0
 fi
 master_type=null
+if [ "${architecture}" = "arm64" ]; then
+  master_type_suffix=ps_v5
```
aleskandro, line 46: `master_type=${master_type}ps_v5`

ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh
```diff
@@ -20,36 +20,62 @@ function getVersion() {
...
+else
+  master_type_suffix=s_v3
```
aleskandro, line 48: `master_type=${master_type}s_v3`

ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh
```diff
@@ -20,36 +20,62 @@ function getVersion() {
...
-  master_type=Standard_D32s_v3
+  master_type_prefix=Standard_D32
 elif [[ "${SIZE_VARIANT}" == "large" ]]; then
-  master_type=Standard_D16s_v3
+  master_type_prefix=Standard_D16
 elif [[ "${SIZE_VARIANT}" == "compact" ]]; then
-  master_type=Standard_D8s_v3
+  master_type_prefix=Standard_D8
 fi
+if [[ -n "${master_type_prefix}" ]]; then
```
aleskandro, line 58: remove the lines 58-60

ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-ref.yaml
lwan-wanglin: ipi-conf-azure is not just for arm64 test,  setting it to "" in ref.yaml and giving default value for both amd64 and arm64 in shell file is more easy.

ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh
lwan-wanglin: @aleskandro  ipi-conf-azure is not just for arm64 test, setting it to "" in ref.yaml config file and giving default value for both amd64 and arm64 in shell file is more easy.   
how can we set two default value for one variable from config file?

vrutkovs: We can default to amd64 (mostly as there's already a lot of amd64 tests) and in arm64 test use a custom COMPUTE_NODE_TYPE.

vrutkovs: >then we will need to set custom COMPUTE_NODE_TYPE to every workflow for arm on azure

Yes

> if we keep the current code, we can set or not set.

Yes, that's the tradeoff here - you'll run less rehearsals as the step code doesn't change anymore and move configuration to the test and not the step-registry. 

>just like what we did for aws now

This is exactly why we should _stop_ doing that - more and more logic would be encoded in the step, causing more rehearsals which are irrelevant and making the logic more and more complicated.

lwan-wanglin: Ok, totally got your point, will fix it, thanks so much for your reply.

ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh
```diff
@@ -31,25 +31,35 @@ if [[ "${SIZE_VARIANT}" == "compact" ]]; then
 fi
 master_type=null
 if [[ "${SIZE_VARIANT}" == "xlarge" ]]; then
-  master_type=Standard_D32s_v3
+  master_type=Standard_D32
```
jianlinliu, line 34: suggest to rename the var name to `master_type_family`

ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh
```diff
@@ -31,25 +31,37 @@ if [[ "${SIZE_VARIANT}" == "compact" ]]; then
...
-  master_type=Standard_D32s_v3
+  master_type_family=Standard_D32
 elif [[ "${SIZE_VARIANT}" == "large" ]]; then
-  master_type=Standard_D16s_v3
+  master_type_family=Standard_D16
 elif [[ "${SIZE_VARIANT}" == "compact" ]]; then
-  master_type=Standard_D8s_v3
+  master_type_family=Standard_D8
 fi
+if [ "${master_type}" != "null" ]; then
```
aleskandro, line 40: this is always null

lwan-wanglin:
@aleskandro @vrutkovs @jianlinliu, most jobs installation part are success, please help to review the pr again, thanks!

vrutkovs (approved):
/lgtm

vrutkovs:
/hold

for others to review

aleskandro:
/lgtm

Thanks @lwan-wanglin

lwan-wanglin:
/assign @kuiwang02
Could you help merge it? thanks

kuiwang02:
i am OK from golang's view.
@lwan-wanglin please request qe produtivity team to approve it.

lwan-wanglin:
/assign @jhou1 @liangxia 
Could you help approve it when your are avaiable? thanks

PR #33050, opened 2022-10-12
Add job to deploy with bonds

creydr:
Add a job to dev-scripts to run the common baremetal e2e test suite with a bond primary interface

PR #33161, opened 2022-10-17
Add bond config to virtualmedia workflow

creydr:
In #33050 we added a job for e2e tests on baremetalds for a bonded primary interface. This PR is the follow up on https://github.com/openshift/release/pull/33050#issuecomment-1280522278 to have this tested in a more regular basis. 

**Hints for reviewers:**
The adding of the bonds to the `baremetalds-e2e-sdn-serial-virtualmedia` workflow was done in ab3f35c417c0ff007ec6a9a18cfc54ce00b342db. The other commit (300783ea9e171e8f978916e3929c1d8dd6a07e19) was only to rename the workflow to have the `-bond` suffix to make it clear that this workflow is based on a bonded primary interface.

/cc @derekhiggins @cybertron

### DIFF
--- a/ci-operator/config/openshift/release/openshift-release-master__nightly-4.12.yaml
+++ b/ci-operator/config/openshift/release/openshift-release-master__nightly-4.12.yaml
@@ -191,12 +191,12 @@ tests:
   steps:
     cluster_profile: equinix-ocp-metal
     workflow: baremetalds-e2e-serial-ovn-ipv6
-- as: e2e-metal-ipi-sdn-serial-virtualmedia
+- as: e2e-metal-ipi-sdn-serial-virtualmedia-bond
   cluster: build05
   interval: 48h
   steps:
     cluster_profile: equinix-ocp-metal
-    workflow: baremetalds-e2e-sdn-serial-virtualmedia
+    workflow: baremetalds-e2e-sdn-serial-virtualmedia-bond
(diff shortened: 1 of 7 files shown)


creydr:
Hello @derekhiggins ,
could you PTAL at this PR? It is the followup to your comment in https://github.com/openshift/release/pull/33050#issuecomment-1280522278
Thanks a lot in advance

ci-operator/config/openshift/release/openshift-release-master__nightly-4.12.yaml
```diff
@@ -196,7 +196,7 @@ tests:
   interval: 48h
```
derekhiggins, line 196: > - as: e2e-metal-ipi-sdn-serial-virtualmedia
I wonder should we also rename this to "e2e-metal-ipi-sdn-serial-virtualmedia-bond"
so its more obvious to the casual observer ?

creydr: Yeah. Probably makes sense too. Let me update it...

creydr: Updated it to include the suffix

creydr:
@wking and @jupierce could you PTAL at this PR. I guess we need both of your approvals, as we rename one of the jobs.
Thanks a lot in advance

PR #33162, opened 2022-10-17
Fixing tests for Heterogeneous Serial jobs

deepsm007:
The heterogeneous serial jobs are failing for the 70 skipped tests which are only part of parallel suite.

/cc @aleskandro


aleskandro:
/lgtm

/hold 

until tests succeed

PR #33164, opened 2022-10-17
quay.io ICSP for aws build clusters

jmguzik:


jmguzik:
> If I recall, it will restart all the nodes. Need some babysitting while rolling.

this is not relevant anymore

PR #33165, opened 2022-10-17
fix: incorrect path

eggfoobar:
/cc @dgoodwin 
/assign @bbguimaraes

Signed-off-by: ehila <ehila@redhat.com>

### DIFF
--- a/ci-operator/jobs/infra-periodics.yaml
+++ b/ci-operator/jobs/infra-periodics.yaml
@@ -2548,7 +2548,7 @@ periodics:
         secretName: github-credentials-openshift-bot
 - agent: kubernetes
   cluster: app.ci
-  cron: 0 0 * * */5
+  cron: 0 1 * * */1
   decorate: true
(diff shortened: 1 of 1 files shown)


ci-operator/jobs/infra-periodics.yaml
```diff
@@ -2548,7 +2548,7 @@ periodics:
         secretName: github-credentials-openshift-bot
 - agent: kubernetes
   cluster: app.ci
-  cron: 0 0 * * */5
+  cron: 0 */1 * * *
```
dgoodwin, line 2551: This looks like every run divisible by 1, i.e. every hour or the same thing as *. Do we want just a 1 here? (run every day at 1am)

eggfoobar: Sure thing, wasn't sure if we wanted to run it every hour to verify and switch it off after, let me change it to 1

PR #33177, opened 2022-10-17
Mount right volume for aws secrets in hypershift-aws-create-chain

enxebre:
We refer to `/etc/hypershift-ci-jobs-awscreds` in the script but we are not mounting it.

I'm not sure where `hypershift-pool-aws-credentials` is coming from on CI, which is referenced in multiple jobs. I can see `hypershift-ci-jobs-awscreds` in vault.


enxebre:
PTAL @sjenning

PR #33187, opened 2022-10-18
kvm-device-plugin: Tolerate CI worker taint

cgwalters:
We want to schedule this pod onto CI workers.

xref https://github.com/openshift/os/issues/1028


travier (approved):

cgwalters:
Hmm actually, we may also need `ci-prowjobs-worker`...

@bear-redhat do you know where these node taints are defined?

cgwalters:
/hold
while I test some stuff

cgwalters:
/hold cancel
OK I added all the tolerations I think we need

travier:
Do we still need this one?

PR #33198, opened 2022-10-18
Health check on CO version after upgrade

shellyyang1989:
CO might not get upgraded and no error is reported. So we need to check each CO version is same with the target version.

### DIFF
--- a/ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
+++ b/ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
@@ -190,6 +190,12 @@ function check_clusteroperators() {
         (( tmp_ret += 1 ))
     fi
 
+    echo "Make sure every operator reports correct version"
+    if incorrect_version=$(${OC} get clusteroperator --no-headers | awk -v var="${TARGET_VERSION}" '$2 != var') && [[ ${incorrect_version} != "" ]]; then
+        echo >&2 "Incorrect CO Version: ${incorrect_version}"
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -190,6 +190,12 @@ function check_clusteroperators() {
         (( tmp_ret += 1 ))
     fi
 
+    echo "Make sure every operator reports correct version"
+    if incorrect_version=$(${OC} get clusteroperator | awk -v var="${TARGET_VERSION}" '$2 != var') && [[ ${incorrect_version} != "" ]]; then
```
jiajliu, line 194: --no-headers?

shellyyang1989: ah, right

shellyyang1989:
The upgrade step got passed in the [job](https://qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/qe-private-deck/pr-logs/pull/openshift_release/33198/rehearse-33198-periodic-ci-openshift-openshift-tests-private-release-4.12-amd64-nightly-4.12-upgrade-from-stable-4.11-aws-ipi-disconnected-private-p1/1582293021855059968)

INFO[2022-10-18T12:53:41Z] Step aws-ipi-disconnected-private-p1-cucushift-upgrade-toimage succeeded after 1h13m40s. 

It's ready for merge if no any more concerns.

jiajliu:
lgtm

PR #33210, opened 2022-10-18
Take a proper Prometheus snapshot

tjungblu:
Just tar'ing the prometheus folder does seem to cause some WAL corruption on GCP jobs


tjungblu:
`--web.enable-admin-api` not available sadly

/hold

PR #33214, opened 2022-10-18
Use 4.12 cluster for testing compliance-operator

rhmdnd:
Let's make sure the compliance-operator works against newer releases.


jhrozek (approved):
/lgtm
/approve
let's see how CI goes..

PR #33223, opened 2022-10-18
openshift-psap/ci-artifacts-priv: remove always_run: true

kpouget:
`main` is not supported, and `perf` isn't convenient

/cc @fcami


fcami (approved):
/lgtm

PR #33224, opened 2022-10-18
`pj-rehearse`: deployment for pj-rehearse plugin

smg247:
`comment-on-pr-creation` being set to false is what allows us to do a soft-rollout with the plugin. It will be completely hidden unless the user issues the proper commands.

Next step is to enable the plugin on the `openshift/release` repo.

### DIFF
--- a/clusters/app.ci/prow/03_deployment/pj-rehearse-plugin.yaml
+++ b/clusters/app.ci/prow/03_deployment/pj-rehearse-plugin.yaml
@@ -1,0 +1,114 @@
+kind: List
+apiVersion: v1
+items:
+- apiVersion: v1
+  kind: Service
+  metadata:
+    name: pj-rehearse-plugin
+    namespace: ci
+  spec:
+    selector:
+      app: prow
+      component: pj-rehearse-plugin
+    ports:
+    - port: 80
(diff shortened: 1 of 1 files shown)


clusters/app.ci/prow/03_deployment/pj-rehearse-plugin.yaml
```diff
@@ -0,0 +1,114 @@
...
+        - name: prowjob-kubeconfig
+          secret:
+            secretName: pj-rehearse
+        - name: hmac
+          secret:
+            secretName: github-webhook-credentials
+        - name: oauth
+          secret:
+            secretName: github-credentials-openshift-ci-robot
+        - name: cache
```
droslean, line 98: Does the tool generates a cache by default in `/var/tmp`?

smg247: yeah, that is the default for the git client:
https://github.com/openshift/ci-tools/blob/f80066832fb41b52cb536cf8ec01bbe8e40cee78/vendor/k8s.io/test-infra/prow/git/v2/client_factory.go#L112-L127

PR #33228, opened 2022-10-18
updating preflight main job to use go 1.19 and operator-sdk version

acornett21:
- Updating go dependency since preflight has a hard dependency on go 1.19 
- Updating operator-sdk version to match what we use in our container

Signed-off-by: Adam D. Cornett <adc@redhat.com>


skattoju (approved):
/lgtm

PR #33236, opened 2022-10-19
Update the upgrade step timeout to 24h

shellyyang1989:
To support upgrade test from 4.3/4.4 to latest, we are increasing the timeout of the upgrade step. [OCPQE-11543](https://issues.redhat.com/browse/OCPQE-11543)


shellyyang1989:
@liangxia could you please help review? Thanks!

PR #33242, opened 2022-10-19
Add TEST_TYPE env var for gcp-sdn minor upgrades.

dgoodwin:
Noticed this discrepancy compared to gcp-ovn minor, and the fact we are
only getting one set of spyglass intervals on these jobs, looks like
we're not running the upgrade conformance suite on this upgrade job.


tjungblu:
/lgtm

don't know if that will go through

PR #33244, opened 2022-10-19
Gather bootstrap logs on Restricted installations with Proxy

MaysaMacedo:
On restricted installations with Proxy, the bootstrap, masters and Proxy have private addresses, consequently is not possible to access the Proxy to gather the logs. This commit fixes the issue by accessing the bastion node, which has access to the restricted resources.

### DIFF
--- a/ci-operator/step-registry/openstack/gather/openstack-gather-commands.sh
+++ b/ci-operator/step-registry/openstack/gather/openstack-gather-commands.sh
@@ -15,6 +15,8 @@ then
     # shellcheck disable=SC1090
     source "${SHARED_DIR}/proxy-conf.sh"
     CREATE_FIPS=0
+    BASTION_FIP=$(<"${SHARED_DIR}/BASTION_FIP")
+    BASTION_USER=$(<"${SHARED_DIR}/BASTION_USER")
 fi
 
 collect_bootstrap_logs() {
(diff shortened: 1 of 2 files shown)


ci-operator/step-registry/openstack/gather/openstack-gather-commands.sh
```diff
@@ -56,9 +58,23 @@ collect_bootstrap_logs() {
...
+				SCP_CMD="scp $SSH_ARGS"
+				echo "Moving credentials and openshift binary to Bastion Proxy"
+				$SCP_CMD ${CLUSTER_PROFILE_DIR}/ssh-privatekey /bin/openshift-install $BASTION_USER@$BASTION_FIP:/tmp
+				SSH_PRIV_KEY_PATH="/tmp/ssh-privatekey"
+				echo "Gathering bootstrap logs from Bastion Proxy"
+				$SSH_CMD $BOOTSTRAP_GATHER_CMD
+				echo "Copying logs"
+				$SCP_CMD $BASTION_USER@$BASTION_FIP:/home/$BASTION_USER/log-bundle-*.tar.gz ${SHARED_DIR}/
+			else
+				cp log-bundle-*.tar.gz "${ARTIFACT_DIR}"
```
mandre, line 75: Hmm, you're no longer gathering bootstrap logs when not using a proxy.

ci-operator/step-registry/openstack/gather/openstack-gather-commands.sh
```diff
@@ -56,9 +58,23 @@ collect_bootstrap_logs() {
...
-			openshift-install gather bootstrap --key "${SSH_PRIV_KEY_PATH}" "${GATHER_BOOTSTRAP_ARGS[@]}"
-			cp log-bundle-*.tar.gz "${ARTIFACT_DIR}"
-			echo "${FIPS[@]}" | xargs --no-run-if-empty openstack floating ip delete
+			BOOTSTRAP_GATHER_CMD="/tmp/openshift-install gather bootstrap --key $SSH_PRIV_KEY_PATH ${GATHER_BOOTSTRAP_ARGS[@]}"
+			if test -f "${SHARED_DIR}/proxy-conf.sh"
+			then
+				SSH_ARGS="-o ConnectTimeout=10 -o "StrictHostKeyChecking=no" -i $SSH_PRIV_KEY_PATH"
+				SSH_CMD="ssh $SSH_ARGS $BASTION_USER@$BASTION_FIP"
+				SCP_CMD="scp $SSH_ARGS"
+				echo "Moving credentials and openshift binary to Bastion Proxy"
```
mandre, line 67: I still don't understand why it would be necessary to run the bootstrap gather from the bastion, as we're already using the bastion as a proxy :thinking:

MaysaMacedo: @mandre From my understanding the installer [uses its ssh client](https://github.com/openshift/installer/blob/master/cmd/openshift-install/gather.go#L165) to gather the logs from all the servers. I don't see any specification of proxy host being defined in the command. I'm not so sure how that would be handled in the installer, maybe specifying a ProxyCommand or the jump host?

In any case, it looks like that is not implemented at the moment and it might be faster to handle it in our step atm, if possible.

mandre: Oh, of course... it uses SSH and not HTTP, the proxy is not going to work in this case. It would be nice to configure the SSH client to jump through the bastion rather than copying the binaries and logs to and from the bastion but I don't think that would be doable via a simple ssh config, we would have to change [the installer code](https://github.com/openshift/installer/blob/5a9ad4c024091c6658813ab3a04c8deb61d684cc/pkg/gather/ssh/ssh.go#L23).

MaysaMacedo:
@EmilienM can you have a look on this one?
I tested locally and it worked. We won't see this in the CI anymore because Proxy CI is progressing and not failing on bootstrap

ci-operator/step-registry/openstack/gather/openstack-gather-commands.sh
```diff
@@ -15,6 +15,8 @@ then
     # shellcheck disable=SC1090
     source "${SHARED_DIR}/proxy-conf.sh"
     CREATE_FIPS=0
+	BASTION_FIP=$(<"${SHARED_DIR}/BASTION_FIP")
```
EmilienM, line 18: tab issue?

PR #33251, opened 2022-10-19
[Prow CI Step Enablement] add step idp-htpasswd to support HyperShift guest cluster e2e test 

LiangquanLi930:
add step `cucushift-hypershift-extended-idp-htpasswd` for support hypershift guest cluster e2e test

### DIFF
--- a/ci-operator/step-registry/cucushift/hypershift-extended/idp-htpasswd/cucushift-hypershift-extended-idp-htpasswd-commands.sh
+++ b/ci-operator/step-registry/cucushift/hypershift-extended/idp-htpasswd/cucushift-hypershift-extended-idp-htpasswd-commands.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+
+export LC_CTYPE=C
+set -o nounset
+set -o errexit
+set -o pipefail
+
+export KUBECONFIG="${SHARED_DIR}/kubeconfig"
+count=$(oc get hostedclusters --no-headers --ignore-not-found -n clusters | wc -l)
+echo "hostedcluster count: $count"
+if [ "$count" -lt 1 ]  ; then
+    echo "namespace clusters don't have hostedcluster"
+    exit 1
+fi
+#Limitation: we always & only select the first hostedcluster to add idp-htpasswd. "
+cluster_name=$(oc get hostedclusters -n clusters -o jsonpath='{.items[0].metadata.name}')
+
+# prepare users
+users=""
+htpass_file=/tmp/users.htpasswd
+
+for i in $(seq 1 40);
+do
+    username="testuser-${i}"
+    password=$(< /dev/urandom tr -dc 'a-z0-9' | fold -w 12 | head -n 1 || true)
+    users+="${username}:${password},"
+    if [ -f "${htpass_file}" ]; then
+        htpasswd -B -b ${htpass_file} "${username}" "${password}"
+    else
+        htpasswd -c -B -b ${htpass_file} "${username}" "${password}"
+    fi
+done
+
+## add users to cluster
+oc create secret generic "$cluster_name" --from-file=htpasswd="$htpass_file" -n clusters
+oc patch hostedclusters $cluster_name -n clusters --type=merge -p '{"spec":{"configuration":{"oauth":{"identityProviders":[{"htpasswd":{"fileData":{"name":"'$cluster_name'"}},"mappingMethod":"claim","name":"htpasswd","type":"HTPasswd"}]}}}}'
+
+# store users in a shared file
+if [ -f "${SHARED_DIR}/runtime_env" ] ; then
+    source "${SHARED_DIR}/runtime_env"
+fi
+runtime_env=${SHARED_DIR}/runtime_env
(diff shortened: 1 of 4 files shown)


LiangquanLi930:
OCPQE-12552

ci-operator/step-registry/cucushift/hypershift-extended/idp-htpasswd/cucushift-hypershift-extended-idp-htpasswd-commands.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+
+export LC_CTYPE=C
+set -o nounset
+set -o errexit
+set -o pipefail
+
+export KUBECONFIG="${SHARED_DIR}/kubeconfig"
+count=$(oc get hostedclusters --no-headers --ignore-not-found -n clusters | wc -l)
+echo "$count"
```
heliubj18, line 10: better to print msg readable like  echo "hostedcluster count: $count"

ci-operator/step-registry/cucushift/hypershift-extended/idp-htpasswd/cucushift-hypershift-extended-idp-htpasswd-commands.sh
```diff
@@ -0,0 +1,46 @@
...
+if [ "$count" -lt 1 ]  ; then
+    echo "namespace clusters don't have hostedcluster"
+    exit 1
+fi
+cluster_name=$(oc get hostedclusters -n clusters -o jsonpath='{.items[0].metadata.name}')
```
heliubj18, line 15: Not sure we add oauth for all of hostedcluster or just the first one ?  If only the first one added, then we should always follow this rule for the guest cluster kubeconfig exposion.

Suggest add description like : "Limitation: we always & only select the first hostedcluster to add idp-htpasswd. "

ci-operator/step-registry/cucushift/hypershift-extended/idp-htpasswd/cucushift-hypershift-extended-idp-htpasswd-commands.sh
```diff
@@ -0,0 +1,46 @@
...
+    if [ -f "${htpass_file}" ]; then
+        htpasswd -B -b ${htpass_file} "${username}" "${password}"
+    else
+        htpasswd -c -B -b ${htpass_file} "${username}" "${password}"
+    fi
+done
+
+## add users to cluster
+oc create secret generic "$cluster_name" --from-file=htpasswd="$htpass_file" -n clusters
+oc patch hostedclusters $cluster_name -n clusters --type=merge -p '{"spec":{"configuration":{"oauth":{"identityProviders":[{"htpasswd":{"fileData":{"name":"'$cluster_name'"}},"mappingMethod":"claim","name":"htpasswd","type":"HTPasswd"}]}}}}'
```
heliubj18, line 35: suggest checking if `spec.configuration.oauth` exists or not

ci-operator/step-registry/cucushift/hypershift-extended/idp-htpasswd/cucushift-hypershift-extended-idp-htpasswd-commands.sh
```diff
@@ -0,0 +1,46 @@
...
+
+# store users in a shared file
+if [ -f "${SHARED_DIR}/runtime_env" ] ; then
+    source "${SHARED_DIR}/runtime_env"
```
heliubj18, line 39: suggest put all the similar codes together special for [cucushift-pre-commands.sh](https://github.com/openshift/release/blob/master/ci-operator/step-registry/cucushift/pre/cucushift-pre-commands.sh) to make it clear.

Another suggestion:  In `cucushift-pre-commands.sh` it just add the new users into the existing idp oauth users instead of removing the existing users.  I'd like to follow their rules to enable all the users in the management cluster...

heliubj18:
You don't rollout/restart OAUTH pods if they exist. This script will not take effect in this case.
Suggestion:
1. Not sure `oc rollout` is valid here, but we need enable the new configuration by rollout or rebuild the pods
2. check the new configuration is valid by "oc login ..."

ci-operator/step-registry/cucushift/hypershift-extended/idp-htpasswd/cucushift-hypershift-extended-idp-htpasswd-commands.sh
heliubj18: another PR for the checkout process together with rollout later.

LiangquanLi930:
> You don't rollout/restart OAUTH pods if they exist. This script will not take effect in this case. Suggestion:
> 
> 1. Not sure `oc rollout` is valid here, but we need enable the new configuration by rollout or rebuild the pods
> 2. check the new configuration is valid by "oc login ..."

do not need rollout, it is set in the hostedcluster CR. The day 2 operation will trigger the oauth update automatically.

PR #33274, opened 2022-10-20
Skip admin ack when source version is 4.7 and earlier

shellyyang1989:


ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -350,8 +350,8 @@ function check_signed() {
 
 # Check if admin ack is required before upgrade
 function admin_ack() {
-    if [[ "${SOURCE_MINOR_VERSION}" -eq "${TARGET_MINOR_VERSION}" ]]; then
-        echo "Upgrade between z-stream version does not require admin ack" && return
+    if (( SOURCE_MINOR_VERSION == TARGET_MINOR_VERSION )) || (( SOURCE_MINOR_VERSION < 8 )); then
```
jiajliu, line 353: Hmm, if we want to avoid unnecessary check, how about skip more? Maybe we could maintain a list. skip_list(<4.8,4.9/4.10/4.12) vs check_list(4.8/4.11), looks like check_list is more shorter?

jiajliu: /cc @evakhoni ^

shellyyang1989: hmm, the change aims to fix the issue found in [4.4 cluster](https://gcsweb-qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/qe-private-deck/pr-logs/pull/openshift_release/33160/rehearse-33160-periodic-ci-openshift-openshift-tests-private-release-4.12-amd64-nightly-4.12-upgrade-from-stable-4.4-aws-ipi/1582664235006038016/artifacts/aws-ipi/cucushift-upgrade-toimage/build-log.txt):
Error from server (NotFound): configmaps "admin-gates" not found

The admin ack was introduced in 4.8, so we skip it in 4.7 and earlier straightforwardly.

For those 4.8 and later versions which does not require admin ack, we have the code block to guard.

```
local out; out="$(oc -n openshift-config-managed get configmap admin-gates -o json | jq -r ".data")"
    if [[ ${out} != *"ack-4.${SOURCE_MINOR_VERSION}"* ]]; then
        echo "Admin ack not required" && return
    fi
```

jiajliu: Good to know.
btw, why we still need the upgrade path from 4.4 in prow?

shellyyang1989: ah, because we have many customer clusters upgrading from very old releases and we have found some [upgrade issues ](https://bugzilla.redhat.com/show_bug.cgi?id=2110501)occurring only for the clusters born in old releases. We have a jira issue [OCPQE-11543](https://issues.redhat.com/browse/OCPQE-11543) for it.

jiajliu: Ack! I thought we only left the long-long path in Jenkins...

PR #33292, opened 2022-10-20
Enabling postsubmits for governance-policy-propagator on arm01

deepsm007:
This will enable building images on arm01 prow cluster and images will be mirrored to apps.ci internal registry

/cc @mprahl @gurnben


gurnben (approved):
/approve

I'll let @mprahl do the rest, but these look good to me!

PR #33298, opened 2022-10-20
Remove single-node Azure tests

omertuc:
Since they are failing often and we do not currently support SNO azure
deployments, it's better to disable these jobs for now


dgoodwin:
Looks good just a config update. Kudos on getting the release controller as well.

PR #33305, opened 2022-10-20
WIP: Test sippy-failure-risk-analysis-post step.

dgoodwin:


dgoodwin:
We don't have access to the artifacts from other steps, and shared dir has some scary size restrictions. I'm going to close this and attempt another way, by uploading at the end of the standard openshift tests step.

/close

PR #33311, opened 2022-10-21
Add more azure jobs for arm64

lwan-wanglin:
According to azure jobs for amd64 now we have, add corresponding jobs for arm64, but adjust the frequency.

refer [ARMOCP-222](https://issues.redhat.com/browse/ARMOCP-222) and [ARMOCP-221](https://issues.redhat.com/browse/ARMOCP-221)

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__arm64-nightly.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__arm64-nightly.yaml
@@ -226,8 +226,72 @@ tests:
     test:
     - chain: openshift-e2e-test-qe
     workflow: cucushift-installer-rehearse-aws-ipi-proxy-cco-manual-security-token-service
-- as: e2e-azure-ipi-ovn-p1
-  cron: 0 15 */2 * *
+- as: e2e-azure-ipi-disconnected-p1
+  cron: 0 0 * * *
+  steps:
+    cluster_profile: azure-arm64-qe
+    dependencies:
+      OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE: release:arm64-latest
+    env:
+      BASE_DOMAIN: qe.azure.devcluster.openshift.com
+      COMPUTE_NODE_TYPE: Standard_D4ps_v5
+      E2E_RUN_TAGS: '@azure-ipi and @arm64 and @disconnected and @network-ovnkubernetes
+        and not @fips'
+      OCP_ARCH: arm64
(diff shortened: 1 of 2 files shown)


lwan-wanglin:
All the jobs installation are success
@aleskandro @liangxia could you help reivew and merge?  cc @MayXuQQ @jianlinliu

aleskandro:
Hello @lwan-wanglin, thanks. you'd look at https://github.com/openshift/release/blob/master/ci-operator/config/openshift/openshift-tests-private/README.md to set the name of the jobs according to the frequency. 

Moreover I'd keep at the same frequency as for amd64 for the initial weeks (especially for azure ipi that is new).

After we reach a steady state (especially on the test suites results), we could re-discuss the frequencies for both the arches (e.g., not halving only the amd64 ones). We already discussed this with @liangxia a few time ago.

lwan-wanglin:
> Hello @lwan-wanglin, thanks. you'd look at https://github.com/openshift/release/blob/master/ci-operator/config/openshift/openshift-tests-private/README.md to set the name of the jobs according to the frequency.
>
I thought I was setting the name according to the frequency, so `0 9 */10 * * ` is not belong to p2(weekly)?

> Moreover I'd keep at the same frequency as for amd64 for the initial weeks (especially for azure ipi that is new).
> 
Thanks, got it, that makes sense! so do we need to make all azure jobs triggered once a day in the initial weeks? Why I ask so, because there are some of jobs for amd64 are setting weekly and monthly.
> After we reach a steady state on the test suites results, we could re-discuss the frequencies for both the arches (e.g., not halving only the arm64 ones). We already discussed this with @liangxia a few time ago.
Ok.

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__arm64-nightly.yaml
```diff
@@ -226,6 +226,55 @@ tests:
...
+      E2E_RUN_TAGS: '@azure-ipi and @arm64 and @disconnected and @network-ovnkubernetes
+        and not @fips'
+      OCP_ARCH: arm64
+      TAG_VERSION: '@4.12'
+      TEST_FILTERS: ~ChkUpgrade&;~NonPreRelease&;~Serial&;~Disruptive&;~ConnectedOnly&
+    test:
+    - chain: openshift-e2e-test-qe
+    workflow: cucushift-installer-rehearse-azure-ipi-disconnected
+- as: e2e-azure-ipi-disconnected-fullyprivate-p2
+  cron: 0 7 */3 * *
```
liangxia, line 247: I think we are running p2 profiles every week (*/7), based on the rules we currently have.

PR #33312, opened 2022-10-21
Try again if API server is not accessible temporarily

shellyyang1989:
API server might be inaccessible temporarily during upgrade and we will get below error from oc cli. In this case, we wait and try again till it's accessible instead of exiting directly. 

```
The connection to the server api.ci-op-vnw3mwrr-abd45.qe.devcluster.openshift.com:6443 was refused - did you specify the right host or port?
```


shellyyang1989:
Upgrade passed in the [job](https://qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/qe-private-deck/pr-logs/pull/openshift_release/33312/rehearse-33312-periodic-ci-openshift-openshift-tests-private-release-4.12-amd64-nightly-4.12-upgrade-from-stable-4.11-aws-ipi-byo-iam-role-p2/1583298584361570304).

shellyyang1989:
cc @jianlinliu @jiajliu @evakhoni for review. Thanks!

PR #33317, opened 2022-10-21
 add ALLOW_ROOTLESS true to the container

xinredhat:
Added ALLOW_ROOTLESS=true as variable before executing openshift_dev_setup.sh


bnallapeta (approved):
/lgtm

PR #33326, opened 2022-10-21
WIP ipi-conf-vsphere-chain.yaml: don't write audit to disk

vrutkovs:
This adds a step which configures Kube API server to send audit logs to /dev/null, saving disk IO


vrutkovs:
Its silly and not helping

PR #33327, opened 2022-10-21
Add optional nfv jobs to CAPO and MAPO

mandre:
These jobs exercises additional networks.


mandre:
The tests ran. That's what we asked them to to.
/hold cancel

PR #33347, opened 2022-10-21
add configurable RELEASE_IMAGE

LiangquanLi930:
now `create hostedcluster` chain Dynamic configuration RELEASE_IMAGE is not supported, So have this pr


sjenning:
Seems fine.  What needs this?

heliubj18:
@sjenning In some QE's HyperShift CI jobs, we need set some specified hosted cluster payload image for OCP components e2e test.

PR #33352, opened 2022-10-21
[CCO e2e-aws-manual-oidc]: Copy generated secret manifests into shared directory.

abutcher:
Also increase verbosity of the provisioning script to make it easier to debug.

### DIFF
--- a/ci-operator/step-registry/ipi/conf/aws/oidc-creds-provision/ipi-conf-aws-oidc-creds-provision-commands.sh
+++ b/ci-operator/step-registry/ipi/conf/aws/oidc-creds-provision/ipi-conf-aws-oidc-creds-provision-commands.sh
@@ -10,16 +10,19 @@ infra_name=${NAMESPACE}-${JOB_NAME_HASH}
 export AWS_SHARED_CREDENTIALS_FILE="${CLUSTER_PROFILE_DIR}/.awscred"
 REGION="${LEASED_RESOURCE}"
 
-# extract aws credentials requests from the release image
+echo "> Extract aws credentials requests from the release image"
 oc registry login
 oc adm release extract --credentials-requests --cloud=aws --to="/tmp/credrequests" "$RELEASE_IMAGE_LATEST"
 
-# create required credentials infrastructure and installer manifests
+echo "> Generated the following CredentialsRequests"
+ls /tmp/credrequests
+
+echo "> Create required credentials infrastructure and installer manifests"
 ccoctl aws create-all --name="${infra_name}" --region="${REGION}" --credentials-requests-dir="/tmp/credrequests" --output-dir="/tmp"
 
-# copy generated service account signing from ccoctl target directory into shared directory
-cp "/tmp/tls/bound-service-account-signing-key.key" "${TPREFIX}_bound-service-account-signing-key.key"
+echo "> Copy generated service account signing from ccoctl target directory into shared directory"
+cp -v "/tmp/tls/bound-service-account-signing-key.key" "${TPREFIX}_bound-service-account-signing-key.key"
 
-# copy generated secret manifests from ccoctl target directory into shared directory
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/ipi/conf/aws/oidc-creds-provision/ipi-conf-aws-oidc-creds-provision-commands.sh
```diff
@@ -10,16 +10,19 @@ infra_name=${NAMESPACE}-${JOB_NAME_HASH}
...
-# copy generated service account signing from ccoctl target directory into shared directory
-cp "/tmp/tls/bound-service-account-signing-key.key" "${TPREFIX}_bound-service-account-signing-key.key"
+echo "> Copy generated service account signing from ccoctl target directory into shared directory"
+cp -v "/tmp/tls/bound-service-account-signing-key.key" "${TPREFIX}_bound-service-account-signing-key.key"
 
-# copy generated secret manifests from ccoctl target directory into shared directory
+echo "> Copy generated secret manifests from ccoctl target directory into shared directory"
 cd "/tmp/manifests"
-for FILE in *; do cp $FILE "${MPREFIX}_$FILE"; done
+for FILE in *; do cp -v $FILE "${MPREFIX}/$FILE"; done
```
abutcher, line 28: In debugging #491, @mtulio pointed out that it seemed like we were copying secret manifests to the wrong directory.

ci-operator/step-registry/ipi/conf/aws/oidc-creds-provision/ipi-conf-aws-oidc-creds-provision-commands.sh
```diff
@@ -10,16 +10,19 @@ infra_name=${NAMESPACE}-${JOB_NAME_HASH}
...
+echo "> Generated the following CredentialsRequests"
+ls /tmp/credrequests
+
+echo "> Create required credentials infrastructure and installer manifests"
 ccoctl aws create-all --name="${infra_name}" --region="${REGION}" --credentials-requests-dir="/tmp/credrequests" --output-dir="/tmp"
 
-# copy generated service account signing from ccoctl target directory into shared directory
-cp "/tmp/tls/bound-service-account-signing-key.key" "${TPREFIX}_bound-service-account-signing-key.key"
+echo "> Copy generated service account signing from ccoctl target directory into shared directory"
+cp -v "/tmp/tls/bound-service-account-signing-key.key" "${TPREFIX}_bound-service-account-signing-key.key"
```
mtulio, line 24: @abutcher should we need to fix here too?
```suggestion
cp -v "/tmp/tls/bound-service-account-signing-key.key" "${TPREFIX}/bound-service-account-signing-key.key"
```

mtulio: You can ignore, as you commented the TLS key should be in the root of install dir

abutcher: I think this one is correct as is since the key goes into the install directory as opposed to the manifests directory.

ci-operator/step-registry/ipi/conf/aws/oidc-creds-provision/ipi-conf-aws-oidc-creds-provision-commands.sh
abutcher: The prefixed name is correct with an `_` to be copied in here https://github.com/openshift/release/blob/25c3e66da3c270819b3898067cc2703e75be7f2a/ci-operator/step-registry/ipi/install/install/ipi-install-install-commands.sh#L316.

PR #33358, opened 2022-10-23
HCO: require specific node type on GCP also on OKD

tiraboschi:
Require COMPUTE_NODE_TYPE: n2-standard-4
also on OKD to be sure we get nodes with
virtualization support.


nunnatsa (approved):
/approve

PR #33363, opened 2022-10-24
several enhancements for behind proxy and disconnected cases

jianlinliu:
1. change "squid-proxy" service name to "proxy" to work with gather-proxy step
2. create "proxyip" file in aws/vsphere bastion host step
3. pull "registry:2" image from quay.io instead of docker.io
4. add `gather-mirror-registry` step in deprovision chains 
5. update `gather-mirror-registry`, public ip has higher priority than private ip
6. [gcp] fix pull secret override issues introduced by https://github.com/openshift/release/commit/77d93204e74370997ecbaf9ec43c5ec480236c16, together with azure, similar fix with https://github.com/openshift/release/pull/32949


jianlinliu:
cc @jinyunma @yunjiang29 @jianli-wei for review.

jinyunma:
checked that deprovision chain for vsphere-ipi-proxy and vsphere-upi-proxy doesn't have step `gather-proxy` yet, do we need to add it in this PR?

jianlinliu:
> checked that deprovision chain for vsphere-ipi-proxy and vsphere-upi-proxy doesn't have step `gather-proxy` yet, do we need to add it in this PR?

Done

PR #33370, opened 2022-10-24
Increase aws qe slices

yunjiang29:
Adding 20 slices to AWS QE


yunjiang29:
@gpei @jianlinliu PTAL.

PR #33372, opened 2022-10-24
MGMT-12334: Remove use of cloud-init in ZTP jobs

osherdp:
It seems like something in the supplied cloud-init makes ZTP jobs unable to acquire machines. This change just tries to mitigate the issue by not using this method, until we can further analyze the situation.

/cc @eliorerz @adriengentil


adriengentil (approved):

osherdp:
seems to work
/unhold

PR #33381, opened 2022-10-24
workflow of enabling user to set lbType in install-config

lihongan:
to address https://issues.redhat.com/browse/OCPQE-12477 which is related to 4.12 epic: https://issues.redhat.com/browse/NE-942

Before we should create manifest to change the LB from default Classic to NLB, with https://issues.redhat.com/browse/NE-942 user can set lbType in install-config directly, the default is still "Classic" in 4.12 but subject to change over time.


lihongan:
the test for `workflow: cucushift-installer-rehearse-aws-ipi-ingress-lbtype-classic` looks good, see ingresscontroller status as below
```
    endpointPublishingStrategy:
      loadBalancer:
        dnsManagementPolicy: Managed
        providerParameters:
          aws:
            classicLoadBalancer:
              connectionIdleTimeout: 0s
            type: Classic
          type: AWS
        scope: External
      type: LoadBalancerService
```

let's test another `workflow: cucushift-installer-rehearse-aws-ipi-ingress-lbtype-nlb`

lihongan:
@jianlinliu could you help take a look?  failed twice at `ipi-deprovision-deprovision` and not sure if new bug in destroy? 
This is only happened when setting `lbType="NLB"` in install-config.

jianlinliu:
> @jianlinliu could you help take a look? failed twice at `ipi-deprovision-deprovision` and not sure if new bug in destroy? This is only happened when setting `lbType="NLB"` in install-config.

This is similar with https://coreos.slack.com/archives/CH76YSYSC/p1666640480489669?thread_ts=1666607038.845479&cid=CH76YSYSC recently we hit several such issues in ci, but we did not found the root cause, maybe you can involve that discussion to post the actions what was done in this case help provide more info there.

lihongan:
/test pj-rehearse
try lbtype-classic workflow but `Step installer-rehearse-aws-ipi-install-install failed`, test again

lihongan:
Test the old ingress-nlb workflow which using manifests in 4.12 and 4.11 to see if any issue.

lihongan:
Interesting, tested the same workflow in 4.11 and passed, but failed in 4.12 `ipi-deprovision-deprovision` step.
@jianlinliu @wking any idea?

lihongan:
AWS tags issue maybe related to https://github.com/openshift/kubernetes/pull/1401, let me retest after the PR merged.

PR #33396, opened 2022-10-24
re-enable bundle index build for CLO

jcantrill:
This PR:

* Re-enables the bundle index build for the CLO


ci-operator/jobs/infra-image-mirroring-multi-arch.yaml
```diff
@@ -1,4 +1,3 @@
-# DO NOT EDIT. This file is auto-generated.
```
smg247, line 1: Was this intentional?

jcantrill: No.  I have no idea why this file changed

jcantrill: rebased. repushed

PR #33409, opened 2022-10-24
update golang version to 1.17 for procfs

raptorsun:
new version of procfs requires golang 1.17+ to run tests.
This PR deblocks the test for: https://github.com/openshift/procfs/pull/7


raptorsun:
/close
We do not use openshift/procfs now.

PR #33411, opened 2022-10-24
WINC-830: ci-operator: Fix getting Windows VM IPv4 address

aravindhp:
Use the `govc vm.ip -a -v4 -wait=20m` command as it will internally use VMTools to return the IPv4 address from the VMXNET3 interface. The previous command was returning a transient IPv4 address.

Remove the redundant wait loop for the IP address.


aravindhp:
/retitle WINC-830: ci-operator: Fix getting Windows VM IPv4 address

aravindhp:
/hold

The wait loop is redundant and should be removed

PR #33413, opened 2022-10-25
Rename baremetal jobs to reflect the profile we are testing with

liangxia:
We'd like the job name to reflect the workflow(profile), so that we can know which profile we are testing with when we are checking the results in ReportPortal.

Changes made in the PR,
1. Rename job from `ocp-e2e-metal-libvirt-ipi-ovn-p1` to `e2e-baremetalds-ipi-ovn-ipv4-p1` to reflect we are testing profile `baremetalds-ipi-ovn-ipv4`
2. Sort the job in alphanumeric order

/cc @lwan-wanglin @LiangquanLi930 @aleskandro


lwan-wanglin:
Look good to me!
/lgtm

PR #31612, opened 2022-08-23
MGMT-11633: removing kube-api-net-suite jobs until problem is resolved

osherdp:

PR #33427, opened 2022-10-25
MGMT-11633: Re enable kube-api job

paul-maidment:
This was previously disabled in https://github.com/openshift/release/pull/31612 and is now being re-enabled after a fix has been merged for the root issue https://github.com/openshift/assisted-test-infra/pull/1891

### DIFF
--- a/ci-operator/config/openshift/assisted-test-infra/openshift-assisted-test-infra-master.yaml
+++ b/ci-operator/config/openshift/assisted-test-infra/openshift-assisted-test-infra-master.yaml
@@ -431,7 +431,7 @@ tests:
       TEST_SUITE: full
     workflow: assisted-baremetal
 - as: e2e-metal-assisted-kube-api-net-suite
-  optional: true
+  skip_if_only_changed: ^docs/|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|OWNERS_ALIASES|PROJECT|LICENSE)$
   steps:
     cluster_profile: packet-assisted
     env:
@@ -443,6 +443,21 @@ tests:
         STATIC_IPS=true
         TEST_FUNC=test_kubeapi
         MAKEFILE_TARGET="test_kube_api_parallel"
+    workflow: assisted-baremetal
+- as: e2e-metal-assisted-kube-api-net-suite-periodic
+  cron: 43 22 */1 * 1-5
+  steps:
+    cluster_profile: packet-assisted
+    env:
+      ASSISTED_CONFIG: |
+        KUBE_API=true
+        ENABLE_KUBE_API=true
+        NUM_MASTERS=1
+        ISO_IMAGE_TYPE=minimal-iso
+        STATIC_IPS=true
+        TEST_FUNC=test_kubeapi
+        MAKEFILE_TARGET="test_kube_api_parallel"
+      TEST_SUITE: full
     workflow: assisted-baremetal
 - as: e2e-metal-assisted-ha-kube-api
   optional: true
--- a/ci-operator/jobs/openshift/assisted-test-infra/openshift-assisted-test-infra-master-periodics.yaml
+++ b/ci-operator/jobs/openshift/assisted-test-infra/openshift-assisted-test-infra-master-periodics.yaml
@@ -711,6 +711,77 @@ periodics:
         secretName: result-aggregator
 - agent: kubernetes
   cluster: build03
+  cron: 43 22 */1 * 1-5
+  decorate: true
+  extra_refs:
+  - base_ref: master
+    org: openshift
+    repo: assisted-test-infra
+  labels:
+    ci-operator.openshift.io/cloud: packet-edge
+    ci-operator.openshift.io/cloud-cluster-profile: packet-assisted
+    ci.openshift.io/generator: prowgen
+    job-release: "4.12"
+    pj-rehearse.openshift.io/can-be-rehearsed: "true"
+  name: periodic-ci-openshift-assisted-test-infra-master-e2e-metal-assisted-kube-api-net-suite-periodic
+  spec:
+    containers:
+    - args:
+      - --gcs-upload-secret=/secrets/gcs/service-account.json
+      - --image-import-pull-secret=/etc/pull-secret/.dockerconfigjson
+      - --lease-server-credentials-file=/etc/boskos/credentials
+      - --report-credentials-file=/etc/report/credentials
+      - --secret-dir=/secrets/ci-pull-credentials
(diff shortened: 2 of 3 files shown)


ci-operator/config/openshift/assisted-test-infra/openshift-assisted-test-infra-master.yaml
```diff
@@ -58,6 +58,10 @@ releases:
       product: ocp
       stream: nightly
       version: "4.12"
+  stable-4-11:
+    integration:
+      name: "4.11"
+      namespace: ocp
```
osherdp, line 64: iiuc this doesn't pick stable 4.11 versions (4.11.z)
but rather you'll get ``ci`` stream, which is what's merged to the ``release-4.11`` branches (e.g. ``4.11.0-0.ci.test-2022-11-23-210525``)

ci-operator/config/openshift/assisted-test-infra/openshift-assisted-test-infra-master.yaml
```diff
@@ -431,9 +435,11 @@ tests:
       TEST_SUITE: full
     workflow: assisted-baremetal
 - as: e2e-metal-assisted-kube-api-net-suite
-  optional: true
+  skip_if_only_changed: ^docs/|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|OWNERS_ALIASES|PROJECT|LICENSE)$
   steps:
     cluster_profile: packet-assisted
+    dependencies:
+      OPENSHIFT_INSTALL_RELEASE_IMAGE: release:stable-4-11
```
osherdp, line 442: why do we need stable 4.11 here? is there a bug-fix that exists in 4.11 but not in latest 4.12 (which should be the default afaik)

ci-operator/config/openshift/assisted-test-infra/openshift-assisted-test-infra-master.yaml
```diff
@@ -512,6 +518,25 @@ tests:
         TEST_FUNC=test_networking
       PACKET_PLAN: m3.large.x86
     workflow: assisted-baremetal
+- as: e2e-metal-assisted-kube-api-net-suite-periodic
```
osherdp, line 521: move it right after the presubmit job, to have it better organized

osherdp (changes requested):

ci-operator/config/openshift/assisted-test-infra/openshift-assisted-test-infra-master.yaml
paul-maidment: I have decided to remove this as I am going with advice you give later on about using the default.
Though I am curious about this because  I notice that the config here uses similar terminology

https://github.com/openshift/release/blob/master/ci-operator/config/openshift/assisted-service/openshift-assisted-service-master__edge.yaml#L170-L177

Maybe I need a better understanding of release streams, I will chat with you about this I think.

ci-operator/config/openshift/assisted-test-infra/openshift-assisted-test-infra-master.yaml
paul-maidment: Removed on your advice here.
At the time of writing, your patch had not yet been merged or it was too close to the merge time to be sure that it would be picked up properly.
That's no longer an issue now so I am removing this.

ci-operator/config/openshift/assisted-test-infra/openshift-assisted-test-infra-master.yaml
paul-maidment: Done

ci-operator/config/openshift/assisted-test-infra/openshift-assisted-test-infra-master.yaml
```diff
@@ -512,6 +518,25 @@ tests:
...
+      OPENSHIFT_INSTALL_RELEASE_IMAGE: release:stable-4-11
+    env:
+      ASSISTED_CONFIG: |
+        KUBE_API=true
+        ENABLE_KUBE_API=true
+        NUM_MASTERS=1
+        ISO_IMAGE_TYPE=minimal-iso
+        STATIC_IPS=true
+        TEST_FUNC=test_kubeapi
+        MAKEFILE_SETUP_TARGET="setup run"
```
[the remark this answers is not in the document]
paul-maidment: Done

ci-operator/config/openshift/assisted-test-infra/openshift-assisted-test-infra-master.yaml
```diff
@@ -512,6 +518,25 @@ tests:
...
+        MAKEFILE_TARGET="test_kube_api_parallel"
+      PACKET_PLAN: m3.large.x86
```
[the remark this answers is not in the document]
paul-maidment: I think this was a side effect of trying to resolve the bug that was caused by putting "setup run" makefile targets in the MAKEFILE_TARGET env.

Now that this has been resolved. I will remove this.

ci-operator/config/openshift/assisted-test-infra/openshift-assisted-test-infra-master.yaml
osherdp: Yes, we probably did the same mistake there
As for the documentation for those streams: https://docs.ci.openshift.org/docs/architecture/ci-operator/#testing-with-an-existing-openshift-release

core-services/prow/02_config/_config.yaml
```diff
@@ -314,15 +314,6 @@ slack_reporter_configs:
     channel: '#ops-testplatform'
     report_template: Job {{.Spec.Job}} of type {{.Spec.Type}} ended with state {{.Status.State}}.
       <{{.Status.URL}}|View logs>
-  openshift-assisted/assisted-installer-deployment:
```
osherdp, line 317: weird to see this change in your PR (I did it a couple of days ago)
it only affect the ownership, making dptp required to approve it. so I suggest you to rebase with master and remove this change from your commit

ci-operator/config/openshift/assisted-service/openshift-assisted-service-master__edge.yaml
```diff
@@ -792,6 +792,8 @@ tests:
   optional: true
   steps:
     cluster_profile: packet-assisted
+    dependencies:
+      OPENSHIFT_INSTALL_RELEASE_IMAGE: release:stable-4-11
```
osherdp, line 796: your removed it from the periodic but here it still exists

osherdp (changes requested):
just two more changes

osherdp (approved):

PR #33445, opened 2022-10-25
hypershift: stagger periodic cron schedule

sjenning:
Currently, all our periodic jobs have the same cron schedule which creates a situation every 6 hours where 4 AWS based jobs all start at once creating stress on the CI cluster and AWS quotas.

This PR staggers job start times and reduce *conformance jobs to once per day.

Chose start times that avoid business hours for team members where PR job activity could be high.

@enxebre @jparrill


jparrill (commented):
/lgtm

PR #33469, opened 2022-10-26
OCPQE-11842: Add e2e for destructive tests on Azure

liangxia:


sunzhaohua2:
Not sure if it's correct, but I feel installer base workflow such as `auzre-ipi`,` azure-ipi-private`, `azure-ipi-proxy`, we don't need a job to test, they will be covered in other complex workflow

liangxia:
> Not sure if it's correct, but I feel installer base workflow such as `auzre-ipi`,` azure-ipi-private`, `azure-ipi-proxy`, we don't need a job to test, they will be covered in other complex workflow

Based on a conversation earlier with Wei Sun, Jianwei and several others, we prefer to keep them there, but set the basic profiles with less test frequency. E.g, `azure-ipi` can be covered by `azure-ipi-private`, we keep the tests for both profile, and set  test frequency of `azure-ipi` to much less than `azure-ipi-private` (every 28 days and every 7 days).
There are several reasons for that. e.g, a tests need to exist (despite of the test frequency) in Prow so that we can manually trigger them when necessary. Also if a installation on `azure-ipi-private` fails, we can check whether `azure-ipi` works or not to quickly narrow down the possible cause.

PR #33478, opened 2022-10-26
fix ginkgo_filter for aws-ipi-disconnected-private-p1

Xia-Zhao-rh:
@kuiwang02 Please help to review it, thanks


ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-nightly-4.12-upgrade-from-stable-4.11.yaml
```diff
@@ -86,6 +86,7 @@ tests:
     cluster_profile: aws-qe
     env:
       BASE_DOMAIN: qe.devcluster.openshift.com
+      TEST_FILTERS: ~ConnectedOnly&
```
kuiwang02, line 89: @Xia-Zhao-rh better to be `TEST_FILTERS: ~ ConnectedOnly&;~HyperShiftMGMT&` although there is no HyperShiftMGMT for upgrade now.

PR #33479, opened 2022-10-26
osde2e: add rosa prod/stage sts e2e tests

jbpratt:


MrSantamaria:
/needs-ok-to-test
We are currently creating Roles and iodcs in the aws account that are not being cleaned up

MrSantamaria:
/assign @MrSantamaria 
We can submit this once it's rebased since our cleanup is working as expected.

PR #33484, opened 2022-10-26
Remove the release-2.7 and main branches from the deprecated sync controllers

mprahl:


dhaiducek (approved):
LGTM!

Good call! Should we also archive the repos upstream and uninstall magic-mirror from them?

mprahl:
> LGTM!
> 
> Good call! Should we also archive the repos upstream and uninstall magic-mirror from them?

Good point! I archived the upstream repos now. I'll go ahead and request that Magic Mirror be disabled on those repos.

PR #33488, opened 2022-10-26
MGMT-12365: Updating release-controller assisted job name

danmanor:
OVNKubernetes is the default network type for OCP version >= 4.12
https://issues.redhat.com/browse/MGMT-11293

### DIFF
--- a/ci-operator/jobs/openshift/release/openshift-release-master-periodics.yaml
+++ b/ci-operator/jobs/openshift/release/openshift-release-master-periodics.yaml
@@ -40002,6 +40002,82 @@ periodics:
       secret:
         secretName: result-aggregator
 - agent: kubernetes
+  cluster: build04
+  cron: 0 0 * * 1
+  decorate: true
+  decoration_config:
+    skip_cloning: true
+  extra_refs:
+  - base_ref: master
+    org: openshift
+    repo: release
+  labels:
+    ci-operator.openshift.io/cloud: packet-edge
+    ci-operator.openshift.io/cloud-cluster-profile: packet-assisted
+    ci-operator.openshift.io/variant: nightly-4.12
+    ci.openshift.io/generator: prowgen
+    ci.openshift.io/no-builds: "true"
+    job-release: "4.12"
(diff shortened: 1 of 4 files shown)


ci-operator/jobs/openshift/release/openshift-release-master-periodics.yaml
```diff
@@ -40071,42 +40070,33 @@ periodics:
...
-  reporter_config:
-    slack:
-      channel: '#single-node-ci'
-      job_states_to_report:
-      - success
-      - failure
-      - error
-      report_template: '{{ if eq .Status.State "success" }} :green_jenkins_circle:
-        {{ else }} :red_jenkins_circle: {{ end }} Job *{{.Spec.Job}}* ended with *{{.Status.State}}*.
-        <{{.Status.URL}}|View logs>'
```
osherdp, line 40109: this one needs to come back (``make update/jobs/etc`` sometimes "forgets" about keeping those configurations)

osherdp (changes requested):
you probably need to change this as well: https://github.com/openshift/release/blob/1caf2ed370a5e2be34a25be601fcf18371b0c129/core-services/release-controller/_releases/release-ocp-4.12.json#L255

osherdp (approved):

osherdp:
@openshift/openshift-release-oversight can you PTAL?

ci-operator/jobs/openshift/release/openshift-release-master-periodics.yaml
```diff
@@ -40002,7 +40002,7 @@ periodics:
       secret:
         secretName: result-aggregator
 - agent: kubernetes
-  cluster: build05
+  cluster: build04
```
dgoodwin, line 40005: I was under the impression metal jobs have to route to build05, is this no longer the case, or not true for metal-assisted jobs?

osherdp: I think it's true (or was true, not sure) for ipi-metal jobs. It allows usage of BM machines from internal labs

For assisted-based jobs we're only using Equinix machines

PR #33490, opened 2022-10-26
hypershift: make roks and iks jobs run on api changes and periodically

sjenning:
@relyt0925 @hasueki @enxebre @jparrill 

Currently roks and iks tests are optional presubmits that run on every PR.  However, each test is served by one underlying cluster that does not support parallel tests.  This results in many failures in the form `could not run steps: step e2e-ibmcloud-iks failed: "e2e-ibmcloud-iks" pre steps failed: "e2e-ibmcloud-iks" pod "e2e-ibmcloud-iks-cluster-claim-check" failed`.  As a result, the signal on these tests is poor and tests with poor signal become ignored.

This PR makes them run on PRs only if there is an API change and adds a once-a-day periodic job with a slack notification if it fails.

### DIFF
--- a/ci-operator/config/openshift/hypershift/openshift-hypershift-main.yaml
+++ b/ci-operator/config/openshift/hypershift/openshift-hypershift-main.yaml
@@ -135,7 +135,7 @@ tests:
     workflow: assisted-baremetal
 - as: e2e-ibmcloud-iks
   optional: true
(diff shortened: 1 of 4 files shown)


hasueki (approved):

jparrill (commented):
/lgtm

PR #33509, opened 2022-10-27
add workflow for vsphere-ipi-disk-encryption-tang

jinyunma:
1. add workflow for vsphere-ipi-disk-encryption-tang
2. use butane to generate tang manifest file

### DIFF
--- a/ci-operator/step-registry/ipi/conf/tang/ipi-conf-tang-ref.yaml
+++ b/ci-operator/step-registry/ipi/conf/tang/ipi-conf-tang-ref.yaml
@@ -1,0 +1,15 @@
+ref:
+  as: ipi-conf-tang
+  from_image:
+    namespace: ocp
+    name: "4.12"
+    tag: upi-installer
+  grace_period: 10m
+  commands: ipi-conf-tang-commands.sh
+  resources:
+    requests:
+      cpu: 10m
+      memory: 100Mi
+  documentation: |-
+    Generate manifest file for tang disk encryption.
+    Tang server is set up on dynmanic bastion host.
--- a/ci-operator/config/openshift/verification-tests/openshift-verification-tests-master__installer-rehearse-4.12.yaml
+++ b/ci-operator/config/openshift/verification-tests/openshift-verification-tests-master__installer-rehearse-4.12.yaml
@@ -75,7 +75,7 @@ tests:
   interval: 960h
   steps:
     cluster_profile: vsphere
-    workflow: cucushift-installer-rehearse-vsphere-ipi-ovn-etcd-encryption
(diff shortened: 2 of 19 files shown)


jinyunma:
@jianlinliu @gpei PTAL, thanks.

ci-operator/step-registry/vsphere/provision/bastionhost/tang/vsphere-provision-bastionhost-tang-ref.metadata.json
```diff
@@ -0,0 +1,12 @@
+{
+	"path": "vsphere/provision/bastionhost/tang/vsphere-provision-bastionhost-tang-ref.yaml",
```
jianlinliu, line 2: Most of implementation in the step is to generate manifest files, so maybe we  can rename it to `ipi-conf-tang`, also we can share the step with other platform in the future.

jinyunma: updated

ci-operator/step-registry/ipi/conf/tang/ipi-conf-tang-ref.yaml
```diff
@@ -0,0 +1,12 @@
+ref:
+  as: ipi-conf-tang
+  from: upi-installer
```
jianlinliu, line 3: Maybe we can replace the image with ` from: cli`

jinyunma: In step `ipi-conf-tang`, require `jq` command, but it is not installed in cli image, so still needs `upi-installer` image here.

jianlinliu: Maybe we can use this one:
```
  from_image:
    namespace: ocp
    name: cli-jq
    tag: latest
```

jinyunma: ah, let me update to use `cli-jq`

jinyunma: get new error, no ssh command in image "cli-jq".
` /bin/bash: line 40: ssh: command not found `

jianlinliu: Per the error - `/bin/bash: line 40: ssh: command not found`, seem like cli-jq does not install ssh client, let us use:
```
  from_image:
    namespace: ocp
    name: "4.12"
    tag: upi-install
```
to replace `from: upi-installer` which has dependency on job definition.

PR #33510, opened 2022-10-27
Use oc adm release extract to get oc instead of curl

shellyyang1989:
Prior to the change, 
1) when the [server](https://openshift-release-artifacts.apps.ci.l2s4.p1.openshiftapps.com/) told us `Extracting tools for ..., may take up to a minute`, we waited for it till the step timed out. 
2) when oc extracted failed we exited directly.

The change uses oc adm release extract cli and retries 5 times when oc extracting fails.

### DIFF
--- a/ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
+++ b/ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
@@ -147,35 +147,15 @@ function rhel_upgrade(){
 
 # Extract oc binary which is supposed to be identical with target release
 function extract_oc(){
-    local route="openshift-release-artifacts"
-    local url
-    local target_arch=""
-    if [ "${OCP_ARCH}" != "amd64" ]; then
-      # Change the route to use for non-amd64 clusters
-      route="${route}-${OCP_ARCH}"
-      if [ "$(uname -m)" = "x86_64" ]; then
-        # Add amd64- to the download url only if we run on amd64 prow clusters
-        # For instance, on an arm64 payload:
-        # - openshift-client-linux-4.xxx.tar.gz is the arm64 client
-        # - openshift-client-linux-amd64-4.xxx.tar.gz is the amd64 client
-        target_arch="amd64-"
-      fi
-    fi
-    url="${route}.apps.ci.l2s4.p1.openshiftapps.com"
-
-    #check tools are ready for download
-    local progress; progress="."
-    while curl -kLs https://${url}/${TARGET_VERSION} | grep -q "Extracting tools"
+    echo -e "Extracting oc\n"
+    local retry=5
+    while ! (oc adm release extract -a "${CLUSTER_PROFILE_DIR}/pull-secret" --command=oc --to=${OC_DIR} ${TARGET});
     do
-        echo -ne "Tools have not yet extracted at the server, please wait ${progress}\r"
-        progress+="."
-        sleep 300
+        echo >&2 "Failed to extract oc binary, retry..."
+        (( retry -= 1 ))
+        if (( retry < 0 )); then return 1; fi
+        sleep 60
     done
-
-    echo -e "downloading oc ${TARGET_VERSION} from server ${url}"
-    if ! (curl -kL\# https://${url}/${TARGET_VERSION}/openshift-client-linux-${target_arch}${TARGET_VERSION}.tar.gz | tar -C ${OC_DIR} -xvz); then
-        echo >&2 "Failed to extract oc binary" && return 1
-    fi
     which oc
     oc version --client
     return 0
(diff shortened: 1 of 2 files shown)


shellyyang1989:
The upgrade step got passed.
INFO[2022-10-27T05:40:52Z] Step aws-ipi-byo-iam-role-p2-cucushift-upgrade-toimage succeeded after 1h8m30s. 

@jiajliu @jianlinliu @liangxia Could you please help review when you have time? Thanks!

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -163,19 +163,24 @@ function extract_oc(){
...
+        if (( retry < 0 )); then return 1; fi
         sleep 300
     done
 
+    retry=5
     echo -e "downloading oc ${TARGET_VERSION} from server ${url}"
-    if ! (curl -kL\# https://${url}/${TARGET_VERSION}/openshift-client-linux-${target_arch}${TARGET_VERSION}.tar.gz | tar -C ${OC_DIR} -xvz); then
-        echo >&2 "Failed to extract oc binary" && return 1
-    fi
+    while ! (curl -kL\# https://${url}/${TARGET_VERSION}/openshift-client-linux-${target_arch}${TARGET_VERSION}.tar.gz | tar -C ${OC_DIR} -xvz); do
```
jianlinliu, line 178: It is possible to use `oc adm release extract` command to get the oc cli instead of `curl`, per my experience, seem like oc extract command is more stable than curl.

liangxia: BTW, curl has the option `--retry` if curl is the only choice.

shellyyang1989: Right. I do find the openshift-release-artifacts.apps.ci.l2s4.p1.openshiftapps.com server is not that stable. I'm currently using ansible image for the upgrade step to have the ansible tool. If i'm looking at the correct [Dockerfile](https://github.com/openshift/openshift-ansible/blob/master/images/installer/Dockerfile), seems like it doesn't have oc. That means if we use ansible, we have to use curl to get oc for the first time we use it. Alternative, we can change to cli image but install ansible in the upgrade step. What's your preference?

jianlinliu: > I do find the openshift-release-artifacts.apps.ci.l2s4.p1.openshiftapps.com server is not that stable.

Do you mean `oc adm release extract` is not stable?

shellyyang1989: ah no. I meant the server we curl from. It bothers other teams as well. https://issues.redhat.com/browse/OCPCRT-157

liangxia: Correct, this is the way to inject oc binary. Also it's the reason why we do not have oc installed by default.

shellyyang1989: aha, let me try! Thanks.

shellyyang1989:
The [upgrade step](https://qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/qe-private-deck/pr-logs/pull/openshift_release/33510/rehearse-33510-periodic-ci-openshift-openshift-tests-private-release-4.12-amd64-nightly-4.12-upgrade-from-stable-4.11-aws-ipi-disconnected-private-p1/1586999712890753024) gets passed.

INFO[2022-10-31T13:13:23Z] Step aws-ipi-disconnected-private-p1-cucushift-upgrade-toimage succeeded after 1h13m40s. 

It's ready for your review. Thanks! @liangxia @jianlinliu 
CC @aleskandro for review from ARM perspective.

jianlinliu:
looks good to me. 

Leave this to @aleskandro for arm part.

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -147,35 +147,15 @@ function rhel_upgrade(){
 
 # Extract oc binary which is supposed to be identical with target release
 function extract_oc(){
-    local route="openshift-release-artifacts"
-    local url
-    local target_arch=""
-    if [ "${OCP_ARCH}" != "amd64" ]; then
```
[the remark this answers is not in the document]
shellyyang1989: ah, right. we don't need it in the step. Let me remove it. Thanks!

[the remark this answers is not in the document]
aleskandro: Thanks!

aleskandro (approved):

PR #33514, opened 2022-10-27
bookstack: PoC

psalajova:
Bookstack PoC.

for https://issues.redhat.com/browse/DPTP-3080
/cc smg247

### DIFF
--- a/clusters/app.ci/bookstack/bookstack.yaml
+++ b/clusters/app.ci/bookstack/bookstack.yaml
@@ -99,8 +99,8 @@ items:
     selector:
       app: bookstack
     ports:
-    - port: 80
-      targetPort: 80
+    - port: 8080
+      targetPort: 8080
       protocol: TCP
     type: LoadBalancer
 - apiVersion: apps/v1
@@ -126,7 +126,7 @@ items:
         - name: bookstack
           image: docker.io/solidnerd/bookstack:22.11.0
           ports:
-          - containerPort: 80
+          - containerPort: 8080
             name: http
           env:
           - name: DB_DATABASE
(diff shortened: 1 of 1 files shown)


clusters/app.ci/bookstack/bookstack.yaml
```diff
@@ -0,0 +1,222 @@
...
+        - name: storage-clockwork
+          emptyDir:
+            sizeLimit: 100Mi
+        - name: dot-env
+          secret:
+            secretName: dot-env
+- apiVersion: v1
+  kind: Secret
+  metadata:
+    name: dot-env
```
smg247, line 161: This `dot-env` name is too generic. This secret will live in the `ci` namespace with hundreds (thousands?) of other, non-related, secrets so it needs to be more specific.

clusters/app.ci/bookstack/bookstack.yaml
```diff
@@ -0,0 +1,222 @@
...
+      APP_DEBUG=${APP_DEBUG:-false}
+      APP_KEY=${APP_KEY:-SomeRandomStringWith32Characters}
+      # The below url has to be set if using social auth options
+      # or if you are not using BookStack at the root path of your domain.
+      APP_URL=${APP_URL:-null}
+      # Database details
+      DB_HOST=${DB_HOST:-localhost}
+      DB_DATABASE=${DB_DATABASE:-bookstack}
+      DB_USERNAME=${DB_USERNAME:-bookstack}
+      DB_PASSWORD=${DB_PASSWORD:-password}
```
smg247, line 176: We can't store stuff like this in the git repo. We will have to create a secret in Vault, and then mirror it to `app.ci` via `ci-secret-bootstrap`. I think the content of this can stay the same in Vault, however. I will add this to Vault myself, as you won't have access, and then I will show you how to achieve the mirroring.

PR #33524, opened 2022-10-27
Upgrade rosa-sts job to 4.12 and update the job priority

yasun1:
4.12 nightly build is enabled on rosa, let us upgrade the rosa-sts job via removing the 4.11 job and adding the 4.12 job.

Besides, update the jobs's priority to p2 for they should be weekly testing.


yasun1:
@liangxia Please help review. Thanks.

PR #33526, opened 2022-10-27
pao: move master jobs to 4.12 and remove tests

yanirq:
move master jobs to point to 4.12.
created 4.11 jobs for the new 4.11 pao upstream branch. removed from master the following tests and related images: e2e-gcp (for pao container and index images tests) e2e-gcp-operator-upgrade
ci-ppc
e2e-render-cmd

all of the above are under cluster-node-tuning-operator.

Signed-off-by: Yanir Quinn <yquinn@redhat.com>


fromanirh:
I think we should do the move and the removal in two separate PRs. First the removal, then the move (but I don't really mind too much about the actual order)

yanirq:
> I think we should do the move and the removal in two separate PRs. First the removal, then the move (but I don't really mind too much about the actual order)

The removal is only from 4.12
The idea behind removing the test lanes in 4.12 (master) is to avoid bumping the version for these lanes since they are not really used and do the bump only for the relevant components. 
4.11 was left intact and will not be cleared from the test lanes so we will always have a reference point

fromanirh:
> > I think we should do the move and the removal in two separate PRs. First the removal, then the move (but I don't really mind too much about the actual order)
> 
> The removal is only from 4.12 The idea behind removing the test lanes in 4.12 (master) is to avoid bumping the version for these lanes since they are not really used and do the bump only for the relevant components. 4.11 was left intact and will not be cleared from the test lanes so we will always have a reference point

ok, then we can perhaps address this by moving master and then removing the lanes?
The lanes have been red for quite some time, one more overridden PR won't hurt.

Anyway I'm not really blocking for this, the exact same reasoning (we had red lanes for quite some time) can also lead us to merge the two things together

ci-operator/jobs/openshift-kni/performance-addon-operators/openshift-kni-performance-addon-operators-release-4.11-postsubmits.yaml
```diff
@@ -13,7 +13,7 @@ postsubmits:
       ci.openshift.io/generator: prowgen
       job-release: "4.11"
     max_concurrency: 1
-    name: branch-ci-openshift-kni-performance-addon-operators-master-images
+    name: branch-ci-openshift-kni-performance-addon-operators-release-4.11-images
```
Tal-or, line 16: master should be renamed 4.12 isn't it?

yanirq: this is the 4.11 postsubmits job. there is another job for master

MarSik: And the filename as well.

yanirq: This is an outcome of creating a 4.11 copy : https://github.com/openshift/release/pull/33526/files#diff-845d53be408e2b5ce81c7d5f2fa10af36677706373f9a2e68d54607a66c42350  from the master one.
Since we branched we should now have a 4.11 job added in addition to the master job

MarSik:
We still keep the must gather lane, right?

yanirq:
> We still keep the must gather lane, right?

@MarSik  there is no specific test lane for must gather. we will need to add it once its completed under: https://github.com/openshift-kni/performance-addon-operators/pull/943 
which should also be done under the 4.11 and master branch for PAO repository (until we move must gather from there)
cc @marioferh

fromanirh:
/hold
feel free to unhold when ready

marioferh (commented):
/lgtm

PR #33552, opened 2022-10-27
AWS Load Balancer Controller: Update build root from repository

thejasn:


Miciah:
I think this should be fine now that https://github.com/openshift/aws-load-balancer-controller/pull/11 has merged.
/retest-required
/lgtm

alebedev87:
This change has been merged as part of https://github.com/openshift/release/pull/34323.

Miciah:
> This change has been merged as part of #34323.

So it looks like this PR can be closed.
/close

PR #33555, opened 2022-10-27
stackrox: run nightly tests at 8am

gavin-stackrox:
In the I don't :heart: OpenShift CI department: the build cluster that runs nightly tests was changed recently: https://github.com/openshift/release/pull/33373. This (I believe) is the root of why we started to see nightly tests fail due to pod scheduling (https://issues.redhat.com/browse/ROX-13279). My theory is that build02 is slower to scale up than the previous cluster build03. I heard somewhere that OpenShift CI is aggressive at scaling down clusters. And at 5am UTC it has probably been scaled to 😴.

It has been suggested that setting limits to our job definitions might resolve this but I think that would result in other flakes particularly w.r.t. memory. And would not resolve the issue because `build02` is just darned slow. And not to mention would be a major pain given how we use the step registry as a 'macro' for test definitions.

I'm going to try to find further evidence to back this up but as a short term attempt at a fix lets change the time the nightlies run so they are more likely to coincide with an active cluster ☕ yet still ready for a post lunch review in EU. Which undermines the entire notion of running nightly tests when CI is idle but 🤷.

/hold
/cc @porridge @house-d

### DIFF
--- a/ci-operator/config/stackrox/stackrox/stackrox-stackrox-nightlies.yaml
+++ b/ci-operator/config/stackrox/stackrox/stackrox-stackrox-nightlies.yaml
@@ -427,7 +427,7 @@ tests:
     workflow: stackrox-automation-flavors-rosa-e2e
   timeout: 5h0m0s
 - as: nightly-tag
-  cron: 0 5 * * *
+  cron: 0 8 * * *
   steps:
     test:
     - as: tag
--- a/ci-operator/jobs/stackrox/stackrox/stackrox-stackrox-nightlies-periodics.yaml
+++ b/ci-operator/jobs/stackrox/stackrox/stackrox-stackrox-nightlies-periodics.yaml
@@ -1,7 +1,7 @@
 periodics:
 - agent: kubernetes
   cluster: build05
-  cron: 0 5 * * *
+  cron: 0 8 * * *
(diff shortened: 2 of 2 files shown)


porridge:
> change the time the nightlies run so they are more likely to coincide with an active cluster

I'm not sure I follow the logic. WDYM by "active cluster"? If there is more activity then we're even less likely to be able to schedule, not more, right?
Also, I'm typically not available for triaging in the afternoon :disappointed: so effectively my triaging might lag by a day.

gavin-stackrox:
> I'm not sure I follow the logic. WDYM by "active cluster"?
Inactive == the cluster has scaled down as many nodes as it can because no jobs have scheduled in a while

porridge:
> > WDYM by "active cluster"?
> 
> Inactive == the cluster has scaled down as many nodes as it can because no jobs have scheduled in a while Active == jobs are scheduled so the cluster machine autoscaler has added more nodes and can handle new scheduling reqs better

Perhaps we should be trying to run these earlier, then, to catch the moment when there are already few jobs, but still plenty of nodes :speak_no_evil:

gavin-stackrox:
We got kicked to another cluster, so I'm closing this.

PR #33566, opened 2022-10-28
node-observability-agent: re-enable linting

alebedev87:


alebedev87:
/retest

Bot, where are you?

PR #33572, opened 2022-10-28
remove HYPERSHIFT_BASE_DOMAIN and aws-credentials

LiangquanLi930:
`HYPERSHIFT_BASE_DOMAIN` are not used for `hypershift-kubevirt-destroy`, so I think we should remove


LiangquanLi930:
@csrwng hi, could you help review this PR, thanks

PR #33575, opened 2022-10-28
update image mirroring

jmguzik:


jmguzik:
@jupierce @bparees we would need your approval. It's the last branch-cutting step.

PR #33582, opened 2022-10-29
append --zone parameter to creating vm command when the region has AZ

gpei:
To fix https://issues.redhat.com/browse/OCPQE-11854

In current RHEL vm provision process, no "--zone" was set to the command so RHEL machines are created without any Availability zone specified, which will lead to the RHEL node not having the following expected labels added. 

A RHEL node without Availability zone set
`failure-domain.beta.kubernetes.io/zone=0,topology.disk.csi.azure.com/zone=,topology.kubernetes.io/zone=0`

A RHCOS node in Availability zone 3
`failure-domain.beta.kubernetes.io/zone=eastus-3,topology.disk.csi.azure.com/zone=eastus-3,topology.kubernetes.io/zone=eastus-3`

Then when pods running on RHCOS nodes which were using PVC created in certain zone, they couldn't be re-scheduled because no RHEL worker satisfied the nodeSelector rule, this will cause RHCOS nodes couldn't be drained, then the workflow failed. 

Tried manually to create a RHEL vm "gpei-1029-6cvgl-rhel2" located in zone "3", after joining to the cluster, it's labeled with the expected 
`failure-domain.beta.kubernetes.io/zone=eastus-3,topology.disk.csi.azure.com/zone=eastus-3,topology.kubernetes.io/zone=eastus-3`
```
  gpei-1029-6cvgl-rhel2                  Ready    worker                 7m53s   v1.25.2+1f53707   beta.kubernetes.io/arch=amd64,beta.kubernetes.io/instance-type=Standard_D4s_v3,beta.kubernetes.io/os=linux,failure-domain.beta.kubernetes.io/region=eastus,failure-domain.beta.kubernetes.io/zone=eastus-3,kubernetes.io/arch=amd64,kubernetes.io/hostname=gpei-1029-6cvgl-rhel2,kubernetes.io/os=linux,node-role.kubernetes.io/worker=,node.kubernetes.io/instance-type=Standard_D4s_v3,node.openshift.io/os_id=rhel,topology.disk.csi.azure.com/zone=eastus-3,topology.kubernetes.io/region=eastus,topology.kubernetes.io/zone=eastus-3
  gpei-1029-6cvgl-worker-eastus3-2th8b   Ready    worker                 79m     v1.25.2+4bd0702   beta.kubernetes.io/arch=amd64,beta.kubernetes.io/instance-type=Standard_D4s_v3,beta.kubernetes.io/os=linux,failure-domain.beta.kubernetes.io/region=eastus,failure-domain.beta.kubernetes.io/zone=eastus-3,kubernetes.io/arch=amd64,kubernetes.io/hostname=gpei-1029-6cvgl-worker-eastus3-2th8b,kubernetes.io/os=linux,machine.openshift.io/interruptible-instance=,node-role.kubernetes.io/worker=,node.kubernetes.io/instance-type=Standard_D4s_v3,node.openshift.io/os_id=rhcos,topology.disk.csi.azure.com/zone=eastus-3,topology.kubernetes.io/region=eastus,topology.kubernetes.io/zone=eastus-3
```

For now, we have Azure jobs running in 
```
        'northcentralus': 10,
        'westus3': 10,
        'centralus': 10
```
And northcentralus doesn't have AZ for size "Standard_D4s_v3", while westus3 and centralus has zone[1,2,3] available. 
```
  [root@preserve-gpei-worker test_place]# for i in northcentralus westus3 centralus; do az vm list-skus -l $i --zone --size ${RHEL_VM_SIZE} --query "[].locationInfo[0].zones[]" -o TLSv1; done
  []
  [
    "2",
    "3",
    "1"
  ]
  [
    "2",
    "1",
    "3"
  ]
```
So also updated the default RHEL_WORKER_COUNT to 3 to make sure RHEL machines are created in each zone.

### DIFF
--- a/ci-operator/step-registry/workers-rhel/azure-provision/workers-rhel-azure-provision-commands.sh
+++ b/ci-operator/step-registry/workers-rhel/azure-provision/workers-rhel-azure-provision-commands.sh
@@ -40,18 +40,21 @@ computeSubnetID=$(az network vnet subnet show --resource-group ${vnet_RG} --vnet
 # Start to provision rhel instances from template in existing VNET and NSG
 for count in $(seq 1 ${RHEL_WORKER_COUNT}); do
   echo "$(date -u --rfc-3339=seconds) - Provision ${infra_id}-rhel-${count} ..."
-  # az command to configure RHEL VM's
-  az vm create --resource-group "${infra_id}-rg" \
-    --name "${infra_id}-rhel-${count}" \
-    --image "${RHEL_IMAGE}" \
-    --ssh-key-values "${SSH_PUB_KEY_PATH}" \
-    --admin-user "${RHEL_USER}" \
-    --public-ip-address "" \
-    --size "${RHEL_VM_SIZE}" \
-    --os-disk-size-gb "${RHEL_VM_DISK_SIZE}" \
-    --nsg "" \
-    --subnet "${computeSubnetID}" > /tmp/tmp.json
-  
+
+  # check if the current region support AZ by checking if $az_num != 0
+  IFS=$'\t' read -r -a zone_list <<< "$(az vm list-skus -l ${AZURE_REGION} --zone --size ${RHEL_VM_SIZE} --query '[].locationInfo[0].zones' -o tsv)"
+  az_num=${#zone_list[@]}
+
+  # az command to create RHEL VM and append --zone when the region has AZ
+  cmd="az vm create --resource-group '${infra_id}-rg' --name '${infra_id}-rhel-${count}' --image '${RHEL_IMAGE}' --ssh-key-values '${SSH_PUB_KEY_PATH}' --admin-user '${RHEL_USER}' --public-ip-address '' --size '${RHEL_VM_SIZE}' --os-disk-size-gb '${RHEL_VM_DISK_SIZE}' --nsg '' --subnet '${computeSubnetID}'"
+
+  if [ "${az_num}" != "0" ]; then
+    cmd="${cmd} --zone ${count}"
+  fi
+
+  echo "Creating RHEL VM: ${cmd}"
+  eval "${cmd}" | tee /tmp/tmp.json 
+ 
   rhel_node_ip=$(cat /tmp/tmp.json | jq -r .privateIpAddress)
   echo "Ip address is ${rhel_node_ip}"
   
--- a/ci-operator/step-registry/workers-rhel/azure-provision/OWNERS
+++ b/ci-operator/step-registry/workers-rhel/azure-provision/OWNERS
@@ -1,12 +1,12 @@
 approvers:
 - jianlinliu
-- Amoghrd
+- jinyunma
 - gpei
 - mgahagan73
 - MayXuQQ
 reviewers:
 - jianlinliu
-- Amoghrd
+- jinyunma
 - gpei
 - mgahagan73
 - MayXuQQ
--- a/ci-operator/step-registry/workers-rhel/azure-provision/workers-rhel-azure-provision-ref.metadata.json
+++ b/ci-operator/step-registry/workers-rhel/azure-provision/workers-rhel-azure-provision-ref.metadata.json
@@ -3,14 +3,14 @@
 	"owners": {
 		"approvers": [
 			"jianlinliu",
-			"Amoghrd",
+			"jinyunma",
 			"gpei",
 			"mgahagan73",
 			"MayXuQQ"
 		],
 		"reviewers": [
 			"jianlinliu",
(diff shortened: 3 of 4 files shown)


MayXuQQ:
/lgtm
shall we replace the region northcentralus ?

gpei:
Also cc @jianlinliu @jinyunma for review.

gpei:
> /lgtm shall we replace the region northcentralus ?

I think it's OK to leave it there, if the job is running in northcentralus then no AZ would be used for RHCOS or RHEL nodes.

ci-operator/step-registry/workers-rhel/azure-provision/workers-rhel-azure-provision-commands.sh
```diff
@@ -40,18 +40,21 @@ computeSubnetID=$(az network vnet subnet show --resource-group ${vnet_RG} --vnet
...
-    --admin-user "${RHEL_USER}" \
-    --public-ip-address "" \
-    --size "${RHEL_VM_SIZE}" \
-    --os-disk-size-gb "${RHEL_VM_DISK_SIZE}" \
-    --nsg "" \
-    --subnet "${computeSubnetID}" > /tmp/tmp.json
-  
+
+  # check if the current region support AZ by checking if $az_num != 0
+  IFS=$'\t' read -r -a zone_list <<< "$(az vm list-skus -l ${AZURE_REGION} --zone --size ${RHEL_VM_SIZE} --query '[].locationInfo[0].zones' -o tsv)"
```
jianlinliu, line 45: I guess we can move this line out of the `for` loop, run it once.

ci-operator/step-registry/workers-rhel/azure-provision/workers-rhel-azure-provision-commands.sh
```diff
@@ -40,18 +40,21 @@ computeSubnetID=$(az network vnet subnet show --resource-group ${vnet_RG} --vnet
...
+  az_num=${#zone_list[@]}
+
+  # az command to create RHEL VM and append --zone when the region has AZ
+  cmd="az vm create --resource-group '${infra_id}-rg' --name '${infra_id}-rhel-${count}' --image '${RHEL_IMAGE}' --ssh-key-values '${SSH_PUB_KEY_PATH}' --admin-user '${RHEL_USER}' --public-ip-address '' --size '${RHEL_VM_SIZE}' --os-disk-size-gb '${RHEL_VM_DISK_SIZE}' --nsg '' --subnet '${computeSubnetID}'"
+
+  if [ "${az_num}" != "0" ]; then
+    cmd="${cmd} --zone ${count}"
```
jianlinliu, line 52: Does that means all the rhel workers always are created in the last zone?

jianlinliu:
Thanks for the deep analysis for the root cause, I wonder, does it also happen on our jenkins jobs, right?

ci-operator/step-registry/workers-rhel/azure-provision/workers-rhel-azure-provision-commands.sh
gpei: the ${count} comes from the `for` loop
`for count in $(seq 1 ${RHEL_WORKER_COUNT}); `
so that the RHEL-1 is in zone 1, RHEL-2 in zone-2, RHEL-3 in zone-3,
to make sure each zone has one RHEL node.

gpei:
> Thanks for the deep analysis for the root cause, I wonder, does it also happen on our jenkins jobs, right?

We don't have RHEL worker only profile for Azure in Jenkins. 
I had a check in 4.12 CI test history, the following two are created for Azure, both of them are created with RHCOS and RHEL.
```
03_Disconnected UPI on Azure with RHCOS & RHEL8.6 & FIPS on & http_proxy & Etcd Encryption on
97_Disconnected IPI on Azure & techpreview & RHCOS & RHEL8.6 & fully Private Cluster
```
This should only fail when all the RHCOS nodes got removed.

ci-operator/step-registry/workers-rhel/azure-provision/workers-rhel-azure-provision-commands.sh
jianlinliu: Ah, got it. So once `RHEL_WORKER_COUNT` is set to >=4, maybe we would hit some issues, so maybe we can:
```
cmd="${cmd} --zone $((${count} % ${az_num} + 1))"
```

gpei: yeah, good point. added it in https://github.com/openshift/release/pull/33623

PR #33586, opened 2022-10-30
Azure test

LiangquanLi930:


kuiwang02:
the updated file is not only under openshift-tests-private, so I can add lgtm, not say "iamok"
please correct conflicts.
from my view, 
/lgtm

PR #33587, opened 2022-10-30
mihawk1 needs firmware update, remove it temporarily

vijay-vin:
Signed-off-by: Vijay Vinnakota <vijay.vinnakota@ibm.com>


vijay-vin:
@deepsm007 
This is the PR to temporarily remove one of the Mihawk system (mihawk1) from ppc64le libvirt quota list. As discussed before, once the firmware is updated I will put this mihawk1 system back into the quota list.

PR #33594, opened 2022-10-31
OCPQE-12669: Add ginkgo upgrade to openshift-upgrade-qe-test-paused-to-latest

liangxia:


shellyyang1989:
lgtm. BTW, the chain covers the stable to stable eus testing by using the `--to-latest` option for upgrade. The stable to nightly eus [chain](https://github.com/openshift/release/blob/master/ci-operator/step-registry/cucushift/upgrade/paused/cucushift-upgrade-paused-workflow.yaml) seems to require the change as well.

liangxia:
> lgtm. BTW, the chain covers the stable to stable eus testing by using the `--to-latest` option for upgrade. The stable to nightly eus [chain](https://github.com/openshift/release/blob/master/ci-operator/step-registry/cucushift/upgrade/paused/cucushift-upgrade-paused-workflow.yaml) seems to require the change as well.

That's a different one and need extra effort.  Will use a follow PR for that.

PR #33596, opened 2022-10-31
e2e-nutanix-assisted: re-add disabled-host-validations

vrutkovs:
Required for installation to complete for now. Seems it was lost after some rebases


eliorerz (approved):
LGTM

PR #33617, opened 2022-10-31
OSSM-2216: adds maistra/api jobs for 2.4

bartoszmajsak:


jwendell:
Let's fix this issue in https://issues.redhat.com/browse/OSSM-2223
After that we can retest

jwendell (approved):
Let's merge this and get it fixed in the api repo.

PR #33618, opened 2022-10-31
stackrox: set limits for some nightly tests

gavin-stackrox:
It has been suggested that setting pod limits might alleviate some of the pod scheduling issues we experience with nightly test failures:
- https://issues.redhat.com/browse/ROX-13279
- https://coreos.slack.com/archives/CBN38N3MW/p1666889370141609

I'm not convinced *but* lets give it a try. Also the nightly build tests have switched away from the problematic `build02` cluster with https://github.com/openshift/release/pull/33597.

This PR sets limits for the 3 repeat job failures.

## Testing

None. We'll see how it fares in nightly runs.

/cc @house-d @porridge 
/hold


sbostick (approved):
LGTM

PR #33630, opened 2022-11-01
[WIP]Add upgrade path 4.3 to 4.12

JianLi-RH:
Please ignore this PR

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-nightly-4.12-upgrade-from-stable-4.3.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-nightly-4.12-upgrade-from-stable-4.3.yaml
@@ -1,0 +1,117 @@
+base_images:
+  ansible:
+    name: "4.3"
+    namespace: ocp
+    tag: ansible
+  cli:
+    name: "4.12"
+    namespace: ocp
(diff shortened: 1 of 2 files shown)


juzhao:
see from slack thread
https://coreos.slack.com/archives/G79AW9Q7R/p1667391466066799

CI jobs and cluster-bot clusters would attach PVs for prometheus by default, in this case, we will see the error
```
level=error msg="Cluster operator monitoring Degraded is True with UpdatingPrometheusK8SFailed: Failed to rollout the stack. Error: running task Updating Prometheus-k8s failed: reconciling Prometheus object failed: creating Prometheus object failed: Prometheus.monitoring.coreos.com \"k8s\" is invalid: spec.storage.volumeClaimTemplate.metadata.creationTimestamp: Invalid value: \"null\": spec.storage.volumeClaimTemplate.metadata.creationTimestamp in body must be of type string: \"null\""
level=fatal msg="failed to initialize the cluster: Cluster operator monitoring is still updating"  
```
but a cluster launched by jenkins don't attach PVs for prometheus by default, if we bind PVs for prometheus after the cluster is ready, will not meet the error

juzhao:
@JianLi-RH 
see from https://coreos.slack.com/archives/G79AW9Q7R/p1667530416192859
since 4.3 is EOF, we can not merge https://github.com/openshift/cluster-monitoring-operator/pull/1810 to 4.3, remove `ipi-install-monitoringpvc` step seems the right solution

PR #33634, opened 2022-11-01
Migrate tests from verification-tests to openshift-tests-private

liangxia:
Migrate tests from verification-tests to openshift-tests-private, so we have a centralized place for the e2e/upgrade tests. It makes the integration with ReportPortal easier.

/cc @pruan-rht @dis016 @jhou1 @JianLi-RH


pruan-rht:
@liangxia this looks like it might affect how junit results are uploaded to ReportPortal.

liangxia:
> @liangxia this looks like it might affect how junit results are uploaded to ReportPortal.

Those tests in folder verification-tests are not processed previously, thus are not uploaded to RP at the moment. 
This is one of the effort that trying to get them uploaded to RP

PR #33635, opened 2022-11-01
Add HyperShift workflow for DPTP

bear-redhat:


ci-operator/step-registry/hypershift/dump/hypershift-dump-chain.yaml
```diff
@@ -5,6 +5,12 @@ chain:
     cli: latest
     commands: |-
       set -exuo pipefail
+
+      if [[ -f ${SHARED_DIR}/hive_kubeconfig ]]; then
+        echo "Using existing hive kubeconfig"
+        export KUBECONFIG=${SHARED_DIR}/hive_kubeconfig
+      fi
```
bbguimaraes, line 12: `KUBECONFIG` is always set (it will point to the build cluster if the test does not put one in the shared directory), so the absence here should be an error (or an early exit, depending on what you want).

ci-operator/step-registry/hypershift/hive/OWNERS
```diff
@@ -0,0 +1,5 @@
+approvers:
+- bear-redhat
+options: {}
+reviewers:
+- bear-redhat
```
bbguimaraes, line 5: Not a very good bus factor XD

Add DPTP.

ci-operator/step-registry/hypershift/dump/hypershift-dump-chain.yaml
```diff
@@ -5,6 +5,12 @@ chain:
     cli: latest
     commands: |-
       set -exuo pipefail
```
bbguimaraes, line 7: Remove the `x`s from these scripts.

ci-operator/step-registry/hypershift/hive/create/hypershift-hive-create-chain.yaml
```diff
@@ -0,0 +1,159 @@
...
+      [[ ${CONTROLPLANE_OPERATOR_IMAGE} = "LATEST" ]] && CONTROLPLANE_OPERATOR_IMAGE="${HYPERSHIFT_RELEASE_LATEST}"
+
+      HASH="$(echo -n $PROW_JOB_ID|sha256sum)"
+      CLUSTER_NAME=${HASH:0:20}
+      INFRA_ID=${HASH:20:5}
+      echo "Using cluster name $CLUSTER_NAME and infra id $INFRA_ID"
+      echo "CLUSTER_NAME=$CLUSTER_NAME" > ${SHARED_DIR}/hosted_cluster.txt
+      echo "INFRA_ID=$INFRA_ID" >> ${SHARED_DIR}/hosted_cluster.txt
+
+      if [[ -f ${SHARED_DIR}/pull-secert-build-farm.json ]]; then
```
bbguimaraes, line 62: Typo.

ci-operator/step-registry/hypershift/dump/hypershift-dump-chain.yaml
```diff
@@ -5,6 +5,12 @@ chain:
     cli: latest
     commands: |-
```
bbguimaraes, line 6: Move these scripts out of the chain definitions and into their own steps in the registry.  Then you'll be able to edit them sensibly in their own `*.sh` files and they'll be verified by our `shellcheck` job.

clusters/hive/hypershift.yaml
```diff
@@ -0,0 +1,97 @@
+apiVersion: v1
```
bbguimaraes, line 1: I imagine you'll want to merge this separately so that we can get some rehearsals in this PR?

ci-operator/step-registry/hypershift/hive/workflow/hypershift-hive-workflow-workflow.yaml
```diff
@@ -0,0 +1,26 @@
...
+    OpenShift 4.13.x are not supported and will cause hypershift-aws-create to stall
+
+  steps:
+    post:
+    - chain: hypershift-dump
+    - chain: hypershift-hive-destroy
+    pre:
+    - ref: openshift-cluster-bot-rbac
+    - chain: hypershift-hive-credentials
+    - chain: hypershift-hive-create
```
bbguimaraes, line 23: These files don't need to be in alphabetical order, so you can use the more natural `pre`/`post` order.

bear-redhat:
Let's see if `cluster-admin` is required without `dump` step
/pj-rehearse

ci-operator/step-registry/hypershift/hive/create/hostedcluster/hypershift-hive-create-hostedcluster-commands.sh
```diff
@@ -0,0 +1,105 @@
...
+echo "CLUSTER_NAME=$CLUSTER_NAME" > ${SHARED_DIR}/hosted_cluster.txt
+echo "INFRA_ID=$INFRA_ID" >> ${SHARED_DIR}/hosted_cluster.txt
+
+if [[ -f ${SHARED_DIR}/pull-secret-build-farm.json ]]; then
+  jq -s '.[0] * .[1]' ${SHARED_DIR}/pull-secret-build-farm.json /etc/ci-pull-credentials/.dockerconfigjson > /tmp/pull-secret.json
+else
+  cp /etc/ci-pull-credentials/.dockerconfigjson /tmp/pull-secret.json
+fi
+
+echo "$(date) Creating HyperShift cluster ${CLUSTER_NAME}"
```
jupierce: What prevents more than a certain number of hosted clusters from being created? I understand that 80 is the recommended maximum at the moment. Seems the workflow should hold a lease for a hosted cluster and the account in which they are trying to establish their guest nodes.

ci-operator/step-registry/hypershift/hive/workflow/hypershift-hive-workflow-workflow.yaml
```diff
@@ -0,0 +1,21 @@
...
+  as: hypershift-hive-workflow
+  documentation: |-
+    The HyperShift hive workflow executes tests against a new ephemeral
+    HyperShift cluster. Administrative access to the control plane is provided
+    via the `KUBECONFIG` environment variable.
+
+    Note that tests will execute once the control plane is ready, regardless of
+    whether nodes have been requested, have joined, or are available.
+
+    The control plane is hosted on hive cluster.
```
jupierce:
```suggestion
    This workflow uses a shared HyperShift management cluster run by the Test Platform team.
```

ci-operator/step-registry/hypershift/hive/workflow/hypershift-hive-workflow-workflow.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+    OpenShift versions supported by this workflow are 4.12.x.
```
jupierce: These should be programmatically enforced.  `oc version -o=json` provides the hive cluster's version. `oc adm release info ..` can supply information about the release image specified.  e.g. https://github.com/openshift/release/blob/12b4cdbbfa4ec9bc32b9dd24e080b13b82983cc7/ci-operator/step-registry/ipi/conf/aws/ipi-conf-aws-commands.sh#L221-L223 .

ci-operator/step-registry/hypershift/hive/create/hostedcluster/hypershift-hive-create-hostedcluster-ref.yaml
```diff
@@ -0,0 +1,60 @@
...
+  - name: HYPERSHIFT_AWS_REGION
+    default: "us-east-1"
+    documentation: "The AWS region of the cluster."
+  - name: CONTROLPLANE_OPERATOR_IMAGE
+    default: ""
+    documentation: "The CPO image to use. Set to LATEST to use the latest one"
+  - name: EXTRA_ARGS
+    default: ""
+    documentation: "Extra args to pass to the create cluster aws command"
+  - name: HYPERSHIFT_GUEST_INFRA_OCP_ACCOUNT
```
jupierce: Perhaps?
```suggestion
  - name: USE_SHARED_GUEST_INFRA_OCP_ACCOUNT
```

ci-operator/step-registry/hypershift/hive/create/hostedcluster/hypershift-hive-create-hostedcluster-commands.sh
```diff
@@ -0,0 +1,105 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+AWS_GUEST_INFRA_CREDENTIALS_FILE="/etc/hypershift-ci-jobs-awscreds/credentials"
```
jupierce: Can you add information about how to use this in the workflow documentation?

jupierce (changes requested):

ci-operator/step-registry/hypershift/hive/create/hostedcluster/hypershift-hive-create-hostedcluster-commands.sh
bear-redhat: The workflow requested a lease. The lease is managed by boskos and set to 80 (I confirmed with HypreShift team).

ci-operator/step-registry/hypershift/hive/create/hostedcluster/hypershift-hive-create-hostedcluster-commands.sh
bear-redhat: I actually plan to remove it (copied from hypershift team workflow), since user should always provide a `cluster_profile`

ci-operator/step-registry/hypershift/hive/create/hostedcluster/hypershift-hive-create-hostedcluster-commands.sh
bear-redhat: @bbguimaraes  can you confirm if what I do is correct (request lease at create step) and can fulfill our goal (to limit the hostedcluster to 80)?

bbguimaraes:
LGTM, but I don't see the `dump` chain used anywhere.

bear-redhat:
> LGTM, but I don't see the dump chain used anywhere.

Yes, in deed. `dump` chain requires admin permission to `clusters-XXX` namespaces. While we cannot assign role with regex, we can only assign the SA with `cluster-admin` privilege, which is not safe.
Thus I removed it, but still keep the chain here in case someone do provide a SA with cluster-admin permission or other possible uses.

PR #33636, opened 2022-11-01
hypershift: increase e2e instance size

sjenning:
We are encountering scheduling pressure in the e2e mgmt clusters.

This increases our instance size from `m5.large` to `m5.4xlarge` and decrease the node count from 9 to  3 (1 per zone).  This increases overall capacity by 33% and decreases resource fragmentation where the cluster might cumulatively have enough resources, and any single node does not.  Also `m5.4xlarge` is 16vcpu/64Gi, which is what we are recommending SD use.

Additionally, I changed some periodic start times.  I didn't realize the times were in UTC before.  Trying to make the run during night hours for all team members.

### DIFF
--- a/ci-operator/step-registry/hypershift/aws/setup-nested-management-cluster/hypershift-aws-setup-nested-management-cluster-chain.yaml
+++ b/ci-operator/step-registry/hypershift/aws/setup-nested-management-cluster/hypershift-aws-setup-nested-management-cluster-chain.yaml
@@ -7,6 +7,9 @@ chain:
     - name: HYPERSHIFT_NODE_COUNT
       default: "3"
       documentation: "The number nodes for the cluster. This number will get mulitplied by three, as there are three AZs"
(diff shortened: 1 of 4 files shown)


ci-operator/step-registry/hypershift/aws/setup-nested-management-cluster/hypershift-aws-setup-nested-management-cluster-chain.yaml
```diff
@@ -7,6 +7,9 @@ chain:
     - name: HYPERSHIFT_NODE_COUNT
       default: "3"
       documentation: "The number nodes for the cluster. This number will get mulitplied by three, as there are three AZs"
+    - name: INSTANCE_TYPE
```
enxebre, line 10: it'd be good to prefix with HYPERSHIFT_ to avoid potential clashes with ci built in env vars, e.g. that happened with namespace

PR #33651, opened 2022-11-01
Patch to resolve flakiness in network migration e2e

nicklesimba:
This patch changes the hostsubnet name. It was originally pulled from a hardcoded value, but is now being changed specifically to use any worker node. This will resolve CI flakiness that can occur when an egressIP is assigned on a master node.

Signed-off-by: nicklesimba <simha.nikhil@gmail.com>


jluhrsen:
looks good. let's see how the rehearsals do

dougbtv:
Sorry on the loose trigger. I just figured it was better than the existing. Feel free to release the hold when you're happy.

jluhrsen:
oh, it wont run the jobs we want to run cause it's not touching a job config. @nicklesimba , can you push a temp update to this PR that changes something (the name is fine) of one of your migration jobs. that will trigger a rehearsal on that job with this PR and we can see how it works. after we see that, you can remove that change and just leave it like this for us to approve.

jluhrsen:
I'd change the commit message to just "resolve flakiness in network migration e2e" and then include in the body of the
commit message that this will ensure we always get a worker node.

PR #33653, opened 2022-11-01
DO NOT MERGE: Make the sippy-e2e scripts more conducive to running locally

DennisPeriquet:
**NOTE**: This allows me to test my changes; I don't want to merge just yet because we intend to put these scripts in the sippy repo and call them here.

Re: [TRT-638](https://issues.redhat.com//browse/TRT-638)

* Use environment variables so we can override them for development mode.
* Abstract out the kubectl/oc command so we can use either in development mode.
* Deal with the difference of the `base64` command on Linux (has the `-w0` option) and MacOS (the `-w0` options results in an error)
* Allow different imagePullPolicy to make it easy to use minikube local container registry.


DennisPeriquet:
done w/ this testing.  https://github.com/openshift/release/pull/33694 does what I need it to do.

PR #28149, opened 2022-04-27
[AWS QE] Support installing cluster on C2S/SC2S

yunjiang29:
Support installing cluster on AWS C2S/SC2S

## Steps
1. Get SHIFT project info from vault, informations are stored as following format:
```
{
  "us-iso-east-1": {
    "source_region": "us-east-1",
    "vpc_id": "vpc-02d3ec22d80ec499f",
    "private_subnet": {
      "id": "subnet-0132f5587ef134486",
      "az": "us-iso-east-1c"
    },
    "public_subnet": {
      "id": "subnet-072a0f8c475a84cca",
      "az": "us-iso-east-1b"
    },
    "project_name": "OpenShift-QE-C2SEmu-CI2",
    "vpc_stack_name": "OpenShift-QE-C2SEmu-CI2",
    "machine_network": "10.143.0.0/16",
    "temporary_credential_endpoint": "https://cap.digitalageexperts.com/api/v1/credentials",
    "cross_account_role": "OpenShift-QE-C2SEmu-CI2-CAR2-CrossAccountRole-XCDV79IJ2XKQ",
    "cert": "LS0t...",
    "private_key": "LS0t..."
  },
}
```
2. Setup a bastion host, includes a proxy and a mirror registry server.
3. Setup temporary token service `aws-c2s-sc2s-init-token-service` (apply and refresh credential)
4. Setup instance metadata service `aws-c2s-sc2s-instance-metadata` (a workaround for emulator, see https://bugzilla.redhat.com/show_bug.cgi?id=1923956#c3)
5. Mirror release images
6. Create cluster

full steps can be found https://github.com/openshift/release/blob/9785bb65e019641b41c2387ca31e77d6d53bba98/ci-operator/step-registry/cucushift/installer/rehearse/aws/ipi/disconnected/private/c2s-sc2s/provision/cucushift-installer-rehearse-aws-ipi-disconnected-private-c2s-sc2s-provision-chain.yaml

## Note
There is no internet connections in C2S/SC2S environment, you must **_set proxy_** if you want to work with C2S/SC2S env.
If you want to connect to internet (e.g. download `jq`, `yq`), you must **_unset proxy_** (to _disconnect_ from the C2S/SC2S environment )

PR #33654, opened 2022-11-01
FIXME: Default $CLUSTER_TYPE for new (s)c2s conditions

2uasimojo:
This is a band-aid for an issue introduced by #28149 which added some checks for a `$CLUSTER_TYPE` variable which apparently is unset in some affected tests. This collides with `set -o nounset` (aka `set -u`) and blows up, spuriously failing tests e.g. in post steps during must-gather.

This commit changes files
- in which a `$CLUSTER_TYPE` check was introduced by #28149
- where `$CLUSTER_TYPE` was not previously referenced (because those must have been fine, or they would have also been exploding tests)

...to default to the empty string if unset.

It also changes `[` style conditions, which spawn a subshell, to `[[` style, which a) don't; and b) allow use of bash regexes to consolidate the check into a single expression.

FIXME: The author of the aforementioned PR should validate these code paths and make sure it is appropriate to default `$CLUSTER_TYPE` therein; and/or make sure it gets set to an appropriate value, whereupon the defaulting should be removed as it could cover up future cases where that variable should be set, but isn't.


2uasimojo:
validated that this at least fixes must-gather via https://github.com/openshift/release/pull/33656

PR #33682, opened 2022-11-02
Add WORKLOAD_KUBECONFIG env variable

chetan-rns:
The workload kubeconfig variable is used to install Argo CD and cluster-agent resources in the workload cluster

Ref: https://github.com/redhat-appstudio/managed-gitops/pull/259


samyak-jn (approved):
/lgtm

PR #33685, opened 2022-11-02
Remove `external-dns` and `external-dns-operator` image promotions in 4.12+

LorbusChris:
These images are now part of an OperatorHub operator, and not part of the core OpenShift payload anymore, yet they've still been promoted into CI release payloads until now.

It also excludes the images from mirroring to OKD imagestreams, to ensure no outdated images possibly still present in OCP CI releases end up in OKD releases.


LorbusChris:
/hold 
This is still missing the excludeTags

PR #33706, opened 2022-11-03
Installer: Disable e2e-agent-mce job by default

zaneb:
This job is permafailing since Oct 21 because the 4.12 operator catalog does not contain the local storage operator, and it won't be released until GA. Make the job on-demand only.


barbacbd (commented):
/lgtm

PR #33715, opened 2022-11-03
Add 4.3 to 4.12 upgrade on Azure

shellyyang1989:
Closes [OCPQE-12661](https://issues.redhat.com/browse/OCPQE-12661)

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-nightly-4.12-upgrade-from-stable-4.3.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-nightly-4.12-upgrade-from-stable-4.3.yaml
@@ -1,0 +1,119 @@
+base_images:
+  ansible:
+    name: "4.12"
+    namespace: ocp
+    tag: ansible
+  cli:
+    name: "4.12"
+    namespace: ocp
+    tag: cli
+  dev-scripts:
+    name: test
+    namespace: ocp-kni
+    tag: dev-scripts
+releases:
+  intermediate44:
+    release:
+      architecture: amd64
+      channel: stable
+      version: "4.4"
+  intermediate45:
+    release:
+      architecture: amd64
+      channel: stable
+      version: "4.5"
+  intermediate46:
+    release:
+      architecture: amd64
+      channel: stable
+      version: "4.6"
+  intermediate47:
+    release:
+      architecture: amd64
+      channel: stable
+      version: "4.7"
+  intermediate48:
+    release:
+      architecture: amd64
+      channel: stable
+      version: "4.8"
+  intermediate49:
+    release:
+      architecture: amd64
+      channel: stable
+      version: "4.9"
+  intermediate410:
+    release:
+      architecture: amd64
+      channel: stable
+      version: "4.10"
+  intermediate411:
+    release:
+      architecture: amd64
+      channel: stable
+      version: "4.11"
+  latest:
+    release:
+      architecture: amd64
+      channel: stable
+      version: "4.3"
+  target:
+    candidate:
+      architecture: amd64
+      product: ocp
+      stream: nightly
+      version: "4.12"
+resources:
+  '*':
+    requests:
+      cpu: 100m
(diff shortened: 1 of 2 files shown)


shellyyang1989:
The install failed twice due to the issue. Maybe we need to consider to change a platform.

```
level=error msg="Cluster operator monitoring Degraded is True with UpdatingPrometheusK8SFailed: Failed to rollout the stack. Error: running task Updating Prometheus-k8s failed: reconciling Prometheus object failed: creating Prometheus object failed: Prometheus.monitoring.coreos.com \"k8s\" is invalid: spec.storage.volumeClaimTemplate.metadata.creationTimestamp: Invalid value: \"null\": spec.storage.volumeClaimTemplate.metadata.creationTimestamp in body must be of type string: \"null\""
level=fatal msg="failed to initialize the cluster: Cluster operator monitoring is still updating"
```

shellyyang1989:
Hi @jianlinliu @gpei I heard that the issue is caused by `- ref: ipi-install-monitoringpvc`. I'm considering removing the step from 4.3 to 4.12 upgrade job. Would there be any side effects if we get rid of it? cc @JianLi-RH @juzhao 

```
level=error msg="Cluster operator monitoring Degraded is True with UpdatingPrometheusK8SFailed: Failed to rollout the stack. Error: running task Updating Prometheus-k8s failed: reconciling Prometheus object failed: creating Prometheus object failed: Prometheus.monitoring.coreos.com \"k8s\" is invalid: spec.storage.volumeClaimTemplate.metadata.creationTimestamp: Invalid value: \"null\": spec.storage.volumeClaimTemplate.metadata.creationTimestamp in body must be of type string: \"null\""
level=fatal msg="failed to initialize the cluster: Cluster operator monitoring is still updating"
```

juzhao:
if developers allow the PR[1] merged to 4.3, then no need to remove `- ref: ipi-install-monitoringpvc`
otherwise we can remove it, the step attached PVs for monitoring, which is not the same for a default cluster(it requires user to configure PVs after the cluster is ready)

[1] https://github.com/openshift/cluster-monitoring-operator/pull/1810

shellyyang1989:
Thanks @juzhao for the confirmation! I'm afraid we would not have new 4.3 release built because it's EOL. Let me remove the step `- ref: ipi-install-monitoringpvc`.

juzhao:
> Thanks @juzhao for the confirmation! I'm afraid we would not have new 4.3 release built because it's EOL. Let me remove the step `- ref: ipi-install-monitoringpvc`.

yes, we only support 4.6+ versions

shellyyang1989:
@juzhao The test purpose is upgrading from 4.3 to latest.

juzhao:
@shellyyang1989 
see from https://coreos.slack.com/archives/G79AW9Q7R/p1667530416192859
since 4.3 is EOF, we can not merge openshift/cluster-monitoring-operator#1810 to 4.3, remove ipi-install-monitoringpvc step seems the right solution

gpei:
@shellyyang1989 seems Junqi already pointed out the cause of the monitoring Degraded issue, actually the `ipi-install-monitoringpvc` failure we talked last week is about: when "storage" operator was disabled, we need to remove this step from the related chain, not like this one.
BTW, I remember we have a CCO bug that blocks fresh install on GCP which version is < 4.6, but I see the install is indeed passed in the rehearse, this is a little confusing...
```
INFO[2022-11-04T11:02:16Z] Step phase pre succeeded after 26m20s.       
```

shellyyang1989:
> CCO

Thanks @gpei . Right, we have [Bug 2022840](https://bugzilla.redhat.com/show_bug.cgi?id=2022840). Seems like 4.4/4.5 install would be affected.

shellyyang1989:
The [test](https://qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/qe-private-deck/pr-logs/pull/openshift_release/33715/rehearse-33715-periodic-ci-openshift-openshift-tests-private-release-4.12-amd64-nightly-4.12-upgrade-from-stable-4.3-azure-ipi/1590724414477963264) passed. 

It's ready for your review and merge. Thanks! cc @liangxia @xingxingxia @jhou1

PR #33721, opened 2022-11-03
MGMT-12415: Use release:latest in all assisted jobs

adriengentil:
This changes aligns peridics and presubmits to use the same OCP release.

### DIFF
--- a/ci-operator/step-registry/assisted/common/setup/prepare/assisted-common-setup-prepare-commands.sh
+++ b/ci-operator/step-registry/assisted/common/setup/prepare/assisted-common-setup-prepare-commands.sh
@@ -89,6 +89,7 @@ export TEST_TEARDOWN=false
 export TEST_FUNC=test_install
 export ASSISTED_SERVICE_HOST={{ IP }}
 export PUBLIC_CONTAINER_REGISTRIES="{{ CI_REGISTRIES | join(',') }}"
+export OPENSHIFT_INSTALL_RELEASE_IMAGE={{ OPENSHIFT_INSTALL_RELEASE_IMAGE }}
 
 {% if ENVIRONMENT == "production" %}
 # Testing against the production AI parameters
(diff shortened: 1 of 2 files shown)


ci-operator/step-registry/assisted/common/setup/prepare/assisted-common-setup-prepare-commands.sh
```diff
@@ -89,6 +89,7 @@ export TEST_TEARDOWN=false
 export TEST_FUNC=test_install
 export ASSISTED_SERVICE_HOST={{ IP }}
 export PUBLIC_CONTAINER_REGISTRIES="{{ CI_REGISTRIES | join(',') }}"
+export OPENSHIFT_INSTALL_RELEASE_IMAGE={{ OPENSHIFT_INSTALL_RELEASE_IMAGE }}
```
osherdp, line 92: so what will happen on the e2e-metal-assisted-4-y jobs?
Doesn't test-infra uses ``OPENSHIFT_INSTALL_RELEASE_IMAGE`` (if it presents) to create an AI instance that only supports one OCP version?

osherdp (changes requested):

ci-operator/step-registry/assisted/common/setup/prepare/assisted-common-setup-prepare-commands.sh
adriengentil: We don't change the logic here as this variable was always set for periodics. The `e2e-metal-assisted-4-y` are using a variable `OPENSHIFT_VERSION`, I guess it overrides `OPENSHIFT_INSTALL_RELEASE_IMAGE` ?

adriengentil: Yep, looks like the case: https://github.com/openshift/assisted-test-infra/blob/master/src/assisted_test_infra/test_infra/utils/utils.py#L466-L498

osherdp: Very well. Good to know we're doing the right thing in test-infra :)

osherdp (approved):

PR #33735, opened 2022-11-03
OSSM-2156: Setup jobs for maistra operator 2.4

jwendell:


bartoszmajsak (approved):

ci-operator/config/maistra/istio-operator/maistra-istio-operator-maistra-2.4.yaml
```diff
@@ -0,0 +1,47 @@
+build_root:
+  image_stream_tag:
+    name: maistra-builder
+    namespace: ci
+    tag: "2.4"
+  use_build_cache: true
+resources:
+  '*':
+    limits:
+      memory: 16Gi
```
yxun, line 10: I don't think we need 16G memory in operator job. what about 4Gi or 8Gi in the limits ?

jwendell: this is a copy of whatever was there initially. We can do a clean up in a follow up pr.

yxun:
Is there a newer version of govet or gofumpt running in the 2.4 builder ? I had some similar lint errors from my laptop when I run make lint with newer versions of tools

jwendell:
> Is there a newer version of govet or gofumpt running in the 2.4 builder ? I had some similar lint errors from my laptop when I run make lint with newer versions of tools

Yes. Typically newer images contain newer versions of tooling. Our 2.4 image is based on Istio 1.16.

PR #33750, opened 2022-11-04
Add liangxia as approver for add workflow to chat-bot

liangxia:
QE are adding dozens of new workflows to cover more test profiles, and we'd like to use cluster-bot to launch clusters with some of the workflows to reproduce/debug issues. So adding workflow to [workflows-config.yaml](https://github.com/openshift/release/blob/master/core-services/ci-chat-bot/workflows-config.yaml) happens often. To avoid disturbing you guys too much, I'm request to be a approver for this.

/cc @jhou1 @kasturinarra

### DIFF
--- a/core-services/ci-chat-bot/OWNERS
+++ b/core-services/ci-chat-bot/OWNERS
@@ -1,3 +1,4 @@
 approvers:
 - AlexNPavel
 - bradmwilliams
+- liangxia


smg247:
/uncc
I am not the one to grant this permission

smg247:
@liangxia I am not the one to grant this permission. @bradmwilliams will have to decide that for you.

bradmwilliams (approved):

PR #33760, opened 2022-11-04
Add run monitor as part of the resource watch observer

xueqzhan:
[TRT-469](https://issues.redhat.com/browse/TRT-469)


dgoodwin:
/lgtm
/hold 

Feel free to release hold when you're happy with the rehersal results. Thanks for fixing the trap as well!

PR #33768, opened 2022-11-04
assisted-nutanix: use minimal ISO

vrutkovs:
No need to upload 1GB ISO when starting new machines
/hold

This fixes "out of memory" error in [periodics](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-assisted-test-infra-master-e2e-nutanix-assisted-periodic/1589178711997943808) as terraform provider crashes due to a large reply from Nutanix


eliorerz (approved):
LGTM

PR #33782, opened 2022-11-05
[Workflow Enablement] Use MCE create HostedCluster on AWS in Prow CI

LiangquanLi930:
The purpose of this PR is to build a ci workflow for installing the hypershift cluster by MCE version 2.2.

Why create an MCE 2.2 folder?
Currently MCE 2.2 is still using the HyperShift CLI, in the future MCE 2.3 will use a different approach, considering the future scenario if 2.2 and 2.3 coexist.

Since this is the first PR to install a hypershift cluster by MCE, we currently only use the upstream MCE Operator to install hypershift. We will support downstream also in future.

Reference:
The MCE operator installation. https://github.com/stolostron/deploy/tree/master/multiclusterengine

Create a hyperShift cluster with MCE operator 2.2.   
https://github.com/stolostron/hypershift-addon-operator/blob/main/docs/provision_hosted_cluster_on_mce_local_cluster.md


LiangquanLi930:
截屏2022-11-08 15 17 27

rokej:
@LiangquanLi930 this looks good to me. Thank you.

PR #33785, opened 2022-11-06
[openshift-loki] Increase lint timeout to 10m0s

periklis:
Address lint timeouts in openshift/loki/pull/67

### DIFF
--- a/ci-operator/config/openshift/loki/openshift-loki-release-5.5.yaml
+++ b/ci-operator/config/openshift/loki/openshift-loki-release-5.5.yaml
@@ -65,6 +65,7 @@ tests:
         requests:
           cpu: 100m
           memory: 200Mi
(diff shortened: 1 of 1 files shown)


ci-operator/config/openshift/loki/openshift-loki-release-5.5.yaml
```diff
@@ -65,6 +65,7 @@ tests:
         requests:
           cpu: 100m
           memory: 200Mi
+      timeout: 10m0s
```
xperimental, line 68: This is the wrong timeout. It needs an option to `golangci-lint` itself. See [here](https://github.com/openshift/cluster-logging-operator/pull/1644).

xperimental (changes requested):

PR #33794, opened 2022-11-07
Add release-4.11 and release-4.12 to NFD operand prow config

yevgeny-shnaidman:


core-services/prow/02_config/openshift/node-feature-discovery/_prowconfig.yaml
```diff
@@ -3,6 +3,7 @@ tide:
   - includedBranches:
     - release-4.0
     - release-4.10
+    - release-4.11
```
yselkowitz, line 6: Given that 4.12 has already branched:
```suggestion
    - release-4.11
    - release-4.12
```

PR #33795, opened 2022-11-07
HCO: always use a named bundle image

tiraboschi:
Always use an explicitly named bundle image
for compatibility with cluster-bot


orenc1 (approved):
/approve

PR #33800, opened 2022-11-07
AGENT-415: Adding (optional) sno ipv6 agent job (dev-scrips repo)

andfasano:
The SNO case it's quite a peculiar case, and since we've added some specific logic in dev-scripts to handle it then it'd be useful to have a test job to verify any dev-scripts related changed, before they could get shipped and used in other repos (in this case, in the openshift-installer one).

Additionally, a previous agent job naming/workflow has been fixed, to use the latest available generic workflow and config vars.


zaneb:
/hold
I agree that we need to test this topology, but I don't think this should run on ~every installer PR.
We also have multiple open proposals to add new jobs that each vary only on parameter. We need to come up with a unified strategy as captured in https://issues.redhat.com/browse/AGENT-417

andfasano:
/hold cancel

As not affecting the installer repo

celebdor:
/retitle AGENT-415: Adding (optional) sno ipv6 agent job

PR #33809, opened 2022-11-07
Use the 2.6 branch assisted image service for 2.6 tests

carbonin:
Without this the 2.6 subsystem tests were using the latest image service image which caused authentication errors because
https://github.com/openshift/assisted-service/pull/4559 is not backported to 2.6

cc @osherdp

### DIFF
--- a/ci-operator/config/openshift/assisted-service/openshift-assisted-service-release-ocm-2.6.yaml
+++ b/ci-operator/config/openshift/assisted-service/openshift-assisted-service-release-ocm-2.6.yaml
@@ -323,6 +323,8 @@ tests:
       dependencies:
       - env: SERVICE
         name: assisted-service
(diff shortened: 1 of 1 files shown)


osherdp (approved):

osherdp:
/pj-rehearse
/hold
just for good measures

osherdp:
Maybe we should do the same for master branch. There is a bit of difference between what we have in prow's registry and ``quay.io/edge-infrastructure/assisted-image-service:latest`` because it's mirroring happens every couple of hours I think
In addition, depending less on quay.io which is always a good idea

carbonin:
>Maybe we should do the same for master branch

:man_shrugging: Let's just get this in for the 2.6 branch then we can make other improvements separately.

osherdp:
/unhold
seems to pass :tada:

PR #33817, opened 2022-11-07
Tag by commit for openshift-knative/eventing-kafka-broker@1.4

aliok:
Creating an additional config file for promoting images of openshift-knative/eventing-kafka-broker@1.4 with tag by commit.

Why 1.4?
- Making the same config working for main/nightly doesn't prove anything
- It is safer to play with that branch instead of 1.5, which is to be released soon with the product

### DIFF
--- a/ci-operator/config/openshift-knative/eventing-kafka-broker/openshift-knative-eventing-kafka-broker-release-v1.4__411_tag_by_commit.yaml
+++ b/ci-operator/config/openshift-knative/eventing-kafka-broker/openshift-knative-eventing-kafka-broker-release-v1.4__411_tag_by_commit.yaml
@@ -32,8 +32,8 @@ images:
 promotion:
   additional_images:
     eventing-kafka-broker-src: src
-  name: knative-v1.4-commit-by-tag
   namespace: openshift
+  tag: knative-v1.4
   tag_by_commit: true
 releases:
   initial:


ci-operator/config/openshift-knative/eventing-kafka-broker/openshift-knative-eventing-kafka-broker-release-v1.4__411_tag_by_commit.yaml
```diff
@@ -0,0 +1,58 @@
...
+  inputs:
+    openshift_release_golang-1.17:
+      as:
+      - registry.ci.openshift.org/openshift/release:golang-1.17
+  to: knative-eventing-kafka-broker-webhook-kafka
+promotion:
+  additional_images:
+    eventing-kafka-broker-src: src
+  # following needs to be unique, but its value is not actually used anywhere
+  name: knative-v1.4-commit-by-tag
```
aliok, line 36: As wrote in the comment, this name needs to be unique. However, it is not used anywhere

pierDipi:
Why not modifying the promotion here to `disabled: false` and then add tag+tag by commit: https://github.com/openshift/release/blob/a2334fe53296dfb452cc27f6cb041b229020edef/ci-operator/config/openshift-knative/eventing-kafka-broker/openshift-knative-eventing-kafka-broker-release-v1.4__46.yaml#L134-L139

aliok:
> Why not modifying the promotion here to disabled: false and then add tag+tag by commit:

I thought we should better use 411 variant also for tag-by-commit.

ci-operator/config/openshift-knative/eventing-kafka-broker/openshift-knative-eventing-kafka-broker-release-v1.4__411_tag_by_commit.yaml
```diff
@@ -0,0 +1,58 @@
...
+  namespace: openshift
+  tag_by_commit: true
```
pierDipi, line 38: name + tag_by_commit wasn't supported last time I checked, we need to replace name with `tag`

pierDipi:
```suggestion
  tag: knative-v1.4
  namespace: openshift
  tag_by_commit: true
```

This should do it.

ci-operator/config/openshift-knative/eventing-kafka-broker/openshift-knative-eventing-kafka-broker-release-v1.4__411_tag_by_commit.yaml
```diff
@@ -0,0 +1,57 @@
...
+resources:
+  '*':
+    requests:
+      cpu: 500m
+      memory: 1Gi
+zz_generated_metadata:
+  branch: release-v1.4
+  org: openshift-knative
+  repo: eventing-kafka-broker
+  variant: 411_tag_by_commit
```
pierDipi, line 57: is this variant correct?

aliok: Yeah, `make ci-operator-config`

pierDipi:
> > Why not modifying the promotion here to disabled: false and then add tag+tag by commit:
> 
> I thought we should better use 411 variant also for tag-by-commit.

it shouldn't really matter since images are container images and I would be surprised if those images are different based on the OCP variant, right?

ci-operator/config/openshift-knative/eventing-kafka-broker/openshift-knative-eventing-kafka-broker-release-v1.4__411_tag_by_commit.yaml
aliok: Yeah, you're right. According to https://docs.ci.openshift.org/docs/architecture/ci-operator/#publishing-images-tagged-by-commit, it should be tag + tagByCommit.

I don't see any diff in generated job yaml though.

Anyway, I changed it.

aliok:
> it shouldn't really matter since images are container images and I would be surprised if those images are different based on the OCP variant, right?

I still want to have this in an additional file. Reasons:
- I don't want to touch existing config
- I see the `zz_generated_metadata -> variant` is changed to `46_tag_by_commit`, which again, changes something I don't want to touch.
- It is more clear what the separate file is for.

We can get rid of it later, when we eventually change from [integration stream](https://docs.ci.openshift.org/docs/architecture/ci-operator/#publishing-to-an-integration-stream) (aka abusing image tags) to tagByCommit completely.

aliok:
/close
Included with https://github.com/openshift/release/pull/33742

PR #33818, opened 2022-11-07
`pj-rehearse` plugin: add rehearsals-ack label

smg247:
We need the label for "phase two" where it is required to merge

/cc @openshift/test-platform @droslean

### DIFF
--- a/core-services/prow/02_config/_labels.yaml
+++ b/core-services/prow/02_config/_labels.yaml
@@ -587,6 +587,12 @@ repos:
         target: both
         prowPlugin: label
         addedBy: anyone
+      - color: 5df542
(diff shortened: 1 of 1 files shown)


core-services/prow/02_config/_labels.yaml
```diff
@@ -587,6 +587,12 @@ repos:
         target: both
         prowPlugin: label
         addedBy: anyone
+      - color: 5df542
```
droslean, line 590: I think it would be better to change this in a different PR. And after the `label-sync` job succeeds, you can add the label to tide's config.

hongkailiu (approved):

PR #33819, opened 2022-11-07
OpenStack: Generate 4.13 periodics

mandre:
Generate 4.13 periodic jobs and disable 4.7 (out of support branch) jobs.
Also add missing slack notification for some 4.12 jobs.

### DIFF
--- a/ci-operator/config/shiftstack/shiftstack-ci/shiftstack-shiftstack-ci-main__periodic-4.13-upgrade-from-stable-4.12.yaml
+++ b/ci-operator/config/shiftstack/shiftstack-ci/shiftstack-shiftstack-ci-main__periodic-4.13-upgrade-from-stable-4.12.yaml
@@ -1,10 +1,10 @@
 base_images:
   openstack-installer:
-    name: '4.12'
+    name: "4.12"
     namespace: ocp
     tag: openstack-installer
   openstack-tests:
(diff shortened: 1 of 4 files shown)


ci-operator/config/shiftstack/shiftstack-ci/shiftstack-shiftstack-ci-main__periodic-4.13-upgrade-from-stable-4.12.yaml
```diff
@@ -0,0 +1,54 @@
...
+      version: '4.13'
+resources:
+  '*':
+    limits:
+      memory: 4Gi
+    requests:
+      cpu: 100m
+      memory: 200Mi
+tests:
+  - as: e2e-openstack-sdn-upgrade
```
MaysaMacedo, line 30: looks like this one will test with ovn which is the default , shoud the sdn me removed from the name?
https://github.com/openshift/release/blob/master/ci-operator/step-registry/openshift/upgrade/openstack/loki/openshift-upgrade-openstack-loki-workflow.yaml

mandre: You're right, I'll do it.

PR #33827, opened 2022-11-07
Increase verify-generated-code step memory/CPU resources

omertuc:
Some performance improvements means we're using more CPU / memory,
to avoid the job getting [OOMKilled](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_assisted-service/4589/pull-ci-openshift-assisted-service-master-edge-verify-generated-code/1589680483808055296), we need more 

https://github.com/openshift/assisted-service/pull/4589

### DIFF
--- a/ci-operator/step-registry/assisted/verify-generated-code/assisted-verify-generated-code-ref.yaml
+++ b/ci-operator/step-registry/assisted/verify-generated-code/assisted-verify-generated-code-ref.yaml
@@ -5,9 +5,9 @@ ref:
   commands: assisted-verify-generated-code-commands.sh
   resources:
     requests:
-      cpu: "3"
-      memory: 600Mi
+      cpu: "4"
+      memory: 850Mi
     limits:
-      memory: 2Gi
+      memory: 1Gi
   documentation: |-
     Make sure all required generated code has been committed.


droslean:
/hold
@omertuc I am holding this PR to gather some data about how many resources the job needs. Did you check that already?

omertuc:
> /hold @omertuc I am holding this PR to gather some data about how many resources the job needs. Did you check that already?

I haven't. However, the more logical core the container thinks it has (`nproc`), the more RAM it will use. So it's a bit hard to tell

droslean:
We have https://pod-scaler-ui-ci.apps.ci.l2s4.p1.openshiftapps.com/usage/pods?branch=master&container=test&org=openshift&repo=assisted-service&target=edge-verify-generated-code&variant= to measure it. I rerun the jobs to gather more data.

omertuc:
Thank you, updated the values to the recommended values

omertuc:
Please unhold if you think it's ok;

ci-operator/step-registry/assisted/verify-generated-code/assisted-verify-generated-code-ref.yaml
```diff
@@ -5,9 +5,9 @@ ref:
   commands: assisted-verify-generated-code-commands.sh
   resources:
     requests:
-      cpu: "3"
-      memory: 600Mi
+      cpu: "4"
+      memory: 1650Mi
```
droslean, line 9: This is the same as the limit.

omertuc: Is that bad?

omertuc: I thought that was the ideal (Guaranteed QoS class etc)

PR #33844, opened 2022-11-08
MGMT-11926: assisted-vsphere: bump vsphere machine specs

vrutkovs:
assisted-vsphere: bump vsphere machine specs, use dedicated workers

Update CPU/RAM settings for worker machines created during periodic.
Unlike presubmit tests, we run conformance test during periodic, so these need more resources to pass tests. This will make periodic job run dedicated workers.

Resources are synced with ipi-conf-vsphere step


vrutkovs:
Two tests failing now:
* [sig-arch][Late] operators should not create watch channels very often [apigroup:config.openshift.io]
* [sig-instrumentation][Late] Alerts shouldn't report any unexpected alerts in firing or pending state

Both seem to be flakes
/cc @eliorerz

PR #33848, opened 2022-11-08
update hypershift-guest 4.12

LiangquanLi930:


heliubj18:
The cucushift is not ready now. just add it here, no case would be selected.

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-nightly.yaml
```diff
@@ -392,12 +392,18 @@ tests:
     cluster_profile: aws-qe
     env:
       BASE_DOMAIN: qe.devcluster.openshift.com
+      E2E_RUN_TAGS: '@aws-ipi and @network-ovnkubernetes and @hypershift-hosted'
```
liangxia, line 395: Also need `and not @fips`

liangxia:
ci/prow/ci-operator-config-metadata — Job failed.

PR #33854, opened 2022-11-08
Remove verify-deps job

Jaanki:
Now that `vendor` directory is no longer a part of submariner-addon source code, remove `verify-deps` job. This job checks if the local vendor/* content is up-to-date with go.mod.

Signed-off-by: Janki Chhatbar <jchhatba@redhat.com>


Jaanki:
> /approve

@skitt looks like this did not work. May be it needs to be approved from the UI options.

PR #33860, opened 2022-11-08
NO-ISSUE: Populate authorized_keys using cloud-init

adriengentil:
Equinix does not always populate the authorized_keys file with the
public keys defined in the console. This change uses cloud-init to
ensure it is done.


osherdp (approved):

adriengentil:
It looks like a problem with the metadata endpoint, so this solution does not work :cry: 
/close

PR #33874, opened 2022-11-09
Add upi-installer for cucushift-installer-rehearse-baremetalds-ipi-ovn-ipv4

liangxia:
/cc @yapei @XiyunZhao @jhou1


yapei:
thank you very much for the updates

PR #33875, opened 2022-11-09
Automating agent-based installation on rdu2 lab

jhou1:
This PR creates a job that deploys an SNO with agent-based installer in rdu2 lab. This is not our final solution for automating cluster deployments with rdu2 machines, our ultimate implementation is described in the `QE-BM-Infra-RDU2` document, which depends on the ARM, Metal, and Agent QE's capacity to work together, until then we have this to kick off Agent QE's CI.

cc @mhanss @zniu1011

### DIFF
--- a/ci-operator/step-registry/baremetal/lab/install/agent/baremetal-lab-install-agent-commands.sh
+++ b/ci-operator/step-registry/baremetal/lab/install/agent/baremetal-lab-install-agent-commands.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+
+RELEASE_IMAGE=$(oc adm release info --registry-config "${CLUSTER_PROFILE_DIR}/pull-secret" "${RELEASE_IMAGE_LATEST}" -o json|jq -r '.metadata.version')
+
+SSHOPTS=(-o 'ConnectTimeout=5'
+  -o 'StrictHostKeyChecking=no'
+  -o 'UserKnownHostsFile=/dev/null'
+  -o 'ServerAliveInterval=90'
+  -o LogLevel=INFO
+  -i "${CLUSTER_PROFILE_DIR}/ssh-key")
+
+# shellcheck disable=SC2087
+ssh "${SSHOPTS[@]}" root@"${AUX_HOST}" <<EOF
+cp "/root/workdir/agent-bm-deployments/prow_${DEPLOYMENT_TYPE}" /root/workdir/agent-bm-deployments/prow_inventory
+cat <<OOO >> /root/workdir/agent-bm-deployments/prow_inventory
+DEPLOYMENT_TYPE=${DEPLOYMENT_TYPE}
+IP_STACK=${IP_STACK}
+CLUSTER_NAME=agent${DEPLOYMENT_TYPE}
+DISCONNECTED=${DISCONNECTED}
+PROXY=${PROXY}
+FIPS=${FIPS}
+RELEASE_IMAGE=registry.ci.openshift.org/ocp/release:${RELEASE_IMAGE}
+OOO
+
+cd /root/workdir/agent-bm-deployments/
+ansible-playbook -i prow_inventory install.yaml
+./openshift-install agent wait-for bootstrap-complete
+./openshift-install agent wait-for install-complete
+EOF
--- a/ci-operator/step-registry/baremetal/lab/post/agent/baremetal-lab-post-agent-commands.sh
+++ b/ci-operator/step-registry/baremetal/lab/post/agent/baremetal-lab-post-agent-commands.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+#
+set -o nounset
+set -o errexit
+set -o pipefail
+
+SSHOPTS=(-o 'ConnectTimeout=5'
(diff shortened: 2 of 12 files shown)


jhou1:
@jianlinliu @liangxia @aleskandro PTAL

ci-operator/step-registry/baremetal/lab/install/agent/baremetal-lab-install-agent-commands.sh
```diff
@@ -0,0 +1,33 @@
...
+RELEASE_IMAGE=$(oc adm release info --registry-config "${CLUSTER_PROFILE_DIR}/pull-secret" "${RELEASE_IMAGE_LATEST}" -o json|jq -r '.metadata.version')
+
+SSHOPTS=(-o 'ConnectTimeout=5'
+  -o 'StrictHostKeyChecking=no'
+  -o 'UserKnownHostsFile=/dev/null'
+  -o 'ServerAliveInterval=90'
+  -o LogLevel=INFO
+  -i "${CLUSTER_PROFILE_DIR}/ssh-key")
+
+ssh "${SSHOPTS[@]}" root@openshift-qe-055.arm.eng.rdu2.redhat.com <<EOF
```
jianlinliu, line 17: Maybe we would get some infosec alert email for the dns ?

ci-operator/step-registry/baremetal/lab/install/agent/baremetal-lab-install-agent-commands.sh
```diff
@@ -0,0 +1,33 @@
...
+CLUSTER_NAME=agent${DEPLOYMENT_TYPE}
+DISCONNECTED=${DISCONNECTED}
+PROXY=${PROXY}
+FIPS=${FIPS}
+RELEASE_IMAGE=registry.ci.openshift.org/ocp/release:${RELEASE_IMAGE}
+OOO
+
+cd /root/workdir/agent-bm-deployments/
+ansible-playbook -i prow_inventory install.yaml
+./openshift-install agent wait-for bootstrap-complete
```
jianlinliu, line 31: If we support multiple versions of agent install, `openshfit-install` need to change accordingly?

ci-operator/step-registry/baremetal/lab/post/agent/baremetal-lab-post-agent-commands.sh
```diff
@@ -0,0 +1,17 @@
...
+SSHOPTS=(-o 'ConnectTimeout=5'
+  -o 'StrictHostKeyChecking=no'
+  -o 'UserKnownHostsFile=/dev/null'
+  -o 'ServerAliveInterval=90'
+  -o LogLevel=INFO
+  -i "${CLUSTER_PROFILE_DIR}/ssh-key")
+
+ssh "${SSHOPTS[@]}" root@openshift-qe-055.arm.eng.rdu2.redhat.com <<EOF
+cd /root/workdir/agent-bm-deployments/
+ansible-playbook -i prow_inventory dns-cleanup.yaml
```
jianlinliu, line 16: Seem like `dns-cleanup.yaml` and `install.yaml` only available in `openshift-qe-055.arm.eng.rdu2.redhat.com`, right?

jhou1: Yes, I had thought about using an image or multiple steps to handle that so the installation program is also under code review, but that makes things difficult because the installation manifests needs to take hostnames, IP addresses, credentials, ssh-key, pull secrets. There are some security implications to do it that way. 

A good solution is WIP: QE-BM-Infra-RDU2, but this solution probably won't complete during our 4.12 timeframe

ci-operator/step-registry/baremetal/lab/install/agent/baremetal-lab-install-agent-commands.sh
jianlinliu: we need multiple concurrent running jobs case, to avoid the openshift-installer is overridden by different jobs.

jhou1: Good point, let me take that into account.

jianlinliu: For the ansible-playbook running, seem like `openshift-qe-055.arm.eng.rdu2.redhat.com` is used for something like bastion, if it is, maybe we can prepare `prow_inventory` on slaves, and inject ssh proxy via `ansible_ssh_common_args` in the `prow_inventory` , even include `openshift-install agent` commands in the playbook together.

jhou1:
After a discussion with @aleskandro yesterday, we decided to preserve bm hosts using steps for BM IPI/UPI by adding an additional column needed by the agent installer, I'll explore that option and use another branch to test.

/hold

ci-operator/step-registry/baremetal/lab/install/agent/baremetal-lab-install-agent-commands.sh
jhou1: Acked, addressed this in an ENV

jhou1:
The implementation of this has diverged and Biagio will submit the PR for agent automation, no need to keep this PR now.

/close

PR #33876, opened 2022-11-09
OCPQE-12807: Add z-stream upgrade for 4.12

liangxia:
Verified the nightly builds are picked up correctly,
```
INFO[2022-11-09T04:03:54Z] Resolved release latest to registry.ci.openshift.org/ocp/release:4.12.0-0.nightly-2022-11-07-104414 
INFO[2022-11-09T04:03:54Z] Resolved release target to registry.ci.openshift.org/ocp/release:4.12.0-0.nightly-2022-11-07-181244
```

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-nightly-4.12-upgrade-from-nightly-4.12.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-nightly-4.12-upgrade-from-nightly-4.12.yaml
@@ -1,0 +1,555 @@
+base_images:
+  ansible:
+    name: "4.12"
+    namespace: ocp
+    tag: ansible
+  cli:
+    name: "4.12"
+    namespace: ocp
+    tag: cli
+  dev-scripts:
+    name: test
+    namespace: ocp-kni
+    tag: dev-scripts
+  openstack-installer:
+    name: "4.12"
+    namespace: ocp
+    tag: openstack-installer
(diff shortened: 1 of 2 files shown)


liangxia:
The failures are caused by wrong master number calculation in 4.12, which is fixed in https://github.com/openshift/release/pull/33884

shellyyang1989:
Looks good to me. cc @jianlinliu to take a look in case any concern.

jianlinliu:
Most looks good to me. Now my only concern is about if our cloudproviders can bare so many concurrent jobs. Recently we received more alerts about usage quota issues.

liangxia:
> Most looks good to me. Now my only concern is about if our cloudproviders can bare so many concurrent jobs. Recently we received more alerts about usage quota issues.

Perhaps we can let it go currently, then we raise the quota issue to the managers, let them decide whether we can request more quota, or set lower test frequency for existing jobs. WDYT ?

jianlinliu:
> Perhaps we can let it go currently, then we raise the quota issue to the managers, let them decide whether we can request more quota, or set lower test frequency for existing jobs. WDYT ?

Yeah, the quota concern is a bit off topic about this PR, we can let the PR go. But the quota issue is a very important issues, we need to discuss it in a short future.

Take aws as example, now there are 26 p1 jobs on aws, plus the adding ones in this PR - 8, would be 34 in total. That means, 34 clusters will be launched in one day, In the future, we are going to add more jobs for old versions, also upgrade testing. In one aws region, there would be 100 vpc quota, that would be definitely become a big problem. 

I will let this PR go, and add the topic in the next ci sync meeting.

PR #33883, opened 2022-11-09
openstack: Add NERC monitoring

pierreprinetti:
Also move the metric files to a container on NERC (from Vexxhost), to keep all monitoring infrastructure (metrics and Prometheus) on the same cloud.

### DIFF
--- a/ci-operator/step-registry/shiftstack/monitor/expose/shiftstack-monitor-expose-ref.yaml
+++ b/ci-operator/step-registry/shiftstack/monitor/expose/shiftstack-monitor-expose-ref.yaml
@@ -8,7 +8,7 @@ ref:
       memory: 100Mi
   credentials:
   - namespace: test-credentials
-    name: cluster-secrets-openstack-vexxhost
+    name: cluster-secrets-openstack-nerc-dev
     mount_path: /var/run/cluster-secrets/openstack
   env:
   - name: OS_CLOUD


ci-operator/jobs/shiftstack/shiftstack-ci/shiftstack-shiftstack-ci-main-periodics.yaml
```diff
@@ -518,6 +518,53 @@ periodics:
...
+  extra_refs:
+  - base_ref: main
+    org: shiftstack
+    repo: shiftstack-ci
+  interval: 5m
+  labels:
+    ci.openshift.io/generator: prowgen
+    pj-rehearse.openshift.io/can-be-rehearsed: "true"
+  name: periodic-ci-shiftstack-shiftstack-ci-main-monitor-nerc-dev
+  spec:
```
mandre, line 535: Don't forget to add slack notification.

pierreprinetti: We don't have that for monitoring jobs. They execute every minute, failures may become extremely noisy. Since this is used in Prometheus, I suggest settings alarms that side if we want that kind of alerts

mandre: Oh my bad, I was under the impression our other monitoring jobs reported on slack. All good then.

mandre (commented):
/lgtm

PR #33884, opened 2022-11-09
Fix master count

liangxia:
When the master nodes has multiple roles, the count is always 0.


shellyyang1989:
/lgtm
Thanks for the quick fix!

PR #33888, opened 2022-11-09
Remove hypershift related tests in 4.10

liangxia:
Hypershift is not supported in 4.10, thus remove it.

/cc @LiangquanLi930 @heliubj18 @jhou1 @wsun1


LiangquanLi930:
i am ok

PR #33891, opened 2022-11-09
[release-1.26] Make serverless 48 consistent with 411

pierDipi:
As per title, this would cause the 4.8 tests to actually run as well.


skonto:
We need to do the same for 4.8 on main too.

pierDipi:
> @pierDipi I think we need to do the same for 4.8 on main branch too.

Done

/pj-rehearse

PR #33904, opened 2022-11-09
Run Agent SNO job only on changes to agent-specific code

zaneb:
We don't need multiple agent jobs running on every change to the core
installer. Run the SNO job only on PRs that touch the agent-specific
parts, on both master and release-4.12.

### DIFF
--- a/ci-operator/config/openshift/installer/openshift-installer-release-4.12.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-release-4.12.yaml
@@ -525,6 +525,16 @@ tests:
         AGENT_E2E_TEST_SCENARIO=COMPACT_IPV4
         AGENT_DEPLOY_MCE=true
     workflow: agent-e2e-generic
+- as: e2e-agent-sno
+  cluster: build05
+  optional: true
+  run_if_changed: /agent/
+  steps:
+    cluster_profile: equinix-ocp-metal
+    env:
+      DEVSCRIPTS_CONFIG: |
+        AGENT_E2E_TEST_SCENARIO=SNO_IPV4_DHCP
+    workflow: agent-e2e-generic
 zz_generated_metadata:
   branch: release-4.12
   org: openshift
--- a/ci-operator/jobs/openshift/installer/openshift-installer-release-4.12-presubmits.yaml
+++ b/ci-operator/jobs/openshift/installer/openshift-installer-release-4.12-presubmits.yaml
@@ -201,6 +201,81 @@ presubmits:
     branches:
     - ^release-4\.12$
     - ^release-4\.12-
+    cluster: build05
+    context: ci/prow/e2e-agent-sno
+    decorate: true
+    labels:
+      ci-operator.openshift.io/cloud: equinix-ocp-metal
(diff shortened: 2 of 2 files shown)


ci-operator/config/openshift/installer/openshift-installer-master.yaml
```diff
@@ -528,7 +528,7 @@ tests:
 - as: e2e-agent-sno
   cluster: build05
   optional: true
-  skip_if_only_changed: (^(docs|upi|hack)/)|((^|/)OWNERS(_ALIASES)?$)|((^|/)[A-Z]+\.md$)|(^(data/.*|images|pkg/.*|terraform/providers)/(alibabacloud|aws|azure|azurestack|baremetal|gcp|ibmcloud|libvirt|nutanix|openstack|ovirt|powervs|vsphere)/)
+  run_if_changed: /agent/
```
andfasano, line 531: If we want to trigger the job only for specific agent code changes, shouldn't be `.*/agent/` (I'm not sure if the regex pattern must be escaped, from the prow [test](https://github.com/kubernetes/test-infra/blob/fe4f3e2158449e8a1b28834bbd2836102d30683f/prow/config/jobs_test.go#L199) seems so, in such case should be then `.*\/agent\/`)?

ci-operator/config/openshift/installer/openshift-installer-release-4.12.yaml
```diff
@@ -525,6 +525,16 @@ tests:
         AGENT_E2E_TEST_SCENARIO=COMPACT_IPV4
         AGENT_DEPLOY_MCE=true
     workflow: agent-e2e-generic
+- as: e2e-agent-sno
```
andfasano, line 528: from our latest team discussion, the new job name must be changed to `e2e-agent-compact-ipv4`

ci-operator/config/openshift/installer/openshift-installer-release-4.12.yaml
```diff
@@ -525,6 +525,16 @@ tests:
...
+  cluster: build05
+  optional: true
+  run_if_changed: /agent/
+  steps:
+    cluster_profile: equinix-ocp-metal
+    env:
+      DEVSCRIPTS_CONFIG: |
+        AGENT_E2E_TEST_SCENARIO=SNO_IPV4_DHCP
```
andfasano, line 536:
```suggestion
        AGENT_E2E_TEST_SCENARIO=SNO_IPV4
```

ci-operator/config/openshift/installer/openshift-installer-master.yaml
zaneb: Golang regexp doesn't have separate match/search functions, so `MatchString()` [matches anywhere](https://pkg.go.dev/regexp@go1.19.3#Regexp.Match) in the string. That's why the previous expression had `^` to match only at the start.
`/` has no special meaning in regexps, so it shouldn't be escaped.

andfasano: Ok thanks

zaneb:
/close
Handled by #34878

PR #33906, opened 2022-11-09
`ci-search` - Jira query: include bugs updated over the last 90 days

hoxhaeris:
This PR raises the time delta for Jira bugs to 90 days. 
As a result, `ci-search` will keep bugs updated within 90 days in the in-memory cache.  
After this period(if a bug does not have an update for more than 90 days), the bug is removed from the in-memory cache and it is marked as `closed` in the disk cache.
The issue can still be indexed, as long as it is not removed from the disk cache.


dgoodwin:
Thanks this is super important for us, few bugs get resolved in 14 days around here. 

/lgtm

PR #33916, opened 2022-11-10
added slack channel to notify

miyadav:
Thanks @jhou1 , I added as suggested in - https://docs.ci.openshift.org/docs/how-tos/notification/ 
PTAL @huali9 @sunzhaohua2 
Not sure about the note it says- 
`Note that we have to modify the job yaml directly even if the job is generated from a ci-operator’s config and regenerating the jobs does not change the existing reporter_config. Moreover, we do not support reporter_config for generated presubmits and postsubmits.`

But when I am running `make update` it does modify the job and `reporter_config` gets removed .


miyadav:
/hold 
Once confirmed will unhold it.

miyadav:
/unhold 
confirmed with team issue was with indent.

miyadav:
@sunzhaohua2 @huali9 @jhou1 PTAL , had to rebase .

PR #33918, opened 2022-11-10
Add 1.6, 1.7 and 1.8 for net-kourier and net-istio

nak3:
As per title.

/cc @skonto @mgencur


mgencur:
LGTM. My understanding is that you're not running any tests in pre-submit checks. This is only for building images in pre-submit and post-submit phase.

nak3:
> LGTM. My understanding is that you're not running any tests in pre-submit checks. This is only for building images in pre-submit and post-submit phase.

Yes, that's correct. net-kourier and net-istio do not have the e2e test in their repo.

PR #33921, opened 2022-11-10
try to re-enable the broken SSP tests

nunnatsa:


ksimon1:
this is manifest for old ss operator which is no longer in use. We can remove this file

PR #33924, opened 2022-11-10
Remove unsupported Eventing releases

pierDipi:
See https://access.redhat.com/support/policy/updates/openshift#ossrvless

v1.2 and below are unsupported


matzew:
This also means the image are _removed_ from the internal registry (just mentioning it, in case that is not clear)

matzew (approved):
/lgtm
/approve

matzew (approved):
/lgtm
/approve

PR #33956, opened 2022-11-11
OCPBUGS-3206: Ensure node reboot in SDN migration script

pliurh:
Add retry logic to node reboot. Make sure all the nodes can be rebooted.


pliurh:
@weliang1 Could you give this patch a try? See if it can fix the issue you found.

weliang1:
> @weliang1 Could you give this patch a try? See if it can fix the issue you found.

@pliurh Both scripts works for vsphere cluster using latest 4.12 build.
/lgtmp

PR #33972, opened 2022-11-11
ci: disable optional-operators-subscribe for addon-operator e2e

Ajpantuso:
### Summary

Temporarily drops the `optional-operators-subscribe step while `infra` node labeling is performed during the test steps.

### DIFF
--- a/ci-operator/config/openshift/addon-operator/openshift-addon-operator-main.yaml
+++ b/ci-operator/config/openshift/addon-operator/openshift-addon-operator-main.yaml
@@ -79,6 +79,16 @@ tests:
       OO_INSTALL_NAMESPACE: '!create'
       OO_PACKAGE: addon-operator
       OO_TARGET_NAMESPACES: '!all'
+    pre:
+    - chain: ipi-aws-pre
+    - as: label-infra-nodes
+      commands: ./mage test:integrationciprepare
+      from: src
+      resources:
+        requests:
+          cpu: 500m
+          memory: 1Gi
+    - ref: optional-operators-subscribe
     test:
     - as: test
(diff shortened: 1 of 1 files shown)


ci-operator/config/openshift/addon-operator/openshift-addon-operator-main.yaml
```diff
@@ -79,6 +79,8 @@ tests:
       OO_INSTALL_NAMESPACE: '!create'
       OO_PACKAGE: addon-operator
       OO_TARGET_NAMESPACES: '!all'
+    pre:
+    - chain: ipi-aws-pre
```
thetechnick, line 83: We also need a post chain for cleanup, or we might leave resources.
Or is `ipi-aws-post` included automatically?

thetechnick (changes requested):

ci-operator/config/openshift/addon-operator/openshift-addon-operator-main.yaml
Ajpantuso: Everything in the workflow is the same except for the missing pre-step https://github.com/openshift/release/blob/master/ci-operator/step-registry/optional-operators/ci/aws/optional-operators-ci-aws-workflow.yaml#L6.

Ajpantuso: https://docs.ci.openshift.org/docs/architecture/step-registry/#ci-operator-test-configuration

thetechnick:
A wait!
We also have to set `ADO_RUN_OPTIONAL_OPERATORS_SUBSCRIBE_STEP` as env-var or we the new script is not actually running.

PR #33973, opened 2022-11-11
openstack clouds: syntax fixes and added logs

pierreprinetti:
The application credentials flow requires the openstack-conf-clouds step
to generate application credentials and apply them to the clouds.yaml
file to be shared with subsequent steps. To generate application
credentials, a complete and functioning clouds.yaml must exist in the
first place; that's why this change moves the code exposing the HTTPS CA
earlier in the process. Before this change, application credentials
generation would fail on a cloud requiring `cacert`.

Additionally, this change fixes quoting issues and adds logs.

Lastly, this change extracts the application credentials generation to a
function in preparation for further changes, and more specifically the
one that will enable application credentials rotation.

### DIFF
--- a/ci-operator/step-registry/openstack/conf/clouds/openstack-conf-clouds-commands.sh
+++ b/ci-operator/step-registry/openstack/conf/clouds/openstack-conf-clouds-commands.sh
@@ -5,12 +5,18 @@ set -o errexit
 set -o pipefail
 
 CLUSTER_TYPE="${CLUSTER_TYPE_OVERRIDE:-$CLUSTER_TYPE}"
-export OS_CLIENT_CONFIG_FILE="${SHARED_DIR}/clouds.yaml"
 
-cp "/var/run/cluster-secrets/${CLUSTER_TYPE}/clouds.yaml" "$OS_CLIENT_CONFIG_FILE"
+clouds_yaml="$(mktemp)"
+cp "/var/run/cluster-secrets/${CLUSTER_TYPE}/clouds.yaml" "$clouds_yaml"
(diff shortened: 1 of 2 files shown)


mandre (commented):
It's not clear to me what the fix consists in. I see you wrapped the creation of the clouds.yaml when using app creds in a function, now use a temp file for clouds.yaml, and moved the part that sets the cacert in the clouds.yaml before generating the app creds... but all that seems like refactoring to me. Could you highlight what the error was in your commit message at least?

pierreprinetti:
> It's not clear to me what the fix consists in. I see you wrapped the creation of the clouds.yaml when using app creds in a function, now use a temp file for clouds.yaml, and moved the part that sets the cacert in the clouds.yaml before generating the app creds... but all that seems like refactoring to me. Could you highlight what the error was in your commit message at least?

good point; I have amended the commit message and edited the PR description. I have moved this code around so much that I forgot basic commit hygiene.

mandre:
> > It's not clear to me what the fix consists in. I see you wrapped the creation of the clouds.yaml when using app creds in a function, now use a temp file for clouds.yaml, and moved the part that sets the cacert in the clouds.yaml before generating the app creds... but all that seems like refactoring to me. Could you highlight what the error was in your commit message at least?
> 
> good point; I have amended the commit message and edited the PR description. I have moved this code around so much that I forgot basic commit hygiene.

Thanks, it's now clear what the problem was: we tried to generate application credentials with a clouds.yaml that did not reference the cacert.

mandre (commented):
/lgtm
/approve

PR #33977, opened 2022-11-11
Watch clusterversion for cluster readiness

csrwng:
Instead of looking at individual cluster operators we should look at the clusterversion to determine when a cluster is ready to be used


sjenning:
CI is blocked without this.  Rehearsal is not passing but it is getting farther than CI is at the moment.
/lgtm

PR #33980, opened 2022-11-11
update camgi to version 0.8.1

elmiko:
this change is being proposed to bring in new functionality in camgi that will prepend `camgi-` to the title of the generated web page. it is being added to help distinguish the camgi output when it is embedded in a prow page through spyglass.


elmiko:
@vrutkovs might i trouble you for an approval? =)

PR #33989, opened 2022-11-12
Skip an OLM that is incompatible with openshift

awgreene:
This commit skips one of OLM's e2e tests that requires OLM to be built with a special go build tag. The go build tag required by this test changes OLM's behavior to support a "Deprecated" CRD type in the bundle manifests, which allows this test to ensure that an alert is surfaced when a bundle includes a deprecated resource.

The team is chosing to disable this test as we do not want to:
- Test a version of OLM that includes this build tag and then release a separate version of OLM that does not include the tag.
- Introduce an additional prow job to test this build of OLM.

### DIFF
--- a/ci-operator/config/openshift/operator-framework-olm/openshift-operator-framework-olm-master.yaml
+++ b/ci-operator/config/openshift/operator-framework-olm/openshift-operator-framework-olm-master.yaml
@@ -84,7 +84,13 @@ tests:
     test:
     - as: test
       cli: latest
-      commands: JUNIT_DIRECTORY=${ARTIFACT_DIR} make e2e/olm SKIP='\[FLAKE\]'
+      commands: |
+        export JUNIT_DIRECTORY=${ARTIFACT_DIR}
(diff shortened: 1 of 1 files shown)


ci-operator/config/openshift/operator-framework-olm/openshift-operator-framework-olm-master.yaml
```diff
@@ -84,7 +84,10 @@ tests:
     test:
     - as: test
       cli: latest
-      commands: JUNIT_DIRECTORY=${ARTIFACT_DIR} make e2e/olm SKIP='\[FLAKE\]'
+      commands: |
+        export JUNIT_DIRECTORY=${ARTIFACT_DIR}
+        export SKIP='\[FLAKE\]|an InstallPlan step contains a deprecated resource version'
```
awgreene, line 89: There's a character limit which the SKIP variable exceeded if inline with the e2e command.

awgreene:
pj-rehearse ack

PR #33996, opened 2022-11-14
MGMT-11926: assisted-common-verify: bump memory limit

vrutkovs:
assisted-vsphere periodic jobs are being OOMKilled as it is running a lot more tests


omertuc:
Maybe we can do a similar thing as we did in https://github.com/openshift/release/pull/33827#issuecomment-1306983518 to know how much memory you need to request

vrutkovs:
That'd be useful, but there's no vsphere-periodic there - probably it has been red the whole time?

Just to be sure I synced it with `openshift-e2e-test` (used by vsphere IPI) - later on we can trim it

PR #33999, opened 2022-11-14
openstack: Delete metrics file if monitoring fails

pierreprinetti:
If the quota file can't be generated, delete the metrics file to signal Prometheus that the data is not available.


mdbooth (commented):
I'm not able to test, but assuming you have this

/lgtm

PR #34002, opened 2022-11-14
Adding assign_on label to get notified as soon as bral is added

kasturinarra:
Adding assign_on label to get notified as soon as the backport risk assessed label is added so that dev does not expliclty reach us asking for qe-cherry-pick-approved label also qe does not need to check on everyday if there is any fixes that are waiting for cherry-pick-approved label.

cc: @zhouying7780


kasturinarra:
@ingvagabund could you please help check and merge ? thanks !!

soltysh (approved):
/lgtm
/approve

PR #34004, opened 2022-11-14
Remove rhel6 job

fossedihelm:
Remove rhel6 since it will not be supported anymore on cnv 4.13
/cc @ksimon1


fossedihelm:
@ksimon1 jobs did not start :(

PR #34007, opened 2022-11-14
[kubevirt] replace azure instances with gcp instances with virtualization

ksimon1:
replace azure instances with gcp instances with virtualization
when running kubevirt on azure clusters, kubevirt VM shut down node on which it runs (probably due to kernel panic). To unblock upstream ci, we have to swith to gcp clusters.

Signed-off-by: Karel Šimon <ksimon@redhat.com>


dominikholler:
The bug motivating this code change is https://issues.redhat.com/browse/OCPBUGS-3641 .

PR #34013, opened 2022-11-14
Ensures push-images happens on post-submit

Stringy:
Also includes consolidation of push-image steps before integration test setup


Molter73 (approved):
/lgtm

PR #34036, opened 2022-11-15
cvo: make lgtm+approve respect gh reviews

petr-muller:
This configuration change will consider an approving GitHub review to have the same effect like `/lgtm` command (and also `/approve` but that is by default)


wking (approved):
I'm fine doing GitHub-approval+explicit-/lgtm, so I don't see much upside to this change, but we don't do a lot of stand-alone approval in the CVO, so I don't see a downside either.

/lgtm

PR #34037, opened 2022-11-15
hypershift: unify instance type and count for mgmt clusters

sjenning:
Instance type change from https://github.com/openshift/release/pull/33636 was not honored.  Mgmt cluster workers are still `m5.xlarge`.   I think it is because the step itself was overriding the `HYPERSHIFT_NODE_COUNT` and possibly clearing the `INSTANCE_TYPE` env var from the job config.

I think we want to just make 3x `m5.4xlarge` workers for mgmt clusters the default.

Also taking the opportunity to rename `INSTANCE_TYPE` to `HYPERSHIFT_INSTANCE_TYPE`.


sjenning:
Failed due to
```
2022-11-15 15:33:42+00:00 Clusteroperators not yet ready
version             False       True          3m52s   Unable to apply 4.12.0-rc.0: some cluster operators are not available
error: timed out waiting for the condition on clusterversions/version
```

It did get past the point of creating the mgmt cluster workers and it is still using 12 m5.xlarge instances.

PR #34055, opened 2022-11-16
Backport E2E tests to 4.8 and 4.9

liangxia:
Back port e2e tests from 4.10 to 4.9, removed those only apply to 4.10+
and back port e2e tests from 4.9 to 4.8 ( no removal as there are no profile marked as 4.9+)


liangxia:
/hold
since we are refining the p1/p2/p3 and f1/f2/f3/...

PR #34062, opened 2022-11-16
remove HYPERSHIFT_HC_RELEASE_IMAGE

LiangquanLi930:


heliubj18:
Developers just changed the default payload of hypershift hosted cluster to 4.12 rc.  remove the 4.11 spec limit to run 4.12 as expected.

PR #34066, opened 2022-11-16
Release 3.73 files

janisz:


gavin-stackrox (approved):
I would add some more OpenShift flavors + EKS. But approving as that can be handled in a separate PR when the release is stable.

PR #34071, opened 2022-11-16
Fix run_before_merge field

osherdp:
Applying the new field which will be autogenerated once https://github.com/openshift/ci-tools/pull/3139 gets in.


osherdp:
It should now be redundant, because we used ``omitempty`` on the ``run_before_merge`` field

PR #34085, opened 2022-11-16
[4.14] Consume job configs

jmguzik:


smg247:
Not sure where the list of deployments to change comes from, but it looks like `qe-private-deck` has been missed for quite awhile. Can you update https://github.com/openshift/release/blob/6d33a9313da5d6e76ad5afbdcc06a30b382435ce/clusters/app.ci/prow/03_deployment/qe_private_deck.yaml as well?

PR #34089, opened 2022-11-16
config/openshift/os: fix cosa image to use for Prow jobs

jlebon:
This is done correctly for 4.10 and 4.11 already. Adapt the newly branched 4.12, and fix 4.9 and older.


jlebon:
These are heavy tests. Since CI is broken anyway on those branches (except 4.12), I'd vote for skipping rehearsals and do more follow-up fixes if it still doesn't work.

/pj-rehearse skip

(Anyone, feel free to `reject` if you disagree.)

PR #34094, opened 2022-11-17
WIP Interop MTA Scenario Execution

calebevans:


ci-operator/step-registry/mtr/interop-scenario/mtr-interop-scenario-chain.yaml
```diff
@@ -0,0 +1,24 @@
+chain:
+  as: mtr-interop-scenario
+  steps:
+  - ref: interop-tooling-operator-install
+  - ref: mtr-orchestrate
+  - ref: mtr-execute
+  - ref: mtr-report
+  env:
```
ascerra, line 8: I don't think this is needed. You are setting these env vars in the config. That should be enough to make use of them in the operator-install script.

ascerra (commented):
Some general comments

- Can we add Miles (madunn) and Diane (dfrazzette) to the OWNERS file.
- Is it possible / what do you think about just using the operatorhub-subscribe ref as is for now in this PR?
- We need to discuss moving this away from your local fork and integrate into the windup_integration_test repo.
- We need to discuss the script in mtr-orchestrate-commands.sh. I don't want us to own this script, PQE should own and maintain this. I'm thinking it might be best to have this script in their test repo, and the container that we build from the dockerfile that will live in their repo as well will execute this script when started. So this orchestrate task will basically happening when the container gets created instead of a post creation shell script. Of course there's a lot we need to discuss here.

Perhaps we do some of these things in a separate PR so that we preserve this working PR.

ci-operator/step-registry/mtr/interop-scenario/mtr-interop-scenario-chain.yaml
calebevans: Yeah, I have a cleanup task set up. I am just trying to get past a few road blocks before I do some cleanup.

1. I will add them to the OWNERS files
2. I can give it a shot. Considering this is working for now, I don't want to make too many changes until I have a successful run
3. I will submit a PR to the windup_integration_test repo when I have successful runs
4. A lot of this needs to be in here. I will be changing it slightly to make it more sane

ci-operator/config/calebevans/windup_integration_test/README.md
```diff
@@ -0,0 +1,90 @@
...
+      SUB_INSTALL_NAMESPACE: mtr
+      SUB_PACKAGE: mtr-operator
+      SUB_SOURCE: redhat-operators
+      SUB_TARGET_NAMESPACES: mtr
+    test:
+    - chain: interop-mtr
+    workflow: cucushift-installer-rehearse-aws-ipi
+```
+
+### Cluster Provisioning and Deprovisioning: `cucushift-installer-rehearse-aws-ipi`
```
ascerra, line 51: Lines 51-72 roughly will be the same in each scenario that we create.

We could abstract this to a higher level piece of documentation then just add a reference to it here. 

I'm leaning towards adding the level of info directly to the workflows README (which doesn't exist right now)
https://github.com/openshift/release/tree/9a70897e1c01c7650df0e1b3dc188092cd600221/ci-operator/step-registry/cucushift/installer/rehearse/aws/ipi

ascerra: Or we could just direct people to the workflow doc that gets auto-created https://steps.ci.openshift.org/workflow/cucushift-installer-rehearse-aws-ipi

calebevans: Agreed - we can try to write our own documentation or reach out to the owners of that Workflow for some documentation.

ci-operator/config/calebevans/windup_integration_test/calebevans-windup_integration_test-mtr.yaml
```diff
@@ -0,0 +1,57 @@
...
+      memory: 200Mi
+tests:
+- as: mtr-scenario
+  steps:
+    cluster_profile: aws-cspi-qe
+    env:
+      BASE_DOMAIN: aws.interop.ccitredhat.com
+      SELENIUM_NAMESPACE: mtr
+      SUB_INSTALL_NAMESPACE: mtr
+      SUB_PACKAGE: mtr-operator
```
ascerra, line 48: Is there a way we can specify MTR version here? Is there a way you've been able to tell which version is being installed? From the artifacts I don't see any version data. [artifacts](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_release/34094/rehearse-34094-pull-ci-calebevans-windup_integration_test-mtr-mtr-scenario/1603892577847742464/artifacts/mtr-scenario/interop-tooling-operator-install/)

I believe since the update to MTR the operator is just at 1.0 but this will become important when we eventually do our expansion to 1.1

Here's an example of specifying the version for the ACM operator
[example](https://github.com/openshift/release/pull/34519/files#diff-e31942882c118aea60054d7a94d29110fd0eb96c45b748bd2cab50c9b49d87f9R38)

Though this method isn't ideal as it will make expansions a manual process ideally we can support multiple versions and support expansions automatically, this needs some brainstorming.

calebevans: Just putting this here for posterity, as we spoke about this in the Slack channel:

We can add another test that specifies which version of the operator to install, that way all required versions are tested:

```yaml
- as: mtr-scenario-1.0.0
  steps:
    cluster_profile: aws-cspi-qe
    env:
      BASE_DOMAIN: aws.interop.ccitredhat.com
      SELENIUM_NAMESPACE: mtr
      SUB_INSTALL_NAMESPACE: mtr
      SUB_PACKAGE: mtr-operator
      SUB_SOURCE: redhat-operators
      SUB_TARGET_NAMESPACES: mtr
      SUB_CHANNEL: stable-v1.0
    test:
    - chain: interop-mtr
- as: mtr-scenario-2.0.0
  steps:
    cluster_profile: aws-cspi-qe
    env:
      BASE_DOMAIN: aws.interop.ccitredhat.com
      SELENIUM_NAMESPACE: mtr
      SUB_INSTALL_NAMESPACE: mtr
      SUB_PACKAGE: mtr-operator
      SUB_SOURCE: redhat-operators
      SUB_TARGET_NAMESPACES: mtr
      SUB_CHANNEL: stable-v2.0
    test:
    - chain: interop-mtr
```

ci-operator/config/calebevans/windup_integration_test/README.md
calebevans: Updated in this commit: [9b549dc8ef74800853783297065b74950afe45f0](https://github.com/openshift/release/pull/34094/commits/9b549dc8ef74800853783297065b74950afe45f0)

ascerra:
@calebevans You should be able to update the config directory name now. You can use the org `config/windup/windup_integration_test` 

You'll need to get the owner of the repo to add the Openshift-ci-robot and openshift-merge-robot to the org. https://docs.ci.openshift.org/docs/how-tos/onboarding-a-new-component/#granting-robots-privileges-and-installing-the-github-app

ci-operator/step-registry/interop/mtr/OWNERS
```diff
@@ -0,0 +1,17 @@
+approvers:
```
ascerra, line 1: We should add Shveta to the approvers and reviewers list for the MTR specific files, including the config once we move it to the windup dir.

calebevans: Agreed, I have updated the org to the windup org rather than my own fork for the repository. I will add Shveta as an owner on the MTR specific files.

ci-operator/step-registry/interop/mtr/execute/README.md
```diff
@@ -0,0 +1,164 @@
...
+The following line of code will stop the local FTP server that was started earlier in the script. If the process is left running, the execute pod will not complete and OpenShift CI will stop the pod after 2 hours, failing the execution.
+
+```bash
+pkill -f ftp_server.py
+```
+
+## Container Used
+
+The container used in this step is named `mtr-runner` in the [configuration file](../../../../windup/windup-windup_integration_test-mtr.yaml). This container created from a custom image located in the [windup/windup_integration_test](https://github.com/windup/windup_integration_test.git) repository within in the `dockerfiles/interop` directory. The code snippet below is the Dockerfile found in that repository.
+
```
rnetser, line 101: Wouldn't a link to the file be better (so if changed, this doc will not need to be updated)?

ci-operator/step-registry/interop/mtr/execute/interop-mtr-execute-commands.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+# Set variables used to update the config file
```
rnetser, line 7: why not use jinja template for config files?

ci-operator/config/windup/windup_integration_test/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Orchestrate, Execute, and Report - `mtr-scenario`
+
+All of the orchestration, test execution, and reporting for the interop MTR scenario is taken care of the by the [`interop-mtr`](../../../step-registry/interop/mtr/README.md) chain. All of the environment variables needed to execute this chain are passed to the chain using the `env` stanza. For more in-depth information on how the [`interop-mtr`](../../../step-registry/interop/mtr/README.md) chain and it' components work, please see the README that is hyperlinked in this paragraph.
+
+## Custom Images
+
+### `mtr-runner`
+
+The `mtr-runner` image is a Python base image with all required packages for test execution installed along with the [windup/windup_integration_test](https://github.com/windup/windup_integration_test.git) repository copied into the `/tmp/integration_tests` directory. The image is used to execute the MTR interop tests.
```
rnetser, line 62: isn't the image stored in quay?

ci-operator/step-registry/interop/mtr/execute/README.md
```diff
@@ -0,0 +1,164 @@
...
+sed -i "s#REPLACE_FTP_PASSWORD#${FTP_PASSWORD}#" $CONFIG_FILE
+sed -i "s#REPLACE_EXECUTOR#${SELENIUM_EXECUTOR}#" $CONFIG_FILE
+sed -i "s#REPLACE_MTR_PASSWORD#${MTR_PASSWORD}#" $CONFIG_FILE
+sed -i "s#REPLACE_MTR_USER#${MTR_USERNAME}#" $CONFIG_FILE
+sed -i "s#REPLACE_MTR_VERSION#${MTR_VERSION}#" $CONFIG_FILE
+```
+
+### Install MTR Tests
+
+The following code snippet uses `pip` to install the MTR tests from the `/tmp/integration_tests` directory within the container. These tests come from the the [windup/windup_integration_test](https://github.com/windup/windup_integration_test.git) repository maintained by MTR product QE. These tests must be installed in this script rather than when the container image is built because OpenShift runs these containers using a user that ends up not having access to [modify the configuration file](#create-mtr-test-configuration-file) needed to execute these tests. Because that configuration file changes with every run, we have to modify it *then* install the tests.
```
rnetser, line 65: `These tests must be installed in this script rather than when the container image is built because OpenShift runs these containers using a user that ends up not having access to ....` - can't we have a service account with the appropriate credentials? otherwise we will end up implementing in custom scripts things that should be pre-configured in the image

ci-operator/step-registry/interop/tooling/deploy-selenium/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+To deploy a Selenium pod, along with the network infrastructure to use it from a different cluster, to the target test cluster. The container uses the [quay.io/redhatqe/selenium-standalone](https://quay.io/repository/redhatqe/selenium-standalone) image and allows us to use the container as a remote executor for Selenium tests.
+
+## Process
+
+This script can be separated into three parts: prepare, deploy, and verify.
+
+### Prepare
+
+#### Retrieve the console URL
```
rnetser, line 31: Can you please explain why the README files contain the code from the shell scripts?
I would suggest just adding a link (to avoid the need to change in multiple places)

ci-operator/step-registry/interop/mtr/execute/README.md
calebevans: See [90600e36f457961ad427db2012821f66007e76db](https://github.com/openshift/release/pull/34094/commits/90600e36f457961ad427db2012821f66007e76db)

ci-operator/step-registry/interop/mtr/execute/interop-mtr-execute-commands.sh
calebevans: I will consider that when this is refactored. PQE is re-factoring their tests into a new repository, so this may not be needed anymore.

ci-operator/config/windup/windup_integration_test/README.md
calebevans: No it isn't. It is built by Prow during execution. I will consider publishing any custom images in the re-factor.

ci-operator/step-registry/interop/mtr/execute/README.md
calebevans: It isn't a service account issue, it is an issue with the user that OpenShift runs the container with. I cannot modify the configuration file in this script if I run pip in the Dockerfile.

ci-operator/step-registry/interop/tooling/deploy-selenium/README.md
calebevans: See [90600e36f457961ad427db2012821f66007e76db](https://github.com/openshift/release/pull/34094/commits/90600e36f457961ad427db2012821f66007e76db)

ci-operator/step-registry/interop/mtr/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+The process for this scenario attempts to follow a general flow of orchestrate, execute, report.
+
+- **Orchestrate**: Install any necessary operators and additional infrastructure for testing.
+- **Execute**: Execute the tests from the test repository.
+- **Report**: Publish XUnit results to Report Portal or any other platform needed.
+
+### MTR Chain Steps
+
+Below is a list of steps and sub-steps executed within this chain. Each step is linked to the step's README file, if you'd like to learn more about a step, please follow the links to those pages. Retrieve
```
ascerra, line 28: I think that once we merge this code we will be able to replace this section with a link to the [auto-generated docs](https://steps.ci.openshift.org/) for this chain.

calebevans: Agreed, I'll keep this open and replace it when we are closer to merging.

calebevans:
/pj-rehearse periodic-ci-calebevans-windup-ui-tests-ocpci_interop-lp-interop-ocp4.12-mtr-interop-scenario

calebevans:
Closing due to refactor: https://github.com/openshift/release/pull/36181

PR #34101, opened 2022-11-17
Fix ssh/scp issue: unknown user 1004160000

yunjiang29:
ssh/scp command failed with error `unknown user 1004160000`.

reference: https://github.com/openshift/release/blob/master/ci-operator/step-registry/vsphere/disconnecting-network/recover/vsphere-disconnecting-network-recover-commands.sh#L10-L17


yunjiang29:
@jianlinliu PTAL

PR #34104, opened 2022-11-17
remove the ocm jobs from the 4.12 stable periodics list

yasun1:
The two jobs e2e-aws-ocm-osd-ccs-p2 and e2e-aws-rosa-sts-p2 are using a nightly build, and need to be removed from the stable periodics list.


yasun1:
@liangxia  Please help review. Thanks.

PR #34095, opened 2022-11-17
Add test frequency to job names

liangxia:
Based on latest name rules in [1], add test frequency to job names

Test frequency update will be in follow up PRs

[1] https://github.com/openshift/release/blob/master/ci-operator/config/openshift/openshift-tests-private/README.md

PR #34106, opened 2022-11-17
adjust test frequency for some jobs and remove duplicates ci coverage

jianlinliu:
Follow up https://github.com/openshift/release/pull/34095 and https://github.com/openshift/release/pull/34096, adjust test frequency for ci coverage.
1. Adjust test frequency for proposed 4.13 4.next install workflow, and update job name accordingly.
2. remove some duplicated coverage


jianlinliu:
cc @gpei @yunjiang29 and @liangxia  for review.

liangxia:
> * Adjust test frequency  for proposed 4.next install workflow
> * Update some e2e job name to p1 to match with our integration Test Configuration Matrix.
> * remove some duplicated coverage for 4.13

@jianlinliu  Perhaps split them into 3 smaller PRs, e.g, first on item1, then item3, item 2 to be the last one, as small and simple PRs are easier to review, and less conflicts.
Another reason is that in the old codes, the P1 has partial meaning of F1. If we get item 2 done without the other items done, then it will be hard to find those original P1 profiles and update its frequency if we missed several of them in the batch.

jianlinliu:
> the P1 has partial meaning of F1. If we get item 2 done without the other items done, then it will be hard to find those original P1 profiles and update its frequency if we missed several of them in the batch.

Thanks for your suggestion,  now shall I roll back the the job name change?

jianlinliu:
for iteam 1 and 3, I think they can be done together, the only purpose is to make get ci workload get down ASAP.

liangxia:
Manually triggered the job to upload existing test results into ReportPortal, so that the test results does not get lost due to the jobs rename.

/unhold

liangxia:
Fine to me to get item 1 and item 3 to done in one PR. Let's hold item 2 for a few more days to let team to act on item 1 and item 3. HDYT ?

jianlinliu:
let me roll back item 2 job name changing to make item 1 and 3 can be merged.

gpei:
@jianlinliu noticed we have some jobs defined with `f10/f12`(while most of the jobs are `f4/f7/f28`), just curious how the `10/12` comes

gpei:
Maybe we can also drop or decrease test frequency of some RHEL worker related profiles? 
```
- as: aws-ipi-proxy-sdn-workers-rhel8-p2-f7
- as: aws-ipi-proxy-workers-rhel8-etcd-encryption-p2-f7
- as: aws-ipi-proxy-workers-rhel8-p2-f7
- as: azure-ipi-proxy-workers-rhcos-rhel8-p2-f7
- as: azure-ipi-workers-rhel8-p2-f7
- as: gcp-ipi-disconnected-workers-rhel8-p2-f7
- as: gcp-ipi-workers-rhel8-fips-p2-f7
- as: gcp-ipi-workers-rhel8-p2-f7
- as: vsphere-ipi-proxy-workers-rhel8-p2-f7
- as: vsphere-upi-disconnected-sdn-workers-rhcos-rhel8-p2-f7
```
The above list is the RHEL related workflows got from 4.13 amd64 nightly, such as 
`- as: aws-ipi-proxy-workers-rhel8-p2-f7`
`- as: gcp-ipi-workers-rhel8-p2-f7`
I guess we can drop them since they're already covered by other workflow.

jianlinliu:
> @jianlinliu noticed we have some jobs defined with `f10/f12`(while most of the jobs are `f4/f7/f28`), just curious how the `10/12` comes

When updating the test frequency, I notice the priority like this:
GA features on heavily testing version (f2/f3/f4/f7) -> techpreview features on heavily testing version (f7)

So for techpreview features on released versions, I think, it should be less than f7, so I set "ibmcloud-ipi" in 4.10 (it is TP feature on 4.10) from f7 to f10. Consider Ibmcloud workflow only has a small number in ci, so I guess I can revert back to f7. Is that okay?

for `ipi-sdn-p1-f12`, it is a different case.
```
$ grep -r "as: .*ipi-sdn-p" *4.12* |grep -v vsphere |grep -v destructive|wc -l
8
```
Before it was on f7, per yesterday's meeting, if the same matrix was landed onto both arm and amd, the frequency should be multiple 2 times, so that we can keep the same ci jobs. e.g: aws-ipi-sdn, before it was on f7, ideally I think it should be f14, but I guess f14 is a bit more aggressive, so as a compromise, I set it to f10.

gpei:
> > @jianlinliu noticed we have some jobs defined with `f10/f12`(while most of the jobs are `f4/f7/f28`), just curious how the `10/12` comes
> 
> When updating the test frequency, I notice the priority like this: GA features on heavily testing version (f2/f3/f4/f7) -> techpreview features on heavily testing version (f7)
> 
> So for techpreview features on released versions, I think, it should be less than f7, so I set "ibmcloud-ipi" in 4.10 (it is TP feature on 4.10) from f7 to f10. Consider Ibmcloud workflow only has a small number in ci, so I guess I can revert back to f7. Is that okay?

yeah, I think that would be great, so that we can at least cover once during each z-stream weekly release.

> 
> for `ipi-sdn-p1-f12`, it is a different case.
> 
> ```
> $ grep -r "as: .*ipi-sdn-p" *4.12* |grep -v vsphere |grep -v destructive|wc -l
> 8
> ```
> 
> Before it was on f7, per yesterday's meeting, if the same matrix was landed onto both arm and amd, the frequency should be multiple 2 times, so that we can keep the same ci jobs. e.g: aws-ipi-sdn, before it was on f7, ideally I think it should be f14, but I guess f14 is a bit more aggressive, so as a compromise, I set it to f10.

for "sdn", I see we also have some other 3 workflows @f7 are covering it:
```
- as: aws-ipi-disconnected-private-sdn-workers-p2-f7
- as: aws-ipi-proxy-sdn-workers-rhel8-p2-f7
- as: aws-ipi-sdn-mtu-migrate-p3-f28
- as: aws-ipi-sdn-p1-f12
- as: azure-ipi-sdn-p1-f12
- as: gcp-ipi-sdn-p1-f12_
- as: gcp-upi-private-xpn-sdn-p2-f7
```
so for aws/azure/gcp-ipi-sdn-p1 actually, f14 is also good to me 😄

jianlinliu:
> so for aws/azure/gcp-ipi-sdn-p1 actually, f14 is also good to me

@zhaozhanqi do you have any comments on this?

jianlinliu:
> The above list is the RHEL related workflows got from 4.13 amd64 nightly, such as `- as: aws-ipi-proxy-workers-rhel8-p2-f7` `- as: gcp-ipi-workers-rhel8-p2-f7` I guess we can drop them since they're already covered by other workflow.

Good points.

zhaozhanqi:
> > so for aws/azure/gcp-ipi-sdn-p1 actually, f14 is also good to me
> 
> @zhaozhanqi do you have any comments on this?

So in summary:
for openshift-sdn or ipsec
nightly build-4.12  f7
multi night -4.12  f14
nighly bild-4.13  f4
right?

jianlinliu:
@liangxia @gpei pls help review again per your comments. I roll back item 2 for most of jobs, only updated some jobs which originally have wrong "p" setting.

zhaozhanqi:
> > > so for aws/azure/gcp-ipi-sdn-p1 actually, f14 is also good to me
> > 
> > 
> > @zhaozhanqi do you have any comments on this?
> 
> So in summary: for openshift-sdn or ipsec nightly build-4.12 f7 multi night -4.12 f14 nighly bild-4.13 f4 right?

For network related related features, we already sync in our team, the test policy can be: 

For sdn ovn, ipsec etc.  weekly is ok but better cover all platforms
For sdn-migration-ovn,  Two weeks are ok (one week for IPI, next week for UPI)

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.13__amd64-nightly.yaml
```diff
@@ -475,6 +475,17 @@ tests:
     test:
     - chain: openshift-e2e-test-qe-destructive
     workflow: cucushift-installer-rehearse-aws-ipi-ovn-ipsec
+- as: aws-ipi-ovn-ingress-nlb-p1-f4
```
liangxia, line 478: `aws-ipi-ovn-ingress-nlb` has been replace by `aws-ipi-ingress-lbtype-nlb` (as we also add `aws-ipi-ingress-lbtype-classic`), and we already have that in the file, so no need to add this one.

jianlinliu: ah, got it. I saw `aws-ipi-ovn-ingress-nlb` is also in arm nightly coverage, shall we replace it in arm ?

liangxia: Let arm take it, or use follow up PR to do that, so that we can get this one merge before the day off.

jianlinliu:
@zhaozhanqi In this PR, I gave up sdn workflow test frequency update, only degrade muliti-arch nightly, because are still in technical preview phase. If we can more details on multi-arch nightly, we can adjust their frequency again.


> the test policy can be:
> For sdn ovn, ipsec etc. weekly is ok but better cover all platforms
> For sdn-migration-ovn, Two weeks are ok (one week for IPI, next week for UPI)

For UPI, we had some discussion in https://coreos.slack.com/archives/GH9JHQFKN/p1668650517256529, that means, we will only have some very basic UPI install workflow for ci coverage, when covering `sdn-migration-ovn`, pls take into consideration, of course, if it does not care about disconnected network environment, basic UPI can cover them. That is okay for me. 

> one week for IPI, next week for UPI
Per my understanding, we are using cron job for the rotation, it is randomly picked by cron job, but not in order. So I guess for sdn-migration-ovn, `*/14` should be meet your requirement.

Thanks for the feedback about test policy for networking type, I guess your team can file another PR to adjust their test frequency. After that, let us see how the ci workload goes.

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.13__amd64-nightly.yaml
jianlinliu: fixed.

zhaozhanqi:
> @zhaozhanqi In this PR, I gave up sdn workflow test frequency update, only degrade muliti-arch nightly and some sdn workflow in 4.13, because muliti-arch feature is still in technical preview phase. If we can more details on multi-arch nightly, we can adjust their frequency again.
> 
> > the test policy can be:
> > For sdn ovn, ipsec etc. weekly is ok but better cover all platforms
> > For sdn-migration-ovn, Two weeks are ok (one week for IPI, next week for UPI)
> 
> For UPI, we had some discussion in https://coreos.slack.com/archives/GH9JHQFKN/p1668650517256529, that means, we will only have some very basic UPI install workflow for ci coverage, when covering `sdn-migration-ovn`, pls take into consideration, of course, if it does not care about disconnected network environment, basic UPI can cover them. That is okay for me.

yes, basic UPI install workflow is ok

> 
> > one week for IPI, next week for UPI
> > Per my understanding, we are using cron job for the rotation, it is randomly picked by cron job, but not in order. So I guess for sdn-migration-ovn, `*/14` should be meet your requirement.
> 
> Thanks for the feedback about test policy for networking type, I guess your team can file another PR to adjust their test frequency. After that, let us see how the ci workload goes.

Ok, sure

liangxia:
I'm going to approve the PR as it solves the quota issue.
For any team that feels that the test frequency needs more adjustment, feel free to open new PRs to get that update to what you want.

/lgtm

yunjiang29:
looks good to me.

PR #34107, opened 2022-11-17
[Prow CI chain Enhancement] update hypershift-aws-create to add configuration HYPERSHIFT_CP_AVAILABILITY_POLICY

LiangquanLi930:
1. now AWS hostedcluster chain Dynamic configuration `HYPERSHIFT_CP_AVAILABILITY_POLICY` is not supported, So have this pr

### DIFF
--- a/ci-operator/step-registry/cucushift/installer/rehearse/aws/ipi/ovn/hypershift/provision/cucushift-installer-rehearse-aws-ipi-ovn-hypershift-provision-chain.yaml
+++ b/ci-operator/step-registry/cucushift/installer/rehearse/aws/ipi/ovn/hypershift/provision/cucushift-installer-rehearse-aws-ipi-ovn-hypershift-provision-chain.yaml
@@ -11,5 +11,7 @@ chain:
     default: "true"
   - name: HYPERSHIFT_AWS_REGION
     default: "us-east-2"
+  - name: HYPERSHIFT_CP_AVAILABILITY_POLICY
+    default: "HighlyAvailable"
   documentation: |-
     Create an IPI cluster on AWS with OVN and hypershift for QE e2e tests.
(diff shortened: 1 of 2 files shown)


enxebre:
I'm curious where are we needing this?

LiangquanLi930:
> I'm curious where are we needing this?

@enxebre It is necessary for some OpenShift component testcases to run in HA mode. This PR is to set a new environment variable HYPERSHIFT_CP_AVAILABILITY_POLICY  to change the HyperShift cluster control plane mode. The default value is SingleReplica , and can be set to HighlyAvailable  to support HA testing scenario.

enxebre:
Thanks @Liangquan Li that explains "what" it does, which is fairly straight forward by looking at the code.
What I need as a reviewer and for anyone looking at the history at any point in time is the "Why" / "Context".
E.g. The QE team needs this for the effort to test the new set up dedicated to test X scenario which is not covered today...

Also, for this to be even more clear and ease assessment can you please include in the PR the code that consumes this?

LiangquanLi930:
> Thanks @LiangQuan Li that explains "what" it does, which is fairly straight forward by looking at the code. What I need as a reviewer and for anyone looking at the history at any point in time is the "Why" / "Context". E.g. The QE team needs this for the effort to test the new set up dedicated to test X scenario which is not covered today...
> 
> Also, for this to be even more clear and ease assessment can you please include in the PR the code that consumes this?

For openshift Component qe, we have many test scenarios that need to use HA, so we need to add this configuration, i think

LiangquanLi930:
> Also, for this to be even more clear and ease assessment can you please include in the PR the code that consumes this?

We need to open other PR, because we need qe team to review, so I am not in this PR consumes it

PR #34108, opened 2022-11-17
Use 4.13 repos and images for ironic-image in OCP 4.13

elfosardo:


elfosardo:
/hold
pending new repos check

PR #34128, opened 2022-11-18
[vsphere] remove unused steps for vsphere

jinyunma:
As we changed to use dynamic bastion host on vSphere as mirro registry/proxy server,  the steps for static ones will be not used any more, delete them.


jinyunma:
@jianlinliu @gpei @sgaoshang @WenXinWei PTAL, thanks.

PR #34141, opened 2022-11-18
MetalLB: pin the operator / metallb version to the right branch

fedepaol:
Both 4.11 and 4.12 were being pinned to main to deploy metallb from the operator repo and viceversa. This makes the CI deploy a component that might not be compatible with the parameters and / or viceversa.


fromanirh (commented):
/lgtm
/approve

PR #34144, opened 2022-11-18
OCPQE-12942: Invoke e2e test on Azure-HyperShift-guest

LiangquanLi930:


kuiwang02:
iamok from golang's view.

PR #34157, opened 2022-11-21
adjust ci coverage

jianlinliu:
1. remove `gcp-upi-private-xpn`, because it is already covered by `gcp-upi-private-xpn-ingress-glb` and `gcp-upi-private-xpn-sdn-p2`
2. Lower test frequency for `gcp-upi-xpn`, because it almost the same as `gcp-upi-private-xpn`
3. add `gcp-ipi-disconnected-private-sts` and `gcp-ipi-xpn-techpreview` ci coverage


jianlinliu:
cc @jianli-wei @gpei @liangxia for review

liangxia:
@jianlinliu I think you need to run `make jobs` to sync the changes between `config` and `jobs`.

jianlinliu:
> @jianlinliu I think you need to run `make jobs` to sync the changes between `config` and `jobs`.

Yeah, I forgot that. Should be fixed now.

PR #34160, opened 2022-11-21
supplemental-ci-images: add openconnect image

jinyunma:
Based on https://docs.ci.openshift.org/docs/how-tos/adding-a-cluster-profile/#image, add openconnect image for ASH wwt vpn connection.


jinyunma:
I tested openconnect docker image on my local host, it works and openconnect can be launched in container as expected.

@jianlinliu @bbguimaraes PTAL, thanks.

PR #34165, opened 2022-11-21
Adds benchmark support to stackrox/collector integration test workflow

Stringy:
Adds a new environment variable to the integration-test workflow which allows them to run the benchmarks. An alternative was to introduce an entire new workflow for the benchmarks but that felt like overkill for the outcome.

### DIFF
--- a/ci-operator/config/stackrox/collector/stackrox-collector-master.yaml
+++ b/ci-operator/config/stackrox/collector/stackrox-collector-master.yaml
@@ -523,6 +523,6 @@ tests:
       VM_TYPE: ubuntu-os
     workflow: stackrox-collector-update-support-packages
 zz_generated_metadata:
-  branch: giles-benchmark-ansible-playbook
+  branch: master
   org: stackrox
   repo: collector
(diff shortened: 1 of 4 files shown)


ci-operator/config/stackrox/collector/stackrox-collector-master.yaml
```diff
@@ -358,11 +358,21 @@ tests:
       VM_TYPE: fedora-coreos
     workflow: stackrox-collector-integration-tests
 - as: flatcar-integration-tests
-  postsubmit: true
```
Molter73, line 361: I believe this postsubmit should be left in, right? Or were we missing the presubmit for flatcar?

Stringy: yeah there were basically two post submit jobs for flatcar (the other being the more usual `merge-flatcar-integration-tests`) so removing this make it run on PRs (admittedly only when the all-integration-tests label is provided)

Molter73 (approved):
/lgtm

PR #34166, opened 2022-11-21
WIP: Add __okd-scos.yaml configs for some ironic components

sherine-k:
…er, ironic-ipa-downloader, ironic-image-downloader


sherine-k:
/cc @lmzuccarelli 
/assign @vrutkovs 
Hello both, would you mind taking a look? I'm trying to fix https://origin-release.ci.openshift.org/#4.13.0-0.okd-scos

lmzuccarelli:
@sherine-k - looks like the '4.14' jobs are failing - not found

sherine-k:
> the '4.14' jobs are failing - not found, maybe we should leave this out for now
The problem is that the `ironic-static-ip-manager` rehearse job needs all images to be available (`ironic-image-downloader` and `ironic-machine-os` images included). but the latter are built the other jobs...
@bradmwilliams Any idea how I can break free of this chicken and egg problem?

core-services/release-controller/_releases/release-ocp-4.13-ci.json
```diff
@@ -35,9 +35,6 @@
           "ironic-agent",
           "ironic-hardware-inventory-recorder",
           "ironic-inspector",
-          "ironic-ipa-downloader",
-          "ironic-machine-os-downloader",
-          "ironic-static-ip-manager",
```
LorbusChris, line 40: @sherine-k I think we need to keep `ironic-ipa-downloader` here? (it's built from https://github.com/openshift/ironic-ipa-downloader/blob/master/Dockerfile.okd)

And we probably want remove the other two from the FCOS mirroring down below too

LorbusChris: And I'd say let's not touch 4.10 and 4.11 configs here at all, but let's bring 4.12,4.13 and 4.14 into good shape first

LorbusChris: `stream8` is the replacement for OCP's ubi/rhel8 here. We could indeed consider switching to `stream9` indeed, since both FCOS and SCOS have new enough kernels. It's not required to do the switch now though

`ironic-hardware-inventory-recorder` has been removed from OCP, but it was still in CI and ended up in OKD releases that way, so we excluded from mirroring (see https://github.com/openshift/release/pull/33696). There's no need to build it anymore.

vrutkovs (approved):
/approve
/lgtm
/pj-rehearse ack

LorbusChris:
/hold
/retitle Add __okd-scos.yaml configs for some ironic components

sherine-k:
Closing in favor of a separate PR per component

PR #34179, opened 2022-11-21
Provide ARM instance type explicitly

yunjiang29:
Background:
* `m6a` was set as the default instance type family by #31684, this caused the `InsufficientInstanceCapacity m6a.xlarge capacity` issue, so #31828 was applied. 
* #31828 only works for public tests, since QE's tests are in the private deck, we need to find another way to fix this issue, a new #34059 was applied: provide instance type explicitly in each provision chain.
* #34059 introduced another issue:  since the default instance types are set in `-provision` chain, these default values are also used by **arm** cluster, this PR is trying to fix this issue.

there are some options for this fix:
1. provide instance type in both AMD and ARM job configurations explicitly, and remove default value from the provision chain.
2. provide instance type in ARM job configurations explicitly, AMD uses the default value in the provision chain. 
3. Same as 2), but reverse ARM and AMD.

I choose 2) in this PR because of the fewer code changes.

@aleskandro @lwan-wanglin @jianlinliu, any comments/ideas on this?

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.11__arm64-nightly.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.11__arm64-nightly.yaml
@@ -53,6 +53,8 @@ tests:
       OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE: release:arm64-latest
     env:
       BASE_DOMAIN: qe.devcluster.openshift.com
+      COMPUTE_NODE_TYPE: m6g.xlarge
+      CONTROL_PLANE_INSTANCE_TYPE: m6g.xlarge
       E2E_RUN_TAGS: '@aws-ipi and @arm64 and @network-ovnkubernetes and not @fips'
       OCP_ARCH: arm64
       TAG_VERSION: '@4.11'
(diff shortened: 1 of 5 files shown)


aleskandro:
Hi @yunjiang29 I'm ok with any of the solutions you propose.

It's good enough just to specify the instance types to use (as we do in flexy).

I'd say the 2nd proposal is the less risky to avoid changing too many x86 prow configs. 

Instead, having the x86 instance types explicitly called out is a "nice-to-have" for the others that create new profiles: they would see very similar configs in both the architectures' prow configs.

yunjiang29:
@aleskandro thanks for your feedback.

lwan-wanglin:
Does this influnce dev's jobs? they use` cluster_profile: aws-arm64`, do we need to explicitly set instance for dev's profiles?

yunjiang29:
> > Does this influnce dev's jobs? they use cluster_profile: aws-arm64 to represent arm clusters, do we need to explicitly set instance for dev's profiles too?
> 
> I guess no, if nothing set for `CONTROL_PLANE_INSTANCE_TYPE` and `COMPUTE_NODE_TYPE`, the same logic as before.

Correct. The types are added in job config level,  which is for QE only.

yunjiang29:
Some jobs' installation get passed, the change works. 
/pj-rehearse ack

@kuiwang02 can you approve this PR? thanks.

kuiwang02:
> Some jobs' installation get passed, the change works. /pj-rehearse ack
> 
> @kuiwang02 can you approve this PR? thanks.

please request qe-productivity team to approve it.
@liangxia @jhou1 please check it. thanks.

PR #34197, opened 2022-11-22
OCPQE-12966: use enable-qe-catalogsource-disconnected for private cluster

jianzhangbjz:
As the title, we should use the `enable-qe-catalogsource-disconnected` step for private clusters.


jianlinliu:
"private" cluster  just means users can not connect the cluster api outside vpc, the clusters itself still have internet connectivity as a common IPI install. So I do not think `we should use the enable-qe-catalogsource-disconnected step for private clusters`.

jianzhangbjz:
Got it, thanks! Closed it.

PR #34199, opened 2022-11-22
add workflow cucushift-installer-rehearse-azure-ipi-private-ovn

lihongan:
To address https://issues.redhat.com/browse/OCPQE-10942
This is only for 4.11 and before release, since OVN is default from 4.12.


lihongan:
@zhaozhanqi @anuragthehatter please help take a look, thanks.

PR #34226, opened 2022-11-22
[redhat-developer/odo] Require only one approval on PRs to be candidates for merging

rm3l:
This updates the list of labels needed on PRs in the [`odo` repository](https://github.com/redhat-developer/odo).
As agreed, we would now require only one approval on PRs to be candidates for merging.
The `lgtm` label makes more sense to keep, so we can also approve PRs right from the GitHub CLI or UI to have them merged.

### DIFF
--- a/core-services/prow/02_config/redhat-developer/odo/_pluginconfig.yaml
+++ b/core-services/prow/02_config/redhat-developer/odo/_pluginconfig.yaml
@@ -17,7 +17,6 @@ lgtm:
 plugins:
   redhat-developer/odo:
     plugins:
-    - approve
     - project
     - milestone
     - milestonestatus


valaparthvi:
> As agreed, we would now require only one approval on PRs to be candidates for merging.

Hi, Armel. I believe I missed this conversation, can you tell me why we're moving to single approval? Is it because of the number of people on team or some other reason?

rm3l:
> > As agreed, we would now require only one approval on PRs to be candidates for merging.
> 
> Hi, Armel. I believe I missed this conversation, can you tell me why we're moving to single approval? Is it because of the number of people on team or some other reason?

@valaparthvi Yes, mostly because of the number of people on the team. With our reduced capacity, we discussed about the need to iterate faster, and not necessarily have to wait too many days until a PR is merged in.
Reviewers will have more responsibility, but I think it should be fine.

rm3l:
@kadel Can you also help review this PR? To confirm that I did not miss any other configuration file. Thanks.

PR #34229, opened 2022-11-22
Use ubi-minimal for central-db

janisz:
https://github.com/stackrox/stackrox/pull/3880

### DIFF
--- a/ci-operator/config/stackrox/stackrox/stackrox-stackrox-release-3.73.yaml
+++ b/ci-operator/config/stackrox/stackrox/stackrox-stackrox-release-3.73.yaml
@@ -53,7 +53,7 @@ images:
   - name: ROX_PRODUCT_BRANDING
     value: RHACS_BRANDING
   dockerfile_path: Dockerfile.gen
-  from: ubi
+  from: ubi-minimal
   inputs:
     central-db-bundle:
       paths:
--- a/ci-operator/config/stackrox/stackrox/stackrox-stackrox-master.yaml
+++ b/ci-operator/config/stackrox/stackrox/stackrox-stackrox-master.yaml
@@ -69,7 +69,7 @@ images:
   - name: ROX_PRODUCT_BRANDING
     value: RHACS_BRANDING
   dockerfile_path: Dockerfile.gen
-  from: ubi
+  from: ubi-minimal
   inputs:
     central-db-bundle:
       paths:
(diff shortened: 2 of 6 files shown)


janisz:
/pj-rehearse ci/rehearse/stackrox/stackrox/remove_python_from_central-db/images

ci-operator/config/stackrox/stackrox/stackrox-stackrox-release-3.73.yaml
```diff
@@ -53,7 +53,7 @@ images:
   - name: ROX_PRODUCT_BRANDING
     value: RHACS_BRANDING
   dockerfile_path: Dockerfile.gen
-  from: ubi
+  from: ubi-minimal
```
RTann, line 56: since 73's code freeze has already begun, I think we should wait until 74 for this

janisz: This PR is mostly for 3.73 as we have CVS for that image and I'd like to fix them before the release

ci-operator/config/stackrox/stackrox/stackrox-stackrox-remove_python_from_central-db.yaml
```diff
@@ -656,6 +656,6 @@ tests:
     test:
     - ref: stackrox-initial
 zz_generated_metadata:
-  branch: master
```
mtodor, line 659: As I understand, this is needed only to test https://github.com/stackrox/stackrox/pull/3880.

Would it be possible to keep the configuration for the `master` branch and just add the configuration for the `remove_python_from_central` branch? So that we have both in parallel. I assume that `master` branch builds will be broken with this change until we revert this PR?

Also, it may be good to add support for the new branch and leave everything else as it is (`nightlies`, `3.73`, etc.). And then, when we see that CI works and the stackrox branch can be merged, we can prepare a new PR on CI that will fix `nightlies`, `3.73`, and also remove support for the new branch. After that, we can merge both PRs in a short time to avoid possible broken CI.

janisz: So I tried that before. To keep master config and this branch but then I have problems with image "promotions" and finally for some reason images were not build for branch rehearsals.
Anyway having both configurations has no benefit for us. After merging to master CI on branch should pass as its targeting master.

PR #34240, opened 2022-11-23
refine ci coverage

jianlinliu:
In lower versions (< 4.12), sdn is the default networking type, so remove the duplicated sdn ci jobs.


jianlinliu:
cc @liangxia @zhaozhanqi for review

PR #32685, opened 2022-09-28
MGMT-12167: Add ZTP job to assisted-installer repo

adriengentil:
- Assisted: run ztp job when controller pod definition changes
- assisted-baremetal-operator-setup: allow redefining ci-index image

PR #34253, opened 2022-11-23
MGMT-12652: Add temporary image streams for ocm-2.8 Assisted Installer migration

eliorerz:
Preparation for the ocm 2.8 FF

/cc @osherdp @gamli75

### DIFF
--- a/clusters/app.ci/supplemental-ci-images/assisted-ci-tools/temp-images-for-migration.yaml
+++ b/clusters/app.ci/supplemental-ci-images/assisted-ci-tools/temp-images-for-migration.yaml
@@ -40,15 +40,15 @@ spec:
       kind: ImageStreamTag
       name: ocm-2.7:cluster-api-provider-agent
       namespace: edge-infrastructure
+  - name: assisted-test-infra-internal
+    from:
+      kind: ImageStreamTag
+      name: ocm-2.7:assisted-test-infra-internal
+      namespace: edge-infrastructure
   - name: assisted-service-scripts
     from:
       kind: ImageStreamTag
       name: ocm-2.7:assisted-service-scripts
-      namespace: edge-infrastructure
-  - name: assisted-service-index
-    from:
-      kind: ImageStreamTag
-      name: ocm-2.7:assisted-service-index
       namespace: edge-infrastructure
 ---
 kind: ImageStream


clusters/app.ci/supplemental-ci-images/assisted-ci-tools/temp-images-for-migration.yaml
```diff
@@ -1,5 +1,58 @@
...
+    from:
+      kind: ImageStreamTag
+      name: ocm-2.7:assisted-test-infra
+      namespace: edge-infrastructure
+  - name: cluster-api-provider-agent
+    from:
+      kind: ImageStreamTag
+      name: ocm-2.7:cluster-api-provider-agent
+      namespace: edge-infrastructure
+  - name: assisted-service-operator-index
```
osherdp, line 43: I don't think we have this image anymore
We have ``assisted-service-index`` which should change its target name imho:
https://github.com/openshift/release/pull/32685#discussion_r982635167

cc @adriengentil

osherdp (changes requested):

clusters/app.ci/supplemental-ci-images/assisted-ci-tools/temp-images-for-migration.yaml
adriengentil: yes, we can get rid of it.
We'll need to mirror the index image once `ocm-2.7` tag is pushed here: https://github.com/openshift/release/blob/master/core-services/image-mirroring/supplemental-ci-images/mapping_supplemental_ci_images_edge_infrastructure

clusters/app.ci/supplemental-ci-images/assisted-ci-tools/temp-images-for-migration.yaml
```diff
@@ -1,5 +1,58 @@
...
+    from:
+      kind: ImageStreamTag
+      name: ocm-2.7:assisted-service-scripts
+      namespace: edge-infrastructure
+  - name: assisted-service-index
+    from:
+      kind: ImageStreamTag
+      name: ocm-2.7:assisted-service-index
+      namespace: edge-infrastructure
+---
```
osherdp, line 53: One more thing, we're gonna need ``assisted-test-infra-internal`` as well, because it's being used in nutanix&vsphere jobs:
https://github.com/openshift/release/blob/b96d29219b85d71197fe9676e64fc4823984921a/ci-operator/config/openshift/assisted-service/openshift-assisted-service-master__edge.yaml#L22-L25

https://github.com/openshift/release/blob/0896191a7c5ec03f0bbb9038bc1fe8ac14f887a2/ci-operator/step-registry/assisted/vsphere/setup/configure/assisted-vsphere-setup-configure-ref.yaml#L3

osherdp (changes requested):

osherdp (approved):

PR #34256, opened 2022-11-23
MGMT-12653: Update the Bugzilla plugin configuration for all the assisted repositories

eliorerz:
Update the Bugzilla plugin configuration for all the assisted repositories repositories to include the release branches 2.7, 2.8 and 2.9.

/cc @osherdp @gamli75

### DIFF
--- a/core-services/prow/02_config/openshift/assisted-image-service/_pluginconfig.yaml
+++ b/core-services/prow/02_config/openshift/assisted-image-service/_pluginconfig.yaml
@@ -10,7 +10,7 @@ bugzilla:
         assisted-image-service:
           branches:
             main:
-              target_release: rhacm-2.7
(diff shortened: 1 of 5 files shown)


osherdp (changes requested):
iiuc this part is redundant (at least for the most part) because they decided to move to Jira in ACM
so there will be no BZ tickets created for ACM (maybe we'll even remove this configuration for the old branches)

As for the Jira alternative, it's still not clear if it will be utilized (cc @filanov)

filanov:
@osherdp is correct, ACM moved to use jira, because the jira plugin is working only with OCPBUGS project we cannot use it.
Maybe instead we should have a list of people that can approve backports

osherdp:
So not needed anymore
/close

PR #34259, opened 2022-11-23
[Baremetal QE] Adds load balancer steps

jadhaj:
This is a draft PR for configuring load balancer steps for the Baremetal @ rdu2 OCP tests.

### DIFF
--- a/ci-operator/step-registry/baremetal/lab/pre/load-balancer/baremetal-lab-pre-load-balancer-commands.sh
+++ b/ci-operator/step-registry/baremetal/lab/pre/load-balancer/baremetal-lab-pre-load-balancer-commands.sh
@@ -238,7 +238,7 @@ api_vip: $api_ip
 "
 echo $vips > external_vips.yaml
 
-docker restart "haproxy-$BUILD_ID"
+docker kill --signal HUP "haproxy-$BUILD_ID"
 
 EOF
 


ci-operator/step-registry/baremetal/lab/pre/load-balancer/baremetal-lab-pre-load-balancer-commands.sh
```diff
@@ -1,2 +1,202 @@
...
+listen api-server-6443
+    bind *:6443
+    mode tcp
+$(
+if [ "${IPI}" != "true" ]; then
+  for bmhost in $(yq e -o=j -I=0 '.[]' "${SHARED_DIR}/hosts.yaml"); do
+    # shellcheck disable=SC1090
+    . <(echo "$bmhost" | yq e 'to_entries | .[] | (.key + "=\"" + .value + "\"")')
+    # shellcheck disable=SC2154
+    if [[ $name == *master* ]]; then
```
aleskandro, line 70: we can change the query with yq instead of using "linear search"

ci-operator/step-registry/baremetal/lab/pre/load-balancer/baremetal-lab-pre-load-balancer-commands.sh
```diff
@@ -1,2 +1,202 @@
-#!/bin/sh
-echo Hello world
+#!/bin/bash
+
+if [ -n "${LOCAL_TEST}" ]; then
+  # Setting LOCAL_TEST to any value will allow testing this script with default values against the ARM64 bastion @ RDU2
+  # shellcheck disable=SC2155
+  export NAMESPACE=test-ci-op AUX_HOST=openshift-qe-bastion.arm.eng.rdu2.redhat.com \
+      SHARED_DIR=${SHARED_DIR:-$(mktemp -d)} CLUSTER_PROFILE_DIR=~/.ssh IPI=false SELF_MANAGED_NETWORK=true \
+      INTERNAL_NET_IP=192.168.90.1
```
aleskandro, line 8: only set the needed variables here.

ci-operator/step-registry/baremetal/lab/pre/load-balancer/baremetal-lab-pre-load-balancer-commands.sh
```diff
@@ -1,2 +1,202 @@
...
+IPI="${2}"
+set +o allexport
+
+# shellcheck disable=SC2174
+mkdir -m 755 -p "/var/builds/$BUILD_ID/haproxy"
+EOF
+
+echo "Uploading the haproxy.cfg file to the auxiliary host..."
+scp "${SSHOPTS[@]}" "${CONF_DIR}/haproxy.cfg" "root@${AUX_HOST}:/var/builds/$NAMESPACE/haproxy"
+scp "${SSHOPTS[@]}" "${CONF_DIR}/dhclient.conf" "root@${AUX_HOST}:/var/builds/$NAMESPACE/haproxy"
```
aleskandro, line 202: I suppose you are still working on the other steps that needs porting here (docker commands and similar).

I would suggest reducing both the number of ssh connections we make and minimizing the amount of remote instructions. Try to make a tradeoff and balance as most as you can.

Two scp will need twice the time than 1 copying both the (small) files and, with the ssh connection earlier to create a directory, we need three times the minimum time needed to reach the server once  and (1) create the directory, (2) push the two files, e.g., from env variables passed via arguments.

jadhaj: Sure, the rollback will be in this PR

As for minimizing scp/ssh connections, WDYT about:

Moving the logic under one ssh connections by using multiple heredocs (EOFs).
EOF will be the main one. Nested heredocs to be added as needed like EOF!, EOF2 and so on 

This way, only 1 ssh connections Is needed and 1 scp for copying hosts.yaml to target machine

I have testes nested heredocs and it seems to be working.

aleskandro: Multiple heredocs are a bit difficult in terms of readability. I'd also keep on having as most logic as possible executing on the pods, rather than on the remote host.

What I'm doing is (1) filling a variable rather than a file and (2) passing the content of the variable as an argument to the remote script.

> 1 scp for copying hosts.yaml to target machine

The hosts.yaml file is expected to be in the /var/builds/$BUILD_ID folder right after the execution of the reserve steps.

jadhaj: For the sake of readability, heredocs can be renamed as needed

>  I'd also keep on having as most logic as possible executing on the pods, rather than on the remote host.

Sorry, i did not got this one. commands are either execute on Jenkins agent or remote host. do you mean Jenkins pod ?


As for passing a var as an arg, substitution inside vat for below code yields errors:

$(
for bmhost in $(yq e -o=j -I=0 '.[] | select(.name == "master*")' hosts.yaml); do
  # shellcheck disable=SC1090
  . <(echo "$bmhost" | yq e 'to_entries | .[] | (.key + "=\"" + .value + "\"")')
  # shellcheck disable=SC2154
  echo "    server $name $ip:6443 check inter 1s"
done
)

aleskandro: They are ocp pods running in the prow build farm ocp clusters, created via the ci-operator.

Get a look at the dhcp pxe step for reference about a for loop.

ci-operator/step-registry/baremetal/lab/pre/load-balancer/baremetal-lab-pre-load-balancer-commands.sh
```diff
@@ -1,2 +1,245 @@
...
+api_ip=nsenter -m -u -n -i -p -t $(docker inspect -f '{{ '{{' }}.State.Pid {{ '}}' }}' haproxy-{{ BUILD_ID }}) -n  \
+/sbin/ip -o -4 a list eth1 | sed 's/.*inet \(.*\)\/[0-9]* brd.*$/\1/'
+
+vips="
+ingress_vip: $api_ip
+api_vip: $api_ip
+"
+echo $vips > external_vips.yaml
+
+docker restart "haproxy-$BUILD_ID"
```
aleskandro, line 241: We don't have to restart the container, but sending sighup to it:

```
docker kill --signal HUP "haproxy-$BUILD_ID"
```

aleskandro:
based on our previous discussion i'm unholding and merging

/unhold

/lgtm

aleskandro:
Refers OCPQE-1181

aleskandro:
Refers OCPQE-11881

PR #34260, opened 2022-11-23
OKD: ironic-agent-image okd-scos specific builds for master, >4.12

sherine-k:
This PR is a partial replacement for https://github.com/openshift/release/pull/34166 for ironic-image

### DIFF
--- a/ci-operator/config/openshift/ironic-agent-image/openshift-ironic-agent-image-main__okd-scos.yaml
+++ b/ci-operator/config/openshift/ironic-agent-image/openshift-ironic-agent-image-main__okd-scos.yaml
@@ -10,17 +10,17 @@ images:
   from: centos
   to: ironic-agent
 promotion:
-  name: "4.13"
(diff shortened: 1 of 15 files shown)


sherine-k:
> I strongly suggest to use stream9, all the packages used come from cs9/rhel9 and this could break without notice ironic upstream has removed support for python3.6 and CS8 since the beginning of the year, we removed support in the ironic openshift containers right after

Thanks @elfosardo. I was wondering if I should make the change for okd-scos and okd in this PR? or just okd-scos?
```
  centos:
    name: stream9
    namespace: coreos
    tag: "9"
```

elfosardo:
@sherine-k better change all to CS9 IMHO, at some point you'll have to change it anyway

elfosardo:
/pj-rehearse ack
missing image origin/scos-4.14, should be addressed to ARt, but it's early anyway
it looks ok

PR #34265, opened 2022-11-23
OKD: Remove ironic-ipa-downloader configs from 4.12+ branches

sherine-k:


LorbusChris:
/retitle OKD: Remove ironic-ipa-downloader configs from 4.12+ branches
/hold cancel
/lgtm
/assign @dhellmann @elfosardo

sherine-k:
Hi @dhellmann 
Do you think this PR can be merged?
Thanks

dhellmann:
@sherine-k & @LorbusChris, I'm not sure I have enough background about what's going on here. We don't need these images any more?

/cc @dtantsur @zaneb for their visibility

zaneb:
> I'm not sure I have enough background about what's going on here. We don't need these jobs any more?

I think that's correct, in metal IPI we now run CoreOS and pull the agent as a container image from the payload, so we no longer need a container with a RHEL image containing IPA. I'll leave for @elfosardo to confirm, but it is safe to approve the image-mirroring changes.

dhellmann:
> > I'm not sure I have enough background about what's going on here. We don't need these jobs any more?
> 
> I think that's correct, in metal IPI we now run CoreOS and pull the agent as a container image from the payload, so we no longer need a container with a RHEL image containing IPA. I'll leave for @elfosardo to confirm, but it is safe to approve the image-mirroring changes.

Thanks. I'll leave the lgtm to one of you.

/approve

PR #34268, opened 2022-11-23
HyperShift: set resource request override for kube apiserver

csrwng:
This will improve scaling of nodes for hosted management clusters created for CI


enxebre:
/lgtm
/hold
to see test passing.

PR #34270, opened 2022-11-23
Enable custom network MTU on AWS Local Zone

yunjiang29:
Resolve https://issues.redhat.com/browse/OCPBUGS-3910


yunjiang29:
@liangxia @gpei @jianlinliu @zhaozhanqi PTAL, all installs get passed

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-nightly-4.12-upgrade-from-stable-4.12.yaml
```diff
@@ -104,7 +104,7 @@ tests:
     test:
     - ref: cucushift-installer-check
     - chain: openshift-upgrade-qe-test
-    workflow: cucushift-installer-rehearse-aws-ipi-localzone
+    workflow: cucushift-installer-rehearse-aws-ipi-ovn-localzone
```
liangxia, line 107: The test name `aws-ipi-localzone-p2-f7` need to reflect the change, too, as we use that to identify which profile we are testing in ReportPortal.

yunjiang29: I think that ovn is the default network for 4.12+,  so the test name should be ok without the network type.

ci-operator/step-registry/cucushift/installer/rehearse/aws/ipi/ovn/localzone/provision/cucushift-installer-rehearse-aws-ipi-ovn-localzone-provision-chain.yaml
```diff
@@ -10,6 +10,8 @@ chain:
   - ref: ipi-conf-telemetry
   - ref: ipi-conf-aws-custom-vpc
   - ref: ipi-conf-aws
+  - ref: ovn-conf
```
jianlinliu, line 13: Per the logic of `ipi-conf-network-mtu`, sounds like we do not need the step.

jianlinliu: Or does that mean only OVN need to set custom MTU?

gpei: Per our testing result, OVN will need NETWORK_MTU:1100, SDN may need a different MTU value. Considering we're testing the localzone in 4.12+ and it's using OVN as default network plugin, I think we can only test  `aws-ipi-localzone-provision` with OVN

jianlinliu: Got it. Per the new step - `ipi-conf-network-mtu` naming, I guess we can reuse it for other cases in the future. So my suggestion is:
1. removing `ovn-conf` step in this workflow, because we start localzone testing since 4.12, its default networkType is ovn.
2. rename the chain name to `cucushift-installer-rehearse-aws-ipi-network-mtu-localzone-provision`

Or if we want to explicitly show networking type in the step, maybe we can rename the `ipi-conf-network-mtu` to `ipi-conf-ovn-mtu`, and update the chain name accordingly.

yunjiang29: Right, the reason is the MTU value is different for OVN and SDN, the ovn/mtu-1100 is the pair in this workflow.

> Considering we're testing the localzone in 4.12+ and it's using OVN as default network plugin, I think we can only test `aws-ipi-localzone-provision` with OVN

Yes, AWS Local Zone is supported by 4.12+.

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-nightly-4.12-upgrade-from-stable-4.12.yaml
liangxia: There are changes from version to version.
Prefer let the name reflect what kind of tests it is. 
It's clear that we know what's changed from 4.11 to 4.12 because we are heavily on 4.12. How about several release later ? E.g, can you recall all the changes we have from 4.7 to 4.8 ?

ci-operator/step-registry/cucushift/installer/rehearse/aws/ipi/ovn/localzone/provision/cucushift-installer-rehearse-aws-ipi-ovn-localzone-provision-chain.yaml
zhaozhanqi: >Or does that mean only OVN need to set custom MTU?
Actually this issue is not related openshift-sdn or OVN,  it's about the host from AZ to local zone only support MTU is 1300 in our testing 

```
sh-4.4# ping -M do -s 1300 -c 5 10.0.54.121
PING 10.0.54.121 (10.0.54.121) 1300(1328) bytes of data.
ping: local error: Message too long, mtu=1300
ping: local error: Message too long, mtu=1300
ping: local error: Message too long, mtu=1300
ping: local error: Message too long, mtu=1300
ping: local error: Message too long, mtu=1300

sh-4.4# ping -M do -s 1200 -c 5 10.0.54.121
PING 10.0.54.121 (10.0.54.121) 1200(1228) bytes of data.
1208 bytes from 10.0.54.121: icmp_seq=1 ttl=64 time=27.0 ms
1208 bytes from 10.0.54.121: icmp_seq=2 ttl=64 time=26.3 ms
1208 bytes from 10.0.54.121: icmp_seq=3 ttl=64 time=26.4 ms
1208 bytes from 10.0.54.121: icmp_seq=4 ttl=64 time=26.4 ms
1208 bytes from 10.0.54.121: icmp_seq=5 ttl=64 time=26.3 ms
```
So here whatever openshift-sdn or ovn,  MTU can be set 1100 I guess. 
even though 1150 also ok for openshift-sdn

zhaozhanqi: >Or if we want to explicitly show networking type in the step, maybe we can rename the ipi-conf-network-mtu to ipi-conf-ovn-mtu, and removing the sdn networking type judgement and codes, and update the chain name accordingly.

My suggestion is:  How about keeping the `sdn networking type judgement` here. if want to test sdn plugin with Local zone in future. So we just add another chain with step `ref: sdn-conf`

jianlinliu: I am okay with keep the snd netowrking type judgement, if that, does it make sense?
```
1. removing ovn-conf step in this workflow, because we start localzone testing since 4.12, its default networkType is ovn.
2. rename the chain name to cucushift-installer-rehearse-aws-ipi-network-mtu-localzone-provision
```

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-nightly-4.12-upgrade-from-stable-4.12.yaml
yunjiang29: Get it, will update it to reflect workflow configuration.

ci-operator/step-registry/cucushift/installer/rehearse/aws/ipi/ovn/localzone/provision/cucushift-installer-rehearse-aws-ipi-ovn-localzone-provision-chain.yaml
yunjiang29: > even though 1150 also ok for openshift-sdn

Per documentation's perspective, we need to provide a maximum MTU value (the suer setting must not be larger than this), so MTU values for SDN and OVN are different, and CI needs to match the value in the doc.

> 1. removing ovn-conf step in this workflow, because we start localzone testing since 4.12, its default networkType is ovn.
2. rename the chain name to cucushift-installer-rehearse-aws-ipi-network-mtu-localzone-provision

If we don't test SDN+LocalZone, I agree with this solution.

@jianlinliu @zhaozhanqi WDTY ^^ ?

jianlinliu: The issue was found when we running local zone testing, we have to adjust MTU, so personally I prefer to focus on localzone testing, but not focus on sdn/ovn + localzone. So I suggested  `cucushift-installer-rehearse-aws-ipi-network-mtu-localzone-provision`

yunjiang29: No objections,  @zhaozhanqi is it good for you?

zhaozhanqi: yes, it's ok for me

yunjiang29:
@jianlinliu @liangxia @zhaozhanqi PTAL.

PR #34274, opened 2022-11-23
MGMT-11803: Use globally defined slack reporting in a few repos

osherdp:
Now that https://github.com/openshift/ci-tools/pull/3140 got in we should be able to define those alerts globally (instead of having to manually edit the relevant jobs after their creation).

This only changes it for assisted-installer-deployment and cluster-api-provider-agent to test the new feature (before applying it on the more heavily-configured assisted-test-infra and assisted-service. and of course all the rest).

/cc @eranco74 @adriengentil


adriengentil (approved):
:+1:

PR #33398, opened 2022-10-24
AGENT-290: Add base agent required job to Assisted repositories

lranjbar:
As the Agent team has discussed, we need to have one required job in the Assisted repositories to reduce regressions. We identified the "e2e-agent-compact" as our base passing job to add to the repositories.

This updates things from 4.11 to 4.12 on `master` and `release-4.12`. 

In 4.12 agent-installer is considered GA so we will need to avoid regressions for this flow.

[AGENT-290](https://issues.redhat.com//browse/AGENT-290)

PR #34280, opened 2022-11-23
AGENT-290: Always run e2e-agent job in assisted-service

zaneb:
This was the intended outcome of https://github.com/openshift/release/pull/33398, but removing the `run_if_changed`
directive from the config unfortunately does not result in `always_run`
being switched to true in the job if it already exists, even though that
would be the default if the job didn't already exist.


osherdp (approved):

osherdp:
``job pull-ci-openshift-assisted-service-release-4.13-e2e-agent-compact is set to always run but also declares run_if_changed targets, which are mutually exclusive``

PR #34296, opened 2022-11-24
OCPQE-13009: Fix test frequency for several destructive jobs

liangxia:
/cc @jianlinliu @yunjiang29 @jhou1 @JianLi-RH


jianlinliu:
good catch.

/lgtm

PR #34315, opened 2022-11-25
Fix cron entries for multi jump upgrade jobs

liangxia:
/hold

Wait for the multi jump upgrade to finish and import the results to RP, then we can merge this PR.


liangxia:
Those upgrade test results has been imported into RP.

PR #34313, opened 2022-11-25
[WIP]move image

kuiwang02:
/cc @Xia-Zhao-rh

PR #34319, opened 2022-11-25
[CIGolang]remove image

kuiwang02:
Following https://github.com/openshift/release/pull/34193, please read the comments
https://github.com/openshift/release/pull/34193#issuecomment-1327042640
https://github.com/openshift/release/pull/34193#issuecomment-1327152533
https://github.com/openshift/release/pull/34193#issuecomment-1327167266
https://github.com/openshift/release/pull/34193#issuecomment-1327187247

the image has nothing with the code of the test-private repo.  it will bring unstable of merging PR.
and the image will not work for step because 
1, no promotion which is not allowed in test-private repo.
2, the test definition and image built definition is in different config files.

the best way is to enhance dev-scripts image. if still needing more one image, should put it in installer image, like https://github.com/openshift/installer/blob/master/images/libvirt/Dockerfile.ci.

another solution which is not good is to put it in verification-tests, shown in https://github.com/openshift/release/pull/34313 although it also has nothing with code of that repo. at least it works because the image could be promoted.
So, you could be temp solution.

cc @aleskandro @jhou1 

/cc @jianzhangbjz

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-master.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-master.yaml
@@ -30,27 +30,6 @@ images:
       - registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.19-openshift-4.12
   to: tests-private
 - dockerfile_literal: |
-    # This Dockerfile builds the image used by baremetal qe installation tests
-    FROM registry.ci.openshift.org/ocp/4.12:cli
-    ENV HOME /output
-    RUN INSTALL_PKGS="ansible openssh rsync wget bc file findutils git hostname lsof make \
-          socat tar tree util-linux wget which zip curl gawk iputils jq iproute ca-certificates \
-          nmap-ncat socat mtr bash net-tools bind-utils iperf3 tcpdump procps conntrack-tools \
-          iputils ethtool strace ipmitool" && \
-      mkdir -p ${HOME} && \
-      yum install --setopt=install_weak_deps=False --nodocs -y $INSTALL_PKGS && \
-      yum clean all && \
-      rm -rf /var/cache/yum/*
-    RUN wget -O /usr/local/bin/yq \
-      https://github.com/mikefarah/yq/releases/download/v4.30.4/yq_linux_$(uname -m \
-        | sed 's/x86_64/amd64/;s/aarch64/arm64/') && \
-      ansible-galaxy collection install -vvvv community.docker community.general && \
-      chmod -R g=u "${HOME}/.ansible/" && \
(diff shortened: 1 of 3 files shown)


jianzhangbjz:
Thanks! And, I guess this `# This Dockerfile builds the image used by baremetal qe installation tests` image is nothing with our private test code, so here's not a good place for it.
/lgtm

aleskandro:
Again, the image is not related to devscripts at all, and it was a temporary place for speeding up the baremetal@rdu2 prow integration (only for QEs and openshift-tests-private configs).

You just faced network issues, the image-building time is not impacted as the images build concurrently, and this image just required less time to build than others.

We'll find another place.

Thanks for understanding and removing the code without caring about others' opinion.

cc @YuLi517

kuiwang02:
> Again, the image is not related to devscripts at all, and it was a temporary place for speeding up the baremetal@rdu2 prow integration (only for QEs and openshift-tests-private configs).
> 
> You just faced network issues, the image-building time is not impacted as the images build concurrently, and this image just required less time to build than others.
> 
> We'll find another place.
> 
> Thanks for understanding and removing the code without caring about others' opinion.
> 
> cc @YuLi517

as I mentioned, there are three reasons I request to remove it.
1) the image has nothing with the code of repo.
2) after it put into the repo, it impact the PR merged.
3) your solution does not work. (I already told you the reason in the comment). you do not meet it now because you have not really run your workflow in the test. when you run it, you will meet the image not found issue.

what you mentioned is only about 2).
cc @jianzhangbjz @chengzhang1016

kuiwang02:
More information about 3)
I know you do not agree 3) because you think the test and image built are defined in same component, it does not need promoted. 
But what I tell you is that test definition is in __(amd64|arm64)-nightly.yaml and image built definition is in -(master|4.13).yaml. they are different file although they are in same component, so it still need promoted.

your step definition is 
```yaml
ref:
  as: baremetal-lab-install-ipi
  from: tests-private-bm
```
here tests-private-bm is just imagestream tag and it must match one imagestream to get image.  when the test job starts, it will use imagestream tag tests-private-bm to match the image, and it could be 
 1) pipeline:tests-private-bm if the job definition is not same with image built definition file, and you need to import it with `base_images` in __(amd64|arm64)-nightly.yaml  (here pipeline is image stream name).
  if you do not import it with `base_images`, you will meet the image not found error. 

 2) stable:tests-private-bm if the job definition is same with image built definition file and the image built will be in stable image stream with tests-private-bm

because the test definition which use your workflow is in __(amd64|arm64)-nightly.yaml and image built definition is in -(master|4.13).yaml, you have to import it with `base_images` in __(amd64|arm64)-nightly.yaml. But the image can not be promoted, so it does not work.

PR #34320, opened 2022-11-25
Add a proxy clusterbot variant for OpenStack

mandre:


mandre:
/close
Nothing to see here, we already have the proxy variant for openstack :man_facepalming:

PR #34325, opened 2022-11-25
Serverless: Make all jobs on 4.8 to only run on demand

mgencur:
* Same for tests with Service Mesh on OCP 4.11

This only changes the behaviour for the "main" branch of serverless-operator. Not the other branches.

Most of the jobs were already with `always_run=false` but defining `skip_if_only_changed: test/ui` actually made them execute always. Removing this stanza to make them executable on demand.


pierDipi (approved):
Thank you!

/lgtm
/approve

PR #34326, opened 2022-11-25
Skip some tests for Machine API components

Fedosin:
This PR skips tests in case if only unrelated files were modified.

For machine-api-operator it additionally disables Machine API test suite for modified unit test files (*_test\.go$).
If only vsphere controller was affected, we don't run e2e tests on other platforms.


elmiko (commented):
nice update, i do wonder about adding the `_test.go` files to the skip list, but i think it makes sense since we aren't adding it to the unit test jobs.

/lgtm

PR #34359, opened 2022-11-28
Add the pj-rehearse command to the auto-bump PR body

smg247:
https://github.com/kubernetes/test-infra/pull/28102 added a way to append text to the prow auto-bump PR. We can use this to command pj-rehearse to run and ack iff all rehearsals pass. This will effectively replace the `pj-rehearse-blocking` check. Once I confirm that this functions properly I will remove that check, and then add the label to the merge criteria.

To be merged after https://github.com/openshift/release/pull/34358

For: https://issues.redhat.com/browse/DPTP-3218

/cc @openshift/test-platform

### DIFF
--- a/core-services/prow/02_config/_bumper.yaml
+++ b/core-services/prow/02_config/_bumper.yaml
@@ -6,6 +6,7 @@ gitHubLogin: "openshift-bot"
 gitHubRepo: "release"
 remoteName: "release"
 assign_to: openshift/test-platform
+additionalPRBody: "/pj-rehearse auto-ack"
 includedConfigPaths:
   - "."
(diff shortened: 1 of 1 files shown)


droslean:
@smg247 Why we can't just add the desired labels to https://github.com/openshift/ci-tools/blob/master/cmd/autoconfigbrancher/main.go#L279-L285?

smg247:
> @smg247 Why we can't just add the desired labels to https://github.com/openshift/ci-tools/blob/master/cmd/autoconfigbrancher/main.go#L279-L285?

Because that won't actually trigger the rehearsals to run. This runs them and adds the label only if they all succeed.

PR #34363, opened 2022-11-29
HyperShift: fix mgmt cluster creation

csrwng:
Uses manually pushed image to quay.io/hypershift/release as the default release image for management clusters. Nightly images are pruned from the CI registry on a regular basis.


sjenning:
I still don't like the idea of pinning some nightly release here but need to get CI back online

/lgtm

PR #34364, opened 2022-11-29
Add 4.6 to 4.12 upgrade job on vSphere

shellyyang1989:
Refers to [OCPQE-12886](https://issues.redhat.com//browse/OCPQE-12886)

### DIFF
--- a/ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
+++ b/ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
@@ -427,6 +427,7 @@ function upgrade() {
 
 # Monitor the upgrade status
 function check_upgrade_status() {
+    # shellcheck disable=SC2153
     local wait_upgrade="${TIMEOUT}" out avail progress
     while (( wait_upgrade > 0 )); do
         sleep 5m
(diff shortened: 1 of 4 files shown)


ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -460,6 +461,120 @@ function check_history() {
...
+    first_node=$(echo ${master_nodes} | cut -d" " -f1)
+    cluster_name=$(<"${SHARED_DIR}"/clustername.txt)
+    vm_path="/${GOVC_DATACENTER}/vm/${cluster_name}"
+
+    current_hw_version=$(govc vm.info -json=true ${vm_path}/${first_node} | jq -r .VirtualMachines[0].Config.Version)
+    echo "Current hardware version is: ${current_hw_version}"
+    hw_version=$(echo ${current_hw_version} | cut -d'-' -f2)
+
+    if (( hw_version >= VSPHERE_HW_VERSION )); then
+        echo "VM hardware version is not less than ${VSPHERE_HW_VERSION}, no upgrade is required"
```
jhou1, line 480: `is not less than ` -> `is greater than`

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -460,6 +461,120 @@ function check_history() {
...
+    fi
+
+    echo "Updating VM hardware version to ${VSPHERE_HW_VERSION}"
+    # shellcheck disable=SC2068
+    for node in ${master_nodes[@]} ${worker_nodes[@]}; do
+        echo "Marking the node as unschedulable"
+        oc adm cordon ${node}
+        if [[ "${worker_nodes[*]}" =~ ${node} ]]; then
+            echo "Evacuate the pods from the compute node"
+            oc adm drain ${node} --force=true --delete-emptydir-data --ignore-daemonsets
```
jhou1, line 491: An active `PodDisruptionBudget` may break and `oc adm drain` indefinitely, having a heavy operation like this does not always guarantee the best results. An ideal state would be that the hardware meets requirements by default. If vSphere has so much added complexity, I suggest running the automation on a different cloud and leaving this to manual operation? 

WDYT @jianlinliu @evakhoni @jinyunma ?

shellyyang1989:
> Thank you @shellyyang1989 , a great effort to get things on vSphere to work!
> 
> Out of curiosity, on what chances are we running on a version smaller than required? I'm asking because the complexity of `vsphere_hw_upgrade` is worth being extracted to a pre-prerequisite step than having it in the toimage step.

AFAIK, if a vsphere cluster is born with 4.9 or lower version, the hardware version will require upgrade when upgrading to 4.10. Please correct me if I'm wrong @jinyunma 

I did have put it as a step https://github.com/openshift/release/tree/master/ci-operator/step-registry/cucushift/upgrade/vsphere-hw before. It only works for the upgrade from 4.9 and higher version. But it doesn't work for the long path upgrade like 4.6 to latest because we need to upgrade cluster to 4.9 and then upgrade vsphere hardware, then upgrade to the latest version. It's a required step in the middle of the upgrade. That's why I'm moving it to the toimage step.

jhou1:
Thank you for the clarification Yang, now I recall I've been told this context before, sorry for asking again.

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-ref.yaml
```diff
@@ -18,6 +18,9 @@
   - name: PLATFORM_VERSION
     default: "8.6"
     documentation: RHEL worker OS version
+  - name: VSPHERE_HW_VERSION
```
jianlinliu, line 21: How about rename the name to `TARGET_VSPHERE_HW_VERSION`

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
jianlinliu: I guess we have to hit `drain node` cases in the future even we did no to that this time.

jinyunma: agree, there is similar bug https://issues.redhat.com/browse/OCPBUGS-2353 when updating machineconfig, drain node failed due to break the pdb of openshift-cluster-csi-drivers, but bug is closed now and refer to the enhancement https://issues.redhat.com/browse/RFE-1367.

I remembered that the operation of drain node is one step in doc [Updating the virtual hardware for compute nodes on vSphere](https://docs.openshift.com/container-platform/4.11/updating/updating-hardware-on-nodes-running-on-vsphere.html#update-vsphere-virtual-hardware-on-compute-nodes_updating-hardware-on-nodes-running-in-vsphere), the options of drain node referred to [here](https://docs.openshift.com/container-platform/4.11/nodes/nodes/nodes-nodes-working.html#nodes-nodes-working-evacuating_nodes-nodes-working) and based on our testing, so agree that it might not guarantee passed on every time.

jianlinliu: Another point, I am thinking if it is possible to split the `drain node` operations into a dedicated step, so that we can opt-in the step when it is required.

chengzhang1016:
> I'm thinking if we can cover vSphere upgrade from 4.9 and higher versions. Thus, we don't need to move the step
I agree with you. Thanks. @shellyyang1989

shellyyang1989:
As discussed above, we will update the job to a non-vSphere platform to reduce the complexity. 

/close

PR #34370, opened 2022-11-29
assisted installer: add assisted-grafana to CI

rccrdpccl:
Move grafana dashboard for assisted-installer to separate repository. This way we can simplify slightly CI execution for both projects and clearer history, etc.

/cc @gamli75 @osherdp

### DIFF
--- a/ci-operator/config/openshift-assisted/assisted-grafana/openshift-assisted-assisted-grafana-master.yaml
+++ b/ci-operator/config/openshift-assisted/assisted-grafana/openshift-assisted-assisted-grafana-master.yaml
@@ -1,0 +1,21 @@
+build_root:
+  image_stream_tag:
+    name: release
+    namespace: openshift
+    tag: golang-1.18
+resources:
+  '*':
+    limits:
+      memory: 4Gi
+    requests:
+      cpu: 100m
+      memory: 200Mi
+tests:
+- as: validate
+  commands: make validate
+  container:
+    from: src
+zz_generated_metadata:
+  branch: master
+  org: openshift-assisted
+  repo: assisted-grafana
--- a/core-services/prow/02_config/openshift-assisted/assisted-grafana/_prowconfig.yaml
+++ b/core-services/prow/02_config/openshift-assisted/assisted-grafana/_prowconfig.yaml
@@ -1,0 +1,12 @@
+tide:
+  queries:
+  - labels:
+    - approved
+    - lgtm
+    missingLabels:
+    - do-not-merge/hold
+    - do-not-merge/invalid-owners-file
+    - do-not-merge/work-in-progress
+    - needs-rebase
+    repos:
+    - openshift-assisted/assisted-grafana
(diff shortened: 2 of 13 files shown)


core-services/prow/02_config/openshift-assisted/assisted-grafana/_prowconfig.yaml
```diff
@@ -0,0 +1,15 @@
+tide:
+  queries:
+  - labels:
+    - approved
+    - lgtm
+    missingLabels:
+    - backports/unvalidated-commits
+    - bugzilla/invalid-bug
```
osherdp, line 8: ``bugzilla/invalid-bug`` seems redundant here, no?
similar question to the ``jira/invalid-bug`` label, as we don't have the jira plugin enabled for this repo

ci-operator/config/openshift-assisted/assisted-grafana/openshift-assisted-assisted-grafana-master.yaml
```diff
@@ -0,0 +1,21 @@
+build_root:
+  image_stream_tag:
+    name: release
+    namespace: openshift
+    tag: golang-1.13
```
osherdp, line 5: can we upgrade to 1.18?
just so we won't have a too old image

osherdp (changes requested):
just two small things

core-services/prow/02_config/openshift-assisted/assisted-grafana/_prowconfig.yaml
rccrdpccl: Sure, this was copy-pasted from other repos so I will also change that in another PR

rccrdpccl:
rehearse will likely fail as the test is not in master https://github.com/openshift-assisted/assisted-grafana/tree/add-readme

osherdp (approved):

rccrdpccl:
@osherdp are you aware if this failure is due to my config?

osherdp:
> @osherdp are you aware if this failure is due to my config?

Seems like it's because we still enable bugzilla plugin globally:
https://github.com/openshift/release/blob/dec7a530455aac1f30bdadca5045d685ab8a93a2/core-services/prow/02_config/openshift-assisted/_pluginconfig.yaml#L62

osherdp (approved):

rccrdpccl:
> Seems like it's because we still enable bugzilla plugin globally:

Thanks, I feel dumb for not seeing it :sweat: 
I take it we can remove it globally as, AFAIU, we do not use it in any longer (and even more so in openshift-assisted)

rccrdpccl:
> Sure, this was copy-pasted from other repos so I will also change that in another PR

I guess I need to change them in this PR after all :sweat_smile:

core-services/prow/02_config/openshift-assisted/assisted-installer-deployment/_prowconfig.yaml
```diff
@@ -26,7 +26,6 @@ tide:
     - lgtm
     missingLabels:
     - backports/unvalidated-commits
-    - bugzilla/invalid-bug
     - do-not-merge/hold
     - do-not-merge/invalid-owners-file
     - do-not-merge/work-in-progress
```
osherdp, line 31: should we remove ``jira/invalid-bug`` as well?

osherdp (changes requested):

osherdp (approved):

PR #34390, opened 2022-11-29
assisted installer grafana: change namespace for base centos image

rccrdpccl:
The image does not exist in `origin` namespace, apparently it's `coreos` namespace


adriengentil:
FYI the mapping is here: https://github.com/openshift/release/blob/master/core-services/image-mirroring/supplemental-ci-images/mapping_supplemental_ci_images_coreos#L4

osherdp (approved):

PR #34402, opened 2022-11-30
Replace aws-ipi-ovn-ingress-nlb with aws-ipi-ingress-lbtype-nlb in 4.12+

liangxia:
Since 4.12+, for ingress, we have `ingress-lbtype-nlb` and `ingress-lbtype-classic`.
`aws-ipi-ovn-ingress-nlb` should be use in 4.11 and earlier versions.

/cc @lihongan @jianlinliu @jhou1 @JianLi-RH @dis016 @pruan-rht

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-nightly-4.12-upgrade-from-stable-4.12.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-nightly-4.12-upgrade-from-stable-4.12.yaml
@@ -93,6 +93,18 @@ tests:
     - ref: cucushift-installer-check
     - chain: openshift-upgrade-qe-test-disconnected
     workflow: cucushift-installer-rehearse-aws-ipi-disconnected-private
+- as: aws-ipi-ingress-lbtype-nlb-p2-f7
+  cron: 22 21 2,10,17,24 * *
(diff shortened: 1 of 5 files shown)


ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-nightly-4.12-upgrade-from-stable-4.11.yaml
```diff
@@ -97,30 +97,30 @@ tests:
     - ref: cucushift-installer-check
     - chain: openshift-upgrade-qe-test-disconnected
     workflow: cucushift-installer-rehearse-aws-ipi-disconnected-private
-- as: aws-ipi-ovn-fips-p2-f7
-  cron: 42 6 16,24,31,7 * *
+- as: aws-ipi-ingress-lbtype-nlb-p2-f7
```
lihongan, line 100: This file is upgrade-from-stable-4.11, but we don't support `ingress-lbtype-nlb` in 4.11.

liangxia: Good catch. Reverted the change for this file.

lihongan:
/lgtm
Thank you @liangxia

PR #34411, opened 2022-11-30
Assisted: run gather chain

vrutkovs:
That would allow us to re-use `gather` steps from OCP,
additionally collecting Prom data, apiserver snapshot for KAS and
Camgi to track Machineset/nodes status

### DIFF
--- a/ci-operator/config/openshift/assisted-test-infra/openshift-assisted-test-infra-master.yaml
+++ b/ci-operator/config/openshift/assisted-test-infra/openshift-assisted-test-infra-master.yaml
@@ -527,6 +527,9 @@ tests:
   skip_if_only_changed: ^docs/|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|OWNERS_ALIASES|PROJECT|LICENSE)$
   steps:
     cluster_profile: packet-sno
+    env:
+      SNO_CONFIG: |
+        BASE_DOMAIN=metalkube.org
     workflow: baremetalds-sno
 - as: e2e-metal-single-node-with-worker-live-iso
   run_if_changed: .*bootstrap_in_place.*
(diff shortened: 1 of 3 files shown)


ci-operator/step-registry/assisted/baremetal/operator/ztp/assisted-baremetal-operator-ztp-workflow.yaml
```diff
@@ -2,7 +2,6 @@ workflow:
   as: assisted-baremetal-operator-ztp
   steps:
     allow_best_effort_post_steps: true
-    allow_skip_on_success: true
```
mkowalski, line 5: Be aware, enabling gather on successful ZTP jobs is potentially going to cause timeouts for successful jobs. At least in the past we were running on a very short buffer (like 5 minutes till the timeout). It's a good change, but it should be monitored. If we see too many timeouts suddenly, we'll need to think about some smart solution

vrutkovs: ah, you mean `post` steps may timeout and make successful jobs fail? Right, this is bad, lets keep it WIP until we figure out a good solution for this

mkowalski: Yes, exactly this scenario... The real meat of the job succeeds, then gather in post times out and the whole job fails as "timeout". Still, this was in the past, maybe this problem now disappeared as we use more modern CPUs, as the universe is in a different position and all sorts of different reasons. By no means I want to stop PR by this, just something we need to pay attention at

vrutkovs: Fixed - not touching allow_skip_on_success

PR #34421, opened 2022-11-30
[Baremetal/QE] Adding dns config step

aleskandro:


ci-operator/step-registry/baremetal/lab/pre/dns/baremetal-lab-pre-dns-commands.sh
```diff
@@ -1,2 +1,87 @@
...
+# shellcheck disable=SC1090
+. <(yq e 'to_entries | .[] | (.key + "=\"" + .value + "\"")' < "${SHARED_DIR}"/external_vips.yaml)
+# shellcheck disable=SC2154
+if [ ${#api_vip} -eq 0 ] || [ ${#ingress_vip} -eq 0 ]; then
+  echo "Unable to parse VIPs"
+  exit 1
+fi
+
+DNS_FORWARD="#DO NOT EDIT; BEGIN $NAMESPACE
+api.${NAMESPACE} IN A ${api_vip}
```
jhou1, line 41: A note, we may also want to consider `AAAA` for ipv6 single stack and `A`,`AAAA` for dualstack clusters.

ci-operator/step-registry/baremetal/lab/pre/dns/baremetal-lab-pre-dns-commands.sh
```diff
@@ -1,2 +1,87 @@
...
+  . <(echo "$bmhost" | yq e 'to_entries | .[] | (.key + "=\"" + .value + "\"")')
+  # shellcheck disable=SC2154
+  if [ ${#name} -eq 0 ] || [ ${#ip} -eq 0 ]; then
+    echo "Error when parsing the Bare Metal Host metadata"
+    exit 1
+  fi
+  DNS_FORWARD="${DNS_FORWARD}
+${name}.${NAMESPACE} IN A ${ip}"
+  DNS_REVERSE_INTERNAL="${DNS_REVERSE_INTERNAL}
+$(echo "${ip}." | ( while read -r -d . b; do rip="$b${rip+.}${rip}"; done; echo "$rip" )).in-addr.arpa. IN PTR ${name}.${NAMESPACE}.${BASE_DOMAIN}"
```
jhou1, line 59: For ipv6 it's `.ip6.arpa.`, but the v6 address has more forms such as
- `2001:0db8:85a3:0000:0000:8a2e:0370:7334`
- `2001:DB8::1`
- `2001:db8:0:0:0:0:2:1`

aleskandro: Yes, I added a to do a few rows later. Either to do that as a follow up or before merging.

My only concern that could lead me to propose doing it in a follow-up is that I don't know how the ipv6 pool is managed in the upstream RH network and how we should do it for the self managed one, when available. It needs some thinking to get a stable solution IMO.

jhou1: Python3 standard library can handle that at ease.

```
>>> import ipaddress
>>> ipaddress.ip_address("2001:db8::1").reverse_pointer
'1.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.8.b.d.0.1.0.0.2.ip6.arpa'
>>> ipaddress.ip_address("192.168.11.1").reverse_pointer
'1.11.168.192.in-addr.arpa'
```

aleskandro: I was wondering more about the ipv6 network topology configuration in the lab

aleskandro: E.g., ipv6 could not reach external hosts in internet at the time I tried

aleskandro: Let's sync offline for that once

ci-operator/step-registry/baremetal/lab/pre/dns/baremetal-lab-pre-dns-commands.sh
aleskandro: Merging to proceed ahead with the other steps @jhou1 and get a first test running asap. I'll follow up with the ipv6 support in the next days.

aleskandro:
Refers OCPQE-11881

PR #34434, opened 2022-11-30
Enable postsubmits for multicloud on arm

deepsm007:
This will enable building images on arm01 prow cluster and images will be mirrored to apps.ci internal registry

/cc @mikeshng @xiangjingli  @gurnben


mikeshng (approved):
/lgtm

/approve

PR #34439, opened 2022-11-30
WIP Add madunn to OWNERS for rhpit/interop-tests and create OADP config/job

madunn:
Add madunn to OWNERS for rhpit/interop-tests

### DIFF
--- a/ci-operator/config/rhpit/interop-tests/rhpit-interop-tests-master__ocp-412-oadp10.yaml
+++ b/ci-operator/config/rhpit/interop-tests/rhpit-interop-tests-master__ocp-412-oadp10.yaml
@@ -39,7 +39,14 @@ tests:
     cluster_profile: aws-cspi-qe
     env:
       BASE_DOMAIN: aws.interop.ccitredhat.com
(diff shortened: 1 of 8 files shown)


ryankwilliams:
@madunn @ascerra The subject of this PR says adding user to owners and the changes include that and a config/job for OADP. Should this PR be kept to the scope of just updating the owners file and a separate PR for the OADP config/job?

If the goal is to do both in this one PR, I would suggest revising the subject and have separate commits be introduced for each change.

ascerra:
What in this PR lead to the job`ci/rehearse/openshift-priv/cluster-api-provider-kubevirt/main/cluster-api-provider-kubevirt-e2` being kicked off?

ascerra (commented):
General comment

- We need to discuss where the config file will live for this scenario. The rhpit dir is just temporary for our development/learning. For teams with github test repos its easy to decide to put the config in the org_name/repo_name of the test repo. Does this scenario have github test repo or is it all gitlab, if so we need to brainstorm a place to put these configs for this scenario and others like it.

PR #34440, opened 2022-11-30
core-services: Add openshift/api allowed_teams

aravindhp:
Add @openshift/openshift-staff-engineers to allowed_teams for applying the `backport-risk-assessed` and `cherry-pick-approved` labels.

### DIFF
--- a/core-services/prow/02_config/openshift/api/_pluginconfig.yaml
+++ b/core-services/prow/02_config/openshift/api/_pluginconfig.yaml
@@ -1,14 +1,18 @@
 label:
   restricted_labels:
     openshift/api:
-    - allowed_users:
+    - allowed_teams:
+      - openshift-staff-engineers
+      allowed_users:
       - xingxingxia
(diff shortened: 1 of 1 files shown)


core-services/prow/02_config/openshift/api/_pluginconfig.yaml
```diff
@@ -5,6 +5,7 @@ label:
       - xingxingxia
       - wangke19
       - gangwgr
+      - openshift-staff-engineers
```
bparees, line 8: i think this needs to go under `allowed_teams`, e.g.:
https://github.com/openshift/release/blob/fd44743075db1228fb4930d1d7bd3037fffb3695/core-services/prow/02_config/_plugins.yaml#L2288-L2295

aravindhp: @bparees Thanks for catching that. Done.

PR #34457, opened 2022-12-01
fix zone issue for azure-ipi-workers-marketimage long-duration-cloud

sunzhaohua2:
Now the job failed with` error msg=Zone Name: "ci.azure.devcluster.openshift.com"`, this will fix this issue.
@miyadav @huali9 @jhou1 PTAL


miyadav:
Thanks @sunzhaohua2 
/lgtm

PR #34458, opened 2022-12-01
MGMT-12789: remove bugzilla prow configurations

osherdp:
Bugzilla no longer used for either OpenShift or ACM, so we shouldn't restrict any PR from merging due to having or not having bugzilla labels.

Another PRs will follow-up to also adjust the Jira plugin configuration.

/cc @filanov @adriengentil 
/hold


osherdp:
```
the following orgs or repos enable the bugzilla plugin but do not forbid the bugzilla/invalid-bug label for merging: [repo: openshift/assisted-image-service ...
```
we can go different directions here (like managing plugins list differently for our repos than other repos under ``openshift``). I think for now I'll just add the restriction of the damn label :)

core-services/prow/02_config/openshift/assisted-image-service/_prowconfig.yaml
```diff
@@ -31,30 +31,7 @@ tide:
...
-    - release-ocm-2.5
-    - release-ocm-2.6
-    labels:
-    - approved
-    - bugzilla/valid-bug
-    - lgtm
-    missingLabels:
-    - backports/unvalidated-commits
-    - bugzilla/invalid-bug
-    - do-not-merge/hold
```
filanov, line 45: don't we need those? hold and wip

filanov (approved):

core-services/prow/02_config/openshift/assisted-image-service/_prowconfig.yaml
osherdp: a duplicated entry basically, so now configurations for master and ``release-ocm-*`` are unified

PR #34495, opened 2022-12-02
[openshift-loki] Fix operator bundle channel for release-5.5

periklis:
Change bundle in CI to use the `stable-5.5` channel instead of `stable`.

Depends on:
* https://github.com/openshift/loki/pull/92

/cc @jcantrill @xperimental


jcantrill (commented):
/lgtm

PR #34498, opened 2022-12-02
Support of OCPQE-13111

liangxia:
To support root cause of [OCPQE-13111](https://issues.redhat.com//browse/OCPQE-13111)


xingxingxia:
@liangxia thank you!

PR #34505, opened 2022-12-02
Use expiration label for pr images

maroroman:
This PR enables the usage of the EXPIRATION arg in the odh-dashboard dockerfile specifically for the case of pr-xxx images.

Should only be merged after: https://github.com/opendatahub-io/odh-dashboard/pull/811 is merged


maroroman:
Closing this PR as an alternative solution was found

PR #34508, opened 2022-12-02
USHIFT-660: MicroShift router smoke test

pmtk:
Adds an optional pre-submit job that deploys a simple `hello-openshift` Pod, exposes it and attempts to curl it from outside the MicroShift VM.


mangelajo (approved):
/lgtm

PR #34531, opened 2022-12-05
Make Zuul job required for mariadb operator

raukadah:
https://github.com/openstack-k8s-operators/mariadb-operator/pull/46 tests the zuul github pr testing on rdoproject zuul tenant.

rdoproject.org/github-check runs on zuul. In order to make it voting, it needs to be included here.

Signed-off-by: Chandan Kumar (raukadah) <raukadah@gmail.com>

PR #34533, opened 2022-12-05
Use github-check for making zuul job voting

raukadah:
https://github.com/openshift/release/pull/34531 added rdoproject.org/github-check job to make it voting.

It is not working. So trying with github-check only.

Signed-off-by: Chandan Kumar (raukadah) <raukadah@gmail.com>


Sandeepyadav93:
/lgtm

lets try this.

PR #34538, opened 2022-12-05
Make zuul job voting by passing job name

raukadah:
pcp_edpm_mariadb_operator_rdo_github_integration runs as a part of rdoproject.org/github-check. Let's try with job name as it makes the job required.

Signed-off-by: Chandan Kumar (raukadah) <raukadah@gmail.com>


Sandeepyadav93:
/lgtm

let's try this.

PR #34539, opened 2022-12-05
MGMT-12654: branch-out release-ocm-2.7

osherdp:
This should include all new configuration files for the new stable branches of ACM 2.7 / MCE 2.2.

I copied configurations from ``master/main`` branches to make sure we're following the same building logic for the new branches, but only allowed certain jobs that are relevant for those branches (removing some publish jobs that are only built against master, and removing some REST-API jobs aside).

This should be dependent on
https://github.com/openshift/release/pull/34253, otherwise we're into the chicken-and-egg problem (component A depends for its building on component B, but also vice-versa).

/cc @eliorerz @adriengentil @gamli75


eliorerz (approved):
LGTM

PR #34545, opened 2022-12-05
[4.10] - Add repo names to match the RHCOS repo names

ravanelli:
- We are changing the repo names in order to improve our internal redhat-coreos repo, to use a single branch for the RHCOS versions. We need to add the names to macth our repos names now.

Signed-off-by: Renata Ravanelli <rravanel@redhat.com>

### DIFF
--- a/core-services/release-controller/_repos/ocp-4.10-rhel8.repo
+++ b/core-services/release-controller/_repos/ocp-4.10-rhel8.repo
@@ -1,5 +1,5 @@
-[rhel-8-baseos]
-name = rhel-8-baseos
+[rhel-8.4-baseos]
+name = rhel-8.4-baseos
 baseurl = https://cdn.redhat.com/content/eus/rhel8/8.4/x86_64/baseos/os/
 enabled = 1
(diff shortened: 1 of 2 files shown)


core-services/release-controller/_repos/ocp-4.10-rhel8.repo
```diff
@@ -1,3 +1,15 @@
+[rhel-8.4-baseos]
```
travier, line 1: Can you add the repos below the existing ones so that we don't change the liveness probe in the file above?

travier (commented):
Minor nit but LGTM

core-services/release-controller/_repos/ocp-4.10-rhel8.repo
ravanelli: Done

ravanelli:
cc @prestist

jupierce:
Please remove the old repos. The duplication could lead to confusion.

jupierce:
Most teams aren't explicitly enabling or disabling repos. If they are, then we silently break them unless we remove the old names. If we remove the names, at least we loudly break them :) .

ravanelli:
> Most teams aren't explicitly enabling or disabling repos. If they are, then we silently break them unless we remove the old names. If we remove the names, at least we loudly break them :) .

SGTM, let's see who screams first. ;)

PR #34571, opened 2022-12-06
MGMT-12804: Make assisted-service-index:ocm-2.7 image available in Prow

eliorerz:
Manually create an ocm-2.7 release of the assisted-service-index from latest

/cc @osherdp 
/cc @adriengentil

### DIFF
--- a/clusters/app.ci/multi-arch/namespaces.yaml
+++ b/clusters/app.ci/multi-arch/namespaces.yaml
@@ -4,11 +4,11 @@ items:
 - apiVersion: v1
   kind: Namespace
   metadata:
-    name: coreos-arm64
+    name: azure-arm64
 - apiVersion: v1
   kind: Namespace
   metadata:
-    name: azure-arm64
+    name: coreos-arm64
 - apiVersion: v1
   kind: Namespace
   metadata:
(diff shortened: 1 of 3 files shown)


osherdp (approved):

adriengentil (approved):

clusters/app.ci/multi-arch/namespaces.yaml
```diff
@@ -4,11 +4,11 @@ items:
 - apiVersion: v1
   kind: Namespace
   metadata:
-    name: coreos-arm64
+    name: azure-arm64
```
osherdp, line 7: is it required?

osherdp (approved):

bear-redhat:
Are the changes in `clusters/app.ci/multi-arch/namespaces.yaml` required?

clusters/app.ci/multi-arch/namespaces.yaml
eliorerz: I got this error on `multi-arch-gen-valid` test
```
ERROR: To update the the multi-arch image mirroring run the following and
ERROR: commit the result:
ERROR: $ make multi-arch-gen
```

mkowalski:
@eliorerz @bear-redhat as I read changes in this file, they are only cosmetic and change order of the objects, but not the number and content

clusters/app.ci/multi-arch/namespaces.yaml
osherdp: Apparently it returns a different order in my env. I think ``ruamel.yaml`` has some regression which makes it returning stuff in different order sometimes
I recreated this change in https://github.com/openshift/release/pull/34728 without reordering

PR #34574, opened 2022-12-06
USHIFT-660: MicroShift Router smoke test - improve failure logs

pmtk:


ci-operator/step-registry/openshift/microshift/e2e/router-smoke-test/openshift-microshift-e2e-router-smoke-test-commands.sh
```diff
@@ -105,12 +105,49 @@ RESULT=$?
...
+for ns in $(oc get namespace -o jsonpath='{.items..metadata.name}'); do
+      for pod in $(oc get pods -n $ns -o name); do
+            oc describe -n $ns $pod || true
+            for container in $(oc get -n $ns $pod -o jsonpath='{.spec.containers[*].name}'); do
+                  oc logs -n $ns $pod $container
+                  oc logs --previous=true -n $ns $pod $container
+            done
+      done
+done
+EOF
```
ggiguash, line 135: We have this code in 2 places already, don't we? Should we add this script under ./hack/ and reuse it in the CI?

ggiguash (changes requested):

PR #34576, opened 2022-12-06
Fix MCE builds missing image dependencies

osherdp:
After https://github.com/openshift/release/pull/34539 we have this kind of failure:
```
step assisted-service is missing dependencies:
"openshift_release_golang-1.18" is neither an imported nor a built image
```

(see
https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/branch-ci-openshift-assisted-service-release-ocm-2.7-mce-images/1600110027492823040)

This is because I somehow forgot to import the base image before using it for building the assisted-service.

This change fixes it as well as implementing image-substitution for master branch (for increasing performance).

/cc @eliorerz 
/hold


eliorerz (approved):
LGTM

PR #34586, opened 2022-12-06
AGENT-350: Collect journal logs and publish them as artifacts

pawanpinjarkar:
Signed-off-by: Pawan Pinjarkar <ppinjark@redhat.com>

### DIFF
--- a/ci-operator/step-registry/agent/gather/agent-gather-commands.sh
+++ b/ci-operator/step-registry/agent/gather/agent-gather-commands.sh
@@ -13,5 +13,5 @@ echo "### Gathering logs..."
 timeout -s 9 5m ssh "${SSHOPTS[@]}" "root@${IP}" bash - <<EOF
 cd dev-scripts
 make agent_gather
-scp agent-gather.tar.xz "${ARTIFACT_DIR}"
+cp agent-gather*.tar.xz "${ARTIFACT_DIR}"
 EOF


ci-operator/step-registry/agent/devscripts/OWNERS
```diff
@@ -0,0 +1,7 @@
+approvers:
+- andfasano
```
andfasano, line 2: Approver list could be simply set as the default agent OWNERS people

ci-operator/step-registry/agent/devscripts/gather/agent-devscripts-gather-commands.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+echo "************ agent gather command ************"
+echo "### Gathering logs..."
+timeout -s 9 5m ssh "${SSHOPTS[@]}" "root@${IP}" bash - <<EOF
```
andfasano, line 9: `SSHOPTS` and `IP` are set only by sourcing the `packet-conf.sh` file, which is currently generated by the [packet-setup step](https://github.com/openshift/release/blob/0fc3adfbd284c6fda81fff80dd903b605ceca4e3/ci-operator/step-registry/baremetalds/packet/setup/baremetalds-packet-setup-commands.sh#L70-L93), ie: 

```
# Fetch packet basic configuration
# shellcheck source=/dev/null
source "${SHARED_DIR}/packet-conf.sh"
```

ci-operator/step-registry/agent/post/agent-post-chain.yaml
```diff
@@ -2,6 +2,7 @@ chain:
   as: agent-post
   steps:
   - ref: baremetalds-devscripts-gather
+  - ref: agent-devscripts-gather
```
andfasano, line 5: About the naming, I'd personally to prefer a simple `agent-gather` (using dev-scripts here it's a tech implementation details, and in future it could change. The scope of the step it's anyhow to gather logs relevant for the agent case)

zaneb:
Artifacts were not generated because the cluster installed successfully (thus no hosts were left running the agent ISO), but at least we know it doesn't regress anything, and in principle it should help if something fails prior to the reboots.
/pj-rehearse ack

PR #34588, opened 2022-12-06
OCPVE-167 - disabling chronyd for rt tests

jeff-roche:


jeff-roche:
This configuration change is not relevant. Chronyd is disabled because the customer the suggestion came from has a different way of synchronizing time that is not relevant to our VMs

PR #34594, opened 2022-12-06
stolostron: update hypershift-addon-operator to use golang 1.19

mikeshng:
Signed-off-by: Mike Ng <ming@redhat.com>


mikeshng:
It seems like it doesn't like this old branch.

PR #34617, opened 2022-12-07
Modify agent-installer-api-server from image

andfasano:
Currently the `base` image points to repos that are not yet updated (see https://issues.redhat.com/browse/CLOUDBLD-12619).
This is causing a problem for https://github.com/openshift/assisted-service/pull/4698, since it requires a newer version of nmstate packages (1.3.3) which is available only in the `rhel-8-golang-1.19-openshift-4.12` image repos.

Once https://issues.redhat.com/browse/CLOUDBLD-12619 will be available, the `from` image could be switched back to `base`

cc @nmagnezi

### DIFF
--- a/ci-operator/config/openshift/assisted-service/openshift-assisted-service-master.yaml
+++ b/ci-operator/config/openshift/assisted-service/openshift-assisted-service-master.yaml
@@ -3,6 +3,10 @@ base_images:
     name: "4.13"
     namespace: ocp
     tag: base
+  base-tmp:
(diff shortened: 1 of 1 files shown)


andfasano:
/hold

Looks like https://issues.redhat.com/browse/CLOUDBLD-12619 already update the 4.13 repo

andfasano:
4.13 base repos have been updated with the required nmstate version (1.3.3)

PR #34619, opened 2022-12-07
Fix gathering of bootstrap logs on non-proxy jobs

stephenfin:
In #33244, we added support for gathering logs via a bastion proxy. This was necessary because the installer does not currently [provide this functionality natively][1] and we wished to gather logs for our proxy jobs. However, we determined whether we should attempt the gather via a bastion by checking whether we have a proxy enabled, i.e. if there is a 'proxy-conf.sh' file present as generated by the 'conf-proxy' step.

This is a flawed check. The OpenStack CI also uses proxies for various NFV e2e tests (in fact, this was the original purpose of the 'conf-proxy' step, then known as 'provision-proxyconf') but does not use bastion hosts. We should have instead relied on a more accurate heuristic for determining whether we're running proxy tests - the presence of a 'CONFIG_TYPE' envvar set to 'proxy' - like we do in other steps like openstack-provision-floatingips, openstack-provision-bastionproxy and openstack-provision-machinesubnet.

Do this now.

[1]: https://issues.redhat.com//browse/CORS-2367

### DIFF
--- a/ci-operator/step-registry/openstack/gather/openstack-gather-commands.sh
+++ b/ci-operator/step-registry/openstack/gather/openstack-gather-commands.sh
@@ -19,6 +19,11 @@ fi
 if [[ "$CONFIG_TYPE" == "proxy" ]]; then
     BASTION_FIP=$(<"${SHARED_DIR}/BASTION_FIP")
     BASTION_USER=$(<"${SHARED_DIR}/BASTION_USER")
+else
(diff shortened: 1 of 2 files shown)


dulek:
/lgtm
/hold

Unhold once you're happy with CI.

EmilienM:
```
/bin/bash: line 20: CONFIG_TYPE: unbound variable
```

stephenfin:
Whoops, I should have refreshed the page first :see_no_evil:

PR #34623, opened 2022-12-07
Multiple OCP CI tests for main branch - NHC

razo7:
Having multiple OCP versions, v4.11 & v 4.12, for CI tests with main branch in [NHC](https://github.com/medik8s/node-healthcheck-operator).


slintes:
oh, that retriggered all...

slintes:
there is an issue with the test pod which gets the boot time. Maybe related to changes done for PSA on 4.12... 🤔 Will investigate tomorrow.

/hold

slintes:
seems OCP 4.11 has no control plane labels yet, so we fail a master node in the e2e test, and CI cluster isn't happy about it... fixed in https://github.com/medik8s/node-healthcheck-operator/pull/152

PR #34624, opened 2022-12-07
Multiple OCP CI tests for main branch - SNR

razo7:
- Having multiple OCP versions, v4.11 & v 4.12, for CI tests with main branch in [SNR](https://github.com/medik8s/self-node-remediation).
- Update Golang version to v1.19


mshitrit:
@razo7 in case one test fail, better not to run pj-rehearse  again, since it'll run both - we can just run the failed job.
/pj-rehearse auto-ack

PR #34629, opened 2022-12-07
SDCICD-865: add osde2e webhook presubmits

jbpratt:
set up a presubmit job to run the webhook tests on change

/hold

### DIFF
--- a/ci-operator/config/openshift/osde2e/openshift-osde2e-main.yaml
+++ b/ci-operator/config/openshift/osde2e/openshift-osde2e-main.yaml
@@ -53,6 +53,28 @@ tests:
   - mount_path: /usr/local/sd-cicd-aws-prod
     name: sd-cicd-aws-prod
   skip_if_only_changed: ^(?:docs|\.github)|\.md$|^(?:\.gitignore|.golang-ci.yml|OWNERS|LICENSE)$
+- as: rosa-stage-mcvw
+  run_if_changed: pkg/e2e/webhooks
+  commands: |
+    export GOFLAGS=""
+    export REPORT_DIR="$ARTIFACT_DIR"
+    export CONFIGS="hypershift"
+    export SECRET_LOCATIONS="/usr/local/osde2e-common,/usr/local/osde2e-credentials,/usr/local/osde2e-rosa-stage,/usr/local/osde2e-hypershift-byovpc"
+
+    make build
+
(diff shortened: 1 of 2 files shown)


ci-operator/config/openshift/osde2e/openshift-osde2e-main.yaml
```diff
@@ -53,6 +53,28 @@ tests:
   - mount_path: /usr/local/sd-cicd-aws-prod
     name: sd-cicd-aws-prod
   skip_if_only_changed: ^(?:docs|\.github)|\.md$|^(?:\.gitignore|.golang-ci.yml|OWNERS|LICENSE)$
+- as: rosa-stage-mcvw
+  run_if_changed: pkg/e2e/webhooks
+  commands: |
+    export GOFLAGS=""
+    export REPORT_DIR="$ARTIFACT_DIR"
+    export CONFIGS="hypershift"
```
ryankwilliams, line 61: @jbpratt Is the plan to try to use `hypershift` cluster (if it supports webhooks) or do we want to stick with ROSA sts stage environment to test with?

ci-operator/config/openshift/osde2e/openshift-osde2e-main.yaml
```diff
@@ -53,6 +53,28 @@ tests:
   - mount_path: /usr/local/sd-cicd-aws-prod
     name: sd-cicd-aws-prod
   skip_if_only_changed: ^(?:docs|\.github)|\.md$|^(?:\.gitignore|.golang-ci.yml|OWNERS|LICENSE)$
+- as: rosa-stage-mcvw
+  run_if_changed: pkg/e2e/webhooks
```
ryankwilliams, line 57: @jbpratt We probably will want to modify the `verify` job config to append to the `skip_if_only_changed` to ignore webhook tests. What do you think? This way if a PR is submitted for webhook test changes, it will not try to run both these pre-submit jobs.

ci-operator/config/openshift/osde2e/openshift-osde2e-main.yaml
jbpratt: the plan was to _try_ to use hypershift because it would be quick but after the investigation this morning, I'll need to switch this to sts.

jbpratt:
closing this, we are going to take a different approach

PR #34630, opened 2022-12-07
dptp-controller: git-sync args fix

danilo-gemoli:
It seems `git-sync` doesn't run at all in the initContainers because of a missing `=`  on the `--depth` switch.
This causes the `dptp-controller-manager` to see the repository only after a delay of [30s](https://github.com/openshift/release/blob/master/clusters/app.ci/assets/dptp-controller-manager.yaml#L123), that is the interval `git-sync` from the sidecar container waits between syncs.

/cc @hongkailiu

### DIFF
--- a/clusters/app.ci/assets/dptp-controller-manager.yaml
+++ b/clusters/app.ci/assets/dptp-controller-manager.yaml
@@ -105,7 +105,7 @@ spec:
         - --branch=master
         - --root=/tmp/git-sync
         - --one-time=true
-        - --depth 1
+        - --depth=1
         env:
         - name: GIT_SYNC_DEST
           value: release


bbguimaraes:
For posterity: `git-sync` amazingly does not report an invalid argument as an error in this setup.

```
$ cat test_git_sync_pod.yaml
apiVersion: v1
kind: Pod
metadata:
  name: test
spec:
  containers:
  - image: gcr.io/google_containers/git-sync:v3.1.6
    name: test
    command:
    - /git-sync
    args:
    - --depth 1
  restartPolicy: Never
$ oc create -f test_git_sync_pod.yaml
pod/test created
$ oc logs test | head -4
INFO: detected pid 1, running init handler
flag provided but not defined: -depth 1
Usage of /git-sync:
  -add-user
$ oc get pod test -o json | jq '.status.containerStatuses[].state'
{
  "terminated": {
    "containerID": "cri-o://217ac3faf5cbcdfd37553eb996076b93cc5f0ddc9c86bbe9549a6363eeedf0b3",
    "exitCode": 0,
    "finishedAt": "2022-12-09T10:21:32Z",
    "reason": "Completed",
    "startedAt": "2022-12-09T10:21:32Z"
  }
}
```

It does so outside of a container, so this seems to be a problem in its "init handler".

PR #34644, opened 2022-12-07
adding e2e-alibaba as a presubmit to the mco

rvanderp3:
SPLAT has a need to test changes in the MCO that are related to alibaba as a presubmit test.

cc: @fabianofranz


yuqi-zhang (approved):
seems fine to me at a glance. Have not tested with alibaba personally but given that this is optional and not always_run I think it is safe to merge

PR #34648, opened 2022-12-08
stolostron: revert hypershift-addon-operator to use golang 1.18

mikeshng:
Signed-off-by: Mike Ng <ming@redhat.com>


mikeshng:
/assign @philipwu08 

/assign @rokej 

There are so few changes here because some lines were magically reverted back to 1.18 by a bot.

PR #34657, opened 2022-12-08
redhat-cps/front-proxy: build image

hardys:
Follow-up to #33813 where we seeded the repo, now we need to build and publish an image for the cps front-proxy repo

### DIFF
--- a/ci-operator/config/redhat-cps/front-proxy/redhat-cps-front-proxy-main.yaml
+++ b/ci-operator/config/redhat-cps/front-proxy/redhat-cps-front-proxy-main.yaml
@@ -1,6 +1,28 @@
+base_images:
+  os:
+    name: centos
+    namespace: origin
+    tag: stream8
+binary_build_commands: make build
 build_root:
   from_repository: true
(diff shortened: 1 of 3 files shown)


ci-operator/config/redhat-cps/front-proxy/redhat-cps-front-proxy-main.yaml
```diff
@@ -43,3 +49,19 @@ zz_generated_metadata:
...
+    ENTRYPOINT ["/usr/bin/cps-front-proxy"]
+  from: os
+  inputs:
+    bin:
+      paths:
+      - destination_dir: .
+        source_path: /go/src/github.com/redhat-cps/front-proxy/bin/cps-front-proxy
+  to: cps-front-proxy
+promotion:
+  namespace: kcp
```
hardys, line 65: @stevekuznetsov do you think this is reasonable or should we create a new org for redhat-cps?

I'm thinking if it's just this one image, having `registry.ci.openshift.org/kcp/cps-front-proxy` may be reasonable?

PR #34663, opened 2022-12-08
Set 'merge' as default merge method and disable blunderbuss plugin

kpouget:
The blunderbuss plugin picks up the default reviewers, see there https://prow.k8s.io/plugins

/cc @fcami


fcami (approved):
/lgtm

PR #34672, opened 2022-12-08
remove ERROR: build log highlighting

gavin-stackrox:
Highlighting any build.log line with 'ERROR:' in it causes quite a bit of confusion for stackrox users when they try to root cause a test failure. An understandable tendency is to jump to the highlighted build log lines and ignore JUnit errors.

e.g. https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/branch-ci-stackrox-stackrox-master-merge-gke-upgrade-tests/1584461683043078144

We have a tool used in a lot of CI that always prints: `ERROR:	Certificate validation error: x509: certificate signed by unknown authority` due to how the tests are setup/run. Changing that would be a major chore.

I'd like to just get rid of 'ERROR:' highlighting but would be happy with a less permissive regex that could meet the needs of the reason it was added: https://bugzilla.redhat.com/show_bug.cgi?id=1789127#c4

/cc @wking @davdhacs

### DIFF
--- a/core-services/prow/02_config/_config.yaml
+++ b/core-services/prow/02_config/_config.yaml
@@ -71,7 +71,6 @@ deck:
         config:
           highlight_regexes:
           - timed out
-          - 'ERROR:'
           - ERRO\[
           - '^error:'
           - '(^FAIL|FAIL: |Failure \[)\b'


danilo-gemoli:
I would not like to remove `ERROR:` for the following reasons:
- Highlight `ERROR:` is actually useful but more importantly it would be a major change in how Spyglass reports to users and I honestly don't know how other teams will be affected from it.
- I would rather trying to modify my tool first instead of changing the CI
- I understand the highlighted text is powerful and it is where everyone put the focus on but it is also responsibility of the user to carefully analyze the whole log

gavin-stackrox:
I would imagine highlighting all instances of `ERROR:` causes more false attention than is useful. But I'd be happy to be proven wrong. And I understand that it is a major change and near impossible to figure out who might complain / be reliant on existing behavior. 

What about:
- Announce the proposed change on #announce-testplatform and see if anyone is troubled
- Make the change and rollback if there _any_ pushback

@danilo-gemoli

jmguzik:
> And I understand that it is a major change and near impossible to figure out who might complain / be reliant on existing behavior.

That's the right statement. In similar situations, we use the rule: _the needs of the many..._ :)

> * Announce the proposed change on #announce-testplatform and see if anyone is troubled

I think that the problem with announcing on a platform such a specific thing is that either it will be ignored or followed up anyway on forum-testplatform with multiple requests by the people not subscribed to the channel. Either way, it will generate additional noise on our side.

Even after your explanation, I see it is made mainly to satisfy the needs of one team, while it should not be the case with the global configuration. I think we can agree that: `ERROR: Certificate validation error: x509: certificate signed by unknown authority` represents some kind of a problem. Even if the problem is not the big one, we can't really exclude these error types based on one output of one tool. Of course, the global configuration could be sharded and made local but that would require a feature implemented upstream which for us is a low priority. I would suggest going that way if you are interested in making it happen.

gavin-stackrox:
Yeah we went with the change on our tool side. I might try that sharding some day. Thanks for the considered responses.

PR #34676, opened 2022-12-08
[WIP] Add ROSA osde2e job to OCP informing suite

MrSantamaria:
[SDCICD-557](https://issues.redhat.com/browse/SDCICD-557)


MrSantamaria:
@wking What do you think of this PR? 
The pipeline will continue to fail until Rosa starts releasing nightlies for 4.13 but if we had implemented this for 4.12 it would have been working.

PR #34678, opened 2022-12-08
azure: upi: sync install workflow with installer changes

r4f4:
~~This change is inter-dependent on https://github.com/openshift/installer/pull/6684~~
This change adds support to both Managed Images and Image Galleries for the UPI workflow. Image Galleries support was added in the installer to enable `aarch64` IPI deployments and https://github.com/openshift/installer/pull/6684 will soon enable it for UPI as well.

### DIFF
--- a/ci-operator/step-registry/upi/install/azure/upi-install-azure-commands.sh
+++ b/ci-operator/step-registry/upi/install/azure/upi-install-azure-commands.sh
@@ -158,9 +158,11 @@ az storage account create -g $RESOURCE_GROUP --location $AZURE_REGION --name $AC
 ACCOUNT_KEY=$(az storage account keys list -g $RESOURCE_GROUP --account-name $ACCOUNT_NAME --query "[0].value" -o tsv)
 
 if openshift-install coreos print-stream-json 2>/tmp/err.txt >/tmp/coreos.json; then
-  VHD_URL="$(jq -r '.architectures.x86_64."rhel-coreos-extensions"."azure-disk".url' /tmp/coreos.json)"
+  VHD_URL="$(jq -r --arg arch "$OCP_ARCH" '.architectures[$arch]."rhel-coreos-extensions"."azure-disk".url' /tmp/coreos.json)"
+  RELEASE_VERSION="$(jq -r --arg arch "$OCP_ARCH" '.architectures[$arch]."rhel-coreos-extensions"."azure-disk".release' /tmp/coreos.json)"
 else
   VHD_URL="$(jq -r .azure.url /var/lib/openshift-install/rhcos.json)"
+  RELEASE_VERSION="$(jq -r .azure.release /var/lib/openshift-install/rhcos.json)"
 fi
 
 echo "Copying VHD image from ${VHD_URL}"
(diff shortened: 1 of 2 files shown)


Prashanth684:
/hold
unitl https://github.com/openshift/installer/pull/6684 merges

r4f4:
/pj-rehearse pull-ci-openshift-installer-master-e2e-azure-ovn-upi

r4f4:
/pj-rehearse pull-ci-openshift-installer-master-e2e-azure-ovn-upi

r4f4:
/pj-rehearse pull-ci-openshift-installer-master-e2e-azure-ovn-upi

r4f4:
/pj-rehearse pull-ci-openshift-installer-master-e2e-azure-ovn-upi

ci-operator/step-registry/upi/install/azure/upi-install-azure-commands.sh
```diff
@@ -218,10 +220,24 @@ az network private-dns link vnet create -g $RESOURCE_GROUP -z ${CLUSTER_NAME}.${
...
 VHD_BLOB_URL=$(az storage blob url --account-name $ACCOUNT_NAME --account-key $ACCOUNT_KEY -c vhd -n "rhcos.vhd" -o tsv)
-az deployment group create -g $RESOURCE_GROUP \
-  --template-file "02_storage.json" \
-  --parameters vhdBlobURL="${VHD_BLOB_URL}" \
-  --parameters baseName="$INFRA_ID"
+STORAGE_ACCOUNT_ID=$(az storage account show -g $RESOURCE_GROUP --name $ACCOUNT_NAME --query id -o tsv)
+
+# Check if it's the new template using Image Galleries instead of Managed Images
+if grep -qs "Microsoft.Compute/galleries" 02_storage.json; then
+  AZ_ARCH=$(echo $OCP_ARCH | sed 's/x86_64/x64/;s/aarch64/Arm64/')
```
Prashanth684, line 227: i think OCP_ARCH will be passed in as arm64 - aarch64 is only used in the context of rhcos/linux

Prashanth684: ok ignore this comment....you do use that to get the image from the rhcos json

MayXuQQ:
create UPI cluster  succeed, and after scaleup, worker created succeed. 

> $ oc get nodes
> NAME                                    STATUS   ROLES                  AGE     VERSION
> maxu-u12-2-5zskw-master-0               Ready    control-plane,master   56m     v1.25.4+77bec7a
> maxu-u12-2-5zskw-master-1               Ready    control-plane,master   56m     v1.25.4+77bec7a
> maxu-u12-2-5zskw-master-2               Ready    control-plane,master   56m     v1.25.4+77bec7a
> maxu-u12-2-5zskw-worker-eastus-1        Ready    worker                 39m     v1.25.4+77bec7a
> maxu-u12-2-5zskw-worker-eastus-2        Ready    worker                 36m     v1.25.4+77bec7a
> maxu-u12-2-5zskw-worker-eastus3-l9bkm   Ready    worker                 5m29s   v1.25.4+77bec7a

/label qe-approved

ci-operator/step-registry/upi/install/azure/upi-install-azure-commands.sh
```diff
@@ -218,10 +220,23 @@ az network private-dns link vnet create -g $RESOURCE_GROUP -z ${CLUSTER_NAME}.${
 
 echo "Deploying 02_storage"
 VHD_BLOB_URL=$(az storage blob url --account-name $ACCOUNT_NAME --account-key $ACCOUNT_KEY -c vhd -n "rhcos.vhd" -o tsv)
-az deployment group create -g $RESOURCE_GROUP \
-  --template-file "02_storage.json" \
-  --parameters vhdBlobURL="${VHD_BLOB_URL}" \
-  --parameters baseName="$INFRA_ID"
+
+# Check if it's the new template using Image Galleries instead of Managed Images
+if grep -qs "Microsoft.Compute/galleries" 02_storage.json; then
```
patrickdillon, line 225: I am reluctant to bring this up without a better suggestion, but I worry that putting conditionals in a template like this can be risky in terms of legibility/debugability. Our old system of templates became so convoluted that it was hard to follow the execution path, so i am perhaps overly hesitant whenever I see any conditionals.

The change introduced here is simple, so I think it is better than any of the alternatives I can come up with, which would be overkill or more confusing than this.

patrickdillon (commented):
/approve
/lgtm

ci-operator/step-registry/upi/install/azure/upi-install-azure-commands.sh
r4f4: This was the best idea I could come up with to make this work with both Managed Images and Galleries. Relying on the installer or release payload versions didn't feel reliable enough. Once 4.12 migrates to Galleries and this workflow is run on 4.12+, then we can drop this conditional. But I agree with you this is not ideal.

r4f4: @patrickdillon @barbacbd  pointed out an interesting idea. We could use the `contentVersion` from the ARM template [1] [2]. It still a conditional but at least it checks for a version instead of a string inside the template. Then it's up to the Installer team to make sure that version is updated accordingly when there are changes to the templates.

[1] https://github.com/openshift/installer/blob/master/upi/azure/02_storage.json#L3
[2] https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/syntax#template-format

PR #34696, opened 2022-12-09
Switched cluster pool from obs-china to obs-usa

dislbenn:
Signed-off-by: dislbenn <dbennett@redhat.com>


subbarao-meduri (approved):
/lgtm

PR #34703, opened 2022-12-09
Switch GCP credentialsMode to Passthrough

patrickdillon:
Mint mode causes custom roles to be created and we are currently hitting a limit. This is intended as a temporary fix.


patrickdillon:
Reverting https://github.com/openshift/machine-api-operator/pull/1087 would have a smaller splash than switching almost all gcp ci jobs to passthrough. I will put up a PR for that.

PR #34707, opened 2022-12-09
SDCICD-776 AWS x1,x2 Instance Types

MrSantamaria:
[SDCICD-776](https://issues.redhat.com/browse/SDCICD-776)
Removed the pipelines from SDCICD-775. 
Set the new Instances to test over the weekend.


ci-operator/jobs/openshift/osde2e/openshift-osde2e-main-periodics.yaml
```diff
@@ -157,17 +190,15 @@ periodics:
       - /osde2e
       env:
       - name: CONFIGS
-        value: aws,stage,e2e-suite,region-random
-      - name: OCM_CCS
+        value: rosa,e2e-suite
```
ritmun, line 193: don't we need stage here, since the the job name suggests `rosa-stage` ?

PR #34713, opened 2022-12-11
MGMT-12789: set the right jira versions for assisted components

osherdp:
It makes no sense to have ACM versions as target-versions for the jira-lifecycle plugin, as either way there's no way to set those in Jira.

This changes those overrides, so that we can use openshift versions when such tickets (OCPBUGS-*) are linked.


AlexNPavel:
This removes handling of all the `release-ocm` branches by the plugin. The only branch that will be handled by the plugin with this change will be the main/master branches, which will follow the default rules for main/master. Is this correct?

osherdp:
> This removes handling of all the `release-ocm` branches by the plugin. The only branch that will be handled by the plugin with this change will be the main/master branches, which will follow the default rules for main/master. Is this correct?

@AlexNPavel Removing handling of `release-ocm` is intended as they are related to the ACM product

Although I thought it should also handle the ``release-4.y`` branches. Doesn't it?

AlexNPavel:
> Although I thought it should also handle the `release-4.y` branches. Doesn't it?

Yes, if those repos have `release-4.y` or `openshift-4.y` branches, those will also follow the default rules from this config.

AlexNPavel (approved):

osherdp:
> > Although I thought it should also handle the `release-4.y` branches. Doesn't it?
> 
> Yes, if those repos have `release-4.y` or `openshift-4.y` branches, those will also follow the default rules from this config.

Ah good. Then that's exactly what we want
Thank you!

PR #34715, opened 2022-12-11
telco5g: Fix kcli parameters for nightly jobs

sshnaidm:
Openshift image parameter is overwritten only when version is "ci", this is used when we run a nightly release check job, to run a usual periodic job on latest available nightly we need to use
tag with version and "version: nightly".


sshnaidm:
cc @SchSeba

PR #34724, opened 2022-12-12
Delete assisted-baremetal jobs from capi-provider-agent

eranco74:
These jobs constently fail, seems that hypershift requires OpenShift as managment cluster for creating a functional HostedCluster. Must of the functionality tested by these jobs is coverged by the operator workflow jobs jobs


osherdp:
What about the job in another repos? I guess it doesn't make sense to maintain them as well

eranco74:
> t doesn't make sense to maintain them as w

Pretty sure we already removed it from all other repos but I'll double check

eranco74:
OK, there's still one optional job in assisted-test-infra.
I'll remove it as well

osherdp (approved):

PR #34729, opened 2022-12-12
NO-ISSUE: Cleanup fix uid script in assisted

adriengentil:
The workaround script fix_uid.sh is not required anymore, the root cause
is fixed in OCP.


adriengentil:
/pj-rehearse pull-ci-openshift-assisted-service-master-edge-e2e-metal-assisted

osherdp (approved):

adriengentil:
so it looks like the issue is still there :thinking:

PR #34730, opened 2022-12-12
MGMT-12804: Use ocm-2.7 tag instead of latest for assisted index

adriengentil:
Since https://github.com/openshift/release/pull/34728, we can use
ocm-2.7 tag for jobs on ocm-2.7 branch for the assisted index.


osherdp (approved):

osherdp:
Don't forget to update hypershift repo next

PR #34744, opened 2022-12-12
MGMT-12439: Add basic job to test heterogeneous infrastructure

adriengentil:
Add first building blocks to test assisted with heterogeneous clusters.
It will first be used to test
https://github.com/openshift/assisted-test-infra/pull/1949.

### DIFF
--- a/ci-operator/step-registry/assisted/equinix/assisted-equinix-workflow.yaml
+++ b/ci-operator/step-registry/assisted/equinix/assisted-equinix-workflow.yaml
@@ -1,0 +1,19 @@
+workflow:
+  as: assisted-equinix
+  steps:
+    cluster_profile: packet-assisted
+    allow_best_effort_post_steps: true
+    allow_skip_on_success: true
+    pre:
+      - ref: assisted-equinix-setup
+      - chain: assisted-common-pre
+    test:
+      - ref: assisted-baremetal-conf
+      - ref: assisted-baremetal-test
+    post:
+      - ref: assisted-common-gather
+      - ref: assisted-equinix-teardown
+    env:
+      TEST_SUITE: minimal
+  documentation: |-
+    This workflow installs an OCP cluster by running assisted-installer on an equinix metal server.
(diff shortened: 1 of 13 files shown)


ci-operator/step-registry/assisted/equinix/assisted-equinix-workflow.yaml
```diff
@@ -0,0 +1,18 @@
...
+  steps:
+    cluster_profile: packet-assisted
+    allow_best_effort_post_steps: true
+    allow_skip_on_success: true
+    pre:
+      - ref: assisted-equinix-setup
+      - chain: assisted-common-pre
+    test:
+      - ref: assisted-baremetal-conf
+      - ref: assisted-baremetal-test
```
osherdp, line 12: (not necessarily in the scope of this PR) in the future we might want to move to ``assisted-common-verify``

ci-operator/step-registry/assisted/equinix/assisted-equinix-workflow.yaml
```diff
@@ -0,0 +1,18 @@
...
+    post:
+      - ref: assisted-common-gather
```
osherdp, line 14: am I wrong or we're not tearing down the machines?

osherdp, line 14: should we add ``baremetalds-packet-gather-metadata``?

osherdp (changes requested):

ci-operator/step-registry/assisted/equinix/assisted-equinix-workflow.yaml
adriengentil: Yes, we already have a ticket for that (assigned to me :see_no_evil: )

ci-operator/step-registry/assisted/equinix/assisted-equinix-workflow.yaml
adriengentil: I'll need to create ticket about that, we now have to 2 machines to report, I don't know yet how to handle the case.

adriengentil: :man_facepalming:

osherdp (approved):

PR #34770, opened 2022-12-13
Setting retries for Vsphere-sdn for install-analysis

deepsm007:
Related to https://github.com/openshift/release/pull/34768 as there is only 1 job for vsphere sdn

/cc @xueqzhan

### DIFF
--- a/core-services/release-controller/_releases/priv/release-ocp-4.12.json
+++ b/core-services/release-controller/_releases/priv/release-ocp-4.12.json
@@ -475,6 +475,7 @@
         },
         "vsphere-sdn": {
             "disabled": true,
+            "maxRetries": 3,
             "optional": true,
             "prowJob": {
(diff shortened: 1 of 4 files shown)


jcpowermac:
@deepsm007 what will be the impact to our vSphere capacity with this?
I am a little concerned about adding more jobs to an already 100% utilized environments

deepsm007:
> @deepsm007 what will be the impact to our vSphere capacity with this? I am a little concerned about adding more jobs to an already 100% utilized environments

We are working on improving the install signal and retries will only affect in case of failures to provide more info if the payload is being tested efficiently. Looking at the history [here](https://prow.ci.openshift.org/job-history/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-nightly-4.13-e2e-vsphere-sdn) it seems pretty stable and shouldn't affect the capacity.

deepsm007:
To avoid capacity issues we will remove the vsphere-sdn from install-analysis job instead of adding retries

PR #34787, opened 2022-12-13
installer: drop linter timeout from invocation

r4f4:
We will configure it via config file in the installer repo.


r4f4:
/pj-rehearse ack
`pull-ci-openshift-installer-master-golint` is expected to fail since removing the `--timeout 5m` argument, it defaults to a 1m timeout. The timeout will be set via https://github.com/openshift/installer/pull/6601 once this merges.

PR #34839, opened 2022-12-15
FDN-112: Add jira/valid-reference label for all repos

bparees:
intent is to make this label available for use on all repos, for use as a future PR merge gating mechanism

### DIFF
--- a/core-services/prow/02_config/_labels.yaml
+++ b/core-services/prow/02_config/_labels.yaml
@@ -369,6 +369,12 @@ default:
       target: prs
       prowPlugin: backport-verifier
       addedBy: prow
+    - color: 15dd18
+      description: Indicates that this PR references a valid Jira ticket of any type.
+      name: jira/valid-reference
+      target: prs
+      prowPlugin: jira
+      addedBy: prow
     - color: e11d21
       description: Indicates that a referenced Jira bug is invalid for the branch this PR is targeting.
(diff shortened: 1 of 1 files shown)


bparees:
@openshift/test-platform ptal

core-services/prow/02_config/_labels.yaml
```diff
@@ -369,6 +369,12 @@ default:
       target: prs
       prowPlugin: backport-verifier
       addedBy: prow
+    - color: 15dd18
+      description: Indicates that this PR references a valid Jira ticket of any type.
```
wking, line 373: It sounds like pulls referencing a valid Jira bug will have both `jira/valid-bug` and `jira/valid-reference`?  Possibly worth pointing that out as an explicit choice in the commit message, so folks don't wonder about trying to remove overlap later.

bparees: > It sounds like pulls referencing a valid Jira bug will have both jira/valid-bug and jira/valid-reference

yes.

i'll fix up the commit message while also changing this to a unique color (forgot to do that)

bparees: actually maybe the color (Green) is as it should be.

bparees: commit text updated

droslean (approved):

PR #34845, opened 2022-12-15
Revert "TRT-732: Fixing the minimum count for vsphere and return code error"

xueqzhan:
Reverts openshift/release#34768


dgoodwin:
/lgtm

Fix worked, we're just now failing payloads. Revert until we can get the response cleaned up.

dgoodwin:
TRT-743

PR #34853, opened 2022-12-15
Sync arm64 4.12 profiles

aleskandro:


ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__arm64-nightly.yaml
```diff
@@ -174,16 +210,17 @@ tests:
...
       COMPUTE_NODE_TYPE: m6g.xlarge
       CONTROL_PLANE_INSTANCE_TYPE: m6g.xlarge
-      E2E_RUN_TAGS: '@aws-ipi and @arm64 and @disconnected and @network-ovnkubernetes
-        and not @fips'
+      E2E_RUN_TAGS: '@arm64 and @aws-ipi and @disconnected and @network-ovnkubernetes
+        and @singlenode and not @fips and not @console'
       OCP_ARCH: arm64
       TAG_VERSION: '@4.12'
       TEST_FILTERS: ~ChkUpgrade&;~NonPreRelease&;~Serial&;~Disruptive&;~ConnectedOnly&
+      TEST_SCENARIOS: MCO|ETCD|PSAP|SDN|API_Server|Authentication|Cluster_Operator|Cluster_Infrastructure|OLM|Network_Edge|Operator_SDK|Workloads|Image_Registry|Container_Engine_Tools|NODE|OTA|PerfScale|Cluster_Observability|Security_and_Compliance|LOGGING
```
kuiwang02, line 218: @aleskandro you do not take default value to remove STORAGE subteam. just double confirm with you: why remove STORAGE? because of this SNO?
if it is caused by SNO, test `aws-ipi-disconnected-private-sno-p3-f28-destructive` is sno, but I do not find STORAGE is removed.

this comment apply to other tests of this PR and also apply to other PRs in https://coreos.slack.com/archives/CPJQCSN49/p1671141487692949

kuiwang02:
@aleskandro general comment for TEST_FILTERS on this PR and the PRs in https://coreos.slack.com/archives/CPJQCSN49/p1671141487692949.

if the cluster is not disconnect, please take default value of golang step and no need to set it in the test.
if the cluster is disconned, need to change `~DisconnectedOnly&` to `~ConnectedOnly&` based on default value of golang step to set it in the test.

aleskandro:
Hi @kuiwang02 the filters should be the same as for the equivalent amd64 definitions, aren't they?

If yes, i expect others will change them later together in a follow up. This pr is to sync the two architectures' tests definitions

kuiwang02:
> Hi @kuiwang02 the filters should be the same as for the equivalent amd64 definitions, aren't they?
> 
> If yes, i expect others will change them later together in a follow up. This pr is to sync the two architectures' tests definitions

@aleskandro yes. maybe some amd64 tests are not good for the definition.
anyway, for both arm64 and amd64, the usage of TEST_FILTERS of golang step are same.
```console
if the cluster is not disconnect, please take default value of golang step and no need to set it in the test.
if the cluster is disconned, need to change ~DisconnectedOnly& to ~ConnectedOnly& based on default value of golang step to set it in the test.
```
So,
if TEST_FILTERS of some amd64 tests of golang step are not following it, they should be corrected.  maybe qe-productivity team need to correct it.
for TEST_FILTERS of arm64 tests of golang step, please do it rightly at the beginning.


anyway, they are my suggestions. if qe-productivity and you do not want to follow it now when making test definition, it is up to you because you are the owner of these tests.
cc @YuLi517 @jhou1 @chengzhang1016

kuiwang02:
@jianzhangbjz @Xia-Zhao-rh FYI. please also be aware of the above.

aleskandro:
> anyway, they are my suggestions. if qe-productivity and you do not want to follow it now when making test definition, it is up to you because you are the owner of these tests. you could decide when to change it to follow the correct definition rule.

I'm ok with your review @kuiwang02, I'm just saying the purpose of this PR is different than fixing the definition of the already enabled tests.

The purpose is to cover a gap that the team has not taken care of: having the same test definitions for both architectures.

So, this PR will make them the same and only covers any diff regarding the initialization of arm64 clusters.

Regarding the bugfix you say:
1. It should need another kind of testing (on amd64 as well)
2. It should modify the amd64 tests definitions
3. It should be part of another PR just because it's correct and good practice that a PR covers one topic and another the other topic.

aleskandro:
I'm not owner of these tests, I'm just responsible of the multi-arch qe efforts: to ensure testing is done against arm64 and amd64 the same way.

kuiwang02:
@aleskandro 
I am sorry for
1, I thought you are the owner of tests when I saw you added them.
2, I just am aware that the usage of filter is not good.

I still would like to enhance it ASAP, so I make https://github.com/openshift/release/pull/34867 to enhance it.
after it is merged, you could rebase the code to make good filter for the PRs in the https://coreos.slack.com/archives/CPJQCSN49/p1671141487692949

aleskandro:
As stated there and in slack, it would be ideal the contrary. To merge this and the related PRs ASAP and then to enhance to already-enabled tests.

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__arm64-nightly.yaml
liangxia: This is because `baselinecaps-none`, which does not install storage operator.
@duanwei33 has more context/detail.

aleskandro:
Refers ARMOCP-466

PR #34856, opened 2022-12-15
Remove template AWS UPI job

patrickdillon:
This job has been replaced by a workflow/steps job.

### DIFF
--- a/core-services/release-controller/_releases/priv/release-ocp-4.10.json
+++ b/core-services/release-controller/_releases/priv/release-ocp-4.10.json
@@ -99,7 +99,7 @@
             "disabled": true,
             "optional": true,
             "prowJob": {
-                "name": "release-openshift-ocp-installer-e2e-aws-upi-4.10-priv"
+                "name": "periodic-ci-openshift-release-master-nightly-4.10-e2e-aws-upi-priv"
             }
         },
         "azure": {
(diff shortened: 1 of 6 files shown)


bradmwilliams:
@patrickdillon In order to clear the `release-controller-config` failure, you'll need to either:
1. Update the following lines to point to the new periodic job that calls the respective workflow
2. Remove the following lines all together

https://github.com/openshift/release/blob/ccb8b19c8181988525d801e7a39e4b3d514f5153/core-services/release-controller/_releases/release-ocp-4.12.json#L161-L166
and then run a `make release-controllers` from the root of the repo and include the changes.

patrickdillon:
@DennisPeriquet your PR https://github.com/openshift/release/pull/34926 fixed this on master. I believe we "backported" the step registry periodics back to 4.9 so I'm updating accordingly in this PR but not sure if it's correct. Would you PTAL?

DennisPeriquet:
@patrickdillon sure thing.   I believe the PR looks good.

But, I wonder if it's possible to remove [these](https://github.com/DennisPeriquet/release/blob/fe9dc9488046a968e43f70cb11d93c9447f7dc51/core-services/template-deprecation/_allowlist.yaml#L483-L491) lines because from what I understand, that file tracks the tests that still need conversion from templates to steps.

bradmwilliams (approved):

PR #34860, opened 2022-12-15
Adds @amd64 for multi-arch purposes in the amd64 QE e2e and upgrade t…

aleskandro:
…ests


liangxia:
Actually, they are intentionally at the moment.
We still have a lot of test cases (1000+ as of 2022-12-19) do not have proper `Arch` value in Polarion, which means those test cases do not have tag amd64/arm64/... Those test cases will be missed if we add `@amd64` explicitly there.
Once we have fixed the `arch` field in polarion for all/most of test cases, and remap that into tags in the code repo, it's good to have this PR then.

PR #34861, opened 2022-12-15
AGENT-350: Handle the case where there are no agent-gather files

zaneb:
If the cluster succeeds there will be no agent-gather files produced, since these are only available from the agent image (which will be long-since gone).

Handle the case where even after running 'make agent-gather', there are no files to copy.


zaneb:
/retitle AGENT-350: Handle the case where there are no agent-gather files
/pj-rehearse

PR #34867, opened 2022-12-16
[CIGolang]enhance filters

kuiwang02:
/cc @Xia-Zhao-rh 

I am sorry I just am aware that the usage of TEST_FILTERS is not good.

for chain openshift-e2e-test-qe, openshift-e2e-test-qe-destructive and openshift-upgrade-qe-test-disconnected, the usage of TEST_FILTERS of golang step are better to be the following
```console
if the cluster is not disconnect, please take default value of golang step and no need to set it in the test.
if the cluster is disconned, need to change ~DisconnectedOnly& to ~ConnectedOnly& based on default value of golang step to set it in the test.
```

beside enhancing it, 
1, some FILTERS are not complete which miss `~HyperShiftMGMT&`. The PR also addd it.
2, some disconnect env use wrong filter, the PR also correct it.

cc @aleskandro @liangxia @jianzhangbjz

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.10__amd64-nightly.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.10__amd64-nightly.yaml
@@ -96,7 +96,6 @@ tests:
       BASE_DOMAIN: qe.devcluster.openshift.com
       E2E_RUN_TAGS: '@aws-ipi and @network-openshiftsdn and not @fips and @destructive'
       TAG_VERSION: '@4.10'
-      TEST_FILTERS: ~ChkUpgrade&;~NonPreRelease&;Serial;Disruptive;~DisconnectedOnly&
     test:
     - chain: openshift-e2e-test-qe-destructive
     workflow: cucushift-installer-rehearse-aws-ipi-byo-iam-role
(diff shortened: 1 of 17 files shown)


aleskandro:
Can we hold until #34867 and related are merged and do the same for them.

Rebasing all the other PRs is a pain with respect to replace all the filters once as here.

I already requested to hold PRs until the multiarch issue is solved in https://coreos.slack.com/archives/CPJQCSN49/p1671141487692949

kuiwang02:
@aleskandro when will those PRs be merged? 
if it is not long time, we could wait it.
if it is long time (for example, by end of next week), we could not wait it. we can not hold to merge PR for one week.

aleskandro:
If you merge this before my ones, I'll not only have to rebase (that is not an issue) but redo a manual copy and paste and review from scratch.
If you merge this after, then it will be straightforward.

The time doesn't depend on me: if the installer, upgrade, or productivity teams want to also test every single profile, it's up to them.

I believe we can already merge them, I just need lgtm and approval. Then the CI watchers loop can help raise or solve any further issues caused by previous work that broke the multi-arch enablement.

Also, if some profiles need to be removed or fixed, it's better to do it in dedicated  PRs rather than in the ones I opened yesterday.

I understand we cannot block all the PRs for a long time, but we are not testing against arm64 stable builds till now, especially the profiles that have never been created on Jenkins and this is a risk from ARM QE side.

kuiwang02:
OK. I can wait here for this PR.
when those PRs are merged, please notice me. hope it is done ASAP. this PR does not only enhance filters. it also correct few wrong filters as I mentioned in https://github.com/openshift/release/pull/34867#issue-1499908478 (I do not know who added the test with wrong filters.)

kuiwang02:
except for https://github.com/openshift/release/pull/34860 which is held, other PRs are merged.
so, 
/unhold

PR #34868, opened 2022-12-16
Use golang 1.18 on main

ReToCode:
With 1.27 golang 1.18 is the default, so we should probably also update the build_root for main.


pierDipi:
/pj-rehearse pull-ci-openshift-knative-serverless-operator-main-4.11-images

PR #34877, opened 2022-12-16
add -ovn- to machine-config-operator presubmits

jluhrsen:
this resolves a flaking test case that checks if the running CNI (ovn, in this case) is part of the job name.

also:
- changes 'agnostic' to 'aws' in the relevant upgrade jobs to match current naming conventions
- adds gcp to relevant upgrade job names for same reason as above

### DIFF
--- a/ci-operator/config/openshift/machine-config-operator/openshift-machine-config-operator-master.yaml
+++ b/ci-operator/config/openshift/machine-config-operator/openshift-machine-config-operator-master.yaml
@@ -74,13 +74,13 @@ tests:
   commands: make test-unit
   container:
     from: src
-- as: e2e-alibabacloud
+- as: e2e-alibabacloud-ovn
   optional: true
   skip_if_only_changed: ^docs/|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|OWNERS_ALIASES|LICENSE)$
   steps:
     cluster_profile: alibabacloud
     workflow: openshift-e2e-alibabacloud
-- as: e2e-aws
+- as: e2e-aws-ovn
   skip_if_only_changed: ^docs/|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|OWNERS_ALIASES|LICENSE)$
   steps:
     cluster_profile: aws
@@ -131,7 +131,7 @@ tests:
         requests:
           cpu: 100m
     workflow: ipi-gcp
-- as: e2e-agnostic-upgrade
+- as: e2e-aws-ovn-upgrade
   skip_if_only_changed: ^docs/|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|OWNERS_ALIASES|LICENSE)$
   steps:
     cluster_profile: aws
--- a/ci-operator/config/openshift/machine-config-operator/openshift-machine-config-operator-master__okd-scos.yaml
+++ b/ci-operator/config/openshift/machine-config-operator/openshift-machine-config-operator-master__okd-scos.yaml
@@ -39,7 +39,7 @@ resources:
       cpu: 100m
       memory: 200Mi
 tests:
-- as: e2e-aws
+- as: e2e-aws-ovn
   optional: true
   steps:
     cluster_profile: aws
@@ -57,7 +57,7 @@ tests:
         requests:
           cpu: 100m
     workflow: ipi-gcp
-- as: e2e-upgrade
+- as: e2e-gcp-ovn-upgrade
   optional: true
   steps:
     cluster_profile: gcp
--- a/ci-operator/config/openshift/machine-config-operator/openshift-machine-config-operator-release-4.12.yaml
+++ b/ci-operator/config/openshift/machine-config-operator/openshift-machine-config-operator-release-4.12.yaml
@@ -125,7 +125,7 @@ tests:
         requests:
           cpu: 100m
     workflow: ipi-gcp
-- as: e2e-agnostic-upgrade
+- as: e2e-aws-ovn-upgrade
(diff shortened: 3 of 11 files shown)


ci-operator/config/openshift/machine-config-operator/openshift-machine-config-operator-master.yaml
```diff
@@ -131,7 +131,7 @@ tests:
         requests:
           cpu: 100m
     workflow: ipi-gcp
-- as: e2e-agnostic-upgrade
+- as: e2e-aws-ovn-upgrade
```
yuqi-zhang, line 134: So, as I understand it, the `agnostic` upgrade job was named as such to act as a front for the platform for upgrade.

Historically (I believe) we started on AWS, moved to GCP, then Azure, then finally earlier this year switched back to AWS. And the agnostic job naming remained as such so we can make the switch behind the scenes a bit easier.

Are you saying that we will no longer do so in the future and AWS-ovn is the catch-all platform for upgrade testing?

jluhrsen: yeah, I think originally the idea for 'agnostic' was to allow for the underlying platform to be
anything. However, in the past year we have changed the default CNI to ovn and also implemented
two new test cases that actually check the name of the job vs the underlying platform and CNI.

```
: [sig-ci] [Early] prow job name should match platform type [apigroup:config.openshift.io] [Suite:openshift/conformance/parallel]
: [sig-ci] [Early] prow job name should match network type [apigroup:config.openshift.io] [Suite:openshift/conformance/parallel]
```

They will only flake if they don't match, but dialing in the job names will get rid of that flake.

No, aws-ovn is not the catch-all platform for upgrade. We run other platforms and CNIs as
periodics that the release team monitors. MCO can run any/all of them too. If we want the
GCP upgrade job here, we'd just add -gcp- to the job name. up to the MCO team.

yuqi-zhang: Ah, sorry, to clarify, there are 3 "main" jobs gating MCO PRs:
1. e2e-aws
2. e2e-gcp-op (this is MCO e2e tests)
3. e2e-agnostic-upgrade (this is the only upgrade test we have as gating)

With this PR we would effectively be renaming the blocking job to be aws-ovn (not that it makes a difference in this case since agnostic is already doing that), which may cause confusion to MCO repo contributors not aware of this change. Since we're here, I just wanted to ask if that is what we want in terms of both naming and platform (aws-ovn as the sole blocking upgrade job that runs)

jluhrsen: right, this won't change any functionality of what presubmits are blocking or not. it's just renaming them to
fix a flaking test case. It's something we've been working on for all of openshift jobs for many months now. I'm looking for jobs that are flaking on a weekly basis and making PRs to fix them up a few at a time. I just happened
to hit MCO this past week :)

so for your list:

1. e2e-aws --> e2e-aws-ovn
2. e2e-gcp-op --> _no change_
3. e2e-agnostic-upgrade --> e2e-aws-ovn-upgrade

yuqi-zhang: Ok, I'll let the team also take a look in case there are any other concerns, thanks!

jkyros: Is this naming scheme by chance captured anywhere in some sort of "canonical naming standard" for jobs so that if/when someone adds a new one, it gets named right? 

Or is this kind of a "well, just name it wrong and eventually someone will clean up the flake" sort of thing? :smile:  

(I'm probably just ignorant and there is probably something somewhere I should have read that I didn't) 

Until this very moment, I assumed the names didn't matter to anything but humans -- should we have known this?

jluhrsen: there was some conversation about some sort of naming structure (notes here), but I don't think anything ever
was finalized. The current e2e-<platform>-<cni> format is from that afaik, but really came about is being
important when we moved from SDN to OVN as the default CNI and we wanted a more proper way of making
sure we didn't have jobs running with one CNI that we assumed was running the other. So the check was added
to make sure the job name matched the actual CNI running. With that we  set about to rename the periodic
jobs (especially the ones that gate release payloads).

anyway that's the recent history, and no it's understandable that this was not really known by everyone.

btw, if there is some reason to not rename your jobs, I don't think anyone will force it. you'll just have flakes
to look at which isn't the worse. I do think some jobs have logic that will check historical data that we save
somewhere to make pass/fail decisions. Those jobs will check based on job names.... I don't know these
details exactly, but it's one reason I set off to try to help each repo get their names fixed up.

jkyros: Oh, okay, cool. Thanks a ton for the background and the doc link -- that makes sense. 

Downside sounds like we'll have to rename our jobs going forward if we move them to a different platform (like we moved the agnostic test) -- we'll just have to remember to do that. 

I think I'm convinced this adds value for the org and we probably need to eat it, but to Jerry's point it's going to be super weird for the MCO since everyone has gotten used to `e2e-agnostic-upgrade` being *the* upgrade job by name. We'll just have to make sure that the MCO socializes that "it's different now". :smile:

jluhrsen:
https://issues.redhat.com/browse/TRT-487

yuqi-zhang (approved):
We discussed it briefly with the team and there were no additional concerns. Let's go ahead and merge this, thanks for fixing this!

/lgtm

PR #34880, opened 2022-12-16
remove duplicate aws-ovn-upgrade presubmit in mco

jluhrsen:
this job is the same as what exists below and currently called e2e-agnostic-upgrade [0], which is also
configured to be skipped if only docs/OWNERS/etc files are changed which is a little more efficient with CI resources.

also note that the e2e-agnostic-upgrade job will be renamed to e2e-aws-ovn-upgrade job eventually [1].

[0] https://github.com/openshift/release/blob/33699caf68273b05077baee7ecb300ebbbc5d25b/ci-operator/config/openshift/machine-config-operator/openshift-machine-config-operator-master.yaml#L134-L138 [1] https://github.com/openshift/release/pull/34877


yuqi-zhang (approved):
/lgtm

PR #34887, opened 2022-12-18
Set up an env variable to allow skipping log verification

mshitrit:
Set up an env variable to allow skipping log verification in ocp 4.12


mshitrit:
/pj-rehearse ack
4.12 failure is expected as this PR is part of the setup that meant to prevent that fauilure.

PR #34895, opened 2022-12-19
Optimize test frequency for 4.12 tests

liangxia:
After sync tests from amd64 to arm64, move 50% of the tests from amd64 to arm64.


liangxia:
Rebase and resolve conflicts.

PR #34904, opened 2022-12-19
Adding non-e2e tests to KMM.

ybettan:
Co-authored-by: Enrique Belarte Luque <ebelarte@redhat.com>

---

Fixes https://github.com/rh-ecosystem-edge/kernel-module-management/issues/237

* `ci/rehearse/rh-ecosystem-edge/kernel-module-management/main/check-commits-count` is failing because the actual commits number to the KMM repo is `0`, therefore, different than expected by the test.
* `ci/rehearse/rh-ecosystem-edge/kernel-module-management/main/check-api-changes` is failing because there are actual diffs in the KMM repo - which means the test (which is making sure all generated files were pushed with the commit) is working as expected and the fix should be applied later to KMM's repo.


ybettan:
/cc @yevgeny-shnaidman 
/cc @qbarrand 

I need `approve` and also `lgtm` please.
My `approve` was removed because (I think) at some point I also pushed a Dockerfile for testing (which is not removed) but seems like it is not adding back the `approve` label.

PR #34912, opened 2022-12-19
OpenStack: Ensure backward compatibility with older bash

mandre:
`test -v` only gained ability to understand array references in bash 4.3. This could be a problem to us if we're running older OCP images where bash is still at version 4.2. This could for example be the case where clusterbot still runs OCP 4.5 based images.

### DIFF
--- a/ci-operator/step-registry/openstack/conf/resourcenames/openstack-conf-resourcenames-commands.sh
+++ b/ci-operator/step-registry/openstack/conf/resourcenames/openstack-conf-resourcenames-commands.sh
@@ -60,7 +60,7 @@ if [[ -z "${OPENSTACK_EXTERNAL_NETWORK:-}" ]]; then
 		exit 1
 	fi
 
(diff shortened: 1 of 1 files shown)


pierreprinetti (approved):

pierreprinetti:
No 4.5 test listed; we probably deactivated all these.
/pj-rehearse periodic-ci-shiftstack-shiftstack-ci-main-periodic-4.10-e2e-openstack-parallel periodic-ci-shiftstack-shiftstack-ci-main-periodic-4.5-e2e-openstack-parallel
/remove-lifecycle stale

mandre:
We don't _need_ this patch anymore, but it doesn't hurt merging it either.

pierreprinetti:
/lgtm cancel
The test didn’t pass

pierreprinetti:
The issue here is that some of the `compute_azs` items contain empty strings, and the script should accept them. In your code, the empty string is treated like unset.

PR #34929, opened 2022-12-19
Add slack reporter_config for hypershift-stage

dustman9000:
This will send Slack notifications to the sd-hypershift-info slack channel.


MrSantamaria:
@dustman9000 but this won't trigger a threshold right? From what I read every pipeline status will shoot an alarm into sd-hypershift-info

dustman9000:
This will not use a threshold yet.

PR #34938, opened 2022-12-20
NHC: use Go 1.19 build root

slintes:
and use stable channel


slintes:
/pj-rehearse skip

test will run on https://github.com/medik8s/node-healthcheck-operator/pull/153

PR #34946, opened 2022-12-20
make the aws-ovn-windows tests required for cno and ovn-kube 4.11-4.8

JacobTanenbaum:
when https://github.com/openshift/ovn-kubernetes/pull/1415 was merged into ovn-kube because there were no required windows testing the failures where ignored and introduced a regression. Make basic windows testing required so that similar regressions are not introduced in the future.


saifshaikh48 (commented):
/lgtm

PR #31430, opened 2022-08-17
OCPBUGS-219 Add ovn to certain jobs to avoid flakes

candita:
Update for new tests that check the cluster's network type matches up with the job name, 4.12 and above.  Update e2e-upgrade test name to e2e-aws-upgrade to standardize in files where it is not already like that.

https://coreos.slack.com/archives/C01CQA76KMX/p1660753233493069?thread_ts=1659554645.706879&cid=C01CQA76KMX

PR #34949, opened 2022-12-20
external dns: release 1.1

alebedev87:
The CI configuration for the release branch `1.1` of ExternalDNS and ExternalDNS Operator.
The mirroring for `1.1` release is made to `quay.io/external-dns-operator/external-dns:1.1` (not `latest`).

### DIFF
--- a/ci-operator/config/openshift/external-dns-operator/openshift-external-dns-operator-release-1.1.yaml
+++ b/ci-operator/config/openshift/external-dns-operator/openshift-external-dns-operator-release-1.1.yaml
@@ -1,0 +1,138 @@
+base_images:
+  base:
+    name: ubi-minimal
+    namespace: ocp
+    tag: "8"
+  operator-index:
+    name: redhat-operator-index
+    namespace: ci
+    tag: v4.9
+build_root:
+  from_repository: true
+images:
+- dockerfile_path: Dockerfile
+  from: base
+  to: external-dns-operator
+operator:
+  bundles:
+  - as: external-dns-operator-bundle
+    base_index: operator-index
+    dockerfile_path: Dockerfile.bundle
+  substitutions:
+  - pullspec: quay.io/openshift/origin-external-dns-operator:latest
+    with: pipeline:external-dns-operator
+releases:
+  initial:
+    integration:
+      name: "4.12"
+      namespace: ocp
+  latest:
+    integration:
+      include_built_images: true
+      name: "4.12"
+      namespace: ocp
+resources:
+  '*':
+    requests:
+      cpu: 100m
+      memory: 200Mi
+tests:
+- as: verify
+  commands: |
+    export GOLANGCI_LINT_CACHE=$(mktemp -d)
+    make verify
+  container:
+    from: src
+- as: unit
+  commands: |
+    export XDG_CACHE_HOME=$(mktemp -d)
+    make test
+  container:
+    from: src
+- as: e2e-aws-ovn-operator
+  steps:
+    cluster_profile: aws
+    dependencies:
+      OO_INDEX: ci-index-external-dns-operator-bundle
+    env:
+      OO_CHANNEL: stable-v1.1
+      OO_INSTALL_NAMESPACE: external-dns-operator
+      OO_PACKAGE: external-dns-operator
+      OO_TARGET_NAMESPACES: '!install'
+    test:
+    - as: test
+      cli: latest
+      commands: make test-e2e
+      from: src
+      resources:
+        requests:
+          cpu: 100m
+    workflow: optional-operators-ci-aws
+- as: e2e-gcp-ovn-operator
+  steps:
+    cluster_profile: gcp
+    dependencies:
+      OO_INDEX: ci-index-external-dns-operator-bundle
+    env:
+      OO_CHANNEL: stable-v1.1
+      OO_INSTALL_NAMESPACE: external-dns-operator
+      OO_PACKAGE: external-dns-operator
+      OO_TARGET_NAMESPACES: '!install'
+    test:
+    - as: test
+      cli: latest
+      commands: make test-e2e
+      from: src
+      resources:
+        requests:
+          cpu: 100m
+    workflow: optional-operators-ci-gcp
+- as: e2e-azure-ovn-operator
+  steps:
+    cluster_profile: azure4
+    dependencies:
+      OO_INDEX: ci-index-external-dns-operator-bundle
+    env:
+      OO_CHANNEL: stable-v1.1
+      OO_INSTALL_NAMESPACE: external-dns-operator
+      OO_PACKAGE: external-dns-operator
+      OO_TARGET_NAMESPACES: '!install'
+    test:
+    - as: test
+      cli: latest
+      commands: make test-e2e
+      from: src
+      resources:
+        requests:
+          cpu: 100m
+    workflow: optional-operators-ci-azure
+- as: e2e-azure-ovn-infoblox-operator
+  steps:
+    cluster_profile: azure4
+    dependencies:
+      OO_INDEX: ci-index-external-dns-operator-bundle
+    env:
+      OO_CHANNEL: stable-v1.1
+      OO_INSTALL_NAMESPACE: external-dns-operator
+      OO_PACKAGE: external-dns-operator
+      OO_TARGET_NAMESPACES: '!install'
+    test:
+    - as: test
+      cli: latest
+      commands: |
+        export DNS_PROVIDER=INFOBLOX
+        export INFOBLOX_CONFIG_DIR=/var/run/infoblox-credentials
+        make test-e2e
+      credentials:
+      - mount_path: /var/run/infoblox-credentials
+        name: infoblox-credentials
+        namespace: test-credentials
+      from: src
+      resources:
+        requests:
+          cpu: 100m
+    workflow: optional-operators-ci-azure
+zz_generated_metadata:
+  branch: release-1.1
+  org: openshift
+  repo: external-dns-operator
--- a/ci-operator/config/openshift/external-dns/openshift-external-dns-release-1.1.yaml
+++ b/ci-operator/config/openshift/external-dns/openshift-external-dns-release-1.1.yaml
@@ -1,0 +1,38 @@
+base_images:
+  base:
+    name: ubi-minimal
+    namespace: ocp
+    tag: "8"
+build_root:
+  from_repository: true
+images:
+- dockerfile_path: Dockerfile.openshift
+  from: base
+  to: external-dns
+promotion:
+  namespace: external-dns-operator
+  tag: "1.1"
+releases:
+  initial:
+    integration:
+      name: "4.12"
+      namespace: ocp
+  latest:
+    integration:
+      include_built_images: true
+      name: "4.12"
+      namespace: ocp
+resources:
+  '*':
+    requests:
+      cpu: 100m
+      memory: 200Mi
+tests:
+- as: unit
+  commands: make test
+  container:
+    from: src
+zz_generated_metadata:
+  branch: release-1.1
+  org: openshift
+  repo: external-dns
--- a/ci-operator/jobs/openshift/external-dns-operator/openshift-external-dns-operator-release-1.1-presubmits.yaml
+++ b/ci-operator/jobs/openshift/external-dns-operator/openshift-external-dns-operator-release-1.1-presubmits.yaml
@@ -1,0 +1,482 @@
+presubmits:
+  openshift/external-dns-operator:
+  - agent: kubernetes
+    always_run: true
+    branches:
+    - ^release-1\.1$
+    - ^release-1\.1-
+    cluster: build01
+    context: ci/prow/ci-index-external-dns-operator-bundle
+    decorate: true
+    labels:
+      ci.openshift.io/generator: prowgen
+      pj-rehearse.openshift.io/can-be-rehearsed: "true"
+    name: pull-ci-openshift-external-dns-operator-release-1.1-ci-index-external-dns-operator-bundle
+    rerun_command: /test ci-index-external-dns-operator-bundle
+    spec:
+      containers:
+      - args:
+        - --gcs-upload-secret=/secrets/gcs/service-account.json
+        - --image-import-pull-secret=/etc/pull-secret/.dockerconfigjson
+        - --report-credentials-file=/etc/report/credentials
+        - --target=ci-index-external-dns-operator-bundle
+        command:
+        - ci-operator
+        image: ci-operator:latest
+        imagePullPolicy: Always
+        name: ""
+        resources:
+          requests:
+            cpu: 10m
+        volumeMounts:
+        - mountPath: /secrets/gcs
+          name: gcs-credentials
+          readOnly: true
+        - mountPath: /etc/pull-secret
+          name: pull-secret
+          readOnly: true
+        - mountPath: /etc/report
+          name: result-aggregator
+          readOnly: true
+      serviceAccountName: ci-operator
+      volumes:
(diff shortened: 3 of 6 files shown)


alebedev87:
Promotion section: `tag_by_commit` is removed as non needed. No impact on the rehearsed images.

candita:
@alebedev87 @miciah since these files contain references to Openshift versions https://github.com/openshift/release/pull/34949/files#diff-c2cb2f4140a4cf2ff73fc4527659ce4d0d45bb9f3887e8041e2a9b06ffc1cf21R10, I wonder if we need to label files with the openshift image number, as we do for most of our CI config/jobs, like https://github.com/openshift/release/tree/master/ci-operator/config/openshift/cluster-ingress-operator, and even for things like https://github.com/openshift/release/tree/master/ci-operator/config/openshift/azure-disk-csi-driver-operator, which may be handled more like external-dns and not tied to specific Openshift versions.  These files are generated automatically for each release so it's not a effort disadvantage.
I did notice that for `maistra` that we label with the maistra version:  https://github.com/openshift/release/tree/master/ci-operator/config/maistra/api, however, they don't contain references to Openshift releases.

alebedev87:
@candita : I believe we should do the same as maistra - keep only the external dns operator versioned configs. We don't use the OCP versioned branches for extrernal dns operator.

candita:
> @candita : I believe we should do the same as maistra - keep only the external dns operator versioned configs. We don't use the OCP versioned branches for extrernal dns operator. Btw ALBO doesn't have OCP branches: https://github.com/openshift/release/tree/master/ci-operator/config/openshift/aws-load-balancer-operator.

I noticed that about ALBO, but you were the author of that one as well.  So, I guess I will be asking some more about why Openshift versions are mentioned in the yaml for external dns if we don't use the versioned branches.

ci-operator/config/openshift/external-dns-operator/openshift-external-dns-operator-release-1.1.yaml
```diff
@@ -0,0 +1,152 @@
+base_images:
+  base:
+    name: "4.12"
```
candita, line 3: ince external-dns isn't tied to an OpenShift release, it is confusing to use or reference OpenShift or OpenShift versions. Can you use a different name, namespace, tag, builder etc?  I expect to not see "4.12", "ocp", or "openshift" anywhere in here.

ci-operator/config/openshift/external-dns-operator/openshift-external-dns-operator-release-1.1.yaml
```diff
@@ -0,0 +1,152 @@
...
+      cli: latest
+      commands: make test-e2e
+      from: src
+      resources:
+        requests:
+          cpu: 100m
+    workflow: optional-operators-ci-azure
+- as: e2e-infoblox-operator
+  steps:
+    cluster_profile: azure4
```
candita, line 125: infoblox uses the same cluster_profile as azure?

ci-operator/config/openshift/external-dns-operator/openshift-external-dns-operator-release-1.1.yaml
```diff
@@ -0,0 +1,152 @@
...
+        name: infoblox-credentials
+        namespace: test-credentials
+      from: src
+      resources:
+        requests:
+          cpu: 100m
+    workflow: optional-operators-ci-azure
+zz_generated_metadata:
+  branch: release-1.1
+  org: openshift
```
candita, line 151: org should not be openshift.

ci-operator/config/openshift/external-dns/openshift-external-dns-release-1.1.yaml
```diff
@@ -0,0 +1,46 @@
+base_images:
+  base:
+    name: "4.12"
+    namespace: ocp
+    tag: base
+  ocp_builder_rhel-8-golang-1.18-openshift-4.12:
```
candita, line 6: golang-1.19 is used for the external-dns-operator, and 1.18 for this?  It would be better to use the same golang version or a comment on the difference.

ci-operator/jobs/openshift/external-dns-operator/openshift-external-dns-operator-release-1.1-presubmits.yaml
```diff
@@ -0,0 +1,482 @@
+presubmits:
+  openshift/external-dns-operator:
+  - agent: kubernetes
+    always_run: true
+    branches:
+    - ^release-1\.1$
```
candita, line 6: Shouldn't the presubmit happen off the master branch?

ci-operator/config/openshift/external-dns-operator/openshift-external-dns-operator-release-1.1.yaml
```diff
@@ -0,0 +1,152 @@
...
+    make verify
+  container:
+    from: src
+- as: unit
+  commands: |
+    export XDG_CACHE_HOME=$(mktemp -d)
+    make test
+  container:
+    from: src
+- as: e2e-aws-operator
```
candita, line 66: Is external-dns supported on ovn?

ci-operator/config/openshift/external-dns/openshift-external-dns-release-1.1.yaml
alebedev87: Yes, it would be better. We have less control over the controller code as it's rebased from the upstream. For this release the upstream wasn't updated to 1.19 yet.

ci-operator/config/openshift/external-dns-operator/openshift-external-dns-operator-release-1.1.yaml
alebedev87: I don't think we have a choice here. Since we use the Openshift CI for external-dns-operator (like a lot of other addon operators do). This is a metadata generated by `make ci-operator-config`, without it the PR wouldn't pass the validation.

ci-operator/config/openshift/external-dns-operator/openshift-external-dns-operator-release-1.1.yaml
alebedev87: You mean the network plugin? OVN is now the default network plugin of OpenShift (couldn't find from which version though). So, I suppose the answer is yes. Also because I cannot remember anything specific to a particular network plugin we used in the operator or controller.

ci-operator/config/openshift/external-dns-operator/openshift-external-dns-operator-release-1.1.yaml
alebedev87: Yes. It doesn't really matter which cloud provider the Infoblox test case will be run at. Since we don't request the credentials from the CloudCredentialsOperator of the cluster like for AWS, GCP and Azure. So, the choice of the cluster profile is pure convenience for Infoblox case - we chose Azure recently due to a lot of issues on GCP and AWS CI accounts.

ci-operator/jobs/openshift/external-dns-operator/openshift-external-dns-operator-release-1.1-presubmits.yaml
alebedev87: Why should it happen off the master branch for `release-1.1` branch?
Btw this file was generated by `make jobs` off the config I added, so I guess it's the normal setup for a new branch.

ci-operator/config/openshift/external-dns-operator/openshift-external-dns-operator-release-1.1.yaml
alebedev87: I don't see a big problem in using the OCP image here. It'll be used as the base (Docker `FROM`) image for the operator.  I think of  it as a sort of minimal Kube version in combination with RHEL which covers the most of the future customers of this operator.

alebedev87:
@candita : thanks for the review!

> I noticed that about ALBO, but you were the author of that one as well :)

As a matter of fact it was Arjun but anyway, I agree with your point about the OCP files in the external-dns configs. I will follow up on this one to exclude the external dns repositories from the OCP job (there seems to be an automation around the branching for OCP dependent repositories).

ci-operator/config/openshift/external-dns-operator/openshift-external-dns-operator-release-1.1.yaml
candita: That sounds reasonable.  Could you please get someone from the release team to validate that you have what is needed here, in order to generate updates for releases after 4.12? If there are OpenShift CI updates after 4.12, this config should be using them, and the things I suggested so far that I know about, you have not agreed would work.  So probably we need an expert on release CI generation files.

ci-operator/config/openshift/external-dns-operator/openshift-external-dns-operator-release-1.1.yaml
candita: You should add the "ovn" to the test name where needed to make sure this runs the OVN tests without flakes. See https://github.com/openshift/release/pull/31430

ci-operator/jobs/openshift/external-dns-operator/openshift-external-dns-operator-release-1.1-presubmits.yaml
candita: I think the release-*.* code is generated using the *-master-presubmits.yaml, or maybe [openshift-external-dns-operator-main.yaml](https://github.com/openshift/release/blob/master/ci-operator/config/openshift/external-dns-operator/openshift-external-dns-operator-main.yaml) in this case.

candita: Something is off - take a look at the structure in https://github.com/openshift/release/tree/master/ci-operator/config/openshift/external-dns-operator

ci-operator/config/openshift/external-dns/openshift-external-dns-release-1.1.yaml
```diff
@@ -0,0 +1,38 @@
...
+    tag: base
+build_root:
+  from_repository: true
+images:
+- dockerfile_path: Dockerfile.openshift
+  from: base
+  to: external-dns
+promotion:
+  namespace: external-dns-operator
+  tag: "1.1"
```
candita, line 14: I think this promotes it into the OpenShift release.  The `promotion` tag is removed in https://github.com/openshift/release/commit/4c06fb054ae2d87564d65c68d0b32aee0a5c16ff

ci-operator/jobs/openshift/external-dns-operator/openshift-external-dns-operator-release-1.1-presubmits.yaml
alebedev87: Sorry, I'm not getting the point. To be the presubmits which target branch `release-1.1` should happen off `release-1.1` branch. The release before the previous release (`1.0` is missing the CI config btw) was dong [the same](https://github.com/openshift/release/blob/615429eab1c4e444f0ac9ac32289deeba136008b/ci-operator/jobs/openshift/external-dns-operator/openshift-external-dns-operator-release-0.1-presubmits.yaml#L6-L7). The cluster-ingress-operator's 4.11 release branch presubmit [uses](https://github.com/openshift/release/blob/615429eab1c4e444f0ac9ac32289deeba136008b/ci-operator/jobs/openshift/cluster-ingress-operator/openshift-cluster-ingress-operator-release-4.11-presubmits.yaml#L6-L7) `release-4.11` branches too.

ci-operator/config/openshift/external-dns-operator/openshift-external-dns-operator-release-1.1.yaml
alebedev87: Found `ubi-minimal` imagestream. Should be enough for external dns repositories.

ci-operator/config/openshift/external-dns-operator/openshift-external-dns-operator-release-1.1.yaml
alebedev87: Done.

ci-operator/jobs/openshift/external-dns-operator/openshift-external-dns-operator-release-1.1-presubmits.yaml
candita: Was this file generated or created manually?

ci-operator/config/openshift/external-dns-operator/openshift-external-dns-operator-release-1.1.yaml
candita: I noticed the failure of `e2e-azure-ovn-infoblox-operator` test.  Just wondering if something needs to change here.

ci-operator/jobs/openshift/external-dns-operator/openshift-external-dns-operator-release-1.1-presubmits.yaml
alebedev87: Generated, by `make jobs` from the config I put in this PR. I believe it's just the naming convention of the config file which contains the branch name.

ci-operator/config/openshift/external-dns-operator/openshift-external-dns-operator-release-1.1.yaml
alebedev87: No, the temporary Infloblox licenses expired, they last 3 months. We still use the temporary licenses for Infoblox product from AWS marketplace as we didn't manage to get any from Infoblox (company). We'll have to re-iterate with DPTP on this.

candita:
@alebedev87 overall, since this product is like maistra, I don't know why you need to specify OCP 4.12 in some of the configuration fields.  I suggest getting a review from someone on the CI team, just to make sure you're not setting a precedent here that will be hard to change later.  So, I mark with lgtm and a hold, which can be removed after the review from the CI team.  If you feel that a review from the CI team isn't necessary, feel free to remove the hold yourself.

/lgtm
/hold

candita:
@alebedev87 Also, I took a look at the current state of these directories, and I think if in this PR you remove the `openshift-external-dns-operator-release-4.*.yaml` and `openshift-external-dns-release-4.*.yaml files, and made sure the generator didn't re-generate these, that would boost my confidence in this.

candita:
Finally, just wanted to know if there was ever a release 1.0?  Or a reason why the version moved from 0.1 to 1.1?

alebedev87:
> Finally, just wanted to know if there was ever a release 1.0? Or a reason why the version moved from 0.1 to 1.1?

Yes, there was. The CI for it was not implemented, I mentioned this in [another thread before](https://github.com/openshift/release/pull/34949#discussion_r1069404071). I'll do a separate PR for the release 1.0.

alebedev87:
> Also, I took a look at the current state of these directories, and I think if in this PR you remove the openshift-external-dns-operator-release-4.*.yaml and `openshift-external-dns-release-4.*.yaml files, and made sure the generator didn't re-generate these, that would boost my confidence in this.

As I [briefly mentioned before](https://github.com/openshift/release/pull/34949#issuecomment-1372077139), I will follow up on this and will push a separate PR to remove all OCP related branch configs.

ci-operator/jobs/openshift/external-dns-operator/openshift-external-dns-operator-release-1.1-presubmits.yaml
LorbusChris: This looks all good, but there should be another config testing the `main` branch (there's no `master` in https://github.com/openshift/external-dns-operator/)

PR #34997, opened 2022-12-21
Albo revert tag by commit

alebedev87:
The tagging by commit to mirror using the wildcard was supposed to replace the custom logic in the mirror job. The custom logic was adding a tag for each image digest published on ALBO's quay.io organization. That, in turn, was a must to preserve the ALBO manifests which used those digests (dangling image digest are cleaned up).
However a tag for a commit is not enough as different branches may have the same Git commit  but different image digests due to different base images, for example.


alebedev87:
/pj-rehearse skip

Mirroring doesn't impact CI config, `tag_by_commit` is optional and was useful only for the mirroring.

hongkailiu (approved):

PR #35006, opened 2022-12-22
NO-ISSUE: Use c3 medium for regular ztp job

adriengentil:
Now that we have stabilized the CI, we can try again to use c3.medium
machine type for ZTP jobs. The goal is to save cost as the c3.medium
costs half of the current m3.large.


osherdp (approved):

adriengentil:
/hold
60% success rate, it would be nice to have the logs of the VMs to understand want is the problem.

adriengentil:
/close
I'll look at it later

PR #35023, opened 2022-12-22
stackrox: disable nightly runs for break

gavin-stackrox:
Per title. The ` 2 ` disables the nightly command until Feb. The `1,2,3,4,5` so that it only runs Mon-Fri (weekend runs are largely ignored).
/cc @vjwilson @BradLugo


BradLugo:
Is there a reason to wait until February as opposed to January?

gavin-stackrox:
> Is there a reason to wait until February as opposed to January?

If I make it January it'll run one on Monday the 2nd which will largely be ignored. What I plan to do is have a follow up PR once this merges to re-enable and `/hold` it until midday Monday the 2nd PDT so that there will be a nightly run ready for Tuesday the 3rd.

BradLugo (approved):
LGTM

PR #35027, opened 2022-12-22
stackrox: enable nightly runs after break

gavin-stackrox:
I'll unhold this Monday 2nd so we have a nightly run waiting for the Tuesday return.
/cc @BradLugo @vjwilson 
/hold


BradLugo (approved):
LGTM

PR #35044, opened 2022-12-23
cloud-provider-openstack: Require tests where CCM is exercised

pierreprinetti:
On 4.12+ (where CCM is the default):
* Remove the experimental/minimal job (is LB comprised in the "minimal" suite?)

On 4.11 and 4.10:
* Stop running e2e-openstack by default, as it is not exercising CCM
* make e2e-openstack-ccm non-optional

/cc mandre


mandre (commented):
/lgtm
/approve

PR #35048, opened 2022-12-23
Remove unit test job for GitOps Service

chetan-rns:
We run unit tests using GH actions in the managed-gitops repository. Since these tests don't require a cluster, we can remove this job from openshift-ci


jgwest (approved):

jgwest:
/lgtm
/approve

Thanks @chetan-rns!

PR #35049, opened 2022-12-23
openstack-conf-clouds: Document appcreds GC

pierreprinetti:
Some application credentials are revoked in the
`openstack-rotate-cloud-credentials` step. In this commit, I document the fact that all remaining expired credentials are deleted by cleanup periodic jobs.


mandre (commented):
/lgtm
/approve

PR #35055, opened 2022-12-25
setting psa labels

mshitrit:
setting psa labels for ocp 4.13 in order to enable e2e tests


slintes:
not needed (and wouldn't work as implemented here)

/close

PR #35059, opened 2022-12-27
fix missing scp error

yunjiang29:
`scp` command is not available in `installer` image, using `rsync` instead.


yunjiang29:
@gpei @jianlinliu PTAL

PR #35063, opened 2022-12-27
OCPQE-13371: remove --continue-on-errror

jianzhangbjz:
Since current index image contains the current version's images only, I remove the `--continue-on-error` flag. It will fail once one image fail to mirror.


jianzhangbjz:
Hi @bandrade , could you help approve it? Thanks!

PR #35090, opened 2023-01-03
DPTP-3267: remove auth log

jianzhangbjz:
As title,


jianzhangbjz:
Since those auth logs are already in the private log, so the people outside redhat cannot see them.
/close

PR #35096, opened 2023-01-03
Modify cron temporarily

LiangquanLi930:
This is a temporary PR to trigger profiles that are not currently running in 4.12.
When our goal is complete, then we will roll back

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__arm64-nightly.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__arm64-nightly.yaml
@@ -254,7 +254,7 @@ tests:
     - chain: openshift-e2e-test-qe
     workflow: cucushift-installer-rehearse-aws-ipi-disconnected-private-sno
 - as: aws-ipi-disconnected-private-sno-p3-f28-destructive
-  cron: 45 6 8 * *
+  cron: 45 6 5 * *
   steps:
     cluster_profile: aws-qe
     dependencies:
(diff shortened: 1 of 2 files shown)


ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__arm64-nightly.yaml
```diff
@@ -864,7 +864,7 @@ tests:
     - chain: openshift-e2e-test-qe
     workflow: cucushift-installer-rehearse-azure-ipi-baselinecaps-v412
 - as: azure-ipi-baselinecaps-v412-p3-f28-destructive
-  cron: 16 18 2 * *
+  cron: 16 18 4 * *
```
liangxia, line 867: Isn't it already executed ?

LiangquanLi930: But executed failed, so re-trigger again

liangxia: We should use rebuild in that case.

LiangquanLi930: okay, got it, thanks, updated

liangxia:
Unfortunately, conflicts at the last minute

LiangquanLi930:
> Unfortunately, conflicts at the last minute

yes, rebase completed

PR #35102, opened 2023-01-03
Remove stackrox/collector PR and master checks

Molter73:
We are moving most of stackrox/collector checks to GHA, since we are nearing completion of the migration, this PR will be responsible for removing existing configuration.


Stringy:
Does this currently contain just deletions of the features that we have so far ported to GHA? I was expecting a full deletion of the configuration

Stringy:
Hang on, I'm an idiot and somehow missed the PR title, ignore me.

Stringy (approved):

Molter73:
The push-drivers and support-packages job seems to be failing due to the CPaaS variant not being ready for running on PRs, I think we can ignore it and I'll take a closer look if it keeps failing after merge.

PR #35107, opened 2023-01-03
Fix driver build failures not sending slack notifications

Molter73:
This small PR makes it so we will have slack notifications when a driver fails to build.


Molter73:
/unhold

This was not the PR I meant to hold.

PR #35121, opened 2023-01-04
Sync serverless CI

pierDipi:
Includes https://github.com/openshift-knative/hack/pull/17


aliok:
Can you write the reason for the change here?

core-services/image-mirroring/knative/mapping_knative_knative-v1.7_eventing-kafka-broker_quay
```diff
@@ -0,0 +1,22 @@
+registry.ci.openshift.org/openshift/knative-eventing-kafka-broker-dispatcher quay.io/openshift-knative/knative-eventing-kafka-broker-dispatcher
```
matzew, line 1: I can't pull these

pierDipi: This works for me: docker pull registry.ci.openshift.org/openshift/knative-eventing-kafka-broker-dispatcher:knative-v1.7

pierDipi: this syntax is omitting the tag to mirror every tag

matzew: OK, so the mirror job knows this, and does than apply the version?

matzew: I have some love for explicit versions, since a string like `registry.ci.openshift.org/openshift/knative-eventing-kafka-broker-dispatcher:knative-v1.7` is a bit easier for debugging, since I can just use it, instead of copy the string, and manually add something to it for testing.

But if the mirror job does this, than I am OK to do all implicit versions, and we fiddle the rest manually, if we need to debug something. Not a fan, but OK

pierDipi: This is mirroring commit tags, we cannot reference every commit here

matzew: Ah, check - ok. sorry, now I get it

matzew (approved):
/lgtm
/approve

PR #35122, opened 2023-01-04
[ibmcloud] add timestamp in the resource group name

MayXuQQ:
in order to easily find the the resource group of ibmcloud prow job , add timestamp in the resource group name.


MayXuQQ:
discard change the resource  group name

PR #35123, opened 2023-01-04
Increate timeout of installer CI jobs

r4f4:
Many jobs are being cancelled during e2e or deprovision steps because 4h is too short a timeout. Let's increase it until we can work on reducing the installer build time.

### DIFF
--- a/ci-operator/config/openshift/installer/openshift-installer-master.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-master.yaml
@@ -228,23 +228,27 @@ tests:
   steps:
     cluster_profile: aws
     workflow: openshift-e2e-aws
+  timeout: 6h0m0s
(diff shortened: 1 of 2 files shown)


r4f4:
/pj-rehearse pull-ci-openshift-installer-master-e2e-gcp-ovn pull-ci-openshift-installer-master-e2e-azure-ovn pull-ci-openshift-installer-master-e2e-aws-ovn

patrickdillon:
Let’s do this so we stop wasting resources, but this should be a temporary fix and we need to fix the underlying build problems

sdodson:
/pj-rehearse ack
None of the jobs have behaved abnormally, worst case this change is ineffective and things still time out at 4hrs and we follow up with another change. Just wanting to go ahead land this change so that we stop spawning jobs which are doomed to fail at 4hrs due to known issue

PR #35132, opened 2023-01-04
Abhinav has left Red Hat :(

wking:
Generated [with][1]:

```console
$ hack/owners-prune.py -r ~/src/openshift/release -m 'Abhinav has left Red Hat :(' abhinavdahiya
$ make registry-metadata
$ git diff | grep -A2 'diff\|"approvers": ' | grep -B10 '\]' | sed -n 's|--- a/\(.*/\)[^/]*[.]json|\1|p' | while read DIR; do (cd "${DIR}"; rm -f OWNERS; ln -s ../OWNERS); done
$ make registry-metadata
```

The symlink setup removes `OWNERS` files where Abhinav was the only approver, replacing them with a symlink to the parent `OWNERS` file, because `ci-operator` gets grumpy if each `step-registry` directory doesn't have its own explicit `OWNERS` entry.

[1]: https://github.com/openshift/release/pull/21876

### DIFF
--- a/ci-operator/step-registry/ipi/azure/pre/stableinitial/ipi-azure-pre-stableinitial-chain.metadata.json
+++ b/ci-operator/step-registry/ipi/azure/pre/stableinitial/ipi-azure-pre-stableinitial-chain.metadata.json
@@ -2,7 +2,11 @@
 	"path": "ipi/azure/pre/stableinitial/ipi-azure-pre-stableinitial-chain.yaml",
 	"owners": {
 		"approvers": [
-			"abhinavdahiya"
+			"jianlinliu",
+			"patrickdillon",
+			"technical-release-team-approvers",
+			"vrutkovs",
+			"yunjiang29"
 		]
 	}
 }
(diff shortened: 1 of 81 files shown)


ci-operator/step-registry/ipi/azure/pre/stableinitial/ipi-azure-pre-stableinitial-chain.metadata.json
```diff
@@ -2,7 +2,12 @@
 	"path": "ipi/azure/pre/stableinitial/ipi-azure-pre-stableinitial-chain.yaml",
 	"owners": {
 		"approvers": [
-			"abhinavdahiya"
+			"vrutkovs",
+			"deads2k",
+			"technical-release-team-approvers",
```
deads2k, line 7: are they aware?

wking: After some back-channel discussion, all of the other parent-directory approvers are ok with being direct approvers in the three previously-approver-less directories, except for @deads2k who who did not want to become a direct approver in any of the three previously-approver-less directories.  I've rerolled to reflect that with 95f97b7e534 -> cb6698cf7dc.

danilo-gemoli:
Who is Abhinav?

PR #35133, opened 2023-01-04
configure artifacts for preflight e2e testing

komish:
Instead of creating the necessary directory in the source image with the right permissions, we use a tempdir.

Signed-off-by: Jose R. Gonzalez <jose@flutes.dev>


acornett21 (approved):

komish:
I readded the `mkdir` command that adds the artifacts directory into the original image to avoid a chicken/egg problem. Once Preflight's PR merges, we should be able to remove that line without issue.

komish:
I'm okay with _this_ failure. This is still a chicken/egg problem, but I'm not sure I can do anything about it until the Preflight PR merges.

komish:
/pj-rehearse skip pull-ci-redhat-openshift-ecosystem-openshift-preflight-main-4.8-e2e

PR #35149, opened 2023-01-05
Refine test frequency based on payloads

liangxia:
We are having 4.8/4.9 stable payloads every two weeks+, so refine the test frequency to match that.

/cc @jianlinliu @jhou1 @JianLi-RH @dis016 @pruan-rht


jianlinliu:
Nice!

/lgtm

PR #35157, opened 2023-01-05
Mark all e2e jobs as required in SO

ReToCode:
- Mark all e2e jobs as required in SO

### DIFF
--- a/ci-operator/jobs/openshift-knative/serverless-operator/openshift-knative-serverless-operator-main-presubmits.yaml
+++ b/ci-operator/jobs/openshift-knative/serverless-operator/openshift-knative-serverless-operator-main-presubmits.yaml
@@ -406,7 +406,6 @@ presubmits:
       job-release: "4.11"
       pj-rehearse.openshift.io/can-be-rehearsed: "true"
     name: pull-ci-openshift-knative-serverless-operator-main-4.11-upgrade-tests-aws-ocp-411
-    optional: true
     rerun_command: /test 4.11-upgrade-tests-aws-ocp-411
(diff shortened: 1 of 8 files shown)


ci-operator/jobs/openshift-knative/serverless-operator/openshift-knative-serverless-operator-main-presubmits.yaml
```diff
@@ -542,7 +540,6 @@ presubmits:
       job-release: "4.11"
       pj-rehearse.openshift.io/can-be-rehearsed: "true"
     name: pull-ci-openshift-knative-serverless-operator-main-4.11-upstream-e2e-kafka-aws-ocp-411
-    optional: true
```
nak3, line 545: We will make `optional: true` for the job with `always_run: false`?

nak3: Never mind... `4.11-upstream-e2e-kafka-aws-ocp-411` is always running. Probably I misunderstood something about the `always_run: false`.

ReToCode: I actually had to look that up yesterday, because I also did not see the relation. What I think I got is: `optional` is not related to `always_run`. `always_run` defaults to  `true`, but one can edit the generated jobs and set it to `false` and `make jobs` won't change it back. There is no setting for `always_run` in the config file.

PR #35178, opened 2023-01-05
Make Loki DaemonSet linux only

saifshaikh48:
This PR limits Loki pods to run on linux nodes only.
Loki need not run on Windows nodes, especially because 
it does not have a Windows compatible container image.


aravindhp:
/cc @ingvagabund @periklis @blockloop

Please review as this is blocking [WMCO](https://github.com/openshift/windows-machine-config-operator) PRs from merging.

PR #35252, opened 2023-01-10
refine sdn migration ovn jobs

zhaozhanqi:
@jianlinliu As we talked in meeting, I removed the following part. 

Remove sdn2ovn job for all version including e2e amd64/arm and upgrade 

cc @YuLi517 @LiangquanLi930


aleskandro:
/hold

In the previous meeting, we decided to have the same tests with the same frequency for both architectures. This PR seems to revert this decision.

Why not just reduce the frequency for both the amd64 and arm64 tests? Multi-arch QE wants the same tests to happen for both architectures unless a specific reason is agreed upon not testing on one of the architectures: is there any? 

By reducing the frequency of tests against amd64 and increasing arm64, so that the sum is still the same as before, we'd even reduce infra costs.

andymcc:
If testing on one arch is enough to cover for other archs it would be more sensible to test against arm64 since the CI costs would be cheaper. If that doesn't seem acceptable we should analyse and discuss why it would then make sense to only test on amd64.

zhaozhanqi:
>In the previous meeting, we decided to have the same tests with the same frequency for both architectures. This PR seems to revert this decision.
> Why not just reduce the frequency for both the amd64 and arm64 tests? Multi-arch QE wants the same tests to happen for both architectures unless a specific reason is agreed upon not testing on one of the architectures: is there any?

Hmm, I'm not aware the decision we need same tests for both architectures. Those will for all testing even though for specified features like ovn related?  IMO, it's not needed,  As the reason now we need to reduce the jobs of duplicated. and since some jobs already added in amd64 but not added arm64 yet. So here we want to use amd64 as the platform so that it's convenient for management and monitor the jobs in the future.

jianlinliu:
> Why not just reduce the frequency for both the amd64 and arm64 tests? Multi-arch QE wants the same tests to happen for both architectures unless a specific reason is agreed upon not testing on one of the architectures: is there any?

Take 4.12 as example, presently we are covering all kinds of networking features in ci, including snd, ovn, ovn-ipsec, snd2ovn, ovn-local-to-shared-gateway-mode-migration on all platforms, run `egrep -r "as: .*sdn|as: .*ovn" *4.12* |grep -v etcd-encryption|grep -v hypershift|grep -v fips-|grep -v baremetalds|grep -v private-xpn |grep -v upgrade|grep -v multi-nightly |grep -v stable.yaml`, for now, there are 45 ci jobs in total, 16 jobs are on f14, 12 jobs are on f7, 17 jobs are on f28, that means the ci triggered 16*2+12*4+17=97 jobs in one month, and it only counts networking related features ci jobs on amd+arm nightly jobs, even did not count stable and upgrade jobs. I do not think our ci quota can bear so many ci jobs, and the whole QE team can complete so many ci test results analysis in a reasonable workflow. We need to  go through the test configurations one by one to identify if the test configuration is really deserved on both arches. If we blindly double all the configuration combination on both arches, our ci will be out of control, ci workload always is hight, and costing money every day, but QE can not find valuable bugs to improve product quality.

aleskandro:
> > Why not just reduce the frequency for both the amd64 and arm64 tests? Multi-arch QE wants the same tests to happen for both architectures unless a specific reason is agreed upon not testing on one of the architectures: is there any?
> 
> We are following our previous decision to set frequency. It is becoming harder to reduce the frequency when the total jobs number is getting larger, especially different requirements from different subteams, e.g: [#34106 (comment)](https://github.com/openshift/release/pull/34106#issuecomment-1319606828).
> 
> Take 4.12 as example, presently we are covering all kinds of networking features in ci, including snd, ovn, ovn-ipsec, snd2ovn, ovn-local-to-shared-gateway-mode-migration on all platforms.
> 
> Run `egrep -r "as: .*sdn|as: .*ovn" *4.12* |grep -v etcd-encryption|grep -v hypershift|grep -v fips-|grep -v baremetalds|grep -v private-xpn |grep -v upgrade|grep -v multi-nightly |grep -v stable.yaml`, for now, there are 45 ci jobs in total, 16 jobs are on f14, 12 jobs are on f7, 17 jobs are on f28, that means the ci triggered 16 * 2 + 12 * 4 + 17=97 jobs in one month, and it even only counts networking related features ci jobs on amd+arm nightly jobs, even did not count stable and upgrade jobs yet.
> 
> I do not think our ci quota can bear so many ci jobs, and the whole QE team can complete so many ci test results analysis in a reasonable workload.
> 

Halving the frequency is still a solution to what you say. If we have a test running twice a week for both amd64 and arm64, we can just set the test to run once a week for arm64 and once a week for amd64. 

For example,
> e.g: https://github.com/openshift/release/pull/34106#issuecomment-1319606828.
> For sdn ovn, ipsec etc. weekly is ok but better cover all platforms
> For sdn-migration-ovn, Two weeks are ok (one week for IPI, next week for UPI)

Let's rotate for the arches as well, with that frequency. Taking the first (sdn ovn, ipsec) one week is for arm64, the other for amd64. 

> We need to go through the test configurations one by one to identify if the test configuration is really deserved on both arches, e.g: does sdn need to be covered on arm (If I remember right, OVN is preferred for arm cluster)? does snd2ovn need to covered on arm? If we blindly double all the configuration combination on both arches, our ci is going to be out of control, ci workload always is high, and costing money every day, but QE can not find valuable bugs to improve product quality. We need refine the ci pool to balance ci jobs and QE workload on a reasonable level.

ARM is now GAed and the preference for OVN is not there anymore. Any test profile covered on amd64 has to be overend on arm64, unless there are unsupported features.

jianlinliu:
> Halving the frequency is still a solution to what you say.

For now, maybe it still works, but if in the feature, the total job number come to 1000+, our minimum test frequency is only once per month. That means all jobs will be triggered in one month. The test frequency can not help us any more. And for some important configuration matrix, we are expecting to see some test results weekly or bi-weekly, but not monthly.

jianlinliu:
@zhaozhanqi I guess now we can  do the following, if you agreed per @aleskandro 's statement:
1. remove some duplicated test coverage for `sdn-migration-ovn` and `sdn2ovn` on both arches
2. adjust test frequency. eg: f14 -> f28, if the testing is being covered on both aches?

aleskandro:
> > In the previous meeting, we decided to have the same tests with the same frequency for both architectures. This PR seems to revert this decision.
> > Why not just reduce the frequency for both the amd64 and arm64 tests? Multi-arch QE wants the same tests to happen for both architectures unless a specific reason is agreed upon not testing on one of the architectures: is there any?
> 
> Hmm, I'm not aware the decision we need same tests for both architectures. Those will for all testing even though for specified features like (sdn migration ovn) features related? IMO, it's not needed, As the reason now we need to reduce the jobs of duplicated. and since some jobs already added in amd64 but not added arm64 yet. So here we want to use amd64 as the platform so that it's convenient for management and monitor the jobs in the future.

> and since some jobs already added in amd64 but not added arm64 yet.

If there are again other jobs being added only for amd64, without involvement of multi-arch qe, or a story on the ARMOCP project to track the enablement issue of the arm64 counterpart, the approvers/reviewers of the openshift-tests-private organization are not doing what we discussed and this becomes a problem for us (multi-arch qe).

We cannot "implement a reconciliation loop" to copy and paste all the test configs from the amd64 files to the arm64 ones. We did [these PRs recently](https://github.com/openshift/release/pulls?q=is%3Apr+is%3Aclosed+author%3Aaleskandro+Sync+arm64+) to re-sync again from amd64 to arm64 and we asked reviewers/approvers to check that any new PR includes, for the platforms supporting arm64 (baremetal, aws, azure IPI as of now), test configs for both of them. 

Given that, for the platforms GAed on versions <= 4.12, a PR has to take into account both the architectures by using the "templates" already available in the relevant confs.

If a PR doesn't, the multi-arch qe team should be aware of this. We can support it or, if adding the config needs more time for lack of support in some steps, we can open a story to support the needed fix and enable the test for arm64 as well. 

If each team helps starts supporting this and shift to this multi-arch behavior by default, eventually it will become as normal as deciding which Cloud Platform we are testing on, and hopefully, "ARM QE" will not exists.

If we don't do this, multi-arch QE doesn't have resources to continue with the enablement of tests for the new platforms to be GAed on 4.13 and need  to stay executing a loop of copy and paste for all the GAed platforms.

Finally, the decision to test on an architecture or another should be discussed with multi-arch qe, and when this is going to be decided, I suggest running the tests on arm64 to reduce the costs.

aleskandro:
> > Halving the frequency is still a solution to what you say.
> 
> For now, maybe it still works, but if in the feature, the total job number come to 1000+, our minimum test frequency is only once per month. That means all jobs will be triggered in one month. The test frequency can not help us any more. And for some important configuration matrix, we are expecting to see some test results weekly or bi-weekly, but not monthly.

Yes, we need to take into account this problem also for the next coming GAing of multi-arch compute nodes clusters (ex-heterogeneous). But the first step is still that the teams adds confs for both.

zhaozhanqi:
IMO, if a feature is not platform related or architecture related.  and we can make sure it.  We would like to use less jobs to cover it.  thus
1.  it can also reduce the resource issue. 
2.  it can also reduce the maintenance so many jobs. 

eg. ` vsphere-ipi-ovn-local-to-shared-gateway-mode-migration-p2-f7`,  ovn-local-to-shared-gateway-mode-migration should not be platform related.  So here we only covered `ovn-local-to-shared-gateway-mode-migration` this feature on vsphere platform, but not other aws/azure. etc. 

So for same reason. I don't think we need to copy one for both amd64 and arm64.  So since some job already added in amd64 platform,  if arm64 is more cheaper than amd64.  We can consider only add job on arm64 in the future if feature is not related architecture. That's my thought. 

Also add @anuragthehatter here. any thought?

aleskandro:
> IMO, if a feature is not platform related or architecture related. and we can make sure it. We would like to use less jobs to cover it. thus
> 
>     1. it can also reduce the resource issue.
> 
>     2. it can also reduce the maintenance so many jobs.
> 
> 
> eg. ` vsphere-ipi-ovn-local-to-shared-gateway-mode-migration-p2-f7`, ovn-local-to-shared-gateway-mode-migration should not be platform related. So here we only covered `ovn-local-to-shared-gateway-mode-migration` this feature on vsphere platform, but not other aws/azure. etc.
> 
> So for same reason. I don't think we need to copy one for both amd64 and arm64. Since some job already added in amd64 platform before, if arm64 is more cheaper than amd64. We can consider only add job on arm64 in the future if feature is not related architecture. That's my thought.
> 
> Also add @anuragthehatter here. any thought?

Indeed, what you say, is the direction that multi-arch qe wants to take. But, till now, this meant, for most of us, that <<I test on amd64 [as it seems easier than arm64], then in the future, [maybe, and someone else] I'll take care of arm64>>.

Therefore, multi-arch qe decided to ask for this enforcement. We believe it allows (1) a fair rotation when testing against the different architectures for any generic feature, (2) the whole team to start embracing that we now support multiple architectures. 

I hope to see what you say in the future, but that future, for multi-arch qe, is 1y ago.

zhaozhanqi:
@jianlinliu @aleskandro 

OK,  we can discuss more later if need tested both arch platform for this kind of specified features, not only this migration one. 
Now I already rollback the code.  and only remove sdn2ovn this dup for amd and arm.  PTAL,Thanks.

aleskandro:
Hi @zhaozhanqi thanks, can you update the description of this PR as well?

zhaozhanqi:
> Hi @zhaozhanqi thanks, can you update the description of this PR as well?

ah, yes, updated, Thanks for reminding this

jianlinliu:
cc @liangxia to review and approve.

PR #35257, opened 2023-01-10
fix lvm job SERVICE_BASE_REF

osherdp:
Seems like the job missed using the right assisted-service branch for its deployment scripts.
It is inconsistent with the behavior of the other jobs, and might cause a situation that assisted-service image is not in sync with its deployment method.


osherdp:
/unhold
temporarily it fixes the problem, and it's also the right thing to do nevertheless

gamli75 (approved):

PR #35278, opened 2023-01-10
stackrox: remove imaging

gavin-stackrox:
Removes imaging from OSCI now that it is performed elsewhere.
/cc @porridge @tommartensen @davdhacs


porridge (approved):
/lgtm

PR #35279, opened 2023-01-10
openshift-ci ClusterPools: Move some capacity to 4.11

2uasimojo:
We started getting claim timeouts on the
ci-ocp-4-11-0-amd64-aws-us-east-1 ClusterPool because we were hitting its `maxSize`. Meanwhile, 4.8 and 4.9 clusterpools had `maxSize: 20` and `size: 10` but are seeing a lot less use. This commit:
- Reduces the 4.8 and 4.9 pools to `maxSize: 10` and `size: 3`
- Creates a new 4.11 ClusterPool in us-west-1. It is identical in every other way to the other 4.11 ClusterPool: `maxSize: 20`, `size: 10`, `runningCount: 4`. (This is now similar to the two-pool setup we have for 4.10.)
- Changes hive master's `e2e` test to use the new pool. Now (for most PRs) we'll have one job using each pool, thus spreading out the load.

### DIFF
--- a/ci-operator/config/openshift/hive/openshift-hive-master.yaml
+++ b/ci-operator/config/openshift/hive/openshift-hive-master.yaml
@@ -77,7 +77,7 @@ tests:
     architecture: amd64
     cloud: aws
     labels:
-      region: us-east-1
+      region: us-west-1
     owner: openshift-ci
     product: ocp
     timeout: 1h0m0s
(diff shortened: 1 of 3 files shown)


2uasimojo:
/assign @bear-redhat 

cf. [thread](https://redhat-internal.slack.com/archives/CBN38N3MW/p1673363340279289)

2uasimojo:
We're going to try to do this more intelligently by looking at metrics.

PR #35297, opened 2023-01-11
cincinnati-operator: use go 1.18 as a build root and test on OCP 4.11

petr-muller:
Companion to https://github.com/openshift/cincinnati-operator/pull/155


wking (approved):
/lgtm

PR #35303, opened 2023-01-11
HCO: move one test to the CNV account

nunnatsa:


orenc1:
looks like it is running on our account:
```
INFO[2023-01-11T20:06:42Z] Acquiring leases for test hco-e2e-image-index-aws: [aws-virtualization-quota-slice] 
INFO[2023-01-11T20:06:43Z] Acquired 1 lease(s) for aws-virtualization-quota-slice: [us-east-2--aws-virtualization-quota-slice-3] 
```
🎉

nunnatsa:
Right. But want to check again after merging #35308

nunnatsa:
Actually @orenc1, I think it's safe to merge it. The fix is about the bootstrap script that failed. If it failed, then we're for sure using the right secret.

orenc1:
> Actually @orenc1, I think it's safe to merge it. The fix is about the bootstrap script that failed. If it failed, then we're for sure using the right secret.

yes, but in #35308 you're changing the secret configuration of `aws-virtualization` profile (removing the `.awscred` key). i think it will be safer to double check if it's still working after the change.

nunnatsa:
rebased to get #35308 changes (not really needed. Just for case)

/pj-rehearse

orenc1:
```
INFO[2023-01-12T12:18:02Z] Step hco-e2e-image-index-aws-ipi-install-install succeeded after 28m0s. 
```
/lgtm
/approve

PR #35305, opened 2023-01-11
Disable infra image mirroring jobs that fail frequently

jmguzik:
Disable == set cron to yearly


jupierce:
If "fail frequently" == 100%, this seems completely reasonable. 
/hold
/lgtm

PR #35311, opened 2023-01-11
Re-enable OSDe2e hypershift pipeline

MrSantamaria:


ci-operator/jobs/openshift/osde2e/openshift-osde2e-main-periodics.yaml
```diff
@@ -32,6 +32,8 @@ periodics:
       command:
       - /osde2e
       env:
+      - name: CHANNEL
+        value: candidate
       - name: CLUSTER_VERSION
         value: 4.12.0-rc.1
```
ryankwilliams, line 38: @MrSantamaria Is it possible to have the job automatically select the latest available version to test? If not, I would say we update this from `4.12.0-rc.1` to `4.12.0-rc.8` as that is the latest version.

MrSantamaria: Updated to 4.12.0-rc.8
OSDe2e can pick the latest version by itself. 
We are gated by ROSA though, I'm not sure how they are updating their approved list for hcp clusters approved versions:
```
ERR: Expected a valid OpenShift version: A valid version number must be specified
Valid versions: 4.12.0-rc.8 4.12.0-rc.7 4.12.0-rc.6 4.12.0-rc.5 4.12.0-rc.4 4.12.0-rc.3 4.12.0-rc.2 4.12.0-rc.1 4.12.0-rc.0
```

ryankwilliams: Thanks for the info Diego. Sounds good to me. Thank you for making the change.

PR #35317, opened 2023-01-12
rosa-hypershift: add CHANNEL_GROUP: nightly

LiangquanLi930:


LiangquanLi930:
@liangxia please talk a look, thanks

jianlinliu:
any backgroup to increase f7 to f2?

LiangquanLi930:
> any backgroup to increase f7 to f2?

@yasun1 @yufchang could you help to answer it? thanks

LiangquanLi930:
> any backgroup to increase f7 to f2?

@yasun1 @yufchang could you help to answer it? thanks

LiangquanLi930:
@jianlinliu We need to run it every two days, so that we can find bugs as early as possible

yasun1:
@jianlinliu @LiangquanLi930 Although provisioning the hypershift rosa cluster with the nightly version is failed, as confirmed with @yufchang, we still want to add a nightly job, and keep the current candidate job. In order to remove the candidate job, we'd like to increase the priority of the nightly job to find when the provisioning is successful ASAP automatically.

jianlinliu:
So far, "f2" is the highest frequency in the whole prow ci, if the test results is really needed and analyzed, I can not object it. But you know, our infrastructure quota is tight, and if in the features (e.g: 4.13, 4.14...), the frequency of "f2" would put more workload on  our infrastructure quota.

LiangquanLi930:
@jianlinliu i think we don't need to worry about this because we will only have one version 4.12 for rosa-hypershift and will change later for 4.13

jianlinliu:
> @jianlinliu i think we don't need to worry about this because we will only have one version 4.12 for rosa-hypershift and will change later for 4.13

Now when new versions come in, generally productivity team will copy the old one to new version, once we forget to update the frequency, issues come out. Pls keep in mind to update the frequency when new versions come in.

/lgtm

LiangquanLi930:
@liangxia could you help to add `approved` ? thanks

heliubj18:
@yasun1 Does this job use the current QE quota in ROSA ?  I remembered you mentioned it need the requirement for ROSA quota additionally instead of using current QE account. cc @jianlinliu

yasun1:
> @yasun1 Does this job use the current QE quota in ROSA ? I remembered you mentioned it need the requirement for ROSA quota additionally instead of using current QE account. cc @jianlinliu

The managed clusters provided by the Prow CI are using QE's personal cloud resource.

PR #35319, opened 2023-01-12
ODS: add ods-notebooks-long test with 4h delay

kpouget:
This test will be used for running longer RHODS notebook tests like the gating test

cc @fcami


fcami (approved):
/lgtm

PR #35326, opened 2023-01-12
STOR 1129: Add vsphere upgrade job with some Pods that use PVs

jsafrane:
OCP 4.13 enables CSI migration and we want to make sure that upgrade from 4.12 does not break any workloads that use storage.

The job creates 50 1-replica StatefulSets and expects they survive upgrade to 4.13.

CSI migration on vSphere is a complex process, therefore I'd like to run this job relatively often during 4.13 development. It can drop to once per week after 4.13 GA. 

cc @openshift/storage


jsafrane:
Added more debug dumps in each step to diagnose potential errors

jsafrane:
Now it's perfect & debuggable.

bertinatto (commented):
/lgtm

PR #35333, opened 2023-01-12
Make hypershift e2e required in autoscaler

enxebre:
/hold

### DIFF
--- a/ci-operator/config/openshift/kubernetes-autoscaler/openshift-kubernetes-autoscaler-master.yaml
+++ b/ci-operator/config/openshift/kubernetes-autoscaler/openshift-kubernetes-autoscaler-master.yaml
@@ -138,7 +138,6 @@ tests:
           cpu: 100m
     workflow: ipi-gcp
 - as: e2e-hypershift
-  optional: true
   skip_if_only_changed: ^docs/|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|PROJECT|LICENSE)$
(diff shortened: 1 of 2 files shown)


csrwng:
/pj-rehearse pull-ci-openshift-kubernetes-autoscaler-master-e2e-hypershift

elmiko:
given the low frequency of changes to the repo i think this is a good pre-submit test to run. i think we should consider adding a `skip_if_only_changed` clause to ensure that this test runs only when we change the cluster-autoscaler parts of the repository as i'm not sure we need to test if only a vertial-pod-autoscaler or docs change is added.

enxebre:
rebased!
>given the low frequency of changes to the repo i think this is a good pre-submit test to run. i think we should consider adding a skip_if_only_changed clause to ensure that this test runs only when we change the cluster-autoscaler parts of the repository as i'm not sure we need to test if only a vertial-pod-autoscaler or docs change is added.

@elmiko is there a precedent I can use as example in this repo? otherwise I'll follow up in a different PR to do it all together for all mapi/capi jobs.

elmiko:
> @elmiko is there a precedent I can use as example in this repo? otherwise I'll follow up in a different PR to do it all together for all mapi/capi jobs.

sadly, i think not, maybe it was wishful thinking on my part. happy to have it in a followup, but i think the current restriction on skipping if there is a docs change is good enough.

/lgtm

PR #35340, opened 2023-01-12
OSDe2e - Hypershift PR Check and Prowgen Periodic

MrSantamaria:
https://issues.redhat.com/browse/SDCICD-896

### DIFF
--- a/ci-operator/jobs/openshift/osde2e/openshift-osde2e-main-presubmits.yaml
+++ b/ci-operator/jobs/openshift/osde2e/openshift-osde2e-main-presubmits.yaml
@@ -54,7 +54,7 @@ presubmits:
     branches:
     - ^main$
     - ^main-
-    cluster: build02
+    cluster: build03
     context: ci/prow/hypershift-pr-check
     decorate: true
     decoration_config:


ci-operator/config/openshift/osde2e/openshift-osde2e-main.yaml
```diff
@@ -290,6 +308,26 @@ tests:
     name: osde2e-credentials
   - mount_path: /usr/local/osde2e-rosa-stage
     name: osde2e-rosa-stage
+- as: hypershift-stage-e2e-default
+  commands: |
+    export REPORT_DIR="$ARTIFACT_DIR"
+    export CONFIGS="hypershift"
+    export CHANNEL="candidate"
+    export CLUSTER_VERSION="4.12.0-rc.8"
```
jbpratt, line 316: Can we set this to latest 4.12? This will fall out of date quickly and I don't see much value in continuously testing the same build for weeks at a time

MrSantamaria: Hadn't seen this. 
Updated to 4.12.0, I would leave this to be updated to the latest but it is gated by Rosa: 
ERR: Expected a valid OpenShift version: A valid version number must be specified
Valid versions: 4.12.0 4.12.0-rc.8 4.12.0-rc.7 4.12.0-rc.6 4.12.0-rc.5 4.12.0-rc.4 4.12.0-rc.3 4.12.0-rc.2 4.12.0-rc.1 4.12.0-rc.0

I'm not sure how they approve hypershift versions so we can align with them but the static version is problematic as versions move up.

jbpratt: Hmm, I think `4.12.0` is possibly older than `4.12.0-rc.8`. Is it possible that we set this to install latest 4.12 at any given time?

MrSantamaria: Tested on my end the stable channel is able to pick up the latest 4.12.0 and it is golden. 
When picking versions for the candidate channel though it ask for specific rc-x.

ci-operator/jobs/openshift/osde2e/openshift-osde2e-main-periodics.yaml
```diff
@@ -1455,6 +1455,84 @@ periodics:
...
+  decorate: true
+  decoration_config:
+    skip_cloning: true
+  extra_refs:
+  - base_ref: main
+    org: openshift
+    repo: osde2e
+  labels:
+    ci.openshift.io/generator: prowgen
+  name: periodic-ci-openshift-osde2e-main-hypershift-stage-e2e-default
```
ryankwilliams, line 1470: @MrSantamaria Should we remove the existing periodic job [osde2e-hypershift-stage](https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift/osde2e/openshift-osde2e-main-periodics.yaml#L12)? This way we do not have the same job running x2,

MrSantamaria: There's two different purposes for these pipelines. 
The periodic job is being used to inform the total health of Hypershift. We have seen that job catch the breaking changes in the environment. 

The PR Check is mainly for QE to have something to test against as changes are made to the test while they live in the OSDe2e repo, so to not break the informing pipeline.

MrSantamaria: Ah I misunderstood your comment. 
Good catch, I should remove the handcrafted job since it is being switched to prowgen.

PR #35370, opened 2023-01-13
Remove vsphere tech-preview jobs for CCM related repos

lobziik:
vSphere cloud-controller-manager gets to GA technically and installs by default now.

xref:
- https://github.com/openshift/library-go/pull/1451


lobziik:
Fro prevent accidental merge
/hold

cc @JoelSpeed

PR #35384, opened 2023-01-13
add tcpdump on aws-sdn-upgrade and w/out -i option

jluhrsen:
need to compare some packet captures between azure and another platform (aws). with `-i any` the captures are pulling in duplicate copies of the same packets which makes analyzing them harder. trying without that flag at all hoping just grabbing traffic on the default interface will be enough.

JIRA: https://issues.redhat.com/browse/OCPBUGSM-45030


jluhrsen:
/hold
this has jobs renamed so that I can check things are ok with rehearsals
/pj-rehearse

PR #35385, opened 2023-01-13
enable rhel-coreos-9 overrides a different way

cheesesashimi:
- overrides rhel-coreos-8 with rhel-coreos-9 for testing purposes
- use the correct rhel-coreos-9-extensions image
- use the rhcos-devel prebuilt release to base the MCO changes on top of
- try using rhcos-devel overrides mixed with a build so we can inject them
- try injecting the images using no-op image builds
- try disabling this maybe?
- try this again...


cheesesashimi:
This didn't work. Closing PR.

PR #35388, opened 2023-01-14
updated to skip doc test

amit-62:
This PR is against issue [#6500](https://github.com/cri-o/cri-o/issues/6500)

### DIFF
--- a/ci-operator/jobs/cri-o/cri-o/cri-o-cri-o-main-presubmits.yaml
+++ b/ci-operator/jobs/cri-o/cri-o/cri-o-cri-o-main-presubmits.yaml
@@ -1,7 +1,7 @@
 presubmits:
   cri-o/cri-o:
   - agent: jenkins
-    always_run: true
+    always_run: false
     branches:
     - ^main$
     - ^main-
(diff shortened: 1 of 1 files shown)


amit-62:
Please suggest improvement if any.

sohankunkerkar:
you might need to generate a new config using ` make jobs`

amit-62:
Hi @sohankunkerkar I have generated new config using `make jobs` but there is an error stating that `always_run:true` and `skip_if_only_changed:` is mutually exclusive. Please help me to solve it.

haircommander:
the docs seem to say that we can drop `always_run` and `skip_if_only_changed` will take its place (and run the tests in all cases except those listed in the regex)

amit-62:
Thank you @haircommander. Should I drop `always_run`. If it is set to false then it will not run automatically against all PRs.

haircommander:
> Thank you @haircommander. Should I drop always_run. If it is set to false then it will not run automatically against all PRs.

I think you can drop it. If it's false, the condition `skip_if_only_changed` will inform when to run the jobs

amit-62:
@haircommander I had removed `always_run` but `make jobs` automatically set it to false.

haircommander:
> @haircommander I had removed `always_run` but `make jobs` automatically set it to false.

fine with me!

/lgtm

haircommander:
/rehearsals-ack

well done with the investigation here, thanks for your work @amit-62!

amit-62:
> /rehearsals-ack
> 
> well done with the investigation here, thanks for your work @amit-62!

Thank you @haircommander.

PR #35408, opened 2023-01-16
MGMT-13325: MGMT-13300: Assisted-Installer: Add support with 4.13 dev preview images

eliorerz:
Add jobs for OCP 4.13, periodic ones will be added in the future after verifying that it working as expected.

/cc @osherdp @adriengentil

### DIFF
--- a/ci-operator/config/openshift/assisted-service/openshift-assisted-service-master__edge.yaml
+++ b/ci-operator/config/openshift/assisted-service/openshift-assisted-service-master__edge.yaml
@@ -856,6 +856,15 @@ tests:
         OPENSHIFT_VERSION=4.12
         OPENSHIFT_INSTALL_RELEASE_IMAGE=""
     workflow: assisted-baremetal
+- as: e2e-metal-assisted-4-13
+  optional: true
+  steps:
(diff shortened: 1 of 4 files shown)


adriengentil (approved):

ci-operator/config/openshift/assisted-service/openshift-assisted-service-master__edge.yaml
```diff
@@ -856,6 +856,14 @@ tests:
         OPENSHIFT_VERSION=4.12
         OPENSHIFT_INSTALL_RELEASE_IMAGE=""
     workflow: assisted-baremetal
+- as: e2e-metal-assisted-4-13
```
osherdp, line 859: maybe also make it optional for now (like in test-infra)

osherdp (changes requested):

ci-operator/config/openshift/assisted-service/openshift-assisted-service-master__edge.yaml
eliorerz: Updated. Thanks

osherdp (approved):

PR #35425, opened 2023-01-16
Updated observability release branches to release-2.8

dislbenn:
Signed-off-by: dislbenn <dbennett@redhat.com>


jlpadilla (approved):
/lgtm
/hold

We also need to update the [_prowconfig.yaml](https://github.com/openshift/release/blob/master/core-services/prow/02_config/stolostron/search-collector/_prowconfig.yaml) file on each project to protect the `release-2.7` branch so PRs and 2 approvals are required to merge into the release-2.7 branches.
**We can make this change in a separate PR because this one is already too big.**

PR #35438, opened 2023-01-17
Remove daily CI report

osherdp:
As it's already has a different infra (that is also much more comprehensive) this removes the existing jobs for the daily CI report.

Additional PR will remove the code from assisted-installer-deployment.

/cc @adriengentil @danmanor @gamli75


danmanor:
Added `hold` so you will trigger merge

PR #35440, opened 2023-01-17
ROX-14352: Mirror new image.

porridge:
Pick up https://github.com/stackrox/rox-ci-image/pull/176


gavin-stackrox (approved):

porridge:
The change to `clusters/` is a result of `make multi-arch-gen` that was complained about by CI.
@janisz can you please /lgtm this again (I bumped the version to the one built from the tag).

jmguzik (approved):

PR #35470, opened 2023-01-18
decrease 4.8 and 4.10 test frequency

jianlinliu:
Now 4.12 GAed, decrease 4.8 and 4.10 ci jobs test frequency.


jianlinliu:
cc @jhou1 @shellyyang1989 @wzheng1 for review.

PR #35495, opened 2023-01-18
METAL-461: Fix up  ipi-packet-check frequency (#32190)

derekhiggins:
This was recently changed as part of an effort
to blanket reduce the frequency of various jobs.
But only running it annually means leaked hosts

### DIFF
--- a/ci-operator/config/openshift-metal3/dev-scripts/openshift-metal3-dev-scripts-master__packetcheck.yaml
+++ b/ci-operator/config/openshift-metal3/dev-scripts/openshift-metal3-dev-scripts-master__packetcheck.yaml
@@ -20,7 +20,7 @@ resources:
       memory: 200Mi
 tests:
 - as: ipi-packet-check
-  cron: 13 11 24 */12 *
+  cron: 0 */2 * * *
   steps:
     cluster_profile: equinix-ocp-metal
     env:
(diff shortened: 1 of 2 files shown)


derekhiggins:
@jmguzik  this happened as part of https://github.com/openshift/release/pull/34511
how can I prevent it happening again?  (leaking baremetal hosts isn't cheap)

derekhiggins:
> @jmguzik this happened as part of #34511 how can I prevent it happening again? (leaking baremetal hosts isn't cheap)

hmm, I guess its because of the 4.6 image being used for the job?

cybertron:
/lgtm

I don't want to think about how much a baremetal server sitting idle for a year would cost. ;-)

andfasano:
Note: previously `0 */2 * * *`  was the recommended frequency value for the job because it's half the (default) job timeout (4h)

andfasano:
@derekhiggins could you please add a rehearse ack

jmguzik:
> hmm, I guess its because of the 4.6 image being used for the job?

@derekhiggins Most likely this is the reason I qualified it for the change. At this moment we do not have any sophisticated tooling that would help us to decrease exec frequency (WIP), but we need to do it because of the costs. I am sorry but such mistakes may happen, I will pay closer attention to the name as well which is clearly saying `master` in this case.

PR #35546, opened 2023-01-19
Run the integr8ly mapping job hourly

trepel:
We need to run this job more frequently, it is still being used.


jmguzik (approved):
I discussed with the author and advised to change this job to green asap to avoid falling back into the 'optimization' category again.

PR #35555, opened 2023-01-19
telco5g: Fix substitution for cluster DNS name and port

sshnaidm:


yuvalk:
you didnt explain why this is needed
is this a 4.12 thing only?

sshnaidm:
> you didnt explain why this is needed is this a 4.12 thing only?

Maybe, we don't care in this patch which version it is. Some kubeconfigs have 2 clusters defined, we need to replace these parameters in all clusters definitions.

yuvalk:
> > you didnt explain why this is needed is this a 4.12 thing only?
> 
> Maybe, we don't care in this patch which version it is. Some kubeconfigs have 2 clusters defined, we need to replace these parameters in all clusters definitions.

I guess I'm asking - are we sure this will work with prior versions too?

sshnaidm:
> > > you didnt explain why this is needed is this a 4.12 thing only?
> > 
> > 
> > Maybe, we don't care in this patch which version it is. Some kubeconfigs have 2 clusters defined, we need to replace these parameters in all clusters definitions.
> 
> I guess I'm asking - are we sure this will work with prior versions too?

yes, it was tested with 11 too, it just expands the previous functionality

gpuckett-redhat (approved):
/lgtm

PR #35574, opened 2023-01-20
Add ofcir service

derekhiggins:
Add the deployment and role required to run ofcir.

### DIFF
--- a/clusters/build-clusters/build05/ofcir/ofcir.yaml
+++ b/clusters/build-clusters/build05/ofcir/ofcir.yaml
@@ -1,0 +1,442 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    control-plane: controller-manager
+  name: ofcir-system
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: cipools.ofcir.openshift
+spec:
+  group: ofcir.openshift
+  names:
+    kind: CIPool
+    listKind: CIPoolList
+    plural: cipools
+    shortNames:
+    - cip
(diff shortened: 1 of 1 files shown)


clusters/build-clusters/build05/ofcir/ofcir.yaml
```diff
@@ -0,0 +1,578 @@
...
+      terminationGracePeriodSeconds: 10
+---
+kind: Group
+apiVersion: user.openshift.io/v1
+metadata:
+  name: ofcir-admin
+users:
+- afasano
+- dhiggins
+- tdomnesc
```
bradmwilliams, line 549: The preferred way to do this is via Rover Groups: https://docs.ci.openshift.org/docs/how-tos/rbac/#rover-groups

derekhiggins: thanks, group created.

clusters/build-clusters/build05/ofcir/ofcir.yaml
```diff
@@ -0,0 +1,569 @@
...
+  name: ofcir-service
+  namespace: ofcir-system
+spec:
+  ports:
+  - name: http
+    port: 8087
+    protocol: TCP
+  selector:
+    control-plane: controller-manager
+  type: NodePort
```
andfasano, line 459: I think this should work (ie visible by normal Prow jobs running on build05 cluster) also by setting a `ClusterIP` just to limit internally the visibility

derekhiggins: done

jmguzik (approved):

PR #35579, opened 2023-01-20
Fix issue DPTP-3132 Retry on retrieving kubeadmin password

bear-redhat:
Add delay to get kubeadmin password to mitigate a bug (kubeadmin secret is not available sometimes when retrieved right after the cluster becomes available)

### DIFF
--- a/ci-operator/step-registry/hypershift/hostedcluster/create/hostedcluster/hypershift-hostedcluster-create-hostedcluster-commands.sh
+++ b/ci-operator/step-registry/hypershift/hostedcluster/create/hostedcluster/hypershift-hostedcluster-create-hostedcluster-commands.sh
@@ -71,21 +71,29 @@ done
 
 # The timeout should be much lower, this is due to https://bugzilla.redhat.com/show_bug.cgi?id=2060091
 echo "Waiting for cluster to become available"
-oc wait --timeout=30m --for=condition=Available --namespace=clusters hostedcluster/${CLUSTER_NAME}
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/hypershift/hostedcluster/create/hostedcluster/hypershift-hostedcluster-create-hostedcluster-commands.sh
```diff
@@ -71,21 +71,29 @@ done
 
 # The timeout should be much lower, this is due to https://bugzilla.redhat.com/show_bug.cgi?id=2060091
 echo "Waiting for cluster to become available"
-oc wait --timeout=30m --for=condition=Available --namespace=clusters hostedcluster/${CLUSTER_NAME}
+oc wait --timeout=30m --for=condition=Available --namespace=clusters hostedcluster/${CLUSTER_NAME} || {
+  echo "Cluster did not become available"
+  oc get hostedcluster --namespace=clusters -o yaml ${CLUSTER_NAME}
```
droslean, line 76: Is it possible to leak any sensitive information with this command?

bear-redhat: Should be fine, the dump does not include any secrets to the ephemeral cluster.

droslean (approved):

PR #35596, opened 2023-01-23
Fix gofmt job of placement-operator

gibizer:
The result of make gofmt was ignored with a || true and the job failed with vendoring. This patch fixes the vendoring issue and ensures that the exit code of the make target is not ignored

### DIFF
--- a/ci-operator/config/openstack-k8s-operators/placement-operator/openstack-k8s-operators-placement-operator-master.yaml
+++ b/ci-operator/config/openstack-k8s-operators/placement-operator/openstack-k8s-operators-placement-operator-master.yaml
@@ -41,7 +41,9 @@ tests:
   container:
     from: src
 - as: gofmt
-  commands: make gofmt || true
+  commands: |
+    export GOFLAGS=-mod=mod
+    make gofmt
   container:
     from: src
 zz_generated_metadata:


gibizer:
not the gofmt job runs without the vendoring error, detects the wrongly formatted placement code, but the job still don't fail :/

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_release/35596/rehearse-35596-pull-ci-openstack-k8s-operators-placement-operator-master-gofmt/1617462978167181312

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_release/35596/rehearse-35596-pull-ci-openstack-k8s-operators-placement-operator-master-gofmt/1617462978167181312/artifacts/test/build-log.txt
```
go: downloading github.com/josharian/intern v1.0.0
test/functional/placementapi_controller_test.go
+ popd
```

gibizer:
`go fmt` does not set the exit code to != 0 if formatting was needed. :/ I guess it would be the same effort to introduce the pre-commit job for placement than to fix the gofmt.sh in the CI repo. So I will do that instead as it gives us more gains.

gibizer:
I will introduce pre-commit instead https://github.com/openshift/release/pull/35600 and I suggest to retire the gofmt job as it does not have value in the current form.

gibizer:
I will introduce pre-commit instead https://github.com/openshift/release/pull/35600 and I suggest to retire the gofmt job as it does not have value in the current form.

PR #35605, opened 2023-01-23
Ensure that manifests can be generated before test

gibizer:


gibizer:
This is needed for https://github.com/openstack-k8s-operators/placement-operator/pull/103

PR #35606, opened 2023-01-23
MGMT-13210: make cloud_hotfix_releases catch-up with latest config changes

osherdp:
Seems like the following changes didn't arrive to the ``cloud_hotfix_releases`` config:
* Using ``ocm-2.8`` image sets (master/main branches) for the components
* Using Dockerfiles as they are defined in the revision of the pull-request (instead of the target branch)
* OCP-specific jobs for 4.12 and 4.13


osherdp:
/cc @eliorerz @adriengentil 
it took me a while to understand why my changes in test-infra didn't reflect in ``cloud_hotfix_releases`` :sweat_smile:

adriengentil (approved):
:sweat_smile:

PR #35616, opened 2023-01-23
Validation fixed on multi-arch mirroring mappings generation

droslean:
It seems that the validation script is not checking the differences on namespaces while generation. So, this PR sorts the namespaces before the generation to avoid flakes and keep maintain the order in the output file and reactors the validation script to check the differences on namespaces too.


clusters/app.ci/multi-arch/namespaces.yaml
```diff
@@ -4,44 +4,44 @@ items:
 - apiVersion: v1
```
droslean, line 4: No changes should be made to this file. Only the ordering changes.

hongkailiu (approved):

PR #35639, opened 2023-01-24
Revert "add dummy excluded branch to verify tide config bug "

smg247:
https://github.com/kubernetes/test-infra/pull/28460 has been included in the image and we no longer need this workaround.

Reverts openshift/release#35532


smg247:
/unhold
https://github.com/openshift/release/pull/35794 was merged which included my bug fix for the second case.

PR #35641, opened 2023-01-24
Removes all checks for the master branch of stackrox/collector

Molter73:
We've removed the mirroring job that provided new versions of the downstream built drivers, so the remaining tests are not really doing anything useful, might as well remove them.


Stringy:
Is it worth also removing the master versions of the steps in `ci-operator/step-registry/stackrox/collector`?

Molter73:
I was thinking we can keep them just in case, at least until the versions we released on OSCI are deprecated, at which point we can go ahead and remove everything else.

PR #35647, opened 2023-01-24
GCP: Updated GCP xpn job to create resources

barbacbd:
** Firewall rules, router(s), and subnets are created for the gcp ipi xpn job. 
** Resources that were created should also be cleaned up.

[CORS-2381](https://issues.redhat.com/browse/CORS-2381)

### DIFF
--- a/ci-operator/step-registry/ipi/conf/gcp/xpn/ipi-conf-gcp-xpn-commands.sh
+++ b/ci-operator/step-registry/ipi/conf/gcp/xpn/ipi-conf-gcp-xpn-commands.sh
@@ -4,13 +4,104 @@ set -o nounset
 set -o errexit
 set -o pipefail
 
+trap 'CHILDREN=$(jobs -p); if test -n "${CHILDREN}"; then kill ${CHILDREN} && wait; fi' TERM
+
+GOOGLE_PROJECT_ID="$(< ${CLUSTER_PROFILE_DIR}/openshift_gcp_project)"
+export GCP_SHARED_CREDENTIALS_FILE="${CLUSTER_PROFILE_DIR}/gce.json"
+sa_email=$(jq -r .client_email ${GCP_SHARED_CREDENTIALS_FILE})
+if ! gcloud auth list | grep -E "\*\s+${sa_email}"
+then
+  gcloud auth activate-service-account --key-file="${GCP_SHARED_CREDENTIALS_FILE}"
+
+fi
+
+XPN_SERVICE_ACCOUNT=`gcloud config get-value account`
+HOST_PROJECT="openshift-dev-installer"
+# set the project in order to create the resources
+gcloud config set project "${HOST_PROJECT}"
+
+echo "$(date -u --rfc-3339=seconds) - Configuring GCP Subnets for Host Project..."
+
+REGION="${LEASED_RESOURCE}"
+INSTANCE_PREFIX="${NAMESPACE}-${JOB_NAME_HASH}"
+HOST_PROJECT_CONTROL_SUBNET="${INSTANCE_PREFIX}-subnet-1"
+HOST_PROJECT_COMPUTE_SUBNET="${INSTANCE_PREFIX}-subnet-2"
+HOST_PROJECT_NETWORK="${INSTANCE_PREFIX}-vpc"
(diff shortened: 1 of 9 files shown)


barbacbd:
/pj-rehearse pull-ci-openshift-installer-master-e2e-gcp-ovn-xpn

barbacbd:
/pj-rehearse pull-ci-openshift-installer-master-e2e-gcp-ovn-xpn

ci-operator/step-registry/ipi/conf/gcp/xpn/ipi-conf-gcp-xpn-commands.sh
```diff
@@ -4,13 +4,104 @@ set -o nounset
...
+
+echo "$(date -u --rfc-3339=seconds) - Configuring GCP Subnets for Host Project..."
+
+REGION="${LEASED_RESOURCE}"
+INSTANCE_PREFIX="${NAMESPACE}-${JOB_NAME_HASH}"
+HOST_PROJECT_CONTROL_SUBNET="${INSTANCE_PREFIX}-subnet-1"
+HOST_PROJECT_COMPUTE_SUBNET="${INSTANCE_PREFIX}-subnet-2"
+HOST_PROJECT_NETWORK="${INSTANCE_PREFIX}-vpc"
+CONTROL_SUBNET_CIDR="10.0.64.0/19"
+COMPUTE_SUBNET_CIDR="10.0.128.0/19"
```
patrickdillon, line 31: This is fine for the PoC, but I think having these as static values will prevent multiple jobs from running simultaneously. The subnets will need to have non-overlapping CIDRs, so we will need to manage these and make the values populated in the install config dynamic.

patrickdillon:
/pj-rehearse pull-ci-openshift-installer-master-e2e-gcp-ovn-xpn

patrickdillon:
I created https://github.com/openshift/installer/pull/6849 so we'll quit getting spammed in the logs here

ci-operator/step-registry/ipi/conf/gcp/xpn/ipi-conf-gcp-xpn-commands.sh
```diff
@@ -4,13 +4,104 @@ set -o nounset
...
+       --project "${HOST_PROJECT}" \
+       --region "${REGION}"
+
+gcloud beta compute networks subnets set-iam-policy "${HOST_PROJECT_CONTROL_SUBNET}" \
+       /tmp/subnet-policy.json \
+       --project "${HOST_PROJECT}" \
+       --region "${REGION}"
+
+# remove the policy created above
+rm -rf subnet-policy.json
```
patrickdillon, line 98: It looks like we're leaking this policy binding... I suspect this needs to be cleaned up as well? 

A pattern we can use to clean this up is at this point write a policy patch and save to the shared dir. In the cleanup step apply the patch, perhaps using `remove-iam-policy-binding`

ci-operator/step-registry/ipi/conf/gcp/xpn/post/ipi-conf-gcp-xpn-post-commands.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+set -euo pipefail
+
+trap 'CHILDREN=$(jobs -p); if test -n "${CHILDREN}"; then kill ${CHILDREN} && wait; fi' TERM
+
+#GOOGLE_PROJECT_ID="$(< ${CLUSTER_PROFILE_DIR}/openshift_gcp_project)"
```
patrickdillon, line 6: nit: delete this

patrickdillon (commented):
This looks really good. Just a couple of comments.

We will want to fix this CIDR thing ASAP--I think this will prevent more than one job running, but it can be done separately.

ci-operator/step-registry/ipi/conf/gcp/xpn/ipi-conf-gcp-xpn-commands.sh
barbacbd: Looks like the policy is cleaned up when the subnet is removed (or I have no idea where this is stored). If this does persist there are more jobs that need to be adjusted too.

PR #35680, opened 2023-01-25
Using the KMM image built in CI instead of a pre-built image.

ybettan:


ybettan:
/hold
I am not sure it is working

PR #35682, opened 2023-01-25
Use the sample from the PR under test for deployment

gibizer:
When OpenStackControlPlane CR is created during the e2e test the job should use the sample from the PR under test instead of the sample from master as the PR might change that sample to adapt it to changes in the PR itself.

### DIFF
--- a/ci-operator/step-registry/openstack-k8s-operators/deploy/openstack-k8s-operators-deploy-commands.sh
+++ b/ci-operator/step-registry/openstack-k8s-operators/deploy/openstack-k8s-operators-deploy-commands.sh
@@ -2,6 +2,7 @@
 
 set -ex
 
+ORG="openstack-k8s-operators"
 OPENSTACK_OPERATOR="openstack-operator"
 
 # We don't want to use OpenShift-CI build cluster namespace
@@ -55,8 +56,8 @@ oc get csv ${INSTALLED_CSV} -o jsonpath='{.spec.install.spec.deployments[*].name
 timeout ${TIMEOUT_OPERATORS_AVAILABLE} xargs -I {} -d ' ' \
 sh -c 'oc wait --for=condition=Available deployment {} --timeout=-1s'
 
-# Deploy openstack services
-make openstack_deploy
+# Deploy openstack services with the sample from the PR under test
+OPENSTACK_CR=/go/src/github.com/${ORG}/${OPENSTACK_OPERATOR}/config/samples/core_v1beta1_openstackcontrolplane.yaml make openstack_deploy
 sleep 60
 
 # Waiting for all services to be ready


ci-operator/step-registry/openstack-k8s-operators/deploy/openstack-k8s-operators-deploy-commands.sh
```diff
@@ -55,8 +56,8 @@ oc get csv ${INSTALLED_CSV} -o jsonpath='{.spec.install.spec.deployments[*].name
 timeout ${TIMEOUT_OPERATORS_AVAILABLE} xargs -I {} -d ' ' \
 sh -c 'oc wait --for=condition=Available deployment {} --timeout=-1s'
 
-# Deploy openstack services
-make openstack_deploy
+# Deploy openstack services with the sample from the PR under test
+OPENSTACK_CR=/go/src/github.com/${ORG}/${OPENSTACK_OPERATOR}/samples/core_v1beta1_openstackcontrolplane.yaml make openstack_deploy
```
viroel, line 60: i think that config dir is missing in path:
"config/samples/core_v1beta1_openstackcontrolplane.yaml"

viroel: Maybe we should try to fetch the entire code user's commit and place in OPERATOR_BASE_DIR.
But this will require more work and tests. We can go with this if works to unblock other PRs

viroel (commented):
Thanks Gigi

ci-operator/step-registry/openstack-k8s-operators/deploy/openstack-k8s-operators-deploy-commands.sh
gibizer: > i think that config dir is missing in path: "config/samples/core_v1beta1_openstackcontrolplane.yaml"

You are right. I've fixed that now. 

> Maybe we should try to fetch the entire code user's commit and place in OPERATOR_BASE_DIR.

I assume that `/go/src/github.com/${ORG}/${OPENSTACK_OPERATOR}` is the clone of the repo where the PR is checked out. Isn't it?

viroel: It is, but I'm considering providing OPENSTACK_REPO and OPENSTACK_BRANCH to install_yamls, so it can fetch the PR code in OPERATOR_BASE_DIR, instead of cloning 'github.com/openstack-k8s-operators/openstack-operator@master'.
I will try that in another PR.

gibizer:
/pj-rehearse ack

It seems install_yamls did what we requested:
```console
~/install_yamls
cp /go/src/github.com/openstack-k8s-operators/openstack-operator/config/samples/core_v1beta1_openstackcontrolplane.yaml /alabama/install_yamls/out/openstack/openstack/cr
```
So I think it is ready now.

ci-operator/step-registry/openstack-k8s-operators/deploy/openstack-k8s-operators-deploy-commands.sh
gibizer: hm, yeah, let's do that in a separate PR, but that make sense too.

PR #35686, opened 2023-01-25
Add secureboot workflow

eranco74:


ci-operator/step-registry/ipi/conf/gcp/secureboot/ipi-conf-gcp-secureboot-commands.sh
```diff
@@ -0,0 +1,22 @@
...
+set -o pipefail
+
+CONFIG="${SHARED_DIR}/install-config.yaml"
+
+PATCH="${SHARED_DIR}/secure_boot.yaml.patch"
+cat > "${PATCH}" << EOF
+controlPlane:
+  platform:
+    gcp:
+      secureBoot: "${SECURE_BOOT}"
```
vrutkovs, line 14: .yml should document supported env vars used in this script

vrutkovs: Also, do we need it parametrized? iiuc the install config section just needs `Enabled` set any time we include this step

eranco74: True, I just copy pasted it from the previous config when it was optional, no need for env var now.

vrutkovs (approved):

PR #35692, opened 2023-01-25
ocm - cut new release 2.8 (config-policy-controller)

JustinKuli:
https://issues.redhat.com/browse/ACM-2734


dhaiducek:
Both `cert` and `config` are in here? I guess that was intentional?

dhaiducek (approved):

JustinKuli:
@dhaiducek 
> Both cert and config are in here? I guess that was intentional?

Nope, but if you're ok with it, I'll just close the cert PR: https://github.com/openshift/release/pull/35700.

I think I just had a local oddity when running a script

PR #35703, opened 2023-01-25
Kill all references to service-idler

danwinship:
We have never shipped this for 4.x and never plan to. (It is not part of the general "service idling" feature; it was an openshift-online-specific add-on in 3.x which is not used any more.) The repo is going to be archived.

The fact that there are references to this in weird places (eg, the top-level `Makefile`) seems to be because (a) it's really old, and (b) it hasn't been updated since 3.x days. Back when the `make service-idler` rule was added, there were also make rules for `prometheus`, `node-exporter`, `oauth-proxy`, and lots of other things. But those all went away as release components and processes got more normalized for 4.x, but the service-idler rule didn't, because service-idler never became a "normal" 4.x component.

/cc @abhat


abhat:
LGTM

PR #35717, opened 2023-01-25
agent-installer-utils: disable golangci-lint

celebdor:
golangci-lint container builds are not able to deal with C library wrapping. This manifests on agent-installer-utils by complaining about the nmstate library saying that it doesn't export data.

For the root cause, I ran the golangci-lint container against the nmstate golang library and found the aforementioned issue. If I use golangci-lint installed directly in my f37 system the issue does not reproduce, so I guess it should be possible to make a new golangci-lint image that had the necessary support for doing type checks on C libraries. I also tested disabling the typecheck linter but that did not work out.

Signed-off-by: Antoni Segura Puimedon <antoni@redhat.com>


ci-operator/config/openshift/agent-installer-utils/openshift-agent-installer-utils-main.yaml
```diff
@@ -33,20 +33,10 @@ tests:
     from: src
   skip_if_only_changed: (^docs/)|((^|/)OWNERS(_ALIASES)?$)|((^|/)[A-Z]+\.md$)
 - as: govet
-  commands: ./hack/go-vet.sh ./...
+  commands: IS_CONTAINER=TURE ./hack/go-vet.sh ./...
```
zaneb, line 36: TURE

celebdor: ouch.

PR #35719, opened 2023-01-26
Delete rpmdb lock files after overriding openshift-hyperkube RPM

jlebon:
Running an rpm transaction will leave around various lock files in the rpmdb. This is usually OK, but here we're building an image and we don't want them in there, so delete them. This is something rpm-ostree does also in postprocessing during regular tree composes and is probably something it should learn to do client-side when pulling down a layered container too. But ideally we'd enforce it at image build time. For now hack around this here.

More practically, this breaks the bdb backend of libsolv for some reason, causing errors like this to get thrown:

    error: db5 error(-30973) from dbenv->open: BDB0087 DB_RUNRECOVERY:
    Fatal error, run database recovery

Fixes https://issues.redhat.com/browse/OCPBUGS-6505

### DIFF
--- a/ci-operator/config/cri-o/cri-o/cri-o-cri-o-main.yaml
+++ b/ci-operator/config/cri-o/cri-o/cri-o-cri-o-main.yaml
@@ -23,7 +23,8 @@ images:
     # See also https://issues.redhat.com/browse/ART-4352
     FROM registry.ci.openshift.org/rhcos-devel/rhel-coreos-8
     COPY --from=build /*.rpm /
-    RUN rpm -Uvh --oldpackage /*.rpm && rm -f /*.rpm
+    # Also delete leftover RPM files; see https://issues.redhat.com/browse/OCPBUGS-6505.
+    RUN rpm -Uvh --oldpackage /*.rpm && rm -vf /*.rpm /usr/share/rpm/__db.* /usr/share/rpm/.*.lock
   inputs:
     base-machine-with-rpms:
       as:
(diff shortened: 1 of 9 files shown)


travier:
Shouldn't all those cases use `rpm-ostree override replace` instead to do things more cleanly?

Won't block on that improvement however so:
/lgtm

travier:
CC @cgwalters
From https://github.com/openshift/release/commit/f3b343eb5298fd926a2fe6fb55bb44c45e679380

jlebon:
> Shouldn't all those cases use `rpm-ostree override replace` instead to do things more cleanly?

That would work too (though interestingly, it does also leave around a lock file but that one doesn't break libsolv), but I think we do want to support `rpm -Uvh` being called like this. I guess if we enable cliwrap in the container, we could clean up after it. Or alternatively do the cleanup as part of `ostree container commit`, though ostree-rs-ext is currently package-agnostic. (Maybe we should have an `rpm-ostree container commit` wrapper that also does rpm-ostree-specific cleanups.)

cgwalters:
> . I guess if we enable cliwrap in the container, we could clean up after it. Or alternatively do the cleanup as part of ostree container commit, though ostree-rs-ext is currently package-agnostic. (Maybe we should have an rpm-ostree container commit wrapper that also does rpm-ostree-specific cleanups.)

I lean more towards cliwrap but...yep.

But bigger picture I think this bug doesn't exist with sqlite anymore, so with rhel9 we can stop caring about it.

travier:
This now needs a `rehearsals-ack` label but no idea who can set that.
\label rehearsals-ack

PR #35733, opened 2023-01-26
[telco5g] configure cnf-tests to use prepared cluster

gpuckett-redhat:
cuts down on wasted time during testing

### DIFF
--- a/ci-operator/step-registry/telco5g/cluster-setup/telco5g-cluster-setup-commands.sh
+++ b/ci-operator/step-registry/telco5g/cluster-setup/telco5g-cluster-setup-commands.sh
@@ -255,7 +255,7 @@ cat << EOF > $SHARED_DIR/check-cluster.yml
 
   - name: Fail when cluster is not available
     shell: "echo Cluster ready: {{ ready_check.stdout }}"
-    failed_when: "True" not in ready_check.stdout 
+    failed_when: "'True' not in ready_check.stdout"
 EOF
 
 # PROCEED_AFTER_FAILURES is used to allow the pipeline to continue past cluster setup failures for information gathering. 


ci-operator/step-registry/telco5g/cluster-setup/telco5g-cluster-setup-commands.sh
```diff
@@ -232,9 +239,9 @@ cat << EOF > ~/fetch-information.yml
...
-ANSIBLE_STDOUT_CALLBACK=debug ansible-playbook -i $SHARED_DIR/inventory ~/ocp-install.yml -vv || status=$?
-ansible-playbook -i $SHARED_DIR/inventory ~/fetch-kubeconfig.yml -vv || true
+if [[ "$T5CI_JOB_TYPE" != "cnftests" ]]; then
+  ANSIBLE_STDOUT_CALLBACK=debug ansible-playbook -i $SHARED_DIR/inventory ~/ocp-install.yml -vv || status=$?
+fi 
+ansible-playbook -i $SHARED_DIR/inventory ~/fetch-kubeconfig.yml -vv
 ANSIBLE_STDOUT_CALLBACK=debug ansible-playbook -i $SHARED_DIR/inventory ~/fetch-information.yml -vv || true
-exit ${status}
-
+exit ${status}
```
sshnaidm, line 247: I think you removed the new line here.
You can configure your IDE always to leave the blank line in the end of file.

ci-operator/step-registry/telco5g/cluster-setup/telco5g-cluster-setup-commands.sh
```diff
@@ -232,9 +239,9 @@ cat << EOF > ~/fetch-information.yml
     shell: kcli ssh root@${CLUSTER_NAME}-installer 'oc get node'
 EOF
 
-status=0
```
sshnaidm, line 235: Why `status` variable initialization is removed?

ci-operator/step-registry/telco5g/cluster-setup/telco5g-cluster-setup-commands.sh
```diff
@@ -232,9 +239,9 @@ cat << EOF > ~/fetch-information.yml
...
-ANSIBLE_STDOUT_CALLBACK=debug ansible-playbook -i $SHARED_DIR/inventory ~/ocp-install.yml -vv || status=$?
-ansible-playbook -i $SHARED_DIR/inventory ~/fetch-kubeconfig.yml -vv || true
+if [[ "$T5CI_JOB_TYPE" != "cnftests" ]]; then
+  ANSIBLE_STDOUT_CALLBACK=debug ansible-playbook -i $SHARED_DIR/inventory ~/ocp-install.yml -vv || status=$?
+fi 
+ansible-playbook -i $SHARED_DIR/inventory ~/fetch-kubeconfig.yml -vv
```
sshnaidm, line 245: `|| true` is removed

ci-operator/step-registry/telco5g/cluster-setup/telco5g-cluster-setup-commands.sh
```diff
@@ -232,9 +239,9 @@ cat << EOF > ~/fetch-information.yml
     shell: kcli ssh root@${CLUSTER_NAME}-installer 'oc get node'
 EOF
 
-status=0
-ANSIBLE_STDOUT_CALLBACK=debug ansible-playbook -i $SHARED_DIR/inventory ~/ocp-install.yml -vv || status=$?
-ansible-playbook -i $SHARED_DIR/inventory ~/fetch-kubeconfig.yml -vv || true
+if [[ "$T5CI_JOB_TYPE" != "cnftests" ]]; then
```
sshnaidm, line 242: We need to keep here possibility for other jobs to fetch kubeconfig if exists even if cluster failed, for next `gather-pao` step which will gather debug information. That's why I put here `|| true` after each call. This should remain the same.
For case of cnf-tests job it shouldn't be the case, if cluster is not installed properly, just fail .
Please rethink the logic here including all above.

sshnaidm (changes requested):

ci-operator/step-registry/telco5g/cluster-setup/telco5g-cluster-setup-commands.sh
gpuckett-redhat: this was unintentional, will be corrected

ci-operator/step-registry/telco5g/cluster-setup/telco5g-cluster-setup-commands.sh
gpuckett-redhat: added a variable `$NOT_CNFTESTS` that is set to 'false' unless the job type is `cnftests`

so in normal execution, it will still evaluate to `|| true`, which will allow the gather-pao step to work. however, during CNF tests, it will evaluate to `|| false`, which accomplishes the desired outcome:

```
[gpuckett@fedora release]$ `/bin/false` || `/bin/false`
[gpuckett@fedora release]$ echo $?
1
[gpuckett@fedora release]$ `/bin/true` || `/bin/false`
[gpuckett@fedora release]$ echo $?
0
[gpuckett@fedora release]$ `/bin/false`
[gpuckett@fedora release]$ echo $?
1
```

ci-operator/step-registry/telco5g/cluster-setup/telco5g-cluster-setup-commands.sh
```diff
@@ -45,6 +45,13 @@ cat << EOF > $SHARED_DIR/bastion_inventory
 ${BASTION_IP} ansible_ssh_user=centos ansible_ssh_common_args="$COMMON_SSH_ARGS" ansible_ssh_private_key_file="${SSH_PKEY}"
 EOF
 
+ADDITIONAL_ARG=""
+if [[ "$T5CI_JOB_TYPE" == "cnftests" ]]; then
+  ADDITIONAL_ARG="--cluster-name cnfdc2 --force"
+else
+  ADDITIONAL_ARG="-e $CL_SEARCH"
```
sshnaidm, line 52: you don't want to allocate cnfdc2 to anyone as well, so you need to use `--exclude cnfdc2`
Maybe cnfdc2 should put into variable

ci-operator/step-registry/telco5g/cluster-setup/telco5g-cluster-setup-commands.sh
sshnaidm: In case of `cnftests` job run and cluster installation failed, we don't want to run any on these playbooks. But we need to ensure cluster is functional or not. I think about adding a specific playbook to check it. Or task in any of existing playbooks, anyway cluster should be up for any job to run.

cluster is up:
```
$ kcli ssh root@cnfdu9-installer "oc get clusterversion"
NAME      VERSION                              AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.12.0-0.nightly-2023-01-24-150415   True        False         20h     Cluster version is 4.12.0-0.nightly-2023-01-24-150415
```
cluster is not up:
```
[kni@cnfdu0 ~]$ kcli ssh root@cnfdu3-installer "oc get clusterversion"
version             False       True          31h     Unable to apply 4.11.0-0.nightly-2023-01-25-044931: the cluster operator monitoring has not yet successfully rolled out
```

ci-operator/step-registry/telco5g/cluster-setup/telco5g-cluster-setup-commands.sh
```diff
@@ -232,9 +239,12 @@ cat << EOF > ~/fetch-information.yml
...
-ANSIBLE_STDOUT_CALLBACK=debug ansible-playbook -i $SHARED_DIR/inventory ~/fetch-information.yml -vv || true
+NOT_CNFTESTS="false"
+status=0 
+if [[ "$T5CI_JOB_TYPE" != "cnftests" ]]; then
+  NOT_CNFTESTS="true"
+  ANSIBLE_STDOUT_CALLBACK=debug ansible-playbook -i $SHARED_DIR/inventory ~/ocp-install.yml -vv || status=$?
+fi 
+ansible-playbook -i $SHARED_DIR/inventory ~/fetch-kubeconfig.yml -vv || `$NOT_CNFTESTS`
+ANSIBLE_STDOUT_CALLBACK=debug ansible-playbook -i $SHARED_DIR/inventory ~/fetch-information.yml -vv || `$NOT_CNFTESTS`
 exit ${status}
```
sshnaidm, line 250: Add a new line please

ci-operator/step-registry/telco5g/cluster-setup/telco5g-cluster-setup-commands.sh
```diff
@@ -232,9 +242,32 @@ cat << EOF > ~/fetch-information.yml
...
-ANSIBLE_STDOUT_CALLBACK=debug ansible-playbook -i $SHARED_DIR/inventory ~/fetch-information.yml -vv || true
+cat << EOF > ~/check-cluster.yml
+---
+- name: Check if cluster is ready
+  hosts: hypervisor
+  gather_facts: false
+  tasks:
+
+  - name: Check if cluster is available
+    shell: kcli ssh root@${CLUSTER_NAME}-installer "oc get clusterversion -o json|jq '.items[0].status.conditions[]|select(.type==\"Available\").status'"
```
sshnaidm, line 253: 1. it failed for me while running:
```
[kni@cnfdd1 ~]$ kcli ssh root@cnfdd3-installer "oc get clusterversion -o json|jq '.items[0].status.conditions[]|select(.type==\"Available\").status'"
jq: error: Available/0 is not defined at <top-level>, line 1:
.items[0].status.conditions[]|select(.type==Available).status                                            
jq: 1 compile error
```

2. please test it on a live environment with possible edge cases - cluster is already installed, not installed, command fails etc
if this fails, then next task won't be executed

3. It requires `jq` to be installed on all hypervisors, which should be handled separately, don't assume it's there.

ci-operator/step-registry/telco5g/cluster-setup/telco5g-cluster-setup-commands.sh
```diff
@@ -232,9 +242,32 @@ cat << EOF > ~/fetch-information.yml
...
+    register: ready_check
+
+  - name: Fail when cluster is not available
```
sshnaidm, line 256: if previous command fails, this won't run (which is OK actually)

ci-operator/step-registry/telco5g/cluster-setup/telco5g-cluster-setup-commands.sh
```diff
@@ -232,9 +242,32 @@ cat << EOF > ~/fetch-information.yml
...
+    shell: "echo Cluster ready: {{ ready_check.stdout }}"
+    failed_when: "False" in ready_check.stdout 
+EOF
+
+
+NOT_CNFTESTS="false"
```
sshnaidm, line 262: can you please rename this variable, it's too much negatives here, hard to understand what it means
also please add a comment explaining its usage

sshnaidm (changes requested):

ci-operator/step-registry/telco5g/cluster-setup/telco5g-cluster-setup-commands.sh
gpuckett-redhat: The command works when run directly in the terminal from the installer machine. It seems like there is a problem with how the command is parsed when fed through `kcli ssh`

```
[root@cnfda3-installer ~]# oc get clusterversion -o json|jq '.items[0].status.conditions[]|select(.type=="Available").status'
"False"
```

I'm trying find a way to escape them so that they'll be fed through properly.

sshnaidm: This task is supposed to run on hypervisor, so better to test it on it.

gpuckett-redhat: The 'oc get clusterversion' command already runs on the installer vm though, isn't it less efficient to pipe that data back out to the hypervisor just to run a parser on it?

gpuckett-redhat: I came up with a solution, I think... it's in the latest commit

ci-operator/step-registry/telco5g/cluster-setup/telco5g-cluster-setup-commands.sh
```diff
@@ -232,9 +242,44 @@ cat << EOF > ~/fetch-information.yml
...
+
+cat << EOF > ~/check-cluster.yml
+---
+- name: Check if cluster is ready
+  hosts: hypervisor
+  gather_facts: false
+  tasks:
+
+  - name: Copy test script to installer vm
+    shell: kcli scp ~/test-jq.sh root@${CLUSTER_NAME}-installer:/tmp/test-jq.sh
```
sshnaidm, line 259:
```suggestion
    shell: kcli scp $SHARED_DIR/test-jq.sh root@${CLUSTER_NAME}-installer:/tmp/test-jq.sh
```

ci-operator/step-registry/telco5g/cluster-setup/telco5g-cluster-setup-commands.sh
```diff
@@ -232,9 +242,44 @@ cat << EOF > ~/fetch-information.yml
...
 EOF
 
-status=0
-ANSIBLE_STDOUT_CALLBACK=debug ansible-playbook -i $SHARED_DIR/inventory ~/ocp-install.yml -vv || status=$?
-ansible-playbook -i $SHARED_DIR/inventory ~/fetch-kubeconfig.yml -vv || true
-ANSIBLE_STDOUT_CALLBACK=debug ansible-playbook -i $SHARED_DIR/inventory ~/fetch-information.yml -vv || true
+cat << EOF > ~/test-jq.sh
+  #!/bin/bash
+  # simple bash script to check if cluster is available
+  oc get clusterversion -o json|jq '.items[0].status.conditions[]|select(.type=="Available").status'
```
sshnaidm, line 248: well, there is definitely no `jq` on the installer, you need to install it firstly.

sshnaidm: Why just not something like `kcli ssh root@cnfdd3-installer 'oc get clusterversion'  | awk {'print $3'} | tail -1`

sshnaidm (changes requested):

ci-operator/step-registry/telco5g/cluster-setup/telco5g-cluster-setup-commands.sh
gpuckett-redhat: When using `kcli ssh root@cnfdd3-installer 'oc get clusterversion'  | awk {'print $3'} | tail -1` it treats the empty 'version' column as a separator, and `$3` ends up being the 'True' from 'Progressing' instead of the 'False' from 'Available'... this means that whether the cluster is available or not, this particular check would return 'True' in a lot of cases where it would not be desirable.

This is initially why I opted to parse the json, since human-readable output can be hard to deal with and the structure of the json output is more predictable.

However, I've found a way to do this without `jq` -- using the -o jsonpath flag for the oc command.

sshnaidm:
I think you can test it now on live: comment on this pr with `/pj-rehearse periodic-ci-openshift-release-master-nightly-4.12-e2e-telco5g-cnftests` and take a look on `cnfdc2` cluster. Make sure it's not taken before.

sshnaidm:
/pj-rehearse periodic-ci-openshift-release-master-nightly-4.12-e2e-telco5g-cnftests

PR #35754, opened 2023-01-26
WIP: Setup cspi-qe oadp-qe-automation private repo

madunn:


ci-operator/config/cspi-qe/oadp-qe-automation/cspi-qe-oadp-qe-automation-main.yaml
```diff
@@ -0,0 +1,82 @@
...
+    name: ubi
+    namespace: ocp
+    tag: "8"
+  upi-installer:
+    name: "4.12"
+    namespace: ocp
+    tag: upi-installer
+build_root:
+  project_image:
+    dockerfile_literal: |
```
ascerra, line 16: Why are we using dockerfile_literal here instead of creating a custom image from a dockerfile.

When we use dockerfile_literal in a config file we are essentially saying that this particular dockerfile is only for this one config file. It will not be able to be used by anything else in OCP CI without directly duplicating it into another config file.

PR #35758, opened 2023-01-27
enable netobserv tests

memodi:
enabling netobserv tests for select profiles

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.13__amd64-nightly.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.13__amd64-nightly.yaml
@@ -500,8 +500,7 @@ tests:
       E2E_RUN_TAGS: '@aws-ipi and @network-ovnkubernetes and @hypershift-hosted and
         not @fips'
       TAG_VERSION: '@4.13'
-      TEST_ADDITIONAL: 
-      
+      TEST_ADDITIONAL: Hypershift
       TEST_FILTERS: ~ChkUpgrade&;~DisconnectedOnly&;~NonPreRelease&;~HyperShiftMGMT&;~MicroShiftOnly&;~NonHyperShiftHOST&;~Serial&;~Disruptive&
       TEST_TIMEOUT: "30"
     test:


jotak:
Hey @memodi / all,

FYI, there's like a criss-crossing between the QE team and the dev team here: QE is enabling netobserv tests through this PR, and in the same time I want to remove the upstream / github integration that we have, cf https://github.com/openshift/release/pull/35766 because we're using github actions.

I don't think this is conflictual, but let me know if you think otherwise

errors
```diff
@@ -0,0 +1,5 @@
+[failed to validate configuration ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.11__amd64-nightly.yaml: Failed resolve MultiStageTestConfiguration test/gcp-ipi-sdn-migration-ovn-p2-f7: workflow/cucushift-installer-rehearse-gcp-ipi-sdn: parameter \"TEST_ADDITIONAL\" is overridden in [test/gcp-ipi-sdn-migration-ovn-p2-f7] but not declared in any step
```
kuiwang02, line 1: @memodi for test gcp-ipi-sdn-migration-ovn-p2-f7 which workflow is not for golang, you do not need to add TEST_ADDITIONAL for it. if you want it, you need to change chain cucushift-sdn-migration-ovn (but it seems the chain's target is not to run any golang and cucu case.)

same for gcp-ipi-sdn-migration-ovn-p2-f14.

memodi: @kuiwang02 - I removed migration tests in commit [1b0f16d](https://github.com/openshift/release/pull/35758/commits/1b0f16daa9c47b0eac66f0f400aa61cbf4796ab8) ,I don't need to enable for them, the errors are gone now.

kuiwang02: @memodi it is good. please remove this errors file from this PR. thanks

memodi: ah, I missed that, deleted it. thanks!

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.13__amd64-nightly.yaml
```diff
@@ -546,7 +551,7 @@ tests:
     cluster_profile: aws-qe
     env:
       BASE_DOMAIN: qe.devcluster.openshift.com
-      TEST_ADDITIONAL: Hypershift
+      TEST_ADDITIONAL: Hypershift&;Network_Observability
```
memodi, line 554: @kuiwang02 - is this correct way to enable for more than one?

kuiwang02: @memodi oh, no. I am sorry I was not aware of this.
it should be `TEST_ADDITIONAL: Hypershift|Network_Observability`

memodi: Fixed it, should be all set now.

kuiwang02:
by the way, frankly it should be approved by qe-productivity team.
because 
1, it does not add new test
2, it is only related to add golang case.
, I approve it.

or other PR which add test, or involved non-golang, have to get approve from qe-produtivity team.

PR #35548, opened 2023-01-19
assisted-nutanix: don't setup CSI and registry operator

vrutkovs:
This seems to block installation now, as CSI operators are not being applied to spoke cluster (assisted only)

This also now randomizes API/Ingress VIPs used so that several simultaneous rehearsals would not use the same IP pairs

PR #35778, opened 2023-01-27
NO-ISSUE: assisted nutanix fix fix_uid.sh

adriengentil:
fix_uid.sh is available in path when using
assisted-test-infra-internal image.


adriengentil:
Did not see https://github.com/openshift/release/pull/35548, closing
/close

PR #35788, opened 2023-01-27
Update osd-network-verifier CI image to Go 1.19

mjlshen:
This will let us update osd-network-verifier to use Go 1.19


boranx (commented):
/lgtm
/approve

PR #35789, opened 2023-01-27
ci-operator/openshift/os: 4.13+: Switch to 9.2

travier:
Remove all 8.6 CI jobs as they will now fail.


travier:
See: https://github.com/openshift/os/pull/1127

cgwalters:
/pj-rehearse ack
Don't need to rehearse removals

PR #35799, opened 2023-01-28
Fix cron entries

liangxia:
The upgrade jobs' cron entries are updated to execute those tests before 4.12 GA. And the e2e jobs' cron entries might be an copy/paste from existing versions/architectures. This PR restore them to the default ones (based on file name and test name)

For review,
/cc @shellyyang1989 @LiangquanLi930 @jianlinliu @jhou1 

For awareness, CC 4.12/4.13 release manager
/cc @chengzhang1016 @jhou1


shellyyang1989:
lgtm on the upgrade jobs. Thanks!

PR #35806, opened 2023-01-29
add osd-ccs for 4.13 EC build

yasun1:


ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.13__amd64-dev-preview.yaml
```diff
@@ -0,0 +1,60 @@
...
+    name: release
+    namespace: openshift
+    tag: golang-1.18
+releases:
+  latest:
+    candidate:
+      architecture: amd64
+      product: ocp
+      stream: nightly
+      version: "4.13"
```
yasun1, line 33: We do not use the version for provisioning an OSD cluster, we get the version in our steps. Is the 'releases' part required for the yaml format?

If the 'releases' part is required, due to now the DPTP does not support get version from the dev-preview and the EC build is a kind of nightly build,  the item 'stream' is set to 'nightly' to not block the OSD jobs. When the OCP side is ready, it can be changed to fit its requirement.

yasun1:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.13-amd64-dev-preview-aws-ocm-osd-ccs-p2-f7

yasun1:
@liangxia  Please help review.

PR #35815, opened 2023-01-29
Updating the image of `check-commits-count` job for KMM.

ybettan:
This image contains a more recent version of `git` which is `2.31.1`. The previous image was using version `1.8` and since we are starting to make use of the `git branch --show-current` command that was introduced in version `2.22` this is a required change.

---

/cc @yevgeny-shnaidman 
This is currently blocking https://github.com/rh-ecosystem-edge/kernel-module-management/pull/411


ybettan:
/pj-rehearse ack
Jobs are failing because there are no actual commits to the repo.

PR #35834, opened 2023-01-30
add manifest for the Dgraph deployment in ci namespace

droslean:
/cc @danilo-gemoli @openshift/test-platform 

### DIFF
--- a/clusters/app.ci/dgraph/admin_dgraph-sts.yaml
+++ b/clusters/app.ci/dgraph/admin_dgraph-sts.yaml
@@ -1,0 +1,106 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: dgraph-public
+  namespace: ci
+  labels:
+    app: dgraph
+spec:
+  type: ClusterIP
+  ports:
+  - port: 5080
+    targetPort: 5080
+    name: grpc-zero
+  - port: 6080
+    targetPort: 6080
+    name: http-zero
+  - port: 8080
+    targetPort: 8080
+    name: http-alpha
+  - port: 9080
+    targetPort: 9080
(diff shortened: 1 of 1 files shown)


droslean:
/hold

Despite the fact that the database will be populated by a periodic job, we don't consider the data sensitive since the source of truth is already exposed to the public. However, everybody will have access to the graph and it can be temporarily mutated only on purpose. I believe that it would be better to expose the database with oauth-proxy with specific permissions. 

@hongkailiu Can we configure an oauth-proxy to authenticate with SSO?

clusters/app.ci/dgraph/admin_dgraph-sts.yaml
```diff
@@ -0,0 +1,106 @@
...
+      matchLabels:
+        app: dgraph
+  template:
+    metadata:
+      labels:
+        app: dgraph
+    spec:
+      containers:
+      - name: ratel
+        image: dgraph/ratel:latest
```
danilo-gemoli, line 47: Should we use one of [those tags](https://hub.docker.com/r/dgraph/ratel/tags) rather than `latest`?

droslean: This is ratel. We shouldn't care about the version. `latest` will be the best option. To be honest we don't even need it for now. We could remove it completely. WDYT?

danilo-gemoli: Is there an alternative user-friendly way to query the db and visualize the resulting graph other than using Ratel? It's a long way to go before having our own UI ready, if that's what you mean.

PR #35852, opened 2023-01-30
introduce the image-graph-generator periodic job

droslean:
/hold
needs https://github.com/openshift/release/pull/35834

/cc @openshift/test-platform @danilo-gemoli 


danilo-gemoli:
I'm not sure "daily" would be the right frequency but we can eventually change it along the way.
/lgtm

PR #35877, opened 2023-01-31
drop codeready-qe-operator addon periodic

jbpratt:
this addon has been deprecated, cleaning up the job

https://redhat-internal.slack.com/archives/CMK13BP4J/p1675172819521009?thread_ts=1675087265.823329&cid=CMK13BP4J


MrSantamaria:
@jbpratt This wasn't a pipeline owned by me but the addon team. I'm aware of the discussion on slack but could you track someone from that team to ack this PR.

jbpratt:
> @jbpratt This wasn't a pipeline owned by me but the addon team. I'm aware of the discussion on slack but could you track someone from that team to ack this PR.

@MrSantamaria technically we do own this pipeline, none of those people are going to be owners of these pipelines according to tide, which is a bit confusing... I think we need to be more proactive about keeping this stuff up to date or removing it if it isn't used. It is just cruft piling up that looks bad when we try to show people using osde2e to test addons and there are 14/15 failures. If we don't own it, I think it should be moved to a different directory.

MrSantamaria:
> > @jbpratt This wasn't a pipeline owned by me but the addon team. I'm aware of the discussion on slack but could you track someone from that team to ack this PR.
> 
> @MrSantamaria technically we do own this pipeline, none of those people are going to be owners of these pipelines according to tide, which is a bit confusing... I think we need to be more proactive about keeping this stuff up to date or removing it if it isn't used. It is just cruft piling up that looks bad when we try to show people using osde2e to test addons and there are 14/15 failures. If we don't own it, I think it should be moved to a different directory.

The history for the addon pipelines living under the osde2e dir was for the team to provide their secrets and be able to keep track of where our OCM token is being used. Addon pipelines are CCS by nature so the team's cloud accounts are not being impacted. 
https://github.com/openshift/osde2e-example-test-harness#configuring-osde2e

The team should have set up the report_config to be alerted of their pipeline health.

jbpratt:
> > > @jbpratt This wasn't a pipeline owned by me but the addon team. I'm aware of the discussion on slack but could you track someone from that team to ack this PR.
> > 
> > 
> > @MrSantamaria technically we do own this pipeline, none of those people are going to be owners of these pipelines according to tide, which is a bit confusing... I think we need to be more proactive about keeping this stuff up to date or removing it if it isn't used. It is just cruft piling up that looks bad when we try to show people using osde2e to test addons and there are 14/15 failures. If we don't own it, I think it should be moved to a different directory.
> 
> The history for the addon pipelines living under the osde2e dir was for the team to provide their secrets and be able to keep track of where our OCM token is being used. Addon pipelines are CCS by nature so the SDCICD team's cloud accounts are not being impacted. https://github.com/openshift/osde2e-example-test-harness#configuring-osde2e
> 
> The team should have set up the report_config to be alerted of their pipeline health.

Why do we require to use our OCM token but not our AWS account? Can't we just push teams to provide their own OCM token and we don't have to have ownership over anything? We know that they don't actually keep up with these pipelines though.

PR #35884, opened 2023-01-31
MCO: don't run these jobs on each PR

cheesesashimi:
These jobs were intended primarily to be used for early testing for the RHCOS 9 changeover. They don't provide much additional signal for most MCO PRs and should only be run on-demand.


cheesesashimi:
These jobs did not materially change, so I will skip rehearsals.

/pj-rehearse skip

yuqi-zhang:
/lgtm
makes sense

PR #35766, opened 2023-01-27
Removing netobserv

jotak:
Netobserv upstream has moved to full github actions based CI.

I checked all references to the project to remove them.

Signed-off-by: Joel Takvorian <jtakvori@redhat.com>

PR #35902, opened 2023-01-31
Reintroduce netobserv just for core-services

jotak:


jotak:
I recently removed bots integration for netobserv (https://github.com/openshift/release/pull/35766) for that was probably a hasty move.
We'd actually like to reintroduce some of the prow/tide  features, but not all ; in particular, no CI builds and testing

droslean (approved):

PR #35916, opened 2023-02-01
USHIFT-821: Fix post chains in microshift workflows

pacevedom:
Reorder post actions to delete resources from gcp in reverse order from creation. Fixes [USHIFT-821](https://issues.redhat.com//browse/USHIFT-821).


pacevedom:
Rehearse failing for a different reason than the one addressed in this commit.

PR #35926, opened 2023-02-01
ci-operator/config/cri-o: disable rhel-integration test

sohankunkerkar:
we need to figure out the seccomp-related issue which is affecting the rhel variant.


haircommander:
/hold

we can't not run integration tests. We probably should just skip those tests in cri-o

PR #35932, opened 2023-02-01
Filestore CSI driver: test mount options and multiplePVsSameID capability

bertinatto:
CC @openshift/storage


jsafrane:
I can see tests with mount options running, sometimes even passing :-)

> External Storage [Driver: filestore.csi.storage.gke.io] [Testpattern: Dynamic PV (default fs)] provisioning should provision storage with mount options

Most of the test failures are because of quota, that's... concerning;

> ProvisioningFailed: failed to provision volume with StorageClass "e2e-provisioning-9782-e2e-scwkxnd": rpc error: code = Internal desc = googleapi: Error 429: Quota limit 'StandardStorageGbPerRegion' has been exceeded. Limit: 20000 in region us-central1.

PR #35942, opened 2023-02-01
MCO: don't run jobs on all PRs

cheesesashimi:
I thought I had fixed this with https://github.com/openshift/release/pull/35884. However, it seems that the true arbiter is the presence of the regex in the `skip_if_only_changed` field. Confusingly, if you set that field to null in the `ci-operator/config` file, the tooling will remove that and not propagate it to the `ci-operator/jobs` file, meaning you have to manually edit the `ci-operator/jobs` file.

¯\_(ツ)_/¯


yuqi-zhang (approved):
That, hmm, sounds really weird in terms of priority ordering of fields, but ¯\\_(ツ)_/¯

/lgtm

PR #35951, opened 2023-02-02
STOR-950: Add CI job for skipping StorageClass creation

dobsonj:
https://issues.redhat.com/browse/STOR-950
/cc @openshift/storage


jsafrane:
I would prefer go tests in https://github.com/openshift/origin/tree/master/test/extended

dobsonj:
/close
see https://github.com/openshift/origin/pull/27704 instead

PR #35980, opened 2023-02-02
Replace current build-root image

viroel:
This patch replaces the current build-image to use a custom Dockerfile which can generate different build root images, from different dependencies.
The previous Dockerfile for build root will be deleted later. We need to maintain pre-commit image for now, since a few repos are consuming it.


viroel:
/pj-rehearse pull-ci-openstack-k8s-operators-openstack-k8s-operators-ci-master-images

PR #35781, opened 2023-01-27
[WINC-921]: add fast-forward workflow 

alinaryan:
This PR adds a windows fast-forward workflow and changes the config to use the new workflow.

PR #35987, opened 2023-02-02
[WINC-921]: Follow-up add fast-forward workflow

alinaryan:
Removes the 'git log' line and sets the upstream branch as origin in the case of fast-forwarding to a new branch.

Follow-up to #35781

### DIFF
--- a/ci-operator/step-registry/windows/ci/fastforward/windows-ci-fastforward-ref.yaml
+++ b/ci-operator/step-registry/windows/ci/fastforward/windows-ci-fastforward-ref.yaml
@@ -1,6 +1,6 @@
 ref:
   as: windows-ci-fastforward
-  from: cli
+  from: tools
   commands: windows-ci-fastforward-commands.sh
   resources:
     requests:


ci-operator/step-registry/windows/ci/fastforward/windows-ci-fastforward-commands.sh
```diff
@@ -61,9 +61,8 @@ if ! git clone -b "$DESTINATION_BRANCH" "$repo_url" ; then
         exit 1
     fi
 
-    echo "INFO Pushing the following commits to $DESTINATION_BRANCH to origin"
-    git --no-pager log --pretty=oneline origin/"$DESTINATION_BRANCH"..HEAD
-    if ! git push origin $DESTINATION_BRANCH; then
+    echo "INFO Pushing to new branch $DESTINATION_BRANCH"
+    if ! git push --set-upstream origin "$DESTINATION_BRANCH" ; then
```
aravindhp, line 65: Please mention why you are making this change in your commit message.

I also don't understand why you need the `--set-upstream`. Why is that needed given it is an ephemeral config change?

aravindhp (changes requested):
Thanks for the follow up, @alinaryan

ci-operator/step-registry/windows/ci/fastforward/windows-ci-fastforward-commands.sh
alinaryan: Updated the commit message and removed the `--set-upstream` given what you said. I was able to run this locally with no errors.

saifshaikh48 (commented):
/lgtm

saifshaikh48:
/remove-lgtm

for [`tools` instead of `cli`](https://github.com/alinaryan/release/blob/master/ci-operator/step-registry/windows/ci/fastforward/windows-ci-fastforward-ref.yaml#L3)

aravindhp (approved):
Thanks for fixing this up, @alinaryan 

/hold

for other reviews.

mtnbikenc:
LGTM

PR #35997, opened 2023-02-03
Fix makefile compatibility with podman for mac

cblecker:
https://github.com/containers/podman/issues/13631#issuecomment-1077643246

We don't want/need to set the SELinux mount label when running on podman for macOS. Also, when we pull images, make sure we pull them for linux/amd64

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,9 +5,14 @@ SHELL=/usr/bin/env bash -o errexit
 export CONTAINER_ENGINE ?= docker
 export SKIP_PULL ?= false
 
+VOLUME_MOUNT_FLAGS = :z
 ifeq ($(CONTAINER_ENGINE), docker)
 	USER=--user $(shell id -u):$(shell id -g)
 else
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -5,9 +5,14 @@ SHELL=/usr/bin/env bash -o errexit
 export CONTAINER_ENGINE ?= docker
 export SKIP_PULL ?= false
 
+VOLUME_MOUNT_FLAGS = :z
 ifeq ($(CONTAINER_ENGINE), docker)
 	USER=--user $(shell id -u):$(shell id -g)
 else
+	ifeq ($(shell uname -s),Darwin)
```
danilo-gemoli, line 12: There is a missing space between the comma and `Darwin`

danilo-gemoli:
I would create two commits in this PR: one for `--platform linux/amd64` and the other for `:z`

cblecker:
@danilo-gemoli Updated per your suggestions

PR #35952, opened 2023-02-02
Add/update AWS Local Zone workflow

yunjiang29:
**Added new workflow to cover more scenarios**
* `cucushift-installer-rehearse-aws-ipi-network-mtu-localzone-sdn`
* `cucushift-installer-rehearse-aws-ipi-disconnected-private-network-mtu-localzone`

**New Local Zone profiles**
* 4.12 x86 stable & nightly
  * aws-ipi-disconnected-network-mtu-localzone-p2-f14
  * aws-ipi-disconnected-network-mtu-localzone-p3-f28-destructive
  * aws-ipi-network-mtu-localzone-sdn-p2-f7
  * aws-ipi-network-mtu-localzone-sdn-p3-f28-destructive

/hold

PR #36007, opened 2023-02-03
Update AWS Local Zone jobs

yunjiang29:
ARM instances are not supported on AWS Local Zone, these jobs were added incorrectly, should be removed.
more details see https://github.com/openshift/release/pull/35952#issuecomment-1413756503

## changes
* Removed all ARM on LZ profiles
* Decrease frequency of all LZ f7 profiles to f14
* Added 4.13 and upgrade profiles, now LZ profiles cover:
  * 4.12 OVN/SDN/Disconnected
  * 4.13 OVN/SDN/Disconnected
  * 4.12 stable -> 4.12 nightly: OVN/SDN/Disconnected
  * 4.12 stable -> 4.13 nightly: OVN/SDN/Disconnected


yunjiang29:
@jianlinliu @liangxia PTAL

jianlinliu:
How about seting all the remaining localzone jobs to `f14` from `f7`? BTW, we also need them in 4.13 testing, right?

PR #36018, opened 2023-02-03
Fix subscription channel to beta

git-hyagi:
We are not using alpha channel anymore, which is causing olm to fail to find the installplan:
```
Failed to find installPlan for subscription
```
https://github.com/pulp/pulp-operator/blob/61b8b940811104ef62d392d6880fc8e6036508c1/bundle/metadata/annotations.yaml#L7-L8


ipanova (approved):

git-hyagi:
unset GOFLAGS

ipanova (approved):

PR #36029, opened 2023-02-03
SPLAT-940: rely on non-zonal ova import to import ova to the IBMCloud datacenter

rvanderp3:
Non-zonal ova import is already importing the OVAs needed in the IBMCloud datacenter.  The change here is to only sync updated OVAs withe datacenter-2.


rvanderp3:
/hold
placing hold while we test this

rvanderp3:
cc: @vr4manta

PR #36051, opened 2023-02-04
Use build_root image from repo

Sandeepyadav93:
This patch changes the build_root image to consume from mariadb-operator repo. Other images will also be generated from build_root, so we can have a single place to swap golang and sdk versions.

Depends-On: https://github.com/openstack-k8s-operators/mariadb-operator/pull/64


abays (commented):
It seems fine to me, but I'll admit I don't really know much about this

abays (approved):
/lgtm

PR #36053, opened 2023-02-04
Set cron expression to very large value so that it would not execute

vaibhavjainwiz:
RHOSE(smart-event) initiative has been merged with RHOC so now we don't need a separate release pipeline for RHOSE. Rather than removing it, I am disabling this JOB in case we need to refer to it in future.


MrSantamaria:
Why not just remove the pipeline? 
If the goal is to not have this execute until it is needed for the operator to be promoted we can re-add the pipeline at that time.

vaibhavjainwiz:
> Why not just remove the pipeline? If the goal is to not have this execute until it is needed for the operator to be promoted we can re-add the pipeline at that time.

There  are many configuration from multiple places that we need to remove to deboared CPaaS/MSP pipeline. We are just waiting for management green light from it.
Once we got the response I would remove all the configuration including this one.

@MrSantamaria Currently PR checks are failing due to invalid cron expression, Could you please suggest a cron expression to disable its execution.

MrSantamaria:
Hmm lets try: 
0 0 0 12 31 2023
//EOY

Thought I would still suggest that if the pipeline isn't being used to trigger other jobs or populate the dashboard the best course is to remove it completely. 
We can always use this thread as a historic point to answer the "why" it was removed.

PR #36067, opened 2023-02-06
NP-603: ushift change default MTU to 1500

pacevedom:


pacevedom:
/pj-rehearse pull-ci-openshift-microshift-main-e2e-openshift-conformance-sig-api-machinery

zshi-redhat:
LGTM

zshi-redhat:
@pacevedom The file `upi-gcp-rhel8-pre-commands.sh` is exclusively used by microshift, right?
Just in case, I don't see microshift name in that script.

pacevedom:
> @pacevedom The file `upi-gcp-rhel8-pre-commands.sh` is exclusively used by microshift, right? Just in case, I don't see microshift name in that script.

Yes, only MicroShift uses this workflow. Actually this workflow was created for it.

PR #36102, opened 2023-02-06
bootc: Trim plugins

cgwalters:
We don't use lgtm or approve.  I think it's those that are the cause of
https://github.com/containers/bootc/issues/57
While we're here trim away other plugins we aren't using, like cherrypick and needs-rebase.


cgwalters:
@baude mind doing approve/lgtm for this and https://github.com/openshift/release/pull/36104 ?

PR #36107, opened 2023-02-07
stackrox: run openshift e2e qa when operator changes

gavin-stackrox:
In part because these tests now install using the operator but also to get more OpenShift coverage.
/cc @porridge @kurlov @connorgorman

### DIFF
--- a/ci-operator/jobs/stackrox/stackrox/stackrox-stackrox-master-presubmits.yaml
+++ b/ci-operator/jobs/stackrox/stackrox/stackrox-stackrox-master-presubmits.yaml
@@ -1216,7 +1216,7 @@ presubmits:
     name: pull-ci-stackrox-stackrox-master-openshift-newest-qa-e2e-tests
     optional: true
     rerun_command: /test openshift-newest-qa-e2e-tests
-    run_if_changed: ^(image|operator|pkg)/.*
+    run_if_changed: ^((generated|image|operator|pkg|make|deploy|scripts)/.*|Makefile|status\.sh|EXPECTED_GO_VERSION)
     spec:
       containers:
       - args:
@@ -1348,7 +1348,7 @@ presubmits:
     name: pull-ci-stackrox-stackrox-master-openshift-oldest-qa-e2e-tests
     optional: true
     rerun_command: /test openshift-oldest-qa-e2e-tests
-    run_if_changed: ^(image|operator|pkg)/.*
+    run_if_changed: ^((generated|image|operator|pkg|make|deploy|scripts)/.*|Makefile|status\.sh|EXPECTED_GO_VERSION)
     spec:
       containers:
       - args:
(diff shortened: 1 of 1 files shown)


porridge (commented):
I'm assuming that (`optional: true` + missing `run_if_changed`) means "don't run unless explicitly requested", right?
Why not use the same expression as for `openshift-*-operator-e2e-tests` jobs? I think it covers some important cases, while not increasing the footprint massively. And using the same expression will be good for consistency in the future...

gavin-stackrox:
> I'm assuming that (`optional: true` + missing `run_if_changed`) means "don't run unless explicitly requested", right? 
It means "do run if there are changes in these paths, otherwise don't run unless explicitly requested". Which is what happens with operator tests today.

> Why not use the same expression as for `openshift-*-operator-e2e-tests` jobs? I think it covers some important cases, while not increasing the footprint massively. And using the same expression will be good for consistency in the future...
Good point. But I'm a bit wary about ramping up the cluster claims too much e.g. for [this PR](https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-stackrox-stackrox-master-openshift-newest-operator-e2e-tests) it should probably not be running operator e2e tests.

porridge:
> > I'm assuming that (`optional: true` + missing `run_if_changed`) means "don't run unless explicitly requested", right?
> 
> It means "do run if there are changes in these paths, otherwise don't run unless explicitly requested".

Hang on. Which paths? I'm asking about the case of unspecified `run_if_changed`.

> > Why not use the same expression as for `openshift-*-operator-e2e-tests` jobs? I think it covers some important cases, while not increasing the footprint massively. And using the same expression will be good for consistency in the future...
> 
> Good point. But I'm a bit wary about ramping up the cluster claims too much e.g. for [this PR](https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-stackrox-stackrox-master-openshift-newest-operator-e2e-tests) it should probably not be running operator e2e tests.

Um, you linked to a job history, not a PR :thinking:

gavin-stackrox:
> > > I'm assuming that (`optional: true` + missing `run_if_changed`) means "don't run unless explicitly requested", right?
> > 
> > 
> > It means "do run if there are changes in these paths, otherwise don't run unless explicitly requested".
> 
> Hang on. Which paths? I'm asking about the case of unspecified `run_if_changed`.

Ah. I missed the word "missing" :sob:. Your initial assumption is correct.

> > > Why not use the same expression as for `openshift-*-operator-e2e-tests` jobs? I think it covers some important cases, while not increasing the footprint massively. And using the same expression will be good for consistency in the future...
> > 
> > 
> > Good point. But I'm a bit wary about ramping up the cluster claims too much e.g. for [this PR](https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-stackrox-stackrox-master-openshift-newest-operator-e2e-tests) it should probably not be running operator e2e tests.
> 
> Um, you linked to a job history, not a PR 🤔

And now I cannot find it. It was a PR that changed a not very related to operator script file. Anyway, I will follow your suggestion to get this closed and get more OCP coverage. If cluster_claims start failing / take too long we can revisit.

porridge (approved):

PR #36122, opened 2023-02-07
[openshift-loki] Use file-based catalog instead of custom indices

periklis:
According to [docs](https://docs.ci.openshift.org/docs/how-tos/testing-operator-sdk-operators/#building-an-index-deprecated) the CI operator is going to remove support for building OLM indices. The following warning encourages to adopt file-based catalogs:
```
DEPRECATION WARNING: Building index images is deprecated and will be removed from ci-operator soon. See https://docs.ci.openshift.org/docs/how-tos/testing-operator-sdk-operators/#moving-to-file-based-catalog for details. 
```

/cc @xperimental

### DIFF
--- a/ci-operator/config/openshift/loki/openshift-loki-main.yaml
+++ b/ci-operator/config/openshift/loki/openshift-loki-main.yaml
@@ -1,6 +1,6 @@
 base_images:
   base:
-    name: "4.11"
+    name: "4.12"
     namespace: ocp
     tag: base
   lokistack-gateway:
@@ -62,7 +62,7 @@ releases:
   latest:
     release:
       channel: stable
-      version: "4.11"
+      version: "4.12"
 resources:
   '*':
     requests:
@@ -76,7 +76,7 @@ tests:
     owner: openshift-ci
     product: ocp
     timeout: 1h0m0s
-    version: "4.11"
+    version: "4.12"
   steps:
     dependencies:
       OO_INDEX: ci-index-loki-operator-bundle
@@ -90,6 +90,9 @@ tests:
       cli: latest
       commands: |
         oc -n openshift-operators-redhat rollout status deployment loki-operator-controller-manager
+      dependencies:
+      - env: OO_BUNDLE
+        name: loki-operator-bundle
       from: src
       resources:
         requests:
--- a/ci-operator/config/openshift/loki/openshift-loki-release-5.5.yaml
+++ b/ci-operator/config/openshift/loki/openshift-loki-release-5.5.yaml
@@ -1,6 +1,6 @@
 base_images:
   base:
-    name: "4.11"
+    name: "4.12"
     namespace: ocp
     tag: base
   loki:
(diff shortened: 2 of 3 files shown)


ci-operator/config/openshift/loki/openshift-loki-main.yaml
```diff
@@ -76,10 +76,10 @@ tests:
     owner: openshift-ci
     product: ocp
     timeout: 1h0m0s
-    version: "4.11"
+    version: "4.12"
   steps:
     dependencies:
-      OO_INDEX: ci-index-loki-operator-bundle
+      OO_BUNDLE: loki-operator-bundle
```
xperimental, line 82: This line is different between `main` and the two release branches. Is it maybe the cause of the failure or are the other two files missing this change?

ci-operator/config/openshift/loki/openshift-loki-main.yaml
```diff
@@ -90,6 +88,9 @@ tests:
       cli: latest
       commands: |
         oc -n openshift-operators-redhat rollout status deployment loki-operator-controller-manager
+      dependencies:
+      - env: OO_BUNDLE
+        name: loki-operator-bundle
```
xperimental, line 93: If I'm reading the docs correctly, then the dependency declaration should be one level up where it originally was, essentially just replacing `OO_INDEX` with `OO_BUNDLE` and using the bundle name directly:

```
dependencies:
  OO_BUNDLE: loki-operator-bundle
```

Similar for the other files.

periklis: Unfortunately it looks that the docs are incomplete. The makefile target `make update` gives me an error like the following if I leave the `OO_BUNDLE` definition only at the top:
```
time="2023-02-13T13:15:27Z" level=error error="failed to validate configuration /ci-operator/config/openshift/loki/openshift-loki-release-5.6.yaml: Failed resolve MultiStageTestConfiguration: test/test-operator: workflow/optional-operators-ci-generic-claim: dependency \"OO_BUNDLE\" is overridden in [test/test-operator] but not declared in any step"
```

periklis:
@xperimental I cannot fully understand the following error vs. what the docs say, but it seems that using `OO_INDEX` in the workflow dependencies is still mandatory:
```
 ERRO[2023-02-13T13:18:27Z] Some steps failed:                           
ERRO[2023-02-13T13:18:27Z] 
  * could not sort nodes
  * steps are missing dependencies
  * step test-operator is missing dependencies: <&api.internalImageStreamTagLink{name:"pipeline", tag:"ci-index", unsatisfiableError:""}> 
```

periklis:
@xperimental Although now we use both `OO_INDEX` and `OO_BUNDLE` we still get the warning:
```
WARN[2023-02-14T12:16:05Z] DEPRECATION WARNING: Building index images is deprecated and will be removed from ci-operator soon. See https://docs.ci.openshift.org/docs/how-tos/testing-operator-sdk-operators/#moving-to-file-based-catalog for details. 
```
seems this OO_BUNDLE feature is not working as expected.

PR #36130, opened 2023-02-07
Use go1.21 for Package Operator tests

eqrx:


eqrx:
well, looks like there is no go 1.20 support yet, leaving this PR open until that changes :)

PR #36145, opened 2023-02-07
OSDe2e Hypershift pipelines version rollback patch

MrSantamaria:
Sets the hypershift pipeline to force 4.12.0. 

This is a temporary patch for the pipelines to match the current rollback of 4.12 not being the new default in the stable channel. 

Slack Thread:
https://redhat-internal.slack.com/archives/CMK13BP4J/p1675715204345139


MrSantamaria:
@jbpratt We could use version 4.12.1 as well, let me know your preference.

MrSantamaria:
> > @jbpratt We could use version 4.12.1 as well, let me know your preference.
> 
> thanks for opening this! Lets bump to 4.12.1 if that is okay

Of course it is and done.

PR #35931, opened 2023-02-01
CONSOLE-3303: nodejs-16 and nginx image mirroring for console plugin template

cajieh:

PR #36149, opened 2023-02-07
CONSOLE-3303: Image mirroring for nodejs-16 and nginx:stable, and build image for console plugin template 

cajieh:


smg247:
I was poking around to see where the base image was coming from, and found: https://github.com/openshift/release/pull/35931. I know it was approved, but we can't mirror images from `docker.io`. Eventually, we end up being rate limited and jobs start to fail. You will have to get this image on quay somehow and then mirror it from there.

cyril-ui-developer:
/assign @jhadvig 
cc: @smg247

smg247:
Feel free to remove the hold whenever

PR #36175, opened 2023-02-08
Running the `release-controller-config` test for any `jobs/openshift` PRs

bradmwilliams:
We are, repeatedly, getting bitten by folks removing jobs, from their respective components, and not the corresponding definition from the release-controller's configuration files.  This PR opens up the trigger regex to any job configuration changes under: `ci-operator/jobs/openshift`.
It also adds some messaging about what these failures might mean if/when they occur, and how to address them.

### DIFF
--- a/hack/validate-release-jobs.py
+++ b/hack/validate-release-jobs.py
@@ -91,6 +91,16 @@ def main(git_repo_path):
     for source, verification, name in missing_jobs:
         logger.error('Unable to locate job definition for: %s:%s:%s', source, verification, name)
 
+    if len(missing_jobs) > 0:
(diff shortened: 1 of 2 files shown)


hack/validate-release-jobs.py
```diff
@@ -91,6 +91,16 @@ def main(git_repo_path):
     for source, verification, name in missing_jobs:
         logger.error('Unable to locate job definition for: %s:%s:%s', source, verification, name)
 
+    if len(missing_jobs) > 0:
```
bbguimaraes, line 94: A minor detail: this could be just `if missing_jobs:` in Python.

bbguimaraes (approved):

PR #36177, opened 2023-02-08
Fixing up 4.13 release configs

bradmwilliams:


bradmwilliams:
These changes address issues caused from: https://github.com/openshift/release/pull/35886

I have a pending PR that should catch this scenario in the future: https://github.com/openshift/release/pull/36175

PR #36179, opened 2023-02-08
test cloning a private repo from dockerfile custom image

ascerra:


ascerra:
Forgot to run make update to let private settings take effect...

PR #36207, opened 2023-02-09
Sort the configs/jobs and fix cron entries

liangxia:
Changes in the PR,
1. Sort the configs/jobs
2. Fix several cron entries


jianli-wei:
@liangxia May I ask why we need them sorted? Thanks!

liangxia:
> May I ask why we need them sorted? Thanks!

When comparing between different versions/architectures, sorted configs are more easier to diff.

jianli-wei (approved):

liangxia:
@jianli-wei E.g, it helps us find issue like this, https://github.com/openshift/release/pull/36208

PR #36219, opened 2023-02-09
RHOC: Add credentials secret mount

asmigala:


MrSantamaria:
Hmmm from the logs I think the harness you are using is the cause of the Ginkgo Leak. 
@ritmun have you seen this error with the addon harnesses?

asmigala:
@MrSantamaria there's definitely a bug in our harness, I'll get to fixing that next week.

MrSantamaria:
If you get too stuck you can reach us at #sd-cicd

PR #36226, opened 2023-02-09
OSSM maistra remove git workaround

yxun:
Signed-off-by: Yuanlin <yuanlin.xu@redhat.com>

This is PR follows  https://github.com/kubernetes/test-infra/pull/28688
The previous Prow uid and git issue has been fixed in upstream. So we can remove our workaround in the image build steps. Thanks,


jmguzik:
@yxun this issue is not yet fixed upstream.

PR #36230, opened 2023-02-09
disable vSphere UPI zonal installs in 4.12

rvanderp3:
UPI zonal installs are broken and not supported in 4.12.  There is a change required in the installer image that requires a backport to work with 4.12.  Thus, we'll disable this periodic.


rvanderp3:
cc: @jcpowermac

bradmwilliams (approved):

rvanderp3:
> /assign @stbenjam

Hi @stbenjam would you mind taking a look at this one when you get a chance?

PR #36240, opened 2023-02-10
Simply the coverage for amd/arm in one config file

liangxia:
When adding tests for new profile(s), teammates sometimes forgot to add tests for arm as we are using separate config files to store the tests.
This PR merge the `*-4.10-__amd64-nightly.yaml` and `*-4.10__arm64-nightly.yaml` into one file `*-4.10__nightly.yaml`, and sort the tests in alphanumeric order as below, it can help remind people adding test for the other arch. Also it help reduce the config files we have.
```
- as: alibaba-ipi-fips-p2-f14-amd64
- as: alibaba-ipi-fips-p3-f28-destructive-amd64
- as: aws-c2s-ipi-disconnected-private-p2-f14-amd64
- as: aws-c2s-ipi-disconnected-private-p3-f28-destructive-amd64
- as: aws-ipi-byo-iam-role-p2-f14-amd64
- as: aws-ipi-byo-iam-role-p2-f14-arm64
- as: aws-ipi-byo-iam-role-p3-f28-destructive-amd64
- as: aws-ipi-byo-iam-role-p3-f28-destructive-arm64
...
```
Currently this PR only use 4.10 nightly as an example to get opinions from the team. I will use follow up PR(s) for the other versions if got agreement on this.


liangxia:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.10-nightly-aws-ipi-ovn-p3-f28-amd64
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.10-nightly-aws-ipi-ovn-p3-f28-arm64

liangxia:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.10-nightly-aws-ipi-ovn-p3-f28-arm64

liangxia:
Both the installation for amd64 and arm64 succed.

/cc @aleskandro @jianlinliu @jhou1

aleskandro:
This seems what we had initially, but yet I'm not sure this will help get others to add configs for both the architectures. If this becomes the path to follow, we need to do a similar thing for the `multi` variant.

liangxia:
> This seems what we had initially, but yet I'm not sure this will help get others to add configs for both the architectures. If this becomes the path to follow, we need to do a similar thing for the `multi` variant.
> 
> Also, this means that everytime the tests run, we will import all the payloads: single-arch arm64, single-arch amd64, multi-arch (that will eventually need 4 x the time needed for a single-arch cluster when the support for manifests-lists in the registry and in the ci-operator land)

For `multi', that's a sure thing. Didn't add it there because this is a POC, also we do not have multi in 4.10 while the poc is on 4.10.
For the second concern, that's not the case. If you check the logs of the two jobs got triggered, amd64 only import image latest for amd, while arm64 import image latest and arm64-latest, both are the same as without the changes in this PR.

aleskandro:
Yeah, right. I was looking at it right now.

Ok, I would collect others' opinion on this. I don't have a strong preference, but still, I would like the team to ensure test configs are duplicated and halved in frequency for both the architectures until we can get better solutions.

jianlinliu:
I do not have strong preference, too. 

About "ensure add configs for both the architectures", I guess we always need to educate teammates to remember that. Just like a similar case, we always remind teammates to destroy clusters after testing, but till now, still often see some long-running clusters forgotten to destroy.

aleskandro:
shall we merge this one?

liangxia:
> shall we merge this one?

It's up on you ;)

liangxia:
//pj-rehearse ack

Let's try to get it move forward.

aleskandro:
> > shall we merge this one?
> 
> It's up on you ;)

If that's the case, I slightly prefer having the variant for each architecture. Mainly because it keeps the config files shorter (they are already very long) and sorted from my perspective...

PR #36246, opened 2023-02-10
gcp-pd-csi-driver: disable auto run of sanity tests

RomanBednar:
This test was also disabled upstream:
https://github.com/kubernetes/test-infra/pull/26354/files

The problem is that sanity tests currently use command faker and only one driver instance. That makes any command faking impossible as the same instance of the driver is re-used in many tests. Relevant discussion upstream is here:
https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/972#issuecomment-1130221537

cc @openshift/storage for approval

### DIFF
--- a/ci-operator/config/openshift/gcp-pd-csi-driver/openshift-gcp-pd-csi-driver-master.yaml
+++ b/ci-operator/config/openshift/gcp-pd-csi-driver/openshift-gcp-pd-csi-driver-master.yaml
@@ -39,7 +39,8 @@ resources:
       cpu: 100m
       memory: 200Mi
 tests:
-- as: test
+- always_run: false
+  as: test
   commands: |
     make test-sanity
   container:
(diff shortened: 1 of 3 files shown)


ci-operator/config/openshift/gcp-pd-csi-driver/openshift-gcp-pd-csi-driver-master.yaml
```diff
@@ -39,7 +39,8 @@ resources:
       cpu: 100m
       memory: 200Mi
 tests:
-- as: test
+- always_run: false
```
jsafrane, line 42: This needs to be updated in 4.13.yaml and 4.14.yaml too

RomanBednar: Done.

PR #34934, opened 2022-12-20
Add e2e tests to the odh-notebook-controller

maroroman:
This PR adds e2e tests to odh-notebook-controller. No changes should be necessary on the nbc side.

Resolves: https://github.com/opendatahub-io/kubeflow/issues/30

I wasn't able to test the e2e tests themselves, but the individual commands in this PR should be working.

/cc @LaVLaS @VaishnaviHire

PR #36250, opened 2023-02-10
Correct odh-notebook-controller config command path

maroroman:
Follow up for #34934 

Fixes the incorrect command path in odh-notebook-controller configs.

/cc @LaVLaS @VaishnaviHire


VaishnaviHire (approved):
/lgtm

PR #36251, opened 2023-02-10
Add verify-deps job to all CSI drivers managed by storage team

jsafrane:
We want to make sure that `vendor/` directory of our CSI driver forks is correct to make sure official offline builds pass. 

I added a new test step `go-verify-deps` that tests we have the vendor correct. We don't want to add `verify-vendor.sh` directly to our forks, because that would be a new `<carry>` patch.

@openshift/storage


bertinatto (commented):
> We want to make sure that vendor/ directory is correct to make sure official offline builds pass.

`go list -m all` will only list the imported modules, it doesn't verify vendor (in fact, it doesn't even need a vendor directory).

In order to verify the vendor directory we probably need to a `go mod vendor`, dumping the resulting directory in an alternative path (e.g., `/tmp/myvendor`) and comparing with the vendor located in the repo. github.com/openshift/build-machinery-go has some code to do that, I believe it's triggered with `make verify-deps`.

mpatlasov:
Change commit title: 's/Add verify-deps job to AWS EBS CSI driver/Add verify-deps job to all CSI drivers managed by storage team' ?

jsafrane:
@mpatlasov good catch, fixed

jsafrane:
/pj-rehearse ack
All rehearse failures actually found a bug in our repos.

bertinatto (commented):
/lgtm

droslean (approved):

PR #36296, opened 2023-02-13
remove unnecessary provison fips chain and use ENV var to set it

jianlinliu:


jianlinliu:
cc @jianli-wei @yunjiang29 @MayXuQQ for review

jianli-wei (approved):

PR #36317, opened 2023-02-13
[logging-view-plugin] Add configuration for release-5.6 branch

xperimental:
This is just the configuration for the branch in the `logging-view-plugin` repository itself. The configuration for `cluster-logging-operator` in 5.6 probably also needs a change because of the image promotion.

/cc @jcantrill

### DIFF
--- a/ci-operator/config/openshift/logging-view-plugin/openshift-logging-view-plugin-release-5.6.yaml
+++ b/ci-operator/config/openshift/logging-view-plugin/openshift-logging-view-plugin-release-5.6.yaml
@@ -1,0 +1,51 @@
+base_images:
+  nginx-120:
+    name: ubi8-nginx
+    namespace: ci
+    tag: "1.20"
+build_root:
+  image_stream_tag:
+    name: release
+    namespace: openshift
+    tag: rhel-8-release-golang-1.17-openshift-4.10
+images:
+- dockerfile_path: Dockerfile.cypress
+  to: plugin-test
+- dockerfile_path: Dockerfile.ci
+  from: nginx-120
+  to: logging-view-plugin
+promotion:
+  name: "5.6"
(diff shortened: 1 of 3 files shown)


xperimental:
PTAL @kyoto @jgbernalp

ci-operator/config/openshift/logging-view-plugin/openshift-logging-view-plugin-release-5.6.yaml
```diff
@@ -0,0 +1,51 @@
+base_images:
+  nginx-120:
+    name: ubi8-nginx
```
jgbernalp, line 3: As we are not using nginx anymore, maybe it will make sense to have a nodejs base image like: https://catalog.redhat.com/software/containers/ubi8/nodejs-16/615aee9fc739c0a4123a87e1, this will allow to simplify the `Docker.ci`

ci-operator/config/openshift/logging-view-plugin/openshift-logging-view-plugin-release-5.6.yaml
```diff
@@ -0,0 +1,51 @@
...
+    namespace: ci
+    tag: "1.20"
+build_root:
+  image_stream_tag:
+    name: release
+    namespace: openshift
+    tag: rhel-8-release-golang-1.17-openshift-4.10
```
jgbernalp, line 10: should this be golang-1.18?

xperimental:
@jgbernalp While I agree that the configuration seems a bit out-of-date, it does reflect how the sources were built in `main` up until the branchcut. The purpose of this PR is to add the configuration that existed to the new branch. I think we should do the updates for `main` first and then backport them to the release-branches where necessary.

PR #36328, opened 2023-02-13
Addition of new containers/libhvee

baude:
containers/libhvee is a library for working with the HyperV hypervisor in support of Podman and ignition.

Signed-off-by: Brent Baude <bbaude@redhat.com>


baude:
could someone review and lgtm ?

PR #36353, opened 2023-02-14
[WIP] Use multi-stage tests for `ci-tools`

bbguimaraes:
For testing only.


jmguzik (approved):
lgtm but let us see if danilo's fix will deliver it as well
/hold

PR #36364, opened 2023-02-14
updating correct file for e2e-console update

memodi:
https://github.com/openshift/release/pull/36315 updated wrong file which will be overwritten by bot, this is for correct file and follow up for https://github.com/openshift/openshift-tests-private/pull/7474


kuiwang02:
this PR does not change the regression test and it changes the PR submit presubmit test.
so, I can approve it.

/lgtm
/approve

PR #36375, opened 2023-02-14
Unify approve/lgtm experience in OTA repos

petr-muller:
Treat GitHub approval as `/lgtm` everywhere. Otherwise I'll *always* need to double tap PRs like https://github.com/openshift/cincinnati-graph-data/pull/3111#issuecomment-1430215775 :grimacing: 

/cc @wking @LalatenduMohanty @PratikMahajan

### DIFF
--- a/core-services/prow/02_config/openshift/cincinnati-operator/_pluginconfig.yaml
+++ b/core-services/prow/02_config/openshift/cincinnati-operator/_pluginconfig.yaml
@@ -1,8 +1,3 @@
-approve:
-- commandHelpLink: https://go.k8s.io/bot-commands
-  repos:
-  - openshift/cincinnati-operator
-  require_self_approval: false
 lgtm:
 - repos:
   - openshift/cincinnati-operator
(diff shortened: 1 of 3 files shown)


core-services/prow/02_config/openshift/cincinnati-operator/_pluginconfig.yaml
```diff
@@ -1,8 +1,3 @@
-approve:
-- commandHelpLink: https://go.k8s.io/bot-commands
-  repos:
-  - openshift/cincinnati-operator
-  require_self_approval: false
```
petr-muller, line 5: This is default so we may as well remove it

PratikMahajan (approved):
/lgtm

petr-muller:
@LalatenduMohanty I assume https://github.com/openshift/cincinnati-graph-data/pull/3111#issuecomment-1430229161 was supposed to go here?

> What happens when PRs get auto approved tag? It is still need approved tag from an approver?

No, such PRs only need `lgtm` (by anyone in the org). It is assumed that approvers will only file PRs with desirable changes. But I think that's secondary in OpenShift. The concept of `approver` exists mainly for upstream to have a decision point about whether a certain change is desirable or not (the implementation quality is separate thing enforced by the `lgtm` label).

PR #36392, opened 2023-02-15
hypershift: @critical

LiangquanLi930:


ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.13__amd64-nightly.yaml
```diff
@@ -581,7 +581,7 @@ tests:
     env:
       BASE_DOMAIN: qe.devcluster.openshift.com
       E2E_RUN_TAGS: '@aws-ipi and @network-ovnkubernetes and @hypershift-hosted and
-        not @fips'
+        and @critical not @fips'
```
liangxia, line 584: Syntax error. 
`and and @critical not @fips`
->
`and @critical and not @fips`

PR #36415, opened 2023-02-15
`release-controller-api` add the jira volumeMount

hoxhaeris:


hack/generators/release-controllers/content/osd_rc_deployments.py
```diff
@@ -362,7 +362,7 @@ def _add_osd_rc_deployment(gendoc):
                                         ],
                             'image': 'release-controller-api:latest',
                             'name': 'controller',
-                            'volumeMounts': get_kubeconfig_volume_mounts(),
+                            'volumeMounts': get_rcapi_volume_mounts(),
```
bradmwilliams, line 365: So, you can't have `volumeMounts` without a corresponding `volumes` definition...

bradmwilliams (changes requested):

hack/generators/release-controllers/content/osd_rc_deployments.py
bradmwilliams: This is why the validation test failed ^^

hoxhaeris: Fixed

bradmwilliams (approved):

PR #36428, opened 2023-02-15
[CVP-3566] Add cluster pools for OCP v4.13

asergienk:


swghosh:
@asergienk Could you try using something similar to:

https://github.com/openshift/release/blob/354bacdd7e7cd22da38a3a1fea7c9865d7a37a23/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.13__amd64-dev-preview.yaml#L32-L35

And, check if it passes the test. Once, 4-stable rc builds are out for 4.13, the cluster pool could switch back again to “4-stable” stream.

PR #36476, opened 2023-02-16
MCO-423: Builds custom OS image for test

cheesesashimi:
This builds a custom OS image which the `TestOSImageURLOverride` uses to verify that the OS layering support within the MCO works as intended.

### DIFF
--- a/ci-operator/config/openshift/machine-config-operator/openshift-machine-config-operator-release-4.13.yaml
+++ b/ci-operator/config/openshift/machine-config-operator/openshift-machine-config-operator-release-4.13.yaml
@@ -35,6 +35,10 @@ base_images:
     name: builder
     namespace: ocp
     tag: rhel-8-golang-1.19-openshift-4.13
+  ocp_os:
+    name: "4.13"
+    namespace: ocp
+    tag: rhel-coreos-9
 build_root:
   from_repository: true
 images:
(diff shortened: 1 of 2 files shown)


cheesesashimi:
/pj-rehearse pull-ci-openshift-machine-config-operator-master-e2e-gcp-op

cheesesashimi:
Although it says that this configuration change touched all of the jobs, the reality is that it mostly affected the image build job and only tangentially affected the `e2e-gcp-op` / `e2e-gcp-op-single-node` jobs because they now have a dependency upon the `custom-os-image` that is produced by this config.

Until https://github.com/openshift/machine-config-operator/pull/3558 lands, this change will have no effect upon either of those jobs. And once it does, it will only directly impact `e2e-gcp-op` which is configured to look for the `OS_IMAGE_URL` env var that this configuration change enables. Also, it does not matter in which order these PRs land since if https://github.com/openshift/machine-config-operator/pull/3558 cannot locate the `OS_IMAGE_URL` env var, it will skip the test.

ci-operator/config/openshift/machine-config-operator/openshift-machine-config-operator-master.yaml
```diff
@@ -35,6 +35,10 @@ base_images:
     name: builder
     namespace: ocp
     tag: rhel-8-golang-1.19-openshift-4.13
+  ocp_os:
+    name: "4.13"
+    namespace: ocp
+    tag: rhel-coreos-8
```
sinnykumari, line 41: should this tag be rhel-coreos-9 ? asking this because we are later on in Dockerfile we are referencing rhel-coreos-9 tag while pulling osimage from 4.13

sinnykumari:
We will need to add this job config to 4.13 and 4.14 branch as well. This will make sure that it runs when 4.13 and 4.14 is no longer points to master. 

Curious, how we will consume this custom image in ci test?

cheesesashimi:
For consuming this in an e2e test, have a look at https://github.com/openshift/machine-config-operator/pull/3558. Basically, we look for the `OS_IMAGE_URL` env var and use that value to create a new MachineConfig which overrides the `osImageURL` field. If the env var is missing, that test will be skipped.

ci-operator/config/openshift/machine-config-operator/openshift-machine-config-operator-master.yaml
cheesesashimi: That's a good point, I'll fix that.

sinnykumari:
looks good. Thanks Zack
/lgtm

PR #36491, opened 2023-02-17
Add flags in openstack-operator pre-commit job

amolkahat:
Added flags in the pre-commit job.


abays:
The CI error seems to be that an older verison of Golang is being used:

```
/go/pkg/mod/sigs.k8s.io/json@v0.0.0-20220713155537-f223a00ba0e2/internal/golang/encoding/json/decode.go:1143:45: predeclared any requires go1.18 or later (-lang was set to go1.16; check go.mod)
```

amolkahat:
Yeah. so actually we were running it using `pre-commit-golang` so it was throwing this error.. and also was not working locally. So I've changed all things to work with make. Added pre-commit hooks to work with make[1].

[1] https://github.com/openstack-k8s-operators/openstack-operator/pull/204

amolkahat:
Not needed. Closing pr.

PR #36522, opened 2023-02-18
Promoting clc-ui-e2e test image

ascerra:
[LPTOCPCI-58](https://issues.redhat.com/browse/LPTOCPCI-58)

This PR is meant to unblock the ACM interop scenario onboarding into OpenShift CI.

There are Dockerfiles that we would like to build images from within OCP CI that exist in multiple test repos (all which are private). In order to make use of them from a single config we can promote these images then use them as base_images from within our config file.

Similar to what is already being done with the [application-ui-tests image](https://github.com/openshift/release/blob/15a6829f2293efc5b9064b9d9a74866ee539d34e/ci-operator/config/stolostron/application-ui/stolostron-application-ui-release-2.5.yaml#L35)

### DIFF
--- a/ci-operator/config/stolostron/clc-ui-e2e/.config.prowgen
+++ b/ci-operator/config/stolostron/clc-ui-e2e/.config.prowgen
@@ -1,0 +1,2 @@
+private: true
+expose: true
--- a/ci-operator/config/stolostron/clc-ui-e2e/OWNERS
+++ b/ci-operator/config/stolostron/clc-ui-e2e/OWNERS
@@ -1,0 +1,9 @@
+approvers:
+- ascerra
+- dtthuynh
+- vboulos
+options: {}
+reviewers:
+- ascerra
+- dtthuynh
+- vboulos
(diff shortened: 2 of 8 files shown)


ci-operator/config/stolostron/clc-ui-e2e/stolostron-clc-ui-e2e-main.yaml
```diff
@@ -0,0 +1,25 @@
...
+    name: release
+    namespace: openshift
+    tag: golang-1.17
+images:
+- dockerfile_path: Dockerfile.interops
+  to: clc-ui-e2e
+promotion:
+  name: "1.0"
+  namespace: acm-qe
+resources:
```
calebevans, line 17: Just curious, is the resources block required if you are just promoting an image?

ascerra: Just checked and it still is required. This tells me that this value must be used for the actually src-build or custom build pods.

calebevans: Sounds good, thanks for clarifying.

ci-operator/config/stolostron/clc-ui-e2e/stolostron-clc-ui-e2e-main.yaml
```diff
@@ -0,0 +1,25 @@
+base_images:
```
ascerra, line 1: Update filename to target the branch release-2.7

vboulos (approved):

PR #36547, opened 2023-02-20
Serverless: Kitchensink upgrade tests and fixes

mgencur:
* Add back Slack notifications for all periodic jobs which were removed by accident
* Add periodic job for new kitchensink upgrade tests


pierDipi (approved):
/lgtm

PR #36573, opened 2023-02-20
MGMT-12442: Use multi-arch assisted images

adriengentil:
Use the multi-arch installer, controller and agent available on quay.

### DIFF
--- a/ci-operator/config/openshift/assisted-test-infra/openshift-assisted-test-infra-master.yaml
+++ b/ci-operator/config/openshift/assisted-test-infra/openshift-assisted-test-infra-master.yaml
@@ -698,6 +698,10 @@ tests:
       OPENSHIFT_INSTALL_RELEASE_IMAGE: release:latest-multi
     env:
       ANSIBLE_PLAYBOOK_CREATE_INFRA: equinix_heterogeneous_create_infra_playbook.yml
+      ASSISTED_CONFIG: |
+        AGENT_DOCKER_IMAGE=quay.io/edge-infrastructure/assisted-installer-agent-multi:latest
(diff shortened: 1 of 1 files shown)


ci-operator/config/openshift/assisted-test-infra/openshift-assisted-test-infra-master.yaml
```diff
@@ -698,6 +698,10 @@ tests:
       OPENSHIFT_INSTALL_RELEASE_IMAGE: release:latest-multi
     env:
       ANSIBLE_PLAYBOOK_CREATE_INFRA: equinix_heterogeneous_create_infra_playbook.yml
+      ASSISTED_CONFIG: |
+        AGENT_DOCKER_IMAGE=quay.io/edge-infrastructure/assisted-installer-agent-multi:latest
```
osherdp, line 702: Actually, I think we can build the prow images multi-arch in the first place by reverting this:
https://github.com/openshift/release/pull/34041

It should be better for performance and for having the jobs more portable to other branches

osherdp (changes requested):

ci-operator/config/openshift/assisted-test-infra/openshift-assisted-test-infra-master.yaml
adriengentil: I think the problem was that the registry is not multi-arch. We have one URI for x86 and one for arm, which is not what I want here.

osherdp: I see
Then looks good to me

osherdp (approved):

PR #36578, opened 2023-02-20
SDCICD-846 - Add Rosa nightly HCP creation Pipeline

MrSantamaria:
https://issues.redhat.com/browse/SDE-2401

This adds a pipeline that installs hypershift with the latest rosa nightly.


ryankwilliams:
@MrSantamaria https://github.com/openshift/osde2e/pull/1655 would resolve the error the job is facing when trying to create the rosa cluster.

PR #36586, opened 2023-02-21
QE 4.13 Zonal GA

sgaoshang:


ci-operator/step-registry/cucushift/installer/rehearse/vsphere/ipi/zones/cucushift-installer-rehearse-vsphere-ipi-zones-chain.yaml
```diff
@@ -0,0 +1,14 @@
+chain:
+  as: cucushift-installer-rehearse-vsphere-ipi-zones
```
jianlinliu, line 2: I think we do not have to introduce this chain, but copy the following plain ref step to provision chain.

sgaoshang: Fixed

PR #36609, opened 2023-02-21
kubevirt: ssp: Set timeout for the whole test instead of a step

akrejcir:
The timeout for step was not used.


akrejcir:
No need to run rehearse jobs.

/pj-rehearse ack

PR #36632, opened 2023-02-22
USHIFT-905: Remove rpm install jobs

pacevedom:
RPM install is already checked in other jobs, plus this one is testing what is included in others too. Removing it as it is redundant.


pmtk:
Looks like we can also remove step/chain/workflow?

pacevedom:
> Looks like we can also remove step/chain/workflow?

Yes, but there is some automation around being allowed to do that. There is the openshift-priv directory which is entirely handled by scripts. Once this PR makes it through that dir will be synced and all references to the workflow removed, then we can delete it.

PR #36645, opened 2023-02-22
Run hypershift external e2e from 4.13

enxebre:
Autoscaler test is failing https://github.com/openshift/kubernetes-autoscaler/pull/250
NodePool is not being scaled out because workload pods are not running because of
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_kubernetes-autoscaler/250/pull-ci-openshift-kubernetes-autoscaler-master-e2e-hypershift/1628126056965738496/artifacts/e2e-hypershift/run-e2e/artifacts/TestAutoscaling_PreTeardownClusterDump/hostedcluster-example-52fxn/namespaces/default/core/events/autoscaling-workload.1745f27b0868a2e9.yaml

The e2e are running from 4.12 which does not have https://github.com/openshift/hypershift/pull/2097/files which I assume is causing the failure above.


elmiko (commented):
/lgtm

PR #36660, opened 2023-02-22
operator-pipelines/preflight-common: add artifact for debug

EmilienM:
When the step fails, it'll be nice to see what operator fails to be
ready.


EmilienM:
/close
in fact we don't need this, we already have it via `gather-extra/artifacts/clusteroperators.json`.

PR #36662, opened 2023-02-22
openstack-gather: don't skip all if no bastion is used

EmilienM:
For NFV jobs, we use a proxy to reach the cluster but we don't have a
bastion node. So instead of skipping the whole step, let's just skip
what is not supported today (`openshift-install gather`) so we can still
collect useful informations.

### DIFF
--- a/ci-operator/step-registry/openstack/gather/openstack-gather-commands.sh
+++ b/ci-operator/step-registry/openstack/gather/openstack-gather-commands.sh
@@ -19,11 +19,6 @@ fi
 if [[ "$CONFIG_TYPE" == "proxy" ]]; then
     BASTION_FIP=$(<"${SHARED_DIR}/BASTION_FIP")
     BASTION_USER=$(<"${SHARED_DIR}/BASTION_USER")
-else
-    if test -f "${SHARED_DIR}/proxy-conf.sh"; then
-        echo "This job uses a proxy but without a bastion, gathering is not supported yet, see CORS-2367"
-        exit 0
-    fi
 fi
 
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/openstack/gather/openstack-gather-commands.sh
```diff
@@ -65,32 +60,36 @@ collect_bootstrap_logs() {
...
+			    	SCP_CMD="scp $SSH_ARGS"
+			    	if ! $SSH_CMD uname -a; then
+			    		echo "ERROR: Bastion proxy is not reachable via $BASTION_FIP"
+			    		exit 1
+			    	fi
+			    	echo "Moving credentials and openshift binary to Bastion Proxy"
+			    	$SCP_CMD "$SSH_PRIV_KEY_PATH" /bin/openshift-install "${BASTION_USER}@${BASTION_FIP}:/tmp"
+			    	echo "Gathering bootstrap logs from Bastion Proxy"
+			    	$SSH_CMD bash - << EOF
+/tmp/openshift-install g    ather bootstrap --key /tmp/ssh-privatekey ${GATHER_BOOTSTRAP_ARGS[@]}
```
MaysaMacedo, line 84: typo

MaysaMacedo (changes requested):

ci-operator/step-registry/openstack/gather/openstack-gather-commands.sh
EmilienM: oh crap

PR #36679, opened 2023-02-23
[openshift-loki] Update bundle image context directory

periklis:
cc @xperimental


periklis:
Skipping rehearsals because of circular dependency with:
- openshift/loki#110

PR #36684, opened 2023-02-23
MGMT-12442: Enable e2e-metal-assisted-day2-arm-workers

adriengentil:
- Set e2e-metal-assisted-day2-arm-workers as optional presubmit in
assisted-service and assisted-image-service.
- Schedule a periodic job 2 times a week


osherdp (approved):

osherdp:
@adriengentil it can be a good idea letting people know of those new jobs, so they can trigger it if their PRs are changing something relevant

PR #36687, opened 2023-02-23
ARMOCP-511: Enable builds on arm01 for the hypershift org

aleskandro:
This PR enables arm64 builds for the hypershift prow org so that QE can use the produced images in the tests pipelines.
The images are mirrored in quay in the same repository as for amd64. The amd64 images are kept without any suffix, while the arm64 images will have the `-arm64` suffix. In future, when the clusters are upgraded to 4.13 and a path for Prow multi-arch builds is available, we will need to change this for producing ManifestList images.

Note: Avoiding <4.13 builds seems not possible yet (cc @droslean).

This is an initial work related to [ARMOCP-511](https://issues.redhat.com//browse/ARMOCP-511).

### DIFF
--- a/ci-operator/config/openshift/hypershift/.config.prowgen
+++ b/ci-operator/config/openshift/hypershift/.config.prowgen
@@ -1,0 +1,2 @@
+additional_architectures:
+- arm64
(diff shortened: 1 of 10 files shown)


deepsm007:
@droslean any thoughts if we want to only create postsubmits for release >=4.13 ?

aleskandro:
cc @LiangquanLi930

aleskandro:
Hello @enxebre  @sjenning, can you PTAL?

bryan-cox:
/retitle ARMOCP-511: Enable builds on arm01 for the hypershift org

PR #36690, opened 2023-02-23
Add aws-ipi-byo-iam-role profile

kasturinarra:
Add aws-ipi-byo-iam-role profile.


kasturinarra:

PR #36697, opened 2023-02-23
Migrate Serverless Interop Testing to OpenShift CI

liswang89:
Jira task: [LPTOCPCI-35](https://issues.redhat.com/browse/LPTOCPCI-35)

For Serverless, there are 4 test suites in current interop testing. We will run them from serverless-operator repo as 3 test suites can be covered in 2 upstream jobs. For now, we cannot decouple installation and test running. 2 upstream jobs need to install the backend one at a time.

The test results are verified in a temp PR using cluster pool:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_release/37297/rehearse-37297-periodic-ci-openshift-knative-serverless-operator-main-ocp4.13-lp-interop-operator-e2e-interop-aws-ocp413/1636718359305261056

Serverless QE is already on OpenShift CI for testing, here is the current config file for reference:
https://github.com/openshift/release/blob/master/ci-operator/config/openshift-knative/serverless-operator/openshift-knative-serverless-operator-main__4.12.yaml

### DIFF
--- a/ci-operator/jobs/openshift-knative/serverless-operator/openshift-knative-serverless-operator-main-periodics.yaml
+++ b/ci-operator/jobs/openshift-knative/serverless-operator/openshift-knative-serverless-operator-main-periodics.yaml
@@ -1384,7 +1384,7 @@ periodics:
   name: periodic-ci-openshift-knative-serverless-operator-main-ocp4.13-lp-interop-operator-e2e-interop-aws-ocp413
   reporter_config:
     slack:
-      channel: '#cspi-qe-slack-poc'
+      channel: '#serverless-qe'
       job_states_to_report:
       - failure
       - error


ascerra:
For scenarios we shouldn't be using the `config/rhpit` directory. This directory is just a sandbox for some simple testing.

The config file should live in a directory that either already exists for the layered products github org or you can create one based on the github org that the layered product we team is using to store their test repos.

ci-operator/config/{layered_product_github_org}/{layered_product_github_test_repo}/filename.

Getting started docs are here: https://cspi-qe.gitbook.io/ocp-ci-onboarding/onboarding/onboarding_guide/scenario_development_guide#getting-started

ci-operator/config/rhpit/interop-tests/rhpit-interop-tests-master__serverless-rehearse-4.12.yaml
```diff
@@ -33,7 +36,7 @@ resources:
       cpu: 100m
       memory: 200Mi
 tests:
-- as: installer-rehearse-aws
+- as: serverless-rehearse-aws
   interval: 960h
   steps:
     cluster_profile: aws-cspi-qe
```
ascerra, line 42: Please use the env var `SPOT_INSTANCES: "true"` when testing so that we save AWS costs.

ryankwilliams: Yes! Thank you @ascerra for pointing this out when doing testing when it does not require the standard AWS instance types.

https://github.com/openshift/release/blob/master/ci-operator/step-registry/ipi/install/install/ipi-install-install-ref.yaml#L40-L42

mgencur:
@liswang89 @ryankwilliams Hi! Recently, there was an email thread "Request for information on "Red Hat Openshift Serverless" Operator testing" where I was providing information requested by your team. For example: How to run tests in a container format. I was talking about [serverless-operator-src](http://registry.ci.openshift.org/knative/openshift-serverless-v1.27.0:serverless-operator-src) image that has all the sources, and now I'm seeing this PR doesn't use anything from the email, it basically copies the approach we use for serverless-operator in CI now, and that includes only one test suite, this is far from what we currently run as part of INTEROP testing.

I'm curious why it's done in this way and you don't use the "container format" as requested in the email.

The current approach in this PR has a couple of problems and we'll need to discuss how to proceed further:
* The "Make" target you're calling actually installs the serverless operator again, just from the current catalog source that is built as part of the step (not from "redhat-operators")
* This is just one test suite, pretty small. For INTEROP testing, we currently run several more test suites which can be seen in [this job](https://master-jenkins-csb-serverless-qe.apps.ocp-c1.prod.psi.redhat.com/job/functional_tests/job/live/103/testReport/). If we want to run all the test suites in CI we'll need a few jobs because the tests together run more than 2 hours and CI has some limits wrt execution time.

Please let me know what you think. Thanks

mgencur:
@liswang89 Please let me know what the limits are for test execution time. Anyway. I think we either need to reduce the testing or split the job. Not sure what's acceptable for your team or how you'll be processing results.

liswang89:
@mgencur If I understand correctly, we want to install it from the latest version of Serverless, so we might need to install it from OperatorHub and the release-1.27 branch. The question is whether it is ok to decouple the [serverless-operator-src](http://registry.ci.openshift.org/knative/openshift-serverless-v1.27.0:serverless-operator-src) image when we want to install serverless operator and serving and Knative eventing separately.
As for test suites, I will add more as I get the first one passed.

mgencur:
I think something like this should work for you:
```
- as: operator-e2e-aws-ocp-413
  cron: 0 1 * * 1
  steps:
    allow_skip_on_success: true
    cluster_profile: aws
    test:
    - as: serverless-install
      commands: make install-all
      dependencies:
      - env: KNATIVE_OPENSHIFT_INGRESS
        name: knative-openshift-ingress
      - env: KNATIVE_OPERATOR
        name: knative-operator
      - env: OPENSHIFT_KNATIVE_OPERATOR
        name: openshift-knative-operator
      from: serverless-operator-src
      resources:
        limits:
          memory: 6Gi
        requests:
          cpu: 100m
          memory: 200Mi
    - as: serverless-e2e
      commands: make test-e2e-with-kafka-testonly
      from: serverless-operator-src
      resources:
        limits:
          memory: 6Gi
        requests:
          cpu: 100m
          memory: 200Mi
    workflow: openshift-e2e-aws
```
Notes:
* The `install-all` target will install Serverless operator from the given branch, install Knative Serving, Knative Eventing and also Kafka. So everything will be installed at this point.
* The `test-e2e-with-kafka-testonly` target will run serverless operator tests (without installation). The test suite is limited. We would also like to run `test-upstream-e2e-kafka-no-upgrade-testonly` and `test-upstream-e2e-no-upgrade-testonly` as part of INTEROP testing. But together they run more than 2 hours, they would need to be split. Probably into three different jobs.
* It uses the `openshift-e2e-aws` workflow which we were using previously when cluster pools were not available. It works without cluster pools.
* Regarding "main" vs. "release-1.27" branch. I'd prefer placing this config on "main" branch as long as we install the operator from current sources (via `install-all`, not using the `install-operators` step which installs a previous version).

ascerra:
@mgencur it looks like we will be grabbing the first channel on the list by default for `OLM_CHANNEL`
[vars.bash](https://github.com/openshift-knative/serverless-operator/blob/f7df04a1491a23a39765bb61722ce616c90ba40f/hack/lib/vars.bash#L68)

But if we wanted to change this all we would need to do is set the var `OLM_CHANNEL` from the config file to override the default I believe, is this correct?

mgencur:
@ascerra OpenShift Serverless currently provides only a single channel named "stable". We don't have other channels at the moment. The list can be seen in https://github.com/openshift-knative/serverless-operator/blob/main/olm-catalog/serverless-operator/project.yaml#L10

liswang89:
@mgencur thanks for the suggestions. Now, 5/91 tests failed. I am using ipi-aws workflow(recommended openshift-e2e-aws) and ocp4.12(recommended 4.13). I wondered whether the test failures were caused by other configuration setups or can be fixed with the recommended change.
@ascerra @ryankwilliams what OCP version we are going to run against? workflow openshift-e2e-aws has extra tests running on provisioned cluster. Is it ok to use?

mgencur:
> Also test-e2e-with-kafka-testonly triggered multiple upstream jobs on OCP4.8. Is this something expected or need to fix?

That's because you're modifying also some jobs for 4.8 which shouldn't be modified, like on [this line](https://github.com/openshift/release/pull/36697/files#diff-bfbd2127cb2a7a50b8a47af5995a05053999cdbcbd0eeb9185dc1752715d05c7R768). This needs to be changed back.

> Please fix it with test-e2e-with-kafka-testonly.

I'm going to send a quick PR for this shortly.

ascerra:
Here are the current tests that we report through the interop program today for this scenario.

@mgencur are the tests that we are targeting with `make test-e2e-with-kafka-testonly` covering these 4 test suites or are there other tests/commands that we need to account for? 
Screenshot from 2023-03-06 10-22-28

mgencur:
@ascerra 
Those 4 test cases are aggregations for the following test suites:
* serverless-operator-e2e is covered by `test-e2e-with-kafka-testonly`
* knative-serving-e2e and knative-eventing-e2e are covered by `test-upstream-e2e-no-upgrade-testonly`
* knative-eventing-kafka-broker-e2e are covered by `test-upstream-e2e-kafka-no-upgrade-testonly`
But together they run more than 4 hours so they need to be split into several jobs (that's also why we have the "upstream" tests split into separate Make tergets

ci-operator/config/openshift-knative/serverless-operator/openshift-knative-serverless-operator-main__ocp4.13-lp-interop.yaml
```diff
@@ -0,0 +1,173 @@
+base_images:
```
calebevans, line 1: Should we be in the OWNERS file for this directory? @ascerra @btjd

ascerra: I believe that we should be. This will allow for us to merge updates to this config if we need to. We may realize that we've made a mistake somewhere or implement a new tool like security checking that we'll want to be able to update this file without requesting PQE to do it.

btjd: +1 it's good to have that option in case we need it.

mgencur:
Looks good in general. I'm wondering who will be watching this job. With the previous approach for INTEROP testing, the PIT QE team triggered our Jenkins job and we (the product team) were notified when the job failed and were looking into the failure.
Asking because it is possible to add Slack notifications to the config like [this one](https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift-knative/serverless-operator/openshift-knative-serverless-operator-main-periodics.yaml#L1082-L1089).

ascerra:
> > Looks good in general. I'm wondering who will be watching this job. With the previous approach for INTEROP testing, the PIT QE team triggered our Jenkins job and we (the product team) were notified when the job failed and were looking into the failure. Asking because it is possible to add Slack notifications to the config like [this one](https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift-knative/serverless-operator/openshift-knative-serverless-operator-main-periodics.yaml#L1082-L1089).
> 
> @mgencur Quick question about the slack example you linked. I see that slack is only mentioned in the Prow job and not the config. Our prow jobs are generated from our config using the Makefile, but I guess the slack feature is not support through the ci-operator config and is just supported through prow. So I'm guessing that this slack example you shared is added manually to the prow job after you use make update to create the prow job, is that right?

mgencur:
> So I'm guessing that this slack example you shared is added manually to the prow job after you use make update to create the prow job, is that right?

Exactly.

mgencur:
This PR looks good to me. No other comments from my side. Let me know when this PR is not WIP and I will stamp it.

ci-operator/jobs/openshift-knative/serverless-operator/openshift-knative-serverless-operator-main-periodics.yaml
```diff
@@ -1364,3 +1364,86 @@ periodics:
...
+    ci-operator.openshift.io/cloud: aws
+    ci-operator.openshift.io/cloud-cluster-profile: aws-cspi-qe
+    ci-operator.openshift.io/variant: ocp4.13-lp-interop
+    ci.openshift.io/generator: prowgen
+    job-release: "4.13"
+    pj-rehearse.openshift.io/can-be-rehearsed: "true"
+  name: periodic-ci-openshift-knative-serverless-operator-main-ocp4.13-lp-interop-operator-e2e-interop-aws-ocp413
+  reporter_config:
+    slack:
+      channel: '#serverless-qe'
```
mgencur, line 1387: This line was previously different and pointed to a different Slack channel. Now it will be sending messages to serverless-qe channel. I previously asked who will be watching this job in https://github.com/openshift/release/pull/36697#issuecomment-1480690502
With this PR, does the workflow between PIT QE and Serverless QE change?
Can we stop using our former jenkins jobs and focus on this CI stuff and slack notifications?
If you're not able to answer this please give me some contact information to somebody who can.

I won't stamp this PR until my questions are answered.

Aside from that this PR looks alright.

ci-operator/config/openshift-knative/serverless-operator/openshift-knative-serverless-operator-main__ocp4.13-lp-interop.yaml
```diff
@@ -0,0 +1,171 @@
...
+resources:
+  '*':
+    limits:
+      memory: 6Gi
+    requests:
+      cpu: 100m
+      memory: 200Mi
+tests:
+- as: operator-e2e-interop-aws-ocp413
+  cron: 0 1 * * 1
```
ascerra, line 112: This should be `0 6 * * 1`

PR #36704, opened 2023-02-23
Update rukpak to use golang 1.19 and 4.13

tmshort:
Update only main and release-4.14


tmshort:
The `e2e` jobs should succeed when https://github.com/openshift/operator-framework-rukpak/pull/19 is merged.

PR #36718, opened 2023-02-23
add -ovn- to cluster-monitoring-operator presubmits

jluhrsen:
this will fix some flakes that are checking that the job name contains the string of the underlying CNI, which is ovn by default in 4.12+

found with this one-liner:
❯ w3m -dump -cols 200 "https://search.ci.openshift.org/?maxAge=96h&type=junit&search=job+name.*does+not+have+network+type+in+name.*expected.*ovn" | sed -n 's/^\([^ ]*\) .*failures match.*/\1/p' | sort | rg cluster-monitoring pull-ci-openshift-cluster-monitoring-operator-master-e2e-agnostic pull-ci-openshift-cluster-monitoring-operator-master-e2e-agnostic-upgrade pull-ci-openshift-cluster-monitoring-operator-master-e2e-aws-single-node

### DIFF
--- a/ci-operator/config/openshift/cluster-monitoring-operator/openshift-cluster-monitoring-operator-master.yaml
+++ b/ci-operator/config/openshift/cluster-monitoring-operator/openshift-cluster-monitoring-operator-master.yaml
@@ -77,11 +77,11 @@ tests:
   commands: make test-unit
   container:
     from: src
-- as: e2e-agnostic
+- as: e2e-aws-ovn
   steps:
     cluster_profile: aws
     workflow: openshift-e2e-aws
-- as: e2e-aws-single-node
+- as: e2e-aws-ovn-single-node
   optional: true
   steps:
     cluster_profile: aws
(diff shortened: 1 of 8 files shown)


JoaoBraveCoding (commented):
So I'm not too familiar with this part of the stack. With that in mind as far as I can see this looks like a rename, could you provide just a bit more insight on the following points:
- why were the monitoring tests moved from `agnostic` to `aws-ovn`? AFAIK the monitoring stack doesn't have any dependency on AWS
- why was the `e2e-agnostic-operator` not moved to the same name?

jluhrsen:
> So I'm not too familiar with this part of the stack. With that in mind as far as I can see this looks like a rename, could you provide just a bit more insight on the following points:
> 
> * why were the monitoring tests moved from `agnostic` to `aws-ovn`? AFAIK the monitoring stack doesn't have any dependency on AWS
> * why was the `e2e-agnostic-operator` not moved to the same name?

@JoaoBraveCoding , that's right, this is just a job rename. In the past year some tests were added that
will check if the job name matches the underlying platform and CNI. These jobs were running on AWS
w/ OVN (that's the default in 4.12+) and the test cases were flaking. NBD and you can live with the
flakes if you want. If not you can approve this PR and they'll go away.

I've been going through jobs/projects slowly in the past 6+ months and making these kinds of PRs. I know
it seems like it came out of the blue :)

JoaoBraveCoding (approved):
Sounds clear! Then I see no reason as to not merge this so adding the green mark

/lgtm

PR #36739, opened 2023-02-27
Promote the image in the registry

cjeanner:
This way, devs will be able to reference the image using its name directly, instead of having to build it on the fly.

Currently, the usecase of this image is:
- allow to run molecule tests
- allow to run pre-commit tests

It may be extended to other tasks, but that's the main focus right now.

Promoting the image will ensure we're able to share a common environment and set of tooling with other related projects.


cjeanner:
/hold

Need to get @viroel feedback. Prevent merge until then.

cjeanner:
/test all

We merged a new thing in the ci-framework repository, let's ensure it's still all good since we have an open PR here.

PR #36746, opened 2023-02-27
Ensure we get the actual code to test

cjeanner:
The "root" build takes master content, not the on from the PR. So we need to get the "src" image, built onto the "root" one, in order to get the actual PR.


cjeanner:
(rebased against master)

PR #36757, opened 2023-02-27
update cron job for zonal ga

sgaoshang:


jianlinliu:
cc @liangxia for review

PR #36761, opened 2023-02-27
load-balancer: remove backup mode from backends

EmilienM:
This was disabling actual balancing over the backends and enabling
active/passive mode. We don't want that.


EmilienM:
/pj-rehearse pull-ci-openshift-installer-master-e2e-openstack-externallb-techpreview

EmilienM:
the CI seems flacky and not under our control, I think this is good to go and I actually deployed a cluster with this config and it worked.

PR #36767, opened 2023-02-27
VPA operator: update to golang 1.19

joelsmith:


joelsmith:
`manifest-diff` is expected to fail until https://github.com/openshift/vertical-pod-autoscaler-operator/pull/139 merges.

PR #36784, opened 2023-02-28
Setup ocp-ci for red-hat-data-services/data-science-pipelines-operator

gmfrasca:


danilo-gemoli:
@gmfrasca hey, have a look at [here](https://docs.ci.openshift.org/docs/architecture/ci-operator/#building-artifacts)

gmfrasca:
Hey @danilo-gemoli - thanks for the ref; think i've fixed it and looks like all checks are passing now.  Is this good to go now?

FWIW this should be near-identical to the setup/configs for the pre-existing opendatahub-io/data-science-pipelines-operator, albeit without some postsubmit/build promotion stuff.

PR #36803, opened 2023-02-28
submariner-addon: bump golangci-lint to v1.48.0

skitt:
This supports Go 1.19.


skitt:
/pj-rehearse

I suspect the verification jobs will fail because of the changed formatting expectations in Go 1.19.

Jaanki (approved):

skitt:
The `golangci-lint` job passed, using the v1.48.0 image, which is what we want to check.

/pj-rehearse ack

PR #36804, opened 2023-02-28
Correct image tagging

cjeanner:
Doc[1] is a bit misleading: apparently, the "latest" tag doesn't seem to exist. Also, the "name" isn't really used, or at least there's no real use to it since the promoted image already has a fixed name, "ci-framework-image".

This patch removes the "name" tag for the image promotion, and ensures we get a "latest" tag around, so that we can use it to point to the actual latest image.

The tag_by_commit is kept, so that we may revert to a previous image in case of issue with the latest one.

[1] https://docs.ci.openshift.org/docs/architecture/ci-operator/#publishing-images-tagged-by-commit


viroel:
/lgtm

lets see how it goes on registry side

PR #36850, opened 2023-03-01
Add hypershift labels to plugin configuration

csrwng:
Adds hypershift labels to repository plugin configuration


bryan-cox (approved):
/lgtm

PR #36852, opened 2023-03-01
step-registry/install: Add `OPERATING_SYSTEM_IMAGE`

cgwalters:
This is a straightforward passthrough of support for overriding the coroes image via MachineConfig.

In particular, I really want to enable things like `workflow-launch ipi-aws nightly OPERATING_SYSTEM_IMAGE=quay.io/cgwalters/ostest@sha256...` to deploy a custom OS image.


cgwalters:
(I have no idea if this is going to work, but I think it will...)

cgwalters:
OK I'm confused about the differences between `launch` and `workflow-launch`.  First, I'd expected that via `workflow-launch` I could directly run a step from the step registry, but that appears to not be the case.  Also, what the heck is "cucushift"?

Is going through `workflow-launch` the right way to do this or should I try to hack something into `launch`?

PR #36864, opened 2023-03-02
Skip generation of cron entries for *baremetal-* jobs

aleskandro:
As anticipated offline, the baremetal jobs to execute in the lab that we are going to introduce, should be skipped by the cron entry generation and be considered manually managed until a logic scheme is decided.

This PR makes the bash script skip the cron generation for `*baremetal-` jobs, unless a `--force` flag is passed in the arguments.

/cc @liangxia @jianlinliu 

Tested up to line 30.

```shell
└ $ ./generate-cron-entry.sh baremetal-asd asd.yaml
The test config baremetal-asd should not get changes in the cron entry as
      the schedule rotation scheme is different than the other tests.
      Use --force to skip this check.

└ $ ./generate-cron-entry.sh baremetal-asd asd.yaml --force
...
└ $ ./generate-cron-entry.sh baremetalds-asd asd.yaml      
...
└ $ ./generate-cron-entry.sh aws-asd asd.yaml
...
```

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/tools/update-cron-entries.py
+++ b/ci-operator/config/openshift/openshift-tests-private/tools/update-cron-entries.py
@@ -7,6 +7,7 @@ import os
 import subprocess
 from datetime import datetime
 
+
 def shell(cmd, debug=False, env_=dict(os.environ)):
     """
     Args:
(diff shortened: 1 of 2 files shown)


ci-operator/config/openshift/openshift-tests-private/tools/update-cron-entries.py
```diff
@@ -34,60 +35,74 @@ def shell(cmd, debug=False, env_=dict(os.environ)):
...
+        print("Updating %s" % (target_file))
+
         with open(target_file, 'r') as file:
             all_data = yaml.safe_load(file)
         file.close()
-    
+
         all_tests_list = all_data['tests']
         new_tests_list = []
+        cron_prefix = "cron:"
```
jianlinliu, line 85: I guess we can move the line out of the for loop, e.g: to line 75.

aleskandro: sure

PR #36884, opened 2023-03-02
OPNET-141: Add clusterbot support for vSphere dual-stack

mkowalski:


dgoodwin:
/hold

Should check with someone from continuous release tooling

ci-operator/jobs/openshift/release/openshift-release-infra-periodics.yaml
```diff
@@ -1120,6 +1120,14 @@ periodics:
               test:
               - ref: clusterbot-wait
               workflow: openshift-e2e-$(CLUSTER_TYPE)-sdn
+          - as: launch-dualstack
+            steps:
+              cluster_profile: vsphere
```
AlexNPavel, line 1125: This should use `$(CLUSTER_PROFILE)` instead of being hardcoded.

AlexNPavel (changes requested):

ci-operator/jobs/openshift/release/openshift-release-infra-periodics.yaml
AlexNPavel: If you do want to use the `vsphere` instead of `vsphere-clusterbot` profile, we can leave this hardcoded. Just wanted to make sure that having it different was intentional.

mkowalski: Yeah so that was my intention, to use "any" cluster and not only `vsphere-clusterbot`; in principle anything that can match so that the `launch vsphere,dualstack` lane has as much as possible resources

AlexNPavel (approved):

ci-operator/jobs/openshift/release/openshift-release-infra-periodics.yaml
rvanderp3: we run specific variants of clusterbot for vSphere in different lease pools.  That is the reason we don't ${CLUSTER_PROFILE} consistently.

mkowalski:
/unhold

Reviewed, approved, lgtm-ed, no reason to hold any more

PR #36890, opened 2023-03-02
DPTP-3326: Add a clusterimageset for using 4.next EC payloads

bparees:


clusters/hive/pools/ocp-release-4.next-x86-64-for-4.13.0-0-to-4.14.0-0_clusterimageset.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: hive.openshift.io/v1
+kind: ClusterImageSet
+metadata:
+  annotations:
+    version_lower: 4.13.0-0
+    version_upper: 4.14.0-0
```
bparees, line 6: until the clusterimageset updater understands the 4-dev-preview stream, i'm not sure what affect this is going to have.  Is there a way to prevent the clusterimageset updater from trying to touch this file, since it's presumably going to stomp the `releaseImage` value below?

bparees:
@bear-redhat ptal

clusters/hive/pools/ocp-release-4.next-x86-64-for-4.13.0-0-to-4.14.0-0_clusterimageset.yaml
```diff
@@ -0,0 +1,11 @@
...
+  creationTimestamp: null
+  name: ocp-release-4.next-x86-64-for-4.13.0-0-to-4.14.0-0
+spec:
+  releaseImage: quay.io/openshift-release-dev/ocp-release:4.13.0-ec.3-x86_64
```
bparees, line 10: i don't think we need a new clusterpool/creds because this clusterimageset is pulling from the same location as the existing 4.12 one (which pulls from `quay.io/openshift-release-dev/ocp-release:4.12.6-x86_64`), it's just a different tag.

bparees:
/hold

i'm working on modifying clsuterimageset updater to be stream aware which will make this much easier.

clusters/hive/pools/cvp/cvp-ocp-4-13-next-amd64-aws-eu-west-3_clusterpool.yaml
```diff
@@ -0,0 +1,40 @@
...
+  labels:
+    architecture: amd64
+    cloud: aws
+    owner: cvp
+    product: ocp
+    region: eu-west-3
+    version: "4.13"
+    version_lower: 4.13.0-0
+    version_stream: 4-dev-preview
+    version_upper: 4.14.0-0
```
bparees, line 14: I had some debate about whether we'd want an imageset+pool for each minor, or just a "4.next" one that always uses the newest 4.next payload (i.e. would automatically transitition from a 4.13.ec to a 4.14.ec once they become available).

I can go either way on it, but specifying the "version" on line 11 above becomes tricky if the version can evolve (unless we can just set it to "4.next", i wasn't sure what was a valid value for that label).

If we did want to go that way, the easiest thing to do would be to just set the version_upper to a really high value like `4.999.0-0` and then i think everything else should just work.

bparees:
This needs to wait for https://github.com/openshift/ci-tools/pull/3314 before merging

Issue #732, opened 2018-03-29
Kill all jobs not run by prow


kargakis:
Please do

Mainly travis at this point; there is also a fedora/27/atomic in openshift-ansible. Unclear where it runs. 

@stevekuznetsov @bbguimaraes

stevekuznetsov:
`fedora/*/atomic` is not going away in o-a

kargakis:
Is it something we cannot move to run in prow?

stevekuznetsov:
Don't think so, @sdodson can we get a contact for that status?

sdodson:
@jlebon is the contact for that job. It relies on Fedora's CI infra.

kargakis:
Is it Jenkins behind the covers? We can still use prow and trigger the job in its original place as long as we can get a bot account there.

jlebon:
`fedora/27/atomic` uses PAPR: https://github.com/projectatomic/papr.
PAPR doesn't run in the Fedora infra, but on an internal OpenStack instance (though we're working towards migrating it to OCP).

I'm not sure how easy it'd be to make a prow --> PAPR compat shim. Seems like the easiest thing to do would be to just move over to Prow completely instead. It's fairly straightforward: it provisions three FAH27 nodes and runs the openshift-ansible installer.

stevekuznetsov:
@jlebon I poked around the repo for a bit but could not find this -- how is all that triggered? If we can clone the PAPR repo into a container and run the scripts, providing the env vars as necessary, that seems like a good shim.

jlebon:
It's triggered by GHPRB/SCM plugins in an internal Jenkins instance which then fire up PAPR. IOW, PAPR isn't a service that waits for triggers, it's a "one-shot" thing that runs tests in an environment which assumes it has access to e.g. an OpenStack instance for deploying VMs and AWS credentials for publishing logs.

Does the Prow infra have any facilities for provisioning multiple VMs? One nice thing PAPR provides is easy access to the latest FAH27 images, though there are symlinks that exist to make downloading those really convenient: https://fedoraproject.org/wiki/Cloud#Fedora_Cloud_Atomic_Image_Download_Links (which is what the PAPR infra uses to keep the OpenStack images up to date).

stevekuznetsov:
[...]

PR #36898, opened 2023-03-02
Increase resources for odh-manifests e2e tests

VedantMahabaleshwarkar:
Currently test runs are failing as resource creation on the cluster fails due to lack of resources. Increasing the resource limits for the tests suite will fix the issue


LaVLaS:
@VedantMahabaleshwarkar This looks like it will only change the resources allocated to the test container that is executing the odh-manfiests test script and not the cluster

Based on the test run in your [odh-manifests PR #732](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/opendatahub-io_odh-manifests/732/pull-ci-opendatahub-io-odh-manifests-master-odh-manifests-e2e/1631388299778068480), it looks like the actual test pod where the test script is running isn't crashing due to memory resources.

PR #36908, opened 2023-03-03
add quay 3.8 test on aro with OCP 4.12

LiZhang19817:
Add quay 3.8 test on aro with OCP 4.12


LiZhang19817:
The Operator Hub is not running in ARO.

PR #36926, opened 2023-03-03
Remove dup jobs from nova- and placemnet-operator

gibizer:
We have pre-commit job that runs fmt, vet, golangci-lint already so those jobs are duplicates. This PR removes the dups from nova- and placement-operator


SeanMooney (approved):
yep precommit does this for us
we do not need these duplicate jobs

PR #36932, opened 2023-03-03
[OKD/SCOS] *: Branch configs for master == 4.14

LorbusChris:
/cc @jmguzik @vrutkovs


LorbusChris:
master / release-4.14 tests are expected to fail as the `scos-4.14` imagestream will only be created when this PR merges

PR #36937, opened 2023-03-03
machine-config-operator: Add gcp-ovn-rt-upgrade

cgwalters:
This is to help test https://github.com/openshift/machine-config-operator/pull/3485

This specific job is not one of the current optional presubmits, it's been executed by the payload jobs.  But the payload jobs are *expensive* and *slow* because they run a lot of tests like 8 times.

With this we should be able to just `/test e2e-gcp-ovn-rt-upgrade` and get it run once, not 8 times.


cgwalters:
Followup to this in https://github.com/openshift/release/pull/36953 cc @cheesesashimi

PR #36943, opened 2023-03-03
CNF-5652: feat: add cpu partitioned cluster pipeline

eggfoobar:
Adds e2e tests running on a CPU Partitioned cluster. Currently the feature is in TechPreview, we will run a few of these tests to confirm behavior before removing the TechPreview designation.

/hold on openshift/origin#27770

### DIFF
--- a/ci-operator/step-registry/ipi/conf/cpu-partitioning/ipi-conf-cpu-partitioning-ref.yaml
+++ b/ci-operator/step-registry/ipi/conf/cpu-partitioning/ipi-conf-cpu-partitioning-ref.yaml
@@ -1,0 +1,13 @@
+ref:
+  as: ipi-conf-cpu-partitioning
+  from_image:
+   namespace: ocp
+   name: "4.13"
+   tag: upi-installer
+  commands: ipi-conf-cpu-partitioning-commands.sh
+  resources:
+    requests:
+      cpu: 10m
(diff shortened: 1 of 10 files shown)


eggfoobar:
/assign @dgoodwin 

Hey Devan, these pipelines are to test out the CPU Partitioning feature, the interval is set momentarily high at 6h to help get some data points and confidence on this feature so we can remove it out of tech preview. We'll revisit this interval in the coming weeks. Hope that's not going to be a problem

ci-operator/step-registry/ipi/conf/cpu-partitioning/ipi-conf-cpu-partitioning-ref.yaml
```diff
@@ -0,0 +1,13 @@
+ref:
+  as: ipi-conf-cpu-partitioning
+  from_image:
+   namespace: ocp
+   name: "4.12"
```
dgoodwin, line 5: Is 4.12 correct here?

eggfoobar: I had assumed that was fine since it's just doing a simple yq transform, but that's a good call out, let me keep it up to date with the min version this feature touches.

PR #36953, opened 2023-03-04
machine-config-operator: e2e-gcp-ovn-rt-upgrade always_run=false

cgwalters:
It's nice to have it available, but in practice it's not super useful for the MCO, because these `-upgrade` tests just do a *synthetic* update and not an OS upgrade.  We want to be using the payload tests.

We'll keep this (it's cheaper and maybe useful sometimes) but make it `always_run=false` so it's not run by default and blocking.


sinnykumari:
Agree. Thanks for the PR.
/lgtm

PR #36955, opened 2023-03-04
hypershift: update enable-qe-catalogsource

LiangquanLi930:


ci-operator/step-registry/cucushift/hypershift-extended/enable-qe/catalogsource/cucushift-hypershift-extended-enable-qe-catalogsource-commands.sh
```diff
@@ -98,6 +98,10 @@ EOF
     set -e
 }
 
+if [[ $SKIP == "true" ]]; then
```
heliubj18, line 101: better to add a log

PR #36974, opened 2023-03-06
Add redhat-appstudio/integration-service repository

jinqi7:


ci-operator/config/redhat-appstudio/integration-service/redhat-appstudio-integration-service-main.yaml
```diff
@@ -0,0 +1,41 @@
...
+    tag: latest
+images:
+- dockerfile_path: Dockerfile
+  to: redhat-appstudio-integration-service-image
+releases:
+  latest:
+    release:
+      architecture: amd64
+      channel: stable
+      version: "4.10"
```
psturc, line 14: probably better to use 4.11 or 4.12 (which should be backward compatible)

ci-operator/config/redhat-appstudio/integration-service/redhat-appstudio-integration-service-main.yaml
```diff
@@ -0,0 +1,41 @@
...
+      memory: 200Mi
+tests:
+- as: integration-service-e2e
+  cluster_claim:
+    architecture: amd64
+    cloud: aws
+    owner: openshift-ci
+    product: ocp
+    timeout: 1h0m0s
+    version: "4.11"
```
psturc, line 30: switching to 4.12 here would probably make sense as well (although I think we're using 4.11 in other jobs - but it should be updated)

psturc (approved):
just minor comments, otherwise lgtm

ci-operator/config/redhat-appstudio/integration-service/redhat-appstudio-integration-service-main.yaml
jinqi7: Updated. Thanks

ci-operator/config/redhat-appstudio/integration-service/redhat-appstudio-integration-service-main.yaml
jinqi7: Thanks. Updated.

dirgim (approved):
/lgtm

PR #36982, opened 2023-03-06
feat: added the test steps for rt-tests to the rt upgrade jobs

jeff-roche:
Adds a chain that calls the realtime suite in the test binary during realtime upgrade cycles


eggfoobar:
/approve
/assign @dgoodwin 

Hey Devan, this addition makes use of https://github.com/openshift/origin/pull/27778 and https://github.com/openshift/origin/pull/27751 and will allow us to run some RT specific tests after the e2e tests for 4.13>= clusters.

PR #37018, opened 2023-03-07
Add ipi-gcp-techpreview workflow and use it in aesgcm jobs

dgrisonnet:
AES-GCM encryption is still in techpreview right now and we will need to a workflow enabling the TechPreviewNoUpgrade cluster configuration in order to run the tests for that feature properly.

Since there was no preexisting ipi for this, only the `openshift-e2e-gcp-techpreview` workflow, I created a new `ipi-gcp-techpreview`.

### DIFF
--- a/ci-operator/config/openshift/cluster-kube-apiserver-operator/openshift-cluster-kube-apiserver-operator-master.yaml
+++ b/ci-operator/config/openshift/cluster-kube-apiserver-operator/openshift-cluster-kube-apiserver-operator-master.yaml
@@ -118,7 +118,7 @@ tests:
         requests:
           cpu: 100m
       timeout: 4h0m0s
-    workflow: ipi-gcp
+    workflow: ipi-gcp-techpreview
 - as: e2e-aws-ovn-upgrade
   steps:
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/ipi/gcp/techpreview/ipi-gcp-techpreview-workflow.metadata.json
```diff
@@ -0,0 +1,12 @@
...
+	"owners": {
+		"approvers": [
+			"jianlinliu",
+			"patrickdillon",
+			"technical-release-team-approvers",
+			"vrutkovs",
+			"yunjiang29"
+		]
+	}
+}
```
dgrisonnet, line 12: This is copied from https://github.com/openshift/release/blob/02e8f2de79b7a6ca4c6b22d89f27c0a4abe91ddf/ci-operator/step-registry/ipi/gcp/ipi-gcp-workflow.metadata.json.

dgrisonnet:
@vrutkovs could you please have a look at the new ipi?

/cc @deads2k

PR #37004, opened 2023-03-07
Setting AWS KMS key in defaultMachinePlatform

yunjiang29:
Resolve issue [CORS-2080](https://issues.redhat.com/browse/CORS-2080) Extend user encryption keys usage for the default storage

Setting KMS key in `installconfig.platform.aws.defaultMachinePlatform.rootVolume.kmsKeyARN` instead of controlPlane and compute nodes separately. 

AWS supports `platform.aws.defaultMachinePlatform.rootVolume.kmsKeyARN` starting from 4.5: https://github.com/openshift/installer/pull/3293

PR #37019, opened 2023-03-07
Fix docs for ipi-conf-aws-kms-key

yunjiang29:
resolve https://github.com/openshift/release/pull/37004#discussion_r1127512545


yunjiang29:
@jianli-wei PTAL

PR #37027, opened 2023-03-07
Re-enable build04

smg247:
[OHSS-18195](https://issues.redhat.com/browse/OHSS-18195) is finished with triage, and a solution is pending implementation. This issue isn't happening on any of our other clusters, and we have had `build04` disabled for 6 weeks now. Lets move some load back to it now with the hope that the issues we were facing will no longer be relevant.

Effectively reverts: https://github.com/openshift/release/pull/35749


smg247:
/close
We are going to wait on this one until the issue is resolved since CI is operational without build04 currently.

smg247:
We updated the registry to only have 2 nodes, and are going to move the load back to `build04` to confirm that the issue was correctly identified. If so, we shouldn't see the errors.

PR #37054, opened 2023-03-08
Kuryr: Disable merge job

dulek:
We're past the feature freeze, so this commit disables Kuryr merge job by setting it to run at 31.02, which is an impossible date.


gryf (approved):
/lgtm

dulek:
Okay, seems like we're branched already, so this is not needed anymore.

PR #36836, opened 2023-03-01
Enable aesgcm encryption tests for cluster-kube-apiserver-operator

dgrisonnet:
This PR adds the following jobs to test encryption for both the aescbc and aesgcm providers:

- e2e-gcp-operator-encryption-aescbc (previously e2e-gcp-operator-encryption-aescbc)
- e2e-gcp-operator-encryption-aesgcm
- e2e-gcp-operator-encryption-rotation-aescbc (previously e2e-gcp-operator-encryption-rotation)
- e2e-gcp-operator-encryption-rotation-aesgcm
- e2e-gcp-operator-encryption-perf-aescbc (previously e2e-gcp-operator-encryption-perf)
- e2e-gcp-operator-encryption-perf-aesgcm

PR #37064, opened 2023-03-08
Temporarily mark aescbc jobs as optional

dgrisonnet:
We need to temporarily mark aescbc jobs as optional because the new workflow added by https://github.com/openshift/release/pull/36836 is breaking the tests until https://github.com/openshift/cluster-kube-apiserver-operator/pull/1449 is merged.


stlaz:
/lgtm
From what I understand, a blocker bug for the optional test exists

PR #37068, opened 2023-03-08
Add aesgcm and aescbc jobs as optional in 4.13

dgrisonnet:
Adding the new jobs introduced by https://github.com/openshift/release/pull/36836 to 4.13 since https://github.com/openshift/cluster-kube-apiserver-operator/pull/1449 will be backported to 4.13.

For now, we make the jobs optional since they will need the backport above to be merged in order to pass and we don't want to break CI for everyone while doing that. Similarly to what happened with master: https://github.com/openshift/release/pull/37064


sanchezl (approved):
/lgtm

PR #37074, opened 2023-03-08
WIP steps: Introduce KubeVirt specific must-gather steps

lyarwood:


lyarwood:
https://github.com/openshift/release/pull/37262 a less involved change that might work but I'll need to see the SSP func tests fail before I can prove it works.

PR #37081, opened 2023-03-08
cluster-capi-operator: make gcp/azure TP OCP e2es optional

damdo:
We are already testing the CAPI E2Es against techpreview on all the TP platforms.
And we are mandatorily testing the capi operator's integration with the TP payload with the e2e-aws-ovn-techpreview,
as such we want to make azure/gcp TP OCP E2Es optional.


damdo:
PR is only making some jobs optional, no actual need to rehearse.
/pj-rehearse ack

PR #37093, opened 2023-03-09
Use disconnected upgrade workflow on disconnected cluster

shellyyang1989:
Closes [OCPQE-14241](https://issues.redhat.com/browse/OCPQE-14241)

We should use disconnected upgrade workflow on disconnected profiles.

cc @JianLi-RH @jhou1 for review. Thanks!


shellyyang1989:
Refers [OCPQE-14339](https://issues.redhat.com/browse/OCPQE-14339)

PR #37101, opened 2023-03-09
Make e2e-openstack-ovn optional

MaysaMacedo:
Given that the cloud where this job runs is currently having issues to create VMs, let's make that job optional and instead one openstack job that runs in another cloud as required. This workaround can be reverted once the cloud is fixed.


pierreprinetti (approved):

pierreprinetti:
/hold
The fix in Vexxhost might be in place already

PR #37108, opened 2023-03-09
Add: set default sc ref

Phaow:
Add set default sc ref to extend qe test CI test coverage.

### DIFF
--- a/ci-operator/step-registry/storage/conf/storageclass/set-default-storageclass/storage-conf-storageclass-set-default-storageclass-commands.sh
+++ b/ci-operator/step-registry/storage/conf/storageclass/set-default-storageclass/storage-conf-storageclass-set-default-storageclass-commands.sh
@@ -29,47 +29,16 @@ function get_default_sc() {
 }
 
 function remove_all_default_sc_annotations () {
-  if [ -z "${CURRENT_DEFAULT_SC}" ]
-  then
-      echo "Currently no default storageclass is set, no need to remove"
-  else
-      for SC in ${CURRENT_DEFAULT_SC}
-      do
-        oc annotate sc "${SC}" "storageclass.kubernetes.io/is-default-class=false" --overwrite
-      done
-  fi
+  run_command "oc annotate sc storageclass.kubernetes.io/is-default-class=false --all --overwrite"
 }
 
-function set_cluster_csi_driver_sc_state_unmanaged () {
-  local ret=0
-  local provisioner=""
-  run_command "oc get sc ${EXPECTED_DEFAULT_STORAGECLASS}" || ret=$?
-  if [[ ! $ret -eq 0 ]]; then
-      echo "Storageclass ${EXPECTED_DEFAULT_STORAGECLASS} is not exist, skip set its clustercsidriver 'storageClassState' as 'Unmanaged' state."
-  else
-      provisioner=$(oc get sc/"${EXPECTED_DEFAULT_STORAGECLASS}" -ojsonpath='{.provisioner}')
-      run_command "oc get clustercsidriver ${provisioner}" || ret=$?
-      if [[ ! $ret -eq 0 ]]; then
-        echo "clustercsidriver/${provisioner} is not exist, skip set 'storageClassState' as 'Unmanaged' state."
-      else
-          oc patch clustercsidriver "${provisioner}" -p '[{"op":"replace","path":"/spec/storageClassState","value":"Unmanaged"}]'  --type json
-      fi  
-  fi
-}
-
-function set_expected_sc_as_default () {
-  ret=0
-  run_command "oc get sc ${EXPECTED_DEFAULT_STORAGECLASS}" || ret=$?
-  if [[ ! $ret -eq 0 ]]; then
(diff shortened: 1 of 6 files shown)


ci-operator/step-registry/storage/conf/storageclass/set-default-storageclass/storage-conf-storageclass-set-default-storageclass-commands.sh
```diff
@@ -0,0 +1,75 @@
...
+      echo "Storageclass ${EXPECTED_DEFAULT_STORAGECLASS} is not exist, skip setting."
+  else
+      oc annotate sc "${EXPECTED_DEFAULT_STORAGECLASS}" "storageclass.kubernetes.io/is-default-class=true" --overwrite
+  fi
+}
+
+set_proxy
+get_default_sc
+set_cluster_csi_driver_sc_state_unmanaged
+remove_all_default_sc_annotations
```
jsafrane, line 73: Why must be the operators marked as `Unmanaged`? They will not overwrite the value of `storageclass.kubernetes.io/is-default-class` annotation, they only add it back when it's missing. 

The whole script could be:
```
oc annotate storageclass.kubernetes.io/is-default-class=false --all --overwrite
oc annotate ${EXPECTED_DEFAULT_STORAGECLASS} storageclass.kubernetes.io/is-default-class=true --overwrite
```

ci-operator/step-registry/storage/conf/storageclass/set-default-storageclass/storage-conf-storageclass-set-default-storageclass-ref.yaml
```diff
@@ -0,0 +1,16 @@
...
+  from: cli
+  # inject oc binary
+  cli: latest
+  commands: storage-conf-storageclass-set-default-storageclass-commands.sh
+  resources:
+    requests:
+      cpu: 10m
+      memory: 100Mi
+  env:
+  - name: EXPECTED_DEFAULT_STORAGECLASS
```
jsafrane, line 12: Nit: `EXPECTED_DEFAULT_STORAGECLASS` sound like the script checks that the actual default SC matches the expected SC. `REQUIRED_DEFAULT_STORAGECLASS` / `NEW_DEFAULT_STORAGECLASS` would emphasize that the script modifies the default SC.

ci-operator/step-registry/storage/conf/storageclass/set-default-storageclass/storage-conf-storageclass-set-default-storageclass-ref.yaml
```diff
@@ -0,0 +1,16 @@
...
+    documentation: |-
+      Name of expecting default storageclass.
```
jsafrane, line 14: "Name of the storage class to set as the new default. The storage class must already exist."

ci-operator/step-registry/storage/conf/storageclass/set-default-storageclass/storage-conf-storageclass-set-default-storageclass-commands.sh
Phaow: Hmm, we have some auto test cases will delete the managed storageClasses, so after it execute maybe role back to the origin status. TBH, currently we don't have a better solution for this. I marked it as `Unmanaged ` so that we could add logic make the disruptive case skip executed. I think I could  also just add a label to the csidriver mark it as CI-preset to process the skip logic.

jsafrane: This script should just change the default SC, at least that's what the description says. If you want to do other things, like deleting the managed SC, you should create a different step that does just that.

Phaow: Agree, thanks for the suggestion!

ci-operator/step-registry/storage/conf/storageclass/set-default-storageclass/storage-conf-storageclass-set-default-storageclass-ref.yaml
Phaow: Addressed, thks!

ci-operator/step-registry/storage/conf/storageclass/set-default-storageclass/storage-conf-storageclass-set-default-storageclass-ref.yaml
Phaow: Addressed, thks!

ci-operator/step-registry/storage/conf/storageclass/set-default-storageclass/storage-conf-storageclass-set-default-storageclass-commands.sh
Phaow: Fixed.

PR #37114, opened 2023-03-09
gather second prometheus snapshot

tjungblu:
We currently only gather the first one, which makes it difficult to debug issues on prometheus restarts. This commit will save both snapshots, leaving the first one it encounters under the existing naming convention to avoid promeciuous from breaking.

### DIFF
--- a/ci-operator/step-registry/gather/extra/gather-extra-commands.sh
+++ b/ci-operator/step-registry/gather/extra/gather-extra-commands.sh
@@ -205,19 +205,25 @@ while IFS= read -r i; do
   FILTER=gzip queue ${ARTIFACT_DIR}/pods/${file}_previous.log.gz oc --insecure-skip-tls-verify logs --request-timeout=20s -p $i
 done < /tmp/containers
 
-# Snapshot the prometheus data from the replica that has the oldest
-# PVC. If persistent storage isn't enabled, it uses the last
-# prometheus instances by default to catch issues that occur when the
-# first prometheus pod upgrades.
-if [[ -n "$( oc --insecure-skip-tls-verify --request-timeout=20s get pvc -n openshift-monitoring -l app.kubernetes.io/name=prometheus --ignore-not-found )" ]]; then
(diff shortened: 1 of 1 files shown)


tjungblu:
I've added the old pvc logic back, not sure that was a remnant from the past, the 4.6/4.7 jobs couldn't find the prometheus pods.

tjungblu:
feeding https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_release/37114/rehearse-37114-pull-ci-openshift-ibm-vpc-block-csi-driver-operator-release-4.13-e2e-ibmcloud-csi/1633848064508694528 into promecieus seems to work fine. 

@dgoodwin I guess we need to also filter the snapshots in the gcsweb frontend?
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_release/37114/rehearse-37114-pull-ci-openshift-ibm-vpc-block-csi-driver-operator-release-4.13-e2e-ibmcloud-csi/1633848064508694528/artifacts/e2e-ibmcloud-csi/gather-extra/artifacts/metrics/

@vrutkovs I suppose that promecieus is still under your control at https://github.com/vrutkovs/promecieus ? :)

vrutkovs:
>I suppose that promecieus is still under your control

Yes, we could make another instance with breaking changes and gradually switch over. However I'm not sure if storing two snapshots will help much - this will simply grab disk state, which doesn't guarantee that either is consistent.

tjungblu:
thanks @vrutkovs, I think we can make this also work somewhat backward compatible with the existing tool. I was thinking to just supply a special query argument to the prow url which should indicate to load the other snapshot if available. UI can be a simple bootstrap toggle, happy to add that when we can merge this.

>  which doesn't guarantee that either is consistent.

That's fine for me, I just want to avoid looking at a blank period because the data might be in another snapshot. It's a bit of a pain right now :/

vrutkovs:
PromeCIeus _should_ be able to work with a direct URL to snapshot, that could be a good start to test it

tjungblu:
@vrutkovs added the logic to promecious too: https://github.com/vrutkovs/promecieus/pull/17

vrutkovs:
PromeCIus change deployed, 2nd snapshot switch seems to work for rehearsed jobs

/approve
/lgtm

PR #32453, opened 2022-09-20
[OCPQE-12211] Enable WINC QE CI workflow for AWS IPI.

jfrancoa:

PR #37125, opened 2023-03-09
Wait for WMCO deployment before creating Windows machineset.

jfrancoa:
The first action done in the ipi-conf-aws-windows-machineset-ref is checking that the windows-user-data exists. However, we need to wait until windows-machine-config-operator deployment is up and running.

Also, this patch parametrizes the AWS image to be used thinking on the moment when Windows 2022 image will be supported on AWS.

### DIFF
--- a/ci-operator/step-registry/ipi/conf/aws/windows-machineset/ipi-conf-aws-windows-machineset-commands.sh
+++ b/ci-operator/step-registry/ipi/conf/aws/windows-machineset/ipi-conf-aws-windows-machineset-commands.sh
@@ -4,6 +4,9 @@ set -o nounset
 set -o errexit
 set -o pipefail
 
+# Wait for WMCO to be up and running
+oc wait deployment windows-machine-config-operator -n openshift-windows-machine-config-operator --for condition=Available=True --timeout=5m
+
 # Ensure userDataSecret exist, fail otherwise. The userDataSecret is required and contains specific information to
 # customize the machine at first boot. For instance, the authorized public key for the SSH server to accept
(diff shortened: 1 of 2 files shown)


jfrancoa:
ping @sebsoto , could I get some review? I was observing that in some releases (4.10 mostly) when deploying the check for WMCO secret fails (oc -n openshift-machine-api get secret "${WINDOWS_USER_DATA_SECRET}") because WMCO is not up and running yet.

aravindhp:
@jfrancoa where do you see the issue that the PR is fixing? It will be nice if you call that out in the PR message.

saifshaikh48:
/lgtm

The commit does 2 separate things, adding a wait and introducing a new parameter. I'm fine with it since the changes are so small but consider splitting into 2 commits.

jfrancoa:
> @jfrancoa where do you see the issue that the PR is fixing? It will be nice if you call that out in the PR message.

I observed this when testing the [IPI AWS WMCO QE rehearsal ](https://github.com/openshift/release/pull/32453) which used this ref. It was occurring mostly on 4.10 and sometimes in 4.11. I am unable to find the logs for it now, but if you see the [previous ref](https://github.com/openshift/release/blob/master/ci-operator/step-registry/windows/conf/operator/windows-conf-operator-commands.sh), there is no wait at all for WMCO to be installed. So if the workflow gets to the [secret check](https://github.com/openshift/release/blob/master/ci-operator/step-registry/ipi/conf/aws/windows-machineset/ipi-conf-aws-windows-machineset-commands.sh#L10) quickly enough, the secret won't be there and the job will halt

PR #37128, opened 2023-03-09
Removing e2e-libvirt job from MAPI operator test suite

lobziik:
Cloud team is not using this job as CI signal,
as for 9.3.2023 this job is permafailing for quite a long time.

Currently there are no memories about this job purpose and origin within the team, and it seems reasonable to
stop running it on every PR.

cc @enxebre, maybe you remember why do we have it? :)


damdo (commented):
/lgtm

PR #37130, opened 2023-03-09
ShiftStack: Remove Bugzilla jobs

dulek:
All the bugs are migrated from Bugzilla into JIRA now, no need to keep Bugzilla automation in place. This commit removes it.


pierreprinetti:
/lgtm
Thank you!

PR #37131, opened 2023-03-09
update storage-conf-wait-for-csi-migration-commands.sh

elmiko:
this change makes it so that the wait for csi migrations will be skipped on ocp version 4.13+. it is being added to mitigate failures where there will be no migration.

we are seeing failures on https://github.com/openshift/cloud-provider-azure/pull/56 which indicate that the condition it is waiting for will never be present. instead of removing this workflow step, which might have unintended side effects, the change makes it so that the wait will be skipped on 4.13+.

### DIFF
--- a/ci-operator/step-registry/storage/conf/wait-for-csi-migration/storage-conf-wait-for-csi-migration-commands.sh
+++ b/ci-operator/step-registry/storage/conf/wait-for-csi-migration/storage-conf-wait-for-csi-migration-commands.sh
@@ -13,11 +13,14 @@ function detect_version() {
             MIGRATED_FEATURE_GATE="CSIMigrationAWS"
             MIGRATED_PLUGIN="kubernetes.io/aws-ebs"
             ;;
-        *)
-            # Use vSphere for 4.12 and newer (unknown) OCP releases, it will be GA as the last.
+        12 )
+            # Use vSphere for 4.12 OCP releases, it will be GA as the last.
             MIGRATED_FEATURE_GATE="CSIMigrationvSphere"
             MIGRATED_PLUGIN="kubernetes.io/vsphere-volume"
             ;;
+        *)
+            # Migration check is not needed on these versions
+            return 1
     esac
 }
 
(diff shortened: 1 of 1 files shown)


elmiko:
cc @lobziik @gnufied , i'm not sure this is the best solution, could use your advice. the failures i am seeing look like this: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cloud-provider-azure/56/pull-ci-openshift-cloud-provider-azure-release-4.13-e2e-azure-ovn-ccm/1633629283375124480

elmiko:
several of the failures seem related to this line

> Error: error writing risk analysis json artifact: open /logs/artifacts/junit/risk-analysis.json: no such file or directory

i don't think that is related to the change i made, unless there is some file generated by the wait for migration script?

elmiko:
so... i'm a little concerned that all the rehearsal failed, but it seems like the failures are not related to the change here. @gnufied @jsafrane wdyt?

gnufied:
I think I agree. I do not see how this change can cause those e2es failure. @jsafrane did you had any other thoughts?

jsafrane:
> Error: error writing risk analysis json artifact: open /logs/artifacts/junit/risk-analysis.json: no such file or directory

That's a symptom of a previous failure. This is [the interesting one](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_release/37131/rehearse-37131-pull-ci-openshift-origin-release-4.13-e2e-aws-csi-migration/1633881070132793344#1:build-log.txt%3A109):

> Skipping tests:
Error: no tests were found matching the TEST_SKIPS regex:
Managed cluster should start all core operators\| Alerts shouldn't report any alerts in firing or pending state\| Prometheus when installed on the cluster shouldn't report any alerts in firing state apart from Watchdog and AlertmanagerReceiversNotConfigured

TEST_SKIPS must match at least one test, otherwise the whole CI job fails >:-o. But this was broken even before this PR, so:
/pj-rehearse ack

PR #37165, opened 2023-03-10
MGDAPI-5364 - feat: use 4.12 clusters for CRO and integreatly branches

KevFan:
* Remove obsolete CRO and integreatly branches
* Use 4.12 clusters for CRO and integreatly branches


KevFan:
Rebased against master to resolve conflicts

PR #37171, opened 2023-03-10
ComplianceAsCode/content: Remove the run_if_changed stanzas to make the tests really opt-in

jhrozek:
I thought (wrong) that the run_if_changed would only be evaluated after
an explicit command to start the test run. But it seems that the CI
tests are always executed when the run_if_changed matches any files
changed in the PR.

Following an advice I got on Slack, let's get rid of those lines.


rhmdnd (approved):
/lgtm

PR #37191, opened 2023-03-10
SPLAT-980: introduce vSphere IBM 7 dedicated lease slice

rvanderp3:
Specific vSphere jobs must run in the IBM vSphere 7 environment. This cluster profile will allow those jobs to be specifically scheduled in that environment.

depends on https://github.com/openshift/ci-tools/pull/3322


rvanderp3:
/hold 

hold until https://github.com/openshift/ci-tools/pull/3322 merges

PR #37201, opened 2023-03-12
PowerVS fail destroy_resources if destroy fails

hamzy:
During destroy_resources, we call openshift-install destroy cluster three times.  If all three of those fail, then we should not create a cluster in this zone.

### DIFF
--- a/ci-operator/step-registry/ipi/install/powervs/install/ipi-install-powervs-install-commands.sh
+++ b/ci-operator/step-registry/ipi/install/powervs/install/ipi-install-powervs-install-commands.sh
@@ -438,6 +438,7 @@ EOF
   #
   # Call destroy cluster on fake metadata file
   #
+  DESTROY_SUCCEEDED=false
   for i in {1..3}; do
     echo "Destroying cluster $i attempt..."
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/ipi/install/powervs/install/ipi-install-powervs-install-commands.sh
```diff
@@ -461,6 +463,12 @@ EOF
       ibmcloud pi network-delete ${UUID}
     done
   ) < <(ibmcloud pi networks --json | jq -r '.networks[] | select(.name|test("rdr-multiarch-'${POWERVS_ZONE}'")) | .networkID')
+
+  if ! ${DESTROY_SUCCEEDED}
+  then
+    echo "Failed to destroy cluster failed after three attempts. Above debug output points to a PowerVS problem."
```
tvardema, line 469: Should be fine with the following line 469 replacement:
echo "Failed to destroy cluster after three attempts."

deepsm007:
/lgtm
/hold
Remove hold when ready to merge

PR #37203, opened 2023-03-13
Fix cron entries to exclude 31th day of the month

liangxia:
We use `DAY_OF_MONTH=${NUMBERS:4:2}%31+1` to hash the jobs to spreed across the month from the 1st till 31th.
But it's buggy as not every month has 31th day. So fix that to `DAY_OF_MONTH=${NUMBERS:4:2}%30+1` to use 1st till 30th.

/cc @jianlinliu @jhou1 @JianLi-RH @dis016 @pruan-rht


liangxia:
Rebase and resolve conflicts.

PR #37217, opened 2023-03-13
Remove AWS tech-preview jobs for CCM related repos

lobziik:
AWS cloud-controller-manager got to GA and installs by default now.

xref:
- https://github.com/openshift/library-go/pull/1485


elmiko (commented):
/lgtm

PR #37219, opened 2023-03-13
Remove Azure tech-preview jobs for CCM related repos

lobziik:
Azure cloud-controller-manager got to GA and installs by default now.

xref:

- https://github.com/openshift/library-go/pull/1484


lobziik:
/hold

https://github.com/openshift/machine-config-operator/pull/3589 - this need to be merged first

PR #37244, opened 2023-03-14
Enable destructive tests in upgrade

liangxia:
/cc @jhou1 @JianLi-RH @dis016 @pruan-rht


JianLi-RH:
Will it break normal tests?

liangxia:
This has been replaced by https://github.com/openshift/release/pull/37261

PR #37270, opened 2023-03-14
[HCO] make externally built FBC index image available

tiraboschi:
Add externally built FBC based index image
quay.io/kubevirt/hyperconverged-cluster-index:1.10.0-fbc-unstable to enable the full transition to FBC based index images for upgrade tests.


tiraboschi:
Done, sorry for the noise

PR #37280, opened 2023-03-14
USHIFT-958: Add 4.14  to testgrid

pacevedom:


bradmwilliams (commented):
/approve

smg247:
@pacevedom Test Platform's automated tooling is attempting to remove this configuration due to `4.14` being the current release. You need to be sure to put these new tests in the `openshift-microshift-main.yaml` file as well so they do not get removed. The only reason these haven't been auto-removed already is that the `release-controller` configuration you have added here would need to be removed as well to allow the tests to pass, and that isn't something that we automate. https://github.com/openshift/release/pull/37521 is the PR, for reference.

PR #37295, opened 2023-03-14
Add e2e-vsphere-ipi-ovn-migration-sdn

weliang1:
This PR is for https://issues.redhat.com/browse/NP-639: Support for ovn-kubernetes to openshift-sdn CNI Migration

Comparing sdn2ovn migration, we do not think many customers will do the ovn2sdn migration, so one workflow testing should be enough.

The reason to pick vsphere cluster is because most customer OCP clusters are running in vsphere.

@zhaozhanqi PTAL

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__amd64-nightly.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__amd64-nightly.yaml
@@ -2845,6 +2845,18 @@ tests:
     test:
     - chain: openshift-e2e-test-qe
     workflow: cucushift-installer-rehearse-vsphere-ipi-sdn
+- as: vsphere-ipi-ovn-migration-sdn-p2-f7
+  cron: 58 4 6,14,22,29 * *
+  steps:
+    cluster_profile: vsphere
+    env:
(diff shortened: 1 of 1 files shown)


weliang1:
@liangxia PTAL!

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.13__amd64-nightly.yaml
```diff
@@ -2845,6 +2845,18 @@ tests:
     test:
     - chain: openshift-e2e-test-qe
     workflow: cucushift-installer-rehearse-vsphere-ipi-sdn
+- as: vsphere-ipi-ovn-migration-sdn-p2-f7
```
liangxia, line 2848: Would you mind to sort the tests based on the name, also add it to 4.14, too

weliang1: @liangxia I though I insert file based on the names as below? Please advise, thanks! 
- as: vsphere-ipi-sdn-p2-f7
- as: vsphere-ipi-ovn-migration-sdn-p2-f7
- as: vsphere-ipi-zones-p2-f7

liangxia: IMO, vsphere-ipi-ovn should sort before vsphere-ipi-sdn

PR #37327, opened 2023-03-15
update rosa-hypershift Test Frequency

LiangquanLi930:


LiangquanLi930:

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-nightly.yaml
```diff
@@ -1152,35 +1152,35 @@ tests:
     test:
     - chain: openshift-e2e-test-qe
     workflow: osd-ccs-aws
-- as: aws-rosa-sts-hypershift-mgmt-p2-f2
-  cron: 28 17 2,4,5,8,10,12,14,16,18,20,22,24,26,28,30 * *
+- as: aws-rosa-sts-hypershift-guest-p1-f7
+  cron: 53 8 1,3,5,7,9,11,13,14,17,19,21,23,25,27,29 * *
```
liangxia, line 1156: This does not looks like a correct cron entry for f7, is it ?

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-stable.yaml
```diff
@@ -810,35 +810,35 @@ tests:
     test:
     - chain: openshift-e2e-test-qe-destructive
     workflow: cucushift-installer-rehearse-aws-ipi-workers-marketplace
-- as: aws-rosa-sts-hypershift-mgmt-p2-f7
-  cron: 14 3 7,14,22,30 * *
+- as: aws-rosa-sts-hypershift-guest-p1-f2
+  cron: 34 13 1,3,5,7,9,11,12,15,17,19,21,23,25,27,29 * *
```
liangxia, line 814: We have a new 4.12 stable payload every week. Using `f2` means we want to test on the same build 3 or 4 times, is it reasonable ? /cc @YuLi517

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-nightly.yaml
LiangquanLi930: Good catch! Thanks a lot!

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-stable.yaml
YuLi517: Following service delivery rules, customer will use stable OCP payload only. This is the reason we reduce frequency of nightly pipeline but increase stable one. Why we still keep nightly one? As it will include all the new features.
This f2 job is our only one ROSA-hypershift GA pipeline. Does QE have confidence to say Green on GA. We count on this job. It will run on Service delivery staging env. In future, we may add one more pipeline with f2 to run on production also.

YuLi517: @liangxia ^^

liangxia: /cc @jianlinliu to ack

jianlinliu: got it. thx.

PR #37336, opened 2023-03-15
Fix reserve step unbound variables

aleskandro:
This PR fixes a bug in the updated reserve step, when the ADDITIONAL_WORKERS variables related to the multi-arch compute nodes enablement are set as empty.

/cc @bmanzari


aleskandro:
Locally tested change with:

```
└ $ cat test.sh 
set -o nounset
MY_VAR=${2}
└ $ bash test.sh 1
test.sh: line 3: 2: unbound variable
└ $ vim test.sh 
└ $ cat test.sh 
set -o nounset
MY_VAR=${2:-}
└ $ bash test.sh 1
└ $ echo $?
0

```

PR #37353, opened 2023-03-15
Add OpenShift system pods disruption scenario

chaitanyaenr:
This commit:
- Adds a scenario to pick OpenShift system pods in openshift-* namespaces and disrupt them to check it's recovery.
- Renames the pod-scenarios workload to etcd-disruption in order to match the behavior of the run.


ci-operator/step-registry/redhat-chaos/random-system-pods-disruption/redhat-chaos-random-system-pods-disruption-command.sh
```diff
@@ -0,0 +1,21 @@
...
+pushd /tmp
+
+ls -la /root/kraken
+git clone https://github.com/redhat-chaos/krkn-hub.git
+pushd krkn-hub/
+
+echo "kubeconfig loc $$KUBECONFIG"
+
+export KRKN_KUBE_CONFIG=$KUBECONFIG
+export NAMESPACE="openshift-.*"
```
paigerube14, line 19: I think there's a way to export an enviornment variable at a higher level so we could keep the scenarios at "pod-scenarios" but at the workflow level of set the variables we want to run. This way if one thing changes we dont need to rewrite every file

paigerube14: documentation: https://docs.ci.openshift.org/docs/architecture/ci-operator/#dependency-overrides

chaitanyaenr:
closing in favor of https://github.com/openshift/release/pull/37359

PR #37380, opened 2023-03-16
Manual testing workflow for the agent team

aleskandro:
This PR adds a workflow in the step-registry to be used for manual testing. It will execute only the pre-provisioning steps to configure the network and leave the users able to execute the configuration and the installation of the clusters on a pre-configured and provisioned baremetal network infrastructure.

Example test config:
```
- as: baremetal-agent-ipv4-dhcp-p1-f14
  cluster: build05
  cron: 30 21 7,21 * *
  steps:
    cluster_profile: equinix-ocp-metal-qe
    dependencies:
      OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE: release:arm64-latest # if you need to mirror the arm64 release for disconnected cases. Blocked by #37221
    env:
      architecture: arm64
      AUX_HOST: openshift-qe-bastion.arm.eng.rdu2.redhat.com
      RESERVE_BOOTSTRAP: "false"
      CLUSTER_DURATION: 12h # The amount of time the network and nodes should stay provisioned 
      DISCONNECTED: "false" # set to true for disconnected scenarios. It depends on #37221
      masters: "3"
      workers: "2"
    workflow: baremetal-lab-agent-manual
```


zniu1011:
lgtm

LiangquanLi930:
@aleskandro Hi we can use this step `cucushift-hypershift-extended-debug`, example: https://github.com/openshift/release/pull/37307/files

aleskandro:
@LiangquanLi930 the solution is still temporary. I'd propose we might do a more generic (with respect to the organization) step that we can then use for all. Or we should catch whether the clusterbot-wait one can get an increased value for the timeout.

PR #37404, opened 2023-03-16
CORS-2588: moving to v5 node type

tjungblu:
The v5 generation has much better sequential IOPS performance, fsync latency and faster CPUs at the same price as v3.

### DIFF
--- a/ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh
+++ b/ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh
@@ -40,7 +40,7 @@ elif [[ "${SIZE_VARIANT}" == "compact" ]]; then
 fi
 if [ -n "${master_type_prefix}" ]; then
   if [ "${OCP_ARCH}" = "amd64" ]; then
-    master_type=${master_type_prefix}s_v3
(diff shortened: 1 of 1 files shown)


tjungblu:
/pj-rehearse pull-ci-openshift-cloud-credential-operator-release-4.14-e2e-azure-upgrade

sdodson:
I don't mind us moving forward with this change, however I would prefer that we not link OCPBUGS-8426 against this change. Something changed in 4.13 + 9.2 and we should strive to understand what that is. It may very well be https://issues.redhat.com/browse/OCPBUGS-10353 that's root cause for OCPBUGS-8426.

I've put in https://issues.redhat.com/browse/CORS-2588 for installer to prefer V5 instances over V3, though we can't make that the default everywhere because certain regions lack V5 instances, tho almost all have them.

sdodson:
/pj-rehearse pull-ci-openshift-cloud-credential-operator-release-4.14-e2e-azure-upgrade

dgoodwin:
I don't think this change helped, looking at the rehersal, I still see the same pattern, lots of disruption, and etcd slowfdata sync as soon as the masters reboot.

sdodson:
/close
Lets let the installer default move this.

PR #37406, opened 2023-03-16
Integration service: Remove branch protection and tide

dirgim:
* Set the branch protection to unmanaged
* Remove the tide settings since we don't want to use it at this point


psturc (approved):

jsztuka (approved):
lgtm

PR #37415, opened 2023-03-16
Update existing 4.12 bug and add a new skiplist for 4.13

mgourin:
Old bug (https://issues.redhat.com/browse/OCPBUGS-4194) resolution caused a new bug (https://issues.redhat.com/browse/OCPBUGS-10424) for both 4.12/4.13;

As part of this resolution, the test name was changed, but only for 4.13.
4.12 test name: "Should have the correct RPS configuration"
4.13 test name: "Check RPS Mask is applied to atleast one single rx queue on all veth interface"

### DIFF
--- a/ci-operator/step-registry/telco5g/cnf/tests/telco5g-cnf-tests-commands.sh
+++ b/ci-operator/step-registry/telco5g/cnf/tests/telco5g-cnf-tests-commands.sh
@@ -40,7 +40,7 @@ cat <<EOF >>"${SKIP_TESTS_FILE}"
 # <feature> <test name>
 
 # SKIPTEST
-# bz### https://issues.redhat.com/browse/OCPBUGS-4194
+# bz### https://issues.redhat.com/browse/OCPBUGS-10424
 # TESTNAME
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/telco5g/cnf/tests/telco5g-cnf-tests-commands.sh
```diff
@@ -49,6 +49,17 @@ EOF
 
 function create_tests_temp_skip_list_13 {
     create_tests_temp_skip_list_12
```
sshnaidm, line 51: I think you need to remove this line then? Let's keep 12 and 13 sip lists separated from now.

sshnaidm (changes requested):

ci-operator/step-registry/telco5g/cnf/tests/telco5g-cnf-tests-commands.sh
mgourin: Removed

sshnaidm (approved):

PR #37416, opened 2023-03-16
set latest tag to the main branch postsubmit build

harshad16:
set latest tag to the main branch postsubmit build

The data-science-pipelines-operator repo would like a `latest` tag , as a mirror to the main tag.


harshad16:
/hold

for discussions

PR #37420, opened 2023-03-16
ROX-15287: fix ocp 4.6 for release-3.73 branch

BradLugo:
/cc @gavin-stackrox 
/assign @BradLugo 
/pj-rehearse auto-ack


gavin-stackrox:
/lgtm
the rehearsal failures could use some more examination but this is a step in the right direction to get rid of the eks placeholders.

PR #37428, opened 2023-03-17
KMM: do not build index images

qbarrand:
Use the new CI Operator feature introduced in openshift/ci-tools#3327 cc @hongkailiu.

/cc @yevgeny-shnaidman @ybettan


ybettan:
SGTM.

CI is failing though.

PR #37433, opened 2023-03-17
odf-console: map CI runner image

alfonsomthd:


alfonsomthd:
@smg247 Here I'm only mapping an external image, can you please merge this?

PR #37436, opened 2023-03-17
odf-console: get root image from repository

alfonsomthd:


alfonsomthd:
@SanjalKatiyar @bipuladh All tests passed: ready to merge.

PR #37441, opened 2023-03-17
set CPU more inline with actual used amount

smg247:
We created a new alert which tells us when memory or CPU actual usage is over 10x greater than the configured amount. These jobs are consistently being flagged. In practice, it does not matter as the `pod-scaler` will increase these requests to the actual used amount, but this new alert helps us to detect memory leaks. We want to update these so they stop firing the alert and creating noise for us.


smg247:
Screenshot 2023-03-17 at 9 52 44 AM

ncdc (approved):

PR #37461, opened 2023-03-20
add operatorhub checking

jianzhangbjz:
The [error below](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/periodic-ci-redhat-openshift-ecosystem-cvp-ocp-4.12-cvp-common-claim/1636222755068186624/artifacts/cvp-common-claim/optional-operators-subscribe/artifacts/sub-oo-klsjd.yaml) means the `default` catalog source not working well, it will block all subscriptions. This PR will check if the default catalog source are disabled correctely.
```yaml
  - message: 'failed to populate resolver cache from source community-operators/openshift-marketplace:
      failed to list bundles: rpc error: code = Unavailable desc = connection error:
      desc = "transport: Error while dialing dial tcp: lookup community-operators.openshift-marketplace.svc
      on 172.30.0.10:53: no such host"'
    reason: ErrorPreventedResolution
    status: "True"
    type: ResolutionFailed
```

### DIFF
--- a/ci-operator/step-registry/optional-operators/subscribe/optional-operators-subscribe-commands.sh
+++ b/ci-operator/step-registry/optional-operators/subscribe/optional-operators-subscribe-commands.sh
@@ -15,6 +15,13 @@ then
 	source "${SHARED_DIR}/proxy-conf.sh"
 fi
 
+# check the global catalog source config, if you don't want to use the default catalog sources, you should disable it by using the `operatorhub`. Like below,
+# oc patch operatorhub cluster -p '{"spec": {"disableAllDefaultSources": true}}' --type=merge
+oc get operatorhub cluster -o yaml
+# if one of global catalog sources is not running well, it will block all subscribption. 
+oc get catalogsource -n openshift-marketplace
+oc get pods -n openshift-marketplace
+
 # In upgrade tests, the subscribe step installs the initial version of the operator, so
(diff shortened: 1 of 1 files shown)


jianzhangbjz:
Hi @dirgim @samvarankashyap could you help approve it when you get a chance? Thanks!

maneeshmehra:
All the default catalog sources should have been disabled within the ephemeral cluster that is provisioned to test the operator-under-test. That should've been done against the cluster where this script also runs as part of the following script the runs prior to this one: https://github.com/openshift/release/blob/master/ci-operator/step-registry/optional-operators/configure-proxy-registry/optional-operators-configure-proxy-registry-commands.sh#L41. It would still be good to confirm that is the case before proceeding to install the operator-under-test on the ephemeral cluster.

maneeshmehra:
@jianzhangbjz: I understood your changes and which is why I am suggesting that should we not put those checks in the same script that actually disables the default catalog sources ? The script I suggested always runs prior to this one as per our workflow: https://github.com/openshift/release/blob/master/ci-operator/step-registry/optional-operators/cvp-common/claim/optional-operators-cvp-common-claim-workflow.yaml#L6

maneeshmehra:
@samvarankashyap @nmars @asergienk: Could you please review this change and provide your sing-offs on it ?

PR #37499, opened 2023-03-20
[wip] DNM - Add periodic

jwendell:


yxun:
Do we need a plugin config folder in https://github.com/openshift/release/tree/master/core-services/prow/02_config/maistra ?

I haven't seen test-infra prow config before.

jwendell:
> Do we need a plugin config folder in https://github.com/openshift/release/tree/master/core-services/prow/02_config/maistra ?
> 
> I haven't seen test-infra prow config before.

I don't think so, at least for now. This is a periodic job, not triggered by any github command.

PR #37501, opened 2023-03-20
stolostron: update AppLC & CLC repos to golang 1.20

lennysgarage:
Addresses:
 - https://issues.redhat.com/browse/ACM-4306
 - https://issues.redhat.com/browse/ACM-4224

### DIFF
--- a/ci-operator/config/stolostron/cluster-curator-controller/stolostron-cluster-curator-controller-backplane-2.3.yaml
+++ b/ci-operator/config/stolostron/cluster-curator-controller/stolostron-cluster-curator-controller-backplane-2.3.yaml
@@ -3,17 +3,21 @@ base_images:
     name: builder
     namespace: stolostron
     tag: go1.18-linux
+  stolostron_builder_go1.19-linux:
+    name: builder
+    namespace: stolostron
+    tag: go1.19-linux
 build_root:
   image_stream_tag:
     name: builder
     namespace: stolostron
-    tag: go1.18-linux
+    tag: go1.19-linux
 images:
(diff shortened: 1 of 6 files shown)


mikeshng (approved):
/approve

/lgtm

ci-operator/config/stolostron/cluster-curator-controller/stolostron-cluster-curator-controller-backplane-2.3.yaml
```diff
@@ -11,7 +11,7 @@ build_root:
 images:
 - dockerfile_path: Dockerfile.prow
   inputs:
-    stolostron_builder_go1.20-linux:
+    stolostron_builder_go1.18-linux:
       as:
```
chenz4027: Will auto update upgrade it to 1.19 automatically?

chenz4027: Sounds good!

lennysgarage: I'd rather actually update it here, just in case it doesn't, so we don't have to open another PR. The auto upgrade should be able to ignore the change if otherwise.

chenz4027: Better to be safe...

mikeshng (approved):
/approve

/lgtm

PR #37528, opened 2023-03-21
USHIFT-984: Enable presubmits for etcd/.* changes

pacevedom:


ci-operator/config/openshift/microshift/openshift-microshift-main.yaml
```diff
@@ -145,15 +145,15 @@ tests:
   run_if_changed: ^(rpm/*|packaging/.*|Makefile)
 - as: e2e-openshift-conformance-reduced
   cluster: build02
-  run_if_changed: ^(assets/.*|cmd/.*|pkg/.*|packaging/.*|vendor/.*|go\.sum|go\.mod|Makefile|.*\.sh)
+  run_if_changed: ^(assets/.*|cmd/.*|etcd/.*|pkg/.*|packaging/.*|vendor/.*|go\.sum|go\.mod|Makefile|.*\.sh)
```
dhellmann, line 148: Would it make sense to use a pattern like `.*\.go` like we do for shell scripts? 

I see we aren't testing on changes to `validate-microshift`, either. I wonder if we change any of the files in other subdirectories often enough that we need this at all?

dhellmann (commented):
This is an improvement, but maybe we can go further in another PR.

/lgtm

ci-operator/config/openshift/microshift/openshift-microshift-main.yaml
pacevedom: We might need to rework this regex a little bit, will do it in a separate PR.

PR #37541, opened 2023-03-21
Branchprotector: unmanaged: true

svor:
Set branch protector unmanaged: true for all branches in eclipse-che/che-plugin-registry repository.


svor:
@smg247 could you take a look please?

PR #37545, opened 2023-03-22
4.9 or less doesn't support IPI on Alibabacloud

jianli-wei:
Remove the periodic job, because 4.9 or less doesn't support IaaS `Alibabacloud`, e.g. one failed [job](https://qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/qe-private-deck/logs/periodic-ci-openshift-openshift-tests-private-release-4.10-amd64-nightly-4.10-upgrade-from-stable-4.9-alibaba-ipi-fips-p2-f14/1637913791771447296), thanks!


jianlinliu:
/lgtm.

Actually because we never announce alibaba get GAed, so I think we can drop all the aliabab upgrade ci jobs.

PR #37559, opened 2023-03-22
Adds pre commit job for cinder

frenzyfriday:
pre commit job is being added to cinder operator repo in : https://github.com/openstack-k8s-operators/cinder-operator/pull/147


marios (commented):
looks correct but we need to wait for https://github.com/openstack-k8s-operators/cinder-operator/pull/147 before we can test it here

PR #37572, opened 2023-03-22
use better name for OS image env var

cheesesashimi:
Use a less generic name for the OS image URL env var to avoid possible collisions.


jkyros:
This should avoid collisions, so I think we're good. 

I _was_ thinking -- should we have it contain 'override' or something because a naive observer might wonder "is this something the MCO needs to live, or is it special?" but nah, I don't want to agonize over it. I think this is good. 

THE_IMAGE_ZACK_IS_FEEDING_IN_TO_THE_E2E_LAYERING_TEST_THAT_EVENTUALLY_GETS_USED_TO_OVERRIDE_OS_IMAGE_URL :smile:

/lgtm

PR #37576, opened 2023-03-22
Add Notebook controller tests downstream

VaishnaviHire:


VaishnaviHire:
/cc @LaVLaS 


The unit test failure is expected, since we need to pull the recent fix downstream

PR #37585, opened 2023-03-22
hypershift: cleanup after 4.14 branch

sjenning:
cleaning up a few things

* BUG `main` presubs are currently testing y-stream upgrades from 4.13 to 4.14 rather than z-stream upgrades within 4.14
* all periodic jobs should be associated with `release-4.*` branches, not `main`
* after periodic jobs are associated with `release-4.*` branch, the version is `periodic-4.1*` is redundant; just use `periodic`
* periodic test names should not contain `periodic` in the name as it will be redundant with the `variant`

NOTE: This will wreck our testgrid dashboard and it will need an update
https://k8s-testgrid.appspot.com/redhat-hypershift


sjenning:
Pulled the presub bug fix into it's own PR https://github.com/openshift/release/pull/37635

Need to reconcile this with CRT effort to do the same thing https://github.com/openshift/release/pull/37630

sjenning:
closing in favor of https://github.com/openshift/release/pull/37630
/close

PR #37529, opened 2023-03-21
Adding ACM steps to the step-registry

ascerra:
## Purpose
To create re-usable steps in the step registry for the ACM interop scenario but also for other use cases (OPP, OCP console dev, ..etc).

## Testing
All of the steps being added here have been tested as a part of the [ACM onboarding scenario](https://github.com/openshift/release/pull/36326/files).
The Prow job with the passing results is [here ](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_release/36326/rehearse-36326-periodic-ci-stolostron-acmqe-autotest-main-acm-ocp4.12-lp-interop-acm-interop-aws/1638018424292511744).
All artifacts are [here](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_release/36326/rehearse-36326-periodic-ci-stolostron-acmqe-autotest-main-acm-ocp4.12-lp-interop-acm-interop-aws/1638018424292511744/)

PR #37593, opened 2023-03-23
updates, mch bug fix, extend fetch ref

ascerra:
Extension of PR https://github.com/openshift/release/pull/37529

It was merged before I could address some things and I also discovered a bug (missing default var for MCH_NAMESPACE) which is fixed in this PR

### DIFF
--- a/ci-operator/step-registry/acm/fetch-managed-clusters/acm-fetch-managed-clusters-commands.sh
+++ b/ci-operator/step-registry/acm/fetch-managed-clusters/acm-fetch-managed-clusters-commands.sh
@@ -16,7 +16,11 @@ export HUB_OCP_PASSWORD
 
 if [ -s /tmp/ci/managed.cluster.name ] && ! grep -q "null" /tmp/ci/managed.cluster.name; then
     echo "managed.cluster.name file was found saving as artifact."
-    cp -r /tmp/ci/managed.cluster.name $ARTIFACT_DIR/
+    cp -r /tmp/ci/managed.cluster.name $SHARED_DIR/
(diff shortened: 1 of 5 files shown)


ci-operator/step-registry/acm/fetch-managed-clusters/acm-fetch-managed-clusters-commands.sh
```diff
@@ -16,7 +16,11 @@ export HUB_OCP_PASSWORD
 
 if [ -s /tmp/ci/managed.cluster.name ] && ! grep -q "null" /tmp/ci/managed.cluster.name; then
     echo "managed.cluster.name file was found saving as artifact."
-    cp -r /tmp/ci/managed.cluster.name $ARTIFACT_DIR/
+    cp -r /tmp/ci/managed.cluster.name $SHARED_DIR/
```
calebevans, line 19: Not really urgent, you couple probably just do `cp -r /temp/ci/managed.cluster.* $SHARED_DIR/`, unless it copies a bunch of unneeded files.

calebevans (commented):
Just a little thing, if you don't want to do that we can push this through. It's good as is.

ci-operator/step-registry/acm/fetch-managed-clusters/acm-fetch-managed-clusters-commands.sh
calebevans: We've decided to push this pr through without this change. May make the change in the future.

calebevans (approved):
/lgtm

PR #37599, opened 2023-03-23
Report failure of image mirroring in serverless-dev channel

nak3:
part of https://issues.redhat.com/browse/SRVKS-1043

Since nobody takes care about mirroring failure, we hit unexpected issue. This patch adds the slack notification.

### DIFF
--- a/ci-operator/jobs/infra-image-mirroring.yaml
+++ b/ci-operator/jobs/infra-image-mirroring.yaml
@@ -1,7 +1,7 @@
 periodics:
 - agent: kubernetes
   cluster: app.ci
-  cron: '@daily'
+  cron: '@hourly'
   decorate: true
   decoration_config:
     timeout: 5h0m0s


ci-operator/jobs/infra-image-mirroring.yaml
```diff
@@ -1,12 +1,18 @@
 periodics:
 - agent: kubernetes
   cluster: app.ci
-  cron: '@hourly'
```
nak3, line 4: This is mysterious :thinking: 
The logs is not hourly but just every 6 hours https://prow.ci.openshift.org/job-history/gs/origin-ci-test/logs/periodic-image-mirroring-knative
But anyway I changes it to daily as every 6 hours or hourly is too frequent.

ReToCode: Hm might be because it runs for 5h (which seems to be the max), then its started after an hour, which would result in 6h intervals.
Nevertheless, this job will fail as long as we either:
- Can create more jobs to split eventing, eventig-kafka and serving
- Or cleanup all tag images from eventing on the internal registry (which is the root cause of the long delay). Not sure if we can do this, though.
But adding reporting is great, thank you.

ReToCode (approved):
/lgtm

ci-operator/jobs/infra-image-mirroring.yaml
ReToCode: @nak3 let's keep `@hourly` here, with the changes to the mirror config coming up, the job should be back to run in minutes.

nak3:
@ReToCode Yeah, I was about to reach out. But can we fix https://prow.ci.openshift.org/job-history/gs/origin-ci-test/logs/periodic-image-mirroring-knative?buildId= soon?
This PR will make a lot of notification (every several hours) in serverless-dev channel so I just worry that the too much notification in the slack...

ReToCode:
Should be fixed when we merge https://github.com/openshift-knative/hack/pull/43/files#diff-a61d7aa0cc903eddb3ac82d139a813d97847a03d2d7ca5b4b5d350a181ddc703R39 and update the Eventing file with it.

ReToCode:
We just updated our mirroring-config, so we should be fine with the job and would like to have a notification if it fails.

Choosing a few random approvers from the OWNERS file, is that change ok for you?
@derekwaynecarr, @imcleod, @sdodson

PR #37610, opened 2023-03-23
add-rosa-aws-hypershift

ropatil010:
Hi @liangxia , 

Can you PTAL. 
PR about/End Goal:
  Need to create cluster for profile:  periodic-ci-openshift-openshift-tests-private-release-4.12-amd64-stable-aws-rosa-sts-hypershift-mgmt-p1-f4 from clusterbot. 

 The above mentioned profile is in: https://github.com/openshift/release/blob/master/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-stable.yaml#L828-L842
 
 Is this much enough or any corrections needs to be done? 
 
 /assign @liangxia

### DIFF
--- a/core-services/ci-chat-bot/workflows-config.yaml
+++ b/core-services/ci-chat-bot/workflows-config.yaml
@@ -70,7 +70,7 @@ workflows:
     platform: aws
   cucushift-installer-rehearse-aws-ipi-ovn-sdn2ovn:
     platform: aws
-  rosa-aws-sts-hypershift:
+  cucushift-installer-rehearse-aws-rosa-sts-hypershift:
     platform: aws
   cucushift-installer-rehearse-aws-ipi-ovn-winc:
     platform: aws


core-services/ci-chat-bot/workflows-config.yaml
```diff
@@ -70,6 +70,8 @@ workflows:
     platform: aws
   cucushift-installer-rehearse-aws-ipi-ovn-sdn2ovn:
     platform: aws
+  rosa-aws-sts-hypershift:
```
liangxia, line 73: Would you mind to sort it ?

core-services/ci-chat-bot/workflows-config.yaml
```diff
@@ -70,6 +70,8 @@ workflows:
     platform: aws
   cucushift-installer-rehearse-aws-ipi-ovn-sdn2ovn:
     platform: aws
+  cucushift-installer-rehearse-aws-rosa-sts-hypershift:
```
liangxia, line 73: Where is the workflow defined ?

PR #37615, opened 2023-03-23
Gather etcd debug artifacts if they exist

dgoodwin:
Part of on-going efforts to debug some performance related problems


dgoodwin:
/hold

This won't work, only can grab in /var/log, and the etcd pod can't write there as it stands today. Need to use oc cp to each etcd pod.

PR #37619, opened 2023-03-23
Configure openstack-ironic-lib repository

elfosardo:
Add the repository to OCP lifecycle
Add Ci jobs
Add branch automation and fast-forward for current dev branch


elfosardo:
going to check 4.10 and 4.11 failures but we're leaving 4.7 as it is for the time being

PR #37639, opened 2023-03-24
Add encrypt cron ci job

sgaoshang:


ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.13__amd64-nightly.yaml
```diff
@@ -2915,6 +2915,16 @@ tests:
     test:
     - chain: openshift-e2e-test-qe
     workflow: cucushift-installer-rehearse-vsphere-upi-disconnected-secureboot
+- as: vsphere-upi-encrypt-p2-f14
+  cron: 49 11 15,30 * *
+  steps:
+    cluster_profile: vsphere-ibm-7
+    env:
+      E2E_RUN_TAGS: '@vsphere-upi and @network-ovnkubernetes and not @fips'
+      TAG_VERSION: '@4.13'
```
jianlinliu, line 2924: I guess we also need the following setting like other tests.
```
TEST_FILTERS: ~ChkUpgrade&;~NonPreRelease&;~Serial&;~Disruptive&;~ConnectedOnly&;~HyperShiftMGMT&;~MicroShiftOnly&
```

sgaoshang: I see some job added this filter, some job not, is there some rule with it?

jianlinliu: Do you mean `vsphere-upi-platform-external-p2-f14` ?  I guess the author forgot to add the filter.

jianlinliu: Hmm, indeed, actually I am not sure the rule there, but seem like only disconnected job add the filter. 

I guess we can drop it again, because the workflow is not disconnected, sorry to mislead you.

jianlinliu:
cc @liangxia to review

PR #37658, opened 2023-03-24
Fix issue DPTP-3337 Use instance storage for building

bear-redhat:
Add a new MachineConfig on worker node to utilise instance-storage for faster building.


droslean (approved):

bear-redhat:
Some instances unexpectedly have `nvme1` device while no instance-storage/secondary EBS volume. Reverting for now.

PR #37664, opened 2023-03-24
test pod with persistent storage

copejon:
Alters the MicroShift reboot tests to deploy a pod with a pvc rather than a deployment.  This better reflects baseline customer envs with persistent storage.

[USHIFT-1020](https://issues.redhat.com//browse/USHIFT-1020)


copejon:
The mass failure is a result of SSL certs being expired on kuttl's releases.

pacevedom (approved):
/lgtm

PR #37670, opened 2023-03-24
Add ubi9:ruby-31 to Ci

jcantrill:
cc @xperimental


ci-operator/config/ViaQ/logging-fluentd/ViaQ-logging-fluentd-ruby31-v1.14.x.yaml
```diff
@@ -86,7 +86,7 @@ base_images:
   ubi8-ruby31:
```
xperimental, line 86: Should this be `ubi9-ruby31`? Similarly below where it is used.

jcantrill: its just a name but i will change it for correctness

PR #37683, opened 2023-03-27
Fixes to setup the openshift-extended test suite environment

aleskandro:
The openshift-extended test suite uses the env var `SSH_CLOUD_PRIV_KEY` to determine the bastion host for reaching internal hosts in the network.

This environment variable is set by looking up the content of the `${SHARED_DIR}/bastion_public_address` file. This PR create that file with the value of the AUX_HOST variable in the BM steps.

Also, this PR adds a minor fix to the determination of the TEST_PROVIDER values in the openshift-extended-test-commands for BM testing.

### DIFF
--- a/ci-operator/step-registry/openshift-extended/test/openshift-extended-test-commands.sh
+++ b/ci-operator/step-registry/openshift-extended/test/openshift-extended-test-commands.sh
@@ -141,7 +141,7 @@ openstack*)
     source "${SHARED_DIR}/cinder_credentials.sh"
     export TEST_PROVIDER='{"type":"openstack"}';;
 ovirt) export TEST_PROVIDER='{"type":"ovirt"}';;
-equinix-ocp-metal)
+equinix-ocp-metal|equinix-ocp-metal-qe)
     export TEST_PROVIDER='{"type":"skeleton"}';;
 nutanix|nutanix-qe)
     export TEST_PROVIDER='{"type":"nutanix"}';;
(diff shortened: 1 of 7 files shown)


ci-operator/step-registry/openshift-extended/test/openshift-extended-test-commands.sh
```diff
@@ -141,7 +141,7 @@ openstack*)
     source "${SHARED_DIR}/cinder_credentials.sh"
     export TEST_PROVIDER='{"type":"openstack"}';;
 ovirt) export TEST_PROVIDER='{"type":"ovirt"}';;
-equinix-ocp-metal)
+equinix-ocp-metal*)
```
aleskandro, line 144: Hi @kuiwang02  can you (dis)confirm we need this and if needed add a /approve for this file's changes?

The cluster type being used for bm clusters is `equinix-ocp-metal-qe`, but a `equinix-ocp-metal` was previously used and exists.

kuiwang02: @aleskandro thanks for the notice.
I suggest `equinix-ocp-metal|equinix-ocp-metal-qe)` to restrict it exactly in case there is other `equinix-ocp-metal-xxx` which is not we want.

aleskandro: Fixed @kuiwang02 (note there is another glob-based case for OpenStack a few rows above, probably to fix as well).

kuiwang02:
@aleskandro 
sorry to miss the comment.
please also change it for openshift-extended-test-disruptive, openshift-extended-test-longduration, openshift-extended-upgrade-post and openshift-extended-upgrade-pre

thanks

PR #37686, opened 2023-03-27
openstack: Use password authentication for cleanup jobs

pierreprinetti:
Since cleanup jobs are also responsible for removing expired application credentials, they cannot operate with restricted application credentials themselves. This change reverts cleanup jobs to use password authentication.


stephenfin:
Ah, good catch.

/lgtm

pierreprinetti:
/pj-rehearse ack
Application credentials were deleted; the job failed because of stale volumes.

PR #37694, opened 2023-03-27
Switch 4.14 ci upgrade-from-stable periodics to use latest accepted 4…

sdodson:
….13 ci

This commit can be reverted once we have a newer candidate than 4.13.0-rc.0


sdodson:
As was done in #37655

sdodson:
This change was reverted in https://github.com/openshift/release/pull/37758

PR #37695, opened 2023-03-27
sdn: refactor how images are built

jcaamano:
So that we don't use references to OCP versions that are not managed by the automation.


jluhrsen:
> @jluhrsen
> 
> PTAL, unsure if are supposed to align 4.14 and 4.15 with what we do on master

seems ok to me. what even is 4.15 at this point?

PR #37698, opened 2023-03-27
Export GOGC vars for openstack-operator

amolkahat:
This patch add vars GOGC to openstack-operator
pre-commit job.


Sandeepyadav93:
/lgtm

We added a similar workaround in nova-operator.

PR #37724, opened 2023-03-27
Revert "Fix issue DPTP-3337 Use instance storage for building"

jupierce:
Reverts openshift/release#37658


bear-redhat:
Some instances unexpectedly have nvme1 device while no instance-storage/secondary EBS volume. Reverting for now.

PR #37791, opened 2023-03-29
OpenStack: Change back 4.12 jobs to password authentication

mandre:
Introduce the new default value `version-default` for
OPENSTACK_AUTHENTICATION_METHOD that is equivalent to `password` for
4.12 and below, and equivalent to `application-credentials` for 4.13 and
above.

This allows us to change the default for OCP 4.12 while at the same time
permitting testing application credentials for 4.12.

Default 4.12 jobs back to password authentication as the jobs would run the
CSI migration tests that are not compatible with application credentials.

Related to OCPBUGS-5920.

### DIFF
--- a/ci-operator/config/shiftstack/shiftstack-ci/shiftstack-shiftstack-ci-main__periodic-4.12.yaml
+++ b/ci-operator/config/shiftstack/shiftstack-ci/shiftstack-shiftstack-ci-main__periodic-4.12.yaml
@@ -106,7 +106,6 @@ tests:
   steps:
     cluster_profile: openstack-vexxhost
     env:
-      ROTATE_CLOUD_CREDENTIALS: "true"
       TEST_SUITE: openshift/conformance/parallel
     workflow: openshift-e2e-openstack-ipi
 - as: e2e-openstack-ovn-serial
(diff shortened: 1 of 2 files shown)


ci-operator/step-registry/openstack/conf/clouds/openstack-conf-clouds-commands.sh
```diff
@@ -57,6 +57,15 @@ new_application_credentials() {
 		" "$OS_CLIENT_CONFIG_FILE"
 }
 
+# Setting a default authentication method based on version number
+if [[ "$OPENSTACK_AUTHENTICATION_METHOD" == "version-default" ]]; then
+	if ! is_openshift_version_gte "4.13"; then
```
mandre, line 62:
```suggestion
	if is_openshift_version_gte "4.13"; then
```

mandre:
/pj-rehearse periodic-ci-shiftstack-shiftstack-ci-main-periodic-4.12-e2e-openstack-ovn-appcreds

mandre:
Both jobs used the authentication method they should have used.
/pj-rehearse ack

PR #37799, opened 2023-03-29
ETCD-409: add etcd-recovery suite to aws

tjungblu:
waiting for merge and suite naming in https://github.com/openshift/origin/pull/27799

### DIFF
--- a/ci-operator/config/openshift/cluster-etcd-operator/openshift-cluster-etcd-operator-master.yaml
+++ b/ci-operator/config/openshift/cluster-etcd-operator/openshift-cluster-etcd-operator-master.yaml
@@ -204,6 +204,11 @@ tests:
   steps:
     cluster_profile: vsphere
     workflow: openshift-e2e-vsphere-etcd-scaling
+- as: e2e-aws-etcd-recovery
+  optional: true
+  steps:
+    cluster_profile: aws
+    workflow: openshift-e2e-aws-etcd-recovery
 zz_generated_metadata:
   branch: master
   org: openshift
--- a/ci-operator/config/openshift/etcd/openshift-etcd-openshift-4.14.yaml
+++ b/ci-operator/config/openshift/etcd/openshift-etcd-openshift-4.14.yaml
@@ -61,6 +61,11 @@ tests:
   steps:
     cluster_profile: aws
     workflow: openshift-perfscale-e2e-configmap-scale
+- as: e2e-aws-etcd-recovery
+  optional: true
+  steps:
+    cluster_profile: aws
+    workflow: openshift-e2e-aws-etcd-recovery
 zz_generated_metadata:
   branch: openshift-4.14
   org: openshift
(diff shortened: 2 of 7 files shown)


ci-operator/config/openshift/cluster-etcd-operator/openshift-cluster-etcd-operator-master.yaml
```diff
@@ -204,6 +204,11 @@ tests:
   steps:
     cluster_profile: vsphere
     workflow: openshift-e2e-vsphere-etcd-scaling
+- as: e2e-aws-etcd-recovery
+  optional: true
+  steps:
+    cluster_profile: aws
+    workflow: openshift-e2e-aws-etcd-recovery
```
hasbro17, line 211: While we'll definitely need presubmits for the `openshift/cluster-etcd-opearator` repo, I think we'll need to add the presubmits for `openshift/origin` so we can actually test it out on PRs where we update the test. Otherwise I don't think we'll have a way to test the changes to the test before it goes in.

That config will go here I believe:
https://github.com/openshift/release/blob/master/ci-operator/config/openshift/origin/openshift-origin-master.yaml

tjungblu: (added it nevertheless)

tjungblu:
/pj-rehearse pull-ci-openshift-cluster-etcd-operator-master-e2e-aws-etcd-recovery

tjungblu:
^ failure here is from the old EtcdRecovery test suite. The ssh test I've added passed:
> : [sig-etcd][Feature:DisasterRecovery][Suite:openshift/etcd/recovery] [Feature:EtcdRecovery] should install ssh keys on CP nodes

ci-operator/config/openshift/cluster-etcd-operator/openshift-cluster-etcd-operator-master.yaml
hasbro17: Ah sorry I wasn't aware if we could just run it with `/test`. I thought we needed the presubmit here in order to invoke the target for `/test e2e-aws-etcd-recovery`.

But if we can just run it directly then we can do without adding it to the config. It's probably not a good idea to run this by default for all PRs on origin.

tjungblu: yep we will be able to run it from the rehearsal and with /test. The presubmits for CEO and etcd should stay I presume? I'll go and remove the origin presubmits then.

tjungblu: added the one for etcd 4.14, CEO stays on master for the time being and removed the origin entry.

vrutkovs (approved):

PR #37800, opened 2023-03-29
slack notify config for 4.14 long-duration-cloud-cases

miyadav:
@sunzhaohua2 @jhou1 @huali9 @liangxia PTAL when time permits.


sunzhaohua2:
/lgtm
Thank you! @miyadav

PR #37803, opened 2023-03-29
update tag

huali9:
It was found by @duanwei33 that storage team has some cases which apply for vsphere-upi were selected but they should not be selected, because the platform is External, not vsphere now. platform: External is treated in the same fashion of platform: none, so seems we should select cases with tag `@baremetal-upi`

https://redhat-internal.slack.com/archives/C02JEKXJ3JS/p1680077797470619?thread_ts=1680051721.646739&cid=C02JEKXJ3JS

@sunzhaohua2 @miyadav @jhou1 PTAL, thanks!

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.13__amd64-nightly.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.13__amd64-nightly.yaml
@@ -2948,7 +2948,7 @@ tests:
   steps:
     cluster_profile: vsphere
     env:
-      E2E_RUN_TAGS: '@vsphere-upi and @network-ovnkubernetes and not @fips'
+      E2E_RUN_TAGS: '@baremetal-upi and @network-ovnkubernetes and not @fips'
       TAG_VERSION: '@4.13'
     test:
     - chain: openshift-e2e-test-qe
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__amd64-nightly.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__amd64-nightly.yaml
@@ -2915,7 +2915,7 @@ tests:
   steps:
     cluster_profile: vsphere
     env:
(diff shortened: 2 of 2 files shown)


ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.13__amd64-nightly.yaml
```diff
@@ -2948,7 +2948,7 @@ tests:
   steps:
     cluster_profile: vsphere
     env:
-      E2E_RUN_TAGS: '@vsphere-upi and @network-ovnkubernetes and not @fips'
+      E2E_RUN_TAGS: '@baremetal-upi and @network-ovnkubernetes and not @fips'
```
miyadav, line 2951: so there will be no `vsphere-upi `anymore ?

huali9: I think that should not work, because there is no such profile, I add a comment here, @miyadav PTAL, does it look better?

miyadav: sorry @huali9 dont think comment is allowed 
```
`ERROR: This check enforces that CI Operator configuration YAML files are generated
ERROR: correctly. We have automation in place that updates these configs and
ERROR: new changes to these configurations should be followed with a re-generation.
ERROR: Run the following command to re-generate the CI Operator configurations:
ERROR: $ make ci-operator-config
ERROR: The following errors were found:
diff -Naupr .//ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.13__amd64-nightly.yaml /tmp/tmp.juUrdbzqdR/ci-operator/config/openshift/openshift-tests-private/openshift-.
.
.

```

Hence I am ok , with it , if everyone is ok as well . 
LGTM

miyadav: you might need to remove comment.

huali9: Yeah, seems that, changed back, thanks @miyadav

PR #37807, opened 2023-03-29
Mirror hac-dev test image from quay

jrichter1:
it's already being used and built outside of openshift-ci, so it should be easier to reuse it rather than rebuild it from the same dockerfile every run
https://github.com/openshift/release/blob/master/ci-operator/config/openshift/hac-dev/openshift-hac-dev-main.yaml#L8


jrichter1:
/assign smg247
Looking for approval :)

PR #37821, opened 2023-03-29
Stackrox release 0.0

ghost:
CI configuration files to support Stackrox release `0.0`.


tommartensen:
The generated content is as expected with the current automation:
All postsubmit jobs from https://github.com/openshift/release/blob/master/ci-operator/config/stackrox/stackrox/stackrox-stackrox-master.yaml.

PR #37825, opened 2023-03-29
Reduce ipa-downloader footprint

elfosardo:
Old repo services have been removed, images jobs are broken. We just configure dummy jobs so they should not cause issues.


derekhiggins:
/lgtm
4.3 is no longer supported, if this keeps automation happy, lgtm

PR #37475, opened 2023-03-20
telco5g: donotmerge Testing PR from cnftests

sshnaidm:

PR #37839, opened 2023-03-29
telco5g: Add git config to be able to switch branches in job

sshnaidm:
Git from container fails when we need to checkout branches, change them, clone PRs. It requires user and mail to be defined. Export required variables in every step.


liornoy:
DId you validate that those env vars are enough and we don't need also the: (?)
git config --global user.name "CI User"
git config --global user.email "cnf-devel@redhat.com"

sshnaidm:
> DId you validate that those env vars are enough and we don't need also the: (?) git config --global user.name "CI User" git config --global user.email "[cnf-devel@redhat.com](mailto:cnf-devel@redhat.com)"

Yes, it's all that is needed. Tested here https://github.com/openshift/release/pull/37475 to be safe

liornoy:
great thanks. lgtm

PR #37840, opened 2023-03-29
StackRox: Remove OCP 4.6, 4.8, and 4.11 from master e2e testing

msugakov:
As part of https://issues.redhat.com/browse/ROX-15284 and as a prerequisite of https://issues.redhat.com/browse/ROX-15337

OCP 4.6 and 4.8 are out of support for ACS 4.0+.
OCP 4.11 is simply redundant since we cover 4.10 and 4.12. It's also redundant according to the matrix
https://docs.engineering.redhat.com/display/StackRox/ACS+-+OpenShift+version+support#ACSOpenShiftversionsupport-SuggestedE2Etestapproachin4.yyera


msugakov:
@gavin-stackrox Regarding
> Do you have a JIRA open already to cover implementation of a test to ensure that installation fails in the 4.9- OCPs? 

Kind of. The jira to bump Operator requirements is https://issues.redhat.com/browse/ROX-15337
I was going to test that manually.
I don't have and did not plan for preventing `roxctl` and Helm installs on OCP 4.9-. I think we should, but it's going to be more difficult and feels less urgent.

After writing this, I created https://issues.redhat.com/browse/ROX-16277

PR #35864, opened 2023-01-31
Added pre-commit job for cinder-operator

amolkahat:
This patch adds pre-commit job for cinder
operator on prow.

Depends-on: https://github.com/openstack-k8s-operators/cinder-operator/pull/147

PR #37859, opened 2023-03-30
Remove golangci-lint job

amolkahat:
Golangci-lint configured to run with pre-commit
patch[1]. Under golangci-lint we are running
govet, gomod etc

[1] https://github.com/openstack-k8s-operators/cinder-operator/pull/147


Sandeepyadav93:
go vet and go fmt can be dropped as well as we are already running them in pre-commit[1]

We can also drop golint but only after we add revive as a replacement.

[1] https://github.com/openstack-k8s-operators/cinder-operator/blob/master/.pre-commit-config.yaml#L5-L7

Sandeepyadav93:
We need to add pre-commit job before we remove the duplicate job, please i.e add pre-commit first in a separate patch or in this same patch.

bshewale:
Closing this PR. As I have clubbed this change with https://github.com/openshift/release/pull/35864

PR #37867, opened 2023-03-30
remove hypershift 4.14 all e2e job.

huangmingxia:
HyperShift operator is unstable to support create hypershift enviorment on 4.14 right now. Currently, hypershift developer mainly focus on SD-HyperShift tasks now as we plan rosa-hypershift GA on Apr 18th. We will enable 4.14 jobs again when we got signal from HyperShift developer that it's stable. 

To saving cloud cost, please remove 4.14 jobs right now.

/cc @LiangquanLi930 @heliubj18 @liangxia

### DIFF
--- a/ci-operator/config/openshift/osde2e/openshift-osde2e-main.yaml
+++ b/ci-operator/config/openshift/osde2e/openshift-osde2e-main.yaml
@@ -31,15 +31,15 @@ tests:
     make check
   container:
     from: src
-- as: rosa-pr-check
+- as: verify
   commands: |
     export GOFLAGS=""
     export REPORT_DIR="$ARTIFACT_DIR"
     export CONFIGS="rosa,pr-check"
-    export ROSA_ENV="stage"
+    export GINKGO_SKIP="Must Gather Operator"
+    export ROSA_ENV="prod"
     export ROSA_STS="true"
-    export MUST_GATHER="false"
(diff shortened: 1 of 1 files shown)


ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__amd64-nightly.yaml
```diff
@@ -654,222 +654,6 @@ tests:
     test:
     - chain: openshift-e2e-test-qe-destructive
     workflow: cucushift-installer-rehearse-aws-ipi-ovn-fips
-- as: aws-ipi-ovn-hypershift-guest-p1-f7
```
liangxia, line 657: Those are needed very soon. Instead of remove it, perhaps update the test frequency makes more sense ?

huangmingxia: This is one approach we discussed internally before. But based on current 4.12 and 4.13 debugging and HyperShift roadmap update, we may not just simplely copy all the 4.13 jobs to 4.14. We will review our profiles again in 4.14 and add some of jobs according to hypershift 4.14 planning.

liangxia: Make sense

LiangquanLi930:
@huangmingxia you can `git rebase` this PR

PR #37877, opened 2023-03-30
openstack-conf-clouds: Remove leftover from auth method selection

pierreprinetti:
Avoid unnecessarily computing the OCP version when the authentication method is explicitly passed.

/cc mandre

### DIFF
--- a/ci-operator/step-registry/openstack/conf/clouds/openstack-conf-clouds-commands.sh
+++ b/ci-operator/step-registry/openstack/conf/clouds/openstack-conf-clouds-commands.sh
@@ -62,17 +62,20 @@ new_application_credentials() {
 # Setting a default authentication method based on version number
 if [[ "$OPENSTACK_AUTHENTICATION_METHOD" == "version-default" ]]; then
 	if is_openshift_version_gte "4.13"; then
+		info 'Detected version gte 4.13: setting application credentials authentication.'
 		OPENSTACK_AUTHENTICATION_METHOD='application-credentials'
 	else
+		info 'Detected version lt 4.13: setting password authentication.'
 		OPENSTACK_AUTHENTICATION_METHOD='password'
 	fi
 fi
 
-# Skip application credentials when testing old OpenShift versions, as appcreds
-# are only supported in OCP v4.12+.
-if ! is_openshift_version_gte "4.12"; then
-	info 'The detected version does not support application credentials.'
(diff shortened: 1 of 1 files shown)


mandre (commented):
I left the check in place in case someone sets up a job with application credentials whenever the OCP version does not support it.
Would you prefer the job to fail fast instead of changing to password authentication? In that case, I suggest changing line 73 instead.

pierreprinetti:
> I left the check in place in case someone sets up a job with application credentials whenever the OCP version does not support it.

what about a warning instead? If someone sets an authentication method explicitly, they certainly don’t expect us to change it.

Another reason to remove this check is that I hit a « oc not found in PATH » in a job where I had set « password », and where there is no Installer.

pierreprinetti:
Opted for fast-fail. PTAL @mandre

ci-operator/step-registry/openstack/conf/clouds/openstack-conf-clouds-commands.sh
```diff
@@ -62,17 +62,20 @@ new_application_credentials() {
...
 fi
 
-# Skip application credentials when testing old OpenShift versions, as appcreds
-# are only supported in OCP v4.12+.
-if ! is_openshift_version_gte "4.12"; then
-	info 'The detected version does not support application credentials.'
-	OPENSTACK_AUTHENTICATION_METHOD='password'
+# Setting a default authentication method based on version number
+if [[ "$OPENSTACK_AUTHENTICATION_METHOD" == "application-credentials" ]]; then
+	if ! is_openshift_version_gte "4.13"; then
```
mandre, line 75: This now prevents testing application credentials on 4.12, which should work for the most part except for the CSI migration. Are we sure we want this? If that's the case, this should be highlighted in the commit message, as we're changing the current behavior.

pierreprinetti: I believe that we should remove this check entirely, and let the users deal with consequences when they explicitly set the auth method to `application-credentials`. A hard crash is for me the next best option, and I have no strong opinion on the version we want to apply this to. Shall I set 4.12?

pierreprinetti:
/pj-rehearse pull-ci-openshift-cloud-credential-operator-release-4.12-e2e-openstack pull-ci-openshift-csi-driver-nfs-release-4.11-e2e-openstack-csi

mandre (commented):
/lgtm

PR #37878, opened 2023-03-30
Remove golint job

mrkisaolamb:
golint check now is replaced by golangci-lint
in pre-commit check


gibizer:
you probably need to run `make update` in the root of the repo to auto generate some changes...

mrKisaoLamb:
> you probably need to run `make update` in the root of the repo to auto generate some changes...

Thanks, done!

gibizer (approved):
Looks good, we replace golint with revive provided by golangci-lint. So this is a good cleanup.

PR #37880, opened 2023-03-30
Disable tide for release-service

mmalina:
We don't want automatic merges for now
and that's what tide does. And we don't
need tide for anything else right now.

Here's an example of where a PR was merged by [penshift-merge-robot](https://github.com/openshift-merge-robot): https://github.com/redhat-appstudio/release-service/pull/171

Cc: @davidmogar


johnbieren:
See https://github.com/openshift/release/pull/37885

mmalina:
> See #37885

Hmm, ok, so it's already done. Closing this.

PR #37881, opened 2023-03-30
Set cpu request more inline with actual used amount

psalajova:
The same as in https://github.com/openshift/release/pull/37441; these changes should take care of the rest of the alerts we keep getting.


psalajova:
prometheus-kcp-e2e

ncdc (approved):

PR #37885, opened 2023-03-30
Release service: Remove tide

johnbieren:


mmalina:
For the record, this was done because we don't want auto merge in our repo for now and that's what tide does.

Here's an example of where a PR was merged by [penshift-merge-robot](https://github.com/openshift-merge-robot): https://github.com/redhat-appstudio/release-service/pull/171

PR #37893, opened 2023-03-30
Remove 5.3 OAL jobs

jcantrill:
This PR
* remove OAL logging 5.3 test jobs

cc @periklis


periklis:
@jcantrill actually we can delete these jobs entirely. We do not have any 5.3 CPaaS pipeline so what is the benefit from keeping these jobs even optional here?

periklis (approved):
/lgtm

PR #37900, opened 2023-03-30
PowerVS: Turn one cluster type into two

hamzy:
We are creating two cluster types in order to split the PowerVS zones into two categories.


hamzy:
This is the second part of this PR:
https://github.com/openshift/ci-tools/pull/3352

PR #37907, opened 2023-03-30
use RELEASE_VERSION instead of IMAGE_TAG

Jooho:
use RELEASE_VERSION instead of IMAGE_TAG to track the build commit


anishasthana (approved):
/lgtm

PR #37914, opened 2023-03-30
openstack-operator: fix DOCKERFILE variable

dprince:
This should resolve a 'DOCKERFILE: unbound variable' error by simply setting it to the name of the actual file instead of this variable which we don't currently need elsewhere.


ci-operator/step-registry/openstack-k8s-operators/build/openstack-k8s-operators-build-commands.sh
```diff
@@ -74,7 +74,7 @@ function build_push_operator_images {
   if [[ -f storage-bundle.Dockerfile ]]; then
     DOCKERFILE=storage-bundle.Dockerfile /bin/bash hack/pin-custom-bundle-dockerfile.sh
```
dprince, line 75: NOTE: This needs to be set here so it passes through the subshell /bin/bash command. It does not get applied below where I set it directly to the .pinned version which is generated by this command.

abays (approved):
/lgtm

PR #37924, opened 2023-03-31
Add check-control-plane-machine-name to check-cluster-health chain to make sure masters are not recreated

sunzhaohua2:
Add health check chain to existing workflows @jhou1 @jianlinliu @yunjiang29 please help to take a look if this makes sense, if so I'll update other workflows, thanks!
Discuss in here: https://redhat-internal.slack.com/archives/CF8SMALS1/p1679641767534869

### DIFF
--- a/ci-operator/step-registry/cucushift/installer/check/cluster-health/cucushift-installer-check-cluster-health-chain.yaml
+++ b/ci-operator/step-registry/cucushift/installer/check/cluster-health/cucushift-installer-check-cluster-health-chain.yaml
@@ -2,5 +2,6 @@ chain:
   as: cucushift-installer-check-cluster-health
   steps:
   - ref: cucushift-installer-check-cluster-health
+  - ref: cucushift-installer-check-control-plane-machine-name
   documentation: |-
     Check cluster is healthy after installation finish
--- a/ci-operator/step-registry/cucushift/installer/rehearse/aws/ipi/usertags/provision/cucushift-installer-rehearse-aws-ipi-usertags-provision-chain.yaml
+++ b/ci-operator/step-registry/cucushift/installer/rehearse/aws/ipi/usertags/provision/cucushift-installer-rehearse-aws-ipi-usertags-provision-chain.yaml
@@ -2,7 +2,6 @@ chain:
   as: cucushift-installer-rehearse-aws-ipi-usertags-provision
   steps:
   - chain: cucushift-installer-rehearse-aws-ipi-provision
-  - ref: cucushift-installer-check-control-plane-machine-name
   env:
   - name: USER_TAGS
     default: |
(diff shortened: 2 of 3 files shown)


jianlinliu:
maybe this will conflicts with what @liangxia is doing in https://github.com/openshift/release/pull/37918, if I remember right, prow ci does not allow call the same step more than one time.

sunzhaohua2:
> maybe this will conflicts with what @liangxia is doing in #37918, if I remember right, prow ci does not allow call the same step more than one time.

Yes, thanks, this will conflict with those jobs. In this pr, I created a chain, add all the check steps to this chain, and then add this chain to the provision chain, so that the check will complete after installation, jobs will not need `test  ref: cucushift-installer-check-cluster-health`. The issue is maybe not every check point is suitable for every version. 

If we add the check in jobs, it will be more flexible for different version, different clusters. what we should do now? Update/add in jobs test with `chain: cucushift-installer-check` for 4.13 aws/gcp/azure, 4.12 aws, do you think this is ok?

jianlinliu:
lxia already introduced `cucushift-installer-check-cluster-health` chain in https://github.com/openshift/release/pull/37971/files, if you like to run `cucushift-installer-check-control-plane-machine-name` checking, I guess we can add the step to `cucushift-installer-check-cluster-health` chain now.

ci-operator/step-registry/cucushift/installer/check/cluster-health/cucushift-installer-check-cluster-health-chain.yaml
```diff
@@ -2,5 +2,6 @@ chain:
   as: cucushift-installer-check-cluster-health
   steps:
   - ref: cucushift-installer-check-cluster-health
+  - ref: cucushift-installer-check-control-plane-machine-name
```
liangxia, line 5: Will this check works on all profiles, e.g, sno, hypershift, rosa, aro, ...

jianlinliu: In the `cucushift-installer-check-control-plane-machine-name` step, I saw the following pre-check code:
```
oc get machine -n openshift-machine-api --selector machine.openshift.io/cluster-api-machine-type=master --no-headers -owide 1>${machine_output} 2>${err_output}

if grep -ir 'No resources found in openshift-machine-api namespace.' ${err_output}; then
    echo "WARN: No machines in openshift-machine-api namespace, skip checking"
    exit 0
fi
```
So I guess it should work in most of situations.

liangxia: Ack

liangxia:
Perhaps we need to update PR title to reflect the code change, too ?

sunzhaohua2:
> Perhaps we need to update PR title to reflect the code change, too ?

Done

liangxia:
I can see Yunfei used a hold above, just in case he has concerns

/assign @yunjiang29

PR #37937, opened 2023-03-31
slack notify for longduration cloud tests to include success/failure/…

miyadav:
…error

@sunzhaohua2 @huali9 @jhou1 PTAL , when time permits.


sunzhaohua2:
Thanks @miyadav cloud you help to add this to 4.12 jobs too?

PR #37940, opened 2023-03-31
stackrox: a release branch template and test branch

gavin-stackrox:
Provides a template for future release branches and is active config that would run in a stackrox/stackrox branch named release-x.y.
/uncc
/hold

### DIFF
--- a/ci-operator/config/stackrox/stackrox/README.md
+++ b/ci-operator/config/stackrox/stackrox/README.md
@@ -9,7 +9,5 @@ and `Pull Request checklist` therein.
 ## Release Template
 
 `stackrox-stackrox-release-x.y*.yaml` files are template YAML used by automation
-to create config for release at branch time. 
-
-*When changes are made to master or nightlies yaml consider if the changes are
-also appropriate for release branches.*
+to create config for release at branch time. Review the `Pull Request checklist`
+for changes required there.


gavin-stackrox:
@tommartensen @0x656b694d @msugakov wdyt?
It would be preferable to fix ci-tools to ignore certain files so the templates could be `.yaml` but that is harder.

msugakov:
We could go further and make these actually `yaml` files pointing to a real `release-x.y` branch in the `stackrox/stackrox` repo.
This way, engineer making changes to CI not only can see the changes but can also test them by force-pushing `release-x.y` to the latest master.

A side-advantage of YAMLs will be that `jobs/` files will be generated and the engineer will configure `always_run: false` so that these can be copied by automation too or at least exist for comparison.

WDYT?

gavin-stackrox:
> WDYT?

worth a try and yes, more useful. (I had originally discussed using `0.0` with @tommartensen but then I decided that release automation might want to overwrite it.)

ci-operator/jobs/stackrox/stackrox/stackrox-stackrox-release-x.y-presubmits.yaml
```diff
@@ -0,0 +1,1459 @@
+presubmits:
+  stackrox/stackrox:
+  - agent: kubernetes
+    always_run: true
```
msugakov, line 4: When I diff `stackrox-stackrox-release-4.0-presubmits.yaml` and this file, I see a bunch of differing `always_run:` values. Could you please adjust them to be consistent (either way)?

ci-operator/config/stackrox/stackrox/README.md
```diff
@@ -1,3 +1,15 @@
...
+CI](https://docs.engineering.redhat.com/display/StackRox/Making+changes+to+OpenShift+CI)
+and `Pull Request checklist` therein.
+
+## Release Template
+
+`stackrox-stackrox-release-x.y*.yaml` files are template YAML used by automation
+to create config for release at branch time. 
+
+*When changes are made to master or nightlies yaml consider if the changes are
+also appropriate for release branches.*
```
msugakov, line 15: In addition (or instead of this) I suggest we rely on the `Pull Request Checklist`. I updated the one by adding the item number 3 there. PTAL and edit as you see fit.

gavin-stackrox: updated.

msugakov (approved):

tommartensen (approved):
Thanks @gavin-stackrox , this looks good. 

The automation will look for all `stackrox-stackrox-release-x.y*.yaml` files and replace `x.y`. 
That's also simpler than what we do right now.

PR #37943, opened 2023-03-31
USHIFT-1040: Fix openshift-conformance-reduced tests

pacevedom:
Fix names for previously skipped tests because of not being able to match them to current names.


pacevedom:
/pj-rehearse pull-ci-openshift-microshift-main-e2e-openshift-conformance-reduced

dhellmann:
This removes a bunch of test cases, but the description here and the linked Jira ticket don't really explain why we want to do that. Can you provide some more background?

pacevedom:
> This removes a bunch of test cases, but the description here and the linked Jira ticket don't really explain why we want to do that. Can you provide some more background?

Actually its adding more than double what we were running. Root cause was unmatched names. Working on not having to do these manually, this is only for the time being.

pacevedom:
/pj-rehearse pull-ci-openshift-microshift-main-e2e-openshift-conformance-reduced

PR #37950, opened 2023-03-31
Add readme to stackrox/stackrox pipelines

msugakov:
This is for future-us so that we can more easily find important information that's documented in Confluence.

/uncc


gavin-stackrox (approved):
/lgtm

PR #37992, opened 2023-04-03
Deployment of gangway component

jmguzik:
/hold


jmguzik:
/cc @hongkailiu 

@hongkailiu please check if you see sth wrong

danilo-gemoli:
/lgtm
let's give it a try, we can eventually add any missing parameters along the way

PR #38004, opened 2023-04-03
CVP-3304: Reduce the cluster sized for EOL versions of Openshift v4.6, v4.7, v4.8, v4.9

samvarankashyap:
The above PR reduces the cluster pool sizes to 1
Resets the running count to 0
Max size of the cluster pools set to 5

The above provisions are made for manual triggering of the Openshift operator bundle testing

### DIFF
--- a/clusters/hive/pools/cvp/cvp-ocp-4-9-amd64-aws-us-west-2_clusterpool.yaml
+++ b/clusters/hive/pools/cvp/cvp-ocp-4-9-amd64-aws-us-west-2_clusterpool.yaml
@@ -24,7 +24,7 @@ spec:
     name: install-config-aws-us-west-2
   labels:
     tp.openshift.io/owner: cvp
-  maxSize: 10
+  maxSize: 5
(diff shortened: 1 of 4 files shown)


clusters/hive/pools/cvp/cvp-ocp-4-9-amd64-aws-us-west-2_clusterpool.yaml
```diff
@@ -24,16 +24,16 @@ spec:
...
     aws:
       credentialsSecretRef:
         name: cvp-aws-credentials
       region: us-west-2
   pullSecretRef:
     name: pull-secret
-  runningCount: 1
-  size: 5
+  runningCount: 0
+  size: 1
```
lslebodn, line 36: The maintenance support for 4.9 will end in April 18, 2023
* https://access.redhat.com/support/policy/updates/openshift#dates

IMHO the change for 4.9 should be done in few weeks and not now.

PR #38017, opened 2023-04-04
Refine cron entry

liangxia:
Frequency f21 does not do what it's expected to do, run tests every 21 days, as we do not have double 21 days in any month.
Remove those frequency levels from the script, and rename existing f21 jobs to f28. If you think f28 does not met your requirements, you can use follow up PRs to update that to any frequency as you wish.

/cc @jianlinliu @YuLi517 @jhou1 @JianLi-RH @dis016 @pruan-rht

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/tools/generate-cron-entry.sh
+++ b/ci-operator/config/openshift/openshift-tests-private/tools/generate-cron-entry.sh
@@ -68,7 +68,7 @@ case "$FN" in
 	1)
 		echo "$MINUTE $HOUR * * *"
 		;;
-	2|3|4|5|6|7|8|9|10|11|12|13|14|15|16|17|18|19|20|21|22|23|24|25|26|27|28|29|30|31)
+	2|3|4|5|6|7|10|14|28|30)
 		DAY_OF_MONTH_TMP=$DAY_OF_MONTH
 		for ((i=1 ; i<31/FN; ++i)) ; do
 			let TMP=(i*FN+DAY_OF_MONTH)%30+1
--- a/ci-operator/jobs/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12-periodics.yaml
+++ b/ci-operator/jobs/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12-periodics.yaml
@@ -9965,7 +9965,7 @@ periodics:
         secretName: result-aggregator
 - agent: kubernetes
   cluster: build03
-  cron: 7 23 16 * *
+  cron: 1 1 10 * *
   decorate: true
   decoration_config:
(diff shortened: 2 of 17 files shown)


YuLi517:
cc: @LiangquanLi930 @heliubj18

ci-operator/config/openshift/openshift-tests-private/tools/generate-cron-entry.sh
```diff
@@ -68,7 +68,7 @@ case "$FN" in
 	1)
 		echo "$MINUTE $HOUR * * *"
 		;;
-	2|3|4|5|6|7|8|9|10|11|12|13|14|15|16|17|18|19|20|21|22|23|24|25|26|27|28|29|30|31)
+	2|3|4|5|6|7|10|14|28|30)
```
YuLi517, line 71: I suggest to keep 21 here. We may use it as one Sprint is 3 weeks (21 days). If any job need to follow Sprint, we have this option here.

liangxia: It's already explained in the description.

YuLi517: I think this code is just providing option of f21. It's not saying all job using f21 right now. Am I right? I just wan to keep this option exist. If in future, we need f21, we can directly use.

liangxia: f21 does not do what we expected it to do. That's why we need to remove it. We do not wan to let people choose f21 as they wish it to execute every 3 weeks, but actually it executes every month.
We may consider to add it back once we found a solution to let jobs execute at every 3 weeks.

YuLi517: My understanding is cron limitation. I am ok with you approach right now.

ci-operator/jobs/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12-periodics.yaml
```diff
@@ -9980,18 +9980,7 @@ periodics:
     ci.openshift.io/generator: prowgen
     job-release: "4.12"
     pj-rehearse.openshift.io/can-be-rehearsed: "true"
-  name: periodic-ci-openshift-openshift-tests-private-release-4.12-amd64-nightly-aws-ipi-ovn-hypershift-install-p2-f21
```
LiangquanLi930, line 9983: @liangxia We should maintenance hypershift `reporter_config` for notification.

liangxia:
Drop the PR as too many conflicts.

PR #38027, opened 2023-04-04
skip e2e-agent-compact job on unrelated changes

osherdp:
Seeing the job running for
https://github.com/openshift/assisted-installer/pull/659 where we only changed approvers list, we can just skip it for similar cases.

### DIFF
--- a/ci-operator/config/openshift/assisted-installer/openshift-assisted-installer-master.yaml
+++ b/ci-operator/config/openshift/assisted-installer/openshift-assisted-installer-master.yaml
@@ -59,6 +59,7 @@ resources:
 tests:
 - as: e2e-agent-compact
   cluster: build05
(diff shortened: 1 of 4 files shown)


ci-operator/config/openshift/assisted-installer/openshift-assisted-installer-master.yaml
```diff
@@ -59,6 +59,7 @@ resources:
 tests:
 - as: e2e-agent-compact
   cluster: build05
+  skip_if_only_changed: ^\.github/|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|OWNERS_ALIASES|LICENSE)$
```
adriengentil, line 62: is this parameter set on other repos as well? Just trying to understand why I runs more than e2e-assisted-metal and e2e-ai-operator-ztp in assisted-service repo.

adriengentil (approved):

PR #38054, opened 2023-04-04
add -ovn- to aws-single-node presubmit in CNO

jluhrsen:
fixes the test flake where job name doesn't match the underlying CNI.


jcaamano:
how is that all the jobs have been shuffled around instead of just a change of that one job e2e-aws-single-node to e2e-aws-ovn-single-node?

jcaamano:
haha ok, got me offguard

/lgtm
/approve

PR #38060, opened 2023-04-04
openstack/gather: skip gather only if no bastion

EmilienM:
We were checking for the wrong file.
Instead of checking for proxy-conf.sh which is used by any job with a
proxy (even those with a bastion), we need to check for
squid-credentials.txt which only exists for clouds with a permanent
proxy, and no bastion.


gryf (approved):
/lgtm

MaysaMacedo:
/pj-rehearse pull-ci-openshift-installer-master-e2e-openstack-proxy

MaysaMacedo:
/pj-rehearse periodic-ci-shiftstack-shiftstack-ci-main-periodic-4.12-e2e-openstack-nfv-intel

MaysaMacedo:
/lgtm
/pj-rehearse ack
I don't think we can exercise the jobs given that the code fragment will only run when bootstrap fails

PR #38095, opened 2023-04-06
OCPQE-14334: Add cluster health check step before running e2e test

liangxia:
/cc @jianlinliu @shellyyang1989 @jhou1 @JianLi-RH @dis016 @pruan-rht


shellyyang1989:
/lgtm

Thank you Liang for the fix!

PR #38101, opened 2023-04-06
Revert "Added random etcd encryption"

jianlinliu:
Reverts openshift/release#37980

It is breaking the old version of ci testing on 4.12-


dpuniaredhat:
@jianlinliu @wangke19 I will raise a separate PR with condition to skip `aesgcm` if cluster versions is less than `4.13`. As of now we can revert this PR to avoid failure in other version.

PR #38102, opened 2023-04-06
installer: make on-demand jobs explicit in config

zaneb:
The default for generating new jobs is for them to always_run, so for existing jobs with always_run: false, set always_run: false in the config so that if they are copied or renamed they will remain on-demand and not start running on every PR.

This is now possible due to openshift/ci-tools#3068.

For now this is only on the master and release-4.14 branches, as these are the most likely to be copied or renamed.


r4f4 (commented):
/lgtm

zaneb:
Looks like the validator needs fixing.

zaneb:
Oops, now with correct alphabetisation.

PR #38135, opened 2023-04-06
[WINC] Use clusterbot profile for vsphere

saifshaikh48:
This PR uses the clusterbot profile for our vsphere jobs.
This is to ensure none of the jobs land in the IBM vCenter environment, 
where we are seeing IPv6 addresses being assigned to Windows nodes 
which is not supported. This is a temporary workaround that should be 
reverted once IPv6 is disabled on the Windows VM golden image template.


saifshaikh48:
/close

We've created a new golden image that works in IBM cloud (ipv6 disabled). No need for this anymore

PR #38136, opened 2023-04-06
OSSM-3716: Add maistra prometheus tests

jwendell:


ci-operator/config/maistra/prometheus/maistra-prometheus-maistra-2.1.yaml
```diff
@@ -0,0 +1,25 @@
+build_root:
+  image_stream_tag:
+    name: maistra-builder
+    namespace: ci
+    tag: "2.1"
+  use_build_cache: true
+resources:
+  '*':
+    limits:
+      memory: 8Gi
```
yxun, line 10: remove this memory limit. The default cluster node may not has 8Gi memory.

jwendell: It worked previously. Other jobs have 8 and some even 16. Note that this is the limit, not the request.

ci-operator/jobs/maistra/prometheus/maistra-prometheus-maistra-2.1-presubmits.yaml
```diff
@@ -0,0 +1,52 @@
+presubmits:
+  maistra/prometheus:
+  - agent: kubernetes
+    always_run: true
+    branches:
+    - ^maistra-2\.1$
+    - ^maistra-2\.1-
+    cluster: build01
```
yxun, line 8: Do you know why make update generates cluster build01 ?
It looks like a small cluster. We may need to run `make update` again.

I had larger memory requests value in maistra/istio jobs and they are on cluster: build03 from my make update.

ci-operator/config/maistra/prometheus/maistra-prometheus-maistra-2.1.yaml
yxun: hm, I don't know other difference. 
You may fetch and merge recent master branch changes. And then run `make update` again. 
I guess they have changed some infrastructure settings and old master branch code  `make update` target need an update.

PR #38145, opened 2023-04-07
add base_images

LiangquanLi930:


LiangquanLi930:
@liangxia Could you help to review this PR?

core-services/ci-chat-bot/workflows-config.yaml
```diff
@@ -66,6 +66,11 @@ workflows:
     platform: aws
   cucushift-installer-rehearse-aws-ipi-ovn-hypershift:
     platform: aws
+    base_images:
+      hypershift-operator:
+        name: "4.14"
```
liangxia, line 71: Do we have jobs in 4.14 ? Shall we use 4.12 as most of hypershift tests are with 4.12 ?

heliubj18: @liangxia Yes, for hypershift operator, currently we always use the latest version to provision different payloads e.g. hostedcluster 4.12/4.13/4.14.  Because hypershift is not GA now, the operator codes are changed quickly every day. So for both QE and Dev we all use the main branch (currently the same as 4.14) to build hypershift operator.

heliubj18: BTW, it seems no way to debug or try the cluster bot command for the newly added workflow in PR status. right? Thanks!

PR #38147, opened 2023-04-07
mtu-migration: oc sa get-token is deprecated

rbbratta:
replace with `oc -n openshift-monitoring create token prometheus-k8s`

```
Command "get-token" is deprecated, and will be removed in the future version. Use oc create token instead.
error: could not find a service account token for service account "prometheus-k8s"
parse error: Invalid numeric literal at line 2, column 10
alerts in firing state:
```


rbbratta:
PTAL @jluhrsen

jluhrsen:
> PTAL @jluhrsen

/approve

PR #38155, opened 2023-04-07
OCPQE-14727: skip health-check

LiangquanLi930:


ci-operator/step-registry/cucushift/installer/check/cluster-health/cucushift-installer-check-cluster-health-commands.sh
```diff
@@ -208,6 +208,14 @@ function check_history() {
 }
 
 ###Main###
+if [ -f "${SHARED_DIR}/cluster-type" ] ; then
+    CLUSTER_TYPE=$(cat "${SHARED_DIR}/cluster-type")
+    if [[ "$CLUSTER_TYPE" == "osd" ]] || [[ "$CLUSTER_TYPE" == "rosa" ]] || [[ "$CLUSTER_TYPE" == "hypershift-guest" ]]; then
```
yasun1, line 213: We do not have the 'cluster_type="hypershift-guest"'.  Till now, the hypershift cluster is a kind of rosa cluster. The cluster_type is the product_id of the cluster.
`PRODUCT_ID=$(rosa describe cluster -c "${CLUSTER_ID}" -o json | jq -r '.product.id')`

yasun1: Is it overwritten in your steps?

LiangquanLi930: https://github.com/openshift/release/blob/294c68c2ce708519cd19b8dc352f79c003d79c4d/ci-operator/step-registry/cucushift/hypershift-extended/enable-guest/cucushift-hypershift-extended-enable-guest-commands.sh#L21
@yasun1

yasun1: ok

jianlinliu: Actually in prow, we are not encouraged to use `if else`, we are supposed to call the needed step in the chain or workflow. So I suggest to remove the health check chain in these ci jobs.

LiangquanLi930: @jianlinliu You mean we should have a chain: `openshift-e2e-test-hypershift-qe` right? 
cc @liangxia

jianlinliu: Yeah, that also works to create a new e2e chain in which does not have the common cluster health check step.

LiangquanLi930:
close this PR, Replace with https://github.com/openshift/release/pull/38172

PR #38103, opened 2023-04-06
OCPQE-14239: Integrate jq in the image to avoid download jq from internet

liangxia:
Changes in this PR,
1. Switch pod image from ansible to verification-tests, as the later one has jq in it.
2. Remove the codes that download jq from github, as interact with external network is unstable.
3. Use proper openshift-ansible branch and ansible version based on OCP version.

PR #38186, opened 2023-04-10
Update the ansible playbook dir

shellyyang1989:
Addresses the [issue](https://github.com/openshift/release/pull/38103#issuecomment-1501576729):
```
ERROR! the playbook: playbooks/upgrade.yml could not be found
```
cc @liangxia for review.

### DIFF
--- a/ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
+++ b/ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
@@ -127,7 +127,7 @@ function rhel_upgrade(){
     echo "Validating parsed Ansible inventory"
     ansible-inventory -i "${SHARED_DIR}/ansible-hosts" --list --yaml
     echo "Running RHEL worker upgrade"
-    ansible-playbook -i "${SHARED_DIR}/ansible-hosts" playbooks/upgrade.yml -vvv
+    ansible-playbook -i "${SHARED_DIR}/ansible-hosts" /usr/share/ansible/openshift-ansible/playbooks/upgrade.yml -vvv
 
     echo "Check K8s version on the RHEL node"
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -127,7 +127,7 @@ function rhel_upgrade(){
     echo "Validating parsed Ansible inventory"
     ansible-inventory -i "${SHARED_DIR}/ansible-hosts" --list --yaml
     echo "Running RHEL worker upgrade"
-    ansible-playbook -i "${SHARED_DIR}/ansible-hosts" playbooks/upgrade.yml -vvv
+    ansible-playbook -i "${SHARED_DIR}/ansible-hosts" /usr/share/ansible/playbooks/upgrade.yml -vvv
```
liangxia, line 130: Is the path correct ?
```
[root@63a3a5083fc3 /]# ls -lhd /usr/share/ansible/openshift-ansible/playbooks/  /usr/share/ansible/playbooks/
ls: cannot access '/usr/share/ansible/playbooks/': No such file or directory
drwxr-xr-x. 1 root root 72 Apr  7 02:33 /usr/share/ansible/openshift-ansible/playbooks/
```

shellyyang1989: ah, you're right. Thank you for the correction! Updated

liangxia: I just realized that maybe this path change need to happen together with the image switch, or else it might break existing RHEL upgrade as we haven't switched the image yet.

shellyyang1989: That's right! Let me close it.

PR #38194, opened 2023-04-10
MicroShift - set golangci-lint explicitly

copejon:
Golangci-lint cannot access the default directory to create the .cache subdir because the home dir does not exist.  Specify the cache dir under /tmp to workaround this.

Unblocks openshift/microshift#1605

### DIFF
--- a/ci-operator/jobs/openshift/microshift/openshift-microshift-main-presubmits.yaml
+++ b/ci-operator/jobs/openshift/microshift/openshift-microshift-main-presubmits.yaml
@@ -979,58 +979,8 @@ presubmits:
         - --gcs-upload-secret=/secrets/gcs/service-account.json
         - --image-import-pull-secret=/etc/pull-secret/.dockerconfigjson
         - --report-credentials-file=/etc/report/credentials
-        - --target=verify
-        command:
-        - ci-operator
-        image: ci-operator:latest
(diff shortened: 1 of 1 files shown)


ci-operator/config/openshift/microshift/openshift-microshift-main.yaml
```diff
@@ -110,9 +110,18 @@ tag_specification:
...
+    - as: verify
+      commands: make verify
+      env:
+      - default: /tmp/
+        name: GOLANGCI_LINT_CACHE
+      from: src
+      resources:
+        requests:
+          cpu: 100m
+          memory: 200Mi
```
pacevedom, line 124: My only nit, is this enough for good scheduling? We can always check after a few executions to adjust.

pacevedom (approved):
/lgtm
/hold
Holding in case you want to check the resources, I am good to look into it in a different PR.

ci-operator/config/openshift/microshift/openshift-microshift-main.yaml
copejon: It's the same as the global values we set for all tests, so it's probably fine. For whatever reason, steps under the test stanza don't inherit the global request values, according to the linter anyway.

PR #38220, opened 2023-04-11
OADP Plugins: make presubmit test optional:true

hhpatel14:


sseago (approved):

hhpatel14:
fixes https://github.com/openshift/oadp-operator/issues/932

PR #38221, opened 2023-04-11
add openshift-e2e-vsphere-ovn-dualstack workflow

rbbratta:


rbbratta:
This doesn't work because we don't select the `vsphere-ibm-7-quota-slice`

rbbratta:
PTAL @zhaozhanqi

core-services/ci-chat-bot/workflows-config.yaml
```diff
@@ -257,6 +257,8 @@ workflows:
     platform: openstack
   openshift-e2e-vsphere:
     platform: vsphere
+  openshift-e2e-vsphere-ovn-dualstack:
```
liangxia, line 260: Is it the right one you want, did you mean to add `cucushift-installer-rehearse-vsphere-ipi-ovn-dualstack` ?

rbbratta: ` cucushift-installer-rehearse-vsphere-ipi-ovn-dualstack` is already added.  We want both eventually.  Everyone should be testing dualstack, it's the future.

liangxia: Ack

PR #38226, opened 2023-04-11
ossm-3607 split maistra istio integ jobs

yxun:
This PR splits the maistra istio integ job into six jobs. Each of them will claim a standby cluster and run `servicemesh-istio-e2e-aws` workflow and then execute an istio integ-suite-kind test target.


jwendell (commented):
nit: can we remove the word `kube` from all jobs? It doesn't add any value.

no need to rehearse the jobs after this change.

PR #38248, opened 2023-04-12
remove duplicated ci coverage on azure

jianlinliu:
1. azure-ipi-ovn is already well covered in azure-ipi-ovn-etcd-encryption
2. azure-ipi-private-ovn is redundant when azure-ipi-private is already covered
3. some azure-ipi-private jobs are duplicated with azure-ipi-private-workers-marketimage


jianlinliu:
cc @liangxia for review.

PR #38251, opened 2023-04-12
Add gitops interop

bluengo:
Updated the necessary files according to the [documentation](https://cspi-qe.gitbook.io/ocp-ci-onboarding/onboarding/onboarding_guide/scenario_development_guide) to configure gitops-operator interop testing.

This PR includes:
1. Update the `config` files with the required information to install gitops-operator.
2. Creation of a new folder under `step-registry` with the specification for the testing container.

PTAL and let me know if something is wrong or missing.

Thanks in advance.

### DIFF
--- a/ci-operator/config/redhat-developer/gitops-operator/redhat-developer-gitops-operator-master.yaml
+++ b/ci-operator/config/redhat-developer/gitops-operator/redhat-developer-gitops-operator-master.yaml
@@ -1,23 +1,49 @@
-binary_build_commands: go mod vendor
+base_images:
+  cli:
+    name: "4.13"
+    namespace: ocp
+    tag: cli
+  os:
+    name: ubi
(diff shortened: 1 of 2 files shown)


Sau1506mya:
@bluengo , First of all you can mark the PR as WIP (Work in Progress) so that everyone knows its just a draft.
Second, before adding the config file and raising the PR, we run the `make update` command to get the automated prow job.

Sau1506mya:
@bluengo , After that you need to be a member of `openshift org`

Sau1506mya:
@bluengo , Can you tell me why you have added the two different config files?
Also, naming convention is not correct, it should be something like this: `ci-operator/config/redhat-developer/gitops-operator/redhat-developer-gitops-operator-master__gitops-ocp4.13-lp-interop.yaml`

PR #38265, opened 2023-04-12
Sdcicd 913 added CCS config for region enablement job

ritmun:
Added CCS creds for ap-se-4 region enablement job

[sdcicd-913](https://issues.redhat.com//browse/sdcicd-913)


ci-operator/jobs/openshift/osde2e/openshift-osde2e-main-periodics.yaml
ryankwilliams, line 1: @ritmun Should we make the job rehearsable? This way we can tell if that secret will work?

PR #38275, opened 2023-04-12
Installer: trigger fewer aws jobs

patrickdillon:
These jobs have limited usefulness and should not be run on installer pulls.

The rhel worker and upgrade jobs aren't providing any meaningful test coverage. (I believe we intended to stop running these jobs in the past: the configs were updated so they wouldn't run but you must manually update the jobs.) 

The proxy job is now permafailing due to the recent aws s3 acl update. We should evaluate whether we can use QE's workflows in stead. Let's stop this permafailing job in the meantime.

### DIFF
--- a/ci-operator/config/openshift/installer/openshift-installer-master__okd-scos.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-master__okd-scos.yaml
@@ -120,7 +120,6 @@ tests:
     workflow: openshift-e2e-gcp
 - as: e2e-aws-upgrade
   optional: true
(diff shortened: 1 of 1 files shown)


rna-afk:
LGTM. needs `make jobs`

patrickdillon:
> LGTM. needs `make jobs`

oops. I think [b8a8642](https://github.com/openshift/release/pull/38275/commits/b8a8642ebba996b3b0e0013775557d3c9f6aeeed) may fix it.

It seems that the jobs only get updated from the configs when you add the triggers, so if you want to remove them you have to manually update both the job and the config. Before this change, I incorrectly updated the okd config so the trigger was getting added back in. I think it's fixed now.

PR #38290, opened 2023-04-13
Add nmstatectl for the unit tests #38290

andfasano:
Some of the agent-based installer unit tests require the presence of nmstatectl, otherwise they will be skipped silently.

*Note:* this PR could be merged only after https://github.com/openshift/installer/pull/7089

### DIFF
--- a/ci-operator/config/openshift/installer/openshift-installer-master.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-master.yaml
@@ -201,6 +201,9 @@ resources:
     requests:
       cpu: "3"
       memory: 5Gi
+test_binary_build_commands: |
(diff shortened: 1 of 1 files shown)


zaneb:
Please make the same changes to the release-4.14 branch config, so that we don't lose this config after branching.
Does the same issue apply to 4.13?

andfasano:
> Please make the same changes to the release-4.14 branch config, so that we don't lose this config after branching. 
Isn't the CI bot copying it already from master -> 4.14? 

> Does the same issue apply to 4.13?
Good point, I'll add it as well

patrickdillon:
/retitle  Add nmstatectl for the unit tests #38290

PR #38317, opened 2023-04-13
config/cri-o: update release target on main jobs

haircommander:
This may also fix CRI-O's openshift e2e woes on main


haircommander:
/pj-rehearse
/cc @sohankunkerkar 
(it's possible this misconfiguration is our problem--1.26 doesn't fail this way and a diff between main and 1.26 revealed this)

PR #38327, opened 2023-04-14
Skip long-duration cases in destructive jobs as there are jobs to run long-duration cases

sunzhaohua2:
Most long-duration cases are destructive, want to skip long-duration cases in destructive jobs, so that we can only cover these cases in only long-duration jobs. @jhou1 @liangxia PTAL


sunzhaohua2:
Close as covered in https://github.com/openshift/release/pull/38328

PR #38328, opened 2023-04-14
Skip long-duration cases in destructive tests

liangxia:
Do the same as in https://github.com/openshift/release/pull/38327

/cc @sunzhaohua2 @jhou1 @JianLi-RH @dis016 @pruan-rht


sunzhaohua2:
LGTM, thanks @liangxia

PR #38371, opened 2023-04-16
MGMT-14208: Adding presubmit lint job

danmanor:
Adding presubmit lint job

### DIFF
--- a/ci-operator/config/openshift-assisted/assisted-events-stream/openshift-assisted-assisted-events-stream-master.yaml
+++ b/ci-operator/config/openshift-assisted/assisted-events-stream/openshift-assisted-assisted-events-stream-master.yaml
@@ -1,3 +1,8 @@
+base_images:
+  golangci-lint:
+    name: golangci-lint
+    namespace: ci
+    tag: v1.37.1
 build_root:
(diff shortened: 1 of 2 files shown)


danmanor:
@osherdp I think I added the config directly in the file instead of adding it in some other file and generate this config from there. If I am correct, can you point me to the file?

osherdp:
> @osherdp I think I added the config directly in the file instead of adding it in some other file and generate this config from there. If I am correct, can you point me to the file?

you need to add the relevant code under https://github.com/openshift/release/blob/c973cbbfafc2d65338f128cbb66f88ab0f793a79/ci-operator/config/openshift-assisted/assisted-events-stream/openshift-assisted-assisted-events-stream-master.yaml#L17

osherdp:
> /lgtm - @danmanor please make sure we merge the lint fixes in the target repo before this or it would be blocking

+1
you can also do ``/pj-rehearse`` to make sure the new job passes (after merging the lint re-formatting PR)

adriengentil:
Is this PR still valid?

danmanor:
@adriengentil no thanks

PR #38403, opened 2023-04-17
Removes gke-kernel-qa-e2e-test from stackrox config

Stringy:
As part of kernel module collection removal in stackrox, this test has been removed. It will be followed with a PR to remove the [gke_kernel_qa_e2e_tests.py](https://github.com/stackrox/stackrox/blob/master/scripts/ci/jobs/gke_kernel_qa_e2e_tests.py) file


ci-operator/jobs/stackrox/stackrox/stackrox-stackrox-master-presubmits.yaml
```diff
@@ -1,7 +1,7 @@
 presubmits:
   stackrox/stackrox:
   - agent: kubernetes
-    always_run: false
+    always_run: true
```
gavin-stackrox, line 4: Not sure how these changes came with your PR but we would not want them.

gavin-stackrox (changes requested):
Also take a look at https://github.com/openshift/release/blob/master/ci-operator/config/stackrox/stackrox/stackrox-stackrox-release-x.y.yaml per https://docs.engineering.redhat.com/display/StackRox/Making+changes+to+OpenShift+CI#MakingchangestoOpenShiftCI-PullRequestchecklist

gavin-stackrox:
@Stringy is there a doc / roadmap to remove COLLECTION_METHOD=kernel from ACS releases? I'd want to double check that we want to do this prior to 4.1.

Stringy:
@gavin-stackrox this is scheduled for 4.1, here's the epic: https://issues.redhat.com/browse/ROX-14421 and we're tracking this through a github project here: https://github.com/orgs/stackrox/projects/6 

The plan is to first disable all uses of kernel modules in testing on the main and collector and then begin removing the collection method from both.

gavin-stackrox (approved):
/lgtm
And welcome back to OpenShift CI ;)

PR #38410, opened 2023-04-17
[LPTOCPCI-168] Update MTR Test Branch to "v1.0" and Change Slack Notifications

calebevans:
The MTR scenario encountered an error this morning due to the branch needing to be changed. We also want to add Slack notifications to the PQE team.


sshveta (approved):
lgtm

PR #38087, opened 2023-04-05
USHIFT-1054 - MicroShift reboot test should not ignore wait error

copejon:
[USHIFT-1054](https://issues.redhat.com//browse/USHIFT-1054)

PR #38415, opened 2023-04-17
MicroShift - increase test-instance workload storage

copejon:
The release-4.12 branch of MicroShift is configured to reserve 10Gb of storage away from workloads, while the workload storage space is also 10Gb.  This PR increases the storage to 20Gb to allow release-4.12 stateful workloads to pass.


copejon:
> Is this linked to a Jira ticket?

I've roped this fix into #38087 so it's no longer relevant. Will close.

PR #38463, opened 2023-04-18
Move all channel bound jobs from stable to fast channels

sdodson:
These were missed from #38347


amisevsk (approved):
Not sure why my review is requested, but changes look fine to me.

cc: @musienko-maxim

andrewballantyne:
/un-cc @andrewballantyne 
I no longer work on OpenShift. I suspect an OWNERS file was not updated.

PR #38489, opened 2023-04-19
refine stable ci jobs pool to cover some specific configurations

jianlinliu:


jianlinliu:
cc @jhou1 @liangxia @YuLi517 @yunjiang29 to review.

jianlinliu:
@liangxia next time when you need to enable stable testing on new versions, you can copy the existing stable ci yaml file for new versions instead of copying nightly ci jobs, our stable ci pool would be relative fixed now.

PR #38499, opened 2023-04-19
Cleanup duplicate linitng code : ironic

bshewale:
As we are added pre-commit job in the same file which basically covers all the linting checks. So removed the duplicate ones from the same file.


viroel (approved):
lgtm, thanks

viroel:
/pj-rehearse ack

this patch only delete jobs

PR #38534, opened 2023-04-20
Cleanup duplicate linitng code: infra

bshewale:
As we are added pre-commit job in the same file which basically covers all the linting checks. So removed the duplicate ones from the same file.


viroel:
We need to merge https://github.com/openstack-k8s-operators/infra-operator/pull/28 first

viroel (approved):
golangci will be added to pre-commit in https://github.com/openstack-k8s-operators/infra-operator/pull/28

PR #38552, opened 2023-04-20
GitOps-service E2E update.

jparsai:
## Description

This PR is to fix E2E, broken because of path change done in https://github.com/redhat-appstudio/managed-gitops/pull/414


jgwest (approved):
LGTM, thanks @jparsai!

PR #38583, opened 2023-04-21
Add golangci-lint tests in CMO

danielmellado:
This commits enables golangci-lint in cluster-monitoring-operator gates
as a new optional check in prow.

Signed-off-by: Daniel Mellado <dmellado@redhat.com>

### DIFF
--- a/ci-operator/config/openshift/cluster-monitoring-operator/openshift-cluster-monitoring-operator-master.yaml
+++ b/ci-operator/config/openshift/cluster-monitoring-operator/openshift-cluster-monitoring-operator-master.yaml
@@ -77,6 +77,13 @@ tests:
   commands: make test-unit
   container:
     from: src
+- as: golangci-lint
+  commands: |
+    export GOCACHE=/tmp/
+    export GOLANGCI_LINT_CACHE=/tmp/.cache
+    make golangci-lint
+  container:
+    from: src
 - as: e2e-aws-ovn
   steps:
     cluster_profile: aws
--- a/ci-operator/jobs/openshift/cluster-monitoring-operator/openshift-cluster-monitoring-operator-master-presubmits.yaml
+++ b/ci-operator/jobs/openshift/cluster-monitoring-operator/openshift-cluster-monitoring-operator-master-presubmits.yaml
@@ -389,6 +389,53 @@ presubmits:
     - ^master$
     - ^master-
     cluster: build03
+    context: ci/prow/golangci-lint
+    decorate: true
+    labels:
+      ci.openshift.io/generator: prowgen
+      pj-rehearse.openshift.io/can-be-rehearsed: "true"
+    name: pull-ci-openshift-cluster-monitoring-operator-master-golangci-lint
+    rerun_command: /test golangci-lint
+    spec:
+      containers:
+      - args:
+        - --gcs-upload-secret=/secrets/gcs/service-account.json
+        - --image-import-pull-secret=/etc/pull-secret/.dockerconfigjson
+        - --report-credentials-file=/etc/report/credentials
+        - --target=golangci-lint
+        command:
+        - ci-operator
+        image: ci-operator:latest
+        imagePullPolicy: Always
+        name: ""
+        resources:
+          requests:
+            cpu: 10m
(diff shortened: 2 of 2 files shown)


danielmellado:
I'm skipping this as I expect this to fail for now until we fix violations on CMO, but this looks good from the config side. Thanks!

jan--f:
This looks good, just one thing :)
Not sure we should merge a failing test. I understand the fixes to make this happy are tracked in https://issues.redhat.com/browse/MON-3165.
Can we add exceptions for the known violations for now and remove those as we fix the issues? This way we could make this test required and avoid merging new violations.

danielmellado:
> This looks good, just one thing :) Not sure we should merge a failing test. I understand the fixes to make this happy are tracked in https://issues.redhat.com/browse/MON-3165. Can we add exceptions for the known violations for now and remove those as we fix the issues? This way we could make this test required and avoid merging new violations.

Violations/small refactor done xD all green!

ci-operator/config/openshift/cluster-monitoring-operator/openshift-cluster-monitoring-operator-master.yaml
```diff
@@ -77,6 +77,13 @@ tests:
   commands: make test-unit
   container:
     from: src
+- as: golangci-lint
+  commands: |
+    export GOCACHE=/tmp/
```
machine424, line 82: I think the env vars are based on `$HOME`, we can just set this?

ci-operator/config/openshift/cluster-monitoring-operator/openshift-cluster-monitoring-operator-master.yaml
```diff
@@ -77,6 +77,13 @@ tests:
   commands: make test-unit
   container:
     from: src
+- as: golangci-lint
```
machine424, line 80: why not just `lint`?

ci-operator/config/openshift/cluster-monitoring-operator/openshift-cluster-monitoring-operator-master.yaml
machine424: (The closer the command is to what we (may) run locally, the easier it is to debug I think)

ci-operator/config/openshift/cluster-monitoring-operator/openshift-cluster-monitoring-operator-master.yaml
danielmellado: to be explicit that this is running golancgi-lint :D

machine424: https://github.com/openshift/release/pull/39954#discussion_r1214217382 :)

machine424: but yeah,  we could have called the makefile target just `lint`. (this is just a detail, just to be consistent with other observability projects, you can ignore this)

danielmellado: Heh, not really a strong opinion on that. I'd stick to golangci-lint to be explicit as I said but feel free to put up a PR to change this if you'd like it ;)

ci-operator/config/openshift/cluster-monitoring-operator/openshift-cluster-monitoring-operator-master.yaml
danielmellado: Take into consideration that this is changing *specific* CI config. In this case, due to the overall prow config we'll be getting into permissions issues unless we do this. As you see, it's basically a cache thing so you'll notice no difference at all when running locally. In fact, you can even use this exact directories locally. I get your point but we wouldn't be able to apply this here.

danielmellado:
I wouldn't think too much on that, totally nitpicking and it's just a)
running a linter and b) check the rehearsal. Happy to have a follow up PR
with that if you have strong opinions about that.

PR #38585, opened 2023-04-21
remove: hypershift-rosa-mgmt e2e

LiangquanLi930:
Because currently we have covered mgmt e2e on the self-managend side, so we remove them


LiangquanLi930:
@liangxia Could you help to review this PR, When your have time

PR #38601, opened 2023-04-21
Remove upi TP featureset

sgaoshang:
Looks like we only run UPI zone jobs on OCP 4.13+, so it's better to remove TP featureSet.

### DIFF
--- a/ci-operator/step-registry/upi/conf/vsphere/zones/upi-conf-vsphere-zones-commands.sh
+++ b/ci-operator/step-registry/upi/conf/vsphere/zones/upi-conf-vsphere-zones-commands.sh
@@ -105,7 +105,6 @@ fi
 
 ${platform_required} && cat >> "${install_config}" << EOF
 baseDomain: $base_domain
-featureSet: TechPreviewNoUpgrade
 controlPlane:
   name: "master"
   replicas: 3


sgaoshang:
/pj-rehearse periodic-ci-openshift-release-master-nightly-4.13-e2e-vsphere-upi-zones

sgaoshang:
@jianlinliu pls hep review it, thanks.

ci-operator/step-registry/upi/conf/vsphere/zones/upi-conf-vsphere-zones-commands.sh
```diff
@@ -105,7 +105,6 @@ fi
 
 ${platform_required} && cat >> "${install_config}" << EOF
 baseDomain: $base_domain
-featureSet: TechPreviewNoUpgrade
```
sgaoshang, line 108: @rvanderp3 Hi, could you help to take a look if we can remove TP featureSet in UPI workflow, thanks.

rvanderp3: yep! we can.  great catch.

/lgtm
/approve

PR #38622, opened 2023-04-21
add CLC create and destroy steps to OPP

vboulos:
This PR adds the Cluster LifeCycle create and destroy tests to OPP

### DIFF
--- a/ci-operator/config/stolostron/policy-collection/stolostron-policy-collection-main__opp-ocp4.12-interop.yaml
+++ b/ci-operator/config/stolostron/policy-collection/stolostron-policy-collection-main__opp-ocp4.12-interop.yaml
@@ -63,9 +63,9 @@ tests:
     - ref: acm-mch
     - ref: acm-policies-openshift-plus-setup
     - ref: acm-policies-openshift-plus
-    - chain: cucushift-installer-check-cluster-health
     - ref: acm-tests-clc-create
     - ref: acm-fetch-managed-clusters
+    - chain: cucushift-installer-check-cluster-health
     - ref: quay-tests-quay-interop-test
 zz_generated_metadata:
   branch: main


ci-operator/config/stolostron/policy-collection/stolostron-policy-collection-main__opp-ocp4.12-interop.yaml
```diff
@@ -55,6 +64,8 @@ tests:
     - ref: acm-policies-openshift-plus-setup
     - ref: acm-policies-openshift-plus
     - chain: cucushift-installer-check-cluster-health
+    - ref: acm-tests-clc-create
```
gparvin, line 67: Could we move the CLC  create after the `acm-policies-openshift-plus`?  I think all of the other "product" tests should run with the managed clusters at least defined.

ci-operator/config/stolostron/policy-collection/stolostron-policy-collection-main__opp-ocp4.12-interop.yaml
```diff
@@ -55,6 +64,8 @@ tests:
...
+    - ref: acm-fetch-managed-clusters
```
gparvin, line 68: Does this just validate the managed clusters are setup?  Could we make it the last test in the `pre` list to make sure OPP validations do not impact managed cluster availability?  I'm not sure if this needs to be immediately after the CLC setup or even if may it could run twice -- after CLC and then as the last test.

gparvin (commented):
I just added a couple ideas around these tests.  Wondering what you think.

ci-operator/config/stolostron/policy-collection/stolostron-policy-collection-main__opp-ocp4.12-interop.yaml
vboulos: acm-fetch-managed-clusters queries the managed clusters, generates their credentials and stores them.
Those credentials will be used as input parameters for other components such as ALC, GRC and OBS

ci-operator/config/stolostron/policy-collection/stolostron-policy-collection-main__opp-ocp4.12-interop.yaml
vboulos: Yes, that's totally fine

gparvin (approved):
Thanks!

PR #38550, opened 2023-04-20
Assisted-Installer: Move assisted test-infra jobs to 4.13

eliorerz:
/cc @eliorerz

PR #38543, opened 2023-04-20
MGMT-14241: Set default OCP for late binding kubeapi sno test to 4.13

paul-maidment:
If we leave the late binding test on the default of 4.12 then we encounter an issue where the MCO will fail to come online with an error in the master machine config pool. 

"getting pending state from journal: invalid character 'U' looking for beginning of value"

According to the MCO team, the error occurs when booting discovery from an OCP 4.13 discovery image and then installing OCP 4.12. For more information, please see the following conversation.

https://redhat-internal.slack.com/archives/C02CZNQHGN8/p1681824679379279

To address this, we altered the late binding kubeapi sno test to install OCP 4.13 by default.

PR #38628, opened 2023-04-23
Revert "MGMT-14241: Set default OCP for late binding kubeapi sno test to 4.13"

eliorerz:
Reverts openshift/release#38543

Reasons for reverting this PR:

- We currently have change for 4.13 PR in testing - https://github.com/openshift/release/pull/38550
- 4.13 can be defined using `OPENSHIFT_VERSION` within the test configuration, and no need to create nightly-4-13 for that.
- What about 4.12? late binding will not work there? That needs to be addressed before changing the OCP version to 4.13 and ignoring that failure. It is better to have it fail on 4.12 (with an open bug) than to hide the failure results (on 4.13).

/cc @osherdp @adriengentil @gamli75 @paul-maidment

### DIFF
--- a/ci-operator/config/openshift/assisted-service/openshift-assisted-service-master__edge.yaml
+++ b/ci-operator/config/openshift/assisted-service/openshift-assisted-service-master__edge.yaml
@@ -155,11 +155,6 @@ releases:
       product: ocp
       stream: nightly
       version: "4.12"
-  nightly-4-13:
-    candidate:
-      product: ocp
-      stream: nightly
-      version: "4.13"
   stable-4-10:
     integration:
       name: "4.10"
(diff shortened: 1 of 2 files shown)


adriengentil:
> What about 4.12? late binding will not work there? That needs to be addressed before changing the OCP version to 4.13 and ignoring that failure. It is better to have it fail on 4.12 (with an open bug) than to hide the failure results (on 4.13).

This job hits this bug https://issues.redhat.com/browse/OCPBUGS-11492, that's why is works only with 4.13 at the moment.

Since this job is supposed to run as a required presubmit job on `assisted-test-infra`, I think it's important to unlock it otherwise the auto-update of the dependencies will fail.

This is a temporary fix, until we have OCP 4.13 and/or the bug fixed.

PR #38650, opened 2023-04-24
Fix OCPQE-14901

liangxia:


ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.13__amd64-nightly-4.13-upgrade-from-stable-4.1.yaml
```diff
@@ -14,11 +14,11 @@ base_images:
   tests-private-postupg:
     name: tests-private
     namespace: ci
-    tag: "4.13"
+    tag: "4.12"
```
kuiwang02, line 17: @liangxia postupg should be 4.13. it is executed after upgraded to 4.13.

same comment for other code in this PR.

liangxia: Good catch. Misread with the cucushift one.

kuiwang02:
I hold it because it has the label to be merged.
but need to check the comment https://github.com/openshift/release/pull/38650#discussion_r1175084241

PR #38687, opened 2023-04-24
Add new project in prow crc-org/vfkit

baude:
i would like to add a new project to prow and the openshift merge control.


jmguzik (approved):

ci-operator/jobs/crc-org/vfkit/OWNERS
```diff
@@ -0,0 +1,7 @@
+approvers:
+  - baude
+  - cfergeau
+reviewers:
```
gbraad, line 4: +should include: `gbraad`

PR #38710, opened 2023-04-25
Add openshift/cluster-api-provider-metal3

honza:
cc @derekhiggins @elfosardo


ci-operator/config/openshift/cluster-api-provider-metal3/openshift-cluster-api-provider-metal3-main.yaml
```diff
@@ -0,0 +1,41 @@
...
+resources:
+  '*':
+    limits:
+      memory: 4Gi
+    requests:
+      cpu: 100m
+      memory: 200Mi
+test_binary_build_commands: make unit
+tests:
+- as: e2e-aws
```
elfosardo, line 34: this should probably not be here, but it's ok, we can change the tests later as far as we're good with the rest

PR #38724, opened 2023-04-25
OSSM-3734: Move header-append-filter push image postsubmit jobs

yxun:


ci-operator/config/maistra/header-append-filter/maistra-header-append-filter-maistra-2.2.yaml
```diff
@@ -20,6 +20,102 @@ tests:
...
+  cluster_claim:
+    architecture: amd64
+    cloud: aws
+    owner: openshift-ci
+    product: ocp
+    timeout: 1h0m0s
+    version: "4.10"
+  steps:
+    env:
+      MAISTRA_BUILDER_IMAGE: registry.ci.openshift.org/ci/maistra-builder:2.4
```
jwendell, line 33: It looks like 2.4 is used everywhere. Is it on purpose?

jwendell (commented):
LGTM, feel free to add the postsubmit:true back

ci-operator/config/maistra/header-append-filter/maistra-header-append-filter-maistra-2.2.yaml
yxun: that's from maistra/test-infra. I was wondering why as well. When I saw the builder image tag 2.4 in test-infra, I try to keep it as it was.

jwendell (approved):

PR #38729, opened 2023-04-25
vSphere: set 4.8 and 4.9 to yearly

jcpowermac:
4.8 and 4.9 are EOL.
We do not have capacity in vSphere
to continue EOL OCP versions.


vrutkovs (approved):
/approve

PR #38753, opened 2023-04-26
OCPCLOUD-2028: Remove old AWS CCM tests

RadekManak:
CCM on AWS is GA in 4.14. We no longer need to run tests with CCM feature flag, as it is deployed by default.

Removed all jobs that matched "aws.*ccm" in files matching ".*master.yaml"

TechPreview jobs on CCM and CCCMO were already removed in previous PR: https://github.com/openshift/release/pull/37217


JoelSpeed (commented):
What is the status of the techpreview jobs for AWS on CCMO and the cloud-provider-aws repositories? Maybe we want to make those on-demand by setting `always_run: false` going forward? Since there should be no-diff between the two, WDYT?

RadekManak:
They were already removed by https://github.com/openshift/release/pull/37217

PR #38764, opened 2023-04-26
OpenStack: Add alternative folks for backports

dulek:
This commit adds an alternative person to be able to label PRs with `backport-risk-assessed` for several OpenStack projects.


dulek:
@mandre, @mdbooth: Looks like your vote can move this through.

mdbooth:
Lets see which of these work:

/approve
/lgtm

mandre (commented):
I'm fine sharing the responsibility of setting the `backport-risk-assessed` with the devs mentioned here as they're clearly experts in their respective areas, however I was under the impression we already had a process in place to avoid depending on a single person (me in this case). We will have to dig the topic a bit more.

PR #38766, opened 2023-04-26
filter out usless pj-rehearse error from logs

smg247:
We don't need errors like this to appear when we look for errors in the logs.


docs/dptp-triage-sop/logs/filter.py
```diff
@@ -1,8 +1,7 @@
 #!/usr/bin/env python3
 
-import sys
-import os
```
smg247, line 4: My IDE made this change for me, but the command works fine like this, so it appears `os` is no longer needed

droslean (approved):

PR #38777, opened 2023-04-26
[knative-client] Add CI jobs for release-v1.8 branch

dsimansk:
Per title adding job for next release.

/cc @rhuss or @matzew @mgencur @pierDipi


rhuss (commented):
/approve
/lgtm

Is this already part of the release cut-off ? I would like to look a bit over your shoulder and we might track the step in Google docs (that what we have agreed during the release sig call). Not sure though, where to start.

PR #38783, opened 2023-04-27
ci-operator/gcp: set CLOUDSDK_PYTHON to use python3

sohankunkerkar:
I think this change might fix the periodic jobs for building images for CRI-O CI


sohankunkerkar:
@jianlinliu PTAL

PR #38787, opened 2023-04-27
ci-operator/config/openshift/release: oldest-supported-z-rollback 4.12.0

wking:
Nobody will be rolling back from recent 4.12.z to the prerelease rc.0. Update to use the oldest supported 4.12.z instead.


wking:
I'm confident that this change is safe:

/pj-rehearse ack

shellyyang1989:
/lgtm
for the change

PR #38803, opened 2023-04-27
increase cpu and memory requirements for ci-search

hoxhaeris:
The requested CPU and memory values are changed to reflect the actual usage. Currently, we do not/will not limit the CPU and memory consumption.


bradmwilliams (approved):

bradmwilliams:
@hoxhaeris Please watch as these changes roll out!  There is a possibility that the pods wont schedule due to the massive memory ask.

PR #38816, opened 2023-04-27
Initial Configuration backplane 2.4

Kyl-Bempah:
/hold


Kyl-Bempah:
@qiujian16 @subbarao-meduri @JakobGray @KevinFCormier @rokej

dhaiducek:
- Should we remove the `release-2.4` configurations entirely rather than just removing the promotion stanzas?
- Also, in the past the latest version was promoted from `main`, though I don't know the background nor the consequences on that--I've just been blindly doing it up until now.

dhaiducek:
@gurnben 
1. A separate PR works for me! Since the `2.4` promotions no longer conflict, I don't think there's harm in leaving the configurations in place, either. I've been thinking about cleaning up GRC repos in the back of my mind lately just to keep things streamlined as our configurations grow.
2. Gotcha. Yeah, then if it's not relevant any more, that sounds good to me! I was wondering if there might be cases where a squad might use the promoted images from `main` in tests or something before fast-forwarding, but that's probably a squad-by-squad thing (and might not even be a thing).

PR #38818, opened 2023-04-27
fix CI jobs for logging fluent 5.8

jcantrill:
This PR fixes the CI jobs for logging-fluent 5.8 release of logging


jcantrill:
/pj-rehearse ack

* e2e job fails because we need https://github.com/openshift/cluster-logging-operator/pull/1962
* lint fails because we need a RP to fix up linting

PR #38829, opened 2023-04-27
ci-operator/cri-o: delete CI images older than 2 weeks

sohankunkerkar:


sohankunkerkar:
@haircommander @mrunalp PTAL

PR #38847, opened 2023-04-28
openstack: Inject oc in conf-clouds

pierreprinetti:
By injecting the `oc` tool in the step definition, we remove an implicit dependency on the base image.


pierreprinetti:
We're fine with the dependency on the base image for now.
/close

PR #38848, opened 2023-04-28
NTO: Bump e2e-gcp-pao timeout for 4.11

jmencak:
e2e-gcp-pao tests are timing out.  Use the same timeout value for them as in the 4.12 branch.


ffromani:
/lgtm

PAO maintainers will keep the investigation and improve the lane, but for now we need this fix.

PR #38860, opened 2023-04-28
Add a PVC for /var/lib/grafana

dgoodwin:
This will allow us to manually install plugins (bigquery is needed), and live persist our dashboards as we iterate on them. We may move to a gitops style for these later, but during creation it will be much faster to iterate in the UI live.

Adding myself to owners file as TRT is taking over this component for loki (which we hope to have back online this week), charting disruption data from bigquery, and possibly some from the sippy postgresql db as well.

### DIFF
--- a/clusters/app.ci/grafana-loki/OWNERS
+++ b/clusters/app.ci/grafana-loki/OWNERS
@@ -2,3 +2,4 @@ approvers:
   - vrutkovs
   - wking
   - periklis
+  - dgoodwin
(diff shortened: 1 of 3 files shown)


dgoodwin:
/hold

I'm going to add a pvc for all of /var/lib/grafana so the db is also persisted. I think I'd like to allow live modifications to dashboards to maintain usability for at least the first while, then perhaps we can move to a gitops style of deploying them.

dgoodwin:
/hold cancel

Thanks Stephen!

PR #38870, opened 2023-04-28
USHIFT: disable arm testing

copejon:


dhellmann (commented):
I see main, 4.13, and 4.15 here but not 4.14. Is that expected?

copejon:
> I see main, 4.13, and 4.15 here but not 4.14. Is that expected?

No, looks like an oversight. Fixed it

PR #38898, opened 2023-05-01
Increase e2e-handler-sdn-ipv4 job timeout to 6h

jluhrsen:
some jobs are failing due to the default 4h timeout being reached.


jluhrsen:
guessing this wasn't needed after all

PR #38911, opened 2023-05-02
Assisted-Installer: Remove slaviered from assisted projects

eliorerz:
/cc @osherdp


ci-operator/config/openshift-assisted/assisted-installer-deployment/OWNERS
```diff
@@ -34,7 +34,6 @@ filters:
     - ori-amizur
     - osherdp
     - romfreiman
-    - slaviered
```
osherdp, line 37: it will get overridden from the repos, so I think there's no point to do it here

osherdp (changes requested):

PR #38913, opened 2023-05-02
openstack bugwatcher: Move notifications to team channel

pierreprinetti:


MaysaMacedo:
What is the purpose of the change?

pierreprinetti:
> What is the purpose of the change?

It’s a proposal to unclutter our public channel. WDYT?

MaysaMacedo:
@pierreprinetti Perhaps it's a good idea to check if the owner of the new channel is open to the change

pierreprinetti:
> @pierreprinetti Perhaps it's a good idea to check if the owner of the new channel is open to the change

you mean @EricArrakis ?

EricArrakis:
> > @pierreprinetti Perhaps it's a good idea to check if the owner of the new channel is open to the change
> 
> you mean @EricArrakis ?

WRT moving the GH slack notifications from forum-shiftstack (btw how does this work since forum-shiftstack doesn't exist anymore) to team-shift-on-stack...I'd defer to the team. What best suits your needs? I can certainly see the argument that those notifications largely only affect the engineering team and maybe it should then be limited to the engineering specific channel.  I can't think off the top of my head if there is value in being in the public channel, maybe folks on the team would have better insight, and thus be better suited to chime in on this change.

pierreprinetti:
>how does this work since forum-shiftstack doesn't exist anymore

The secret `forum-shiftstack` in the CI vault contains a URL that points to the channel by ID, and not by name. TL;DR we can rename the channel in Slack; the SLACK_HOOK remains valid but its name in the CI will not change.

PR #38929, opened 2023-05-02
enabling promotion of agent-installer-utils image

andfasano:
Allow the promotion of the agent-installer-utils image so that it could be included in the release payload


andfasano:
Not required

PR #38951, opened 2023-05-03
Move OpenStack CCCMO jobs to vh-managed

mandre:
vh-mecha-central was recently re-deployed and we added a proxy to reach the provider network where the FIPs are. This caused the LB UDP tests to fail as it's not compatible with the environment.

Move the jobs to vexxhost managed cloud, where we have public IP addresses for the FIPs.


mandre:
@JoelSpeed if you were wondering why the OpenStack LB tests started failing in your PRs.

/assign @JoelSpeed

mandre:
/pj-rehearse pull-ci-openshift-cluster-cloud-controller-manager-operator-release-4.14-e2e-openstack-ovn

EmilienM:
@mandre the thing that I worry is that managed VEXXHOST doesn't have multiple AZ therefore we won't use that cloud to test failure domains in CCPMSO... I wonder why we need that job then?
I think we should work on making that job running on mecha in multiple AZ.

damdo (commented):
/approve
/lgtm

elmiko (commented):
/approve

PR #38960, opened 2023-05-03
Use oauth proxy sidecar for Grafana

dgoodwin:
It appears Grafana is not capable of using fixed oauth token, only the jwt obtained from the users SSO login, which is not authorized on our loki instance. We need to add a proxy sidecar (same as we do for promtail) to inject the oauth authorization header.


dgoodwin:
/close 

superceded by #39003

PR #38979, opened 2023-05-03
Rework a bit the pluginconfig associated to ci-framework

cjeanner:
Cleaning up a bit the active plugins, and switching the self_approval
requirement.

### DIFF
--- a/core-services/prow/02_config/openstack-k8s-operators/ci-framework/_pluginconfig.yaml
+++ b/core-services/prow/02_config/openstack-k8s-operators/ci-framework/_pluginconfig.yaml
@@ -1,8 +1,9 @@
 approve:
 - commandHelpLink: ""
+  ignore_review_state: true
   repos:
   - openstack-k8s-operators/ci-framework
-  require_self_approval: false
+  require_self_approval: true
 external_plugins:
   openstack-k8s-operators/ci-framework:
   - endpoint: http://refresh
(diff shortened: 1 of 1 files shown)


core-services/prow/02_config/openstack-k8s-operators/ci-framework/_pluginconfig.yaml
```diff
@@ -2,7 +2,8 @@ approve:
 - commandHelpLink: ""
   repos:
   - openstack-k8s-operators/ci-framework
-  require_self_approval: false
+  require_self_approval: true
+  lgtm_acts_as_approve: false
```
viroel, line 6: it seems that OpenShift CI will block us from merging this configuration. the `make prow-config` target deletes this line:
`-  lgtm_acts_as_approve: false`

Sandeepyadav93: Based on the discussion with Stephen from DPTP team, `lgtm_acts_as_approve: false` is the default so no need to mention it and our patches were getting merged because approvers were using the github option to approve PRs, `ignore_review_state: true` will help us with that.

more about this parameter is here: https://github.com/kubernetes/test-infra/blob/master/prow/plugins/config.go#L316-L319

~~~
	// IgnoreReviewState causes the approve plugin to ignore the GitHub review state. Otherwise:
	// * an APPROVE github review is equivalent to leaving an "/approve" message.
	// * A REQUEST_CHANGES github review is equivalent to leaving an /approve cancel" message.
~~~

cjeanner: :+1:

PR #39007, opened 2023-05-04
CAPO: use rhel8 base image to run unit tests

mandre:
The golang-1.17 image is based on centos 7, which is too old for our
needs. Use a rhel8 based image instead.

Also unset GOFLAGS to remove `-mod=vendor` since that prevents us from
building envtest, and set XDG variables to workaround permission issues.

### DIFF
--- a/ci-operator/config/openshift/cluster-api-provider-openstack/openshift-cluster-api-provider-openstack-main.yaml
+++ b/ci-operator/config/openshift/cluster-api-provider-openstack/openshift-cluster-api-provider-openstack-main.yaml
@@ -2,7 +2,7 @@ build_root:
   image_stream_tag:
     name: release
     namespace: openshift
-    tag: golang-1.17
+    tag: rhel-8-release-golang-1.19-openshift-4.14
 resources:
   '*':
     requests:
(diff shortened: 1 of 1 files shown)


mandre:
This is ready for review
/hold cancel
/assign @pierreprinetti

ci-operator/config/openshift/cluster-api-provider-openstack/openshift-cluster-api-provider-openstack-main.yaml
```diff
@@ -2,15 +2,16 @@ build_root:
...
+    tag: rhel-8-release-golang-1.19-openshift-4.14
 resources:
   '*':
     requests:
       cpu: 100m
       memory: 200Mi
 tests:
 - as: unit
-  commands: make test
+  commands: GOFLAGS= XDG_DATA_HOME=/tmp/.local/share XDG_CACHE_HOME=/tmp/.cache make
```
pierreprinetti, line 13: Shouldn't we vendor dependencies (as a downstream patch) and remove the GOFLAGS override?

mandre: Possibly, but since we're only using CAPO as a library in MAPO (we vendor in MAPO), is it important?

PR #39012, opened 2023-05-04
hypershift: add HO install flags for CI

sjenning:
So we can remove the magic here
https://github.com/openshift/hypershift/blob/main/cmd/install/install.go#L151-L154


sjenning:
/pj-rehearse periodic-ci-openshift-hypershift-release-4.14-periodics-e2e-aws-ovn

PR #39034, opened 2023-05-05
Ensure windows-user-data secret creation.

jfrancoa:
This patch improves the check for windows-user-data secret, adding a busy loop until it gets created. If it didn't the job will fail after reaching the job time out.

It was observed in [4.12](https://qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/qe-private-deck/pr-logs/pull/openshift_release/37431/rehearse-37431-pull-ci-openshift-openshift-tests-private-release-4.12-debug-winc-gcp-ipi/1651856965418094592) and [master](https://qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/qe-private-deck/pr-logs/pull/openshift_release/37431/rehearse-37431-pull-ci-openshift-openshift-tests-private-master-debug-winc-gcp-ipi/1651857086402793472) Windows Containers rehearse executions, that the job was failing because the windows-user-data secret was not created in openshift-machine-api namespace. When digging into the WMCO logs, we could see that WMCO didn't have time enough to create it and the job failed after the first check:
`{"level":"info","ts":"2023-04-28T08:59:12Z","logger":"controller.secret","msg":"secret not found, creating the secret","secret":"openshift-windows-machine-config-operator/cloud-private-key","name":"windows-user-data"}`

Converting the simple check for windows-user-data secret existence into a busy-loop will make sure that the job continues only if the secret gets created, but it will give some extra time for the check at the same time.

### DIFF
--- a/ci-operator/step-registry/ipi/conf/aws/windows-machineset/ipi-conf-aws-windows-machineset-commands.sh
+++ b/ci-operator/step-registry/ipi/conf/aws/windows-machineset/ipi-conf-aws-windows-machineset-commands.sh
@@ -10,7 +10,7 @@ oc wait deployment windows-machine-config-operator -n openshift-windows-machine-
 # Ensure userDataSecret exist, fail otherwise. The userDataSecret is required and contains specific information to
 # customize the machine at first boot. For instance, the authorized public key for the SSH server to accept
 # incoming connections, firewall rules, etc.
-oc -n openshift-machine-api get secret "${WINDOWS_USER_DATA_SECRET}"
+timeout 3m bash -c 'until oc -n openshift-machine-api get secret "${WINDOWS_USER_DATA_SECRET}" 2> /dev/null; do echo -n "." && sleep 15; done'
 
 # config AWS profile
 export AWS_SHARED_CREDENTIALS_FILE="${CLUSTER_PROFILE_DIR}/.awscred"
--- a/ci-operator/step-registry/ipi/conf/gcp/windows-machineset/ipi-conf-gcp-windows-machineset-commands.sh
+++ b/ci-operator/step-registry/ipi/conf/gcp/windows-machineset/ipi-conf-gcp-windows-machineset-commands.sh
@@ -10,7 +10,7 @@ oc wait deployment windows-machine-config-operator -n openshift-windows-machine-
 # Ensure userDataSecret exist, fail otherwise. The userDataSecret is required and contains specific information to
 # customize the machine at first boot. For instance, the authorized public key for the SSH server to accept
 # incoming connections, firewall rules, etc.
-oc -n openshift-machine-api get secret "${WINDOWS_USER_DATA_SECRET}"
+timeout 3m bash -c 'until oc -n openshift-machine-api get secret "${WINDOWS_USER_DATA_SECRET}" 2> /dev/null; do echo -n "." && sleep 15; done'
 
 # Get machineset name to generate a generic template
 ref_machineset_name=$(oc -n openshift-machine-api get -o 'jsonpath={range .items[*]}{.metadata.name}{"\n"}{end}' machinesets | grep worker | head -n1)


ci-operator/step-registry/ipi/conf/aws/windows-machineset/ipi-conf-aws-windows-machineset-commands.sh
```diff
@@ -10,7 +10,7 @@ oc wait deployment windows-machine-config-operator -n openshift-windows-machine-
 # Ensure userDataSecret exist, fail otherwise. The userDataSecret is required and contains specific information to
 # customize the machine at first boot. For instance, the authorized public key for the SSH server to accept
 # incoming connections, firewall rules, etc.
-oc -n openshift-machine-api get secret "${WINDOWS_USER_DATA_SECRET}"
+until oc -n openshift-machine-api get secret "${WINDOWS_USER_DATA_SECRET}" 2> /dev/null; do echo -n "." && sleep 2; done
```
sebsoto, line 13: This is going to be an infinite loop if the secret was not created.
In that case this job will not fail until the 4ish hour timeout is hit.
There needs to be some reasonable limit to retries here, you can use a for loop for that.

sebsoto (changes requested):

ci-operator/step-registry/ipi/conf/aws/windows-machineset/ipi-conf-aws-windows-machineset-commands.sh
aravindhp: In addition please increase the time between API calls to 5 seconds.

aravindhp (changes requested):

ci-operator/step-registry/ipi/conf/gcp/windows-machineset/ipi-conf-gcp-windows-machineset-commands.sh
```diff
@@ -10,7 +10,7 @@ oc wait deployment windows-machine-config-operator -n openshift-windows-machine-
 # Ensure userDataSecret exist, fail otherwise. The userDataSecret is required and contains specific information to
 # customize the machine at first boot. For instance, the authorized public key for the SSH server to accept
 # incoming connections, firewall rules, etc.
-oc -n openshift-machine-api get secret "${WINDOWS_USER_DATA_SECRET}"
+until oc -n openshift-machine-api get secret "${WINDOWS_USER_DATA_SECRET}" 2> /dev/null; do echo -n "." && sleep 5; done
```
saifshaikh48, line 13: Consider increasing the sleep time. For what it's worth, WMCO waits 15 seconds before retrying API calls.

saifshaikh48: Also, if you want to fail fast here (rather than waiting on the entire job's timeout), consider using the [timeout command](http://manpages.courier-mta.org/htmlman1/timeout.1.html). example:
```
timeout 3m bash -c 'until oc -n openshift-machine-api get secret "${WINDOWS_USER_DATA_SECRET}" 2> /dev/null; do echo -n "." && sleep 15; done'
```

ci-operator/step-registry/ipi/conf/aws/windows-machineset/ipi-conf-aws-windows-machineset-commands.sh
jfrancoa: As suggested by @saifshaikh48 , I made use of the timeout function to ensure that if after 3 minutes the secret isn't created the step will fail.

jfrancoa:
/pj-rehearse pull-ci-openshift-openshift-tests-private-release-4.14-debug-winc-aws-ipi

jfrancoa:
/pj-rehearse pull-ci-openshift-openshift-tests-private-release-4.12-debug-winc-aws-ipi

jfrancoa:
/pj-rehearse pull-ci-openshift-openshift-tests-private-master-debug-winc-aws-ipi

jfrancoa:
/pj-rehearse pull-ci-openshift-openshift-tests-private-release-4.13-debug-winc-aws-ipi

jfrancoa:
ping @saifshaikh48 @sebsoto could you please have a look? all comments were already addressed.

saifshaikh48:
/lgtm

rehearsals look like they are failing, I don't have permission to view the output so not sure if that is expected

jfrancoa:
The rehearsals are failing due to few WINC tests which are failing (we still need to adapt our machinesets, as the naming convention is different in Prow's ref). But I checked all rehearsals and they are all installing WMCO and passing the stage of windows-user-data creation

PR #39038, opened 2023-05-05
remove duplicated workflow for fips and compact

jianlinliu:
Similar with https://github.com/openshift/release/pull/38872, we are configuring env var for fips and compact cluster at job config level to reuse the existing workflow to avoid duplicates.


jianlinliu:
cc @liangxia to review.

PR #39046, opened 2023-05-05
add api call for 4.12 upgrade jobs

jianzhangbjz:
As the title shows, add `remote_api` for all 4.12 upgrade jobs.


kuiwang02:
my understanding is that only test which is upgraded to stable build needs it.
and only test upgraded from stabe 4.11 needs it.

kuiwang02:
actually test we trigger is only for install and upgrade, not run case.
it means
for install, the test has chain cucushift-installer-check-cluster-health only, like
```yaml
    test:
```
for upgrade, the test has chain openshift-upgrade-qe-sanity, like
```yaml
    test:
    - chain: openshift-upgrade-qe-sanity
```

only these test should be triggered by api, so we'd better to add remote api for these tests only.

kuiwang02:
so, do we need to add remote api for all upgraded tests?

jianzhangbjz:
> do we need to add remote api for all upgraded tests?

Yes, because this `API Call` way can be used by other teams too, not only for errata. Originally, I want to update all Prow jobs, but for now, I just update the 4.12 upgrade jobs for them.

PR #38914, opened 2023-05-02
dpcr loki

dgoodwin:
- Only enable new loki for very specific set of jobs to start with
- Modify loki config to push to the new instance on DPCR

PR #39060, opened 2023-05-05
Revert "dpcr loki"

dgoodwin:
Reverts openshift/release#38914

Pod missing limits is breaking payloads.


dgoodwin:
/pj-rehease ack

No rehersals as this is a revert to known good state to get payloads back online.

PR #39068, opened 2023-05-05
Add e to list of gcp types we understand

sdodson:
Looks like the jobs are currently assuming max 8 vCPU, not entirely sure what that does but we're runnng these on e2-standard-4 so support those.


sdodson:
/pj-rehearse periodic-ci-openshift-release-master-ci-4.13-upgrade-from-stable-4.12-e2e-gcp-ovn-rt-upgrade

sdodson:
Looks like not having this machine type yields `isolated_cpus=4` where as having it defined yields `isolated_cpus=2` (half max_cpus)

https://github.com/openshift/release/blob/91579ab298e008a03ac05bce263ee27cf903f032/ci-operator/step-registry/optional-operators/configure-realtime-tuned-profile/optional-operators-configure-realtime-tuned-profile-commands.sh#L7-L27

PR #39087, opened 2023-05-06
update 4.12 stable jobs to support API call

jianzhangbjz:
As the title shows, following #39046 , only update those `stable` jobs, no `nightly`.


jianzhangbjz:
Hi @kuiwang02 @Xia-Zhao-rh , could you help approve it when you get a chance? Thanks!

PR #39106, opened 2023-05-08
[LPTOCPCI-210] Clean Up VPN PoC Resources

calebevans:
This cluster profile was used in a proof of concept and is no longer needed


danilo-gemoli:
/lgtm
Should we remove [this](https://github.com/openshift/ci-tools/blob/8598d1f45fe6ac0e9cf9e58385cf9bd2f8d220a9/pkg/api/types.go#L1130) as well?

calebevans:
@danilo-gemoli yes, we should, I created a PR to remove that at the same time as I created this PR: https://github.com/openshift/ci-tools/pull/3420

PR #39120, opened 2023-05-08
additional quota in vsphere 8

jcpowermac:


core-services/prow/02_config/generate-boskos.py
```diff
@@ -308,6 +308,8 @@
 
 for i in range(200,204):
     CONFIG['vsphere-8-quota-slice']['ci-segment-{}'.format(i)] = 1
+for i in range(205,214):
```
smg247, line 311: Why not just extend the above range?

jcpowermac: there is no 204

PR #39125, opened 2023-05-08
[core-services] Update prow and plugin configs

mansikulkarni96:
This PR removes the plugin approve from the plugin config and tide approve label requirement from prow config to allow manual merges in the prometheus-community-windows-exporter repo.
[openshift/prometheus-community-windows_exporter](https://github.com/openshift/prometheus-community-windows_exporter/) repository does not contain an OWNERS file, hence we need to go back to doing manual merges as automatic approval is not supported.
Ref:
https://docs.ci.openshift.org/docs/how-tos/onboarding-a-new-component/#enabling-automatic-merges
https://docs.ci.openshift.org/docs/how-tos/onboarding-a-new-component/#repositories-under-existing-organizations


aravindhp (changes requested):
@mansikulkarni96 thanks for doing this. Please mention why we want to go back to doing manual merges

aravindhp (approved):

PR #39140, opened 2023-05-09
update test Frequency

LiangquanLi930:
Currently, we can see many failure cases on this job but as it's f21. We cannot trigger a rerun to verify again immediately after the sub-team raised PR. This job is important to OCP 4.13 GA and HyperShift GA. So we need to increase frequency to f7 now to monitor for a while. When it is stable, we will consider reducing frequency later.


LiangquanLi930:

PR #39147, opened 2023-05-09
setup image build&push for runtime image on main branch

harshad16:
setup image build&push for runtime image on main branch

Reference: https://github.com/openshift/release/pull/38165/files
The image would be pushed on: https://quay.io/repository/modh/runtime-images


atheo89 (approved):
/lgtm

PR #39152, opened 2023-05-09
MON-3173: manually disable ci/prow/versions test on release branches …

machine424:
…from 4.10 to 4.15


machine424:
Thanks to you both.

@jan--f  Yes, as I mentioned in the ticket, this is not ideal, but will buy us some time to think about a better/automated solution.

I'll add the step to the template, I hope we'll not get used to doing this :)

machine424:
Some of this was reverted by a lovely bot https://github.com/openshift/release/pull/39366/files

I'll see

jan--f:
hm its possible that the brancher bot ensures that the release branches are tested just like the master branch is tested. Maybe we should fix this is our `make versions` target?

machine424:
Yep, I'm trying to apply solution #2 in here https://github.com/openshift/cluster-monitoring-operator/pull/1969

PR #39160, opened 2023-05-09
Enable loki for all aws/metal/vsphere jobs that had it previously

dgoodwin:


dgoodwin:
/pj-rehearse ack

No rehearsals necessary, we're just adding more jobs, the previous config should be well tested on what was covered prior.

PR #39189, opened 2023-05-10
update hypershift-extended-health-check

LiangquanLi930:


ci-operator/step-registry/cucushift/hypershift-extended/health-check/cucushift-hypershift-extended-health-check-commands.sh
```diff
@@ -12,6 +12,7 @@ set -x
 # It then reads the output of "oc get pod" command in the corresponding HostedCluster namespace and checks if the status is "Running" or "Completed".
 # If any pod is not in the expected state, it prints an error message and returns 1. Otherwise, it returns 0.
 function check_control_plane_pod_status {
+    HYPERSHIFT_NAMESPACE=$(oc get hostedclusters -A '-o=jsonpath={.items[0].metadata.namespace}')
```
heliubj18, line 15: we should add `--ignore-not-found`,  otherwise it would be crashed here

PR #39204, opened 2023-05-10
Fix the default loki time window in prow job debug tools links.

dgoodwin:
I mistakenly thought this ran after the job but it's actually a
pre-install step, so the loki links were prepopulated with a time window
that was missing most of the logs.

This window should be safer and include all of the logs for each job run
by default.

### DIFF
--- a/ci-operator/step-registry/ipi/install/hosted-loki/ipi-install-hosted-loki-commands.sh
+++ b/ci-operator/step-registry/ipi/install/hosted-loki/ipi-install-hosted-loki-commands.sh
@@ -693,9 +693,11 @@ echo "Promtail manifests created, the cluster can be found at https://grafana-lo
 
 
 if [[ -f "/usr/bin/python3" ]]; then
-  # Try to prepopulate the loki time window to match the job (with some leeway), so the user is never staring at no logs when they're actually there:
-  LOKI_EPOCH_MILLIS_FROM="$(date -d '-8 hours' +%s%N | cut -b1-13)"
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/ipi/install/hosted-loki/ipi-install-hosted-loki-commands.sh
```diff
@@ -693,9 +693,11 @@ echo "Promtail manifests created, the cluster can be found at https://grafana-lo
...
 
 if [[ -f "/usr/bin/python3" ]]; then
-  # Try to prepopulate the loki time window to match the job (with some leeway), so the user is never staring at no logs when they're actually there:
-  LOKI_EPOCH_MILLIS_FROM="$(date -d '-8 hours' +%s%N | cut -b1-13)"
-  LOKI_EPOCH_MILLIS_TO="$(date -d '+1 hours' +%s%N | cut -b1-13)"
+  # Try to prepopulate the loki time window to match the job (with some leeway), so the user is never staring at no logs when they're actually there.
+  # Note that this step runs prior to install, so we use a window 2 hours before now (to be ridiculously safe) and 8 hours after, which should cover
+  # the runtime of just about any job. There's no risk over overloading the UI or seeing logs you don't want because we filter by invoker (this job).
+  LOKI_EPOCH_MILLIS_FROM="$(date -d '-2 hours' +%s%N | cut -b1-13)"
+  LOKI_EPOCH_MILLIS_TO="$(date -d '+8 hours' +%s%N | cut -b1-13)"
```
dgoodwin, line 700: If anyone is curious, this cut is good until 2300 or so, so if anyone finds this in 80 years and is mad, my bad.

PR #39214, opened 2023-05-10
Allow console-operator feature development without bugzilla/valid-bug label

jhadvig:
/assign @spadgett

### DIFF
--- a/core-services/prow/02_config/openshift/console-operator/_prowconfig.yaml
+++ b/core-services/prow/02_config/openshift/console-operator/_prowconfig.yaml
@@ -68,10 +68,7 @@ tide:
     labels:
     - approved
     - bugzilla/valid-bug
-    - docs-approved
     - lgtm
-    - px-approved
-    - qe-approved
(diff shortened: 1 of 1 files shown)


spadgett:
/hold

We only have an exception for frontend changes. Holding for now.

core-services/prow/02_config/openshift/console-operator/_prowconfig.yaml
```diff
@@ -88,8 +85,10 @@ tide:
     - master
     labels:
     - approved
-    - bugzilla/valid-bug
+    - docs-approved
```
jerolimov, line 88: I recommend that we also add the requirement to link a jira ticket here:

```suggestion
    - docs-approved
    - jira/valid-reference
```

This works fine with all CONSOLE and ODC stories and makes it easier to find the reason for a change later.

jhadvig:
@jerolimov thanks for comment. It's addressed 👍

PR #39240, opened 2023-05-11
Fix the dhcp request payload of the HAProxy LB

aleskandro:


aleskandro:
Tested downstream

/pj-rehearse ack

PR #39246, opened 2023-05-11
HOSTEDCP-966: Gate more Operators on Hypeshift e2e

enxebre:


kpouget:
I don't know why I was picked up as a reviewer for this PR, seems to be a bug from Prow/Github 🤔

enxebre:
cc @dagrayvid

dagrayvid:
This seems like a great idea, thanks @enxebre. FYI @jmencak

ci-operator/config/openshift/cluster-ingress-operator/openshift-cluster-ingress-operator-master.yaml
```diff
@@ -43,6 +51,9 @@ tests:
   commands: make test
   container:
     from: src
+- as: e2e-hypershift
+  steps:
```
candita, line 55: Could we add `optional: true` until we know this doesn't introduce any issues for cluster-ingress-operator CI?

enxebre: We are actually intentionally setting as not optional at this point we GAing to avoid operators breaking us at it has happened in the past.
We are doing the same in multiple operators and we are also ci stream promotion blocker now. I think it's a fair organic step and we can always relax it back if it cause unncessary issues. Makes sense?

enxebre: This is the newest example https://github.com/openshift/cluster-ingress-operator/pull/930.
We need to get this merged asap right after fixing this new issue so teams are able to operate with healthy safety checks that ensure GA level stability.

enxebre:
Revert merged https://github.com/openshift/cluster-ingress-operator/pull/938
Let's see it go back green and then merge
/pj-rehearse pull-ci-openshift-cluster-ingress-operator-master-e2e-hypershift

enxebre:
> util.go:452: Container route-controller-manager in pod openshift-route-controller-manager-6b8747c96c-p6hzf has a restartCount > 0 (8)

I assume that was still using conflicting code

/pj-rehearse pull-ci-openshift-cluster-ingress-operator-master-e2e-hypershift

sjenning:
> I assume that was still using conflicting code

Nope a different break https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/288 and https://github.com/openshift/route-controller-manager/pull/22

PR #39261, opened 2023-05-11
Disable loki on proxy jobs, add remaining jobs that had loki before

dgoodwin:
We were mostly scaled up already, removing the temporary filter to let
whatever was hanging around in as well.

Proxy jobs are now disabled, we think this was broken before but now
with the new prod-bearer-token container, this crash loops because it
can't reach SSO, whereas we think previously the promtail container just
silently didn't do anything but didn't crash.

Maybe a future enhancement, but for now proxy disabled until we figure
out how to get all the egress working.

[TRT-1029](https://issues.redhat.com//browse/TRT-1029)


dgoodwin:
/pj-rehearse ack

As with prior prs, the config is well tested, random rehersals are unlikely to show anything as this impacts so many jobs.

PR #39262, opened 2023-05-11
Update Observability components to build release-2.9

dislbenn:
Cutting `release-2.9` branches for Observability components.


jlpadilla:
@dislbenn We also need to update the branch protection for `release-2.8` and `release-2.9` for all projects.
Check this file for a sample.
https://github.com/openshift/release/blob/master/core-services/prow/02_config/stolostron/search-v2-api/_prowconfig.yaml#L19-L20

PR #38755, opened 2023-04-26
Add tempest job for manila-operator

cescgina:

PR #39289, opened 2023-05-14
Build default share type when Manila is deployed

fmount:
This patch allows to execute a subset of openstack related commands to make sure `Manila` has a default `share type` after the deployment of the share service.
A default share type is required by tempest to avoid failures like:

```
Default share type is not found, please check default_share_type config: Share type with name default could not be found.:
manila.exception.ShareTypeNotFoundByName: Share type with name default could not be found.
```

### DIFF
--- a/ci-operator/step-registry/openstack-k8s-operators/tempest/openstack-k8s-operators-tempest-commands.sh
+++ b/ci-operator/step-registry/openstack-k8s-operators/tempest/openstack-k8s-operators-tempest-commands.sh
@@ -53,7 +53,12 @@ if [[ "$REF_ORG" != "$ORG" ]]; then
     fi
     BASE_OP=${EXTRA_REF_REPO}
 fi
+SERVICE_NAME=$(echo "${BASE_OP^^}" | sed 's/\(.*\)-OPERATOR/\1/'| sed 's/-/\_/g')
 
+# NOTE: if manila is deployed, build a default share required by tempest
+if [[ "${SERVICE_NAME}" == "MANILA" ]]; then
+    oc exec -it pod/openstackclient -- openstack share type create default false
+fi
 
 set +e
 


ci-operator/step-registry/openstack-k8s-operators/deploy/openstack-k8s-operators-deploy-commands.sh
```diff
@@ -240,3 +240,9 @@ oc exec -it  pod/mariadb-openstack -- mysql -uroot -p${MARIADB_PASSWD} -e "show
 # Post tests for keystone-operator
 # Check to confirm you can issue a token.
 openstack token issue
+
+# NOTE: if manila is deployed, build a default share required by tempest
+if [[ "${SERVICE_NAME}" == "MANILA" ]]; then
```
viroel, line 245: I would say that the best place is in tempest step[1]. But tempest container doesn't have openstack client installed [2]. We can update the container if needed. Or we can just merge this here.
Thoughts?

[1] https://github.com/openshift/release/blob/master/ci-operator/step-registry/openstack-k8s-operators/tempest/openstack-k8s-operators-tempest-commands.sh
[2] https://github.com/openstack-k8s-operators/openstack-k8s-operators-ci/blob/master/images/tempest-container/Dockerfile

ci-operator/step-registry/openstack-k8s-operators/deploy/openstack-k8s-operators-deploy-commands.sh
```diff
@@ -240,3 +240,9 @@ oc exec -it  pod/mariadb-openstack -- mysql -uroot -p${MARIADB_PASSWD} -e "show
...
+    openstack share type create default false
```
cescgina, line 246: Testing this in https://github.com/openshift/release/pull/38755 and getting the following error: `openstack: 'share type create default false' is not an openstack command. See 'openstack --help`. Logs: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_release/38755/rehearse-38755-pull-ci-openstack-k8s-operators-manila-operator-main-manila-operator-build-deploy-tempest/1658011422828793856/artifacts/manila-operator-build-deploy-tempest/openstack-k8s-operators-deploy/build-log.txt

fmount: ah, ok I suspect the openstackcli isn't able to discover manila. Let me try something different, we'll keep investigating this part.

ci-operator/step-registry/openstack-k8s-operators/deploy/openstack-k8s-operators-deploy-commands.sh
```diff
@@ -240,3 +240,9 @@ oc exec -it  pod/mariadb-openstack -- mysql -uroot -p${MARIADB_PASSWD} -e "show
 # Post tests for keystone-operator
 # Check to confirm you can issue a token.
 openstack token issue
+
+# NOTE: if manila is deployed, build a default share required by tempest
+if [[ "${SERVICE_NAME}" == "MANILA" ]]; then
+    oc exec -it pod/openstackclient -- openstack share type create default false
```
fmount, line 246: @cescgina maybe if we run the command by using the `openstackclient` pod it works because it's able to discover the right endpoints and should be aware about `share` service.

ci-operator/step-registry/openstack-k8s-operators/deploy/openstack-k8s-operators-deploy-commands.sh
gouthampacha: hi! agree that adding this here is like a simple post test/config for manila-api.. it's good for now. if we ever have to have a more elaborate default share type (i suspect we won't need to), we can move the concern to tempest bootstrapping..

fmount:
Tested via [1]  where we can actually see tempest running and using the default Manila backend.

[1] https://github.com/openshift/release/pull/38755

ci-operator/step-registry/openstack-k8s-operators/deploy/openstack-k8s-operators-deploy-commands.sh
fmount: @viroel ack and thanks for clarifying that aspect. My previous comment was made under different considerations, but I think one of the most important things is keeping our job steps as `individual` and `reusable`. Being that said, I updated the PR and  moved that piece of code in the tempest step.
Let me know if it looks good or requires more iterations.

fmount:
Closing in favor of https://github.com/openshift/release/pull/38755 where the patch has been added already.

PR #39312, opened 2023-05-15
nova-operator master branch renamed to main

gibizer:


gibizer:
/pj-rehearse ack

Two tempest test failed both with no valid host. So the job itself is OK after the rename. But we still need to fine tune the tempest allow list in this job to filter out test that needs compute nodes.
```
tempest.api.compute.admin.test_live_migration_negative.LiveMigrationNegativeTest.test_invalid_host_for_migration
tempest.api.compute.admin.test_live_migration_negative.LiveMigrationNegativeTest.test_live_block_migration_suspended
```

PR #39269, opened 2023-05-11
Cluster density v2 - perfscale 

paigerube14:
Adding performance and scale qe workload cluster-density-v2 with gen_csv options

PR #39331, opened 2023-05-15
Workers scaleup 

paigerube14:
Need to be able to scale up a cluster using the machine sets that are already there

### DIFF
--- a/ci-operator/step-registry/openshift-qe/cluster-density/openshift-qe-cluster-density-ref.metadata.json
+++ b/ci-operator/step-registry/openshift-qe/cluster-density/openshift-qe-cluster-density-ref.metadata.json
@@ -15,7 +15,8 @@
 			"paigerube14",
 			"rpattath",
 			"skordas",
-			"jtaleric"
+			"jtaleric",
+			"svetsa-rh"
 		]
 	}
 }
--- a/ci-operator/step-registry/openshift-qe/cluster-density/openshift-qe-cluster-density-workflow.metadata.json
+++ b/ci-operator/step-registry/openshift-qe/cluster-density/openshift-qe-cluster-density-workflow.metadata.json
@@ -15,7 +15,8 @@
 			"paigerube14",
 			"rpattath",
 			"skordas",
-			"jtaleric"
+			"jtaleric",
+			"svetsa-rh"
 		]
 	}
 }
--- a/ci-operator/step-registry/openshift-qe/network-perf/openshift-qe-network-perf-ref.metadata.json
+++ b/ci-operator/step-registry/openshift-qe/network-perf/openshift-qe-network-perf-ref.metadata.json
@@ -15,7 +15,8 @@
 			"paigerube14",
 			"rpattath",
 			"skordas",
-			"jtaleric"
+			"jtaleric",
+			"svetsa-rh"
 		]
 	}
 }
--- a/ci-operator/step-registry/openshift-qe/network-perf/openshift-qe-network-perf-workflow.metadata.json
+++ b/ci-operator/step-registry/openshift-qe/network-perf/openshift-qe-network-perf-workflow.metadata.json
@@ -15,7 +15,8 @@
 			"paigerube14",
 			"rpattath",
 			"skordas",
-			"jtaleric"
+			"jtaleric",
+			"svetsa-rh"
 		]
 	}
 }
--- a/ci-operator/step-registry/openshift-qe/network-smoke/openshift-qe-network-smoke-ref.metadata.json
+++ b/ci-operator/step-registry/openshift-qe/network-smoke/openshift-qe-network-smoke-ref.metadata.json
@@ -15,7 +15,8 @@
 			"paigerube14",
 			"rpattath",
 			"skordas",
-			"jtaleric"
+			"jtaleric",
+			"svetsa-rh"
 		]
 	}
 }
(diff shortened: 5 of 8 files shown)


paigerube14:
/pj-rehearse periodic-ci-openshift-qe-ocp-qe-perfscale-ci-main-ocp-qe-perfscale-ci-tests

paigerube14:
/pj-rehearse periodic-ci-openshift-qe-ocp-qe-perfscale-ci-main-ocp-qe-perfscale-ci-tests

paigerube14:
@skordas @qiliRedHat @svetsa-rh ptal, scaled to 5 for now to be able to see it run and use the current number of nodes when deciding the number of iterations

qiliRedHat:
@paigerube14 code looks good to me. I think this is the first machinest scale step in prow ci, what about making it more general? Like passing env MACHINESET_NAME OR (MACHINESET_LABEL_INCLUDE, MACHINESET_LABEL_EXCLUDE) AND REPLICA_COUNT. 
Some use case I can think of

1. AWS Local zone, scale machinesets with label 'edge', instead of 'workers'.
2. Hypershift team said scal up is something they were asked if QE covered and they hope perfscale QE to add tests. I said scale up only without perf test is a very simple step, they can add to their test flow before we are ready to run perfscale test on those openshift services.

paigerube14:
> @paigerube14 code looks good to me. I think this is the first machinest scale step in prow ci, what about making it more general? Like passing env MACHINESET_NAME OR (MACHINESET_LABEL_INCLUDE, MACHINESET_LABEL_EXCLUDE) AND REPLICA_COUNT. 
> 
> Some use case I can think of
> 
> 
> 
> 1. AWS Local zone, scale machinesets with label 'edge', instead of 'workers'.
> 
> 2. Hypershift team said scal up is something they were asked if QE covered and they hope perfscale QE to add tests. I said scale up only without perf test is a very simple step, they can add to their test flow before we are ready to run perfscale test on those openshift services.


This doesn't create any new machine sets so we wouldn't need a machine name. I could change worker count to replica count to be more general. This is the same code we use when we scale up our machine sets before adding the infra and workload nodes

paigerube14:
/pj-rehearse periodic-ci-openshift-qe-ocp-qe-perfscale-ci-main-ocp-qe-perfscale-ci-tests

paigerube14:
@jtaleric can you please take a look at this pr introducing scaling chain. I will have the [cluster-density-v2 pr](https://github.com/openshift/release/pull/39269) edit the command to get the number of nodes without grep.

svetsa-rh:
@paigerube14 The changes look good to me. I need to see if it works for Rosa clusters or not.

ci-operator/step-registry/workers-scale/workers-scale-commands.sh
```diff
@@ -0,0 +1,93 @@
...
+log(){
+    echo -e "\033[1m$(date "+%d-%m-%YT%H:%M:%S") " "${*}\033[0m"
+}
+
+function waitForReady() {
+    echo "Wait for Ready"
+    set +x
+    local retries=0
+    local attempts=140
+    while [[ $(oc get nodes --no-headers -l node-role.kubernetes.io/worker | grep -v "NotReady\\|SchedulingDisabled" | grep worker -c) != "$1" ]]; do
```
svetsa-rh, line 12: @paigerube14 This is not getting current worker count so your change https://github.com/openshift/release/pull/39269 won't effect , but may be we can use similar command pointed by Joe here as well?

ci-operator/step-registry/workers-scale/workers-scale-commands.sh
```diff
@@ -0,0 +1,93 @@
...
+        oc get nodes --no-headers -l node-role.kubernetes.io/worker | cat -n
+        log "Sleeping for 60 seconds"
+        sleep 60
+        ((retries += 1))
+        if [[ "${retries}" -gt ${attempts} ]]; then
+            for node in $(oc get nodes --no-headers -l node-role.kubernetes.io/worker | egrep -e "NotReady|SchedulingDisabled" | awk '{print $1}'); do
+                oc describe node $node
+            done
+
+            for machine in $(oc get machines -n openshift-machine-api --no-headers | grep -v "master" | grep -v "Running" | awk '{print $1}'); do
```
svetsa-rh, line 26: Same here too.

svetsa-rh (changes requested):

ci-operator/step-registry/workers-scale/workers-scale-commands.sh
paigerube14: I will have the https://github.com/openshift/release/pull/39269 edit the command to get the number of nodes without grep since this is passing already here, will fine toon in later pr

ci-operator/step-registry/workers-scale/workers-scale-commands.sh
svetsa-rh: Discussed with Paige. We will update these references in later PRs.

ci-operator/step-registry/workers-scale/workers-scale-commands.sh
svetsa-rh: The above commands are not addressed as part of https://github.com/openshift/release/pull/39269 but Paige or I will make sure to edit the commands in next PR.

PR #39342, opened 2023-05-15
Update sanitize-prowp-jobs for openstack-k8s-operators

viroel:
After moving almost all repos from 'master' to 'main' default branch, all related config/jobs where also updated.
This patch updates sanitize-prow-jobs/_config.yaml to be in sync with new job filenames.


viroel:
Already updated by CI bots:
https://raw.githubusercontent.com/openshift/release/master/core-services/sanitize-prow-jobs/_config.yaml

PR #39085, opened 2023-05-06
[Deprecated] OCPQE-15091: Fix chain-upgrade 4.1->4.13

liangxia:
Update on May 17, 2023
  This PR is deprecated, as in favor of https://github.com/openshift/release/pull/39348

PR #39348, opened 2023-05-16
OCPQE-15091: Fix chain-upgrade 4.1->4.13

liangxia:
This is to replace https://github.com/openshift/release/pull/39085


kuiwang02:
i am ok from golang's view.

PR #39388, opened 2023-05-16
Block any console repository PR that are not delivering major updates

jhadvig:
This change is due to https://github.com/openshift/console/pull/12821 which is addressing a major TypeScript version bump. For that reason we need to block any other console repositories from being merged to prevent rebases.

/assign @spadgett


spadgett (approved):
/lgtm

PR #39390, opened 2023-05-16
Release Branch Override Change

Kyl-Bempah:
In accordance with the changes made in https://github.com/stolostron/build-harness-extensions/commit/29a41b2429545116ffc1af428515ed98c5a443b8, https://github.com/stolostron/build-harness-extensions/pull/203, & https://github.com/stolostron/build-harness-extensions/pull/205 the `OSCI_COMPONENT_BRANCH` no longer needs to be overwritten with RELEASE_REF and will be deprecated when https://github.com/openshift/release/pull/40187 is merged. This change should provide the pipeline the correct release while allowing the actual git branch to be referenced in the manifest.


ci-operator/config/openshift/assisted-image-service/openshift-assisted-image-service-main__mce.yaml
```diff
@@ -45,7 +45,6 @@ tests:
         OSCI_PUBLISH_DELAY=0
         OSCI_PIPELINE_PRODUCT_PREFIX=backplane
         OSCI_PIPELINE_REPO=backplane-pipeline
-      RELEASE_REF: backplane-2.4
```
osherdp, line 48: Why? I mean what's suppose to happen without this value?

ci-operator/config/openshift/assisted-image-service/openshift-assisted-image-service-main__mce.yaml
osherdp, line 1: There are a couple of other components that I would expect be updated similarly (assisted-service, assisted-installer, etc.)
Any reason not to update their configuration?

osherdp (changes requested):

ci-operator/config/openshift/assisted-image-service/openshift-assisted-image-service-main__mce.yaml
Kyl-Bempah: This value was initially set due to needing to override your branch value in the pipeline and identify where to place it, that shouldn't be required anymore and there have been updates to better track where component updates are coming from so removing the override acts as data sanitation.

ci-operator/config/openshift/assisted-image-service/openshift-assisted-image-service-main__mce.yaml
Kyl-Bempah: I don't know what all your components are off-hand and initially requested this in the [slack thread](https://redhat-internal.slack.com/archives/CUPJTHQ5P/p1683740839696969) but I was pointed to this config, if you'd like to use this one as a guinea pig we can confirm and circle back to the others if you'd like to do them all at once you can use what I did here and duplicate it where it may apply.

ci-operator/config/openshift/assisted-image-service/openshift-assisted-image-service-main__mce.yaml
osherdp: thanks for the explanation
please have that in the PR description and commit message as future documentation to this change

ci-operator/config/openshift/assisted-image-service/openshift-assisted-image-service-main__mce.yaml
osherdp: the components are:
```openshift/{assisted-service,assisted-image-service,assisted-installer,assisted-installer-agent,cluster-api-provider-agent}```

osherdp (approved):

PR #39394, opened 2023-05-16
o/cincinnati-operator: test on all OCPs

petr-muller:


ci-operator/config/openshift/cincinnati-operator/openshift-cincinnati-operator-master.yaml
```diff
@@ -97,6 +106,214 @@ tests:
...
+      - env: RELATED_IMAGE_OPERATOR
+        name: pipeline:cincinnati-operator
+      - env: GRAPH_DATA_IMAGE
+        name: pipeline:cincinnati-graph-data-container
+      from: src
+      resources:
+        requests:
+          cpu: 100m
+          memory: 200Mi
+    workflow: hypershift-hostedcluster-workflow
```
wking, line 247: Moving from cluster-claims to HyperShift is interesting.  But before removing the WIP, we probably want to pick one approach and consolidate around that.

LalatenduMohanty (approved):
/lgtm

PR #39395, opened 2023-05-16
add `gangway-api` tenant_id to `qe-private-deck`

smg247:
`Deck` instances can have multiple `tenant_ids` and this addition will allow `qe-private-deck` to properly display jobs that enable the `gangway-api` as well as standard `qe-private` tenanted jobs.

We also need to add the `gangway-api` id to public deck and `deck-internal`. To do this the global default id will also need to be specified as you must specify all tenant ids if specifying any.

For: https://issues.redhat.com/browse/DPTP-3446

### DIFF
--- a/clusters/app.ci/prow/03_deployment/deck.yaml
+++ b/clusters/app.ci/prow/03_deployment/deck.yaml
@@ -108,6 +108,8 @@ objects:
           - --supplemental-plugin-config-dir=/etc/plugins
           - --kubeconfig-dir=/etc/build-farm-credentials
           - --kubeconfig-suffix=config
+          - --tenant-id=GlobalDefaultID
+          - --tenant-id=gangway-api
           env:
           - name: GITHUB_APP_ID
             valueFrom:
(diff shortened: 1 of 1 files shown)


clusters/app.ci/prow/03_deployment/qe_private_deck.yaml
```diff
@@ -125,6 +125,7 @@ objects:
                 - --kubeconfig-dir=/etc/build-farm-credentials
                 - --kubeconfig-suffix=config
                 - --tenant-id=qe-private
+                - --tenant-id=gangway-api
```
jmguzik, line 128: That's nice but I guess we need to add it as well to our deck. This will result in seeing qe jobs triggered by gangway in public deck as well, right? And vice versa? The solution for that could be adding 2 gangway tenants (if possible). One for public jobs and second one for QE.

smg247: The `tenant_id` isn't actually used for that much in `deck`. It merely provides filtering on the `prowjobs` that are visible. The actual `gangway-api` tenanted jobs won't show in public deck as their logs are still stored in the private bucket that public deck has no access to. Even now, the `gangway-api` tenanted jobs from the qe-private repos are not appearing in public deck.

As for public deck needing the `gangway-api` id added, I can do some additional research and see if that is necessary.

smg247: It does appear that `gangway-api` tenanted jobs are not showing at all in public deck. I will have to add the tenant id, **and** the default tenant id there as well as deck-internal.

jmguzik (approved):

PR #39403, opened 2023-05-17
Remove csi-cluster-drive-registrar

jsafrane:
It is a failed experiment. It's deprecated upstream and OCP has never used the image in any 4.x version.

cc @openshift/storage


hongkailiu (commented):
/lgtm

PR #39448, opened 2023-05-17
Use make test for OKO dataplane-operator.

bshephar:
The dataplane operator uses envtest for functional test coverage as outlined by the kubebuilder standards [1]. This change ensures envtest is setup correctly by using the make file target `test`.

### DIFF
--- a/ci-operator/config/openstack-k8s-operators/dataplane-operator/openstack-k8s-operators-dataplane-operator-main.yaml
+++ b/ci-operator/config/openstack-k8s-operators/dataplane-operator/openstack-k8s-operators-dataplane-operator-main.yaml
@@ -51,10 +51,10 @@ resources:
       cpu: 100m
       memory: 200Mi
 tests:
-- as: unit
+- as: functional
   commands: |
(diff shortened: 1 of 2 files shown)


slagle (approved):

bshephar:
> /pj-rehearse

I don't think we can rehearse it unless the patch is merged in the dataplane-operator unfortunately. But we can essentially determine that it would work based on the outcome of the Run tests job here:
https://github.com/openstack-k8s-operators/dataplane-operator/actions/runs/5008462239/jobs/8976357889?pr=187

This does the same thing proposed in this PR.

slagle:
> > /pj-rehearse
> 
> I don't think we can rehearse it unless the patch is merged in the dataplane-operator unfortunately. But we can essentially determine that it would work based on the outcome of the Run tests job here: https://github.com/openstack-k8s-operators/dataplane-operator/actions/runs/5008462239/jobs/8976357889?pr=187
> 
> This does the same thing proposed in this PR.

Ok, sounds good.

slagle (approved):

PR #39473, opened 2023-05-18
PowerVS: Swap 4.13 and 4.14

hamzy:
Since 4.13 has been released, we will now swap 4.13 jobs and 4.14 jobs.  Currently there are two sets of leases: one reliable and one less reliable.  We now need to be focused on the 4.14 jobs.

### DIFF
--- a/ci-operator/config/openshift/multiarch/openshift-multiarch-master__nightly-4.13.yaml
+++ b/ci-operator/config/openshift/multiarch/openshift-multiarch-master__nightly-4.13.yaml
@@ -471,7 +471,7 @@ tests:
   cluster: build01
   cron: 0 2,8,14,20 * * *
   steps:
-    cluster_profile: powervs-1
+    cluster_profile: powervs-2
     dependencies:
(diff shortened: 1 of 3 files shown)


ci-operator/jobs/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.13-periodics.yaml
```diff
@@ -44537,8 +44537,8 @@ periodics:
     org: openshift
     repo: openshift-tests-private
   labels:
-    ci-operator.openshift.io/cloud: powervs-1
-    ci-operator.openshift.io/cloud-cluster-profile: powervs-1
+    ci-operator.openshift.io/cloud: powervs-2
```
deepsm007, line 44540: I am not sure why QE test was also updated

hamzy: It is using the good pool, and the swap was meant to be consistent.

PR #39489, opened 2023-05-18
Export loki creds as env vars for openshift-tests

dgoodwin:
We'll soon be using these to upload intervals at the end of test runs.


ci-operator/step-registry/openshift/e2e/test/openshift-e2e-test-ref.yaml
```diff
@@ -4,6 +4,10 @@ ref:
   grace_period: 10m
   commands: openshift-e2e-test-commands.sh
   timeout: 14400s
+  credentials:
+  - namespace: test-credentials
+    name: loki-prod-collector-test-secret
+    mount_path: /var/run/loki-secret
```
dgoodwin, line 10: Borrowed from the hosted-loki step.

dgoodwin:
/pj-rehearse ack

No sign of issues in the rehersals and several successful jobs.

PR #39513, opened 2023-05-19
Set always_run:false for dataplane-operator build-deploy test

slagle:
This test is also run in the kuttl test, so it does not need to always
run as a separate test.

Signed-off-by: James Slagle <jslagle@redhat.com>


cescgina (commented):
lgtm, thanks!

fao89 (approved):

PR #39523, opened 2023-05-19
stackrox: use the same build root for 3.73 OCP variants

gavin-stackrox:
I expect this to fix the `Deployment.clone()` missing method failures. 
/cc @BradLugo @msugakov 
/hold


BradLugo (approved):
/lgtm

PR #39526, opened 2023-05-19
`tide`: increase memory request to match reality

smg247:
Tide has not had its memory request updated in more than 3 years. It is way off now, and can cause issues if it gets on a node that is under memory pressure.

I gathered [usage](https://console-openshift-console.apps.ci.l2s4.p1.openshiftapps.com/monitoring/query-browser?query0=container_memory_working_set_bytes%7Bcontainer%3D%7E%22tide%22%7D) for the last 2 weeks, and it hovers around `10Gi`.

### DIFF
--- a/clusters/app.ci/prow/03_deployment/tide.yaml
+++ b/clusters/app.ci/prow/03_deployment/tide.yaml
@@ -88,7 +88,7 @@ items:
             readOnly: true
           resources:
             requests:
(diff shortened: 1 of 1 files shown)


bradmwilliams (approved):

danilo-gemoli:
Why is it consuming over `10Gi`?

PR #39540, opened 2023-05-21
OCPQE-14716 Arch migration - Select release from channel candidate

evakhoni:
fixing a failure reported in [OCPQE-14716](https://issues.redhat.com/browse/OCPQE-14716), where there is a delay of release promotion to a candidate channel, causing test failures in prow CI. this pr switching to using the candidate as the source to begin with, so no such mismatch will occur.

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.10__amd64-stable.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.10__amd64-stable.yaml
@@ -30,12 +30,10 @@ build_root:
     tag: golang-1.17
 releases:
   latest:
-    prerelease:
+    release:
       architecture: amd64
-      product: ocp
-      version_bounds:
-        lower: 4.10.0
-        upper: 4.11.0-0
+      channel: candidate
+      version: "4.10"
 resources:
   '*':
     limits:
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.10__arm64-stable.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.10__arm64-stable.yaml
@@ -38,19 +38,15 @@ build_root:
     tag: golang-1.18
 releases:
   arm64-latest:
-    prerelease:
+    release:
       architecture: arm64
-      product: ocp
-      version_bounds:
-        lower: 4.10.0
-        upper: 4.11.0-0
+      channel: candidate
+      version: "4.10"
   latest:
-    prerelease:
+    release:
       architecture: amd64
-      product: ocp
-      version_bounds:
-        lower: 4.10.0
-        upper: 4.11.0-0
+      channel: candidate
+      version: "4.10"
 resources:
   '*':
     limits:
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.11__amd64-stable.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.11__amd64-stable.yaml
@@ -30,12 +30,10 @@ build_root:
     tag: golang-1.18
 releases:
   latest:
-    prerelease:
+    release:
       architecture: amd64
-      product: ocp
-      version_bounds:
-        lower: 4.11.0-0
-        upper: 4.12.0-0
+      channel: candidate
+      version: "4.11"
 resources:
   '*':
     limits:
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.11__arm64-stable.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.11__arm64-stable.yaml
@@ -38,19 +38,15 @@ build_root:
     tag: golang-1.18
 releases:
   arm64-latest:
-    prerelease:
+    release:
       architecture: arm64
-      product: ocp
-      version_bounds:
-        lower: 4.11.0
-        upper: 4.12.0-0
+      channel: candidate
+      version: "4.11"
   latest:
-    prerelease:
+    release:
       architecture: amd64
-      product: ocp
-      version_bounds:
-        lower: 4.11.0
-        upper: 4.12.0-0
+      channel: candidate
+      version: "4.11"
 resources:
   '*':
     limits:
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-stable.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-stable.yaml
@@ -38,12 +38,10 @@ build_root:
     tag: golang-1.18
 releases:
   latest:
-    prerelease:
+    release:
       architecture: amd64
-      product: ocp
-      version_bounds:
-        lower: 4.12.0-0
-        upper: 4.13.0-0
+      channel: candidate
+      version: "4.12"
 resources:
   '*':
     limits:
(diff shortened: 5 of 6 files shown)


ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.13__amd64-stable.yaml
```diff
@@ -50,19 +50,17 @@ build_root:
     tag: golang-1.18
 releases:
   latest:
-    prerelease:
+    release:
       architecture: amd64
+      channel: candidate
+      version: "4.13"
       product: ocp
```
shellyyang1989, line 57: Looks like `product: ocp` need to be removed.

evakhoni: done lets see...

shellyyang1989:
Looks good to me.

@aleskandro @jianlinliu could you please take a look? Thank you.

jianlinliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.13-amd64-stable-aws-ipi-to-multiarch-p2-f14

jianlinliu:
cc @liangxia to review.

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.13__amd64-stable.yaml
```diff
@@ -50,19 +50,15 @@ build_root:
     tag: golang-1.18
 releases:
   latest:
-    prerelease:
+    release:
```
liangxia, line 53: Perhaps you want a candidate release, rather than a candidate channel from official release ?
```
releases:
  latest:
    candidate:
      architecture: amd64
      product: ocp
      version: "4.13"
```
Official release are releases as customers would see them, which are already released. Are we supposed to test released payloads here ?

evakhoni: >Perhaps you want a candidate release, rather than a candidate channel from official release  
candidate releases should appear on official, with candidate channel aren't they?

shellyyang1989: Hi Liang,
We just want a release which is promoted to the production cincinnati because arch migration relies on cincinnati. We are testing payloads in candidate channel which includes all EC, RC and GA releases. Also a release goes to [candidate channel](https://docs.openshift.com/container-platform/4.13/updating/understanding-upgrade-channels-release.html#candidate-version-channel_understanding-upgrade-channels-releases) doesn't mean it's officially supported.

liangxia: They are different thing, at different level.
For releases, we have
1. candidate release -> candidate
2. pre-release -> prerelease
3. official release -> release
And for channels, e.g, under official release, we have
1. candidate channel
2. fast channel
3. stable channel
4. eus channel

liangxia: > Hi Liang,
> We just want a release which is promoted to the production cincinnati because arch migration relies on cincinnati. We are testing payloads in candidate channel which includes all EC, RC and GA releases. Also a release goes to [candidate channel](https://docs.openshift.com/container-platform/4.13/updating/understanding-upgrade-channels-release.html#candidate-version-channel_understanding-upgrade-channels-releases) doesn't mean it's officially supported.

Those are not only used by arch migration, they are also used by other tests in this file.
At least for stable e2e, we are not supposed to test a GAed/released version.

shellyyang1989: > And for channels, e.g, under official release, we have

> candidate channel

Maybe there is some gap here and ci doc needs update. Candidate channel only offers **unsupported** early access to releases as soon as they are built. It means when an ec/rc/standard GA release is built/created, it will be published to candidate channel, let's say 4.13.0-rc.7, when it's present on release controller and accepted, it will go to candidate channel asap. There is a delay here so we change to use the candidate channel.

liangxia: The issue is the change only consider arch migration, but not the other e2e tests. So it will cause issue for the other e2e tests. I would suggest separate those arch migration tests to a different config file from the e2e tests.

jianlinliu: If arch migration depends on official candidate channel, we need to rename the yaml file name for dedicated arch migration testing, at least, for now, in the 4.13 stable yaml, we only have arch migration testing.

jianlinliu: > for now, in the 4.13 stable yaml, we only have arch migration testing.

Correction, in openshift-openshift-tests-private-release-4.13__amd64-stable.yaml, there are some rosa test jobs, so if that, we need to create a dedicated arch migration yaml file, and move the arch migration jobs to the new ci job yaml file.

shellyyang1989: Hmm, I don't see the difference the change brings on the e2e test. They both select the recent non-nightly version. Am I miss anything?

jianlinliu: Take an example, if an candidate payload  build came out, e.g: 4.13.5, the build need to be tested against in e2e testing, if it passed, then we would go through errata process, and announce it is released, and place it in candidate channel. So lxia's point is to stress the stable yaml file is focus on QE's e2e testing before we release it public.

shellyyang1989: Like I said before, once the non-nightly payload is built and accepted, it will be published to candidate channel. Although there is a delay but it's very short. The release process is like: 
1. ART creates, let's say 4.13.5
2. ART signs the build
3. Dev & CI
4. ART publishes it to **candidate channel** if step#3 pass
5. QE does errata testing
6. Errata is shipped alive and the release is published to **fast channel**
7. The release is published to **stable channel** after a period of time watching

So as we can see, the release goes to candidate channel in very early stage before it's public.

jianlinliu: Good to know that. Now the key step is 3, I am not sure how often the build is created, but "Dev & CI" failed which blocks publishing it candidate channel, while QE still need to test them.

jianlinliu: One more thing is if 4.13 stable yaml get updated to use build from candidate channel, I think all the stable yaml should be updated to keep consistent.

shellyyang1989: > Now the key step is 3, I am not sure how often the build is created, but "Dev & CI" failed which blocks publishing it candidate channel, while QE still need to test them.

If "Dev & CI" blocking jobs failed, the release would be rejected. Even we use `prerelease`, we cannot test it as it selects the accepted release. Please correct me if I'm wrong.

> I think all the stable yaml should be updated to keep consistent

It's fine with me.

jianlinliu: @liangxia any comments here?

liangxia: +1 for update them all.

evakhoni: done https://github.com/openshift/release/pull/39540/commits/7a37d55d61688c7e48c1e0dac687f05561baa9fb

jianlinliu:
cc @liangxia to confirm

PR #39550, opened 2023-05-22
Refine the triggering of e2e-presubmit job

osherdp:
This should ignore changes in files that doesn't have any effect on the product itself, the build process or the e2e testing logic. For those cases we'll skip running the e2e-presubmit job, and for the other cases we'll always run this flow.

Closes [OCM-830](https://issues.redhat.com//browse/OCM-830)


bardielle (approved):
/lgtm

PR #39552, opened 2023-05-22
[WIP] Add heterogeneous IBM cloud VPC setup

werled:
Adds code for setting up additional worker nodes on IBM cloud VPC and a workflow that includes setup and e2e test.


werled:
Included in #40253

PR #39563, opened 2023-05-22
installer: add a terraform variant job

r4f4:
This job will build the terraform provider binaries for all platforms/arches needed by the Installer images. This should stop them from being built every time there is a change to the Installer repo, since the providers are changed very sparingly.

We should see a decrease in total Installer CI jobs runtime once this image is used by the Installer container images.

Update: since images are only promoted in the post-submit job, when there is a change to terraform providers the Installer's build script will detect changes and rebuild them. See https://github.com/openshift/installer/pull/7688.

### DIFF
--- a/ci-operator/config/openshift/installer/openshift-installer-master__terraform.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-master__terraform.yaml
@@ -1,0 +1,56 @@
+base_images:
+  base:
+    name: "4.15"
+    namespace: ocp
+    tag: base
+  ocp_builder_rhel-8-golang-1.20-openshift-4.15:
+    name: builder
+    namespace: ocp
+    tag: rhel-8-golang-1.20-openshift-4.15
+build_root:
+  from_repository: true
+images:
+- dockerfile_path: images/infrastructure-providers/Dockerfile
+  from: base
+  inputs:
+    ocp_builder_rhel-8-golang-1.20-openshift-4.15:
+      as:
+      - registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.20-openshift-4.15
+  to: installer-terraform-providers
+promotion:
+  name: "4.15"
+  namespace: ocp
+  tag_by_commit: true
+releases:
+  initial:
+    integration:
+      name: "4.15"
+      namespace: ocp
+  latest:
+    integration:
+      include_built_images: true
+      name: "4.15"
+      namespace: ocp
+resources:
+  '*':
+    requests:
+      cpu: 100m
+      memory: 200Mi
+  terraform-artifacts:
+    requests:
+      cpu: "5"
+      memory: 5Gi
+tests:
+- always_run: false
+  as: verify-vendor
+  commands: |
+    make -C terraform go-mod-tidy-vendor
+    git diff --exit-code
+  container:
+    from: src
+  run_if_changed: ^terraform/.*
+zz_generated_metadata:
+  branch: master
+  org: openshift
+  repo: installer
+  variant: terraform
--- a/ci-operator/jobs/openshift/installer/openshift-installer-master-presubmits.yaml
+++ b/ci-operator/jobs/openshift/installer/openshift-installer-master-presubmits.yaml
@@ -8768,6 +8768,119 @@ presubmits:
     - ^master$
     - ^master-
     cluster: build04
+    context: ci/prow/terraform-images
+    decorate: true
+    labels:
+      ci-operator.openshift.io/variant: terraform
+      ci.openshift.io/generator: prowgen
+      pj-rehearse.openshift.io/can-be-rehearsed: "true"
+    name: pull-ci-openshift-installer-master-terraform-images
+    rerun_command: /test terraform-images
+    run_if_changed: ^terraform/.*
+    spec:
+      containers:
+      - args:
+        - --gcs-upload-secret=/secrets/gcs/service-account.json
+        - --image-import-pull-secret=/etc/pull-secret/.dockerconfigjson
+        - --report-credentials-file=/etc/report/credentials
+        - --target=[images]
+        - --target=[release:latest]
+        - --variant=terraform
+        command:
+        - ci-operator
+        image: ci-operator:latest
+        imagePullPolicy: Always
+        name: ""
+        resources:
+          requests:
+            cpu: 10m
+        volumeMounts:
+        - mountPath: /secrets/gcs
+          name: gcs-credentials
+          readOnly: true
+        - mountPath: /secrets/manifest-tool
+          name: manifest-tool-local-pusher
+          readOnly: true
+        - mountPath: /etc/pull-secret
+          name: pull-secret
+          readOnly: true
+        - mountPath: /etc/report
+          name: result-aggregator
+          readOnly: true
+      serviceAccountName: ci-operator
+      volumes:
+      - name: manifest-tool-local-pusher
+        secret:
+          secretName: manifest-tool-local-pusher
+      - name: pull-secret
+        secret:
+          secretName: registry-pull-credentials
+      - name: result-aggregator
+        secret:
+          secretName: result-aggregator
+    trigger: (?m)^/test( | .* )terraform-images,?($|\s.*)
+  - agent: kubernetes
+    always_run: false
+    branches:
+    - ^master$
+    - ^master-
+    cluster: build04
+    context: ci/prow/terraform-verify-vendor
+    decorate: true
+    labels:
+      ci-operator.openshift.io/variant: terraform
+      ci.openshift.io/generator: prowgen
+      pj-rehearse.openshift.io/can-be-rehearsed: "true"
+    name: pull-ci-openshift-installer-master-terraform-verify-vendor
+    rerun_command: /test terraform-verify-vendor
+    run_if_changed: ^terraform/.*
+    spec:
+      containers:
+      - args:
+        - --gcs-upload-secret=/secrets/gcs/service-account.json
+        - --image-import-pull-secret=/etc/pull-secret/.dockerconfigjson
+        - --report-credentials-file=/etc/report/credentials
+        - --target=verify-vendor
+        - --variant=terraform
+        command:
+        - ci-operator
+        image: ci-operator:latest
+        imagePullPolicy: Always
+        name: ""
+        resources:
+          requests:
+            cpu: 10m
+        volumeMounts:
+        - mountPath: /secrets/gcs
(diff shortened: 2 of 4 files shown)


ci-operator/config/openshift/installer/openshift-installer-master__terraform.yaml
```diff
@@ -0,0 +1,110 @@
...
+    }
+    nplatforms=4 # (linux|darwin)_(amd|arm)64
+    [ $(ls terraform/bin/ | wc -w) == $nplatforms ] || fail "Artifacts were not generated for all platforms"
+    nbins=17 # 16 providers + terraform bin itself
+    for dir in terraform/bin/*; do
+      [ $(ls $dir | wc -w) == $nbins ] || fail "Missing binaries on $dir"
+    done
+  container:
+    from: terraform-artifacts
+  run_if_changed: terraform
```
andfasano, line 105: As per the [Prow docs](https://docs.prow.k8s.io/docs/jobs/#triggering-jobs-based-on-changes), `run_if_changed` is matched against the path of each changed file (and it's sufficient that _any_ will match). The current form should work, but just to be a slightly bit safer side we could use `  run_if_changed: ^terraform/.*`

ci-operator/config/openshift/installer/openshift-installer-master__terraform.yaml
```diff
@@ -0,0 +1,110 @@
...
+tests:
+- as: tf-check
+  commands: |
+    function fail() {
+      echo "$1"
+      exit 1
+    }
+    nplatforms=4 # (linux|darwin)_(amd|arm)64
+    [ $(ls terraform/bin/ | wc -w) == $nplatforms ] || fail "Artifacts were not generated for all platforms"
+    nbins=17 # 16 providers + terraform bin itself
```
andfasano, line 99: As a starting point it could be fine, but maybe we could check explicitly the list of expected binaries (names)

r4f4: I'm thinking it's probably better to add a script to the Installer repo, e.g., `hack/tf-check.sh` and use that here. Then it's much easier to keep the check synced with any tf providers changes.

ci-operator/config/openshift/installer/openshift-installer-master__terraform.yaml
```diff
@@ -0,0 +1,110 @@
...
+    COPY --from=linuxarmbuilder /go/src/github.com/openshift/installer/terraform/bin/ terraform/bin/
+  from: base
+  inputs:
+    ocp_builder_rhel-8-golang-1.18-openshift-4.12:
+      as:
+      - registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.18-openshift-4.12
+    ocp_builder_rhel-8-golang-1.19-openshift-4.14:
+      as:
+      - registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.19-openshift-4.14
+    ocp_builder_rhel-8-golang-1.20-openshift-4.14:
```
andfasano, line 64: Looks like only this one is really required (1.18-4.12 and 1.19-4.14 are not referenced in the dockerfile)?

r4f4: Removed.

r4f4:
/pj-rehearse pull-ci-openshift-installer-master-terraform-tf-check

ci-operator/config/openshift/installer/openshift-installer-master__terraform.yaml
andfasano, line 1: I think that maybe it could be more convenient to land this PR with just the variant job, and then later into a different PR (try to) modify the `openshift-installer-master` job. In this way we could be sure to effectively push the terraform-artifacts image in the CI registry - ie not sure it could work all together.
The variant job is anyhow fully isolated, so in case of need it this PR could be also reverted without impacting any other jobs.

r4f4:
/pj-rehearse pull-ci-openshift-installer-master-terraform-tf-check

ci-operator/config/openshift/installer/openshift-installer-master__terraform.yaml
```diff
@@ -0,0 +1,110 @@
...
+      as:
+      - registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.20-openshift-4.14
+  to: terraform-artifacts
+promotion:
+  name: "4.14"
```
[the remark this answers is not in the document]
andfasano: (also, I may be wrong, but I don't think that custom is will be automatically created, in my previous example we reused an already existing one)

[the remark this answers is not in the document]
r4f4: Ok. Done.

r4f4:
/assign @jupierce 
PTAL. @patrickdillon talked to you about our efforts to decrease Installer container images build time. This is the first step.

ci-operator/jobs/openshift/installer/openshift-installer-master-presubmits.yaml
```diff
@@ -5519,6 +5519,106 @@ presubmits:
           requests:
             cpu: 500m
     trigger: (?m)^/test shellcheck
+  - agent: kubernetes
+    always_run: true
```
patrickdillon, line 5523: I think this job should have `always_run: false` and a `run_if_changed` field, right?

ci-operator/config/openshift/installer/openshift-installer-master__terraform.yaml
```diff
@@ -0,0 +1,96 @@
...
+    namespace: ocp
+    tag: base
+  ocp_builder_rhel-8-golang-1.20-openshift-4.14:
+    name: builder
+    namespace: ocp
+    tag: rhel-8-golang-1.20-openshift-4.14
+build_root:
+  from_repository: true
+images:
+- dockerfile_literal: |
```
patrickdillon, line 13: The `dockerfile_literal` is good to be able to trigger rehearsals and CI config changes in a single PR, but at the end we want this dockerfile to live in the installer, repo, right?

andfasano: Right, this is convenient right now for experimenting without impacting too many repos. Once we'll be happy of the final dockerfile shape, we could refactor and extract it into its own file in the installer repo, and remove the literal here.

ci-operator/jobs/openshift/installer/openshift-installer-master-presubmits.yaml
r4f4: Do you know where to set that? I couldn't find a trigger-based container image generation example.

andfasano: Isn't this job triggered only in case of success of the terraform-check one? I wasn't expecting to find a run_if here (but on the other yes)

r4f4:
Update: rebased on top of master and added `tag_by_commit` to the promotion job.

ci-operator/jobs/openshift/installer/openshift-installer-master-presubmits.yaml
patrickdillon: Oh, I see. It looks like these triggers are not supported in the config, but it appears that image jobs may support them. We may be able to add the fields manually into the job here.

core-services/image-mirroring/openshift/mapping_origin_4_14
```diff
@@ -167,6 +167,7 @@ registry.ci.openshift.org/origin/4.14:ingress-node-firewall-daemon quay.io/opens
 registry.ci.openshift.org/origin/4.14:insights-operator quay.io/openshift/origin-insights-operator:4.14 quay.io/openshift/origin-insights-operator:4.14.0 quay.io/openshift/origin-insights-operator:latest
 registry.ci.openshift.org/origin/4.14:installer quay.io/openshift/origin-installer:4.14 quay.io/openshift/origin-installer:4.14.0 quay.io/openshift/origin-installer:latest
 registry.ci.openshift.org/origin/4.14:installer-artifacts quay.io/openshift/origin-installer-artifacts:4.14 quay.io/openshift/origin-installer-artifacts:4.14.0 quay.io/openshift/origin-installer-artifacts:latest
+registry.ci.openshift.org/origin/4.14:installer-terraform-artifacts quay.io/openshift/origin-installer-terraform-artifacts:4.14 quay.io/openshift/origin-installer-terraform-artifacts:4.14.0 quay.io/openshift/origin-installer-terraform-artifacts:latest
```
patrickdillon, line 170: hm. how/why was this generated?

r4f4: By promoting the image to the ocp namespace in the config.

r4f4: And by running `make openshift-image-mirror-mappings`

r4f4:
/hold
https://github.com/openshift/installer/pull/7716 is renaming the dir where the Dockerfile is.

core-services/image-mirroring/openshift/mapping_origin_4_14
patrickdillon: gh is missing the button to `Resolve Conversation`? This is clear now.

ci-operator/config/openshift/installer/openshift-installer-master__terraform.yaml
```diff
@@ -0,0 +1,56 @@
...
+build_root:
+  from_repository: true
+images:
+- dockerfile_path: images/terraform-providers/Dockerfile
+  from: base
+  inputs:
+    ocp_builder_rhel-8-golang-1.20-openshift-4.15:
+      as:
+      - registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.20-openshift-4.15
+  to: installer-terraform-providers
```
patrickdillon, line 19: > /hold [openshift/installer#7716](https://github.com/openshift/installer/pull/7716) is renaming the dir where the Dockerfile is.

Yes, it looks like it takes some effort to get an image promoted, so I think we should rename the image to `installer-infrastructure-providers` or something more generic, so that we will not need to update the image if we want to use the same pattern for CAPI provders .

ci-operator/jobs/openshift/installer/openshift-installer-master-presubmits.yaml
r4f4: I have added/edited the fields manually.

r4f4:
/assign @bparees 
for the image mirror changes.

PR #39574, opened 2023-05-22
Update ACM Installer repos fast-forwading for new versions

eemurphy:
Signed-off-by: Erin Murphy erinmurp@redhat.com
issue: https://issues.redhat.com/browse/ACM-5609


JakobGray (commented):
Why does `multiclusterhub-operator` only have one file change while the other components have multiple?
There is a change to the release-2.4 branch of discovery, but I wouldn't expect that to need a change because that branch is several releases old.

eemurphy:
@JakobGray thanks forgot to remove the disabled:true for the old branches in MCH. I had to add disabled:true to that discovery one because I got an error on this PR saying: `output tag stolostron/2.4:discovery-operator is promoted from more than one place: stolostron/discovery@main, stolostron/discovery@release-2.4` so had to add disabled:true. Same is true of release-2.3, it needs the disabled:true or else it give the same error because of the backplane-2.3 file

PR #39596, opened 2023-05-22
stackrox: import CI image for release branch 3.73 with golang 1.19.9 (take 2)

gavin-stackrox:
The first attempt updated nodejs to an incompatible version. These versions only have golang updated.
/hold
/cc @tommartensen


BradLugo (approved):
/lgtm

PR #39613, opened 2023-05-23
[openshift-loki]: Add missing test jobs for release-5.7

periklis:
Adds the same job sequence as in previous supported Loki Operator release branches, e.g. `release-5.6`.

/cc @xperimental

### DIFF
--- a/ci-operator/config/openshift/loki/openshift-loki-release-5.7.yaml
+++ b/ci-operator/config/openshift/loki/openshift-loki-release-5.7.yaml
@@ -17,9 +17,9 @@ base_images:
     tag: opa-openshift
 build_root:
   image_stream_tag:
-    name: release
-    namespace: openshift
-    tag: golang-1.18
+    name: builder
+    namespace: ocp
+    tag: rhel-8-golang-1.20-openshift-4.14
 images:
 - context_dir: operator
   dockerfile_path: Dockerfile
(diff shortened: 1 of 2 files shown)


ci-operator/config/openshift/loki/openshift-loki-release-5.7.yaml
```diff
@@ -72,7 +123,7 @@ tests:
       OO_PACKAGE: loki-operator
```
xperimental, line 123: Can not comment on the correct line:

Should `OO_CHANNEL` be `stable-5.7`?

periklis: There is no `stable-5.7` in the annotations yet.

ci-operator/config/openshift/loki/openshift-loki-release-5.7.yaml
```diff
@@ -17,9 +17,9 @@ base_images:
     tag: opa-openshift
 build_root:
   image_stream_tag:
-    name: release
-    namespace: openshift
-    tag: golang-1.18
+    name: builder
+    namespace: ocp
+    tag: rhel-8-golang-1.19-openshift-4.12
```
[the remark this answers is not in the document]
periklis: Good catch

ci-operator/config/openshift/loki/openshift-loki-release-5.7.yaml
xperimental: There should be. It's just that it exists on both `stable` and `stable-5.7`. But I guess you're right. This is stable because 5.8 isn't out yet. :+1:

PR #39631, opened 2023-05-23
rhcos-image-layering-examples: execute linter with correct name

sinnykumari:
Hopefully this is last thing to fix.


yuqi-zhang:
file seems to be referring to the right location

/lgtm

PR #39641, opened 2023-05-23
HOSTEDCP-966: require hypershift e2e for operators in control plane

sjenning:

### DIFF
--- a/ci-operator/config/openshift/cluster-storage-operator/openshift-cluster-storage-operator-master.yaml
+++ b/ci-operator/config/openshift/cluster-storage-operator/openshift-cluster-storage-operator-master.yaml
@@ -130,7 +130,6 @@ tests:
       TEST_SUITE: openshift/conformance/parallel
     workflow: openshift-e2e-openstack-ipi
 - as: e2e-hypershift-ovn-conformance
-  optional: true
   steps:
     workflow: hypershift-aws-conformance
 - as: verify-deps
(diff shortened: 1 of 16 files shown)


enxebre:
/lgtm
/approve
/hold
until this is solved https://github.com/openshift/cluster-ingress-operator/pull/930#issuecomment-1560186364

sjenning:
cc @deads2k @cgwalters @jsafrane

yuqi-zhang:
I think it makes sense from an MCO perspective, the main concern being how stable this would be.

Looking at the recent runs: https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-openshift-machine-config-operator-master-e2e-hypershift

Seems about 50/50? I guess we can retest until passing, but it may be hard to determine whether a test failure is due to a flake (looks like it would happen a lot atm) vs actually failing.

Is the MCO tests reflective of the general pass rate? Or are we an outlier? Would there be plans to get the e2e test to say, 80% or 90% success rate? Then I think we would be a bit more confident in the signal and having it required

sjenning:
The hypershift team + TRT.  It is a release blocking job so if it breaks, we'll know almost immediately and it will be high priority to fix it.

Note `hypershift-e2e` in the blocking jobs list
https://amd64.ocp.releases.ci.openshift.org/releasestream/4.14.0-0.ci/release/4.14.0-0.ci-2023-05-24-220329

Making this required really just provides pre-merge protection for your teams so that you don't break at the release level.

sinnykumari:
This really helps, thanks for the additional context Seth!

yuqi-zhang (approved):
approving from the MCO context, should be good to go, thanks Seth!

PR #39659, opened 2023-05-23
add image build for crd-schema-checker

deads2k:
so we can play with the admission webhook if we like.


bparees:
/approve

naming was agreed on in https://redhat-internal.slack.com/archives/CE4L0F143/p1685047015803859?thread_ts=1685028098.636719&cid=CE4L0F143

PR #39663, opened 2023-05-24
stackrox: fix concurrency

gavin-stackrox:
Some recently added jobs were backing up. 
/cc @tommartensen


gavin-stackrox:
> IIUC this change is to reduce queue length for these specific jobs?

Yes. The queue length of 1 means jobs on merge back up and then the #test-failures signal comes too late to have meaning for anyone.

PR #39676, opened 2023-05-24
cover more region test on cluster profile azuremag-qe

jinyunma:
update lease resource for cluster profile `azuremag-qe` to cover region `usgovtexas` test


jinyunma:
/unhold

cc @jianlinliu to review

PR #39722, opened 2023-05-25
OSSM-4014 Add 3.0 job that updates istio charts in istio-operator repo

dgn:
This also fixes the push-image job by using the new targets


dgn:
The prowjob rehearse fails because there's a cross-dependency between the changes in https://github.com/maistra/istio-operator/pull/1196 and this change. I think we should merge this first (I tested the steps locally) and then /retest the other PR

jwendell (approved):

PR #39728, opened 2023-05-25
Add a condition to skip running the evented pleg presubmit job

sairameshv:
This condition helps to skip running the evented pleg based presubmit ci job for unrelated changes.

/cc @sohankunkerkar


sohankunkerkar (changes requested):
`s/always_run: true/always_run: false`

sairameshv:
> `s/always_run: true/always_run: false`

Addressed @sohankunkerkar , Thanks for the catch !!

sohankunkerkar (approved):
Thanks!
/lgtm

PR #39731, opened 2023-05-25
[amq streams] - add 0.34.x image

see-quick:
This PR adds a new image for AMQ Streams product.


see-quick:
^^ @droslean @hongkailiu

PR #39732, opened 2023-05-25
loki-operator: updates operator to rhel 9

JoaoBraveCoding:
Relates to https://issues.redhat.com/browse/LOG-4059

### DIFF
--- a/ci-operator/config/openshift/loki/openshift-loki-main.yaml
+++ b/ci-operator/config/openshift/loki/openshift-loki-main.yaml
@@ -38,6 +38,13 @@ images:
 - context_dir: operator
   dockerfile_path: Dockerfile
   from: base
+  inputs:
+    base:
+      as:
+      - gcr.io/distroless/static:nonroot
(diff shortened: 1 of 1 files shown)


ci-operator/config/openshift/loki/openshift-loki-main.yaml
```diff
@@ -38,6 +38,11 @@ images:
 - context_dir: operator
   dockerfile_path: Dockerfile
   from: base
+  inputs:
+    ocp_builder_rhel-9-golang-1.20-openshift-4.14:
+      as:
+      - golang:1.20.4
+      - gcr.io/distroless/static:nonroot
```
periklis, line 45: We should use the 4.14 base image w/o the builder for the runner FROM directive. See https://github.com/openshift/loki/blob/main/Dockerfile.ocp#L7

periklis, line 45:
```suggestion
  base:
    as:
      - gcr.io/distroless/static:nonroot
```

PR #39746, opened 2023-05-25
Fix unnecessary base64 encoding of loki cred env vars

dgoodwin:
This was unintential, thought it was decoding them, but the source I
worked from was acutally encoding them for use in secrets.


dgoodwin:
I believe this is safe to merge without rehersals, the env var is unused until I can get it working in https://github.com/openshift/origin/pull/27930, and I have verified the bash syntax works locally.

/pj-rehearse ack

PR #39765, opened 2023-05-26
skip hive auto in openshift-tests-private repo

LiangquanLi930:
skip hive auto, we can run auto case by Jenkins job or local env


kuiwang02:
@LiangquanLi930 please execute `make update` or `make jobs` to update config.

PR #39769, opened 2023-05-26
Run assisted installer upgrade agent test without workers

jhernand:
Currently the assisted installer test that verifies the agent upgrade feature runs with three control plane nodes and two worker nodes because that is the default. That test does the following:

1. Starts the service configured with an old agent image.
2. Waits for the cluster to be ready to install.
3. Restarts the service configured with an agent image that doesn't exist.
4. Waits for the cluster to move to the insufficient state.
5. Restarts the service configured with the current agent image.
6. Waits for the cluster to be ready to install again.

But in some cases the cluster will move to the ready state when only the three master nodes are ready and some of the worker nodes have not even started yet. Those late worker nodes will register with the initial agent image and then will try to start the next step runner with the non existing agent image. In that particular situation the logic of the agent is to wait one hour before trying again, and that causes a timeout and makes the test fail.

It is unlikely that this situation will reproduce in production environments, and even if it does reproduce it is acceptable to wait that hour and fail in that case.

To avoid this situation in the test this patch change it so that it doesn't use worker nodes.

Related: https://issues.redhat.com/browse/MGMT-14691

### DIFF
--- a/ci-operator/config/openshift/assisted-test-infra/openshift-assisted-test-infra-master.yaml
+++ b/ci-operator/config/openshift/assisted-test-infra/openshift-assisted-test-infra-master.yaml
@@ -741,6 +741,7 @@ tests:
     cluster_profile: packet-assisted
     env:
       ASSISTED_CONFIG: |
+        NUM_WORKERS=0
         TEST=./src/tests/test_upgrade_agent.py
         TEST_FUNC=test_upgrade_agent
     workflow: assisted-baremetal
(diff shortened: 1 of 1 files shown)


eliorerz:
/pj-rehearse pull-ci-openshift-assisted-test-infra-master-e2e-metal-assisted-upgrade-agent
/pj-rehearse periodic-ci-openshift-assisted-test-infra-master-e2e-metal-assisted-upgrade-agent-periodic

eliorerz:
/pj-rehearse periodic-ci-openshift-assisted-test-infra-master-e2e-metal-assisted-upgrade-agent-periodic

PR #39802, opened 2023-05-29
OKD periodics: update FCOS stable/testing runs

vrutkovs:
Test FCOS stable on OKD 4.13 and FCOS testing-devel on 4.14


barbacbd (commented):
/lgtm

PR #39823, opened 2023-05-30
only run healthcheck for some jobs

zhaozhanqi:
@liangxia @jiajliu ^^


ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.10__amd64-nightly.yaml
```diff
@@ -582,10 +582,8 @@ tests:
     cluster_profile: aws-qe
     env:
       BASE_DOMAIN: qe.devcluster.openshift.com
-      E2E_RUN_TAGS: '@amd64 and @aws-ipi and @network-openshiftsdn and not @fips'
-      TAG_VERSION: '@4.10'
     test:
-    - chain: openshift-e2e-test-qe
+    - ref: cucushift-installer-check-cluster-health
```
liangxia, line 586: Maybe use `chain: cucushift-installer-check-cluster-health` which contains several checks ?

zhaozhanqi: if the cluster can be setup and all operator are running well, that should be fine.  so `ref: cucushift-installer-check-cluster-health` should be enough

PR #39825, opened 2023-05-30
add a printed message for downloaded camgi version

elmiko:
this change adds an echo statement to the must-gather commands script so that we can tell that the camgi tool has been successfully downloaded.


vrutkovs (approved):
/approve
/lgtm

PR #39836, opened 2023-05-30
Update python to version 3.9 with base image ubi9

btaani:
Upgrade `ubi8/python-3.6` to `ubi9/python:3.9` as a base for the curator image

Ref: [[LOG-4169](https://issues.redhat.com/browse/LOG-4169)

cc @periklis


ci-operator/config/openshift/origin-aggregated-logging/openshift-origin-aggregated-logging-master.yaml
```diff
@@ -102,7 +102,7 @@ base_images:
   ubi9-python311:
```
periklis:
```suggestion
  ubi9-python39:
```

periklis (changes requested):

PR #39854, opened 2023-05-30
HOSTEDCP-966: hypershift: mco: make hypershift e2e required

sjenning:


sjenning:
cc @yuqi-zhang

yuqi-zhang (approved):
/lgtm

PR #39856, opened 2023-05-30
HOSTEDCP-966: hypershift: rcm: require hypershift e2e

sjenning:

RCM is deployed directly by hypershift's control-plane-operator (CPO), not the RCMO.  So we should make sure changes to the RCM don't break hypershift.

https://github.com/openshift/hypershift/blob/main/control-plane-operator/controllers/hostedcontrolplane/routecm/deployment.go


sjenning:
/pj-rehearse pull-ci-openshift-route-controller-manager-main-e2e-hypershift

PR #39862, opened 2023-05-30
create pre-submit jobs for vSphere static IP

rvanderp3:
This optional job only runs when manually invoked and is intended to perform a e2e installation/test and verification that static IP addresses are provided.


elmiko (commented):
/approve
/lgtm

PR #39866, opened 2023-05-30
Vboulos promote images from main

vboulos:
This PR promotes the ACM components images from the components main branch. This allows us to perform testing for the latest ACM 2.x version that is not yet GA. Example: ACM 2.8 that is currently a release candidate and supposed to go GA in a couple of weeks. When 2.8 becomes GA, ACM 2.9 could be tested instead as its testing code will be available in the main branch, and so on.

### DIFF
--- a/ci-operator/config/stolostron/acmqe-grc-test/OWNERS
+++ b/ci-operator/config/stolostron/acmqe-grc-test/OWNERS
@@ -5,9 +5,9 @@
 # See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md
 
 approvers:
-- drkho
+- dtthuynh
 - vboulos
 options: {}
 reviewers:
-- drkho
+- dtthuynh
 - vboulos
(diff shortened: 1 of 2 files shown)


ascerra:
The files that you have created all look good to me!

The only thing that I believe is missing is a bit of a hack but it is helpful. In order to allow us to run rehearsal jobs that use that latest changes to a promoted image. These images need to be used in a job that runs in production.

We have a file where we are adding all images to trick the system into propagating the latest image changes to the build cluster so that they are available to runs using pj-rehearse.

That file is here: https://github.com/openshift/release/blob/master/ci-operator/config/rhpit/interop-tests/rhpit-interop-tests-main__image-hack.yaml

You'll see the images for the ACM 2.7 versions are here in baase_image and they are used in a "dummy" test stanza in the test section in the 2nd half of the file so we should add these "main" images to this file as well.

morvencao (commented):
/lgtm on Obs part.

PR #39881, opened 2023-05-31
OCPRHV-834: remove e2e-ovirt-sdn due to EOL

engelmi:
This PR removes the `e2e-ovirt-sdn` test step as ovirt will be disabled in the installer (PR: https://github.com/openshift/installer/pull/7213)


vrutkovs (approved):
/approve
/lgtm

barbacbd:
It looks like there is a 4.15 presubmit file. Should we make those changes here too ?

PR #39886, opened 2023-05-31
ETCD-434: add periodics for etcd recovery tests

tjungblu:


ci-operator/config/openshift/cluster-etcd-operator/openshift-cluster-etcd-operator-master__periodics.yaml
```diff
@@ -0,0 +1,47 @@
+base_images:
```
hasbro17, line 1: I'm curious if there's a difference (in terms of visibility or build images) in where we place these periodic presubmits?
Seems like I placed the scaling ones in https://github.com/openshift/release/blob/master/ci-operator/config/openshift/release/openshift-release-master__nightly-4.14.yaml#L955-L986

So long as this ends up being visible here this would be fine too:
https://sippy.dptools.openshift.org/sippy-ng/jobs/4.14?filters=%7B%22items%22%3A%5B%7B%22columnField%22%3A%22current_runs%22%2C%22operatorValue%22%3A%22%3E%3D%22%2C%22value%22%3A%227%22%7D%2C%7B%22columnField%22%3A%22variants%22%2C%22operatorValue%22%3A%22contains%22%2C%22value%22%3A%22never-stable%22%2C%22not%22%3Atrue%7D%5D%7D&sortField=net_improvement&sort=asc

tjungblu: @dgoodwin do you guys prefer to have it in the release config or can we keep it "team local"? 

I do assume that sippy wouldn't pick it up from here, I also targeted the master branch currently

tjungblu: @hasbro17 wdyt? I guess we can leave this on master presubmits for now, after 4.14 is out we can transfer it if we want to start blocking payloads on it. I think we should, but only when we have a good pass rate here.

hasbro17: Yep so long as we get a pass rate to look at on sippy that should be good enough.

PR #39896, opened 2023-05-31
Add cherrypick endpoint to podman-py

umohnani8:


umohnani8:
@rhatdan @giuseppe PTAL

lsm5:
@umohnani8 looks like this commit created a bunch of additional openshift tasks? Is that the expected result? Might be worth mentioning in the commit message for others (like me) to understand. Other than that, LGTM.

lsm5:
> @umohnani8 looks like this commit created a bunch of additional openshift tasks? Is that the expected result? Might be worth mentioning in the commit message for others (like me) to understand. Other than that, LGTM.

ohh, it's openshift release repo. Ok lol. My bad.

lsm5:
ah, ok so we want to be able to `slash-cherrypick $FOO_BRANCH`, that's what this does, right? LGTM. Do we also need to separately account for `cherry-pick`  or is that automatically handled with this?

PR #39899, opened 2023-05-31
Add JIRA epic for firewatch

smpsgill:
Updating the default project & adding the JIRA epic for firewatch config


Simranpal:
cc @FilipB

PR #39915, opened 2023-06-01
powervs-ccm: Use from_repository setting

RadekManak:


Karthik-K-N:
I see the test is failing because error message mismatch due to env variable being not set
```
ibm_loadbalancer_test.go:1662: Expected Error: 'Error running calicoctl: resource does not exist: bla bla bla, exit status 1'. Actual Error: 'Error running calicoctl: resource does not exist: bla bla blawarning: GOCOVERDIR not set, no coverage data emitted
```

and I also see that the in the new PR its being set https://github.com/openshift/cloud-provider-powervs/pull/35/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52


How to handle this scenario? Can we override this error and merge it. Or should we submit a PR to add the environmetal variable .

RadekManak:
Here's what we going to do:

1. Revert .ci-operator.yaml image to the one currently being used https://github.com/openshift/cloud-provider-powervs/pull/36/files
2. This pull requests should be able to merge
3. Revert back to go 1.20 in the rebase to kube 1.27

Karthik-K-N:
Seems like image is not available
```
 * could not resolve inputs: could not determine inputs for step [input:root]: could not resolve base image from openshift/release:rhel-8-golang-1.18-openshift-4.11: imagestreamtags.image.openshift.io "release:rhel-8-golang-1.18-openshift-4.11" not found
 ```

PR #39932, opened 2023-06-01
[LPTOCPCI-300] enhancements for interop testing - Quay scenario.

ascerra:
## Purpose
After viewing the first run of the quay testing against OCP 4.14 on OpenShift CI I noticed the following.

1. Scenario is not capable of trigger via the api, this PR adds that capability. (As of right now this is only needed if their is an intermittent failure when run by the cron, instead of waiting a week we can retrigger on demand when needed.)
2. The scenarios slack channel appears to not exist #quay-qe
3. The XMLS being generated to not start with `junit_`

## Add Jira ticket creation on failures
We've built a tool called [firewatch](https://github.com/CSPI-QE/firewatch) that allows us to create a Jira issue in a specific project based on where in the prow job we see the failure.
- This is configurable by a json blob in the config file
- This is brought into the config file via this ref [firewatch-report-issues](https://steps.ci.openshift.org/reference/firewatch-report-issues).
- We've combined this ref with the ipi-aws workflow already ([firewatch-ipi-aws](https://steps.ci.openshift.org/workflow/firewatch-ipi-aws)) and I am proposing that we use that here. Though it is not a requirement I see we've used cucushift-installer-rehearse-aws-ipi as the workflow here, which is very similar to ipi-aws. I believe the cucushift workflow does 3 additional things ([enable-qe-catalogsource](https://steps.ci.openshift.org/reference/enable-qe-catalogsource), [gather-core-dump](https://steps.ci.openshift.org/chain/gather-core-dump), [gather-network](https://steps.ci.openshift.org/chain/gather-network)) if these are required here and we cannot use the ipi-aws workflow then I can create a workflow to add the firewatch ref as a post step to a new workflow to extend the current cucushift workflow. Please let me know what you think.


ci-operator/jobs/quay/quay-tests/quay-quay-tests-master-periodics.yaml
```diff
@@ -482,6 +482,8 @@ periodics:
     job-release: "4.14"
     pj-rehearse.openshift.io/can-be-rehearsed: "true"
   name: periodic-ci-quay-quay-tests-master-ocp-414-quay38-quay-e2e-tests-quay38-ocp414-lp-interop
+  prowjob_defaults:
+    tenant_id: gangway-api
   reporter_config:
     slack:
       channel: '#quay-qe'
```
ascerra, line 489: @LiZhang19817 For some reason I am unable to find this slack channel in the Red Hat internal slack. Does this channel exist? If it does could you please share a link. If not we just need to update this line to an existing slack channel.

ci-operator/step-registry/quay-tests/test-quay-e2e/quay-tests-test-quay-e2e-commands.sh
```diff
@@ -16,6 +16,7 @@ mkdir -p $ARTIFACT_DIR
 
 function copyArtifacts {
     cp -r ./cypress/results/* $ARTIFACT_DIR && cp -r ./cypress/videos/smoke/* $ARTIFACT_DIR
+    #TODO: Find any XMLS generated and preprend "junit_" to the filename.
```
ascerra, line 19: In order for tests within the XML to be visible from the TestGrid dashboard automatically the syntax `junit_` needs to be prepended to the XML file name. The artifacts for the test that ran this past monday is[ here](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/periodic-ci-quay-quay-tests-master-ocp-414-quay38-quay-e2e-tests-quay38-ocp414-lp-interop/1663062757404053504/artifacts/quay-e2e-tests-quay38-ocp414-lp-interop/quay-tests-test-quay-e2e/artifacts/) 

If we configure this to change the filename from `quay_automation_testiong_report-4f2491ad7c1e96fb4ec891df158fe636.xml` to `junit_quay_automation_testiong_report-4f2491ad7c1e96fb4ec891df158fe636.xml`

then that should be all that is needed.

ci-operator/jobs/quay/quay-tests/quay-quay-tests-master-periodics.yaml
ascerra: Screenshot from 2023-06-06 18-50-49

ci-operator/config/quay/quay-tests/quay-quay-tests-master__ocp-414-quay38.yaml
```diff
@@ -29,17 +29,28 @@ resources:
...
+        ]
+      FIREWATCH_DEFAULT_JIRA_PROJECT: LPTOCPCI
+      FIREWATCH_JIRA_SERVER: https://issues.redhat.com
     test:
     - ref: quay-tests-deploy-quay-aws-s3
     - ref: quay-tests-test-quay-e2e
     - ref: quay-tests-qbo-qe-test
     - ref: quay-tests-cso-qe-test
-    workflow: cucushift-installer-rehearse-aws-ipi
+    workflow: firewatch-ipi-aws
```
LiZhang19817, line 53: @ascerra  the workflow "cucushift-installer-rehearse-aws-ipi" is defined by Openshift QE installation team, and mapping to profile "aws-qe", we should follow the rule. Otherwise we need to use other profile name to mapping your workflow.

ci-operator/step-registry/quay-tests/test-quay-e2e/quay-tests-test-quay-e2e-commands.sh
LiZhang19817: @ascerra OK, we can update our code to add this prefix "junit_"

ci-operator/config/quay/quay-tests/quay-quay-tests-master__ocp-414-quay38.yaml
ascerra: Sounds good I will update this thank you for the response.

ci-operator/jobs/quay/quay-tests/quay-quay-tests-master-periodics.yaml
ascerra: have you been receiving prow job reports in this channel? As far as I know the prow plugin will not update private channels. If we can direct this to a public channel then we can use this notification system to create messages in that channel like this is doing for another layered product for example. https://redhat-internal.slack.com/archives/C053E0P1H6J/p1686094558383499

LiZhang19817: Yes, we can receive the report and monitor the testing results.

LiZhang19817: @ascerra we've asked PROW CI team to give support to send test report to multiple slack channels. cc @dongboyan77 @chengzhang1016 
https://github.com/openshift/ci-tools/issues/3477

ascerra: The mtr-interop channel was just an example of another product teams public channel we'd want a public channel specific to Quay to send results too. The most important thing is that you get the results in a channel that you view regularly though. I wasn't aware results could be posted to private channels.

And awesome thanks for sharing this, I think we can leave it as is for now! I'll track this ticket and once if this feature is added we can add an additional channel here.

ci-operator/jobs/quay/quay-tests/quay-quay-tests-master-periodics.yaml
```diff
@@ -477,22 +477,13 @@ periodics:
...
-      channel: '#quay-qe'
-      job_states_to_report:
-      - success
-      - failure
-      - error
-      report_template: '{{if eq .Status.State "success"}} :rainbow: Job *{{.Spec.Job}}*
-        ended with *{{.Status.State}}*. <{{.Status.URL}}|View logs> :rainbow: {{else}}
-        :volcano: Job *{{.Spec.Job}}* ended with *{{.Status.State}}*. <{{.Status.URL}}|View
-        logs> :volcano: {{end}}'
+  name: periodic-ci-quay-quay-tests-master-ocp414-quay38-quay-e2e-tests-quay38-ocp414-lp-interop
```
calebevans, line 484: Looks like your Slack config got removed.

calebevans (changes requested):
The only thing I can see to change is to add the Slack config back in.

ci-operator/jobs/quay/quay-tests/quay-quay-tests-master-periodics.yaml
ascerra: Good catch! Fixing now!

PR #39934, opened 2023-06-01
USHIFT-1298: disable cgo when cross-compiling

dhellmann:
The MicroShift build sets CGO_ENABLED=1 by default, but that does not
work when cross-compiling to aarch64 so disable it in the CI build step.


dhellmann:
openshift-tests panicked, but that means we want past the point where the compilation of MicroShift would be different. The actual change to the Makefile is coming separately in https://github.com/openshift/microshift/pull/1869

/pj-rehearse ack

dhellmann:
/pj-rehearse ack

The e2e tests passed but the job failed to collect the debug data so it failed. The conformance job shows the same panic that has been fixed, which comes from running openshift-tests after successfully building. The ARM jobs are failing the tests, which also means the build was successful.

pacevedom (approved):
/lgtm

PR #39949, opened 2023-06-02
Fix apps stack not deleted on AWS GovCloud

yunjiang29:


ci-operator/step-registry/ipi/install/install/aws/ipi-install-install-aws-commands.sh
```diff
@@ -330,6 +330,7 @@ EOF
   fi
 
   if [ "${CLUSTER_TYPE}" == "aws-usgov" ]; then
+    echo ${APPS_DNS_STACK_NAME} >> "${SHARED_DIR}/to_be_removed_cf_stack_list"
```
jianlinliu, line 333: Can we replace line 310 with this line?

yunjiang29: fixed

ci-operator/step-registry/ipi/install/install/aws/ipi-install-install-aws-commands.sh
```diff
@@ -307,7 +307,7 @@ EOF
   fi
 
   APPS_DNS_STACK_NAME="${CLUSTER_NAME}-apps-dns"
-  echo ${APPS_DNS_STACK_NAME} > "${SHARED_DIR}/apps_dns_stack_name"
+  echo ${APPS_DNS_STACK_NAME} > "${SHARED_DIR}/to_be_removed_cf_stack_list"
```
jianlinliu, line 310: seem like should be `>>` here?

yunjiang29: Thanks!

PR #39953, opened 2023-06-02
openshift-pipelines/operator: Add some e2e jobs as well…

vdemeester:
… They will run no-op initially, but once they are there we can iterate over it on openshift-pipelines/operator repository.


jkandasa (approved):
/lgtm

PR #39956, opened 2023-06-02
loosen exit failure check to ensure deletion commands always happen

jianlinliu:


jianlinliu:
cc @yunjiang29 to review.

PR #39986, opened 2023-06-05
Sync serverless ci

nak3:
This patch syncs serving, net-istio and net-kourier CI.
It adds serving 1.10 CI config by running:

```
$ make generate-serving-ci REMOTE=git@github.com:nak3/release.git
```


ci-operator/config/openshift-knative/net-istio/openshift-knative-net-istio-release-v1.10__410.yaml
```diff
@@ -1,6 +1,24 @@
+base_images:
+  openshift_release_golang-1.19:
```
skonto, line 2: In productization env we use 1.20.3 should we have that instead?

ReToCode (approved):
/lgtm

/unhold

PR #39992, opened 2023-06-05
MGMT-14816: Use oci-assisted cluster profile

adriengentil:
Use oci-assisted cluster profile instead of raw credentials for
assisted job.


adriengentil:
/pj-rehearse ack
`e2e-oci-assisted` is still in development

osherdp (approved):

eliorerz (approved):

PR #40021, opened 2023-06-05
Fix gather-logs step to properly look in correct namespace

viroel:


gibizer:
/pj-rehearse ack
The log collector now works as expected in a multi namespace deployment. The failed jobs are failed due to running out of claimable test clusters.

gibizer (approved):

PR #40044, opened 2023-06-06
rpm-ostree: skip tests if doc changed

HuijingHei:


travier:
/lgtm
/hold
For @cgwalters or @jmarrero to confirm here

PR #40051, opened 2023-06-06
ci-operator/openshift/os: Bump QEMU tests timeouts to 4h

travier:
RHCOS Prow CI is regularly reaching the 3h timeout that we currently have. Bump the timeout to get CI passing until we have a better solution in place.


dustymabe (approved):
/lgtm

marmijo (approved):
LGTM

PR #40057, opened 2023-06-06
inscrease mirror registry disk size

yunjiang29:
C2S/SC2S mirror all testing images into local registry, it looks like 120GB is not enough now:

```
      sha256:fa920188f567e51d75aacd723f0964026e42ac060fed392036e8d4b3c7a8129f -> 7a9d35ac
  stats: shared=469 unique=4083 size=113.7GiB ratio=0.83
error: unable to retrieve source image registry.redhat.io/openshift-logging/elasticsearch6-rhel8
```


yunjiang29:
@jianlinliu PTAL

PR #40059, opened 2023-06-06
Run openstack-build-deploy-kuttl on glance-operator PRs

fmount:
When `glance-operator` receives `PRs`, only run `kuttl` jobs that already includes the `build-deploy` steps.
This avoid reducing the amount of duplicated job runs and help saving resources in CI.


fmount:
@viroel @konan-abhi @fultonj FYI

PR #40069, opened 2023-06-06
remove non-existent jobs from kcp-dev/kcp branchprotection rules

xrstf:
These two Github Actions have been removed recently in https://github.com/kcp-dev/kcp/pull/2963 and so must not be checked before merging any further PRs.


xrstf:
superseded by #40132

PR #40111, opened 2023-06-07
Remove typo,fix registry configure

xiuwang:
The pvc fails to create.
```
2023-05-20 05:13:16+00:00 - Setting the image registry storage to use the manual setting storage...
error: the path "${SHARED_DIR}}/registry-pvc.yaml" does not exist
```
@jianlinliu Could you help review this, thanks?

### DIFF
--- a/ci-operator/step-registry/ipi/install/vsphere/registry/ipi-install-vsphere-registry-commands.sh
+++ b/ci-operator/step-registry/ipi/install/vsphere/registry/ipi-install-vsphere-registry-commands.sh
@@ -21,7 +21,7 @@ spec:
     requests:
       storage: 10Gi
 EOF
-  oc create -f '${SHARED_DIR}}/registry-pvc.yaml' -n openshift-image-registry &&
+  oc create -f "${SHARED_DIR}/registry-pvc.yaml" -n openshift-image-registry &&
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/ipi/install/vsphere/registry/ipi-install-vsphere-registry-commands.sh
```diff
@@ -21,7 +21,7 @@ spec:
     requests:
       storage: 10Gi
 EOF
-  oc create -f '${SHARED_DIR}}/registry-pvc.yaml' -n openshift-image-registry &&
+  oc create -f '${SHARED_DIR}/registry-pvc.yaml' -n openshift-image-registry &&
```
jianlinliu, line 24: Beside that, I think here should use `""` instead  of `''`

xiuwang: Updated

xiuwang:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.11-amd64-nightly-vsphere-ipi-disconnected-registry-pvc-p2-f28

ci-operator/step-registry/ipi/install/vsphere/registry/ipi-install-vsphere-registry-commands.sh
xiuwang: @jianlinliu I have trigger periodic-ci-openshift-openshift-tests-private-release-4.11-amd64-nightly-vsphere-ipi-disconnected-registry-pvc-p2-f28 job

jianlinliu:
Oh, recently vsphere slice quota is tight, I would approve it, we can verify it in the future testing.

PR #40119, opened 2023-06-07
Enable aesgcm encryption tests for cluster-openshift-apiserver-operator

dgrisonnet:
Port of the work done in https://github.com/openshift/release/pull/36836 to openshift-apiserver-operator.

The jobs will be temporarily optional until I add the new tests logic in the repo. This is required to avoid breaking CI and causing disruption.

Prior to this change, the perf tests were not run in openshift-apiserver-operator repo, so I'll take this opportunity to all add them.


dinhxuanvu (approved):
/lgtm

PR #40120, opened 2023-06-07
Enable aesgcm encryption tests for cluster-authentication-operator

dgrisonnet:
Port of the work done in https://github.com/openshift/release/pull/36836 to cluster-authentication-operator.

The jobs will be temporarily optional until I add the new tests logic in the repo. This is required to avoid breaking CI and causing disruption.


dgrisonnet:
I need to follow up on that effort.

/remove-lifecycle stale

PR #40137, opened 2023-06-07
Add missing queries to console-plugin-template

jhadvig:
/assign @spadgett 

Just FYI we are missing `release-4.13` and `release-4.12` branches in the template repo. Will add then so they track the different `ConsolePlugin`'s API version.

### DIFF
--- a/core-services/prow/02_config/openshift/console-plugin-template/_prowconfig.yaml
+++ b/core-services/prow/02_config/openshift/console-plugin-template/_prowconfig.yaml
@@ -1,14 +1,65 @@
 tide:
   queries:
   - includedBranches:
+    - release-4.10
+    - release-4.11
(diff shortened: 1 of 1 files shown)


core-services/prow/02_config/openshift/console/_prowconfig.yaml
```diff
@@ -44,6 +44,26 @@ tide:
     - needs-rebase
     repos:
     - openshift/console
+  - includedBranches:
```
spadgett, line 47: Was this meant for the console-plugin-template prowconfig?

jhadvig:
not sure why the console config was update... anyway I've removed the changes... should be good to review again :)

spadgett (approved):
/lgtm

PR #40155, opened 2023-06-08
Rename custom-ingresscontroller to custom-controller

liangxia:
The test name `azure-ipi-ingress-custom-ingresscontroller-fips-p3-f28-destructive` contains 66 chars, which exceed the limit of 61 chars. While shortening the test name, some use `ingress-custom-controller`, and some use `ingress-custom-ingress`. We've changed the test names to `ingress-custom-controller` in an earlier PR to make it consistent.
To avoid the inconsistent name to happen in the future version, and keep the test name and workflow name consistent, I'd like to rename the workflow, too.

The original workflow is `cucushift-installer-rehearse-azure-ipi-ingress-custom-ingresscontroller`,
we have two ways to rename to shorten it, which one do you prefer ?
1. cucushift-installer-rehearse-azure-ipi-ingress-custom-controller
2. cucushift-installer-rehearse-azure-ipi-custom-ingresscontroller


lihongan:
Both OK to me. Thank you!

PR #40158, opened 2023-06-08
Updating oldest-supported-z-rollback to 4.13.0

shellyyang1989:
Update the oldest rollback version to 4.13.0 to pick up the fix https://github.com/openshift/origin/pull/27929


wking (approved):
/lgtm

PR #40160, opened 2023-06-08
fix: increase number of nodes for appstudio e2e

psturc:
in previous hypershift workflow we were using: 3x worker node [m6i.4xlarge (16vCPU, 64GB RAM)](https://aws.amazon.com/ec2/instance-types/m6i/)
now: 6x worker node [m5.xlarge (4vCPU, 16GB RAM)](https://aws.amazon.com/ec2/instance-types/m5/)

### DIFF
--- a/ci-operator/config/redhat-appstudio/e2e-tests/redhat-appstudio-e2e-tests-main.yaml
+++ b/ci-operator/config/redhat-appstudio/e2e-tests/redhat-appstudio-e2e-tests-main.yaml
@@ -43,7 +43,7 @@ tests:
     cluster_profile: aws-2
     env:
       HYPERSHIFT_HC_RELEASE_IMAGE: quay.io/openshift-release-dev/ocp-release:4.12.19-multi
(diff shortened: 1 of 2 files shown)


ci-operator/config/redhat-appstudio/e2e-tests/redhat-appstudio-e2e-tests-main.yaml
```diff
@@ -43,7 +43,7 @@ tests:
     cluster_profile: aws-2
     env:
       HYPERSHIFT_HC_RELEASE_IMAGE: quay.io/openshift-release-dev/ocp-release:4.12.19-multi
-      HYPERSHIFT_NODE_COUNT: "6"
+      HYPERSHIFT_NODE_COUNT: "12"
```
Michkov, line 46: Staging has 9 nodes (3 masters, 6 workers), production has 12 nodes (3 masters, 9 workers), I don't think that our testing env should have more workers than staging.

PR #40172, opened 2023-06-08
Add node with BMO to UPI deployed cluster

sgoveas:
Adds new step for to add BMH to an existing cluster
https://issues.redhat.com/browse/OCPQE-14592

### DIFF
--- a/ci-operator/step-registry/baremetal/lab/post/add-nodes/baremetal-lab-post-add-nodes-commands.sh
+++ b/ci-operator/step-registry/baremetal/lab/post/add-nodes/baremetal-lab-post-add-nodes-commands.sh
@@ -12,11 +12,6 @@ master00_bootMacAddr=$(yq e '.[] | select(.name == "master-00") | .provisioning_
 master00_username=$(yq e '.[] | select(.name == "master-00") | .bmc_user' ${SHARED_DIR}/hosts.yaml)
 master00_pass=$(yq e '.[] | select(.name == "master-00") | .bmc_pass' ${SHARED_DIR}/hosts.yaml)
 
-workera_addr=$(yq e '.[] | select(.name == "worker-a-00") | .bmc_address' ${SHARED_DIR}/hosts.yaml)
-workera_bootMacAddr=$(yq e '.[] | select(.name == "worker-a-00") | .provisioning_mac' ${SHARED_DIR}/hosts.yaml)
-workera_username=$(yq e '.[] | select(.name == "worker-a-00") | .bmc_user' ${SHARED_DIR}/hosts.yaml)
-workera_pass=$(yq e '.[] | select(.name == "worker-a-00") | .bmc_pass' ${SHARED_DIR}/hosts.yaml)
-
 cat > ${DIR}/bmo.yaml <<EOF
 apiVersion: metal3.io/v1alpha1
 kind: Provisioning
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/baremetal/lab/post/baremetal-lab-post-chain.yaml
```diff
@@ -2,6 +2,7 @@ chain:
   as: baremetal-lab-post
   steps:
     - chain: gather
+    - ref: baremetal-lab-add-nodes
```
aleskandro, line 5: we should add  this as first step of the `test:` stanza, before running the E2Es. We should do this at the testconfig level

ci-operator/step-registry/baremetal/lab/post/add-nodes/baremetal-lab-post-add-nodes-commands.sh
```diff
@@ -0,0 +1,143 @@
...
+CLUSTER_NAME="${NAMESPACE}"
+DIR=/tmp/bmo
+mkdir -p ${DIR}
+
+master00_addr=$(yq e '.[] | select(.name == "master-00") | .bmc_address' ${SHARED_DIR}/hosts.yaml)
+master00_bootMacAddr=$(yq e '.[] | select(.name == "master-00") | .provisioning_mac' ${SHARED_DIR}/hosts.yaml)
+master00_username=$(yq e '.[] | select(.name == "master-00") | .bmc_user' ${SHARED_DIR}/hosts.yaml)
+master00_pass=$(yq e '.[] | select(.name == "master-00") | .bmc_pass' ${SHARED_DIR}/hosts.yaml)
+
+workera_addr=$(yq e '.[] | select(.name == "worker-a-00") | .bmc_address' ${SHARED_DIR}/hosts.yaml)
```
aleskandro, line 15: we should do this in a loop so that we can address the case where more than 1 additional worker is requested.

aleskandro: See https://github.com/openshift/release/blob/524564cec347aceaa462d41ac942b277077f9fd7/ci-operator/step-registry/upi/install/heterogeneous/upi-install-heterogeneous-commands.sh#L177-L186

ci-operator/step-registry/baremetal/lab/post/add-nodes/baremetal-lab-post-add-nodes-commands.sh
```diff
@@ -0,0 +1,143 @@
...
+kind: Provisioning
+metadata:
+  name: provisioning-configuration
+spec:
+  provisioningNetwork: "Disabled"
+  watchAllNamespaces: false
+
+EOF
+
+cat > ${DIR}/master00-bmc.yaml <<EOF
```
aleskandro, line 31: do we need to do this only for one master, or all?

ci-operator/step-registry/baremetal/lab/post/add-nodes/baremetal-lab-post-add-nodes-commands.sh
```diff
@@ -0,0 +1,143 @@
...
+    name: worker-user-data-managed
+    namespace: openshift-machine-api
+
+EOF
+
+
+oc create -f ${DIR}/bmo.yaml 
+echo "Waiting up to 60 seconds for the metal3 pods to get created..."
+sleep 60
+while IFS= read -r line; do
```
aleskandro, line 101: we also need to allow failing for timeout, right?

ci-operator/step-registry/baremetal/lab/post/add-nodes/baremetal-lab-post-add-nodes-commands.sh
```diff
@@ -0,0 +1,143 @@
...
+  if [[ $line != *"Running"* ]]; then
+    echo "Creating resources to enable Metal platform components failed"
+    exit 1
```
aleskandro, line 104: `sleep 60 && continue` maybe?

ci-operator/step-registry/baremetal/lab/post/add-nodes/baremetal-lab-post-add-nodes-commands.sh
```diff
@@ -0,0 +1,143 @@
...
+  password: "$worker_pass"
+---
+apiVersion: metal3.io/v1alpha1
+kind: BareMetalHost
+metadata:
+  name: worker-a-00
+  namespace: openshift-machine-api
+spec:
+  bmc:
+    address: "redfish-virtualmedia://$workera_addr/redfish/v1/Systems/1"
```
aleskandro, line 83: we have a variable within the inventory to get the scheme and the base URL to use. Respectively, `bmc_scheme` and `bmc_base_uri`

ci-operator/step-registry/baremetal/lab/post/add-nodes/baremetal-lab-post-add-nodes-commands.sh
```diff
@@ -0,0 +1,143 @@
...
+    namespace: openshift-machine-api
+
+EOF
+
+cat > ${DIR}/workera-bmc.yaml <<EOF
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: workera-bmc
```
aleskandro, line 69: this should be `worker-a-00-bmc` so that we can allow nodes with different credentials (see the comment above regarding the loop to do this for multiple workers)

aleskandro:
There is one more step to be considered before this can start working. The BMO components expose the iso to mount in the hosts BMC. Therefore, the BMC hosts need to be able reaching the HTTP server exposing this iso. However, this HTTP server is hosted in a private network that is not routable from the current BMC net.

The current network is the fully possible set described in the doc: Baremetal Network + Provisioning Network (not needed here and optional for us)  + Out-of-band management network (unable to route traffic to the Baremetal Network).

Let's discuss what we're doing to address this problem offline.

ci-operator/step-registry/baremetal/lab/post/add-nodes/baremetal-lab-post-add-nodes-commands.sh
sgoveas: Done

ci-operator/step-registry/baremetal/lab/post/add-nodes/baremetal-lab-post-add-nodes-commands.sh
sgoveas: Only once

ci-operator/step-registry/baremetal/lab/post/add-nodes/baremetal-lab-post-add-nodes-commands.sh
sgoveas: Where to use these variables? I have used `name` `bmc_user` `bmc_pass` and `bmc_address`.
Are the username and password variables used correct? or do I need to use `redfish_*` variables?

ci-operator/step-registry/baremetal/lab/post/add-nodes/baremetal-lab-post-add-nodes-commands.sh
sgoveas: This step creates the ironic and metal3 pods and I expect them to be up and running in less than a minute, but as you suggested, it is better to loop and check in sometime. I have set it to checking every 30 sec in 3 mins.

ci-operator/step-registry/baremetal/lab/post/add-nodes/baremetal-lab-post-add-nodes-commands.sh
aleskandro: right, for the redfish credentials, it's better to use redfish_user and redfish_password (rationale: some nodes have different credentials for IPMI and redfish and they are kind of hard-coded :( )

PR #40187, opened 2023-06-08
Remove override for external components

Kyl-Bempah:
This should allow external components to provide their branches without breaking ones that do not


gurnben (approved):
/approve
/lgtm

PR #40202, opened 2023-06-08
Hive: Use 4.13 for all e2es

2uasimojo:
The referenced card was rescoped to make presubmits all run against the latest GAed release (currently 4.13). Subsequent work, via HIVE-2235, will add coverage for the in-flight release (currently 4.14) via periodics.

NOTE: This `unused` business is due to a quirk of using ClusterPools. Quoting from [1]:

> For any test using cluster_claim, ci-operator creates a release whose
> name is specified by cluster_claim.as which is latest by default. The
> payload of the release is the one used to install the ephemeral cluster
> for the test. That means if there is a release in the releases stanza
> with the same name, it is overridden throughout the test. To avoid the
> release overriding, cluster_claim.as can be given as a value which does
> not appear in the releases stanza.

[1] https://docs.ci.openshift.org/docs/architecture/ci-operator/#testing-with-a-cluster-from-a-cluster-pool

[HIVE-2234](https://issues.redhat.com//browse/HIVE-2234)

### DIFF
--- a/ci-operator/config/openshift/hive/openshift-hive-master.yaml
+++ b/ci-operator/config/openshift/hive/openshift-hive-master.yaml
@@ -25,12 +25,12 @@ promotion:
 releases:
   initial:
     integration:
-      name: "4.14"
+      name: "4.13"
       namespace: ocp
   latest:
     integration:
       include_built_images: true
-      name: "4.14"
+      name: "4.13"
       namespace: ocp
 resources:
   '*':
@@ -75,6 +75,7 @@ tests:
 - as: e2e
   cluster_claim:
     architecture: amd64
+    as: unused
     cloud: aws
     labels:
       region: us-east-1
@@ -125,6 +126,7 @@ tests:
 - as: e2e-pool
   cluster_claim:
     architecture: amd64
+    as: unused
     cloud: aws
     labels:
       region: us-east-1


2uasimojo:
We need a better way to confirm the semver of the release image being used by e2e*. Here's the breadcrumb trail I'm following right now for the e2e rehearsal:
- [This line](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_release/40202/rehearse-40202-pull-ci-openshift-hive-master-e2e/1666902222404849664#1:build-log.txt%3A22) has both "4.13" and "release:latest" in it. Good, but not sufficient.
- L8 of [this file](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_release/40202/rehearse-40202-pull-ci-openshift-hive-master-e2e/1666902222404849664/artifacts/e2e/test/build-log.txt) -- which is the actual output of e2e-test.sh -- tells us exactly what image we're getting... but it's a digest URI. Let's remember the first few chars of that sha: c62e618
- Bottom of [this file](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_release/40202/rehearse-40202-pull-ci-openshift-hive-master-e2e/1666902222404849664/artifacts/release/build-log.txt) implies that 4.13.0-0.ci.test-2023-06-08-205211-ci-op-rtgb841i-latest is either the same as, or copied to, the image with sha c62e618 from above.

2uasimojo:
> We need a better way to confirm the semver of the release

https://github.com/openshift/hive/pull/2044 may help.

PR #40225, opened 2023-06-09
o/cincinnati-graph-data: fast-track PRs by `openshift-bot`

petr-muller:
We already trust `openshift-ota-bot` and do not require `lgtm` and
`approved` labels, we can do the same for `openshift-bot` that ART uses.

### DIFF
--- a/core-services/prow/02_config/openshift/cincinnati-graph-data/_prowconfig.yaml
+++ b/core-services/prow/02_config/openshift/cincinnati-graph-data/_prowconfig.yaml
@@ -1,5 +1,14 @@
 tide:
   queries:
+  - author: openshift-bot
+    missingLabels:
+    - do-not-merge/hold
+    - do-not-merge/invalid-owners-file
+    - do-not-merge/work-in-progress
+    - jira/invalid-bug
+    - needs-rebase
+    repos:
+    - openshift/cincinnati-graph-data
   - author: openshift-ota-bot
     missingLabels:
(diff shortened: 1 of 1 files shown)


core-services/prow/02_config/openshift/cincinnati-graph-data/_prowconfig.yaml
```diff
@@ -9,6 +9,15 @@ tide:
     - needs-rebase
     repos:
     - openshift/cincinnati-graph-data
+  - author: openshift-bot
```
wking, line 12: Do we want to add this to [`trusted-apps`][1] too?

[1]: https://github.com/openshift/release/blob/6eaf4546cfeed7cf6970ff7ffa8783ed43a598ab/core-services/prow/02_config/openshift/cincinnati-graph-data/_pluginconfig.yaml#L12-L14

petr-muller: It's not an app, it's an account: https://github.com/openshift-bot

Apps have a little `bot` label in GH UI:

image

petr-muller: So it should not be necessary. `trusted-apps` control whose PRs can trigger CI jobs without `/ok-to-test`, which is normally controlled by org membership. apps are not accounts, so they cannot be org members, hence `trusted-apps` was invented.

PR #40232, opened 2023-06-09
Use release-2.8 for observability-e2e

dislbenn:
Switch Observability E2E to use `release-2.8` builds until `2.9-dev` image is available.


subbarao-meduri (approved):
/lgtm

PR #40236, opened 2023-06-10
PowerVS: update ibmcloud to 2.16.1

hamzy:
The ibmcloud command line released version 2.16.1


mjturek:
Is there a way to always pull latest? Is that something we'd want to do?

PR #40237, opened 2023-06-11
USHIFT-1297: Add iso boot step

pacevedom:
Building on top of USHIFT-1296, this step takes the results from the iso build and boots a VM with it. Waits for MicroShift running before finishing.

### DIFF
--- a/ci-operator/step-registry/openshift/microshift/infra/iso-boot/openshift-microshift-infra-iso-boot-commands.sh
+++ b/ci-operator/step-registry/openshift/microshift/infra/iso-boot/openshift-microshift-infra-iso-boot-commands.sh
@@ -27,11 +27,18 @@ cd ~/microshift
 ./scripts/image-builder/create-vm.sh edge default \$(find _output/image-builder -name "*.iso")
 VMIPADDR=\$(./scripts/devenv-builder/manage-vm.sh ip -n edge)
 timeout 5m bash -c "until ssh -oStrictHostKeyChecking=accept-new redhat@\${VMIPADDR} 'echo hello'; do sleep 5; done"
-VMHOSTNAME=\$(ssh "redhat@\${VMIPADDR}" hostname)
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/openshift/microshift/infra/iso-boot/openshift-microshift-infra-iso-boot-commands.sh
```diff
@@ -0,0 +1,41 @@
...
+chmod 0600 "${HOME}/.ssh/config"
+
+cat <<EOF > /tmp/boot.sh
+#!/bin/bash
+set -xeuo pipefail
+cd ~/microshift
+./scripts/image-builder/create-vm.sh edge default \$(find _output/image-builder -name "*.iso")
+VMIPADDR=\$(./scripts/devenv-builder/manage-vm.sh ip -n edge)
+timeout 5m bash -c "until ssh -oStrictHostKeyChecking=accept-new redhat@\${VMIPADDR} 'echo hello'; do sleep 5; done"
+VMHOSTNAME=\$(ssh "redhat@\${VMIPADDR}" hostname)
```
ggiguash, line 30: Why do we need this? Isn't the host name 'edge'?
I think the hostname command on the VM may return "localhost.localdomain" by default.

ci-operator/step-registry/openshift/microshift/infra/iso-boot/openshift-microshift-infra-iso-boot-commands.sh
```diff
@@ -0,0 +1,41 @@
...
+echo "\${VMIPADDR} \${VMHOSTNAME}" | sudo tee -a /etc/hosts
```
ggiguash, line 31: If we need a host name for remote connection, we can set it using `hostnamectl` command and restart microshift service.

PR #40243, opened 2023-06-11
release: setup image mirroring

Tal-or:


Tal-or:
not needed for now. we'll reevaluate if needed.

Tal-or:
Not needed for now. we'll reevaluate if needed.

PR #40260, opened 2023-06-12
Stop collecting logs from non-openshift namespaces in CI clusters.

dgoodwin:
I discovered that the vast majority of our new namespace label
cardinality is coming from non-openshift namespaces created temporarily
in the tests. While the system is handling this fine, we still do not
want this cardinality and would rather use it for app/container names.

This change drops logs for all temporary namespaces. I'm not aware of
anyone needing this as their tests can collect the namespace fine, and
they usually have random names regardless, so labelling is of little
use. If anyone does need these logs, we can likely duplicate the config
and label them differently, without namespace, or with namespace="TEMP",
etc.

### DIFF
--- a/ci-operator/step-registry/ipi/install/hosted-loki/ipi-install-hosted-loki-commands.sh
+++ b/ci-operator/step-registry/ipi/install/hosted-loki/ipi-install-hosted-loki-commands.sh
@@ -120,6 +120,9 @@ data:
       - match:
           selector: '{app="event-exporter", namespace="openshift-e2e-loki"}'
           action: drop
+      - match:
+          selector: '{namespace!~"openshift-.*"}'
+          action: drop
       - pack:
           labels:
           - app
@@ -196,13 +199,13 @@ data:
       pipeline_stages:
       - cri: {}
       - match:
-          selector: '{app="event-exporter", namespace="openshift-e2e-loki"}'
-          stages:
-          - static_labels:
-              type: kube-event
+          selector: '{namespace!="openshift-e2e-loki", app!="event-exporter"}'
+          action: drop
       - labelallow:
           - invoker
           - type
+      - static_labels:
+          type: kube-event
       relabel_configs:
       - action: replace
         source_labels:


ci-operator/step-registry/ipi/install/hosted-loki/ipi-install-hosted-loki-commands.sh
```diff
@@ -196,13 +199,13 @@ data:
       pipeline_stages:
       - cri: {}
       - match:
-          selector: '{app="event-exporter", namespace="openshift-e2e-loki"}'
-          stages:
-          - static_labels:
-              type: kube-event
+          selector: '{namespace!="openshift-e2e-loki", app!="event-exporter"}'
+          action: drop
```
dgoodwin, line 203: Cleaning this up as I discovered it has potential to double log things in other namespaces in weird ways, I added a static pod in default ns that just logged Hello world every 5s. This config was picking it up. Now we drop everything that isn't in namespace openshift-e2e-loki, and isn't app event exporter. (as the promtail pods are in here too, we want those logged by the above)

dgoodwin:
/pj-rehearse periodic-ci-openshift-release-master-nightly-4.14-e2e-aws-sdn

dgoodwin:
Looks like it's working both in rehersal and my test cluster.

https://grafana-loki.ci.openshift.org/explore?orgId=1&left=%7B%22datasource%22:%22PCEB727DF2F34084E%22,%22queries%22:%5B%7B%22expr%22:%22%7Binvoker%3D%5C%22openshift-internal-ci%2Frehearse-40260-periodic-ci-openshift-release-master-nightly-4.14-e2e-aws-sdn%2F1668250551978037248%5C%22,type%3D%5C%22pod%5C%22,namespace%21~%5C%22openshift.%2A%5C%22%7D%20%22,%22refId%22:%22A%22,%22editorMode%22:%22code%22,%22queryType%22:%22range%22%7D%5D,%22range%22:%7B%22from%22:%221686580829787%22,%22to%22:%221686588106610%22%7D%7D

This link shows we have no logs outside openshift- namespaces. Playing with that query I see that we do have our normal openshift ns logs, as well as kube-events.

PR #40276, opened 2023-06-13
Perf scale slack

paigerube14:
Add slack posts when periodic job runs


paigerube14:
@qiliRedHat I had to add to the jobs file itself but it stays between the auto generation of the job file. Documentation I followed here: https://docs.ci.openshift.org/docs/how-tos/notification/

paigerube14:
@qiliRedHat I think we will have to manually edit this file again when we add more jobs and the job file gets a new items added to the list

PR #40332, opened 2023-06-14
Remove tide and other plugins

RobGeada:
We don't want auto-merging from tide, so disabling it along with some other plugins we don't need


RobGeada:
\lgtm

danielezonca (approved):

ruivieira (approved):

PR #40337, opened 2023-06-14
more loki label fixes

dgoodwin:
Fix promtail pod logs duplicating as type=kube-event. The selector for the events job was doing an AND when I needed an OR. As a result, promtail logs were getting double logged, once as a pod log, once as type=kube-event. The kube-event one was hitting rate limits easily as promtail logs are VERY chatty right now. (the #2 most logged namespace)

Lower promtail log level to just warnings, as that is what I'm interested in for debugging what label combos are hitting rate limits.

Strong suspicion this will be the fix for most of the remaining 60 million dropped log lines a day.


dgoodwin:
These changes have been tested via this step-registry script on my aws cluster in a method that has proven 100% effective in all previous PRs, as such I'm not going to wait on rehersals.

/pj-rehearse ack

PR #40345, opened 2023-06-14
Add namespace labels to loki logs of type=kube-event

dgoodwin:
Extract namespace for loki logs with type=kube-event
    
    We need to ignore the normal namespace, which would be the pod
    event-exporter we're using to get kube events. This change uses a
    temp label of podnamespace, which is not exported as a real label to
    loki. (as it's not in labelallow)
    
    Then we use two json stages to get at the nested metadata.namespace
    json, and finally export this as the actual 'namespace' label.

### DIFF
--- a/ci-operator/step-registry/ipi/install/hosted-loki/ipi-install-hosted-loki-commands.sh
+++ b/ci-operator/step-registry/ipi/install/hosted-loki/ipi-install-hosted-loki-commands.sh
@@ -193,27 +193,37 @@ data:
       relabel_configs:
       - action: labelmap
         regex: __journal__(.+)
-    - job_name: events
+    - job_name: kubernetes-events
       kubernetes_sd_configs:
       - role: pod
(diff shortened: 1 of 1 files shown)


dgoodwin:
Results from my live cluster can be seen here: https://grafana-loki.ci.openshift.org/explore?orgId=1&left=%7B%22datasource%22:%22PCEB727DF2F34084E%22,%22queries%22:%5B%7B%22expr%22:%22%7Binvoker%3D%5C%22openshift-internal-ci%2Fdgoodwin-crc-test%2F1%5C%22,%20type%3D%5C%22kube-event%5C%22%7D%22,%22refId%22:%22A%22,%22editorMode%22:%22code%22,%22queryType%22:%22range%22%7D%5D,%22range%22:%7B%22from%22:%221686762055847%22,%22to%22:%221686762072096%22%7D%7D

/pj-rehearse ack

PR #40347, opened 2023-06-14
Add Anaconda notebook build into weekly build

Xaenalt:
This adds the weekly build and push for the Anaconda workbench images

Out of curiosity, how stable are the tags for it? I know we need to use sha256s, how long do they stay around?


Xaenalt:
@atheo89 @harshad16 PTAL

harshad16:
> Out of curiosity, how stable are the tags for it? I know we need to use sha256s, how long do they stay around?

We didn't put any expiration on the image in quay.io.
Till I know there is no default expiration time.

PR #40356, opened 2023-06-15
Unify name for profile cco-manual-sts and cco-manual-workload-identity

liangxia:
Changes in this PR,
1. Add `fips` to profile `aws-ipi-disconnected-cco-manual-sts-ep` for 4.10/4.12 stable tests and several upgrade tests.
2. Unify the profile name from `*-cco-manual-sts-ep`, `*-cco-manual-sts-ep-fips`, `*-cco-man-sts-ep-fips` to `*-sts-ep-fips` for amd.
3. To keep consistent with item 2, use `*-sts-ep` for arm (arm does not support fips at the moment).

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.10__amd64-stable.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.10__amd64-stable.yaml
@@ -53,13 +53,14 @@ tests:
     test:
     - chain: cucushift-installer-check-cluster-health
     workflow: cucushift-installer-rehearse-aws-c2s-ipi-disconnected-private
-- as: aws-ipi-disconnected-cco-manual-sts-ep-p2-f28
-  cron: 3 8 23 * *
+- as: aws-ipi-disconnected-sts-ep-fips-p2-f28
+  cron: 24 9 10 * *
   remote_api: true
   steps:
     cluster_profile: aws-qe
     env:
       BASE_DOMAIN: qe.devcluster.openshift.com
+      FIPS_ENABLED: "true"
     test:
     - chain: cucushift-installer-check-cluster-health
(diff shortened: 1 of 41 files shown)


ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.10__amd64-stable.yaml
```diff
@@ -53,13 +53,14 @@ tests:
     test:
     - chain: cucushift-installer-check-cluster-health
     workflow: cucushift-installer-rehearse-aws-c2s-ipi-disconnected-private
-- as: aws-ipi-disconnected-cco-manual-sts-ep-p2-f28
-  cron: 3 8 23 * *
+- as: aws-ipi-disconnected-cco-man-sts-ep-fips-p2-f28
```
jianlinliu, line 56: We had ever talked with sts naming issue before, cc @gpei and @yunjiang29 , we suggest to remove "cco-manual" keyword directly for any sts ci jobs.

gpei: yes, we had a discussion with CCO team and reached the following agreement:

- use `sts` to replace `cco-manual-sts` for aws
- use `workload-identity` to replace `cco-manual-workload-identity` for gcp

liangxia: @jianlinliu @gpei  Replaced `cco-manual-sts` and `cco-man-sts` with `sts`.
I think we can do the other rename in a follow up PR, as this PR already contains many lines of changes.

liangxia: Applied the changes for both `cco-manual-sts` and `cco-manual-workload-identity`

PR #40364, opened 2023-06-15
unify job name to shorten techpreview as tp and replace sts with workload-identity on gcp

jianlinliu:


ci-operator/jobs/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12-periodics.yaml
```diff
@@ -12186,6 +12186,7 @@ periodics:
         secretName: result-aggregator
 - agent: kubernetes
   cluster: build03
+<<<<<<< HEAD
```
liangxia, line 12189: Unexpected line ?

jianlinliu: fixed.

jianlinliu:
FYI, @jianli-wei I update gcp sts job names together in this PR.

jianli-wei:
> FYI, @jianli-wei I updated gcp sts job names together in this PR.

It's said we could use "**wif**" (workload Identity Feature) instead. @gpei Any suggestion? Thanks!

gpei:
> > FYI, @jianli-wei I updated gcp sts job names together in this PR.
> 
> It's said we could use "**wif**" (workload Identity Feature) instead. @gpei Any suggestion? Thanks!

I remember the CCO dev team said they're using `workload-identity` for their e2e jobs, and agreed that we can use this words for both GCP and Azure. The problem of `wif` is managers doesn't know the meaning of it.

PR #40365, opened 2023-06-15
ROX-17751: Run ocp sensor-integration only if explicitly requested

vikin91:
Sensor-integration tests run fine on GKE currently. On OCP, they fail for the reasons specific to the tests. We want to work on them to make them pass on OCP (see ticket https://issues.redhat.com/browse/ROX-17875), but that work is not urgent.

For now, let's disable those tests on master presubmits to save some costs.
The tests should be executed in PRs only when command `/test ocp-4-*-sensor-integration-tests` is run.

### DIFF
--- a/ci-operator/jobs/stackrox/stackrox/stackrox-stackrox-master-presubmits.yaml
+++ b/ci-operator/jobs/stackrox/stackrox/stackrox-stackrox-master-presubmits.yaml
@@ -852,7 +852,8 @@ presubmits:
       pj-rehearse.openshift.io/can-be-rehearsed: "true"
     name: pull-ci-stackrox-stackrox-master-ocp-4-10-qa-e2e-tests
     optional: true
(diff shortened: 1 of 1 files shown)


gavin-stackrox (changes requested):
Can you also remove from nightlies? https://github.com/openshift/release/blob/b37cf43b5dd43f5537df31c8a9c7af9787d61a18/ci-operator/config/stackrox/stackrox/stackrox-stackrox-nightlies__ocp-4-10.yaml#L44 
+4.12, +4.13

vikin91:
@gavin-stackrox , I am starting my PTO. In case there are further changes required, I would kindly ask you to apply them and merge this. Thanks a lot for the help!

vikin91:
@gavin-stackrox could you give this PR another look?

gavin-stackrox (approved):
/lgtm

PR #40369, opened 2023-06-15
[release-4.13] ovn-kubernetes: switch back to normal Dockerfiles

dcbw:
Now that the normal Dockerfiles are synced with the .rhel9 ones (via https://github.com/openshift/ovn-kubernetes/pull/1695), we can switch back to them and then eventually remove the .rhel9 ones to reduce confusion.


dcbw:
/pj-rehearse ack

We only need to test that images get built and clusters come up, which they all did

PR #40370, opened 2023-06-15
upgrading go version to 1.19 and ocp4.13 for kn-pluginc-func job

jrangelramos:
upgrading go version to 1.19 and ocp4.13 for kn-pluginc-func job


jrangelramos:
@dsimansk could you please take a look and lgtm again?

PR #40413, opened 2023-06-16
test understore in job profilename

pruan-rht:


pruan-rht:
closing this as `_` is not a valid k8s object name
https://kubernetes.io/docs/concepts/overview/working-with-objects/names/

PR #40415, opened 2023-06-16
hypershift: enable boskos leases on mgmt clusters

sjenning:


sjenning:
cc @wking

ci-operator/step-registry/hypershift/aws/setup-nested-management-cluster/hypershift-aws-setup-nested-management-cluster-chain.yaml
```diff
@@ -3,6 +3,9 @@ chain:
   steps:
   - as: create-management-cluster
     cli: latest
+    leases:
```
wking, line 6: Docs for [explicit lease configuration][1], for folks like me who aren't all that familiar with this syntax.

[1]: https://docs.ci.openshift.org/docs/architecture/step-registry/#explicit-lease-configuration

wking (approved):
Would be good to understand how we arrived at 30, what constraints were considered, and what constraints (e.g. AWS resource quotas in the relevant account/region/whatever) that are currently high enough to not need consideration, but which we may need to consider if we grow this number in the future.  Getting that context is not a blocker though, and the implementation:

/lgtm

PR #40419, opened 2023-06-16
MicroShift: also create lvm thinpool

copejon:
This is required for testing csi-snapshotting.


pmtk:
@copejon looks like I cannot ack the rehearsals

PR #40438, opened 2023-06-19
Enable oc binary for Far

razo7:
1. Pull image regardless of semver version (which is `latest` for main branch only).
2. Enable `oc` binary (and more from `cli: latest`) which will be used to patch the CredentialsRequest for E2E test on AWS


slintes:
please update title and description

/lgtm
/hold

PR #40456, opened 2023-06-19
telco5g: Record cluster name for later statistics

sshnaidm:


ci-operator/step-registry/telco5g/cluster-setup/telco5g-cluster-setup-commands.sh
```diff
@@ -104,6 +104,7 @@ ansible-playbook -i $SHARED_DIR/bastion_inventory $SHARED_DIR/get-cluster-name.y
 # shellcheck disable=SC2046,SC2034
 IFS=- read -r CLUSTER_NAME CLUSTER_API_IP CLUSTER_API_PORT CLUSTER_HV_IP CLUSTER_ENV <<< "$(cat ${SHARED_DIR}/cluster_name)"
 PLAN_NAME="${CLUSTER_NAME}_ci"
+echo "${CLUSTER_NAME}" > ${ARTIFACT_DIR}/job-cluster
```
yuvalk, line 107: maybe 
`echo "${CLUSTER_NAME}" | tee ${ARTIFACT_DIR}/job-cluster`
is better so we also have that in the visible log?

PR #40460, opened 2023-06-20
Add vSphere disconnected profile

sgaoshang:
Add vSphere disconnected profile for IBMC

### DIFF
--- a/core-services/prow/02_config/generate-boskos.py
+++ b/core-services/prow/02_config/generate-boskos.py
@@ -185,6 +185,7 @@ CONFIG = {
     'vsphere-quota-slice':{},
     'vsphere-8-quota-slice':{},
     'vsphere-discon-quota-slice':{},
+    'vsphere-dis-quota-slice':{},
     'vsphere-clusterbot-quota-slice':{},
     'vsphere-connected-quota-slice':{},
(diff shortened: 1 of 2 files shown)


sgaoshang:
@jinyunma @jianlinliu pls help to review it, thanks.

core-services/prow/02_config/generate-boskos.py
```diff
@@ -295,6 +296,9 @@
 for i in range(60,64):
     CONFIG['vsphere-discon-quota-slice']['qe-discon-segment-{}'.format(i)] = 1
 
+for i in range(230,235):
+    CONFIG['vsphere-dis-quota-slice']['devqe-segment-{}'.format(i)] = 1
```
jianlinliu, line 300: shall we add `discon` keyword in `devqe-segment` ? Or `devqe-segment` is the name configured in ibmc vsphere cluster, can not be changed.

jinyunma: seems missing `-disconnected` here? should be `devqe-segment-{}-disconnected`.

sgaoshang: Good catch, I added it back, thanks.

PR #40475, opened 2023-06-20
[WIP] MGDAPI-5854 - bump: use go1.19 on integreatly operator feature branches

KevFan:
* Remove obselete branches
* Bump to use go 1.19 on feature branches


Boomatang:
This wont work. The go version comes from https://github.com/integr8ly/integreatly-operator/blob/master/openshift-ci/Dockerfile.tools If this is merge the ci-operator will override the changes.

The files that are being delete possible could be merged. I only looked at the bump to the go version.

KevFan:
> This wont work. The go version comes from https://github.com/integr8ly/integreatly-operator/blob/master/openshift-ci/Dockerfile.tools If this is merge the ci-operator will override the changes.
> 
> The files that are being delete possible could be merged. I only looked at the bump to the go version.

@Boomatang I believe that is true for master which have been auto dated, but not sure about feature branches, do you know does this apply for feature branches also ?

KevFan:
> Yes, you would need to merge the changes to the dockerfile.tools on all the feature branches.

Thanks ! I'll rebase the feature branches instead and update the PR

KevFan:
\hold

KevFan:
Going to just close this PR since it's not required. Obselete branches can be deleted at some other stage

PR #40485, opened 2023-06-20
ostreedev/ostree: Update Prow configuration

cgwalters:
This will pair with https://github.com/ostreedev/ostree/pull/2893


jmarrero (approved):
/lgtm

PR #40508, opened 2023-06-21
Bump oldest-supported-z-rollback to 4.13.3

shellyyang1989:
Bump oldest-supported-z-rollback to 4.13.3 to pick up https://github.com/openshift/origin/pull/27956 and https://github.com/openshift/cluster-network-operator/pull/1824


shellyyang1989:
/close

As we may want to skip the test `openshift-tests.[sig-network][Feature:tuning] sysctl allowlist update should start a pod with custom sysctl only when the sysctl is added to whitelist`

PR #36495, opened 2023-02-17
AGENT-502: Use assisted-test-bin image for unit test jobs

andfasano:
Required by https://github.com/openshift/assisted-installer-agent/pull/482, due the nmstate dependencies.

Note: docker (and docker-compsoe) do not seem to be needed to pass the unit test job

PR #36356, opened 2023-02-14
Build assisted-test-bin image with the required agent-tui dependencies

andfasano:
This patch supports running succesfully the linting job for https://github.com/openshift/assisted-installer-agent/pull/482.
A new image is assembled with the required dependecies (nmstate)

PR #40513, opened 2023-06-21
AGENT-595: cleanup agent-installer based code

andfasano:
Since agent-installer based is now shipped through the agent-installer-utils image, there's no need to keep this code here anymore.

This patch reverts the changes introduced in https://github.com/openshift/release/pull/36495 and https://github.com/openshift/release/pull/36356


andfasano:
/hold

Requires https://github.com/openshift/installer/pull/7266

osherdp (approved):

PR #40564, opened 2023-06-22
AGENT-662: agent-installer-utils add e2e tests

andfasano:
This patch adds the sno/compact/ha agent jobs for the agent-installer-utils repo

### DIFF
--- a/ci-operator/config/openshift/agent-installer-utils/openshift-agent-installer-utils-main.yaml
+++ b/ci-operator/config/openshift/agent-installer-utils/openshift-agent-installer-utils-main.yaml
@@ -3,6 +3,10 @@ base_images:
     name: "4.14"
     namespace: ocp
     tag: base
+  dev-scripts:
+    name: test
+    namespace: ocp-kni
+    tag: dev-scripts
   ocp_builder_rhel-8-golang-1.20-openshift-4.14:
     name: builder
     namespace: ocp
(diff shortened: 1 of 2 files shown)


zaneb:
Would be good to use skip_if_only_changed to ignore docs/OWNERS file changes.
Is there a bot that syncs this to the release-4.14 and release-4.15 configs? If not we need to copy these same configs over so we don't lose them on release-4.14 once it branches off master.

andfasano:
> Would be good to use skip_if_only_changed to ignore docs/OWNERS file changes. Is there a bot that syncs this to the release-4.14 and release-4.15 configs? If not we need to copy these same configs over so we don't lose them on release-4.14 once it branches off master.

Good point for skip, added it. I've checked about the config sync,  and from the [docs](https://docs.ci.openshift.org/docs/architecture/branching/#do-i-need-to-make-sure-branched-ci-operator-configs-are-up-to-date-for-release-branches) it seems it will happen for 4.15 but not for 4.14, so added also it

andfasano:
> > Would be good to use skip_if_only_changed to ignore docs/OWNERS file changes. Is there a bot that syncs this to the release-4.14 and release-4.15 configs? If not we need to copy these same configs over so we don't lose them on release-4.14 once it branches off master.
> 
> Good point for skip, added it. I've checked about the config sync, and from the [docs](https://docs.ci.openshift.org/docs/architecture/branching/#do-i-need-to-make-sure-branched-ci-operator-configs-are-up-to-date-for-release-branches) it seems it will happen for 4.15 but not for 4.14, so added also it

From a double check with DPTP it seems that the promotion target of `main` (so `4.14`) and the release branches based on that (so `4.15`) will be kept in sync by the CI

rwsu (approved):

PR #40590, opened 2023-06-23
Bulk frequency reduce of 4.12 periodics to limit costs of CI platform

jmguzik:
If you were affected and think it's a mistake either revert new value or contact us on #forum-testplatform


jmguzik:
I will open a new one and link here.

PR #40617, opened 2023-06-25
Add tide configurations for bug-master repos

osherdp:
Making it possible for PRs that are approved and passing tests to just get automatically merged. This is probably what's currently missing for it to happen.


eliorerz (approved):
LGTM

PR #40620, opened 2023-06-25
Drop 4.9 to 4.10 sts upgrade jobs

shellyyang1989:
4.9 doesn't include AWS s3 ACLs policy [fix](https://github.com/openshift/cloud-credential-operator/pull/531), so drop the related tests to avoid false negative signals.

cc @jianping-shu @yunjiang29 @jhou1 for review. Thank you.


jianping-shu:
LGTM

PR #40639, opened 2023-06-26
Unify the name for aws-ipi-ovn-efs-fips

liangxia:
The workflow is `cucushift-installer-rehearse-aws-ipi-ovn-efs`, and `fips` is enabled via env, so rename the test name from `aws-ipi-ovn-fips-efs` to `aws-ipi-ovn-efs-fips`


Phaow:
LGTM

PR #39762, opened 2023-05-26
OCPQE-15002 && OCPQE-15003: Migrate aws efs ci profiles

Phaow:
## Migrate aws efs ci profiles from jenkins to Prow
- [OCPQE-15002](https://issues.redhat.com/browse/OCPQE-15002) Migrate 10_aarch64_Disconnected UPI on AWS & EFS CI profile to Prow
- [OCPQE-15003](https://issues.redhat.com/browse/OCPQE-15003) Migrate 13_aarch64_IPI on AWS & OVN & STS & EFS CI profile to Prow

PR #40647, opened 2023-06-26
OCPQE-15726: Restore periodic ci jobs reporter config

Phaow:
## Fix: restore periodic ci jobs reporter config
**Root cause**
- It seems sometimes  when we `make jobs` the auto generate jobs config removed the `reporter_config`(Maybe a bug of prow `make jobs`). Some `reporter_config` was deleted in https://github.com/openshift/release/pull/39762 caused some longDuring jobs don't notify the results in slack.

**Fix solution**
- Restore the `reporter_config` back


Phaow:
@jianping-shu @LiangquanLi930 @fxierh @sunzhaohua2  Could you please help take a look when you get a chance? It should no risk as the change only have added `reporter_config` which was removed somehow by `make jobs` previously. Thanks!

sunzhaohua2:
Thank you!
/lgtm

PR #40649, opened 2023-06-26
remove console cases when console is not installed

yapei:


yapei:
@yanpzhan Could you help review?

yanpzhan:
@yapei the update if ok for exclude console cucushift case, I see openshift-extended-web-tests is in openshift-e2e-test-qe chain, seems we need also exclude console case in cypress.

yapei:
@yanpzhan Good point, for Cypress cases, we need update [openshift-extended-web-tests](https://github.com/openshift/release/blob/master/ci-operator/step-registry/openshift-extended/web-tests/openshift-extended-web-tests-commands.sh)

yanpzhan:
Leave Cypress part to later pr, we could merge this update first.
/lgtm

PR #40653, opened 2023-06-26
Publish appliance image with `latest` tag

jhernand:
This patch changes the job that publishes the appliance image so that it uses the `latest` tag instead of `latest-<git hash>`.


jhernand:
I am closing this because the alternative in #40654 has already been approved and merged.

PR #40654, opened 2023-06-26
Publish openshift appliance image periodically

jhernand:
This patch changes the job that pusblishes the appliance image so that it will run periodically, once per day. An important side effect of this is that the image tag will automatically be `latest`.


ci-operator/config/openshift/appliance/openshift-appliance-master.yaml
```diff
@@ -45,7 +45,7 @@ tests:
   container:
     from: test-bin
 - as: publish
-  postsubmit: true
+  cron: '@daily'
```
danielerez, line 48: we also need 'IMAGE_TAG: latest' in env, right?

jhernand: I am not sure, I am rehearsing the job to check that.

jhernand: Apparently without the `IMAGE_TAG` the tag isn't `latest`: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_release/40654/rehearse-40654-periodic-ci-openshift-appliance-master-publish/1673295636100812800/artifacts/publish/assisted-baremetal-images-publish/build-log.txt . I added `IMAGE_TAG: latest`, will try again.

PR #40686, opened 2023-06-27
Pull latest codes for openshift-ansible

liangxia:
As we integrate openshift-ansible repo in verification-tests image, when openshift-ansible repo is updated, while the image is not refreshed, then we might got some upgrade failures.
Discussion in https://redhat-internal.slack.com/archives/C01UHADN91U/p1687336547378229

/cc @shellyyang1989 @gpei @jhou1 @jianlinliu


shellyyang1989:
Looks good to me. Thank you Liang for the fix!

PR #40700, opened 2023-06-27
Switch to hardcode list of openshift namespaces that get indexed in loki

dgoodwin:
We thought openshift-* was safe, 600k active streams later we realized
there are a pile of openshift-must-gather-RANDOM namespaces getting
matches, as well as amq, debug, and test.

Moving to a fixed list as I think this will be less maintenance than
playing wack-a-mole every time we see active streams blow up.

[TRT-1109](https://issues.redhat.com//browse/TRT-1109)


dgoodwin:
Tested locally
/pj-rehearse ack

PR #40708, opened 2023-06-27
ipi install: use integration cincinnati in CI clusters

petr-muller:


petr-muller:
> A total of 10556 jobs have been affected by this change.
:shipit: 

/pj-rehearse

PR #40740, opened 2023-06-28
use ruamel.yaml instead of yaml module

jianlinliu:
use ruamel.yaml module instead of yaml to avoid changing doulbe-quotes values to sigle-quote


jianlinliu:
cc @liangxia and @jianli-wei to review.

PR #40747, opened 2023-06-28
Enhance generate-cron-entry.sh to compatible with/without priority in…

liangxia:
… test name

```
$ ./generate-cron-entry.sh   alibaba-ipi-fips-p3-f28             openshift-openshift-tests-private-release-4.10__amd64-nightly.yaml
cron: 53 6 23 * *
$ ./generate-cron-entry.sh   alibaba-ipi-fips-p3-f28-destructive openshift-openshift-tests-private-release-4.10__amd64-nightly.yaml
cron: 8 12 23 * *
$ ./generate-cron-entry.sh   alibaba-ipi-fips-f28                openshift-openshift-tests-private-release-4.10__amd64-nightly.yaml
cron: 13 23 14 * *
$ ./generate-cron-entry.sh   alibaba-ipi-fips-f28-destructive    openshift-openshift-tests-private-release-4.10__amd64-nightly.yaml
cron: 24 2 8 * *
```

/cc @jianlinliu @jianli-wei @jhou1 @JianLi-RH @dis016

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/tools/generate-cron-entry.sh
+++ b/ci-operator/config/openshift/openshift-tests-private/tools/generate-cron-entry.sh
@@ -32,17 +32,15 @@ if [[ $DEBUG = "true" ]] ; then
 	echo "TEST_NAME: $TEST_NAME"
 	echo "YAML_FILE: $YAML_FILE"
 fi
-if ! [[ "$TEST_NAME" =~ p[1-3]-f[0-9]+ ]] ; then
-	echo "test_name must match [a-z0-9-]-p[1-3]-f[0-9]+(-.*)?"
+if ! [[ "$TEST_NAME" =~ (p[1-3])?-f[0-9]+ ]] ; then
+	echo "test_name must match [a-z0-9-](-p[1-3])?-f[0-9]+(-.*)?"
 	display_usage
 	exit 2
 fi
 
-FN_TMP="${TEST_NAME#*-p[1-3]-f}"
-FN="${FN_TMP%%-*}"
+FN="$(echo $TEST_NAME | sed -E 's/.*-f([0-9]+)(.*)?/\1/')"
 NUMBERS="$(echo $TEST_NAME $YAML_FILE | md5sum | tr -d [0a-z])"
 if [[ $DEBUG = "true" ]] ; then
-	echo "FN_TMP: $FN_TMP"
 	echo "FN: $FN"
 	echo "NUMBERS: $NUMBERS"
 fi


jianlinliu:
@liangxia when `priority` is set in the job name:
```
$ ./generate-cron-entry.sh   alibaba-ipi-fips-p3-f28             openshift-openshift-tests-private-release-4.10__amd64-nightly.yaml
cron: 53 6 23 * *
$ ./generate-cron-entry.sh   alibaba-ipi-fips-p3-f28-destructive openshift-openshift-tests-private-release-4.10__amd64-nightly.yaml
cron: 8 12 23 * *
```
The corn are set to the same day, is it possible to make it more different?

liangxia:
> The corn are set to the same day, is it possible to make it more different?

Once the `-p3` is remove, they are on different dates. Currently one of the job has been set with p2 to workaround the issue.

jianli-wei:
@liangxia I tried a bit with your updated script (see below results), I wonder after removing the "-pN", is it possible to allow a random number, i.e. "-r[0-9]", in order to get better differences among similar jobs? Thanks! 

- without "-pN", the 2 jobs would be 2  days away
- with "-p1" for the first job and "-p[1-3]" for the second job, they could be 10+ days away, which looks better

```
[jiwei@jiwei Downloads]$ ./generate-cron-entry.sh  gcp-ipi-proxy-tp-f28 openshift-openshift-tests-private-release-4.13__amd64-nightly.yaml
cron: 41 11 18 * *
[jiwei@jiwei Downloads]$ ./generate-cron-entry.sh  gcp-ipi-proxy-tp-f28-destructive openshift-openshift-tests-private-release-4.13__amd64-nightly.yaml
cron: 18 2 16 * *
[jiwei@jiwei Downloads]$ 
[jiwei@jiwei Downloads]$ ./generate-cron-entry.sh  gcp-ipi-proxy-tp-p1-f28 openshift-openshift-tests-private-release-4.13__amd64-nightly.yaml
cron: 42 9 6 * *
[jiwei@jiwei Downloads]$ ./generate-cron-entry.sh  gcp-ipi-proxy-tp-p2-f28 openshift-openshift-tests-private-release-4.13__amd64-nightly.yaml
cron: 17 4 18 * *
[jiwei@jiwei Downloads]$ ./generate-cron-entry.sh  gcp-ipi-proxy-tp-p3-f28 openshift-openshift-tests-private-release-4.13__amd64-nightly.yaml
cron: 35 8 26 * *
[jiwei@jiwei Downloads]$ 
[jiwei@jiwei Downloads]$ ./generate-cron-entry.sh  gcp-ipi-proxy-tp-p1-f28-destructive openshift-openshift-tests-private-release-4.13__amd64-nightly.yaml
cron: 38 1 23 * *
[jiwei@jiwei Downloads]$ ./generate-cron-entry.sh  gcp-ipi-proxy-tp-p2-f28-destructive openshift-openshift-tests-private-release-4.13__amd64-nightly.yaml
cron: 45 14 23 * *
[jiwei@jiwei Downloads]$ ./generate-cron-entry.sh  gcp-ipi-proxy-tp-p3-f28-destructive openshift-openshift-tests-private-release-4.13__amd64-nightly.yaml
cron: 15 20 23 * *
[jiwei@jiwei Downloads]$

**Comment by jianlinliu** [comment]
> is it possible to allow a random number, i.e. "-r[0-9]", in order to get better differences among similar jobs

I hope we do not introduce a new field just for generate more different cron item, or else, it sounds meaningless to remove priority filed. 

To generate more difference, my original thought is to always append a random seed to generate a much more different cron date inside the script to avoid such coincidence.

**Comment by jianli-wei** [comment]
> > is it possible to allow a random number, i.e. "-r[0-9]", in order to get better differences among similar jobs
> 
> I hope we do not introduce a new field just for generating more different cron items, or else, it sounds meaningless to remove the priority filed.
> 
> To generate more difference, my original thought is to always append a random seed to generate a much more different cron date inside the script to avoid such coincidence.

@jianlinliu Let me clarify a bit. My concerns are, on considering the script seems unaware of similar profiles, i.e. different profiles which use the same workflow, I guess it's hard for the script itself to generate very good cron entries, at least for the DAY_OF_MONTH in cron entries. So I suggest to allow manual interventions when necessary. Thanks!

**Comment by jianlinliu** [comment]
> different profiles which use the same workflow, I guess it's hard for the script itself to generate very good cron entries, at least for the DAY_OF_MONTH in cron entries

I understand the concern, and my proposal is - ` always append a random seed to generate a much more different cron date inside the script`. @liangxia wdyt?

**Comment by jianli-wei** [comment]
> my proposal is - ` always append a random seed to generate a much more different cron date inside the script`

@jianlinliu As the whole range of DAY_OF_MONTH is just 1 - 30, how the script knows whether a much more different cron date, if it's unaware of the cron dates of interested existing cron entries?

**Comment by liangxia** [comment]
> always append a random seed to generate a much more different cron date inside the script

Sounds like that will result in a different cron entry each time when we run the script for a specific test name.

**Comment by jianlinliu** [comment]
> Sounds like that will result in a different cron entry each time when we run the script for a specific test name.

This is a good point. How about this, take an example, the job name is `gcp-ipi-proxy-tp-f28`, when the script is starting to calculate the date, we can silently change the job name inside the script to `gcp-ipi-proxy-tp-N-f28`, here `N` equals the length of `gcp-ipi-proxy-tp`, I did not dig into the logic of calculating the cron date, did not run a real test there.

**Comment by liangxia** [comment]
The logic of the shell script is quite simple,
```
NUMBERS="$(echo $TEST_NAME $YAML_FILE | md5sum | tr -d [0a-z])"
```
Then take the first two bit of NUMBERS (%60) as MINUTE, and next wo bit (%24) as HOUR, ...
After you run this with thousands of test names, you will get several tests on the same date (or even the same hour).

**Comment by jianlinliu** [comment]
> After you run this with thousands of test names, you will get several tests on the same date (or even the same hour).

I agree with that. I do not think we can always generate a satisfied cron date in any time. I am just trying my best to relief @jianli-wei 's concern.

Per the jianli's example, if we use the above proposal to generate cron date, it would generate:
```
[jianlinliu@jialiu-laptop ~]$ sh mytest.sh gcp-ipi-proxy-tp-f28 openshift-openshift-tests-private-release-4.13__amd64-nightly.yaml
cron: 47 21 26 * *
[jianlinliu@jialiu-laptop ~]$ sh mytest.sh  gcp-ipi-proxy-tp-f28-destructive openshift-openshift-tests-private-release-4.13__amd64-nightly.yaml
cron: 33 13 3 * *
```

I added the following two lines:
```
TEST_NAME_LENGTH=${#TEST_NAME}
TEST_NAME="${TEST_NAME}-${TEST_NAME_LENGTH}"
```
The idea is trying to add more difference seed beside `TEST_NAME` and `YAML_FILE`.

liangxia:
Adding (or removing) any characters to md5sum will results to a different date/time, which means it makes some cron entries looks better, at the same time, some of them become worse.

jianlinliu:
@jianli-wei just like what @liangxia mentioned, though adding one more seed may resolve your issue on `gcp-ipi-proxy-tp-f28` case, but may make others worse. Another example is `alibaba-ipi-fips-f28` case,  the generated cron date looks good without adding more seeds. So looks like `gcp-ipi-proxy-tp-f28` is just a coincidence?

Another side, the reason of not adding one more field for user manual intervention is to make the job name be easier to understand, not have to repeat explaining each field's meaning to someone who are not familiar with the ci job name story. After all, different people has different insight on the naming conventions, we may can not leverage all the insights,  we want the job naming rule to be as much as easy-understanding.

For me, either (adding one more hidden seed or not) is okay for me.

jianli-wei:
@jianlinliu @liangxia Thanks! And let's live with it for now.

PR #40780, opened 2023-06-28
Test a `code-freeze-approved` label

dhaiducek:
What I'm unclear on right now is whether this will entirely override the org settings or if they can play nicely together. If it's the former, I'll have to rethink whether going this route is feasible.

The thinking here is there's an `excludedBranches` field we can add to this configuration that will allow us to exclude branches like `main` and any branch that's unfrozen. I've created a script that can be run to update all of our repos in `release` at once.

ref: https://issues.redhat.com/browse/ACM-6175


JustinKuli (approved):
Seems like a good place to experiment!

PR #40798, opened 2023-06-29
Fix HAC URL.

Katka92:
Change the old Stonesoup URL to the new `application-pipeline` one.


rhopp:
This shoud fix https://issues.redhat.com/browse/RHTAPBUGS-425

PR #40803, opened 2023-06-29
Add loadbalancer ref to the openstack ipi noramfs chain

gryf:


dulek:
But in general:
/lgtm

Let's see if the job runs normally.

pierreprinetti (approved):

PR #40820, opened 2023-06-29
Exclude `main` from code freeze

dhaiducek:
ref: https://issues.redhat.com/browse/ACM-6175

/cc @JustinKuli


JustinKuli (approved):

dhaiducek:
GitHub queries only have OR logic (which are leveraged by Tide), so using labels to gate PRs during code freezes isn't currently feasible.

PR #40843, opened 2023-06-30
Taking out incorrect pre submit 

paigerube14:
Pre submit had a test run in it. Talked with @svetsa-rh and she said the profile that was causing this issue is duplicated of another test. Removing the bad presubmit test


paigerube14:
CC: @nathan-weinberg this will remove that test as a required test in your pr

PR #38121, opened 2023-04-06
cri-o: add integration tests using kata containers

littlejawa:
kata containers are currently tested against cri-o PRs using a job running on the Jenkins infrastructure from the kata-containers project.
This is cumbersome, as most of the job's scripts are hosted and maintained on the kata side, independently from cri-o changes.

The reason for doing it was because cri-o was using VMs for the tests that did not support nested virtualization. This is not the case anymore, so we want to bring the job back to the cri-o framework, simplifying maintenance.

PR #40847, opened 2023-06-30
cri-o: disable the kata jenkins job

littlejawa:
Now that kata is tested with the same code as the other CI jobs for cri-o (see #38121), we don't need this job to run anymore.


littlejawa:
@haircommander @sohankunkerkar - I think this is all we need to disable the kata-jenkins job?
There are sources and VMs on the kata side, but we will take care of them separately.

sohankunkerkar (approved):

mrunalp (approved):

PR #40849, opened 2023-06-30
upkeep: change pv creation to be dynamic

eggfoobar:
PV creation was previously based on knowing the instance type, as we use things like metal it becomes error prone to have to keep updating the if statement. With this it's not 100% better but now we query based off of size and no longer need to know which volume path to attach.

### DIFF
--- a/ci-operator/step-registry/openshift/microshift/infra/aws/ec2/openshift-microshift-infra-aws-ec2-commands.sh
+++ b/ci-operator/step-registry/openshift/microshift/infra/aws/ec2/openshift-microshift-infra-aws-ec2-commands.sh
@@ -23,11 +23,6 @@ fi
 
 ami_id=${EC2_AMI}
 instance_type=${EC2_INSTANCE_TYPE}
-host_device_name="/dev/xvdc"
-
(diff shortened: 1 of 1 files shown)


copejon:
Looks like some file dependencies aren't being handled identically to non-metal tests

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_release/40849/rehearse-40849-pull-ci-openshift-microshift-main-microshift-metal-tests/1674822069247807488#1:build-log.txt%3A1054

dhellmann:
> Looks like some file dependencies aren't being handled identically to non-metal tests
> 
> https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_release/40849/rehearse-40849-pull-ci-openshift-microshift-main-microshift-metal-tests/1674822069247807488#1:build-log.txt%3A1054

The greenboot check for the VM timed out, so some of the files the data collection steps expect do not exist. I have that timeout set to 8m in my PR and it is more reliable. We should make the data collection more reliable, too.

dhellmann:
Is this the same as #40891? Which one do we need?

eggfoobar:
Closing this in favor of current work in #40891

PR #40853, opened 2023-06-30
Updating OWNERs for Cilium

jtaleric:
Adding myself and a few Isovalent/Cilium engineers to the OWNERs file for the Cilium directory.


nbusseneau:
@jtaleric Since this is not merged yet, can we also add `asauber` to the list?

@anybody Not sure how to move forward with the checks above, what are the additional steps we need to take to become members of the org? Is there a self-service tool, are we supposed to check in with someone specific?

PR #40873, opened 2023-07-03
Removed success slack report for periodics (unspam).

sagidayan:
Kept the format of the message as is, in case we want to add it back easily.


fabiendupont (approved):
/lgtm

PR #40726, opened 2023-06-27
Add feature-rearchitecture branch for odh operator

VaishnaviHire:
- Add `feature-rearchitecture` branch for opendatahub-operator repo.
- Add `feature-rearchitecture` odh-manifests bundle to the PR image

PR #40886, opened 2023-07-03
odh-operator: Set OCP 4.12 as test cluster

LaVLaS:
Updates the cluster version to 4.12 to bring the odh-operator tests inline with odh-manifest default cluster


LaVLaS:
Closing this since we are doing feature work on a branch that was added in #40726

PR #40887, opened 2023-07-03
update cuda image

heyselbi:
Updating cuda image to 12.2.0 in OVMS


Xaenalt:
It'll need updates to the cuda package versions in the dockerfiles as well

PR #40922, opened 2023-07-04
upi/gcp/nested: Update gcp image family to rhel9

praveenkumar:


ci-operator/step-registry/upi/gcp/nested/pre/upi-gcp-nested-pre-commands.sh
```diff
@@ -40,7 +40,7 @@ gcloud compute firewall-rules create "${INSTANCE_PREFIX}" \
 # image-family openshift4-libvirt must exist in ${GOOGLE_PROJECT_ID} for this template
```
cfergeau, line 40: `openshift4-libvirt-rhel9`

cfergeau (commented):
Looks good to me, but I suspect there are more users of this than just crc?

praveenkumar:
> Looks good to me, but I suspect there are more users of this than just crc?

@cfergeau yes, https://github.com/openshift/release/pull/40922#issuecomment-1619874806 have all the jobs which is going to affected.

cfergeau:
Looks good to me, I'm assuming even openshift-install 4.6 was supported when running on rhel 9? Do we need to keep jobs for all these openshift versions?
/lgtm

I'll approve once https://github.com/ironcladlou/openshift4-libvirt-gcp/pull/46/ lands, as before this lands we don't have the required image.

PR #40925, opened 2023-07-04
ROX-17337: increase Stackrox OCP 4.13 cluster pool size

tommartensen:
AWS Quotas have been increased to accomodate 20 parallel clusters.
Starting with a max size of 10, allowing all to hibernate, test, eventually increase max size (only!) to 20.


gavin-stackrox (approved):
/lgtm

PR #40932, opened 2023-07-04
Use latest ubi8/nodejs-16 img to build gpu plugin

rawagner:


rawagner:
/pj-rehearse ack

IIUC we need this merged in order to mirror to happen & build to pass

PR #40968, opened 2023-07-05
[openshift/cert-manager-operator] Protect branches cert-manager-*,master in prow config

swghosh:
Signed-off-by: Swarup Ghosh <swghosh@redhat.com>


swghosh:
/cc @thejasn
PTAL

PR #40978, opened 2023-07-05
OCP Logging: use vector 5.8 with CLO 5.7

syedriko:
OCP Logging: use vector 5.8 with CLO 5.7


jcantrill:
/retitle OCP Logging: use vector 5.8 with CLO 5.7

PR #40991, opened 2023-07-06
Check config/worker format

liangxia:


ci-operator/step-registry/openshift/upgrade/qe/test/cert-rotation/openshift-upgrade-qe-test-cert-rotation-commands.sh
```diff
@@ -119,6 +119,11 @@ oc adm restart-kubelet nodes --all --directive=RemoveKubeletKubeconfig
 oc adm reboot-machine-config-pool mcp/worker mcp/master
 oc adm wait-for-node-reboot nodes --all
 
+# WARNING: On some platforms this step may prevent new nodes from joining the cluster. To verify whether this step is safe to perform on your cluster, please follow this process:
```
jhou1, line 122: cc @sunzhaohua2

ci-operator/step-registry/openshift/upgrade/qe/test/cert-rotation/openshift-upgrade-qe-test-cert-rotation-commands.sh
```diff
@@ -119,6 +119,11 @@ oc adm restart-kubelet nodes --all --directive=RemoveKubeletKubeconfig
...
+workerUserData="$(oc -n openshift-machine-api get secret/worker-user-data -o=jsonpath='{.data.userData}' | base64 -d)" || true
+workerUserDataManaged="$(oc -n openshift-machine-api get secret/worker-user-data-managed -o=jsonpath='{.data.userData}' | base64 -d)" || true
```
sunzhaohua2, line 124: @liangxia the two commands' result will be in the following format:
`{"ignition":{"config":{"merge":[{"source":"$URL"}]},"security":{"tls":{"certificateAuthorities":[{"source":"data:text/plain;charset=utf-8;base64,..."}]}},"version":"3.2.0"}}`

For the URL, if it looks like "https://api-int.$CLUSTERNAME.$DOMAIN:22623/config/worker", the cluster is safe to run the following steps. If you see the IP address directly instead, looks like https://192.168.0.1:22623/config/worker do NOT perform this step.

jhou1:
This can make the automation not breaking, but given we already know on this condition things would break, we should not run those profiles we know would be broken.

Considering a proper fix would have to land, there would be a need for us to remove this condition after we merge this. Personally I feel we are short of a good reason to include this in automation at the moment.

liangxia:
> This can make the automation not breaking, but given we already know on this condition things would break, we should not run those profiles we know would be broken.
> Considering a proper fix would have to land, there would be a need to remove this condition after a fix is built. Personally I feel we are short of a good reason to include this in automation at the moment.

The cert-rotation tests have 29 steps, we are conditionally skip one step. We still can validate the other 28 steps.

PR #40992, opened 2023-07-06
disable azure marketimage testing

jianlinliu:
The azure marketplace testing is running in a specific account, the account owner reminds QE again not to use it until the bill paid issue get resolved.

Set `always_run: false` to replace the cron, also add "disabled-" keyword to avoid it getting updated when running `update-cron-entries.py` tools.

Once we got notification from the account owner, we would enable it again.


jianlinliu:
cc @liangxia @jinyunma and @gpei  to review

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.11__amd64-nightly.yaml
```diff
@@ -1092,8 +1092,8 @@ tests:
     - ref: idp-htpasswd
     - ref: openshift-extended-test-longduration
     workflow: cucushift-installer-rehearse-azure-ipi-private
-- as: azure-ipi-private-workers-marketimage-p2-f14
-  cron: 18 7 13,28 * *
+- as: baremetal-azure-ipi-private-workers-marketimage-p2-f28
```
liangxia, line 1095: Adding `disabled` to the ignore keyword in https://github.com/openshift/release/pull/40993
So that instead of using prefix baremetal for azure profiles, we can use `disabled` as the prefix.

PR #41013, opened 2023-07-06
Migrate from ipi-aws workflow to hypershift workflows

flacatus:
During last weeks we start to see that running our e2e in ipi-aws is very flaky. This PR changes from ipi-aws to hypershift tests which seems more stable.


albarbaro (approved):

elsony (approved):
/lgtm

PR #41027, opened 2023-07-06
Add yearly cron to 3 node worker tests that will be used for adhoc te…

svetsa-rh:
…sting. This is to by-pass tests getting added to presubmit jobs, which is forcing each and every PR to run these tests unnecessarily.


svetsa-rh:
@paigerube14 Can you ptal?

PR #41045, opened 2023-07-07
Update cert-manager to 1.12.2 on master

thejasn:


swghosh:
OO_LATEST_CSV: 1.12.2 is fine, since we have https://github.com/openshift/cert-manager-operator/pull/134 upcoming. The rehearsal test was failing as a result of the fact that master still uses 1.11.4, test will pass on master once 1.12 is in.

/pj-rehearse ack

PR #41049, opened 2023-07-07
hypershift: temporary job

LiangquanLi930:
We need to obtain the FIPS'RP this week, so we have submitted this temporary PR (Pull Request). We will roll it back next week.


liangxia:
@LiangquanLi930 Check if this met your requirements,
https://github.com/openshift/release/blob/master/ci-operator/config/openshift/openshift-tests-private/README.md#trigger-a-new-test-for-an-existing-periodic-job-in-prow

PR #40385, opened 2023-06-15
Move CCO periodics to a separate file

suhanime:
Follow guidelines as per
https://docs.ci.openshift.org/docs/how-tos/naming-your-ci-jobs/#periodics-outside-of-openshiftrelease and move CCO periodics to a separate file.

xref: https://issues.redhat.com/browse/CCO-367

PR #41065, opened 2023-07-07
CCO-368: Add a periodic version of the STS e2e tests

bentito:
This test will be called weekly.


bentito:
Does this seem right for causing the `make test-e2e-sts` target to get called weekly? And weekly seems like the right interval for this given the PR frequency for CCO?

suhanime:
@bentito I have added the make target to the periodic test in [40385](https://github.com/openshift/release/pull/40385) (Thanks for pointing it out). I think we can close this PR in favour of 40385.

PR #41066, opened 2023-07-07
migtools/volume-snapshot-mover: `unmanaged: true`

kaovilai:
So that we can set custom rules such as requiring two approvals prior to merge

### DIFF
--- a/core-services/prow/02_config/migtools/volume-snapshot-mover/_prowconfig.yaml
+++ b/core-services/prow/02_config/migtools/volume-snapshot-mover/_prowconfig.yaml
@@ -3,16 +3,7 @@ branch-protection:
     migtools:
       repos:
         volume-snapshot-mover:
-          branches:
-            master:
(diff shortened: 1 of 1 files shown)


hhpatel14 (approved):
This change will not have Prow manage the branch protection. 

Since this repo is not part of openshit, we can use github actions to push images. I see the changes are enabled for that. I will make further changes to remove this repo from openshift/release.

sseago (approved):

sseago:
@hhpatel14 Semi-related note: we don't want to make this change to the openshift/velero* repos. When we were running into force-push issues related to fork rebases, we tried this setting and it failed. Not an issue for volume-snapshot-mover, but just be aware that we don't want to do this for the fork repos.

kaovilai:
> Since this repo is not part of openshit, we can use github actions to push images. 

That would probably make it faster for e2e in operator to become green again.

PR #41072, opened 2023-07-07
[WIP] Update region

svetsa-rh:
@paigerube14 @qiliRedHat Can you ptal? This is with reference to: https://redhat-internal.slack.com/archives/C03ABT5822W/p1688758000292339 and https://redhat-internal.slack.com/archives/C03ABT5822W/p1688758342856999

Couple of noteworthy mention of **NOTES** below.

**NOTE:** I had to add default REGION to all ref's to by-pass ci-config-operator error checks which otherwise fail to resolve REGION variable. See below sample error.

Example error:
time="2023-07-07T20:25:36Z" level=error error="failed to validate configuration openshift-qe/ocp-qe-perfscale-ci/openshift-qe-ocp-qe-perfscale-ci-main__aws-4.14.yaml: Failed resolve MultiStageTestConfiguration: test/ocp-qe-perfscale-ci-tests-net-perf: workflow/openshift-qe-network-perf: parameter \"REGION\" is overridden in [test/ocp-qe-perfscale-ci-tests-net-perf] but not declared in any step"
time="2023-07-07T20:25:36Z" level=fatal msg="error validating configuration files"

**NOTE:** With this change, we need to update our GCP, Azure and ARO tests to add REGION for each test.

I am open to other ideas, if there are better ways of setting the REGION. 

Option1) Go with changes made in this PR.
Option 2) I don't know the exact step where AWS cluster is being created for non-Rosa clusters to set it there but also apprehensive about what impact that would have for rest of the organization to change it at that level since many many teams might be affected by that change.
Option 3) If it's easier, I can update for Rosa clusters alone for now and we can update non-Rosa clusters at a different time in a different PR after we come to a conclusion. LMK what you guys think.
Update:
Option 4) Once we identify the right step in common workflow for classic OCP AWS cluster creation, add an option for REGION with "" (blank) value and see if it is acceptable solution for everyone that will get affected. This needs to be done from a separate PR to avoid confusion and get approvals from correct OWNERS. 
Then update this PR.

### DIFF
--- a/ci-operator/step-registry/openshift-qe/network-perf/openshift-qe-network-perf-ref.yaml
+++ b/ci-operator/step-registry/openshift-qe/network-perf/openshift-qe-network-perf-ref.yaml
@@ -5,6 +5,10 @@ ref:
     name: ocp-qe-perfscale-ci
     tag: latest
   commands: openshift-qe-network-perf-commands.sh
+  env:
+  - name: REGION
+    default: "us-west-2"
+    documentation: Use a specific AWS region, overriding the LEASED_RESOURCE environment variable in the cluster_profile.
   credentials:
   - namespace: test-credentials # this entry injects the custom credential
     name: ocp-qe-perfscale-es
--- a/ci-operator/config/openshift-qe/ocp-qe-perfscale-ci/openshift-qe-ocp-qe-perfscale-ci-main__aws-4.14.yaml
+++ b/ci-operator/config/openshift-qe/ocp-qe-perfscale-ci/openshift-qe-ocp-qe-perfscale-ci-main__aws-4.14.yaml
@@ -40,6 +40,7 @@ tests:
     cluster_profile: aws-qe
     env:
       BASE_DOMAIN: qe.devcluster.openshift.com
+      REGION: us-west-2
       WORKER_REPLICA_COUNT: "6"
     post:
     - chain: ipi-aws-post
@@ -53,6 +54,7 @@ tests:
     cluster_profile: aws-qe
     env:
       BASE_DOMAIN: qe.devcluster.openshift.com
+      REGION: us-west-2
       WORKER_REPLICA_COUNT: "20"
     post:
     - chain: ipi-aws-post
@@ -67,6 +69,7 @@ tests:
     env:
       BASE_DOMAIN: qe.devcluster.openshift.com
       PODS_PER_NODE: "200"
+      REGION: us-west-2
       WORKER_REPLICA_COUNT: "20"
     post:
     - chain: ipi-aws-post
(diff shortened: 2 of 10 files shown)


qiliRedHat:
@svetsa-rh Please check the use-west-2's quota is enough in the account.

svetsa-rh:
> @svetsa-rh Please check the use-west-2's quota is enough in the account.

@qiliRedHat I have checked quota on AWS console before making changes here. Hope that suffices.

ci-operator/step-registry/openshift-qe/network-perf/openshift-qe-network-perf-ref.yaml
```diff
@@ -5,6 +5,10 @@ ref:
     name: ocp-qe-perfscale-ci
     tag: latest
   commands: openshift-qe-network-perf-commands.sh
+  env:
+  - name: REGION
```
paigerube14, line 9: This variable has no application in our tests. I dont think it should be added here

paigerube14:
Take a look at https://steps.ci.openshift.org/reference/ipi-conf-aws, there is no region environment variable thats why you're getting the error. Need to find which variables we can properly use to set this

paigerube14:
We may need to edit [this line](https://github.com/openshift/release/blob/f7a153e44d65965bc9ccf6d331e0e1e608843d58/ci-operator/step-registry/ipi/conf/aws/ipi-conf-aws-commands.sh#L20) to use something [like this](https://github.com/openshift/release/blob/f7a153e44d65965bc9ccf6d331e0e1e608843d58/ci-operator/step-registry/osd-ccs/fleet-manager/provision-shards/osd-ccs-fleet-manager-provision-shards-commands.sh#L10) and add 'REGION' as an environment variable to the ipi-aws-conf step 
I'm not sure where LEASED_RESOURCE is set but I see in some refs they try to edit that variable

svetsa-rh:
@paigerube14 Did you see Option 4 I mentioned here https://github.com/openshift/release/pull/41072#issue-1794428975 
What do you think of going with that option?

paigerube14:
@svetsa-rh ah yes, need to do option 4. But also take out the addition of the regions in the ref files in our steps in this PR and take out region from the aws-4.14 config file

svetsa-rh:
> @svetsa-rh ah yes, need to do option 4. But also take out the addition of the regions in the ref files in our steps in this PR and take out region from the aws-4.14 config file

Yeah @paigerube14, once that PR for AWS common code gets merged, I will rebase and make changes here. Yes, no need at that point.
I'll edit subject to "WIP" to avoid confusion.

paigerube14:
@svetsa-rh do we need this pr still? I think the profile we have set includes this region without this update

svetsa-rh:
All needed changes were already merged. This is old PR and no longer needed.

PR #41073, opened 2023-07-07
Add kuttl tests in e2e workflow for infra

vhariria:


ci-operator/jobs/openstack-k8s-operators/infra-operator/openstack-k8s-operators-infra-operator-main-presubmits.yaml
```diff
@@ -6,6 +6,68 @@ presubmits:
     - ^main$
     - ^main-
     cluster: build02
+    context: ci/prow/infra-operator-build-deploy-kuttl
```
cescgina, line 9: We should set `always_run: false` in the build-deploy job in line 160, so that we don't have both the build-deploy and build-deploy-kuttl jobs running for each PR

PR #41101, opened 2023-07-10
Adjust build_root config for console-plugin-nvidia-gpu

mresvanis:
After switching to the latest nodejs-16 `build_root` image with [this](https://github.com/openshift/release/pull/40932) PR, it seems that the former does not contain the appropriate permissions on the `/go` [directory](https://github.com/rh-ecosystem-edge/console-plugin-nvidia-gpu/pull/60#issuecomment-1625036007). Looking into the CI documentation regarding the [build root image](https://docs.ci.openshift.org/docs/architecture/ci-operator/#build-root-image), it seems that we don't need a specific nodejs base image defined there, but a go image that contains the necessary permissions and tools (like one of the base go images).


mresvanis:
@mareklibra @jelkosz would you have some time to review this change? We seem to need it in order

mareklibra:
/lgtm

@mresvanis , Thank you for maintaining it.

PR #41122, opened 2023-07-11
remove 4-dev-preview stream when 4.13 get GAed

jianlinliu:
4.13 get GAed, so remove the 4-dev-preview stream setting now.


jianlinliu:
cc @liangxia to review.

PR #41144, opened 2023-07-11
HyperShift: add external-dns env

LiangquanLi930:
We need to add a env:`HYPERSHIFT_EXTERNAL_DNS_DOMAIN `, to meet the current testing requirements.


bryan-cox (approved):
/lgtm

PR #41154, opened 2023-07-11
Tweak rukpak jobs:

ncdc:
- Add verify-deps
- Add verify-manifests
- Add verify-commits

### DIFF
--- a/ci-operator/config/openshift/operator-framework-rukpak/openshift-operator-framework-rukpak-main.yaml
+++ b/ci-operator/config/openshift/operator-framework-rukpak/openshift-operator-framework-rukpak-main.yaml
@@ -42,6 +42,22 @@ tests:
   commands: make verify
   container:
     from: src
+- as: verify-deps
+  commands: |
+    # Verify main repo deps
+    make -f openshift/Makefile verify-deps
+    # Verify openshift dir
+    cd openshift && make verify-deps
+  container:
+    from: src
+- as: verify-manifests
+  commands: make -f openshift/Makefile verify-manifests
+  container:
+    from: src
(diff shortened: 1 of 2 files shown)


ncdc:
The verify-commits rehearsal is failing because it's using the PULL_BASE_SHA from this repo, when it should be from the target repo. This is expected in the rehearsal, and I expect an actual PR against the target repo to work.

/pj-rehearse ack

ci-operator/config/openshift/operator-framework-rukpak/openshift-operator-framework-rukpak-main.yaml
```diff
@@ -42,6 +42,22 @@ tests:
   commands: make verify
   container:
     from: src
+- as: verify-deps
+  commands: |
+    # Verify main repo deps
+    make -f openshift/Makefile verify-deps
```
dtfranz, line 48: I'm not seeing this make target in the repo, is it something you're in the process of adding?

grokspawn: It's part of the build-machinery: https://github.com/openshift/operator-framework-rukpak/blob/main/openshift/Makefile#L3

grokspawn: By the transitive process of [lots-of-includes]...
https://github.com/openshift/operator-framework-rukpak/blob/main/openshift/vendor/github.com/openshift/build-machinery-go/make/targets/openshift/deps-gomod.mk

PR #41158, opened 2023-07-11
Undo yearly cron for adhoc tests. 

svetsa-rh:
This change should prevent future tests to force and run these tests as mandatory.


paigerube14:
I think this will still add these as jobs to run when a pr is opened in the main branch... guess we could move common scripts out of main branch to somewhere else

svetsa-rh:
Thanks Paige, was just going to try testing "config" :-)

paigerube14:
very weird it said all passed but one was listed as failed...

PR #41176, opened 2023-07-12
[SSP-operator] Adding user to subuid

assafad:
Adding user to subuid within unittest in kubevirt-ssp-operator-main.yaml.


lyarwood:
@assafad Stupid question but why do we suddenly need this? There's no justification in the PR description or commit message.

0xFelix:
@lyarwood Assaf got in contact with me to try fixing https://github.com/kubevirt/ssp-operator/pull/621, this was my suggestion after also talking to Karel. But yeah, would be good to add a ref to the initial PR.

lyarwood:
I wonder if we just need the `preset-podman-in-container-enabled: true` label enabled here as we use downstream to enable the use of podman within these containers.

PR #41284, opened 2023-07-14
HOSTEDCP-413: Added HO upgrade step before e2e

muraee:


ci-operator/step-registry/hypershift/aws/e2e/nested/hypershift-aws-e2e-nested-workflow.yaml
```diff
@@ -31,4 +31,6 @@ workflow:
     pre:
     - ref: ipi-install-rbac
     - chain: hypershift-aws-setup-nested-management-cluster
+    - ref: hypershift-install-initial
```
csrwng, line 34: Is there a way to simply call the existing `hypershift-install` step with a different `HYPERSHIFT_RELEASE_LATEST` value? That would help to keep the code DRY

muraee: it might be possible, but the problem is that we also need to override the `from` image so we also use the initial version of the CLI. This is not possible afaik.

sjenning:
nice!
/approve
/lgtm

muraee:
/pj-rehearse pull-ci-openshift-hypershift-release-4.13-e2e-aws

PR #41316, opened 2023-07-17
Remove skip if condition from bundle e2e test job config

malayparida2000:


iamniting (approved):

malayparida2000:
Due to the skip if condition the test is not always run. But in ocs operator, we want to mark this test as a required one, that needs to pass for a PR to merge. So it being optional here & being required in the branch settings at the same time gives a conflict & we are unable to mark it as required in ocs operator. So we are trying to remove the skip if config & see if marking it as required works.

PR #41343, opened 2023-07-17
SDN-3398 add 4.x-4.y aws-ovn upgrade presubmits for CNO

jluhrsen:


tssurya:
/retitle SDN-3398 add 4.x-4.y upgrade presubmits for CNO

PR #41345, opened 2023-07-17
Partially revert commit that removed jira/valid-bug in storage reposotories

bertinatto:
Partially reverts  66c151b4ae3be3ade693cac1396b52e64017615e:

1. Only targeting storage repositories.
2. Avoid having 2 rules for `main/master` requiring `jira/valid-bug`

CC @openshift/storage


jsafrane:
Since csi-driver-nfs is here, I miss csi-driver-manila-operator and openstack-cinder-csi-driver-operator. Or NFS should not be here, our choice.

gabemontero (approved):
/approve

PR #41346, opened 2023-07-17
remove opentelemetry

smg247:
app is not installed in org, breaking tide. I believe if we simply remove the tide config we will get around the issue


smg247:
/cc @IshwarKanse
Once you get to the bottom of https://redhat-internal.slack.com/archives/CBN38N3MW/p1689598760755169 and either install the app or fork the repo, you can revert this

PR #41371, opened 2023-07-18
vSphere build02 dex config

jcpowermac:


jcpowermac:
sync directly into `dex` namespace

```
secretsync/target-name vsphere-build02
secretsync/target-namespace dex
vsphere-build02-id
vsphere-build02-secret
```

PR #33905, opened 2022-11-09
core-services/image-mirroring: Mirror 4.12 image for COSA

travier:

PR #41372, opened 2023-07-18
update RHCOS CI for 4.14

mike-nguyen:


ci-operator/config/coreos/coreos-assembler/coreos-coreos-assembler-rhcos-4.14.yaml
```diff
@@ -0,0 +1,31 @@
+build_root:
+  project_image:
+    dockerfile_literal: |
+      FROM registry.fedoraproject.org/fedora:37
```
travier, line 4:
```suggestion
      FROM registry.fedoraproject.org/fedora:38
```
?

travier: Wondering if we missed other instances of that for 4.13

travier: 4.13 is on F37: https://github.com/coreos/coreos-assembler/blob/rhcos-4.13/Dockerfile

So it's only here

dustymabe (approved):

ci-operator/config/openshift/os/openshift-os-release-4.14.yaml
```diff
@@ -2,7 +2,7 @@ base_images:
   coreos_coreos-assembler_latest:
```
marmijo, line 2:
```suggestion
  coreos_coreos-assembler_4.14:
```

In the referenced PR: https://github.com/openshift/release/pull/33905, this line was changed to indicate the version. Want to do that here too?

mike-nguyen: I missed that change!  Thanks @marmijo

mike-nguyen:
This job adds a mirror for a container image that is used in the job so pj-rehearse jobs are failing.

marmijo (approved):

PR #41380, opened 2023-07-18
WIP: arm64: test SNO jobs with max-parallel-tests set to 6

Prashanth684:
seeing a bunch of failures in the arm64 sno jobs which we feel is related to performance when setting parallel tests to 15.

### DIFF
--- a/ci-operator/step-registry/single-node/e2e/test/single-node-e2e-test-commands.sh
+++ b/ci-operator/step-registry/single-node/e2e/test/single-node-e2e-test-commands.sh
@@ -262,6 +262,12 @@ function upgrade_paused() {
 
 # Preserve the && chaining in this function, because it is called from and AND-OR list so it doesn't get errexit.
 function suite() {
+    if [[ "${OCP_ARCH}" == "arm64" ]]; then
+        MAX_PARALLEL_TESTS=6
+    else
+        MAX_PARALLEL_TESTS=15
+    fi
+
     if [[ -n "${TEST_SKIPS}" ]]; then
         TESTS="$(openshift-tests run --dry-run --provider "${TEST_PROVIDER}" "${TEST_SUITE}")" &&
         echo "${TESTS}" | grep -v "${TEST_SKIPS}" >/tmp/tests &&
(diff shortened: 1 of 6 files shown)


Prashanth684:
/pj-rehearse periodic-ci-openshift-multiarch-master-nightly-4.14-ocp-e2e-aws-ovn-arm64-single-node

ci-operator/step-registry/single-node/e2e/test/single-node-e2e-test-commands.sh
```diff
@@ -262,6 +262,12 @@ function upgrade_paused() {
 
 # Preserve the && chaining in this function, because it is called from and AND-OR list so it doesn't get errexit.
 function suite() {
+    if [[ "${OCP_ARCH}" == "arm64" ]]; then
+        MAX_PARALLEL_TESTS=6
```
jeffdyoung, line 266: @Prashanth684 based on the load I observed on x86, I'd say 6 tests is the right number for both archs. The excessive load is causing x86 flakes.

Prashanth684: adding some representatives from the sno team if we are going to change this for x86 as well: @eranco74 @omertuc

omertuc: >The excessive load is causing x86 flakes.

Could you elaborate on what kind of flakes you think it's causing? Any particular examples? 

I'm for changing it to 6 from 15, if it doesn't extend the runtime too long, as that might lead to flakes in the form of CI timeouts

Prashanth684:
/pj-rehearse periodic-ci-openshift-multiarch-master-nightly-4.14-ocp-e2e-aws-ovn-arm64-single-node

ci-operator/step-registry/single-node/e2e/test/single-node-e2e-test-commands.sh
Prashanth684: @jeffdyoung are the flakes seen in your testing or in general CI as well? this does significantly raise test run time so we have to be careful which is why I wanted to enable on arm64 first.

jeffdyoung: @Prashanth684 its during my test on x86 (m5d.2xlarge):
6 max = 5 to 6 failures. runtime ~2hr10mins. 
Max6 load screenshot:
max6


15 max = 7 to 12 failures. runtime ~1hr10mins.
Max15 load screenshot:
max15


6 threads seems to give more consistent results at a better load at the cost of an extra hour. I would expect arm64 to behave the same once the openvswitch fix lands in rhcos.

PR #41386, opened 2023-07-18
openstack: deploy ctlplane with rootVolume too

EmilienM:
When testing the rootVolume in a deployment, also include the
controlplane, so later this can be tested by the CPMS.

This might have a negative impact on the performance, since etcd will be
on a cinder volume attached via Ceph, and we know the performance
impact, so we need to be careful about this change.

### DIFF
--- a/ci-operator/step-registry/openstack/conf/installconfig/openstack-conf-installconfig-commands.sh
+++ b/ci-operator/step-registry/openstack/conf/installconfig/openstack-conf-installconfig-commands.sh
@@ -95,8 +95,11 @@ esac
 
 if [[ "${ZONES_COUNT}" -gt '0' ]]; then
 	yq --yaml-output --in-place ".
-		| .compute[0].platform.openstack.zones = ${ZONES_JSON}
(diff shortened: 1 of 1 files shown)


EmilienM:
/pj-rehearse periodic-ci-shiftstack-shiftstack-ci-main-periodic-4.14-e2e-openstack-kuryr

EmilienM:
/hold cancel

/cc mandre
are you fine with it? this should only have effects on the Kuryr job now and the CPMS job running in az0.

mandre:
> /cc mandre are you fine with it? this should only have effects on the Kuryr job now and the CPMS job running in az0.

From https://github.com/openshift/release/blob/29c1c67/ci-operator/step-registry/openstack/conf/resourcenames/openstack-conf-resourcenames-commands.sh#L37-L45, I can see it's only affecting jobs that run on the `openstack-kuryr` or `openstack-vh-mecha-az0` platforms since that's where we define compute AZs.

mandre (commented):
/lgtm

PR #41399, opened 2023-07-19
remove output xml files from test grid

Kartikey-star:
The output.xml files returned by helm test acceptance contains the word 'robot', which is causing the test grid to report failure. We also have another artifact for results so we would like to ignore the output files


dperaza4dustbit (commented):
/lgtm

PR #41405, opened 2023-07-19
Storing target cluster-name under SHARED_DIR

omrirh:


omrirh:
Hello @yasun1 ,
let me know if those changes can be approved as this blocks some of our jobs.
Thanks in advance!

ci-operator/step-registry/osd-ccs/cluster/provision/aws/osd-ccs-cluster-provision-aws-commands.sh
```diff
@@ -3,6 +3,7 @@
 set -o nounset
 set -o errexit
 set -o pipefail
+set -o verbose
```
yasun1, line 6: Let us close the verbose model.

omrirh: Done

PR #41415, opened 2023-07-19
subnet tags are requires for hcp

yasun1:


yasun1:
@yunjiang29 Please help review. Thanks.

PR #41422, opened 2023-07-19
MON-3295: add openshift/prometheus-promu config

jan--f:


jan--f:
Adding `OWNERS` is part of https://github.com/openshift/prometheus-promu/pull/1

jan--f:
@danilo-gemoli @droslean This is ready to review.
I removed the unit test job for now. I'll re-add that once I can update the code base based on the config proposed here. We don't build any images for openshift for this, so no tests don't hurt anything.

PR #41432, opened 2023-07-19
MGMT-14367: move assisted jobs to OCP 4.14

adriengentil:


gamli75 (approved):

osherdp:
@adriengentil I guess we can try the same thing here no https://github.com/openshift/release/pull/41031 ?

adriengentil:
> @adriengentil I guess we can try the same thing here no #41031 ?
> (if ztp failures are indeed just a mismatch of versions between openshift and hypershift)

thanks, I also moved `cluster-api-provider-agent` to 4.14

osherdp (approved):

PR #41437, opened 2023-07-19
openstack/rotate/cloud-credentials: add proxy support

EmilienM:
This is needed when the cluster is running behind a proxy.


EmilienM:
Unfortunately, there is no job that can be used for testing as the clusters deployed by jobs doing cloud credentials rotations don't run behind a proxy.

/pj-rehearse ack

PR #41447, opened 2023-07-19
cluster-olm-operator: add file skips for e2e

ncdc:


ncdc:
No need to rehearse; this just changes when to skip some jobs.
/pj-rehearse skip

PR #41457, opened 2023-07-19
Making the 4.13 Rollback job blocking

bradmwilliams:
Per [slack discussion](https://redhat-internal.slack.com/archives/CJARLA942/p1689747278596559), it has been decided to make the 4.13 Rollback job Blocking.


wking (approved):
/lgtm

PR #41499, opened 2023-07-20
hypershift: increase conformance cluster destroy timeout

sjenning:
We are observing timeouts in the `hypershift-aws-destroy` chain during aggregated conformance tests.  I believe this is due to AWS ratelimiting and so many jobs starting (lots of AWS create) and stopped (lots of AWS delete) happening all at once.

https://testgrid.k8s.io/redhat-hypershift#4.14-aws-ovn-conformance&width=20

The jobs that overran the timeout seemed to overrun it by <5m.  Doubling to the timeout to see if this avoids the failure.


bryan-cox (approved):
/lgtm

PR #41504, opened 2023-07-20
OCPBUGS 15573: Remove non-4.14 libvirt IBM Power/Z e2e tests from the release controller.

jaypoulz:
We're seeing some spikes of node-related outages in our P/Z libvirt jobs. By unlinking these e2e runs from the release controller, resources will be available for 4.14 runs. This will hopefully result in a more stable signal while we work on resolving the underlying cause. The intent is to potentially add these jobs back as informing jobs if we can adjust their frequency via the release controller, but the full implications of that are still being evaluated.


deads2k:
/approve

and I'll happily approve re-adding them if/when the timing and load concerns are resolved.

PR #41526, opened 2023-07-21
Fix run condition for e2e-agent-ha-dualstack tests

andfasano:
This patch will avoid that unrelated installer PRs will trigger e2e-agent-ha-dualstack test, thus helping in lowering the number of (equinix) instances allocated


pawanpinjarkar (approved):
/lgtm

PR #41547, opened 2023-07-21
openstack jobs skip CI for renovate/readme updates

dprince:
This updates the openstack jobs to skip CI if only renovate.json or the README.md changes

### DIFF
--- a/ci-operator/jobs/openstack-k8s-operators/cinder-operator/openstack-k8s-operators-cinder-operator-main-presubmits.yaml
+++ b/ci-operator/jobs/openstack-k8s-operators/cinder-operator/openstack-k8s-operators-cinder-operator-main-presubmits.yaml
@@ -63,7 +63,7 @@ presubmits:
           secretName: result-aggregator
     trigger: (?m)^/test( | .* )cinder-operator-build-deploy,?($|\s.*)
   - agent: kubernetes
-    always_run: true
+    always_run: false
     branches:
     - ^main$
     - ^main-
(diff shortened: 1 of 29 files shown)


ci-operator/jobs/openstack-k8s-operators/cinder-operator/openstack-k8s-operators-cinder-operator-main-presubmits.yaml
```diff
@@ -14,6 +14,7 @@ presubmits:
     name: pull-ci-openstack-k8s-operators-cinder-operator-main-cinder-operator-build-deploy
     optional: true
     rerun_command: /test cinder-operator-build-deploy
+    skip_if_only_changed: ^renovate.json|^README.md
```
viroel, line 17: Not sure what Prow is going to do here, because the job is already set as "always_run: false". 
This means that if PR is not renovate.json or README only, it will trigger the job?
if yes, we probably don't want to run this job in favor of kuttl/tempest one.

viroel: But we can merge to see how it works and fix if needed aftwerwards

viroel: According with the docs[1], we shouldn't add `skip_if_only_changed` if we don't really want to run these jobs that are already set to `always_run: false`. Thanks @dprince.

[1] https://docs.prow.k8s.io/docs/jobs/#standard-triggering-and-execution-behavior-for-jobs

viroel:
@Sandeepyadav93 can you merge this on monday and monitor if works as expected?

PR #41549, opened 2023-07-21
No churn option 

paigerube14:
Do not need to churn/delete the namespaces, just want to be able to have some load on a cluster. This will cut down on run time 
CC: @chaitanyaenr


chaitanyaenr (approved):
/lgtm

PR #41590, opened 2023-07-24
Add chaos cluster profile

chaitanyaenr:
This commit adds a chaos profile to use for the test runs that will need access to custom AWS account.

PR in ci-tools: https://github.com/openshift/ci-tools/pull/3549. Thanks.


chaitanyaenr:
@danilo-gemoli @hongkailiu need your review when you get time please.

PR #41595, opened 2023-07-24
AWS Shared VPC: Count Availability Zones

patrickdillon:
Check the number of availability zones. This prevents an error when trying to use 3 availability zones but the region only has two.

Particularly, us-central-1 has 2 zones and we see failures like this regularly: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-nightly-4.14-e2e-aws-ovn-shared-vpc-phz-techpreview/1682697371017809920

Should fix: https://issues.redhat.com/browse/OCPBUGS-16700

### DIFF
--- a/ci-operator/step-registry/aws/provision/vpc/shared/aws-provision-vpc-shared-commands.sh
+++ b/ci-operator/step-registry/aws/provision/vpc/shared/aws-provision-vpc-shared-commands.sh
@@ -378,7 +378,10 @@ Outputs:
       ]
 EOF
 
-ZONES_COUNT=$(aws --region "${REGION}" ec2 describe-availability-zones --filter Name=state,Values=available Name=zone-type,Values=availability-zone | jq '.AvailabilityZones | length')
+MAX_ZONES_COUNT=$(aws --region "${REGION}" ec2 describe-availability-zones --filter Name=state,Values=available Name=zone-type,Values=availability-zone | jq '.AvailabilityZones | length')
+if (( ZONES_COUNT > MAX_ZONES_COUNT )); then
+  ZONES_COUNT=$MAX_ZONES_COUNT
+fi
 
 # The above cloudformation template's max zones account is 3
 if [[ "${ZONES_COUNT}" -gt 3 ]]


patrickdillon:
/hold
for review from @yunjiang29 @jianlinliu

ci-operator/step-registry/aws/provision/vpc/shared/aws-provision-vpc-shared-commands.sh
```diff
@@ -378,6 +378,8 @@ Outputs:
       ]
 EOF
 
+ZONES_COUNT=$(aws --region "${REGION}" ec2 describe-availability-zones --filter Name=state,Values=available Name=zone-type,Values=availability-zone | jq '.AvailabilityZones | length')
```
jianlinliu, line 381: The step has a `ZONES_COUNT` env var which is default to "3", so once we have this line, the env would be ignored.

@yunjiang29 do you have some opinions on it?

yunjiang29, line 381: to preserve `ZONES_COUNT`, I would like suggest:
```suggestion
MAX_ZONES_COUNT=$(aws --region "${REGION}" ec2 describe-availability-zones --filter Name=state,Values=available Name=zone-type,Values=availability-zone | jq '.AvailabilityZones | length')
if (( ZONES_COUNT > MAX_ZONES_COUNT )); then
  ZONES_COUNT=$MAX_ZONES_COUNT
fi
```

yunjiang29: to preserve `ZONES_COUNT`, I would like suggest:
```suggestion
MAX_ZONES_COUNT=$(aws --region "${REGION}" ec2 describe-availability-zones --filter Name=state,Values=available Name=zone-type,Values=availability-zone | jq '.AvailabilityZones | length')
if (( ZONES_COUNT > MAX_ZONES_COUNT )); then
  ZONES_COUNT=$MAX_ZONES_COUNT
fi
```

patrickdillon:
ROSA rehearsal failures appear to be due to quota issues: 'Resource handler returned message: "The maximum number of addresses has been reached. (Service: Ec2, Status Code: 400, Request ID: 4befbdbc-3045-467a-a0db-17204128d156)" (RequestToken: 472e7de8-f917-26f9-e9df-c972e76c9610, HandlerErrorCode: GeneralServiceException)'

PR #41599, opened 2023-07-24
Update approvers list

svetsa-rh:
@paigerube14 Can you ptal?


ci-operator/step-registry/openshift-qe/installer/OWNERS
```diff
@@ -3,6 +3,7 @@ approvers:
 - paigerube14
```
paigerube14, line 3: If this folder OWNERS file is the same as the file in the openshift-qe main folder think we should have a symbolic link here so we dont need to update every sub folder every time we have an edit to the owners

Example: https://github.com/openshift/release/blob/master/ci-operator/step-registry/openshift-qe/cluster-density-v2/OWNERS

svetsa-rh: @paigerube14 Updated, thanks!

ci-operator/jobs/openshift-qe/ocp-qe-perfscale-ci/OWNERS
```diff
@@ -9,6 +9,7 @@ approvers:
 - mffiedler
 - paigerube14
 - qiliredhat
+- svetsa-rh
```
paigerube14, line 12: this file should probably also use the symbolic link

svetsa-rh: Updated.

PR #41601, opened 2023-07-24
Gcp configurable disk size

cgwalters:
gcp: Add configurablility for disk size

To aid in reproducing https://bugzilla.redhat.com/show_bug.cgi?id=2224081

---

ipi/gcp: Expand non-default disks

To provoke https://bugzilla.redhat.com/show_bug.cgi?id=2224081
better.

---

### DIFF
--- a/ci-operator/config/openshift/machine-config-operator/openshift-machine-config-operator-master.yaml
+++ b/ci-operator/config/openshift/machine-config-operator/openshift-machine-config-operator-master.yaml
@@ -205,6 +205,13 @@ tests:
           cpu: 100m
       timeout: 2h30m0s
     workflow: ipi-gcp
+- as: ipi-gcp-disk-large
+  skip_if_only_changed: ^docs/|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|OWNERS_ALIASES|LICENSE)$
+  steps:
+    cluster_profile: gcp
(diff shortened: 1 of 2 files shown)


cgwalters:
Anyone know if with rehearsals I can pass non-default options?  It doesn't seem so...but hopefully this is enough to at least expose it to cluster-bot?  Ah, I may need another PR to allowlist this in https://github.com/openshift/ci-chat-bot/blob/7b811646cdd4b7c47bef6531fa630fbcb7c9d8e4/pkg/manager/prow.go#L61C1-L61C1 ?

cgwalters:
OK I took a stab at adding another optional presubmit to the MCO which forces on large disks, which I can explicitly test in rehearsals.

Incidentally when I looked at testing manually with a 4TiB disk in GCP I paused when the cost estimator showed over $900/month. (From [disk image pricing](https://cloud.google.com/compute/disks-image-pricing - $0.204 per GB in us-west-2 as of 20230725).  But...that's prorated per hour.  So that works out to just $1.13 per hour if I did my math right.  And we're unlikely to use the nodes for more than an hour.  So yeah, that's Quite Cheap.

cgwalters:
/pj-rehearse pull-ci-openshift-machine-config-operator-master-ipi-gcp-disk-large

PR #41618, opened 2023-07-25
Use Go 1.19 on main

nak3:
This is just same with https://github.com/openshift/release/pull/34868
Go 1.19 is used since SO 1.29.

/cc @ReToCode @pierDipi


ReToCode (approved):
/lgtm

PR #41620, opened 2023-07-25
tag vpc for ocm

yasun1:
For the managed clusters provisioned from the ocm side, the additional tags on the subnets are required in the preflight check before installation. For the OCP step `aws-provision-tags-for-byo-vpc`, it introduces the tag with 'infra_id' which is required for the OCP cluster, however, we only get the infra_id after we create the cluster. So an individual step only for the managed cluster.

[OCPQE-17770](https://issues.redhat.com//browse/OCPQE-17770)


jianlinliu:
> For the OCP step aws-provision-tags-for-byo-vpc, it introduces the tag with 'infra_id' which is required for the OCP cluster, however, on the ocm side, we only get the infra_id after we create the cluster.

If I remember right, OCP is the same case,  only get the infra_id after we create the cluster.

yasun1:
@jianlinliu On the ocm side, it does the preflight check for tags before the cluster install, so we need to put the step before creation. This conflicts with the OCP process(do cluster installation -> tag vpc) .

yunjiang29:
I think three options for this requirement:
1. if post-action is acceptable, apply tag after cluster creation: get VPC after OCM cluster created, and apply tag to it.
2. if post-action is not acceptable,  add a switch to `aws-provision-tags-for-byo-vpc` to indicate if add infra_id tag or not
3. if post-action is not acceptable,  use different process for OCP and OCM, which is what this PR does.

Personally, I prefer 1, as the same behaviour applies to clusters (OCP and OCM).
2 & 3 are the easiest way but a little hard to maintain.

@yasun1  @lihongan can you confirm if 1 is acceptable?

yunjiang29:
> it does the preflight check for tags before the cluster install

So 1 is not acceptable? what about 2?

lihongan:
Please correct me if I'm wrong, I think for OCM the requirement is adding `elb` tags for subnets before installing the cluster, no need to add tag for vpc, right?

yasun1:
Certainly, option 2 is acceptable.  In the previous, I do not want to maintain too many switches in the script, but now seems like I can use one switch 'MANAGED_CLUSTER' to control the features for the ocm clusters.

lihongan:
The tag "kubernetes.io/cluster/${CLUSTER_ID}" is required by AWS Load Balancer Operator (ALBO) testing, it can be added to VPC and subnets in post-action.

yasun1:
Not sure whether need to separate the step into two, one is pre, another is post? It may be based on the requirement in the future.

ci-operator/step-registry/aws/provision/tags-for-byo-vpc/aws-provision-tags-for-byo-vpc-commands.sh
```diff
@@ -13,24 +13,32 @@ fi
...
+  if [[ -z $infra_id ]]; then
+    echo "Error: infra_id is empty"
+    exit 1
+  fi
+
+  echo "Adding tags for VPC: $vpc_id, tags: kubernetes.io/cluster/${infra_id}, value: shared."
+  aws --region $REGION ec2 create-tags --resources $vpc_id --tags Key=kubernetes.io/cluster/${infra_id},Value=shared
+fi
 
+# Add tags on subnets
```
yunjiang29, line 37:
```suggestion
# Add tags on private subnets
```

ci-operator/step-registry/aws/provision/tags-for-byo-vpc/aws-provision-tags-for-byo-vpc-commands.sh
```diff
@@ -13,24 +13,32 @@ fi
...
 echo "Adding tags for private subnets:$private_subnet_ids, tags: kubernetes.io/role/internal-elb, value is empty."
 aws --region $REGION ec2 create-tags --resources $private_subnet_ids --tags Key=kubernetes.io/role/internal-elb,Value=
 
-if [[ ${APPLY_TAGS_TO_PUBLIC_SUBNETS} == "yes" ]]; then
+if [[ ${MANAGED_CLUSTER} == "yes" ]]; then
   echo "Adding tags for public subnets:$public_subnet_ids, tags: kubernetes.io/role/elb, value is empty."
   aws --region $REGION ec2 create-tags --resources $public_subnet_ids --tags Key=kubernetes.io/role/elb,Value=
 fi
```
yunjiang29, line 44: How about combine [L25-35](https://github.com/openshift/release/pull/41620/files#diff-d2363855530a9951bd709a70c003a445c28e71b7cf050ba815b21ee645288b77R25-R35) together:
```
if [[ ${MANAGED_CLUSTER} == "yes" ]]; then
  # Adding tag for public subnets
else
  # Adding infra_id tag for VPC
fi
```

yunjiang29 (commented):
@yasun1 Added two comment, others look good to me

ci-operator/step-registry/aws/provision/tags-for-byo-vpc/aws-provision-tags-for-byo-vpc-commands.sh
yasun1: Below there is a step to add the public subnets, so `Add tags on subnets` should be more reasonable.

ci-operator/step-registry/aws/provision/tags-for-byo-vpc/aws-provision-tags-for-byo-vpc-commands.sh
yasun1: The reason why I do not combine these together is that they are two objects, one is vpc, another is subnet, there is no causation between the two. To be more scalability and readable, I'd like to put them into different blocks.

yunjiang29: Get it

ci-operator/step-registry/aws/provision/tags-for-byo-vpc/pre/aws-provision-tags-for-byo-vpc-pre-commands.sh
```diff
@@ -0,0 +1,30 @@
...
+  export AWS_SHARED_CREDENTIALS_FILE="${CLUSTER_PROFILE_DIR}/.awscred"
+fi
+
+REGION="${LEASED_RESOURCE}"
+
+vpc_id=$(head -n 1 ${SHARED_DIR}/vpc_id)
+private_subnet_ids=$(yq-go r -j ${SHARED_DIR}/private_subnet_ids | jq -r '[ . | join(" ") ] | @csv' | sed "s/\"//g")
+public_subnet_ids=$(yq-go r -j ${SHARED_DIR}/public_subnet_ids | jq -r '[ . | join(" ") ] | @csv' | sed "s/\"//g")
+
+if [[ -z $vpc_id ]] || [[ -z $private_subnet_ids ]]; then
```
yunjiang29, line 20:
```suggestion
if [[ -z $private_subnet_ids ]] || [[ -z $public_subnet_ids ]]; then
```

ci-operator/step-registry/aws/provision/tags-for-byo-vpc/pre/aws-provision-tags-for-byo-vpc-pre-commands.sh
```diff
@@ -0,0 +1,30 @@
...
+  echo "Error: Can not get VPC id or private subnets, exit"
```
yunjiang29, line 21:
```suggestion
  echo "Error: Can not get subnet ids, exit"
```

ci-operator/step-registry/aws/provision/tags-for-byo-vpc/pre/aws-provision-tags-for-byo-vpc-pre-commands.sh
```diff
@@ -0,0 +1,30 @@
...
+  echo "vpc: $vpc_id, private_subnet_ids: $private_subnet_ids"
```
yunjiang29, line 22:
```suggestion
  echo "private_subnet_ids: $private_subnet_ids, public_subnet_ids: $public_subnet_ids"
```

ci-operator/step-registry/aws/provision/tags-for-byo-vpc/pre/aws-provision-tags-for-byo-vpc-pre-commands.sh
```diff
@@ -0,0 +1,30 @@
...
+if [[ ${ENABLE_SHARED_VPC} == "yes" ]]; then
+  echo "Using shared account"
+  export AWS_SHARED_CREDENTIALS_FILE="${CLUSTER_PROFILE_DIR}/.awscred_shared_account"
+else
+  export AWS_SHARED_CREDENTIALS_FILE="${CLUSTER_PROFILE_DIR}/.awscred"
+fi
+
+REGION="${LEASED_RESOURCE}"
+
+vpc_id=$(head -n 1 ${SHARED_DIR}/vpc_id)
```
yunjiang29, line 16: `vpc_id` is not involved in this step

ci-operator/step-registry/aws/provision/tags-for-byo-vpc/pre/aws-provision-tags-for-byo-vpc-pre-ref.yaml
```diff
@@ -0,0 +1,18 @@
+ref:
+  as: aws-provision-tags-for-byo-vpc-pre
```
yunjiang29, line 2: As this step is only for pre-flight check, I suggest to add a `ocm`/`managed-cluster` or something like that to the step name, e.g. `aws-provision-tags-for-byo-vpc-for-mc` or `aws-provision-tags-for-byo-vpc-for-mc-precheck`

yunjiang29 (commented):
@yasun1

ci-operator/step-registry/aws/provision/tags-for-byo-vpc/pre/aws-provision-tags-for-byo-vpc-pre-ref.yaml
yasun1: The step is not only for ocm, but now it is only used by ocm. Need us to limit it only for ocm?

yasun1:
@yunjiang29 Please help review again.

yunjiang29:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.12-amd64-nightly-aws-rosa-sts-hypershift-guest-p1-f7

yasun1:
@yunjiang29 Let us merge the PR first so that it will not be a blocker for the PR https://github.com/openshift/release/pull/41664

yunjiang29:
Merging this PR first and the following error will be fixed in https://github.com/openshift/release/pull/41664 
```
INFO[2023-07-27T02:24:23Z] Step aws-rosa-sts-hypershift-guest-p1-f7-aws-provision-tags-for-byo-vpc-ocm-pre succeeded after 14s. 
...
INFO[2023-07-27T02:52:27Z] Logs for container test in pod aws-rosa-sts-hypershift-guest-p1-f7-aws-provision-tags-for-byo-vpc: 
INFO[2023-07-27T02:52:27Z] jq: error: Could not open file /tmp/secret/metadata.json: No such file or directory
```

PR #41628, opened 2023-07-25
ODF: remove special prow config for ceph-csi repository

nixpanic:
The standard ODF (red-hat-storage) Prow configuration should be sufficient. There have been two Tide queries active, which caused PRs to be merged unintentionally. Using the default configuration will make the behavior easier to understand and is less prone to making mistakes.

/assign agarwal-mudit


Rakshith-R (changes requested):
OWNERS file seems to be have deleted too. 
I think we leave it as it is.
Other changes lgtm.

nixpanic:
> OWNERS file seems to be have deleted too. I think we leave it as it is. Other changes lgtm.

That is intentional, the whole directory is removed. There is an `OWNERS` file in the parent directory, which manages the standard configuration for the red-hat-storage organization.

agarwal-mudit (approved):
/lgtm

Rakshith-R (approved):

PR #41671, opened 2023-07-26
MGMT-15343: adding dependabot to trusted apps

eifrach:


core-services/prow/02_config/openshift-assisted/assisted-installer-deployment/_pluginconfig.yaml
```diff
@@ -1,3 +1,8 @@
+triggers:
```
osherdp, line 1: let's do this for all required repos, shall we?
* let's move this to ``core-services/prow/02_config/openshift-assisted/_pluginconfig.yaml``. this way it will apply for all the repositories in the ``openshift-assisted`` organization
* assisted-image-service misses this configuration as well

osherdp (changes requested):

core-services/prow/02_config/openshift-assisted/assisted-installer-deployment/_pluginconfig.yaml
osherdp, line 1: you can delete it now. it's enabled for the whole org

core-services/prow/02_config/openshift-assisted/_pluginconfig.yaml
```diff
@@ -61,3 +61,8 @@ plugins:
     - jira
     - size
     - welcome
+triggers:
+- repos:
+  - openshift-assisted
+  trusted_apps:
+  - dependabot
```
osherdp, line 68: tip: change IDE/vim settings to add blank lines at end of files
it will save you a lot of time with repos that we usually use as most of them rely on it

osherdp (approved):

core-services/prow/02_config/openshift-assisted/_pluginconfig.yaml
eifrach: I'm deleting it manually - usaully linter's complaining about empty line not requiring one 
i'm doing it on auto drive mode :)

PR #41674, opened 2023-07-26
stackrox: increase timeout for FIPS-related tests

BradLugo:
We saw an issue with the release-4.1 OSCI FIPS jobs where it would hit the 5h timeout.

/hold
/uncc
/assign @BradLugo


jschnath (commented):
Hi Brad I don't think we actually need to increase the timeout, at least not for release 4.1.2. There was a different problem there that has now been resolved. If the timeouts should be increased regardless I won't block you, but don't do it if it's just for the release 4.1.2.

PR #41710, opened 2023-07-27
Add gcp cert-rotation profiles

wangke19:
We don't have gcp cert-rotation on  amd64-nightly-4.13-upgrade-from-stable-4.12, and for upgrade path  amd64-nightly-4.13-upgrade-from-stable-4.13, we have one with workload-identity option which always ran into failure, cert-rotation is required to follow steps in bug https://issues.redhat.com/browse/OCPBUGS-15749  manually, so removed existing profile and added new one.


ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.13__amd64-nightly-4.13-upgrade-from-stable-4.13.yaml
```diff
@@ -1,44 +1,3 @@
-base_images:
```
liangxia, line 1: Are those removal indented ?

PR #41720, opened 2023-07-27
Fix maistra operator integration test

jwendell:
- Fix docker login command
- Use `&&` between shell commands to stop on first error


mattmahoneyrh:
If docker login is not needed, why not just remove the login?

PR #41737, opened 2023-07-27
Remove the cap on COMPUTE_NODE_COUNT

krishvoor:


svetsa-rh:
@jianlinliu @yunjiang29 @technical-release-team-approvers PerfScale team needs to test various e2e workloads on AWS clusters for different cluster sizes to check for performance and scalability issues. 
Not sure why the max is being limited to 5 nodes. Could you ptal? 
Can we increase this to a larger number if we cannot remove it completely? We sometimes scale clusters up to 250 nodes.

svetsa-rh:
@krishvoor Are you able to scale nodes using WORKER_REPLICA_COUNT instead of COMPUTE_NODE_REPLICAS? If so, we can close this PR instead of updating defaults for creation time.

krishvoor:
> @krishvoor Are you able to scale nodes using WORKER_REPLICA_COUNT instead of COMPUTE_NODE_REPLICAS? If so, we can close this PR instead of updating defaults for creation time.

I have not yet, but can sure try this.

> Are you able to scale nodes using

I am thinking, this will work well for a use-case in the Day-2 job and will still be a problem during Day-1  scenario?

svetsa-rh:
> I am thinking, this will work well for a use-case in the Day-2 job and will still be a problem during Day-1 scenario?

That is true. For purpose of our testing, do we care for self managed clusters? Also, for self managed clusters, will it have significant impact on performance as seen by customers?

krishvoor:
> > I am thinking, this will work well for a use-case in the Day-2 job and will still be a problem during Day-1 scenario?
> 
> That is true. For purpose of our testing, do we care for self managed clusters? Also, for self managed clusters, will it have significant impact on performance as seen by customers?

Agreed for our case might not be impactful.

But thinking from a large cluster testing (say 120 nodes).
It might have an impact on overall cluster-install timings.

jianlinliu:
> Not sure why the max is being limited to 5 nodes.

Searched git commits, found this - https://github.com/openshift/release/pull/34716#discussion_r1054079591.

Personally I think the PR change make scenes. 

/lgtm

PR #41740, opened 2023-07-27
Remove 4.12 rosa testing in Prod.

svetsa-rh:
Originally added this file for us to be able to perform reliability testing for managed services to workaround longevity issues. For multiple reasons, that testing is now being done on Jenkins and there is no other reason for us to add/test in Prod. 
Deleting the file.


svetsa-rh:
@paigerube14 @jtaleric Can you ptal?

svetsa-rh:
`make jobs` in progress

svetsa-rh:
@qiliRedHat Can you ptal?

PR #41783, opened 2023-07-28
Run branch-protector every 3h

kaovilai:
Give us more iterations per day to update/test/fix branch protection rules.


smg247:
We don't want to do this. It uses a lot of tokens. I triggered a run ad-hoc.
/close

PR #41795, opened 2023-07-31
NO-ISSUE: Assisted-Installer: Make ODF and CNV periodics run on 4.13 due to lso failure

eliorerz:
/cc @osherdp @adriengentil


adriengentil:
why LSO fails in this case? It's a similar issue to https://github.com/openshift/assisted-service/pull/5323/files ?

osherdp:
> why LSO fails in this case? It's a similar issue to https://github.com/openshift/assisted-service/pull/5323/files ?

yes it's the same issue, but we don't have this kind of hack in the assisted-service code

osherdp (approved):

adriengentil (approved):

PR #41827, opened 2023-07-31
Add port 1337 to tcpdump filter

trozet:
This will allow us to debug pdb load balancer disruption tests: https://github.com/openshift/origin/pull/28112


tssurya (approved):
/lgtm

PR #41830, opened 2023-07-31
[openshift-loki]: Fix operator pullspec substitutions

periklis:
Based on the way we tag upstream by the following PR, we switched to tags without the `v` prefix:
- grafana/loki/pull/10050

Required by:
- openshift/loki/pull/171


JoaoBraveCoding (approved):
/lgtm

PR #41840, opened 2023-08-01
HCO: Add images for the new version, 1.11.0

nunnatsa:


clusters/app.ci/supplemental-ci-images/hyperconverged-cluster-bundle.yaml
```diff
@@ -17,3 +17,9 @@ spec:
       name: quay.io/kubevirt/hyperconverged-cluster-bundle:1.10.0-unstable
     importPolicy:
       scheduled: false
+  - name: 1.11.0-unstable
```
droslean, line 20: You can remove this, since this is added in the mirroring configuration.

nunnatsa: Thanks @droslean. 

This is copied from the previous versions in the same file. This is working for us. Is it a blocker?

droslean: All of the tags are not needed. It's not a blocker for now, but in the effort of maintaining the images for multi-arch, this can affect it.

Let's merge this. I will remove all of them once we are ready to support multi-architecture in our clusters.

/lgtm
/approve

nunnatsa: Makes sense. Thanks

PR #41843, opened 2023-08-01
Refine keyword position for disabled jobs

liangxia:
Move the keyword `disabled` from test name prefix to test name suffix, so that the jobs sort in a better order.


pruan-rht:
@liangxia this could possibly break the previously agreed naming convention of having the format of `p<x>-f<y>`.  Just something to consider.

liangxia:
> @liangxia this could possibly break the previously agreed naming convention of having the format of `p<x>-f<y>`. Just something to consider.

p<x> is no longer used/needed, while f<y> is still there. We are following the rules.

liangxia:
Rebase the codes so that the required tests can pass, no code changes.

PR #41854, opened 2023-08-01
post-installation check for Single-Node OpenShift cluster

jianli-wei:
- add post-installation step `cucushift-installer-check-sno` to ensure it is a single-node cluster


jianli-wei:
/pj-rehearse periodic-ci-openshift-verification-tests-master-installer-rehearse-4.14-installer-rehearse-gcp

PR #41860, opened 2023-08-01
ROX-17337: enable 4.13 cluster pool for PRs

tommartensen:
Enabled claiming clusters from stackrox' 4.13 cluster pool for all presubmit jobs to master.

The pool size was adapted in #41853 .

Also we seemed to have forgotten one job in the nightlies, and had a pool test job in master, that is cleaned up.


gavin-stackrox (approved):
/lgtm
/honk

PR #41863, opened 2023-08-01
Update skip list for neutron/ovn operators

karelyatin:
Update the filter list in order to avoid running
kuttl and functional jobs when not required.


karelyatin:
@Sandeepyadav93 @slawqo please have a look

viroel:
looks good, this is probably a good solution for all other operators too

PR #41889, opened 2023-08-02
make okd-e2e-gcp-ovn optional job always_run:false

jluhrsen:
this job pass current pass rate is under 50% and not required. It can still be run with the following
comment:

    /test okd-e2e-gcp-ovn


LorbusChris (approved):

LorbusChris:
This is now an AWS test. Does the problem persist?

PR #41908, opened 2023-08-02
Removes tests for Ceph storage plugin from Openshift console

bipuladh:
We have migrated away from console monorepo and have not maintained our Ceph E2E tests.


jhadvig (approved):
Approving since ceph static plugin will get removed as well from console codebase

/lgtm
/approve

PR #41923, opened 2023-08-02
albo, extdns: ignore review approval, disallow self approval

alebedev87:
This PR disallows the PR self approval and ignores the review state for the following repositories: `aws-load-balancer-operator`, `aws-load-balancer-controller`, `external-dns-operator`, `external-dns`.


alebedev87:
I LGTMed a wrong PR.

PR #41955, opened 2023-08-03
ROX-17337/fix: enable auto updater for image set in stackrox OCP 4.13 cluster pool

tommartensen:
`version_lower` and `version_upper` fields allow the auto updater bot to update the image set reference : https://docs.ci.openshift.org/docs/how-tos/cluster-claim/#create-a-manifest-for-your-cluster-pool

Example from other cluster pools:
https://github.com/openshift/release/pull/41748/commits/91e6f20a8415e2827c48a2cebbe0aa0536c79dfd#diff-da119b0311c8065ed07405dd7006ba48d92ae6309db583a30b9e18fbb7865e62R21

Manually updated the image set ref to 4.13.7 to bring level with OSCI default cluster pool.

Removed the `channel: fast` option, that is undocumented and did not have any effect.


gavin-stackrox (approved):
/lgtm

PR #42006, opened 2023-08-04
send slack message for the job rosa-aws-sts-rosacli-e2e-f3

yuwang-RH:
send slack message for the job rosa-aws-sts-rosacli-e2e-f3


yuwang-RH:
@kuiwang02 @liangxia Could you help review and approve this PR?Thanks.

PR #42009, opened 2023-08-04
`arm01`: ci-operator build framework

danilo-gemoli:
This PR creates what is necessary to build an `arm64` version of the `ci-operator`.
In order to stay as close as possible to the `amd64` building process I have tried to mimic what `ci-operator` already does to build itself, that translates in what follows:
- Make the build root `ci-tools-build-root-arm64:1.19` that is the `arm64` version of [the one](https://github.com/openshift/release/blob/64fe83bb0b9d875ab3991092059b073666b80a87/clusters/app.ci/supplemental-ci-images/ci-tools-build-root.yaml#L305-L403) we already use 
- Starting from the previous build root, spin up a new build `ci-tools-bin-arm64` that clones the latest version of [o/ci-tools](https://github.com/openshift/ci-tools) and then issue a `make production-install`. This is the same as [here](https://github.com/openshift/release/blob/0116a6300414a81d9b03a6c1c26798e802c43db9/ci-operator/config/openshift/ci-tools/openshift-ci-tools-master.yaml#L18)
- Finally run the build `ci-operator-arm64` that grabs the `ci-operator` binary built in `ci-tools-bin-arm64` and then constructs the image `ci/ci-operator-arm64` using this [Dockerfile](https://github.com/openshift/ci-tools/blob/2a2449aeec11633d87e84e2a5ccb88f200c681bf/images/ci-operator/Dockerfile)

Jira card: [DPTP-3589](https://issues.redhat.com/browse/DPTP-3589)

/cc @deepsm007  @droslean

### DIFF
--- a/clusters/build-clusters/arm01/ci-tools/100_bc_ci-tools-build-root-1.19-arm64.yaml
+++ b/clusters/build-clusters/arm01/ci-tools/100_bc_ci-tools-build-root-1.19-arm64.yaml
@@ -1,0 +1,102 @@
+kind: BuildConfig
+apiVersion: build.openshift.io/v1
+metadata:
+  namespace: ci
+  name: ci-tools-build-root-1.19-arm64
+spec:
+  successfulBuildsHistoryLimit: 1
+  failedBuildsHistoryLimit: 2
+  output:
+    to:
+      kind: ImageStreamTag
+      name: ci-tools-build-root-arm64:1.19
+  source:
+    images:
+    # FIXME: no arm64 version of boskos so far
+    - from:
+        kind: ImageStreamTag
+        name: boskos:latest
+        namespace: ci
+      paths:
+      - destinationDir: .
+        sourcePath: /app
+    - from:
+        kind: ImageStreamTag
+        name: prometheus:latest
+        namespace: ci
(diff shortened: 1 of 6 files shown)


clusters/build-clusters/arm01/ci-tools/100_bc_ci-tools-build-root-1.19-arm64.yaml
```diff
@@ -0,0 +1,102 @@
...
+  output:
+    to:
+      kind: ImageStreamTag
+      name: ci-tools-build-root-arm64:1.19
+  source:
+    images:
+    # FIXME: no arm64 version of boskos so far 
+    - from:
+        kind: ImageStreamTag
+        name: boskos:latest
```
droslean, line 18: There should be an arm64 version of boskos.

droslean: @danilo-gemoli This is happening because of:
```
    importPolicy:
      importMode: Legacy

```

in the boskos imagestream.

droslean (approved):

PR #42051, opened 2023-08-07
openstack-ansiblee-operator: Do not run build-deploy job by default

kajinamit:
We don't need to run build-deploy job by default because we run the kuttl job which contains the build-deploy steps.


kajinamit:
In https://github.com/openstack-k8s-operators/openstack-ansibleee-operator/pull/206 I noticed both build-deploy job and build-deploy-kuttl job are executed.

rebtoor (approved):

PR #42052, opened 2023-08-07
`arm01`: Postsubmit that triggers a ci-operator-arm64 BuildConfig

danilo-gemoli:
A postsubmit that triggers the BuildConfig `ci-operator-arm64` every time something gets merged on [o/ci-tools](https://github.com/openshift/ci-tools).
The arguments `--follow` and `--wait` make sure that the ProwJob streams the build log in realtime and waits until it (the build) either fails or succeeds.
The `config-updater` SA has rights to trigger any BuildConfig from the `ci` NS.

Jira Card: [DPTP-3591](https://issues.redhat.com/browse/DPTP-3591)

/cc @droslean @deepsm007

### DIFF
--- a/clusters/build-clusters/arm01/admin_ci-images-builder_rbac.yaml
+++ b/clusters/build-clusters/arm01/admin_ci-images-builder_rbac.yaml
@@ -1,0 +1,50 @@
+---
+kind: ServiceAccount
+apiVersion: v1
+metadata:
+  annotations:
+    openshift.io/description: This SA is used to start ci builds
+  name: ci-images-builder
+  namespace: ci
+---
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: ci-images-builder
+  namespace: ci
+rules:
+- apiGroups:
+  - build.openshift.io
+  resources:
+  - buildconfigs
+  - builds
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - build.openshift.io
+  resources:
+  - buildconfigs/instantiate
+  verbs:
+  - create
+- apiGroups:
+  - build.openshift.io
+  resources:
+  - builds/log
+  verbs:
+  - get
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: ci-images-builder
+  namespace: ci
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: ci-images-builder
+subjects:
+- kind: ServiceAccount
+  name: ci-images-builder
+  namespace: ci


ci-operator/jobs/openshift/ci-tools/openshift-ci-tools-master-postsubmits.yaml
```diff
@@ -4,7 +4,44 @@ postsubmits:
...
+          name: build-farm-credentials
+          readOnly: true
+      serviceAccountName: config-updater
+      volumes:
+      - name: build-farm-credentials
+        secret:
+          items:
+          - key: sa.config-updater.arm01.config
+            path: kubeconfig
+          secretName: config-updater
```
droslean, line 39: Are you sure that this is the correct credentials to trigger the build? I am not sure if the `config-updater` has permission on buildConfigs.

danilo-gemoli: Yeah it's the account I'm using to do some tests before opening the PRs as mine doesn't have any power there. Give it a shoot eventually, so to have a double check.

droslean: You need to use the `origin-release-images-builder` service account.

danilo-gemoli: I agree, but shouldn't then we
- create `origin-release-images-builder` on `arm01` as well as it exists on `build01` only (check [here](https://github.com/openshift/release/blob/dc969377f33de08e69553cee297c0575b592daa7/clusters/build-clusters/01_cluster/ci/admin_origin-release-images-builder_rbac.yaml))?
- and then force the postsubmit of this PR to always run on `arm01` with the SA set to `origin-release-images-builder` as you're suggesting?

ci-operator/jobs/openshift/ci-tools/openshift-ci-tools-master-postsubmits.yaml
```diff
@@ -1,5 +1,31 @@
...
+    spec:
+      containers:
+      - args:
+        - --namespace=ci
+        - start-build
+        - --follow
+        - buildconfig/ci-operator-arm64
+        command:
+        - oc
+        image: image-registry.openshift-image-registry.svc:5000/ocp-arm64/4.13:cli
```
droslean, line 22: How can we be sure that this job will run on an arm node?

danilo-gemoli:
```yaml
cluster: arm01
...
labels:
  ci-operator.openshift.io/cluster: arm01
```
will ensure Prow is going to spawn the pod on the arm cluster.
`image-registry.openshift-image-registry.svc:5000/ocp-arm64/4.13:cli` holds `oc` version `arm64` which will authenticate to the cluster as `ci-images-builder`.

droslean: We have both amd64 and arm64 nodes in the arm01 cluster. We need to make sure that the pod will be spawned in the arm64 node.

droslean:
Let's merge it and deal with my comment in a follow-up

/lgtm
/approve

PR #42071, opened 2023-08-07
OCPBUGS-17437: Validate CSO assets for kustomize changes

gnufied:
Validate CSO assets for any kustomize related changes.

Fixes OCPBUGS-17437

### DIFF
--- a/ci-operator/config/openshift/cluster-storage-operator/openshift-cluster-storage-operator-master.yaml
+++ b/ci-operator/config/openshift/cluster-storage-operator/openshift-cluster-storage-operator-master.yaml
@@ -43,6 +43,10 @@ tests:
   commands: make verify
   container:
     from: src
+- as: verify-kustomize-assets
+  commands: ./hack/verify-manifest.sh
+  container:
+    from: src
 - as: e2e-aws-ovn
   steps:
     cluster_profile: aws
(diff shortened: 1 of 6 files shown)


RomanBednar:
I've noticed a couple of things that might be worth addressing (in CSO, before merging this PR ideally):

- `verify-manifests.sh` should not be calling `generate-manifests.sh`
	- it does not make much sense to run `generate-manifests.sh` in CI (the change it does has to be commited to pass the verification)
	- there's a hint when it fails to run `make update` that runs `generate-manifests.sh` so `verify-manifests.sh` should not run it itself like it does now
- `verify-manifests.sh` should be checking only directories that `generate-manifests.sh` acts upon
	- if anything outside `assets/csidriveroperators/aws-ebs/` dir is modified `verify-manifests.sh` will fail because it checks entire `assets/` dir and `make update`  (as hinted) won't help because it runs `oc kustomize` only for `assets/csidriveroperators/aws-ebs` subdir

gnufied:
> it does not make much sense to run generate-manifests.sh in CI (the change it does has to be commited to pass the verification)

We are doing that to ensure that user has not forgotten to run it in their local environment. The idea is not to pass verification, but fail the verification if there are changes because of rerunning ./generate-manifesh.sh 

> if anything outside assets/csidriveroperators/aws-ebs/ dir is modified verify-manifests.sh will fail because it checks entire assets/ dir and make update (as hinted) won't help because it runs oc kustomize only for assets/csidriveroperators/aws-ebs subdir

verify-manifest.sh is for entire `assets` directory. We are going to move other drivers that run on hypershift to have same structure. Also, regardless, why will there be any uncomitted changes in `/assets` ? If there are, that seems like a red flag too.

PR #42087, opened 2023-08-08
infra-operator: Skip prow jobs when updating zuul job definitions

kajinamit:
We don't have to run prow jobs when we update only definitions of zuul jobs.


stuggi (approved):
/lgtm

PR #42091, opened 2023-08-08
oko: Fix wrong test directory name

kajinamit:
Some operators have the 'tests' directory instead of the 'test' directory. This fixes the pattern according to the correct name.


Sandeepyadav93:
/lgtm

Cross checked all these operators have tests directory.

PR #42093, opened 2023-08-08
MGMT-14710: Change the regex for ci operator file names

danmanor:
Change the regex for ci operator master configs so it will end with `master` or `__master` instead of include `master` so `bug-master-bot` will not match


osherdp (approved):
need an approval from ``#forum-ocp-testplatform`` though

PR #42105, opened 2023-08-08
onboard ART aos-cd-jobs to prow

ashwindasr:
This is a config repository. We do not build images or run prow tests.


joepvd (approved):
/lgtm

PR #42135, opened 2023-08-09
[OCPQE-16286] Remove WINC jobs for 4.10 on AWS.

jfrancoa:
Due to the removal of the Windows Server 2019 image with Containers from AWS EC2, currently there is no way to create a Windows worker on AWS using any release previous than 4.10. Therefore, we are dropping this job from the 4.10 profile.


jfrancoa:
ping @sergiordlr , could you please have a look?

jfrancoa:
ping @liangxia could you please review? we need to remove this jobs as AWS can't run Windows Containers on 4.10 anymore.

PR #42155, opened 2023-08-09
odf: remove the individual config files

agarwal-mudit:
Every repo should follow common set of rules.


nixpanic (changes requested):
Actually, this doesn't `/lgtm`. If the ``OWNERS` file is the only one remaining in the directory, remove the whole directory including the `OWNERS` file. The file is used form maintenance of that particular directory only, and hence is useless.

Don't confuse these `OWNERS` files in the openshift/release` repository with `OWNERS` files in the repository that contains the actual component/code.

iamniting (approved):

PR #42159, opened 2023-08-09
ci-operator: add e2e job for crun-wasm

sohankunkerkar:


ci-operator/jobs/openshift/kubernetes/openshift-kubernetes-master-presubmits.yaml
```diff
@@ -264,6 +264,79 @@ presubmits:
         secret:
           secretName: result-aggregator
     trigger: (?m)^/test( | .* )e2e-aws,?($|\s.*)
+  - agent: kubernetes
+    always_run: true
```
haircommander, line 268: I don't think we need a presubmit to always run. I think a periodic is sufficient then we can run the presubmit if needed

sohankunkerkar: Fixed

mrunalp (approved):

vrutkovs (approved):
/lgtm

PR #42177, opened 2023-08-10
OCPQE-16303: Correct tests executed in destructive jobs on alicloud

liangxia:
Destructive jobs are supposed to use `chain: openshift-e2e-test-qe-destructive` as they execute different test sets.

The wrongly used test chain cause no cucushift test cases executed in destructive jobs on alicloud related profiles.


gpei:
cc @jianli-wei for the related info

PR #42180, opened 2023-08-10
Correct tests executed in destructive job on aws localzone

liangxia:
Correct tests executed in destructive job on profile `aws-ipi-localzone-byo-subnet-sdn`, destructive jobs are supposed to use `chain: openshift-e2e-test-qe-destructive` as they execute different test sets.


yunjiang29:
@liangxia thanks for the correction

PR #42187, opened 2023-08-10
Set auto merge method to squash for ART repos

Ximinhan:
We want to CI bot automatically merge a pr with squash method so commit history in upstream is clean
ref: https://docs.prow.k8s.io/docs/components/core/tide/config/#general-configuration


joepvd:
/hold

Not sure we want this

joepvd:
/close

Let's try the default of merge commits for the coming while.

PR #42190, opened 2023-08-10
baremetal/agent: make post-test gather jobs optional to save time

mkowalski:


mkowalski:
/cc @aleskandro @gpei
(cucushift approvers)

/cc @davidvossel @nirarg @csrwng
(hypershift approvers as you use baremetalds step)

aleskandro:
Hi @mkowalski 

/lgtm

cc @LiangquanLi930 for visibility

ci-operator/step-registry/agent/gather/agent-gather-ref.yaml
```diff
@@ -1,7 +1,10 @@
 ref:
   as: agent-gather
+  best_effort: true
```
andfasano, line 3: I guess that also the workflows in https://github.com/openshift/release/tree/master/ci-operator/step-registry/agent/e2e will require ` allow_best_effort_post_steps: true` in addition to the `best_effort: true`?

mkowalski: It is optional, not really required, that's why I didn't modify all the workflows we have in our universe. Once this merges and steps get this property, what happens is

* workflows that have been not modified do not change their behaviour - steps still run always and still make job fail if they fail
* workflows that have been modified (i.e. some from this PR) - steps will not run when the core test passed; apart from this, failure of the step will not be listed as a reason of job's failure

TLDR: no workflow *requires* a change but it's recommended

andfasano: Ack, for clarity it would be nice to mention that in the [description](https://github.com/openshift/release/pull/42190#issue-1844746096) of the PR for future memory

mkowalski: I do plan to modify this in a subsequent PR just after this merges; I just don't want to `push --force` where because it already got some LGTMs that would be lost

PR #42191, opened 2023-08-10
must-gather: don't include gather-must-gather in failures for the test

mkowalski:
This PR changes the way how gather-must-gather is failing and reportig the failure in the Prow's summary of the test. Similarly to how gather-extra is configured (best effort and optional-on-success) in ci-operator/step-registry/gather/extra/gather-extra-ref.yaml, we want gather-must-gather to also

* not cause the test to fail
* not be listed as a failed test

given it's only post-test step and doesn't provide a real insight into why the scenario failed if any.

### DIFF
--- a/ci-operator/step-registry/gather/audit-logs/gather-audit-logs-ref.yaml
+++ b/ci-operator/step-registry/gather/audit-logs/gather-audit-logs-ref.yaml
@@ -1,8 +1,10 @@
 ref:
   as: gather-audit-logs
+  best_effort: true
   optional_on_success: true
   from: cli
   commands: gather-audit-logs-commands.sh
+  timeout: 15m
   resources:
     requests:
       cpu: 300m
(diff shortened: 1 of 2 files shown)


mkowalski:
/pj-rehearse skip

No need to run the jobs because we are only changing the conditions for step to run&fail

mkowalski:
/pj-rehearse skip

No need to run the jobs because we are only changing the conditions for step to run&fail

vrutkovs:
I don't think `best_effort` is sufficient - you also need to set `allow_best_effort_post_steps` in the workflows too - https://docs.ci.openshift.org/docs/architecture/step-registry/#marking-post-steps-best-effort. Lets add some in this PR - or are you planning to do this in follow-ups?

vrutkovs (approved):
/approve
/lgtm

Workflows will be updated in follow-up PRs

vrutkovs:
/hold

Wait, maybe workflows are required here

PR #42192, opened 2023-08-10
vsphere: make diagnostic collection optional

mkowalski:
This PR makes the vSphere diagnostic collection step optional and removes its ability to fail the whole test scenario. The rationale is that in case when a functional test succeeded but for any reason we failed to collect the vSphere's diagnostic data, we do not want to rerun the whole job.

The PR aims to speed up jobs where vSphere is involved by focusing on the functional test and letting diagnostic and log collection be only additional stuff that does not impact the result.

### DIFF
--- a/ci-operator/step-registry/ipi/deprovision/vsphere/diags/ipi-deprovision-vsphere-diags-ref.yaml
+++ b/ci-operator/step-registry/ipi/deprovision/vsphere/diags/ipi-deprovision-vsphere-diags-ref.yaml
@@ -1,5 +1,8 @@
 ref:
   as: ipi-deprovision-vsphere-diags
+  best_effort: true
+  optional_on_success: true
+  timeout: 15m
   from: upi-installer
   commands: ipi-deprovision-vsphere-diags-commands.sh
   resources:
--- a/ci-operator/step-registry/openshift/e2e/vsphere/ovn/dualstack/openshift-e2e-vsphere-ovn-dualstack-workflow.yaml
+++ b/ci-operator/step-registry/openshift/e2e/vsphere/ovn/dualstack/openshift-e2e-vsphere-ovn-dualstack-workflow.yaml
@@ -1,6 +1,8 @@
 workflow:
   as: openshift-e2e-vsphere-ovn-dualstack
   steps:
+    allow_best_effort_post_steps: true
+    allow_skip_on_success: true
     pre:
     - ref: ipi-install-hosted-loki
     - chain: ipi-conf-vsphere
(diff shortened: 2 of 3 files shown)


mkowalski:
/pj-rehearse skip

No need to run the jobs because we are only changing the conditions for step to run&fail

mkowalski:
/pj-rehearse skip

No need to run the jobs because we are only changing the conditions for step to run&fail

mkowalski:
Let me know if the change in all vSphere workflows is too big. I am happy to keep it contained to only `/e2e/vsphere/ovn/[...]` but I believe it makes sense everywhere

jcpowermac:
@mkowalski these options are new to me can you describe what is changing here?
Or point me to docs?

ci-operator/step-registry/ipi/deprovision/vsphere/diags/ipi-deprovision-vsphere-diags-ref.yaml
```diff
@@ -1,5 +1,8 @@
 ref:
   as: ipi-deprovision-vsphere-diags
+  best_effort: true
```
jcpowermac, line 3: This should not be best effort, this should always run.

mkowalski:
The "best effort" means that you have ability to say "even if this step failed, the CI job can still pass"
The "skip on success" means that you have ability to say "if this step is in post and the main test passed, you can skip"

As you want `ipi-deprovision-vsphere-diags` to always run, even if the test itself passed, I will remove `allow_skip_on_success` from your workflows.

As for `allow_best_effort_post_steps`, keeping it will mean that if your test itself passed but diagnostic collector failed, the Prow job passes nevertheless. Up to you if you want this

Overall the change in `ipi-deprovision-vsphere-diags-ref.yaml` only adds ability for workflows to decide how to behave. Without modifying the workflow the behaviour won't change.

Rationale for the PR is that in OVN/vsphere dual-stack jobs we want to skip anything not related to the test itself (so any kind of must-gather, diagnostic and whatnot) if the openshift conformance succeeded. I have had cases where the main test succeeded but then we failed during log collection and because of this the whole job failed. As we run on limited resources and also often fail to acquire a lease, I'm using all means to remove "false failures"

ci-operator/step-registry/ipi/deprovision/vsphere/diags/ipi-deprovision-vsphere-diags-ref.yaml
mkowalski: Described in a long way in https://github.com/openshift/release/pull/42192#issuecomment-1673930136, but TLDR: it's not changing when the job runs; it adds an opt-in ability for workflow to resign from this job

mkowalski:
Below is a sample of how I want visibility to be configurable

image

You can see that what ultimately caused the failure of this job was `operator conditions kube-apiserver` (it never deployed so the installation failed).

The list contains also e.g. `gather-must-gather` as a test which failed. But when swiping quickly through failed runs, this is a false flag unless investigated deeper. We have `gather-*` failing but this is a symptom of something else failing faster. I could also say that if the cluster never deployed correctly, must-gather has no chance of working properly.

With the changes in this PR (and others, some merged some not yet) I can configure my jobs to not include `gather-must-gather` & co. in the list of failed tests

PR #42194, opened 2023-08-10
USHIFT-1258: Add rhel-eus suffix to main jobs

pacevedom:
Skip metal tests because they run scenario tests with their own ISO.

### DIFF
--- a/ci-operator/config/openshift/microshift/openshift-microshift-main.yaml
+++ b/ci-operator/config/openshift/microshift/openshift-microshift-main.yaml
@@ -164,7 +164,7 @@ tests:
       EC2_AMI: ami-08c7e766098f8cb4f
       EC2_INSTANCE_TYPE: c6g.metal
     workflow: openshift-microshift-e2e-metal-tests
-- as: ocp-conformance-nightly
+- as: ocp-conformance-rhel-eus-nightly
   cluster: build02
   cron: 0 1,13 * * *
   steps:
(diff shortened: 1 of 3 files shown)


dhellmann:
@pacevedom @dgoodwin are these name changes going to interfere with data collection for TRT? Should we wait to rename until we need the 9.3 job? Or go ahead and do it early?

dgoodwin:
Changing them will destroy any history, up to you if you decide to go ahead with that or not. 

I don't think I see any jobs with '.' in the name, if that's the case, it might be best to avoid being the first. rhel92 instead?

ci-operator/config/openshift/microshift/openshift-microshift-main.yaml
```diff
@@ -131,7 +131,7 @@ tests:
       DRY_RUN: "y"
     workflow: openshift-microshift-rebase
 - always_run: false
-  as: e2e-openshift-conformance-reduced
+  as: e2e-openshift-conformance-reduced-rhel9.2
```
dhellmann, line 134: Should we say `rhel-eus` instead of including the version number? Or do we want the history of 9.2 and 9.4 to be separate?

dgoodwin: That sounds fine assuming with 9.4 you do not intend to also keep testing 9.2.

ci-operator/config/openshift/microshift/openshift-microshift-main.yaml
```diff
@@ -141,7 +141,7 @@ tests:
       EC2_AMI: ami-02b8534ff4b424939
     workflow: openshift-microshift-e2e-openshift-conformance-reduced-aws
 - always_run: false
-  as: e2e-openshift-conformance-reduced-arm
+  as: e2e-openshift-conformance-reduced-arm-rhel9.2
```
pmtk, line 144: Why not `el92`? It's the same information with less chars

ci-operator/config/openshift/microshift/openshift-microshift-main.yaml
pacevedom: Renamed jobs to rhel-eus.

PR #42206, opened 2023-08-10
Fix issue DPTP-3553 Remove Secret object

bear-redhat:
The generated `hypershift-install.yaml` file includes a `Secret` object with AWS access key. We shall filter and remove it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -427,20 +427,22 @@ __check_defined = \
 	$(if $(value $1),, \
 		$(error Undefined environment variable $1$(if $2, ($2))))
 
-generate-hypershift-deplyment:
+# yq: https://github.com/mikefarah/yq
+yq ?= yq
+
+generate-hypershift-deployment:
 	@:$(call check_defined, MGMT_AWS_CONFIG_PATH)
-	@:$(call check_defined, TMPDIR)
 
-	mkdir -p ${TMPDIR}/hypershift-cli
-	oc image extract registry.ci.openshift.org/ci/hypershift-cli:latest --path /usr/bin/hypershift:${TMPDIR}/hypershift-cli --confirm
-	chmod +x ${TMPDIR}/hypershift-cli/hypershift
-
-	${TMPDIR}/hypershift-cli/hypershift \
+	$(SKIP_PULL) || $(CONTAINER_ENGINE) pull --platform linux/${GO_ARCH} registry.ci.openshift.org/ci/hypershift-cli:latest
+	$(CONTAINER_ENGINE) run $(USER) --platform linux/${GO_ARCH} \
+		--rm \
+		-v "$(MGMT_AWS_CONFIG_PATH):/mgmt-aws$(VOLUME_MOUNT_FLAGS)" \
+		registry.ci.openshift.org/ci/hypershift-cli:latest \
 		install \
 		--oidc-storage-provider-s3-bucket-name=hypershift-oidc-provider \
-		--oidc-storage-provider-s3-credentials=${MGMT_AWS_CONFIG_PATH} \
+		--oidc-storage-provider-s3-credentials=/mgmt-aws \
 		--oidc-storage-provider-s3-region=us-east-1 \
 		--hypershift-image=registry.ci.openshift.org/ci/hypershift-cli:latest \
 		--enable-uwm-telemetry-remote-write=false \
-		render > $(CURDIR)/clusters/hive/hypershift/hypershift-install.yaml
-.PHONY: generate-hypershift-deplyment
+		render | $(yq) eval 'select(.kind != "Secret")' > clusters/hive/hypershift/hypershift-install.yaml
+.PHONY: generate-hypershift-deployment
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -442,5 +444,5 @@ generate-hypershift-deplyment:
 		--oidc-storage-provider-s3-region=us-east-1 \
 		--hypershift-image=registry.ci.openshift.org/ci/hypershift-cli:latest \
 		--enable-uwm-telemetry-remote-write=false \
-		render > $(CURDIR)/clusters/hive/hypershift/hypershift-install.yaml
-.PHONY: generate-hypershift-deplyment
+		render | yq eval 'select(.kind != "Secret")' > clusters/hive/hypershift/hypershift-install.yaml
```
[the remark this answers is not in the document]
bear-redhat: Added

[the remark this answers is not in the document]
hongkailiu: > Could you please give a link to THE `yq` it requires?

Have you addressed this?

core-services/ci-secret-bootstrap/_config.yaml
```diff
@@ -2228,6 +2228,21 @@ secret_configs:
...
+      item: hypershift-operator-oidc-provider-s3-credentials
+    credentials:
+      field: credentials
+      item: hypershift-operator-oidc-provider-s3-credentials
+    region:
+      field: region
+      item: hypershift-operator-oidc-provider-s3-credentials
+  to:
+  - cluster_groups:
+    - build_farm
```
hongkailiu, line 2243: I thought the secret is on hive only, no?

Makefile
```diff
@@ -427,20 +427,39 @@ __check_defined = \
...
-	${TMPDIR}/hypershift-cli/hypershift \
+	$(SKIP_PULL) || $(CONTAINER_ENGINE) pull --platform linux/${GO_ARCH} registry.ci.openshift.org/ci/hypershift-cli:latest
+	$(CONTAINER_ENGINE) run $(USER) --platform linux/${GO_ARCH} \
+		--rm \
+		-v "$(MGMT_AWS_CONFIG_PATH):/mgmt-aws$(VOLUME_MOUNT_FLAGS)" \
+		registry.ci.openshift.org/ci/hypershift-cli:latest \
 		install \
 		--oidc-storage-provider-s3-bucket-name=hypershift-oidc-provider \
-		--oidc-storage-provider-s3-credentials=${MGMT_AWS_CONFIG_PATH} \
+		--oidc-storage-provider-s3-credentials=/mgmt-aws \
```
hongkailiu, line 443: Could we also do a variable for this?

Makefile
```diff
@@ -427,20 +427,39 @@ __check_defined = \
...
 		--oidc-storage-provider-s3-region=us-east-1 \
 		--hypershift-image=registry.ci.openshift.org/ci/hypershift-cli:latest \
 		--enable-uwm-telemetry-remote-write=false \
-		render > $(CURDIR)/clusters/hive/hypershift/hypershift-install.yaml
-.PHONY: generate-hypershift-deplyment
+		render | $(yq) eval 'select(.kind != "Secret")' > clusters/hive/hypershift/hypershift-install.yaml
+
+	# The following lines will avoid using docker image
```
hongkailiu, line 449: Let us remove those unused cmds or use them with if-else.

hongkailiu: I do not feel this is the right place to keep things.
You can save it in your own fork in case you need it in the future.

Makefile
hongkailiu: Oh. I see the point now.

Makefile
```diff
@@ -427,20 +427,22 @@ __check_defined = \
...
-	@:$(call check_defined, MGMT_AWS_CONFIG_PATH)
-	@:$(call check_defined, TMPDIR)
+# yq: https://github.com/mikefarah/yq
+yq ?= yq
 
-	mkdir -p ${TMPDIR}/hypershift-cli
-	oc image extract registry.ci.openshift.org/ci/hypershift-cli:latest --path /usr/bin/hypershift:${TMPDIR}/hypershift-cli --confirm
-	chmod +x ${TMPDIR}/hypershift-cli/hypershift
+generate-hypershift-deployment:
+	@:$(call check_defined, MGMT_AWS_CONFIG_PATH)
```
hongkailiu, line 434: I knew this is not introduced in this PR, but it would be much simpler if `MGMT_AWS_CONFIG_PATH` is defined with a default value (you could use the one when you call it from your laptop as you are probably the one most often) as we do for many of others. Thus, we do not need `check_defined` and `__check_defined` function at all.

I also checked the linked stackoverflow page. Even the following as the other suggestion looks more reasonable to me:
- I do not need to jump to another part of the makefile to understand what it does.
- The two `check` functions themselves are too esoteric. If there is a simple way of avoiding them, I would go there.

```
ifndef MY_FLAG
$(error MY_FLAG is not set)
endif
```

bear-redhat: `MGMT_AWS_CONFIG_PATH` has no and should have no default value. That's why its `mandatory`.

> I do not need to jump to another part of the makefile to understand what it does.
You don't need to. The name of that function already suggests the purpose of it

> The two check functions themselves are too esoteric. If there is a simple way of avoiding them, I would go there
I don't see repeatedly calling `ifndef` is better than using a function

PR #42215, opened 2023-08-10
ci-operator/config/openshift/release: Run credentials-request-freeze every 48h

wking:
Rolling back changes from 26ebda9d3d (#40924) and similar, CC @jmguzik.  `credentials-request-freeze` jobs are very cheap, because they do not launch a cluster (they just compare release image content), and running them more frequently gives plenty of post-merge time before a release gets cut to discuss what to do about changes like openshift/cluster-ingress-operator#954 adding `route53:ListTagsForResources` as an ingress permission in 4.12.


wking:
Nothing to rehearse; I'm just adjusting the interval.
 
 /pj-rehearse ack

PR #42219, opened 2023-08-11
add global hub image mirror.

morvencao:


ci-operator/step-registry/ocm/ci/image-mirror/ocm-ci-image-mirror-commands.sh
```diff
@@ -66,9 +66,9 @@ if [[ "$branch" == "main" || "$branch" == "master" ]]; then
 fi
 
 # Validate release branch. We can only run on release-x.y branches.
-if [[ ! ("$branch" =~ ^release-[0-9]+\.[0-9]+$ || "$branch" =~ ^backplane-[0-9]+\.[0-9]+$) ]]; then
-    log "ERROR Branch $branch is not a release or backplane branch."
-    log "      Base branch of PR must match release-x.y or backplane-x.y"
+if [[ ! ("$branch" =~ ^release-[0-9]+\.[0-9]+$ || "$branch" =~ ^backplane-[0-9]+\.[0-9]+$ || "$branch" =~ ^global-hub-[0-9]+\.[0-9]+$) ]]; then
```
gurnben, line 69: To avoid regex issues thorughout the remainder of the pipeline, we should limit the product name (in this case global-hub, in previous cases release/backplane) to a single word with no `-` separator.  

I know it may be disruptive, so let me know if its an issue, but could we rename the globalhub versioned branches to `globalhub-x.y`?  Moving forward, I think that will be our standard naming strategy if we leverage this mirroring workflow.

morvencao: Thanks @gurnben it makes total sense.
I updated the PR, we will update the branch names for global hub repos.

gurnben (approved):
/lgtm
/approve

PR #42234, opened 2023-08-11
fix: Trying a new Expression

sawood14012:
https://issues.redhat.com/browse/SANDBOX-57 

signed off by msawood@redhat.com


mfrancisc (approved):

ci-operator/config/codeready-toolchain/toolchain-e2e/codeready-toolchain-toolchain-e2e-master.yaml
```diff
@@ -18,7 +18,7 @@ resources:
       memory: 250Mi
 tests:
 - as: e2e
-  skip_if_only_changed: ^docs/|^\.github/|^\.tekton/|\.md$|\.adoc$|\.properties$|^(?:.*/)?(?:\.gitignore|.codecov.yml|.yamllint|OWNERS|PROJECT|LICENSE)$
+  skip_if_only_changed: ^doc/|^\.github/|\\.(md|adoc)$|^(?:.*/)?(?:\.gitignore|.codecov.yml|.yamllint|sonar-project.properties|OWNERS|PROJECT|LICENSE)$
```
MatousJobanek, line 21: I just realized, could you please also remove .github? 
the directory contains the part that builds the operators and pushes them into quay repo

sawood14012: if somebody modifies the github action workflows , it will just be a workflow change why do you want to run e2e on it

ci-operator/config/codeready-toolchain/toolchain-e2e/codeready-toolchain-toolchain-e2e-master.yaml
```diff
@@ -18,7 +18,7 @@ resources:
       memory: 250Mi
 tests:
 - as: e2e
-  skip_if_only_changed: ^docs/|^\.github/|^\.tekton/|\.md$|\.adoc$|\.properties$|^(?:.*/)?(?:\.gitignore|.codecov.yml|.yamllint|OWNERS|PROJECT|LICENSE)$
+  skip_if_only_changed: ^doc/|\\.(md|adoc)$|^(?:.*/)?(?:\.gitignore|.codecov.yml|.yamllint|sonar-project.properties|OWNERS|PROJECT|LICENSE)$
```
MatousJobanek, line 21: why not
```suggestion
  skip_if_only_changed: ^doc/|\.md$|\.adoc$|^(?:.*/)?(?:\.gitignore|.codecov.yml|.yamllint|sonar-project.properties|OWNERS|PROJECT|LICENSE)$
```

MatousJobanek: and if you want to keep the original one - are you sure about the two back-slashes?

MatousJobanek: could you please check the regex one more time - some of the parts look weird to me:
* you don't escape the forward slashes
* you don't escape the dots in .codecov.yml, .yamllint, and sonar-project.properties
* codecov can be also in the format .codecov.yaml (which is what we use in our projects)
* could you please explain why you need to use two non capturing groups?
* why the first part `^doc/|\.md$|\.adoc$` is not included in the group of the second part? 
* the beginning of the line char `^` seems to be redundant in front of the first non capturing group expression
* the `\.md$|\.adoc$|` matches only the suffixes, not the entire file names - is this sufficient? if so, then why we need to non capturing groups in the second part at all? 

TBH, I'm not good at regexes, so please correct me if I'm missing some important detail

sawood14012: both are fine  ref: https://docs.prow.k8s.io/docs/jobs/#triggering-jobs-based-on-changes

sawood14012: > * you don't escape the dots in .codecov.yml, .yamllint, and sonar-project.properties
> * codecov can be also in the format .codecov.yaml (which is what we use in our projects)

good one i will modify it to support yml | yaml

MatousJobanek: > both are fine ref: https://docs.prow.k8s.io/docs/jobs/#triggering-jobs-based-on-changes

I doubt that the version with two backslashes would be fine. Two backslashes result into one backslash, so you would expect the files to end with: `\.md`

Just to be sure, I tried it in go playground:

https://go.dev/play/p/zuVWv_jd_UI

MatousJobanek (approved):

PR #42240, opened 2023-08-11
HyperShift: make kubevirt e2e optional until we can get it passing again

csrwng:
Make kubevirt e2e job optional until it can be fixed


bryan-cox (approved):
/lgtm

PR #42242, opened 2023-08-11
`annotate.sh`: Add keep_going on failure

danilo-gemoli:
Keep annotating even after a failure
/cc @bradmwilliams


hack/annotate.sh
```diff
@@ -25,6 +26,10 @@ retry() {
       sleep $delay
     else
       echo "Retry $count/$retries exited $rc, no more retries left." > /dev/stderr
+	  if [ $keep_going -eq 1 ]; then
+	  	echo '$keep_going is set, failure is going to be skipped.' > /dev/stderr
+	  	return 0
+	  fi
```
bradmwilliams, line 32: I think moving on with the job is a good idea, but ultimately the jobs still needs to fail if there is an error on any attempt.

bradmwilliams (changes requested):

bradmwilliams (approved):

PR #42243, opened 2023-08-11
add os extension to ipsec

yuvalk:
this is to enable host support
needed for the north-south ipsec feature


yuvalk:
found out the info I was missing
closing this infavour of the old one

PR #42244, opened 2023-08-11
Update Owners file for rosa workflow to include folks from NA timezone

svetsa-rh:
Blocker issue background https://redhat-internal.slack.com/archives/CHWDU0CP9/p1691491358756139

### DIFF
--- a/ci-operator/step-registry/rosa/aws/non-sts/rosa-aws-non-sts-workflow.metadata.json
+++ b/ci-operator/step-registry/rosa/aws/non-sts/rosa-aws-non-sts-workflow.metadata.json
@@ -5,7 +5,9 @@
 			"yasun1",
 			"xueli181114",
 			"yuwang-RH",
-			"yufchang"
+			"yufchang",
+			"jtaleric",
+			"svetsa-rh"
 		],
 		"reviewers": [
 			"yasun1",
@@ -13,7 +15,9 @@
 			"yuwang-RH",
 			"tzhou5",
 			"yingzhanredhat",
-			"yufchang"
+			"yufchang",
+			"jtaleric",
+			"svetsa-rh"
 		]
 	}
 }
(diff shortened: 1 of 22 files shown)


svetsa-rh:
@mffiedler @rpattath Here is the PR requesting access to be able to update rosa workflows, to incorporate NA timezones.

svetsa-rh:
@yasun1 @yufchang @xueli181114 Can you ptal?

svetsa-rh:
cc: @jtaleric

jtaleric (approved):
yes please. This would be super helpful!

yasun1:
@svetsa-rh @jtaleric Many teams' jobs are based on the steps. To be the approver of the rosa cli provision steps, it requires the owner be the expert of the rosa cli features.

svetsa-rh:
> @svetsa-rh @jtaleric Many teams' jobs are based on the steps. To be the approver of the rosa cli provision steps, it requires the owner to be the expert on the rosa cli features. Could you explain more about the requirement?

@yasun1 So, the ask here is, to have ability to approve PRs during NA timezone so that we are unblocked and have a global coverage of approvers during multiple timezones. 

Both @jtaleric and I understand the sensitivity of approving any PRs at this level as we both are very well aware of the impact it has on many teams. And we agree to not abuse this power and be sensitive to use only when we hit blockers.

Without having any approvers in NA timezone and outside of China timezone, I am afraid that we will continue to be hampered if we run into any issues and want to stop bleeding. For ex: We had multiple QE and developers impacted by the issue discussed here https://redhat-internal.slack.com/archives/CHWDU0CP9/p1691491358756139 and lost productivity for multiple days. 

Another big one I had in mind was also China's Mid-Autumn festival where you folks won't be available for multiple days. 

Hope this helps!

cc: @mffiedler

yasun1:
@svetsa-rh It makes sense. I will grant the apply for you. But I advice that if there are changes on the core step `rosa-cluster-provision`, please ping ocm qes to have a quick review to sync the changes on both of our two sides. Thanks.

PR #41867, opened 2023-08-01
[LPTOCPCI-579] Add a firewatch rule to ignore test_failures in "gather-*" steps in interop scenarios

calebevans:

PR #42246, opened 2023-08-11
Update jira projects for firewatch config.

ascerra:
This PR is meant to correct the jira projects that we report to using the firewatch config.

This is a result of the comment here: https://github.com/openshift/release/pull/41867#discussion_r1285303866

We are also changing the cron value here to allow for the use of the new trigger job
- Details are [here](https://cspi-qe.gitbook.io/ocp-ci-onboarding/tutorials/triggering_guide#weekly-trigger-job)
- This will not effect the current cadence, it will still run weekly on Monday morning.
- This just allows the interop program to more easily manage the 40+ interop scenarios.


calebevans (approved):
/lgtm

PR #42273, opened 2023-08-14
Before-branching 4.14: Bump missed 4.13 references

psalajova:
Bump references to 4.13 that were missed in https://github.com/openshift/release/pull/41475

/cc @jmguzik @danilo-gemoli

### DIFF
--- a/core-services/release-controller/_releases/priv/release-ocp-4.15.json
+++ b/core-services/release-controller/_releases/priv/release-ocp-4.15.json
@@ -62,7 +62,7 @@
             "upgradeFromRelease": {
                 "candidate": {
                     "stream": "nightly",
-                    "version": "4.13"
+                    "version": "4.14"
                 }
             }
         },
@@ -293,7 +293,7 @@
             "upgradeFromRelease": {
                 "candidate": {
                     "stream": "nightly",
-                    "version": "4.13"
+                    "version": "4.14"
                 }
             }
         },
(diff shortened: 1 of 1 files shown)


danilo-gemoli:
It's good but we need to understand some bits of config before going on.
/hold
Feel free to unhold if we are in hurry.

core-services/release-controller/_releases/priv/release-ocp-4.15.json
```diff
@@ -62,7 +62,7 @@
             "upgradeFromRelease": {
```
danilo-gemoli, line 62: @deepsm007 Hey sorry for the interrupt but I need some help here. Do you know if 
```json
"upgradeFromRelease": {
  "candidate": {
    "stream": "nightly",
     "version": "4.14"
 }
```
can be safely replaced by
```json
  "upgradeFrom": "PreviousMinor"
```
in this context? It would be easier for us to maintain as the won't have to bump the version every time we perform the branch cut.

danilo-gemoli: @bradmwilliams words from [this thread](https://redhat-internal.slack.com/archives/CNHC2DK2M/p1692023891223699):
> No, the release-controller would return an rc version always and not the corresponding nightly or ci stream.

So no, it doesn't work here.

PR #42274, opened 2023-08-14
Update OWNER_ALIASES: Add members to cspi-qe-ocp-lp

oharan2:


OWNERS_ALIASES
```diff
@@ -267,6 +267,7 @@ aliases:
   - Sau1506mya
   - amp-rh
   - macharya0907
```
ascerra, line 269:
```suggestion
```

Let's remove Jay as he's no longer on the team

OWNERS_ALIASES
```diff
@@ -267,6 +267,7 @@ aliases:
...
+  - oharan2
```
ascerra, line 270: Can we also add other new team members

```suggestion
  - shakyav
  - anramesh27
  - aviavraham
  - dfrazzette
```

PR #42289, opened 2023-08-14
Removing the branch protection rule

jpodivin:
in order to unify behavior across the project


jpodivin:
We are probably going to expand the rule instead.

PR #42315, opened 2023-08-14
increase timeout value for aggregator step

xueqzhan:
This value should be >= what we allow for aggregator's internal timeouts.


dgoodwin:
/lgtm

We have to figure out where to doc this in all places, it seems we missed one when we did the bump for the 1.28 rebase again.

PR #42335, opened 2023-08-15
Allow underscores in filters for longduration and disruptive jobs

fxierh:
Currently underscores are allowed in filters for the `openshift-extended-test` step, but not the `-longduration` and `-disruptive` equivalents. 

We the Hive QE team intend to use filters (in our longduration jobs) which look like the following:
- HIVE_SD_A
- HIVE_SD_R
- HIVE_SPEC

We kindly request support for underscores in filters through this PR.


fxierh:
@kuiwang02 Thanks for your suggestions. The PR is ready for review.

PR #42362, opened 2023-08-15
hive periodics: Do not use $PULL_NUMBER

2uasimojo:
The e2e-pool periodic failed due to the `$CLUSTER_NAME` variable being unset, which happened because (apparently) the test runner is running with `-o nounset`, causing the subshell in this line to fail (but not torpedo the test at that stage):

```
export CLUSTER_NAME=$(printf "hiveci-%x-%x" $(date +%s) $PULL_NUMBER)
```

I'm guessing this didn't fail in rehearsals because `$PULL_NUMBER` would have been set to the pull number of the release PR in play. (That's arguably a bug in the rehearsal runner.)

So with this commit we remove the use of `$PULL_NUMBER` and just suffix the `$CLUSTER_NAME` with `-0` to ensure it doesn't conflict with the same variable in presubmit e2e tests.

----

Also fix an issue where `USE_MANAGED_DNS` wasn't being sent to the test
properly. This was fixed in the presubmits via f74af17.

[HIVE-2235](https://issues.redhat.com//browse/HIVE-2235)

### DIFF
--- a/ci-operator/config/openshift/hive/openshift-hive-master__periodic.yaml
+++ b/ci-operator/config/openshift/hive/openshift-hive-master__periodic.yaml
@@ -63,10 +63,9 @@ tests:
         # Configure a cluster name that our monitor/reaper can use to
         # - discover clusters created by this test (prefix `hiveci-`)
         # - know when they were created (embedded date stamp)
-        # Add the PR number for extra uniqueness.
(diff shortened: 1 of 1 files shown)


2uasimojo:
/test build-clusters

wha?

PR #42364, opened 2023-08-15
Temporarily remove build01-dry check and build-clusters check

psalajova:
Remove these jobs while build01 is in failover


deepsm007:
You might have to run `make update-ci-build-clusters` to pass the check

PR #42378, opened 2023-08-15
add image creation for check-payload

rphillips:
Adds check-payload image to OCP. The goal is to get this image into the OCP payload.


rphillips:
There was a [Slack discussion](https://redhat-internal.slack.com/archives/CB95J6R4N/p1692283225151219). This PR is ready.

PR #42392, opened 2023-08-16
sts:AssumeRole is required for Shared-VPC install

yunjiang29:
Fix https://issues.redhat.com/browse/OCPBUGS-17751


yunjiang29:
@jianlinliu PTAL

PR #42395, opened 2023-08-16
upgrade osd flavors to 0.7.6 image

davdhacs:
/hold
/unassign
/uncc


gavin-stackrox:
@davdhacs the osd-* failures are a known flake with that test. 
/lgtm

PR #42400, opened 2023-08-16
update test frequency for 4.13 gcp disasterrecovery job

jianlinliu:
gcp disasterrecovery job in 4.12 and 4.14 are on f14, seem like only 4.13 missed some change, sync them onto f14.


jianlinliu:
cc @liangxia to review.

PR #42417, opened 2023-08-16
CVP-3866: Add a new step to copy Brew registry credentials

gastoner:


nmars:
One doc string nit, otherwise lgtm

PR #42428, opened 2023-08-16
hypershift: cluster-image-registry-operator: add conformance presubmit test

sjenning:
https://github.com/openshift/cluster-image-registry-operator/pull/867 took out hypershift conformance, which blocks ART-built nightlies and feed the EC stream.

TRT investigation https://issues.redhat.com/browse/TRT-1193

Need conformance protection as this component can break hypershift in ways our standard e2e does not detect i.e. functioning internal image registry.

### DIFF
--- a/ci-operator/config/openshift/cluster-image-registry-operator/openshift-cluster-image-registry-operator-master.yaml
+++ b/ci-operator/config/openshift/cluster-image-registry-operator/openshift-cluster-image-registry-operator-master.yaml
@@ -161,6 +161,10 @@ tests:
   steps:
     cluster_profile: hypershift
     workflow: hypershift-aws-e2e-external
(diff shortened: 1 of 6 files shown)


flavianmissi:
Is there no way the standard e2e can cover this? feels like a simple test.
I understand and agree that the registry should not break hypershift, but I'm hesitant to adding more suites to the image registry and end up slowing down our merge rate even more... we have plenty of flakes with our current runs :/

sjenning:
/pj-rehearse ack

for reference, this is the testgrid dashboard monitoring the health of this job
https://testgrid.k8s.io/redhat-hypershift#4.14-aws-ovn-conformance

It blocks nightly promotions so, if it does break, we know almost immediately and it is urgent priority to fix it.

PR #42430, opened 2023-08-16
USHIFT-958: Always run openshift conformance reduced

pacevedom:


ci-operator/config/openshift/microshift/openshift-microshift-main.yaml
```diff
@@ -130,8 +130,7 @@ tests:
     env:
       DRY_RUN: "y"
     workflow: openshift-microshift-rebase
-- always_run: false
-  as: e2e-openshift-conformance-reduced
+- as: e2e-openshift-conformance-reduced
```
pmtk, line 133: What is the functional impact of this change?
I had impression this job runs on each PR anyway, so what will change?

pacevedom: It shouldnt run in every PR right now. Anyway this is the first step towards the final status, which is to require the job again (once it passes).

pmtk: Does `always_run` overrides `run_if_changed`? And in such case no matter which file changed (even docs/), the job would run?

PR #42439, opened 2023-08-16
redhat-appstudio/e2e-tests: Skip e2e-tests job when changes are only in load-test files

pmacik:
Changes to any file under `tests/load-tests` as well as changes to `cmd/loadTests.go` are not involved in the e2e-tests so running the e2e tests for PR that are only related to the load test does not bring any value.

On the other hand, skipping them in such PRs significantly accelerate load test development and maintanence.


flacatus (approved):

flacatus (approved):

jhutar (commented):
Nice!

PR #42453, opened 2023-08-17
Replace vsphere platform to vsphere-dis for disconnected jobs

mhanss:


mhanss:
@liangxia and @hoxhaeris please review.

hoxhaeris (approved):

hoxhaeris:
I can't see a respective job defined here: 
https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift/release/openshift-release-infra-periodics.yaml
for this workflow. I believe this won't work as intended.

liangxia:
@hoxhaeris Those workflow comes from QE, and are defined in https://github.com/openshift/release/tree/master/ci-operator/step-registry/cucushift/agent/vsphere/disconnected/install

hoxhaeris:
> @hoxhaeris Those workflow comes from QE, and are defined in https://github.com/openshift/release/tree/master/ci-operator/step-registry/cucushift/agent/vsphere/disconnected/install

Cluster-Bot will reference the jobs that are defined [here](https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift/release/openshift-release-infra-periodics.yaml). 
The one you are referring to, the step-registry configuration, is part of what's necessary, but not all.

PR #42524, opened 2023-08-21
Add commitchecker to storage repositories

jsafrane:
Adding commitchecker to all repos that are fork of some upstream and are maintained by the storage team.

I avoided repos that are maintained by other teams, like OpenStack, IBM, oVirt or Alibaba.

Note: verify-commits rehearsals should fail with:

`INFO[2023-08-21T09:11:24Z] ERROR: couldn't find commits from 92a04b63235ce880ebe08d4d277a40d9ee2d24f1..HEAD: error executing git log: fatal: Invalid revision range 92a04b63235ce880ebe08d4d277a40d9ee2d24f1..HEAD`

It tries to check this PR with commit https://github.com/openshift/release/commit/92a04b63235ce880ebe08d4d277a40d9ee2d24f1 in the CSI driver / sidecar repository, but this PR is in openshift/release

@openshift/storage


jsafrane:
Add secrets-store

PR #42477, opened 2023-08-17
stackrox: add initial OCP 4.14 tests

gavin-stackrox:
/uncc
/hold

PR #42541, opened 2023-08-22
Draft: stackrox: add initial OCP 4.14 tests

BradLugo:
/hold
/uncc


BradLugo:
Related to https://github.com/openshift/release/pull/42477

BradLugo:
We've confirmed that the operator works in our manual testing because of _how_ the operator was deployed, not which version of OCP 4.14 we were using. Closing

PR #42558, opened 2023-08-22
Add steps for oidc-config and operator-roles

yasun1:


ci-operator/step-registry/rosa/sts/operator-roles/create/rosa-sts-operator-roles-create-commands.sh
```diff
@@ -0,0 +1,61 @@
...
+# Create operator roles
+echo "Create the operator roles with the prefix ${OPERATOR_ROLES_PREFIX}..."
+rosa create operator-roles -y --mode auto \
+                           --prefix ${OPERATOR_ROLES_PREFIX} \
+                           --oidc-config-id ${oidc_config_id} \
+                           --installer-role-arn ${account_installer_role_arn} \
+                           ${HOSTED_CP_SWITCH} \
+                           ${SHARED_VPC_SWITCH}
+# Store the operator-roles-prefix for the post steps and the operator roles deletion
+echo -n "${OPERATOR_ROLES_PREFIX}" > "${SHARED_DIR}/operator-roles-prefix"
```
yuwang-RH, line 61: This step is to create operator-roles prior to cluster spec, is there any other step to create operator roles after cluster spect?

yuwang-RH:
@yasun1 And the account roles and operator roles creation support the channel-group flag which is mainly used to decide the latest supported version. Not sure if we want to support it in this PR, just a reminder

yuwang-RH:
Others LGTM

ci-operator/step-registry/rosa/sts/operator-roles/create/rosa-sts-operator-roles-create-commands.sh
yasun1: This is only for byo operator roles. I assume there is no user case with '--cluster' for the operator-roles in Prow CI. We merged it into the cluster provision.

PR #42562, opened 2023-08-22
VolSync: CI-replace golang1.20 builder with ocp builder

tesshuflower:
Make sure we're building with the ocp golang1.20 builder in CI docker build for FIPS support


JohnStrunk (commented):
/lgtm

PR #42572, opened 2023-08-22
[OKO] Hive: Update CDC-23 and cluster-pools

viroel:
This patch do some updates for oko cluster pools:
 - Update oko-23 cdc with new floating ips
 - Remove one cluster from oko-op-04 and move to oko-op-02, to test new configurations
 - Disable another 4.10 cluster pool which is not being used anymore


rlandy (commented):
/lgtm

PR #42617, opened 2023-08-23
Microshift rpm-upgrade: fix race condition post-upgrade

copejon:
There is a small race window during which there may be more pods than greenboot expects. Adding a retry loop on the greenboot script should provide topolvm enough time to terminate the old pods


copejon:
/hold

This may be handled by openshift/microshift#2255

copejon:
The microshift PR fixes this, will close

PR #40760, opened 2023-06-28
Add health check for post install

yunjiang29:

PR #42632, opened 2023-08-24
OCPQE-16275: Enable cucushift-installer-check to all AWS workflows

yunjiang29:
Previous AWS work https://github.com/openshift/release/pull/40760


yunjiang29:
@jianlinliu @gpei @jinyunma PTAL

PR #42672, opened 2023-08-24
Parameterizing sdn to ovn migration timeout

vishnuchalla:
Adding a generic timeout for the ovn-sdn-migration ref as it is resulting in some of the jobs failures. More context [here](https://github.com/openshift/release/pull/42306#issuecomment-1685910024). 
This option should provide us a flexibility to decide on a generic timeout for all the steps involved. If not set, It will run with the default values.

### DIFF
--- a/ci-operator/step-registry/ovn/sdn-migration/ovn-sdn-migration-commands.sh
+++ b/ci-operator/step-registry/ovn/sdn-migration/ovn-sdn-migration-commands.sh
@@ -5,8 +5,13 @@ set -o nounset
 set -o pipefail
 
 TARGET=${TARGET:-OVNKubernetes}
+# Check if the OVN_SDN_MIGRATION_TIMEOUT environment variable is set and is equal to "0s"
+if [ -n "$OVN_SDN_MIGRATION_TIMEOUT" ] && [ "$OVN_SDN_MIGRATION_TIMEOUT" = "0s" ]; then
+  unset OVN_SDN_MIGRATION_TIMEOUT
+fi
 
-timeout 1200s bash <<EOT
+co_timeout=${OVN_SDN_MIGRATION_TIMEOUT:-1200s}
+timeout "$co_timeout" bash <<EOT
 until
   oc wait co --all --for='condition=AVAILABLE=True' --timeout=10s && \
   oc wait co --all --for='condition=PROGRESSING=False' --timeout=10s && \
@@ -18,7 +23,8 @@ done
 EOT
 
 oc patch Network.operator.openshift.io cluster --type='merge'   --patch '{"spec":{"migration":null}}'
-timeout 60s bash <<EOT
+cno_timeout=${OVN_SDN_MIGRATION_TIMEOUT:-60s}
+timeout "$cno_timeout" bash <<EOT
 until 
   ! oc get network -o yaml | grep migration > /dev/null
 do
@@ -30,10 +36,12 @@ EOT
 # Change network to target network in Network.operator.openshift.io the CR to trigger machine config update by MCO.
 oc patch Network.operator.openshift.io cluster --type='merge' --patch "{\"spec\":{\"migration\":{\"networkType\":\"${TARGET}\"}}}"
 # Wait until MCO starts applying new machine config to nodes
-oc wait mcp --all --for='condition=UPDATING=True' --timeout=300s
+mco_timeout=${OVN_SDN_MIGRATION_TIMEOUT:-300s}
+oc wait mcp --all --for='condition=UPDATING=True' --timeout="$mco_timeout"
 
 # Wait until MCO finishes its work or it reachs the 20mins timeout
-timeout 2700s bash <<EOT
+mcp_timeout=${OVN_SDN_MIGRATION_TIMEOUT:-2700s}
+timeout "$mcp_timeout" bash <<EOT
 until
   oc wait mcp --all --for='condition=UPDATED=True' --timeout=10s && \
   oc wait mcp --all --for='condition=UPDATING=False' --timeout=10s && \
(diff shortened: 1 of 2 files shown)


vishnuchalla:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.12-amd64-nightly-gcp-ipi-sdn-migration-ovn-f14

vishnuchalla:
/pj-rehearse periodic-ci-openshift-release-master-ci-4.14-e2e-aws-ovn-network-migration

vishnuchalla:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.12-amd64-nightly-gcp-ipi-sdn-migration-ovn-f14

vishnuchalla:
> @vishnuchalla I guess you want to increase the timeout when running this script in a large cluster. May I ask what is the size of the cluster? And at which step the current timeout is not big enough?

We have cluster with 120 worker nodes on which we plan to trigger some large scale tests.

danwinship:
> I guess you want to increase the timeout when running this script in a large cluster.

It seems like it would be better to just count the number of nodes at the top and then adjust the timeouts based on that? (eg `timeout $((num_nodes * 20)) bash`)

vishnuchalla:
> > I guess you want to increase the timeout when running this script in a large cluster.
> 
> It seems like it would be better to just count the number of nodes at the top and then adjust the timeouts based on that? (eg `timeout $((num_nodes * 20)) bash`)

That might work in some cases and we cannot guarantee it to me stable across all types of runs and we might often re-visit this part of code to change that fixed value if something breaks. Instead I think we should have a variable exposed as in this PR, so that users can have flexibility to experiment and decide on the right number based on their use case. That way the code will have minimal changes and provides flexibility at the user end to set their desired value on timeouts.

jluhrsen:
> > > I guess you want to increase the timeout when running this script in a large cluster.
> > 
> > 
> > It seems like it would be better to just count the number of nodes at the top and then adjust the timeouts based on that? (eg `timeout $((num_nodes * 20)) bash`)
> 
> That might work in some cases and we cannot guarantee it to be stable across all types of runs and we might often re-visit this part of code to change that fixed value if something breaks. Instead I think we should have a variable exposed as in this PR, so that users can have flexibility to experiment and decide on the right number based on their use case. That way the code will have minimal changes and provides flexibility at the user level to set their desired value on timeouts.

if this variable is set to some value (non zero) then that is the timeout used for all steps, whereas before the
timeouts were specific to each step. e.g., 1200s for co to stop progressing, 60s for CNO to be patched, etc.

That could really blow up on us if there is an issue. would it be overkill to add specific timeouts for each
step so you can keep defaults like 60s for one thing, but tweak something more major like CO progressing?

vishnuchalla:
> > > > I guess you want to increase the timeout when running this script in a large cluster.
> > > 
> > > 
> > > It seems like it would be better to just count the number of nodes at the top and then adjust the timeouts based on that? (eg `timeout $((num_nodes * 20)) bash`)
> > 
> > 
> > That might work in some cases and we cannot guarantee it to be stable across all types of runs and we might often re-visit this part of code to change that fixed value if something breaks. Instead I think we should have a variable exposed as in this PR, so that users can have flexibility to experiment and decide on the right number based on their use case. That way the code will have minimal changes and provides flexibility at the user level to set their desired value on timeouts.
> 
> if this variable is set to some value (non zero) then that is the timeout used for all steps, whereas before the timeouts were specific to each step. e.g., 1200s for co to stop progressing, 60s for CNO to be patched, etc.
> 
> That could really blow up on us if there is an issue. would it be overkill to add specific timeouts for each step so you can keep defaults like 60s for one thing, but tweak something more major like CO progressing?

Yes defaults will stay in place. This change is to provide users flexibility to modify timeout according to their use case. If they set a lower value its their own set failure, but if they set it to a greater one then it will be beneficial. And also I don't think the user who is going to set this timeout value is going to set it without totally being aware of where and how it is being used.

jluhrsen:
> > > > > I guess you want to increase the timeout when running this script in a large cluster.
> > > > 
> > > > 
> > > > It seems like it would be better to just count the number of nodes at the top and then adjust the timeouts based on that? (eg `timeout $((num_nodes * 20)) bash`)
> > > 
> > > 
> > > That might work in some cases and we cannot guarantee it to be stable across all types of runs and we might often re-visit this part of code to change that fixed value if something breaks. Instead I think we should have a variable exposed as in this PR, so that users can have flexibility to experiment and decide on the right number based on their use case. That way the code will have minimal changes and provides flexibility at the user level to set their desired value on timeouts.
> > 
> > 
> > if this variable is set to some value (non zero) then that is the timeout used for all steps, whereas before the timeouts were specific to each step. e.g., 1200s for co to stop progressing, 60s for CNO to be patched, etc.
> > That could really blow up on us if there is an issue. would it be overkill to add specific timeouts for each step so you can keep defaults like 60s for one thing, but tweak something more major like CO progressing?
> 
> Yes defaults will stay in place. This change is to provide users flexibility to modify timeout according to their use case. If they set a lower value its their own set failure, but if they set it to a greater one then it will be beneficial. And also I don't think that the user who is going to set this timeout value is going to set it without totally being aware of where and how it is being used. And also if someone reaches out with a failure reason due to timeout, we can simply redirect them to adjust `OVN_SDN_MIGRATION_TIMEOUT` value according to their own use case. Its totally up to them to decide.

gotcha, and it makes sense. I like the idea. my only concern is that I would assume that if someone is going
to tweak this timeout on purpose they are going to use a larger value (> 2700 for example) and that is
going to override a 60s timeout which is pretty dramatic. There are two cases where that is the case. Anyway,
if you think it's overkill to add a new variable for each specific timeout case, then no worries on my side.

vishnuchalla:
> > > > > > I guess you want to increase the timeout when running this script in a large cluster.
> > > > > 
> > > > > 
> > > > > It seems like it would be better to just count the number of nodes at the top and then adjust the timeouts based on that? (eg `timeout $((num_nodes * 20)) bash`)
> > > > 
> > > > 
> > > > That might work in some cases and we cannot guarantee it to be stable across all types of runs and we might often re-visit this part of code to change that fixed value if something breaks. Instead I think we should have a variable exposed as in this PR, so that users can have flexibility to experiment and decide on the right number based on their use case. That way the code will have minimal changes and provides flexibility at the user level to set their desired value on timeouts.
> > > 
> > > 
> > > if this variable is set to some value (non zero) then that is the timeout used for all steps, whereas before the timeouts were specific to each step. e.g., 1200s for co to stop progressing, 60s for CNO to be patched, etc.
> > > That could really blow up on us if there is an issue. would it be overkill to add specific timeouts for each step so you can keep defaults like 60s for one thing, but tweak something more major like CO progressing?
> > 
> > 
> > Yes defaults will stay in place. This change is to provide users flexibility to modify timeout according to their use case. If they set a lower value its their own set failure, but if they set it to a greater one then it will be beneficial. And also I don't think that the user who is going to set this timeout value is going to set it without totally being aware of where and how it is being used. And also if someone reaches out with a failure reason due to timeout, we can simply redirect them to adjust `OVN_SDN_MIGRATION_TIMEOUT` value according to their own use case. Its totally up to them to decide.
> 
> gotcha, and it makes sense. I like the idea. my only concern is that I would assume that if someone is going to tweak this timeout on purpose they are going to use a larger value (> 2700 for example) and that is going to override a 60s timeout which is pretty dramatic. There are two cases where that is the case. Anyway, if you think it's overkill to add a new variable for each specific timeout case, then no worries on my side.

Yes, Even if the value is larger timeout is just a timeout. As a process need not wait until the timeout to finish and as its just a limit on execution time, I think it should be fine.

PR #42683, opened 2023-08-25
Add slack alert for failed 4.11 rollback job

shellyyang1989:


shellyyang1989:
cc @jianlinliu @wking @jiajliu for review

wking (approved):
Catching up with #42020, which made this job a nightly blocker.  4.12 doesn't have this reporter yet, but it's also not a nightly blocker yet.

/lgtm
/apprive

PR #42707, opened 2023-08-26
Added govulncheck to backplane-cli CI

Dee-6777:
This PR is for the issue [OSD-18227](https://issues.redhat.com/browse/OSD-18227) to add govulncheck in backplane-cli CI check.

### DIFF
--- a/ci-operator/config/openshift/backplane-cli/openshift-backplane-cli-main.yaml
+++ b/ci-operator/config/openshift/backplane-cli/openshift-backplane-cli-main.yaml
@@ -7,7 +7,7 @@ build_root:
   image_stream_tag:
     name: release
     namespace: openshift
-    tag: golang-1.20
+    tag: golang-1.19
 images:
 - dockerfile_path: Dockerfile
   from: base


ci-operator/jobs/openshift/backplane-cli/openshift-backplane-cli-main-presubmits.yaml
```diff
@@ -252,3 +252,52 @@ presubmits:
...
+    cluster: build04
+    context: ci/prow/govulncheck
+    decorate: true
+    decoration_config:
+      skip_cloning: true
+    labels:
+      ci.openshift.io/generator: prowgen
+      pj-rehearse.openshift.io/can-be-rehearsed: "true"
+    name: pull-ci-openshift-backplane-cli-main-test
+    rerun_command: /test govulncheck
```
samanthajayasinghe, line 269: Is it not `/test scan` ?

supreeth7 (changes requested):
This change is missing the job configuration.
You will need to run `make jobs` to generate Prow job configuration files out of ci-operator configuration files.

Suggestion: Try running `make jobs WHAT=openshift/backplane-cli` from the root of the repository to fix the issue.

ci-operator/config/openshift/backplane-cli/openshift-backplane-cli-main.yaml
```diff
@@ -28,6 +28,11 @@ tests:
     make test
   container:
     from: src
+- as: scan
+  commands: |
+    make scan
+  container:
+    from: src
```
supreeth7, line 35: This should be an optional check

```suggestion
- as: scan
  commands: make scan
  container:
    from: src
  optional: true
```

supreeth7 (changes requested):

ci-operator/config/openshift/backplane-cli/openshift-backplane-cli-main.yaml
```diff
@@ -28,6 +28,10 @@ tests:
     make test
   container:
     from: src
+- as: scan
+  commands: make scan
+  container:
+    from: src
```
supreeth7, line 34: This should be optional as stated earlier.

```suggestion
- as: scan
  commands: make scan
  container:
    from: src
  options: true
```

supreeth7 (changes requested):

supreeth7 (changes requested):
I've had a look at the Backplane CLI Makefile and I do not see govulncheck as a prerequisite. Please ensure you first install govulncheck before running `make scan` target. This change needs to happen in the Backplane CLI Makefile.

supreeth7:
@Dee-6777 In the mentioned PR, you've added the make target which is correct. But the assumption here that `govulncheck` binary will already present isn't right, we have to first ensure that the binary is present, if not, we'll have to install (download) the binary and then run the `scan` make target.

For example. [here](https://github.com/openshift/backplane-cli/blob/main/Makefile#L77) we first ensure that `goreleaser` binary is present and then run the release target.

Dee-6777:
@supreeth7 Thanks!
I carried out the changes in this [PR](https://github.com/openshift/backplane-cli/pull/198), it will first check if govulncheck binary present or not and then it will install

supreeth7 (approved):
Rehearsals are ought to fail as they rehearse (force-test) the newly added CI step against the repository.

/pj-rehearse ack
/unhold
/lgtm

ci-operator/config/openshift/backplane-cli/openshift-backplane-cli-main.yaml
```diff
@@ -7,7 +7,7 @@ build_root:
   image_stream_tag:
     name: release
     namespace: openshift
-    tag: golang-1.18
+    tag: golang-1.21
```
supreeth7, line 10: ubi-minimal/ubi base images don't have support for go1.21 yet, only till go1.20.
```suggestion
    tag: golang-1.20
```

supreeth7 (changes requested):

ci-operator/config/openshift/backplane-cli/openshift-backplane-cli-main.yaml
```diff
@@ -7,7 +7,7 @@ build_root:
   image_stream_tag:
     name: release
     namespace: openshift
-    tag: golang-1.18
+    tag: golang-1.20
```
supreeth7, line 10: I reckon go1.20 tag isn't published yet as well, let's fallback to go1.19.
```suggestion
    tag: golang-1.19
```

supreeth7 (changes requested):

Dee-6777:
@supreeth7 I think the version of CI and the go version in bp-cli should be same to fix this error. Because go version in go.mod for bpcli is 1.21 and all test cases passed when I made CI version to 1.21

supreeth7:
@Dee-6777 The rehearsals didn't pass back then as go1.21 image tag doesn't exist in the `release/openshift` registry.
The `make scan` is expected to fail as every release of go will have vulnerabilities that will be fixed in the next release, but we cannot keep bumping our base image to the latest go version and expect it to be a support base image in the OCP image registry.

I'd like to make the control over our CI easier, I'll address it in a couple of PRs soon.

supreeth7 (approved):
/lgtm
/pj-rehearse ack

Dee-6777:
@supreeth7 for **ci/rehearse/openshift/backplane-cli/main/lint** to pass shall I created a new PR replacing all the ioutil to io

PR #42714, opened 2023-08-27
Temporarily change build cluster for assisted repos due to outage

osherdp:
Changing the build-cluster from build03 to build01, because there's a major outage for that build farm:
```
kubelet: Error: kubelet may be retrying requests that are timing out in
CRI-O due to system load. Currently at stage container storage creation:
context deadline exceeded: error reserving ctr name
k8s_sidecar_e2e-ai-operator-ztp-baremetalds-devscripts-proxy_ci-op-1k2d0bx9_346ac336-95a1-4957-bcce-d7bbd2cea00a_0
for id 50a99700233edd619b8b2d3a9d0b5f8e1b0f8d5d24596e8ea15d5b8e1da7788d:
name is reserved
```

(for example
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_assisted-service/5446/pull-ci-openshift-assisted-service-release-ocm-2.8-e2e-ai-operator-ztp/1695757257972649984)

Hopefully after the outage will get mitigated we can revert it back to the original state.

/cc @filanov @eliorerz

### DIFF
--- a/ci-operator/config/openshift/assisted-image-service/openshift-assisted-image-service-main__edge.yaml
+++ b/ci-operator/config/openshift/assisted-image-service/openshift-assisted-image-service-main__edge.yaml
@@ -85,6 +85,7 @@ test_binary_build_commands: |
   go mod vendor
 tests:
 - as: mirror-nightly-image
+  cluster: build01
   cron: '@daily'
   steps:
     dependencies:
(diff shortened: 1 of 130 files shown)


filanov (approved):

osherdp:
Original issue seems to be solved. I do see in rehearsals 2 other issues:
* pip no longer supports Python 3.6 which we probably still use (in the provisioning OS, not for the container image)
* some infra issues with vsphere env (``error fetching network: network 'ci-segment-130' not found``)

/pj-rehearse ack

PR #42730, opened 2023-08-28
Add test for checking manual changes in zuul jobs

adrianfusco:
The test generates again the molecule.yaml and project.yaml files. If there's a diff it means there's a manual change that should be fixed


Sandeepyadav93:
/lgtm cancel

We discussed another approach over Slack conversation about running this in pre-commit

adrianfusco:
Done in https://github.com/openstack-k8s-operators/ci-framework/pull/481

PR #42731, opened 2023-08-28
GitOps | Fixing Periodic job

varshab1210:


anandrkskd (changes requested):
LGTM 
but can you add the Slack notification config here as well

PR #42734, opened 2023-08-28
test: add three test cases to osd fm prow job

pawelpaszki:
## WHAT
test: add three test cases to osd fm prow job

## Verification
1. tests should pass
2. Manually go through the code and confirm that it resembles tests carried out for the test cases: `OCP-63511`, `OCP-60338` and `OCP-63998` in polarion

### DIFF
--- a/ci-operator/step-registry/osd-ccs/fleet-manager/osd-ccs-fleet-manager-workflow.yaml
+++ b/ci-operator/step-registry/osd-ccs/fleet-manager/osd-ccs-fleet-manager-workflow.yaml
@@ -13,8 +13,6 @@ workflow:
     - ref: osd-ccs-fleet-manager-sc-provision
     - chain: rosa-aws-sts-provision-byo-vpc
     - ref: osd-ccs-conf-idp-htpasswd-multi-users
-    test:
-    - ref: osd-ccs-fleet-manager-test
     post:
     - chain: rosa-aws-sts-deprovision-byo-vpc
     - ref: osd-ccs-fleet-manager-sc-deprovision
--- a/ci-operator/step-registry/osd-ccs/fleet-manager/sc/provision/osd-ccs-fleet-manager-sc-provision-commands.sh
+++ b/ci-operator/step-registry/osd-ccs/fleet-manager/sc/provision/osd-ccs-fleet-manager-sc-provision-commands.sh
@@ -66,7 +66,7 @@ fi
 sc_cluster_id=$(echo '{"region": "'${OSDFM_REGION}'", "cloud_provider": "aws"}' | ocm post /api/osd_fleet_mgmt/v1/service_clusters | jq -r '.id')
 # Save Service Cluster info
 echo "Service Cluster fm id:${sc_cluster_id}"
-echo "${sc_cluster_id}" > "${ARTIFACT_DIR}/osd-fm-sc-id"
+echo "${sc_cluster_id}" > "${SHARED_DIR}/osd-fm-sc-id"
 
 # Check if Service Cluster is ready
 echo "Waiting for Service Cluster ready..."
--- a/ci-operator/step-registry/osd-ccs/fleet-manager/test/osd-ccs-fleet-manager-test-commands.sh
+++ b/ci-operator/step-registry/osd-ccs/fleet-manager/test/osd-ccs-fleet-manager-test-commands.sh
@@ -217,7 +217,7 @@ function test_monitoring_disabled ()
 function test_labels() 
 {
   TEST_PASSED=true
-  sc_cluster_id=$(cat "${ARTIFACT_DIR}"/osd-fm-sc-id)
+  sc_cluster_id=$(cat "${SHARED_DIR}"/osd-fm-sc-id)
   mc_cluster_id=$(cat "${ARTIFACT_DIR}"/osd-fm-mc-id)
 
   #Set up region
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-stable.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-stable.yaml
@@ -101,6 +101,7 @@ tests:
       TAG_VERSION: '@4.12'
       TEST_FILTERS: ~ChkUpgrade&;~NonPreRelease&;~Serial&;~Disruptive&;~DisconnectedOnly&;~HyperShiftMGMT&;~MicroShiftOnly&;~NonHyperShiftHOST&;ROSA&
     test:
+    - ref: osd-ccs-fleet-manager-test
     - chain: openshift-e2e-test-hypershift-qe
     workflow: osd-ccs-fleet-manager
 - as: aws-rosa-sts-hypershift-guest-integration-critical-f2


ci-operator/step-registry/osd-ccs/fleet-manager/osd-ccs-fleet-manager-workflow.yaml
```diff
@@ -13,8 +13,9 @@ workflow:
     - ref: osd-ccs-fleet-manager-sc-provision
     - chain: rosa-aws-sts-provision-byo-vpc
     - ref: osd-ccs-conf-idp-htpasswd-multi-users
+    - ref: osd-ccs-fleet-manager-test
```
jianping-shu, line 16: Suggest to put "osd-ccs-fleet-manager-test" right after "osd-ccs-fleet-manager-sc-provision".
Because if any FM configuration test case failed and exit 1, the provisioning steps of hosted cluster is not necessary and can be avoided.

ci-operator/step-registry/osd-ccs/fleet-manager/test/OWNERS
```diff
@@ -0,0 +1,10 @@
+approvers:
+- yasun1
+- jianping-shu
+- heliubj18
+- LiangquanLi930
+reviewers:
+- yasun1
+- jianping-shu
+- heliubj18
+- LiangquanLi930
```
jianping-shu, line 10: Pls. add yourself as reviewer/approver in every owner kind of file.

ci-operator/step-registry/osd-ccs/fleet-manager/test/osd-ccs-fleet-manager-test-commands.sh
```diff
@@ -0,0 +1,357 @@
...
+###### end of disable workload monitoring tests (OCP-60338) ######
+
+##################################################################
+
+###### Sector predicates to support multiple sectors by labels tests (OCP-63998) ######
+
+sc_cluster_id=$(cat "${SHARED_DIR}"/osd-fm-sc-id)
+mc_cluster_id=$(cat "${SHARED_DIR}"/osd-fm-mc-id)
+
+INITIAL_MC_COUNT=$(ocm get /api/osd_fleet_mgmt/v1/management_clusters  --parameter search="region is 'ap-northeast-1'" | jq -r .total)
```
jianping-shu, line 193: Maybe we can define OSDFM_REGION=${LEASED_RESOURCE} like others, in theory the region can be customized in the future.

jianping-shu:
@yasun1 @heliubj18 PTAL when you got time, thanks

ci-operator/step-registry/osd-ccs/fleet-manager/osd-ccs-fleet-manager-workflow.yaml
pawelpaszki: done

ci-operator/step-registry/osd-ccs/fleet-manager/test/osd-ccs-fleet-manager-test-commands.sh
pawelpaszki: done

ci-operator/step-registry/osd-ccs/fleet-manager/sc/provision/osd-ccs-fleet-manager-sc-provision-commands.sh
```diff
@@ -65,15 +65,25 @@ fi
...
+sc_ocm_cluster_id=$(ocm get /api/osd_fleet_mgmt/v1/service_clusters/$sc_cluster_id | jq -r .cluster_management_reference.cluster_id)
+echo "Save kubeconfig for Service Cluster:${sc_ocm_cluster_id}"
+ocm get /api/clusters_mgmt/v1/clusters/${sc_ocm_cluster_id}/credentials | jq -r .kubeconfig > "${SHARED_DIR}/hs-sc.kubeconfig"
+
+# Save MC kubeconfig and cluster info
+mc_ocm_cluster=$(ocm get /api/osd_fleet_mgmt/v1/management_clusters -p search="parent.id='${sc_cluster_id}' and status is 'ready'" -p size=1)
+mc_ocm_cluster_id=$(echo $mc_ocm_cluster | jq -r '.items[0].cluster_management_reference.cluster_id')
+mc_cluster_id=$(echo $mc_ocm_cluster | jq -r '.items[0].id')
+echo "Management Cluster fm id:${mc_cluster_id}"
+echo "${mc_cluster_id}" >> "${SHARED_DIR}/osd-fm-mc-id"
```
yasun1, line 85: Here, using `>` should be better than `>>`.  Besides, I assume putting the id in the exposed folder "${ARTIFACT_DIR}` could be better.

ci-operator/step-registry/osd-ccs/fleet-manager/osd-ccs-fleet-manager-workflow.yaml
```diff
@@ -11,10 +11,11 @@ workflow:
       ROSA_LOGIN_ENV: "integration"
     pre:
     - ref: osd-ccs-fleet-manager-sc-provision
+    - ref: osd-ccs-fleet-manager-test
```
yasun1, line 14: It is acceptable that the test step includes the validation, but it is not good to do testing in the workflow(e.g.. add/remove labels). It is better to put the test into the `test` step so that the workflow is not blocked by a testing scenario.

ci-operator/step-registry/osd-ccs/fleet-manager/sc/provision/osd-ccs-fleet-manager-sc-provision-commands.sh
pawelpaszki: thanks. I will change it. does it also apply to kubeconfig location?

ci-operator/step-registry/osd-ccs/fleet-manager/osd-ccs-fleet-manager-workflow.yaml
jianping-shu: @yasun1 Yes. You are right! But the test step is for hosted cluster actually. There won't be too many FM test cases like this and I think the lightweight solution is to use a ref to group a few FM cases here. How about change its name to osd-ccs-fleet-manager-healthcheck?

ci-operator/step-registry/osd-ccs/fleet-manager/sc/provision/osd-ccs-fleet-manager-sc-provision-commands.sh
yasun1: No, kubeconfig is a credential, we should put it in the unexposed folder `${SHARED_DIR}`.

ci-operator/step-registry/osd-ccs/fleet-manager/osd-ccs-fleet-manager-workflow.yaml
pawelpaszki: @yasun1 - I have changed that. please let me know if that's what you had in mind. @jianping-shu - can you take another look at my recent changes as well please?

yasun1:
@pawelpaszki I realized that you also changed the folder for sc_cluster_id. That could be incorrect. `${ARTIFACT_DIR}` is the exposed folder where we can see the information in the artifact of the prow ci job, but this folder will not be shared across the whole workflow. The `sc_cluster_id` is used for the deprovision in the post step, it should be under the shared folder `${SHARED_DIR}`. The reason why I ask you to put the `mc_cluster_id` is that I think there is no next step to use the `mc_cluster_id`, and we can easily get it from `${ARTIFACT_DIR}` during debug.

ci-operator/step-registry/osd-ccs/fleet-manager/osd-ccs-fleet-manager-workflow.yaml
```diff
@@ -11,9 +11,10 @@ workflow:
       ROSA_LOGIN_ENV: "integration"
     pre:
     - ref: osd-ccs-fleet-manager-sc-provision
-    - ref: osd-ccs-fleet-manager-test
     - chain: rosa-aws-sts-provision-byo-vpc
     - ref: osd-ccs-conf-idp-htpasswd-multi-users
+    test:
+    - ref: osd-ccs-fleet-manager-test
```
jianping-shu, line 17: I was thinking the test ref shall be in ci job 'aws-ocm-osd-ccs-fleet-manager-f7' instead of workflow.
Not sure if it is ok/appropriate to put it here.
See https://steps.ci.openshift.org/workflow/generic-claim
@yasun1 @LiangquanLi930 Do you think that? Thx

yasun1: +1

pawelpaszki:
@yasun1 , sorry for confusion about my changes. so to summarise, I need to to the following to address the outstanding comments:
- reference the testing step as per this comment: https://github.com/openshift/release/pull/42734#discussion_r1311008557
- revert the `sc_cluster_id` location to `${SHARED_DIR}`

does that sound ok to you?

yasun1:
@pawelpaszki Perfect to me. Thanks very much.

pawelpaszki:
@liangxia - can you approve please?

PR #42759, opened 2023-08-28
Update to golang 1.19 for devfile registry operator

michael-valdron:
# Description

Devfile registry operator will be using golang 1.19 after pending changes: https://github.com/devfile/registry-operator/pull/45

blocking devfile/api#881, devfile/api#1106


johnmcollier (approved):
/approve
/lgtm

PR #42055, opened 2023-08-07
Add certificate rotation test for AWS

vrutkovs:
Similar to baremetal-based cert rotation tests this tests that AWS cluster can recover after being suspended for n days (default: 30) in order to ensure cloud provider features are still available.

Includes https://github.com/openshift/release/pull/42773

PR #42773, opened 2023-08-29
aws-provision-bastionhost: fetch vpcID and subnet from working cluster

vrutkovs:
Adapts aws-provision-bastionhost step so that it could use values from installed cluster and could be used as ssh bastion

### DIFF
--- a/ci-operator/step-registry/aws/provision/bastionhost/aws-provision-bastionhost-commands.sh
+++ b/ci-operator/step-registry/aws/provision/bastionhost/aws-provision-bastionhost-commands.sh
@@ -24,14 +24,24 @@ if [[ "${CLUSTER_TYPE:-}" =~ ^aws-s?c2s$ ]]; then
   REGION=$(jq -r ".\"${LEASED_RESOURCE}\".source_region" "${CLUSTER_PROFILE_DIR}/shift_project_setting.json")
 fi
 
+CLUSTER_NAME="${NAMESPACE}-${UNIQUE_HASH}"
 # 1. get vpc id and public subnet
-VpcId=$(cat "${SHARED_DIR}/vpc_id")
+if [[ ! -f "${SHARED_DIR}/vpc_id" && ! -f "${SHARED_DIR}/public_subnet_ids" ]]; then
+  if [[ -e ${SHARED_DIR}/metadata.json ]]; then
+    # for OCP
+    echo "Reading infra id from file metadata.json"
+    infra_id=$(jq -r '.infraID' ${SHARED_DIR}/metadata.json)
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/aws/provision/bastionhost/aws-provision-bastionhost-commands.sh
```diff
@@ -24,14 +24,24 @@ if [[ "${CLUSTER_TYPE:-}" =~ ^aws-s?c2s$ ]]; then
   REGION=$(jq -r ".\"${LEASED_RESOURCE}\".source_region" "${CLUSTER_PROFILE_DIR}/shift_project_setting.json")
 fi
 
+CLUSTER_NAME="${NAMESPACE}-${UNIQUE_HASH}"
 # 1. get vpc id and public subnet
-VpcId=$(cat "${SHARED_DIR}/vpc_id")
+if [[ ! -f "${SHARED_DIR}/vpc_id" && ! -f "${SHARED_DIR}/public_subnet_ids" ]]; then
+  if [[ -e ${SHARED_DIR}/metadata.json ]]; then
```
jianlinliu, line 30: Once that, the vpc would fail to be deleted  on ipi-deprovision step, because the bastion become a dependency of the vpc.

vrutkovs: Correct, you'd need to add
```
    - ref: aws-deprovision-stacks
    - ref: aws-deprovision-s3buckets
```
before deprovision steps first

jianlinliu: for QE's existing provision chain where need a bastion, we always set customer vpc as a pre-condition, so that is not a problem for us. I meat, if you applied the step to a common IPI install, then need to do some adjustment for the ipi deprovsion chain.

vrutkovs: Yeah, I'm using additional steps in https://github.com/openshift/release/pull/42055, but I don't think its worth updating common ipi-deprovision chain for this.

I suppose just a warning in bastion ref should suffice?

jianlinliu, line 30: How about removing the check of metadata.json, so that the step can failed earlier.

vrutkovs: good idea, created https://github.com/openshift/release/pull/43314 for this

PR #42778, opened 2023-08-29
OCPQE-16287: Alter cron for baremetal ec jobs 

dis016:
[OCPQE-16287](https://issues.redhat.com//browse/OCPQE-16287):  Alter cron for baremetal ec jobs to keep resources available for nightly jobs.

/cc @liangxia @jhou1 @pruan-rht

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__arm64-ec.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__arm64-ec.yaml
@@ -1558,7 +1558,7 @@ tests:
     workflow: cucushift-installer-rehearse-baremetalds-ipi-ovn-ipv4
 - as: baremetal-pxe-compact-agent-ipv4-static-connected-f14
   cluster: build05
-  cron: 30 14 11,26 * *
+  cron: 47 15 4,19 * *
   steps:
     cluster_profile: equinix-ocp-metal-qe
     dependencies:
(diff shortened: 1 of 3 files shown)


pruan-rht:
@dis016 how do you determine the new launch frequency for the jobs?  Did you do it manually or use [this](https://github.com/openshift/release/blob/master/ci-operator/config/openshift/openshift-tests-private/tools/generate-cron-entry.sh) script

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__arm64-ec.yaml
```diff
@@ -1579,7 +1579,7 @@ tests:
     workflow: baremetal-lab-agent-install-pxe
 - as: baremetal-pxe-ha-agent-ipv4-static-connected-f14
   cluster: build05
-  cron: 30 14 5,19 * *
+  cron: 30 14 11,26 * *
```
liangxia, line 1582: This one and the above one are running at the same date/time.

dis016:
> 

@pruan-rht i had manually altered the cron, as those ec jobs are disturbing the nightly

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__arm64-ec.yaml
dis016: @liangxia i had observed other baremetal jobs, and looks 11,26 is having less jobs. so used it twice for ec jobs.

dis016:
@pruan-rht @liangxia @jhou1 updated the values as per generate-cron-entry.sh script

jhou1:
/lgtm

Thanks @dis016

PR #42783, opened 2023-08-29
[oko] add hold-the-node post step to install_yamls deploy job

viroel:
Adds a new hold the node step to install_yamls deploy job since it is not defined as workflow in step-registry.


fmount (approved):

viroel:
/pj-rehearse ack

this step is already running in all other workflows, and is a noop if the specific string is not added to the PR message.

PR #42816, opened 2023-08-30
Add not @workloads tag

kasturinarra:
Adding `not @workloads` to make sure that test cases which has this tag does not get picked up when baselinecapability is set to None


kasturinarra:
@liangxia could you please help review? thanks !!

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__arm64-nightly.yaml
```diff
@@ -116,7 +116,8 @@ tests:
       COMPUTE_NODE_TYPE: m6g.xlarge
       CONTROL_PLANE_INSTANCE_TYPE: m6g.xlarge
       E2E_RUN_TAGS: '@arm64 and @aws-ipi and @network-ovnkubernetes and not @fips
-        and not @console and not @storage and @destructive'
+        and not @console and not @storage and not @@baseLineCapabilitiesSetNone and
```
liangxia, line 119: Bad format of `@@`

PR #42839, opened 2023-08-30
MULTIARCH-3741 - Updating the 4.10 through 4.14 periodic cron configs

tvardema:
Updating the cron definitions for each of the ppc64le and s390x jobs to alleviate potential resource issues, and have a more predictable periodic "calendar."

### DIFF
--- a/ci-operator/config/openshift/multiarch/openshift-multiarch-master__nightly-4.10.yaml
+++ b/ci-operator/config/openshift/multiarch/openshift-multiarch-master__nightly-4.10.yaml
@@ -220,7 +220,7 @@ tests:
     workflow: openshift-e2e-aws-techpreview-serial
 - as: ocp-e2e-remote-libvirt-s390x
   cluster: build01
-  cron: 0 0 * * 1,3,5
+  cron: 0 0 1 12 *
   steps:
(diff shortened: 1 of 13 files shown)


ci-operator/config/openshift/multiarch/openshift-multiarch-master__nightly-4.10.yaml
```diff
@@ -233,7 +233,7 @@ tests:
     workflow: openshift-e2e-libvirt
 - as: ocp-image-ecosystem-remote-libvirt-s390x
   cluster: build01
-  cron: 5 1 7 */12 *
+  cron: 0 0 * * 0
```
Prashanth684, line 236: for future reference - do you want to add a comment on what this translates to ?

tvardema: Yes, I do.  I'll update with human-readable descriptions in comments.  Good call.

tvardema: Unfortunately the comments are removed when running `make update`.  I think we'll have to go without :(

PR #42878, opened 2023-08-31
Update jobs to use go-1.19

divyansh42:
Since source to image upstream [repo](https://github.com/openshift/source-to-image) is updated to go `1.19` changing the same in the CI jobs.


divyansh42:
@coreydaley could you please add `rehearsals-ack` label?

PR #42888, opened 2023-08-31
Fixing iterations on control plan smoke tests

vishnuchalla:
This PR fixes the `< 1 iterations` error in control plane smoke tests.
Kube-burner uses this equation to calculate iterations.
`iterations = ((current_worker_count * pods_per_node) - current_pod_count) / 2)`

So previously we tried to keep the iterations count minimal .i.e  (1 or 2), so we calculated pods per node based on that which was very aggressive leaving no rooms for even 1 to 2 pods per node to spin up within the time our test actually got triggered. 
An example case:
```
++ oc get nodes --no-headers -l 'node-role.kubernetes.io/worker=,node-role.kubernetes.io/infra!=,node-role.kubernetes.io/workload!=' --output 'jsonpath={.items[?(@.status.conditions[-1].type=='\''Ready'\'')].status.conditions[-1].type}'
++ xargs
++ wc -w
+ current_worker_count=24
++ oc get nodes '--selector=node-role.kubernetes.io/worker=,node-role.kubernetes.io/infra!=,node-role.kubernetes.io/workload!=' -o name
++ cut -d / -f 2
++ xargs -I '{}' oc get pods '--field-selector=status.phase=Running,spec.nodeName={}' -o name -A
++ wc -l
+ current_pod_count=288
```
And based on these values we calculated
`PODS_PER_NODE=13` i.e
```
pods_per_node=$(( (2 + current_pod_count) / current_worker_count + 1 ))
pods_per_node= (2 + 288)/24 + 1 = 13
```
With this we should have achieved 
`iterations = ((24 * 13) - 288) / 2)` ie. iterations=12, but looks like with in that short span, between our calculation and test trigger, the cluster had 24 pods spinned up which led the numerator to go `<0`. 

So in order to fix this, now we have added `+10` at the end of the `--pods-per-node` we get through our calculation, leaving extra room for 10 pods in each node to get created in even before our test triggers, which is ideally not possible to occur in reality. So we should start seeing a positive number on the iterations and our tests should start running.


vishnuchalla:
/pj-rehearse periodic-ci-openshift-qe-ocp-qe-perfscale-ci-main-gcp-4.13-ocp-qe-perfscale-ci-tests-kube-burner

afcollins (approved):

jtaleric:
I am a bit concerned with the time this smoke-test takes. 

`Step XXXXXX-perfscale-ci-tests-kube-burner-XXXXXXXXXXXX-kube-burner-smoke succeeded after 23m1s.` 

Seems like a awfully long smoke test... and for large scale I can only assume this will take longer to complete.

vishnuchalla:
/pj-rehearse periodic-ci-openshift-qe-ocp-qe-perfscale-ci-main-gcp-4.13-ocp-qe-perfscale-ci-tests-kube-burner

jtaleric:
/pj-rehearse periodic-ci-openshift-qe-ocp-qe-perfscale-ci-main-aws-4.14-ocp-qe-perfscale-aws-ci-tests-medium-scale-kube-burner

afcollins:
While we are touching these steps, we would also benefit from lower timeouts on each kube-burner job. I see some GCP runs that (correctly) timeout because one of the nodes in the cluster is in bad shape, but we will know that way sooner than 2h.

I think 15m is a very conservative amount of time to allow a tiny number of iterations to complete. We may need to model smoke test runtimes for larger clusters.

vishnuchalla:
> While we are touching these steps, we would also benefit from lower timeouts on each kube-burner job. I see some GCP runs that (correctly) timeout because one of the nodes in the cluster is in bad shape, but we will know that way sooner than 2h.
> 
> I think 15m is a very conservative amount of time to allow a tiny number of iterations to complete. We may need to model smoke test runtimes for larger clusters.
> 
> From the medium run triggered here, the longest running job was "Job cluster-density-v2 took 3m32s". So I think we could even go down to 10m timeout.

Acknowledged and changed as suggested.  And also keeping the smoke tests only for the tests that are present in our workload chain.

vishnuchalla:
/pj-rehearse periodic-ci-openshift-qe-ocp-qe-perfscale-ci-main-aws-4.14-ocp-qe-perfscale-aws-ci-tests-medium-scale-kube-burner
/pj-rehearse periodic-ci-openshift-qe-ocp-qe-perfscale-ci-main-gcp-4.13-ocp-qe-perfscale-ci-tests-kube-burner

vishnuchalla:
/pj-rehearse periodic-ci-openshift-qe-ocp-qe-perfscale-ci-main-gcp-4.13-ocp-qe-perfscale-ci-tests-kube-burner

ci-operator/step-registry/openshift-qe/kube-burner-smoke/openshift-qe-kube-burner-smoke-commands.sh
```diff
@@ -14,28 +14,20 @@ source ./venv_qe/bin/activate
...
 }
 
 set_pods_per_node
-export EXTRA_FLAGS="--pods-per-node=$PODS_PER_NODE"
-export WORKLOAD=node-density-cni
-./run.sh
-
-set_pods_per_node
-export EXTRA_FLAGS="--pods-per-node=$PODS_PER_NODE --pod-ready-threshold=180000ms"
+export EXTRA_FLAGS="--pods-per-node=$PODS_PER_NODE --pod-ready-threshold=180000ms --timeout=10m"
```
afcollins, line 25: thanks!

PR #42788, opened 2023-08-29
Rename kube-burner-smoke to control-plane-smoke

krishvoor:
This PR allows users to understand the nature of workloads to run,
We are suffixing it with control-plane-tests to identify the nature of jobs easily

PR #42933, opened 2023-09-01
Creating a common script for cluster density and node density* workloads

shahsahil264:
@paigerube14

### DIFF
--- a/ci-operator/jobs/openshift/release/openshift-release-infra-periodics.yaml
+++ b/ci-operator/jobs/openshift/release/openshift-release-infra-periodics.yaml
@@ -1513,7 +1513,6 @@ periodics:
     job-env: aws
     job-test: e2e
     job-type: upgrade
-    tag: "1"
   name: release-openshift-okd-fcos-installer-e2e-aws-upgrade-from-fcos-stable
   spec:
     containers:
@@ -1530,10 +1529,6 @@ periodics:
       command:
       - ci-operator
       env:
-      - name: RELEASE_IMAGE_LATEST
-      - name: CLUSTER_VARIANT
-      - name: BRANCH
-        value: "4.10"
       - name: CLUSTER_TYPE
         value: aws
       - name: UNRESOLVED_CONFIG
@@ -1553,9 +1548,6 @@ periodics:
               requests:
                 cpu: 100m
                 memory: 200Mi
-          tag_specification:
-            name: "$(BRANCH)"
-            namespace: origin
           tests:
           - as: e2e-$(CLUSTER_TYPE)
             commands: TEST_SUITE=openshift/conformance/parallel run-tests
@@ -1621,7 +1613,6 @@ periodics:
     job-env: aws
     job-test: e2e
     job-type: upgrade
-    tag: "1"
   name: release-openshift-okd-scos-installer-e2e-aws-upgrade-from-scos-stable
   spec:
     containers:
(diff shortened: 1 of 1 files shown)


shahsahil264:
/pj-rehearse periodic-ci-openshift-qe-ocp-qe-perfscale-ci-main-ocp-qe-perfscale-ci-tests

shahsahil264:
/pj-rehearse periodic-ci-openshift-qe-ocp-qe-perfscale-ci-main-ocp-qe-perfscale-ci-tests

paigerube14:
After the below PR, we could put the common script into the control-plane folder because all related tests are under that chain/workflow ttps://github.com/openshift/release/pull/42788/files

shahsahil264:
/pj-rehearse periodic-ci-openshift-qe-ocp-qe-perfscale-ci-main-ocp-qe-perfscale-ci-tests

ci-operator/step-registry/openshift-qe/node-density/openshift-qe-node-density-ref.yaml
```diff
@@ -26,5 +26,9 @@ ref:
     default: "kube-burner-ocp.out"
     documentation: |-
       File in which we will parse to find uuid, start and end time
+  - name: GC
+    default: "true"
+    documentation: |-
+      Default is true, which means clean up the pod/resource that kube-burner ocp created, you can set it to false to keep the resouce
```
krishvoor, line 32: The default is true per [kube-burner docs](https://cloud-bulldozer.github.io/kube-burner/v1.7.6/ocp/#openshift-wrapper) so I do not think, this would be needed.

Incase a user wants to supply extra flags, this can be done via, 
`EXTRA_FLAGS` parameters
https://steps.ci.openshift.org/chain/openshift-qe-ocp-qe-perfscale-ci-tests#environment

ci-operator/step-registry/openshift-qe/node-density-heavy/openshift-qe-node-density-heavy-ref.yaml
```diff
@@ -20,7 +20,11 @@ ref:
   - name: PODS_PER_NODE
     default: "100"
     documentation: |-
-      Define the maximum number of pods to deploy on each labeled node.  
+      Define the maximum number of pods to deploy on each labeled node.
+  - name: GC
+    default: "true"
+    documentation: |-
+      Default is true, which means clean up the pod/resource that kube-burner ocp created, you can set it to false to keep the resouce
```
krishvoor, line 27: Same as below.

krishvoor (changes requested):
minor queries/suggestions.

krishvoor:
/hold
This should wait until the https://github.com/openshift/release/pull/42788 gets merged.

ci-operator/step-registry/openshift-qe/node-density/openshift-qe-node-density-ref.yaml
paigerube14: The e2e-benchmarking run script sets this for us so we don't need to use EXTRA_FLAGS explicitly. Having this as an environment variable will allow us at the config level to set to cleanup or not. For the loaded-upgrade tests we have we set this to false to keep the load on the cluster till after the upgrade  

https://github.com/cloud-bulldozer/e2e-benchmarking/blob/c1196a3cf0e2b34952d286fb8f33a94f19755f7b/workloads/kube-burner-ocp-wrapper/run.sh#L12

paigerube14: The one thing I'm not sure of is if we need to set a default parameter in this file (I'm not sure if its required) but I feel its helpful to know what the value is without searching into e2e-benchmarkin/kube-burner

PR #42943, opened 2023-09-03
Test OCP-54741 automation fix for BM Kubevirt

jechen0648:


jechen0648:
have some mistake, close it, will open a new one

PR #42960, opened 2023-09-04
Remove priority field from test names for 4.13

liangxia:
Priority field is no longed needed in test names. It has been remove for other version, except 4.13 jobs.


jianlinliu:
@liangxia I saw there are still some ci job name including priority field in openshift-openshift-tests-private-release-4.13__arm64-nightly.yaml and openshift-openshift-tests-private-release-4.13__amd64-nightly.yaml, will we have some following PR to remove them?

liangxia:
Yes, I'm still working on it.  The `reporter_config` part make it harder to track when there are renames.

PR #42969, opened 2023-09-04
steps, cnv: Select CNV version from OCP version

qinqon:
To make the install cnv step more generic this change add a fallback in case the CNV_PREREASE_VERSION is not set that will map ocp version to it's cnv version.

### DIFF
--- a/ci-operator/step-registry/hypershift/kubevirt/install/hypershift-kubevirt-install-commands.sh
+++ b/ci-operator/step-registry/hypershift/kubevirt/install/hypershift-kubevirt-install-commands.sh
@@ -2,7 +2,11 @@
 
 set -ex
 
-CNV_PRERELEASE_VERSION=${CNV_PRERELEASE_VERSION:-}
+function ocp_version() {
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/hypershift/kubevirt/install/hypershift-kubevirt-install-commands.sh
```diff
@@ -2,7 +2,19 @@
 
 set -ex
 
-CNV_PRERELEASE_VERSION=${CNV_PRERELEASE_VERSION:-}
+function calculate_cnv_version() {
+    local ocp_version=$(oc get clusterversion --no-headers | awk '{ print $2 }' | awk -F "." '{print $1"."$2}')
```
orenc1, line 6: i think something like this will be more clean/elegant than parsing cli output with awk:
```
oc get clusterversion version -o jsonpath='{.status.desired.version}' | awk -F "." '{print $1"."$2}'
```

orenc1:
although most of the CI lanes are failing, the CNV installation part works with a version that matches the OCP version it runs on.

PR #42977, opened 2023-09-04
test: add additional check for osd fm deprovision job

pawelpaszki:
## WHAT
test: add additional check for osd fm deprovision job


pawelpaszki:
ping @jianping-shu

PR #42994, opened 2023-09-05
remove azure resource group ci jobs from 4.12

jianlinliu:
resource group feature is already well covered by [azure stack](https://github.com/openshift/release/blob/master/ci-operator/step-registry/ipi/conf/azurestack/creds/ipi-conf-azurestack-creds-commands.sh#L33) and [azure sts jobs](https://github.com/openshift/release/blob/master/ci-operator/step-registry/cucushift/installer/rehearse/azure/ipi/cco-manual-workload-identity/provision/cucushift-installer-rehearse-azure-ipi-cco-manual-workload-identity-provision-chain.yaml#L7) since 4.12, so remove them.


jianlinliu:
cc @jinyunma @liangxia to review

PR #43018, opened 2023-09-05
MGMT-15505: assisted-service 2.9 config

adriengentil:


ci-operator/config/openshift/assisted-service/openshift-assisted-service-release-ocm-2.9.yaml
```diff
@@ -0,0 +1,759 @@
...
+- as: mirror-vcsref-image
+  postsubmit: true
+  steps:
+    dependencies:
+      SOURCE_IMAGE_REF: assisted-service
+    env:
+      RELEASE_TAG_PREFIX: ocm-2.9
+    test:
+    - ref: assisted-baremetal-images-publish
+- as: push-pr-image
```
osherdp, line 195: not sure if we need it. I think it was mainly a hack back then to allow testing the publishing from a PR (when the code still needed more debugging)

ci-operator/config/openshift/assisted-service/openshift-assisted-service-release-ocm-2.9.yaml
```diff
@@ -0,0 +1,759 @@
...
+- as: e2e-ai-operator-ztp-hypershift-zero-nodes-periodic
+  cron: 00 02 */1 * 1-4
+  steps:
+    cluster_profile: packet-assisted
+    dependencies:
+      ASSISTED_OPENSHIFT_INSTALL_RELEASE_IMAGE: release:latest
+      HYPERSHIFT_IMAGE: hypershift-operator
+      OPENSHIFT_INSTALL_RELEASE_IMAGE: release:latest
+    workflow: assisted-baremetal-operator-hypershift
+- as: e2e-metal-assisted-kube-api-net-suite
```
osherdp, line 679: I think we don't need the test-infra based tests. it will be easier to only maintain one kind of tests and I don't think we're missing coverage because of it

osherdp (changes requested):

ci-operator/config/openshift/assisted-service/openshift-assisted-service-release-ocm-2.9.yaml
adriengentil: those one are kube-api tests, don't we want to keep them?

ci-operator/config/openshift/assisted-service/openshift-assisted-service-release-ocm-2.9.yaml
adriengentil: we still have it in 2.8/2.7

ci-operator/config/openshift/assisted-service/openshift-assisted-service-release-ocm-2.9.yaml
osherdp: in my opinion they require more maintenance than helping us
in a world where ztp and kube-api tests are equally simulating and have the same coverage then I would instead prefer to deprecate the ztp ones, but currently ztp has more coverage and simulates the user case much better

adriengentil: ok then I'll clean them up in other branches as well

osherdp (approved):

PR #43031, opened 2023-09-05
telco5g: Run 4.11 telco5g jobs on weekends only

sshnaidm:
[CNF-9596](https://issues.redhat.com//browse/CNF-9596)

We don't need 4.11 so often and don't have resources for its running on baremetal every day. Let's reduce its frequency to twice a week in Sat and Sun.


deepsm007:
/assign @stbenjam 
Needs an approval

PR #43045, opened 2023-09-06
increase test frequency from f28 to f14 on 4.14 eus ci

jiajliu:
Current test frequency is low for 4.14 eus test, which is not enough for us to check latest eus result from ci, so increase it to f14.

@jianlinliu @shellyyang1989  pls help review when you have time, thanks!

cc @wzheng1


jianlinliu:
cc @liangxia to review.

PR #43084, opened 2023-09-06
Add environment variable to CI config

brucejcong:


sjenning:
/hold
Why do we want to disable PKI on the AWS flow?

brucejcong:
Hey @sjenning from my understanding, this is to address this issue here: https://github.com/openshift/hypershift/issues/2455. According to Hide after convos with Cesar, this would help in extending the CI tests. 

From my understanding, this should catch if any PKI contracts has been broken on patch updates, but let me know if there's anything I'm missing! This PR goes with it: https://github.com/openshift/hypershift/pull/2989

csrwng:
@brucejcong The thinking was to disable PKI reconciliation on upgrade for branches != main

csrwng:
@brucejcong you'll need to add the env var to https://github.com/openshift/release/blob/5a157e42551ae816941865abaed833468a6021ca/ci-operator/step-registry/hypershift/aws/run-e2e/hypershift-aws-run-e2e-chain.yaml#L24
and default it to "false"

csrwng:
/hold cancel
This is to test that no new PKI is introduced in previous minor releases.

PR #43104, opened 2023-09-07
manila-operator: Make kuttl job and tempest job required

kajinamit:
These are fundamental tests and should pass when merging any PRs.


Sandeepyadav93:
/pj-rehearse ack 

Job history is good.

PR #43141, opened 2023-09-08
app.ci okd-admins: add release-controller-modify role

vrutkovs:
This role allows modifying imagestreams and enables running custom release controllers for experiments.

Previous ClusterRole - `registry-editor` - is no longer available

### DIFF
--- a/clusters/app.ci/okd/admin_okd.yaml
+++ b/clusters/app.ci/okd/admin_okd.yaml
@@ -42,16 +42,16 @@ items:
 - kind: RoleBinding
   apiVersion: rbac.authorization.k8s.io/v1
   metadata:
-    name: okd-registry-editor
+    name: okd-release-controller-modify
     namespace: origin
   subjects:
   - kind: Group
     apiGroup: rbac.authorization.k8s.io
     name: okd-admins
   roleRef:
-    kind: ClusterRole
+    kind: Role
     apiGroup: rbac.authorization.k8s.io
-    name: registry-editor
+    name: release-controller-modify
 - kind: RoleBinding
   apiVersion: rbac.authorization.k8s.io/v1
   metadata:


clusters/app.ci/okd/admin_okd.yaml
```diff
@@ -49,9 +49,10 @@ items:
     apiGroup: rbac.authorization.k8s.io
     name: okd-admins
   roleRef:
-    kind: ClusterRole
+    kind: Role
     apiGroup: rbac.authorization.k8s.io
-    name: registry-editor
```
bradmwilliams, line 54: @vrutkovs When did/does the `registry-editor` ClusterRole get removed?  I've checked 2 clusters and both still have the it defined...
``` shell
$ oc --context app.ci get clusterversion
version   4.13.11   True        False         4d2h    Cluster version is 4.13.11
$ oc --context app.ci get clusterroles registry-editor
NAME              CREATED AT
registry-editor   2020-04-16T19:19:25Z
```
``` shell
$ oc --context dpcr get clusterversion
version   4.14.0-ec.4   True        False         38d     Cluster version is 4.14.0-ec.4
$ oc --context dpcr get clusterroles registry-editor
NAME              CREATED AT
registry-editor   2020-10-28T21:32:24Z
```
Regardless, I'm fine with this change, but you'll need to resolve the test failures before it can merge.

bradmwilliams (changes requested):

clusters/app.ci/okd/admin_okd.yaml
vrutkovs: Oh, it must be some standard role then. I looked up in other `release` role and it was not used.

In any case, seems `release-controller-modify` role includes permissions from `registry-editor`

bradmwilliams: Yeah, it's probably better to use `release-controller-modify` anyway.

bradmwilliams (approved):

PR #43158, opened 2023-09-08
USHIFT-1674: Add 4.14 periodics to testgrid

pacevedom:


bradmwilliams:
/cc @openshift/openshift-release-oversight 
We're in that grey area of what releases TRT is paying attention to.  So, I'd prefer their blessing on adding new jobs...

neisw:
I think we will be ok with new jobs, might wait until next week to let things settle after branch cut.  Should these jobs be added for 4.15 as well?

pacevedom:
> I think we will be ok with new jobs, might wait until next week to let things settle after branch cut. Should these jobs be added for 4.15 as well?

Not really needed for 4.15/main, right now we are good with 4.14 only. Thanks!

bradmwilliams (approved):

PR #43160, opened 2023-09-08
node-density-cni update

jtaleric:
Have node-density-cni match similar to node-density-heavy.

### DIFF
--- a/ci-operator/step-registry/openshift-qe/node-density-cni/openshift-qe-node-density-cni-ref.yaml
+++ b/ci-operator/step-registry/openshift-qe/node-density-cni/openshift-qe-node-density-cni-ref.yaml
@@ -15,9 +15,17 @@ ref:
     documentation: |-
       File in which we will parse to find uuid, start and end time
   - name: PODS_PER_NODE
-    default: "200"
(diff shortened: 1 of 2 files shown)


ci-operator/step-registry/openshift-qe/node-density-cni/openshift-qe-node-density-cni-ref.yaml
```diff
@@ -15,9 +15,17 @@ ref:
     documentation: |-
       File in which we will parse to find uuid, start and end time
   - name: PODS_PER_NODE
-    default: "200"
+    default: "100"
```
paigerube14, line 18: do we need to lower this?

jtaleric: @paigerube14 this was to mimic what we are doing with node-density heavy --- Last I heard we wanted those things to mimic each other. https://github.com/openshift/release/blob/master/ci-operator/step-registry/openshift-qe/node-density-heavy/openshift-qe-node-density-heavy-ref.yaml#L21

PR #43167, opened 2023-09-08
remove osde2e harness test job from avo postsubmit pipeline

ritmun:
There should not be a test harness   postsubmit job for AVO in prow. Removing. 

Team Cicada is implementing app-interface pipelines for this purpose in [SDE-2000](https://issues.redhat.com//browse/SDE-2000)


mjlshen (approved):
/approve

PR #43199, opened 2023-09-11
openstack-test: Kuryr job optional

pierreprinetti:
Kuryr testing signal is currently not reliable.


mandre (commented):
If the job is broken and it prevents merging code then it's fair game to make it optional until it stabilizes.
Do we have an issue where we track kuryr CI problems?

dulek:
> If the job is broken and it prevents merging code then it's fair game to make it optional until it stabilizes. Do we have an issue where we track kuryr CI problems?

https://issues.redhat.com/browse/OCPBUGS-17066 is where I'm tracking it at the moment.

dulek:
/lgtm

We intend to remove Kuryr in 4.15 anyway.

PR #43210, opened 2023-09-11
Include missing openstack-provision-vips-ports step

MaysaMacedo:
The ipi-conf-openstack-noramfs conf should be the same as ipi-conf-openstack with exception of the step ipi-conf-etcd-on-ramfs. This commit adds the openstack-provision-vips-ports step that was missing.


mandre (commented):
Good catch.
/lgtm
/approve

PR #43222, opened 2023-09-12
openshift-loki: Use loki v2.9.0 in operator release-5.8

periklis:
Refs: [LOG-4492](https://issues.redhat.com//browse/LOG-4492)

/cc @xperimental


xperimental:
Does it make sense to split this up into three parts (main, 5.6, 5.7), so that we can merge main & 5.7 earlier than 5.6?

PR #43227, opened 2023-09-12
auto ocp-66839

jiajliu:
Case ocp-66839 automation and integraion. Summary:(refer to [comment1](https://github.com/openshift/release/pull/43227#issuecomment-1720648105) and [comment2](https://github.com/openshift/release/pull/43227#issuecomment-1731028623))

1. integrate aws cloud credential update steps into upgrade-toimage step;
2. add preupgrade and postupgrade steps for ota upgrade cases;
3. auto ocp-66839
4. jq refactor 

[Test result](https://qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/qe-private-deck/pr-logs/pull/openshift_release/43227/rehearse-43227-periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28/1706138674179084288) 

- preupgrade [failed](https://gcsweb-qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/qe-private-deck/pr-logs/pull/openshift_release/43227/rehearse-43227-periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28/1706138674179084288/artifacts/aws-ipi-disconnected-sts-basecap-none-addcap-f28/cucushift-ota-preupgrade/artifacts/ota-test-result.txt) due to known issue https://issues.redhat.com/browse/OCPBUGS-18961
```
Tobe gained CRs with cap annotation: MachineAPI, but expected: MachineAPI and ImageRegistry
```
- upgrade succeed
- postupgrade [succeed](https://gcsweb-qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/qe-private-deck/pr-logs/pull/openshift_release/43227/rehearse-43227-periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28/1706138674179084288/artifacts/aws-ipi-disconnected-sts-basecap-none-addcap-f28/cucushift-ota-postupgrade/build-log.txt) 

@jianlinliu @shellyyang1989 Please help review when you have time, thanks!

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__amd64-nightly-4.14-upgrade-from-stable-4.13.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__amd64-nightly-4.14-upgrade-from-stable-4.13.yaml
@@ -184,7 +184,7 @@ tests:
     test:
     - chain: openshift-upgrade-qe-test-disconnected
     workflow: cucushift-installer-rehearse-aws-ipi-disconnected-private-sno
-- as: aws-ipi-disconnected-sts-basecap-none-addcap-f28
+- as: aws-ipi-disconnected-sts-basecap-none-f28
   cron: 18 22 7 * *
   steps:
     cluster_profile: aws-qe
@@ -194,7 +194,9 @@ tests:
       ENABLE_OTA_TEST: "66839"
       FIPS_ENABLED: "true"
       PERSISTENT_MONITORING: "false"
+      TEST_FILTERS: ~ConnectedOnly&;~HyperShiftMGMT&;~MicroShiftOnly&
       UPGRADE_CCO_MANUAL_MODE: oidc
+      UPGRADE_CHECK_RUN_TAGS: '@upgrade-check and @4.13 and @aws-ipi and @disconnected'
       UPGRADE_PRE_RUN_TAGS: '@upgrade-prepare and @4.13 and @aws-ipi and @disconnected'
     test:
     - chain: openshift-upgrade-qe-test-disconnected
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__amd64-nightly-4.15-upgrade-from-stable-4.14.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__amd64-nightly-4.15-upgrade-from-stable-4.14.yaml
@@ -186,12 +186,12 @@ tests:
     test:
     - chain: openshift-upgrade-qe-test-disconnected
     workflow: cucushift-installer-rehearse-aws-ipi-disconnected-private-sno
-- as: aws-ipi-disconnected-sts-basecap-none-addcap-f28
+- as: aws-ipi-disconnected-sts-basecap-none-f28
   cron: 16 18 12 * *
   steps:
     cluster_profile: aws-qe
     env:
-      ADDITIONAL_ENABLED_CAPABILITIES: CSISnapshot
+      ADDITIONAL_ENABLED_CAPABILITIES: MachineAPI
       BASE_DOMAIN: qe.devcluster.openshift.com
       BASELINE_CAPABILITY_SET: None
       FIPS_ENABLED: "true"
--- a/ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
+++ b/ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
@@ -46,6 +46,7 @@ function extract_ccoctl(){
     local retry=5 
     tmp_ccoctl="/tmp/upgtool"
     mkdir -p ${tmp_ccoctl}
+    export PATH=/tmp:${PATH}
     cco_image=$(oc adm release info --image-for='cloud-credential-operator' ${TARGET} -a "${CLUSTER_PROFILE_DIR}/pull-secret") || return 1
     while ! (env "NO_PROXY=*" "no_proxy=*" oc image extract $cco_image --path="/usr/bin/ccoctl:${tmp_ccoctl}" -a "${CLUSTER_PROFILE_DIR}/pull-secret");
     do
@@ -88,7 +89,6 @@ function update_cloud_credentials_oidc(){
                 export AWS_SHARED_CREDENTIALS_FILE="${CLUSTER_PROFILE_DIR}/.awscred"
                 infra_name=${NAMESPACE}-${UNIQUE_HASH}
                 oidc_provider=$(head -n1 ${SHARED_DIR}/aws_oidc_provider_arn)
-                export PATH=/tmp:${PATH}
                 extract_ccoctl
                 if ! ccoctl aws create-iam-roles --name="${infra_name}" --region="${LEASED_RESOURCE}" --credentials-requests-dir="${tobeCredsDir}" --identity-provider-arn="${oidc_provider}" --output-dir="${toManifests}"; then
                     echo "Failed to update iam role!" && exit 1
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__arm64-nightly-4.14-upgrade-from-stable-4.13.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__arm64-nightly-4.14-upgrade-from-stable-4.13.yaml
@@ -150,7 +150,7 @@ tests:
     test:
     - chain: openshift-upgrade-qe-test-disconnected
     workflow: cucushift-installer-rehearse-aws-ipi-disconnected-private
-- as: aws-ipi-disconnected-sts-basecap-none-addcap-f28
+- as: aws-ipi-disconnected-sts-basecap-none-f28
   cron: 33 14 30 * *
   steps:
     cluster_profile: aws-qe
@@ -158,14 +158,15 @@ tests:
       OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE: release:arm64-latest
       OPENSHIFT_UPGRADE_RELEASE_IMAGE_OVERRIDE: release:arm64-target
     env:
-      ADDITIONAL_ENABLED_CAPABILITIES: CSISnapshot
       BASE_DOMAIN: qe.devcluster.openshift.com
       BASELINE_CAPABILITY_SET: None
       COMPUTE_NODE_TYPE: m6g.xlarge
       CONTROL_PLANE_INSTANCE_TYPE: m6g.xlarge
+      ENABLE_OTA_TEST: "66839"
       OCP_ARCH: arm64
       PERSISTENT_MONITORING: "false"
       TEST_FILTERS: ~ConnectedOnly&;~HyperShiftMGMT&;~MicroShiftOnly&
+      UPGRADE_CCO_MANUAL_MODE: oidc
       UPGRADE_CHECK_RUN_TAGS: '@arm64 and @upgrade-check and @4.13 and @aws-ipi and
         @disconnected and not @fips'
       UPGRADE_PRE_RUN_TAGS: '@arm64 and @upgrade-prepare and @4.13 and @aws-ipi and
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__arm64-nightly-4.15-upgrade-from-stable-4.14.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__arm64-nightly-4.15-upgrade-from-stable-4.14.yaml
@@ -150,7 +150,7 @@ tests:
     test:
     - chain: openshift-upgrade-qe-test-disconnected
     workflow: cucushift-installer-rehearse-aws-ipi-disconnected-private
-- as: aws-ipi-disconnected-sts-basecap-none-addcap-f28
+- as: aws-ipi-disconnected-sts-basecap-none-f28
   cron: 39 17 28 * *
   steps:
     cluster_profile: aws-qe
@@ -158,7 +158,7 @@ tests:
       OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE: release:arm64-latest
       OPENSHIFT_UPGRADE_RELEASE_IMAGE_OVERRIDE: release:arm64-target
     env:
-      ADDITIONAL_ENABLED_CAPABILITIES: CSISnapshot
+      ADDITIONAL_ENABLED_CAPABILITIES: MachineAPI
       BASE_DOMAIN: qe.devcluster.openshift.com
       BASELINE_CAPABILITY_SET: None
       COMPUTE_NODE_TYPE: m6g.xlarge
--- a/ci-operator/jobs/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14-periodics.yaml
+++ b/ci-operator/jobs/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14-periodics.yaml
@@ -1584,7 +1584,7 @@ periodics:
     ci-operator.openshift.io/variant: amd64-nightly-4.14-upgrade-from-stable-4.13
     ci.openshift.io/generator: prowgen
     pj-rehearse.openshift.io/can-be-rehearsed: "true"
-  name: periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28
+  name: periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-f28
   spec:
     containers:
     - args:
@@ -1594,8 +1594,8 @@ periodics:
       - --oauth-token-path=/usr/local/github-credentials/oauth
       - --report-credentials-file=/etc/report/credentials
       - --secret-dir=/secrets/ci-pull-credentials
-      - --secret-dir=/usr/local/aws-ipi-disconnected-sts-basecap-none-addcap-f28-cluster-profile
-      - --target=aws-ipi-disconnected-sts-basecap-none-addcap-f28
+      - --secret-dir=/usr/local/aws-ipi-disconnected-sts-basecap-none-f28-cluster-profile
+      - --target=aws-ipi-disconnected-sts-basecap-none-f28
       - --variant=amd64-nightly-4.14-upgrade-from-stable-4.13
       command:
       - ci-operator
@@ -1612,7 +1612,7 @@ periodics:
       - mountPath: /secrets/ci-pull-credentials
         name: ci-pull-credentials
         readOnly: true
-      - mountPath: /usr/local/aws-ipi-disconnected-sts-basecap-none-addcap-f28-cluster-profile
+      - mountPath: /usr/local/aws-ipi-disconnected-sts-basecap-none-f28-cluster-profile
         name: cluster-profile
       - mountPath: /secrets/gcs
         name: gcs-credentials
@@ -43482,7 +43482,7 @@ periodics:
     ci-operator.openshift.io/variant: arm64-nightly-4.14-upgrade-from-stable-4.13
     ci.openshift.io/generator: prowgen
     pj-rehearse.openshift.io/can-be-rehearsed: "true"
-  name: periodic-ci-openshift-openshift-tests-private-release-4.14-arm64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28
+  name: periodic-ci-openshift-openshift-tests-private-release-4.14-arm64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-f28
   spec:
     containers:
     - args:
@@ -43492,8 +43492,8 @@ periodics:
       - --oauth-token-path=/usr/local/github-credentials/oauth
       - --report-credentials-file=/etc/report/credentials
       - --secret-dir=/secrets/ci-pull-credentials
-      - --secret-dir=/usr/local/aws-ipi-disconnected-sts-basecap-none-addcap-f28-cluster-profile
-      - --target=aws-ipi-disconnected-sts-basecap-none-addcap-f28
+      - --secret-dir=/usr/local/aws-ipi-disconnected-sts-basecap-none-f28-cluster-profile
+      - --target=aws-ipi-disconnected-sts-basecap-none-f28
       - --variant=arm64-nightly-4.14-upgrade-from-stable-4.13
       command:
       - ci-operator
@@ -43510,7 +43510,7 @@ periodics:
       - mountPath: /secrets/ci-pull-credentials
         name: ci-pull-credentials
         readOnly: true
-      - mountPath: /usr/local/aws-ipi-disconnected-sts-basecap-none-addcap-f28-cluster-profile
+      - mountPath: /usr/local/aws-ipi-disconnected-sts-basecap-none-f28-cluster-profile
         name: cluster-profile
       - mountPath: /secrets/gcs
         name: gcs-credentials
--- a/ci-operator/jobs/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15-periodics.yaml
+++ b/ci-operator/jobs/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15-periodics.yaml
@@ -3938,7 +3938,7 @@ periodics:
     ci-operator.openshift.io/variant: amd64-nightly-4.15-upgrade-from-stable-4.14
     ci.openshift.io/generator: prowgen
     pj-rehearse.openshift.io/can-be-rehearsed: "true"
-  name: periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-nightly-4.15-upgrade-from-stable-4.14-aws-ipi-disconnected-sts-basecap-none-addcap-f28
+  name: periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-nightly-4.15-upgrade-from-stable-4.14-aws-ipi-disconnected-sts-basecap-none-f28
   spec:
     containers:
     - args:
@@ -3948,8 +3948,8 @@ periodics:
       - --oauth-token-path=/usr/local/github-credentials/oauth
       - --report-credentials-file=/etc/report/credentials
       - --secret-dir=/secrets/ci-pull-credentials
-      - --secret-dir=/usr/local/aws-ipi-disconnected-sts-basecap-none-addcap-f28-cluster-profile
-      - --target=aws-ipi-disconnected-sts-basecap-none-addcap-f28
+      - --secret-dir=/usr/local/aws-ipi-disconnected-sts-basecap-none-f28-cluster-profile
+      - --target=aws-ipi-disconnected-sts-basecap-none-f28
       - --variant=amd64-nightly-4.15-upgrade-from-stable-4.14
       command:
       - ci-operator
@@ -3966,7 +3966,7 @@ periodics:
       - mountPath: /secrets/ci-pull-credentials
         name: ci-pull-credentials
         readOnly: true
-      - mountPath: /usr/local/aws-ipi-disconnected-sts-basecap-none-addcap-f28-cluster-profile
+      - mountPath: /usr/local/aws-ipi-disconnected-sts-basecap-none-f28-cluster-profile
         name: cluster-profile
       - mountPath: /secrets/gcs
         name: gcs-credentials
@@ -37961,7 +37961,7 @@ periodics:
     ci-operator.openshift.io/variant: arm64-nightly-4.15-upgrade-from-stable-4.14
     ci.openshift.io/generator: prowgen
     pj-rehearse.openshift.io/can-be-rehearsed: "true"
-  name: periodic-ci-openshift-openshift-tests-private-release-4.15-arm64-nightly-4.15-upgrade-from-stable-4.14-aws-ipi-disconnected-sts-basecap-none-addcap-f28
+  name: periodic-ci-openshift-openshift-tests-private-release-4.15-arm64-nightly-4.15-upgrade-from-stable-4.14-aws-ipi-disconnected-sts-basecap-none-f28
   spec:
     containers:
     - args:
@@ -37971,8 +37971,8 @@ periodics:
       - --oauth-token-path=/usr/local/github-credentials/oauth
       - --report-credentials-file=/etc/report/credentials
       - --secret-dir=/secrets/ci-pull-credentials
-      - --secret-dir=/usr/local/aws-ipi-disconnected-sts-basecap-none-addcap-f28-cluster-profile
-      - --target=aws-ipi-disconnected-sts-basecap-none-addcap-f28
+      - --secret-dir=/usr/local/aws-ipi-disconnected-sts-basecap-none-f28-cluster-profile
+      - --target=aws-ipi-disconnected-sts-basecap-none-f28
       - --variant=arm64-nightly-4.15-upgrade-from-stable-4.14
       command:
       - ci-operator
@@ -37989,7 +37989,7 @@ periodics:
       - mountPath: /secrets/ci-pull-credentials
         name: ci-pull-credentials
         readOnly: true
-      - mountPath: /usr/local/aws-ipi-disconnected-sts-basecap-none-addcap-f28-cluster-profile
+      - mountPath: /usr/local/aws-ipi-disconnected-sts-basecap-none-f28-cluster-profile
         name: cluster-profile
       - mountPath: /secrets/gcs
         name: gcs-credentials
--- a/ci-operator/step-registry/openshift/upgrade/qe/test/disconnected/openshift-upgrade-qe-test-disconnected-chain.yaml
+++ b/ci-operator/step-registry/openshift/upgrade/qe/test/disconnected/openshift-upgrade-qe-test-disconnected-chain.yaml
@@ -2,12 +2,16 @@ chain:
   as: openshift-upgrade-qe-test-disconnected
   steps:
   - ref: idp-htpasswd
+  - ref: cucushift-pre
+  - ref: openshift-extended-upgrade-pre
   - ref: cucushift-upgrade-prepare
   - ref: cucushift-upgrade-mirror-images-by-oc-adm
   - ref: cucushift-upgrade-healthcheck
   - ref: cucushift-ota-preupgrade
   - ref: cucushift-upgrade-toimage
   - ref: cucushift-ota-postupgrade
+  - ref: openshift-extended-upgrade-post
+  - ref: cucushift-upgrade-check
   - ref: cucushift-upgrade-report
   documentation: |-
     The chain executes upgrade-prepare, which creates test data on a disconnected cluster;


jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
The test is still ongoing, cc @jiajliu @shellyyang1989 for a workflow review in advance, and we could have a discussion next week. Here is a brief intro:

- ENABLE_OTA_TEST is used as a main switch, which determine if ota test will be ran during an upgrade job.
- cucushift-ota-preupgrade and cucushift-ota-postupgrade steps are new steps for adding ota test only, will be placed before and after upgrade step.(dropping other unnecessary steps in upgrade chain for test only, will add it back)
- in pre/post upgrade step, ran_ota_test is used as a case switch, which determine if the specified case should be ran or not, and collect test result of each case to output into ota-test-result.txt
- each case will be a func, which called in ran_ota_test, it will not exit, but return 1 once fail, then the failure result can be collected in ota-test-result.txt(considering the workaround pr strongly depends on the case owner, it may block an upgrade job if the owner take leave, so I don't choose exit directly way here)

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
New update:

1. split run_ota_test into run_ota_single_case and run_ota_multi_test to distinguish two kinds of test cases:

- run_ota_single_case, run case which need to be ran on specified profile or may break other test cases
- run_ota_multi_test, all cases included in this func could be ran in one cluster’s upgrade

2. move cloud credential update process from the case to main upgrade step as a critical step to cover more platforms
3. refactor all `jq` commands to `jsonpath` in upgrade-toimage due to no `jq` tool in image `cloud-credential-operator`

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
Add back unrelated steps of upgrade chain in https://github.com/openshift/release/compare/5198ccc50635292b47c5efbccbdbe6b3e05c23cc..6b0bf815e501db6fd5f45f4ec160959056a63cc0

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -41,6 +41,41 @@ EOF
...
+    # TODO: add gcp and azure
+    # Update iam role with ccoctl based on tobeCredsDir
+    if [[ "${platform}" == "AWS" ]]; then
+        if [[ ! -e ${SHARED_DIR}/aws_oidc_provider_arn ]]; then
+            echo "No aws_oidc_provider_arn file in SHARED_DIR" && exit 1
+        else
+            oidc_provider=$(head -n1 ${SHARED_DIR}/aws_oidc_provider_arn)
+        fi
+    fi
+    if ! diff -r "${preCredsDir}" "${tobeCredsDir}" &> /dev/null; then
```
jianlinliu, line 70: Is this a restrict diff? Even just some permission difference, would start to create new credentials?

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -241,12 +276,6 @@ function check_clusteroperators() {
     done < "${input}"
     rm -f "${tmp_clusteroperator}"
 
-    echo "Make sure every operator column reports version"
```
jianlinliu, line 244: Any background to remove this check?

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -525,12 +556,12 @@ export OC="run_command_oc"
 mkdir -p /tmp/client
 export OC_DIR="/tmp/client"
 export PATH=${OC_DIR}:$PATH
-
+export AWS_SHARED_CREDENTIALS_FILE="${CLUSTER_PROFILE_DIR}/.awscred"
```
jianlinliu, line 559: Only aws upgrade has this file, so this line should be moved to aws if condition in `update_cloud_credentials`

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-ref.yaml
```diff
@@ -1,13 +1,10 @@
 ref:
   as: cucushift-upgrade-toimage
-  from_image:
-    namespace: ci
-    name: verification-tests
-    tag: latest
+  from: cloud-credential-operator
```
jianlinliu, line 3: Hmm, I am afraid this change would break some other upgrade scenarios, I can not remember exact reason why we used `verification-tests`, maybe related to rhel worker upgrade stuff, cc @shellyyang1989 to confirm.

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -41,6 +41,41 @@ EOF
...
+        echo "No need update cloud credentials" && return
+    fi
+    preCredsDir="/tmp/pre-include-creds"
+    tobeCredsDir="/tmp/tobe-include-creds"
+    mkdir "${preCredsDir}" "${tobeCredsDir}"
+    # Extract all CRs from live cluster with --included
+    if ! oc adm release extract --to "${preCredsDir}" --included --credentials-requests; then
+        echo "Failed to extract CRs from live cluster!" && exit 1
+    fi
+    if ! oc adm release extract --to "${tobeCredsDir}" --included --credentials-requests "${OPENSHIFT_UPGRADE_RELEASE_IMAGE_OVERRIDE}"; then
```
jianlinliu, line 57:
```suggestion
    if ! oc adm release extract --to "${tobeCredsDir}" --credentials-requests "${TARGET}"; then
```

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -41,6 +41,41 @@ EOF
...
+    if [[ "${platform}" == "AWS" ]]; then
+        if [[ ! -e ${SHARED_DIR}/aws_oidc_provider_arn ]]; then
+            echo "No aws_oidc_provider_arn file in SHARED_DIR" && exit 1
+        else
+            oidc_provider=$(head -n1 ${SHARED_DIR}/aws_oidc_provider_arn)
+        fi
+    fi
+    if ! diff -r "${preCredsDir}" "${tobeCredsDir}" &> /dev/null; then
+        infra_name=${NAMESPACE}-${UNIQUE_HASH}
+        if ! ccoctl aws create-iam-roles --name="${infra_name}" --region="${LEASED_RESOURCE}" --credentials-requests-dir="${tobeCredsDir}" --identity-provider-arn="${oidc_provider}"; then
```
jianlinliu, line 72: I think this command should run in the if block on line 63.

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
jiajliu: yes, it will

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
jiajliu: This step is covered in `Make sure every operator's AVAILABLE column is True`.

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
jiajliu: ack, updated

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
jiajliu: good catch, updated

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
jiajliu: updated, thanks

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-ref.yaml
jiajliu: Using the image is for ccoctl tool:(

ci-operator/step-registry/cucushift/ota/postupgrade/cucushift-ota-postupgrade-commands.sh
```diff
@@ -0,0 +1,85 @@
...
+    newCap=$(grep -rh "capability.openshift.io/name:" "${credsDir}"|awk -F": " '{print $NF}'|sort -u|xargs)
+    if [[ "${newCap}" != "MachineAPI ImageRegistry" ]] && [[ "${newCap}" != "ImageRegistry MachineAPI" ]]; then
+        echo "Tobe gained CRs with cap annotation: ${newCap}, but expected: MachineAPI and ImageRegistry"
+        return 1
+    fi
+    echo "Test Passed: ${FUNCNAME[0]}"
+    return 0
+}
+
+# This func run all test cases with with checkpoints which will not break other cases,
```
shellyyang1989, line 47: nit: with with -> with, ran -> run

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
jianlinliu: so that would generate some duplicates IAM credentials, when only some permission difference between source and target payload?

ci-operator/step-registry/cucushift/ota/postupgrade/cucushift-ota-postupgrade-commands.sh
```diff
@@ -0,0 +1,85 @@
...
+        sleep 60
+    done
+    mv ${tmp_oc}/oc /tmp -f
+    which oc
+    oc version --client
+}
+
+# Define the checkpoints/steps needed for the specific case
+function post-ocp-66839(){
+    if [[ "${ADDITIONAL_ENABLED_CAPABILITIES}" != "CSISnapshot" ]] || [[ "${BASELINE_CAPABILITY_SET}" != "None" ]]; then
```
shellyyang1989, line 25: Looks like the test-case defines baseline as `None` only. Just curious why we have the additional cap CSISnapshot as a prerequisite.

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
jianlinliu: I think the step "Make sure every operator's AVAILABLE column is True"  is checking "AVAILABLE" column, while this step is checking "VERSION" column.

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -41,6 +41,47 @@ EOF
...
+        echo "No need update cloud credentials" && return
+    fi
+    preCredsDir="/tmp/pre-include-creds"
+    tobeCredsDir="/tmp/tobe-include-creds"
+    mkdir "${preCredsDir}" "${tobeCredsDir}"
+    # Extract all CRs from live cluster with --included
+    if ! oc adm release extract --to "${preCredsDir}" --included --credentials-requests; then
+        echo "Failed to extract CRs from live cluster!" && exit 1
+    fi
+    if ! oc adm release extract --to "${tobeCredsDir}" --included --credentials-requests "${TARGET}"; then
```
jianlinliu, line 57: The `--include` would help only extract necessary credential CR manifests against `TARGET` even the current  cluster is not using the TARGET payload?

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
jiajliu: If my understanding correct, according to the discussion with jianping in slack, ccoctl will judge current role and only update necessary ones.

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
jiajliu: my bad, a wrong paste, i meant it's covered in `"Make sure every operator do not report empty column"`

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
jiajliu: yes

ci-operator/step-registry/cucushift/ota/postupgrade/cucushift-ota-postupgrade-commands.sh
jiajliu: It's a requirement from installer by jianlin, so we cover the same profile(`aws-ipi-disconnected-sts-basecap-none-addcap-f28`) here.

ci-operator/step-registry/cucushift/ota/postupgrade/cucushift-ota-postupgrade-commands.sh
jiajliu: > ran -> run
You mean run -> ran? in `This func run`

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
jianlinliu: Got it. Thanks.

shellyyang1989: Hmm, assuming `VERSION` field is empty, but we check `AVAILABLE` by using `if unavailable_operator=$(${OC} get clusteroperator | awk '$3 == "False"' | grep "False")`, then we will consider `PROGRESSING` status as `AVAILABLE` status by mistake.

```
# oc get co
NAME                                       VERSION                              AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
authentication                              False        True         False      14m     
baremetal                                      False        True         False      33m     
```

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-ref.yaml
shellyyang1989: We are using `verification-tests` is because we have asked productivity team to integrate `jq` and `ansible` in it. If we need `ccoctl`, we can add it to `verification-tests` as well.

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -41,6 +41,47 @@ EOF
     fi
 }
 
+function update_cloud_credentials(){
```
shellyyang1989, line 44: Loop @jianping-shu to take a look at the change about CR update before upgrade

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
jiajliu: @shellyyang1989 you are talking about the step `Make sure every operator's AVAILABLE column is True`, right? I've corrected myself in https://github.com/openshift/release/pull/43227/files#r1339475755. It should be `Make sure every operator do not report empty column`

ci-operator/step-registry/cucushift/ota/postupgrade/cucushift-ota-postupgrade-commands.sh
shellyyang1989: sorry, should be ran -> should be run

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
shellyyang1989: Yeah, If I recall correctly, oc outputs all the CR files needed by the target image instead of the different files as compared with source image. And ccoctl is able to apply the differences to the cluster when we specify the target CR set

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-ref.yaml
jiajliu: > If we need ccoctl, we can add it to verification-tests as well.

We had a simple discussion around it in the begging, considering it need maintain a dockerfile so drop it. But if we could not find better way, i think we should add it. @jianlinliu hdyt?
@shellyyang1989 could you share the link or jira issue for the requirement to add tool in base image `verification-tests`

shellyyang1989: Here you go https://issues.redhat.com/browse/OCPQE-11640

ci-operator/step-registry/cucushift/ota/postupgrade/cucushift-ota-postupgrade-commands.sh
jiajliu: updated

ci-operator/step-registry/cucushift/ota/postupgrade/cucushift-ota-postupgrade-commands.sh
jianlinliu: Hmm, `ADDITIONAL_ENABLED_CAPABILITIES` is set with different values in different versions, e.g: for 4.13 source version, the optional cap is CSISnapshot, but on 4.14 source cluster, it is CSISnapshot, and Machine-API.

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-ref.yaml
jianlinliu: For the ccoctl, I think we can extract it from payload image just like what we did for oc client.

ci-operator/step-registry/cucushift/ota/postupgrade/cucushift-ota-postupgrade-commands.sh
jiajliu: Then we could control it in upgrade [config file](https://github.com/openshift/release/pull/43227/files#diff-48817126d567bdb6f9b42dc07689ff7d2a387dd3eb9eccc516448f7792282e10L201) in future versions, right? I think variable additional caps is not the checkpoint for this case, so we could keep CSISnapshot only in the following upgrade config.

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-ref.yaml
jiajliu: Would u like a static tool pre-installed in the image or extract it dynamically in each upgrade? A little concern if it will bring one more unstable point to upgrade process.

jiajliu: Let me update it to use dynamic extract way.

jiajliu: updated

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-ref.yaml
```diff
@@ -3,11 +3,10 @@ ref:
   from_image:
     namespace: ci
     name: verification-tests
-    tag: latest
```
shellyyang1989, line 6: Is it intentional to remove the tag?

jiajliu: not really:) add it back~

jiajliu:
Hi @jianlinliu @shellyyang1989 @jianping-shu Could u take a look if any further comments? if not, i will trigger one round test against the update:)

shellyyang1989:
LGTM

ci-operator/step-registry/cucushift/ota/postupgrade/cucushift-ota-postupgrade-commands.sh
jianlinliu: For the `ADDITIONAL_ENABLED_CAPABILITIES`, jinyum did a lot of work to cover all kinds of combination, the additional cap can be set via ENV vars, also some random selection via a dedicated step. Whatever the additional cap is set from where, the final setting would be landed into install-config.yaml, maybe we can get the additional setting from https://github.com/openshift/release/blob/master/ci-operator/step-registry/cucushift/installer/check/capability/cucushift-installer-check-capability-commands.sh#L143 to be more dynamic?

jianlinliu: And actually `CSISnapshot` setting on installation at the beginning has some configuration concerns, because storage operator is disabled, but  CSISnapshot actually can not work, we added it just because we want to ensure the additional cap can cause some difference on credential CR manifest for our case testing purpose. In the future, especially in next version, e.g: 4.14, more operators become optional, maybe we consider to change the setting. So I think we can discuss more about how to make the `ADDITIONAL_ENABLED_CAPABILITIES` judgement to be more reasonable.

jiajliu: Hmm. yes, i know we have supported all kinds of combination for caps. And there are other cvo/upgrade cases covering them. For this case, the checkpoint is for `--include` cmd instead of optional caps, and so we drop the idea to integrate the checkpoint into existed upgrade case for optional caps in our case review stage, but add one more dedicated case is to cover it in upgrade process.
Jinyun's preface work could be used in optional caps cases integration, but it's not quite the same here, because not all caps have corresponding CRs, that means we need maintain another map for it. But actually it's not quite valuable because if `--include` cmd works in one combination, it would work in other combination. As for the command itself, no difference.

jiajliu: FYI https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitems/testcase?query=trello%3AOTA%5C-555%20AND%20subteam.KEY%3Aota%20AND%20NOT%20status%3Ainactive 
Just checked there were 8 upgrade cases about optional caps, so above combination will be covered in those cases. As for `CSISnapshot`, if it makes us confused, i could remove it. Actually case ocp-66839 only cares  BASELINE_CAPABILITY_SET be None, due to no extra value, so I keep it there.

jiajliu: Drop `ADDITIONAL_ENABLED_CAPABILITIES` to match the case checkpoint correctly.

ci-operator/step-registry/cucushift/ota/postupgrade/cucushift-ota-postupgrade-commands.sh
```diff
@@ -0,0 +1,85 @@
...
+    echo "Test Start: ${FUNCNAME[0]}"
+    credsDir="/tmp/post-include-creds"
+    mkdir "${credsDir}"
+    if ! oc adm release extract --to "${credsDir}" --included --credentials-requests; then
+        echo "Failed to extract manifests!"
+        return 1
+    fi
+    # New gained cap annotation should be in extracted creds 
+    newCap=$(grep -rh "capability.openshift.io/name:" "${credsDir}"|awk -F": " '{print $NF}'|sort -u|xargs)
+    if [[ "${newCap}" != "MachineAPI ImageRegistry" ]] && [[ "${newCap}" != "ImageRegistry MachineAPI" ]]; then
```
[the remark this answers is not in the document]
jiajliu: It should be. But I remember the output is different between my local test and pr test, so I add it to be stronger...

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -41,6 +41,71 @@ EOF
...
+    else
+        chmod 775 /tmp/ccoctl
+    fi
+}
+
+function update_cloud_credentials(){
+    set -x
+    platform=$(oc get infrastructure cluster -o jsonpath='{.status.platformStatus.type}')
+    cco_mode=$(oc get cloudcredential cluster -o jsonpath='{.spec.credentialsMode}')
+    if (( SOURCE_MINOR_VERSION == TARGET_MINOR_VERSION )) || (( SOURCE_MINOR_VERSION < 13 )) || [[ "${cco_mode}" != "Manual" ]]; then
```
jianlinliu, line 69: I think we can use `EXTRACT_MANIFEST_INCLUDED` env to decide if run `update_cloud_credentials` in main function.

jiajliu: could u share more info about this env?

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -41,6 +41,71 @@ EOF
...
+        mkdir "${toManifests}"
+        case "${platform}" in
+        "AWS")
+            if [[ ! -e ${SHARED_DIR}/aws_oidc_provider_arn ]]; then
+                echo "No aws_oidc_provider_arn file in SHARED_DIR" && exit 1
+            else
+                export AWS_SHARED_CREDENTIALS_FILE="${CLUSTER_PROFILE_DIR}/.awscred"
+                infra_name=${NAMESPACE}-${UNIQUE_HASH}
+                oidc_provider=$(head -n1 ${SHARED_DIR}/aws_oidc_provider_arn)
+                if ! ccoctl aws create-iam-roles --name="${infra_name}" --region="${LEASED_RESOURCE}" --credentials-requests-dir="${tobeCredsDir}" --identity-provider-arn="${oidc_provider}" --output-dir="${toManifests}"; then
```
jianlinliu, line 93: Suggest to move `extract_ccoctl` prior to this line when ccoctl is necessary to be used.

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -41,6 +41,71 @@ EOF
...
+        mkdir "${toManifests}"
+        case "${platform}" in
+        "AWS")
+            if [[ ! -e ${SHARED_DIR}/aws_oidc_provider_arn ]]; then
+                echo "No aws_oidc_provider_arn file in SHARED_DIR" && exit 1
+            else
+                export AWS_SHARED_CREDENTIALS_FILE="${CLUSTER_PROFILE_DIR}/.awscred"
+                infra_name=${NAMESPACE}-${UNIQUE_HASH}
+                oidc_provider=$(head -n1 ${SHARED_DIR}/aws_oidc_provider_arn)
+                export PATH=/tmp:${PATH}
```
jianlinliu, line 91: I guess we can move this line into `extract_ccoctl` function?

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__amd64-nightly-4.14-upgrade-from-stable-4.13.yaml
```diff
@@ -189,13 +189,12 @@ tests:
   steps:
     cluster_profile: aws-qe
     env:
-      ADDITIONAL_ENABLED_CAPABILITIES: CSISnapshot
```
jianlinliu, line 192: If we keep the setting, would that affect the following testing?

jiajliu: No, we have discussed it in thread https://github.com/openshift/release/pull/43227#discussion_r1339470158. So let's updated it to comply with the case correctly to avoid misunderstanding here.

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
jiajliu: I think so, let me updated, and do u think it's worth a re-test since the latest test result passed in https://qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/qe-private-deck/pr-logs/pull/openshift_release/43227/rehearse-43227-periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-disconnected-sts-basecap-none-addcap-f28/1711214278461100032

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__amd64-nightly-4.14-upgrade-from-stable-4.13.yaml
jianlinliu: Do you mind to update the same job in openshift-openshift-tests-private-release-4.15__amd64-nightly-4.15-upgrade-from-stable-4.14.yaml

ci-operator/step-registry/cucushift/ota/postupgrade/cucushift-ota-postupgrade-commands.sh
```diff
@@ -0,0 +1,85 @@
...
+        sleep 60
+    done
+    mv ${tmp_oc}/oc /tmp -f
+    which oc
+    oc version --client
+}
+
+# Define the checkpoints/steps needed for the specific case
+function post-ocp-66839(){
+    if [[ "${BASELINE_CAPABILITY_SET}" != "None" ]]; then
```
jianlinliu, line 25: I am thinking if we should remove the condition, so that we can easily find out any mis-configuration issue in job config level.

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
jianlinliu: I think we can directly update, no need to re-run the testing, but I suggest to run some other common upgrade jobs to ensure the big change would not break other things.

ci-operator/step-registry/cucushift/ota/postupgrade/cucushift-ota-postupgrade-commands.sh
jiajliu: Prefer to keep it there to distinguish the case is skipped or failed.

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__amd64-nightly-4.14-upgrade-from-stable-4.13.yaml
jiajliu: sure, let me updated all other related jobs

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
jiajliu: ack, let me restore pre-upgrade and post-upgrade steps and trigger one common upgrade

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-gcp-ipi-ovn-ipsec-f14

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__amd64-nightly-4.14-upgrade-from-stable-4.13.yaml
jianlinliu: Ah, sounds we also need to update the ci job name to remove `addcap` keyword together.

If possible, can you also help update the fresh install e2e jobs ones togher?

```
[jianlinliu@jialiu-laptop openshift-tests-private]$ grep -r "as: aws.*sts" *release-4.1{4,5}*|grep base
openshift-openshift-tests-private-release-4.14__amd64-nightly-4.14-upgrade-from-stable-4.13.yaml:- as: aws-ipi-disconnected-sts-basecap-none-addcap-f28
openshift-openshift-tests-private-release-4.14__amd64-nightly.yaml:- as: aws-ipi-disconnected-sts-basecap-none-addcap-f28
openshift-openshift-tests-private-release-4.14__amd64-nightly.yaml:- as: aws-ipi-disconnected-sts-basecap-none-addcap-f28-destructive
openshift-openshift-tests-private-release-4.14__arm64-nightly-4.14-upgrade-from-stable-4.13.yaml:- as: aws-ipi-disconnected-sts-basecap-none-addcap-f28
openshift-openshift-tests-private-release-4.14__arm64-nightly.yaml:- as: aws-ipi-disconnected-sts-basecap-none-addcap-f28
openshift-openshift-tests-private-release-4.14__arm64-nightly.yaml:- as: aws-ipi-disconnected-sts-basecap-none-addcap-f28-destructive
openshift-openshift-tests-private-release-4.15__amd64-nightly-4.15-upgrade-from-stable-4.14.yaml:- as: aws-ipi-disconnected-sts-basecap-none-addcap-f28
openshift-openshift-tests-private-release-4.15__amd64-nightly.yaml:- as: aws-ipi-disconnected-sts-basecap-none-addcap-f28
openshift-openshift-tests-private-release-4.15__amd64-nightly.yaml:- as: aws-ipi-disconnected-sts-basecap-none-addcap-f28-destructive
openshift-openshift-tests-private-release-4.15__arm64-nightly-4.15-upgrade-from-stable-4.14.yaml:- as: aws-ipi-disconnected-sts-basecap-none-addcap-f28
openshift-openshift-tests-private-release-4.15__arm64-nightly.yaml:- as: aws-ipi-disconnected-sts-basecap-none-addcap-f28
openshift-openshift-tests-private-release-4.15__arm64-nightly.yaml:- as: aws-ipi-disconnected-sts-basecap-none-addcap-f28-destructive
```

jiajliu: @jianlinliu Are you sure we need update fresh install jobs too? I see there are ADDITIONAL_ENABLED_CAPABILITIES for them.
```
- as: aws-ipi-disconnected-sts-basecap-none-addcap-f28
  cron: 24 21 27 * *
  steps: 
    cluster_profile: aws-qe
    env:
      ADDITIONAL_ENABLED_CAPABILITIES: CSISnapshot MachineAPI
      BASELINE_CAPABILITY_SET: None
```

jianlinliu: Yeah, the jobs were added on purpose to cover the test case, on that time, I thought the additional cap is a requirement for the case, so I added  `CSISnapshot` coverage for both fresh install and upgrade. For `MachineAPI`, there is a known issue when running ipi install + none basecap, so enable `MachineAPI` as a workaround, but that is not a test point.

jiajliu: ack, then let me update install job accordingly

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__amd64-nightly-4.15-upgrade-from-stable-4.14.yaml
```diff
@@ -191,12 +191,13 @@ tests:
   steps:
     cluster_profile: aws-qe
     env:
-      ADDITIONAL_ENABLED_CAPABILITIES: CSISnapshot
```
jianlinliu, line 194: For 4.14, there is a known issue, we have to enable `MachineAPI` here.

jiajliu: You mean we should keep `ADDITIONAL_ENABLED_CAPABILITIES: MachineAPI` for 4.14 install but with name `aws-ipi-disconnected-sts-basecap-none-f28`? If so, how about drop `aws-ipi-disconnected-sts-basecap-none-f28` for 4.14 until the bug fix since the test checkpoint is for None cap?

jianlinliu: `MachineAPI` becomes a optional cap in 4.14, so once basecap is set to none, the machine-api would be disabled too, but that would lead the install failed, actually I am afraid there would be no real fix for such case, but just report some pre-check from installer or tell customers via document to guide them to configure the additional caps correctly in this case.

jianlinliu: > You mean we should keep ADDITIONAL_ENABLED_CAPABILITIES: MachineAPI for 4.14 install but with name aws-ipi-disconnected-sts-basecap-none-f28

Right.

jiajliu: ack, then i could update it in 4.14 install job config, and drop the test in 4.14-4.15 upgrade job. I will file another pr for 4.14-4.15 upgrade on such profile if MachineAPI would be a must add cap, is that ok for u?

jianlinliu: For sure, I can update the other jobs in another PR if that would introduce some more noise in this PR.

jiajliu: I just updated but not touch the 4.14-4.15 upgrade job which does not enable the case test yet. The case need extra work if we need set additional cap for 4.14 install, so i will have another pr to deal with it. PTAL

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__amd64-nightly.yaml
```diff
@@ -484,12 +484,11 @@ tests:
     test:
     - chain: openshift-e2e-test-qe-destructive
     workflow: cucushift-installer-rehearse-aws-ipi-disconnected-private-sdn-workers-rhcos-rhel8
-- as: aws-ipi-disconnected-sts-basecap-none-addcap-f28
+- as: aws-ipi-disconnected-sts-basecap-none-f28
   cron: 21 14 14 * *
   steps:
     cluster_profile: aws-qe
     env:
-      ADDITIONAL_ENABLED_CAPABILITIES: CSISnapshot MachineAPI
```
jianlinliu, line 492: Are we going to update this job in this PR? If yes, I think we need to keep `MachineAPI` here.

jiajliu: I thought installer only need the workaround for 4.14 installation? we also need it for 4.15 installation, right?

jiajliu: ok, i'm not sure the difference between 4.14 and 4.15 behavior on the workaround, to save time, i will not touch install part in this pr, @jiajliu pls help handle it in dedicated pr for install part.

jianlinliu: > we also need it for 4.15 installation, right?
I think so.

jianlinliu: > pls help handle it in dedicated pr for install part.

okay.

jianlinliu:
@jiajliu the PR looks good to me after `make jobs` to make `ci/prow/generated-config` happy.

jiajliu:
> @jiajliu the PR looks good to me after `make jobs` to make `ci/prow/generated-config` happy.

Finally:) and would u mind lgtm for it if no other missing here.

jianlinliu:
> > @jiajliu the PR looks good to me after `make jobs` to make `ci/prow/generated-config` happy.
> 
> Finally:) and would u mind lgtm for it if no other missing here.

The required job - https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_release/43227/pull-ci-openshift-release-master-generated-config/1711601630689890304 failed, I think you need to run `make jobs` to get it fixed, after that, I can add lgtm there.

jiajliu:
> > > @jiajliu the PR looks good to me after `make jobs` to make `ci/prow/generated-config` happy.
> > 
> > 
> > Finally:) and would u mind lgtm for it if no other missing here.
> 
> The required job - https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_release/43227/pull-ci-openshift-release-master-generated-config/1711601630689890304 failed, I think you need to run `make jobs` to get it fixed, after that, I can add lgtm there.

Wired, it has passed from my side.  Refer to https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_release/43227/pull-ci-openshift-release-master-generated-config/1711606628907749376

jianlinliu:
> Wired, it has passed from my side. Refer to https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_release/43227/pull-ci-openshift-release-master-generated-config/1711606628907749376

Okay, after refresh the page, good now.

/lgtm

jianlinliu:
cc @liangxia to review

PR #43306, opened 2023-09-13
remove ipi-install-times-collection step from upi install

jianlinliu:
Per https://github.com/openshift/installer/pull/3065, openshift-install configmap can be used to determinate the install is a UPI or IPI install, while `ipi-install-times-collection` always create the configmap with the same name, so remove it avoid the conflicts.


jianlinliu:
cc @jianli-wei and @jinyunma to review.

PR #43317, opened 2023-09-13
Mount Route53 credentials in the openshift-extended-test-longduration step

fxierh:
Mount Route53 credentials in the openshift-extended-test-longduration step so test cases running on non-AWS platforms (e.g. vSphere) can make use of them.


fxierh:
/cc @kuiwang02 
Would you please take a look ?
Also, thanks for your help :)

kuiwang02:
@fxierh after it is merged, please help monitor if the longduration step still works (as long as the step start to work, it is ok)
thanks

/lgtm

fxierh:
> @fxierh after it is merged, please help monitor if the longduration step still works (as long as the step start to work, it is ok) thanks

Okay.

PR #43319, opened 2023-09-13
Retire 4.10 periodic jobs for OpenStack

MaysaMacedo:
The 4.10 release has recently had the maintenance support ended. This commit retires the 4.10 jobs, with exception of 4.10 upgrades, which customers might run.


mandre (commented):
Thanks. Could you also do the same for the 4.9 -> 4.10 upgrade jobs?

MaysaMacedo:
@mandre done

PR #43323, opened 2023-09-13
Add not workloads amd64

kasturinarra:
Add not workloads tag to make sure that dc cases does not run on the profile where baselineCaps is set to none.

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__amd64-nightly.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__amd64-nightly.yaml
@@ -153,7 +153,7 @@ tests:
       BASE_DOMAIN: qe.devcluster.openshift.com
       BASELINE_CAPABILITY_SET: None
       E2E_RUN_TAGS: '@amd64 and @aws-ipi and @network-ovnkubernetes and not @fips
-        and not @console and not @storage'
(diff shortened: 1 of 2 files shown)


kasturinarra:
@liangxia @zhouying7780 could you please help review ? thanks !!

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__amd64-nightly.yaml
```diff
@@ -168,7 +168,7 @@ tests:
       BASE_DOMAIN: qe.devcluster.openshift.com
       BASELINE_CAPABILITY_SET: None
       E2E_RUN_TAGS: '@amd64 and @aws-ipi and @network-ovnkubernetes and not @fips
-        and not @console and not @storage'
+        and not @console and not @storage and not @workloads'
```
liangxia, line 171: We have several `baselinecaps-none` profiles. Perhaps we should apply the same change to the other `baselinecaps-none` profiles, too ?

kasturinarra: sure, let me check and update, thankyou !!

kasturinarra: @liangxia done for all, could you please help check ? thanks !!

PR #41649, opened 2023-07-25
Create ref for interop weekly testing trigger.

ascerra:
This PR introduces a new ref that can be used to create a triggering job with active and inactive jobs

This will enable a new triggering strategy for weekly interop testing.

See JIRA for more info https://issues.redhat.com/browse/LPTOCPCI-490

Example of build.log.txt output when the API is down: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_release/41649/rehearse-41649-periodic-ci-rhpit-interop-tests-main-weekly_trigger-weekly-trigger/1689057637297033216/artifacts/weekly-trigger/cspi-qe-weekly-trigger/build-log.txt

Example of build-log.txt output when API is up: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_release/41649/rehearse-41649-periodic-ci-rhpit-interop-tests-main-weekly_trigger-ocp-self-managed-layered-product-interop/1701676873211187200/artifacts/ocp-self-managed-layered-product-interop/trigger-jobs/build-log.txt

PR #43330, opened 2023-09-13
Setting all interop program cron to run once per year.

ascerra:
We are implementing a new triggering strategy that will use the gangway api to trigger our jobs.
- This allows us to read a file that we store in vault to determine if we want to skip or not skip a specific scenario for the weekly run
- Here is the PR to implement this new trigger job: https://github.com/openshift/release/pull/41649
- The cron has been updated to run once a year during company shutdown. This can either remain if we want to run that week or we can skip it completely with another PR similar to this.


calebevans (approved):
/lgtm

PR #43352, opened 2023-09-14
shiftstack/prune: Ignore vendor when running static code checks

pierreprinetti:


mandre (commented):
For the record, `go fmt` is calling `gofmt` with the `-l` and `-w` options, and has logic to ignore the `vendor` directory, as I had to ask @pierreprinetti .

/lgtm
/approve

PR #43354, opened 2023-09-14
Update job triggers for CI to run less jobs

SimonBaeumer:
Update the job triggers to reduce the amount of jobs.

 - Skip e2e tests when documentation changed
 - Skip image-push when documentation changed
 - Run all jobs on main branch


SimonBaeumer:
@johannes94 @kurlov As you pointed out offline, it is not possible currently to enable coditional builds because of github branch protection.
It would be possible by:
 - labelling the pr as `skip-e2e-tests` and checking for a label
 - early exit e2e tests via script when specific files were changed

These changes would require more effort, so I am abandoning this PR for now.

PR #43361, opened 2023-09-14
Extend permissions to more users so that they could add cherry-pick-approved labels

kasturinarra:
Currently users in the PR does not have permissions to add `cherry-pick-approved` labels, adding them to these repos so that they could add the label.


soltysh (approved):
/lgtm
/approve

PR #43367, opened 2023-09-14
openstack: Add missing step to ipi-conf-openstack-noramfs chain

stephenfin:
The `ipi-conf-openstack-noramfs` chain describes itself as "identical to the `ipi-conf-openstack` except that it omits the `ipi-conf-etcd-on-ramfs` step." Well it also omits the `openstack-provision-vips-ports` when it shouldn't. This is a no-op since we don't run dual stack jobs with this chain, but it keeps us consistent and avoid a potential source of confusion in the future.


MaysaMacedo:
I had this fixed here https://github.com/openshift/release/pull/43210
just forgot to add the pj-rehearse ack.

Thanks @stephenfin

stephenfin:
/close

@MaysaMacedo had already spotted this

PR #43372, opened 2023-09-14
Add boskos config for GCP arm64

SaravanaStorageNetwork:
Ref: [ARMOCP-563](https://issues.redhat.com/browse/ARMOCP-563)

### DIFF
--- a/core-services/ci-secret-bootstrap/_config.yaml
+++ b/core-services/ci-secret-bootstrap/_config.yaml
@@ -1117,6 +1117,10 @@ secret_configs:
     - non_app_ci
     name: cluster-secrets-gcp
     namespace: ci
+  - cluster_groups:
+    - non_app_ci
+    name: cluster-secrets-gcp-arm64
+    namespace: ci
 - from:
     pull-secret:
       dockerconfigJSON:
--- a/core-services/prow/02_config/_boskos.yaml
+++ b/core-services/prow/02_config/_boskos.yaml
@@ -995,6 +995,39 @@ resources:
   state: free
   type: fleet-manager-qe-quota-slice
 - names:
+  - us-central1--gcp-arm64-quota-slice-00
+  - us-central1--gcp-arm64-quota-slice-01
+  - us-central1--gcp-arm64-quota-slice-02
+  - us-central1--gcp-arm64-quota-slice-03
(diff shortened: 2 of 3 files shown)


core-services/ci-secret-bootstrap/_config.yaml
```diff
@@ -1204,6 +1204,10 @@ secret_configs:
     - non_app_ci
     name: cluster-secrets-gcp-qe
     namespace: ci
+  - cluster_groups:
```
aleskandro, line 1207: Maybe this should not go together with the QE secrets, as QE uses other accounts to host the ephemeral clusters.

core-services/prow/02_config/_boskos.yaml
```diff
@@ -994,6 +994,39 @@ resources:
   - ap-northeast-1--fleet-manager-qe-quota-slice-2
   state: free
   type: fleet-manager-qe-quota-slice
+- names:
```
aleskandro, line 997: I also wonder whether we actually need this additional cluster-profile, since the books config for the non-arm64 dev one is using only the us-central1 zone... @deepsm007 wdyt?

core-services/ci-secret-bootstrap/_config.yaml
SaravanaStorageNetwork: Ack. will move it from here.

core-services/prow/02_config/_boskos.yaml
SaravanaStorageNetwork: IMHO separation of concerns helps.

deepsm007: Better to have separate cluster-profile, it allows us to control zone we want to use in future.

Prashanth684: is arm64 available in us-central?

SaravanaStorageNetwork: > are arm64 instances available in us-central?

Yes.

Please refer https://github.com/openshift/release/blob/ec17dd1983ff77374a8dcf7da9085b9a602bd7a3/core-services/prow/02_config/generate-boskos.py#L138-L146

PR #43365, opened 2023-09-14
openstack: Ensure resource names are configured

stephenfin:
The `openstack-conf-clouds` step now has a dependency on the configuration generated by the `openstack-conf-resourcenames` step (as of #43348). Ensure all usage of the former first uses the latter.

PR #43373, opened 2023-09-14
openstack: Replace symlink with actual file

stephenfin:
It seems prow doesn't handle steps from different "root" step-registry directories very well.

This is a speculative change to help us figure out what's going on in #43365.


pierreprinetti:
/assign vrutkovs
pretty please?

stephenfin:
fwiw, this worked and allowed us to merge #43365

PR #43376, opened 2023-09-14
make CSI presubmit jobs required for cluster-storage-operator

dobsonj:
The generic e2e jobs for each cloud provider don't provide much value compared to runnning the CSI test suite for each cloud provider. If for example we break RBAC permissions in this repo, it's the CSI test suite that will surface the issue(s).

This PR makes the generic e2e jobs optional and instead requires the CSI test suite for AWS, GCP, vSphere, and Azure. It also adds IBM VPC as optional.

cc @openshift/storage


jsafrane:
I've got second thoughts. Can you please keep the old e2e jobs there, just not run them by default? I.e. set

```
always_run: false
optional: true
```

Bigger operator changes could deserve `/test all` .

dobsonj:
> I've got second thoughts. Can you please keep the old e2e jobs there, just not run them by default?

Ack, updated

PR #43379, opened 2023-09-14
Fix unbound variable

Kyl-Bempah:
This is a quick fix to resolve the unbound variable error for OSCI_COMPONENT_VERSION


gurnben (approved):
/approve
/lgtm

PR #43395, opened 2023-09-15
OCPBUGS-18979: update agent-installer-utils base image to RHEL 9

zaneb:


andfasano:
The `builder` image is still getting resolved as rhel8, I think we need to land something like https://github.com/openshift/agent-installer-utils/pull/28/files first

zaneb:
Right, this is just the base image. The builder image gets updated by https://github.com/openshift-eng/ocp-build-data/pull/3520, after which a bot will sync to the agent-installer-utils repo.
In theory, if this were used as a real image then the base image should get updated first, because a binary built on RHEL 8 will work in a RHEL 9 base image, but the reverse is not true.
In practice, I suspect we are linking against a dynamic library in the build image and then copying the binary thus built along with the library from the _base_ image to the host to run, in which case... everything has to be updated together (or at least using the same library versions), which is not actually possible to do atomically given all the different places that configuration lives.

The rehearsals show that this doesn't break CI at least, which I think is probably a good enough reason to merge it.

zaneb:
@andfasano so based on our discussion I think we're ok with merging this?

andfasano:
> @andfasano so based on our discussion I think we're ok with merging this?

Yes, the only bit: I think it'd be better to limit the scope of this change to master (4.15) only

zaneb:
OK, it has to go back to 4.14 eventually, but we can do a separate PR for that after we've tested this in 4.15. I dropped the 4.14 part from this PR.
/pj-rehearse ack

zaneb:
@andfasano I believe this can merge now? The builder image was updated in https://github.com/openshift/release/commit/599571752549de09f777e07eacb865571316e6a9#diff-56e565dcada4c981b67132618a20dec6c8ad64669d96465375b6989f768e6d4c

rwsu:
/test build09-dry
build09-dry failure:
```
{"component":"applyconfig","file":"/go/src/github.com/openshift/ci-tools/cmd/applyconfig/applyconfig.go:145","func":"main.commandExecutor.runAndCheck","level":"info","msg":"oc apply -f clusters/build-clusters/common/rh-splunk/deploy.yaml -o name --kubeconfig /etc/build-farm-credentials/kubeconfig --dry-run=server --validate=true: failed to apply\nerror: resource mapping not found for name: \"splunkforwarder\" namespace: \"rh-corp-logging\" from \"clusters/build-clusters/common/rh-splunk/deploy.yaml\": no matches for kind \"SplunkForwarder\" in version \"splunkforwarder.managed.openshift.io/v1alpha1\"\nensure CRDs are installed first\n","severity":"info","time":"2023-09-28T16:41:23Z"}
{"component":"applyconfig","file":"/go/src/github.com/openshift/ci-tools/cmd/applyconfig/applyconfig.go:236","func":"main.(*configApplier).doWithRetry.func1","level":"info","msg":"Running in server-side dry run mode, attempting to recover previous failure by providing missing namespaces","severity":"info","time":"2023-09-28T16:41:23Z"}
{"component":"applyconfig","file":"/go/src/github.com/openshift/ci-tools/cmd/applyconfig/applyconfig.go:239","func":"main.(*configApplier).doWithRetry.func1","level":"error","msg":"Previous failure was not caused by missing namespaces, cannot recover","severity":"error","time":"2023-09-28T16:41:23Z"}
{"component":"applyconfig","file":"/go/src/github.com/openshift/ci-tools/cmd/applyconfig/applyconfig.go:280","func":"main.(*configApplier).doWithRetry","level":"error","msg":"Apply command failed (not recoverable)","output":"error: resource mapping not found for name: \"splunkforwarder\" namespace: \"rh-corp-logging\" from \"clusters/build-clusters/common/rh-splunk/deploy.yaml\": no matches for kind \"SplunkForwarder\" in version \"splunkforwarder.managed.openshift.io/v1alpha1\"\nensure CRDs are installed first\n","severity":"error","time":"2023-09-28T16:41:23Z"}
```

rwsu (commented):
/lgtm

PR #43431, opened 2023-09-18
kccm: Remove CNV_PRERELEASE env var

qinqon:
The CNV prerelease env var is no longer needed, this change remove that.


orenc1:
the correct CNV versions have been installed.
/lgtm

PR #43436, opened 2023-09-18
ODH DW: Use ODH 2.x for PR checks

sutaakar:
Switched back to install CodeFlare operator from OLM as ODH operator requires CodeFlare subscription to be created.


sutaakar:
Postponing until https://github.com/opendatahub-io/opendatahub-operator/issues/425 is implemented and available.
Then a PR can be directly applies as shown in https://github.com/opendatahub-io/odh-manifests/blob/master/tests/scripts/install.sh#L80

anishasthana (approved):
/lgtm

PR #43452, opened 2023-09-18
assisted-service: disable hypershift-zero-nodes-periodic

danielerez:
Removing 'e2e-ai-operator-ztp-hypershift-zero-nodes-periodic' jobs. They are redundant as the feature is currently unused.


adriengentil (approved):

adriengentil:
/hold
Don't we want to remove the presubmits as well?

adriengentil:
/unhold

>  Don't we want to remove the presubmits as well?
I understand this PR is only about periodic, let's open another one if needed

PR #43485, opened 2023-09-19
Remove the arm64 hypershift variant job configuration

droslean:
/cc @openshift/test-platform @danilo-gemoli @deepsm007 @bryan-cox 


droslean:
@bryan-cox Once we implement the necessary pieces to start supporting the arm64 builds, we can re-introduce the configuration. For now, since all images are optional, the jobs have no purpose to even exist.

PR #43508, opened 2023-09-19
Promote test-catalog image

awgreene:
This commit promotes the test-catalog image so it may be used in an upcoming origin e2e test.

### DIFF
--- a/ci-operator/config/openshift/operator-framework-catalogd/openshift-operator-framework-catalogd-release-4.15.yaml
+++ b/ci-operator/config/openshift/operator-framework-catalogd/openshift-operator-framework-catalogd-release-4.15.yaml
@@ -27,8 +27,6 @@ images:
   to: test-catalog
 promotion:
   disabled: true
-  excluded_images:
-  - test-catalog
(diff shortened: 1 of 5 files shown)


ci-operator/config/openshift/operator-framework-catalogd/openshift-operator-framework-catalogd-release-4.15.yaml
```diff
@@ -27,8 +27,6 @@ images:
   to: test-catalog
 promotion:
   disabled: true
```
awgreene, line 29: I saw that this was added in [this PR](https://github.com/openshift/release/commit/8dc1acba96c0c17f9b5bd060a4a7f8c14595f0bb), I suspect that this is supposed to be present on y stream releases that haven't GA'd.

awgreene:
I'm going to promote [this catalog](https://github.com/openshift/operator-framework-operator-controller/tree/df95362a2a16c9ab2293f7197fd6198298c8ef18/testdata) instead as we can create the bundles defined within.

PR #43527, opened 2023-09-19
Disable branch protection for Tempo operator

pavolloffay:


IshwarKanse:
@pavolloffay We need to disable branch protection for [OpenTelemetry](https://github.com/openshift/release/blob/master/core-services/prow/02_config/openshift/open-telemetry-opentelemetry-operator/_prowconfig.yaml) too. We can add the config for the same to this PR.

frzifus (approved):

PR #43537, opened 2023-09-20
update fips_enable to dr

geliu2016:
@kuiwang02 we want to add fips_enable into dr test(aws and gcp, but not azure), could you please take a view? thanks
cc: @xingxingxia @wsun1

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__amd64-nightly.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__amd64-nightly.yaml
@@ -1093,6 +1093,19 @@ tests:
     - ref: idp-htpasswd
     - ref: openshift-extended-test-longduration
     workflow: cucushift-installer-rehearse-aws-ipi-private
+- as: aws-ipi-private-fips-f14-disasterrecovery
+  cron: 11 11 12,27 * *
+  steps:
+    cluster_profile: aws-qe
+    env:
+      BASE_DOMAIN: qe.devcluster.openshift.com
+      FIPS_ENABLED: "true"
+      TEST_FILTERS: ~ChkUpgrade&;~ConnectedOnly&;~MicroShiftOnly&
+      TEST_SCENARIOS: DR_Testing
+    test:
+    - ref: idp-htpasswd
+    - ref: openshift-extended-test-longduration
+    workflow: cucushift-installer-rehearse-aws-ipi-private
 - as: aws-ipi-private-shared-vpc-phz-sts-f14
   cron: 36 23 8,23 * *
   steps:
@@ -1778,6 +1791,19 @@ tests:
     cluster_profile: azure-qe
     env:
       BASE_DOMAIN: qe.azure.devcluster.openshift.com
+      TEST_FILTERS: ~ChkUpgrade&;~ConnectedOnly&;~MicroShiftOnly&
+      TEST_SCENARIOS: DR_Testing
+    test:
+    - ref: idp-htpasswd
+    - ref: openshift-extended-test-longduration
+    workflow: cucushift-installer-rehearse-azure-ipi-private
+- as: azure-ipi-private-fips-f14-disasterrecovery
+  cron: 35 19 6,21 * *
+  steps:
+    cluster_profile: azure-qe
+    env:
+      BASE_DOMAIN: qe.azure.devcluster.openshift.com
+      FIPS_ENABLED: "true"
       TEST_FILTERS: ~ChkUpgrade&;~ConnectedOnly&;~MicroShiftOnly&
       TEST_SCENARIOS: DR_Testing
     test:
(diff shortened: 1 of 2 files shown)


kuiwang02:
@geliu2016 
please execute `make update` before submitting the PR and then it resolves the ci-operator-config-metadata job failure.

I am OK for the change after you correct it.

please ping qe-productivity team to approve it.

xingxingxia:
Agree. Also, any reason for `but not azure`? And the file ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__amd64-nightly.yaml is only about 4.15, how about 4.14? And how about arm64?

geliu2016:
> Agree. Also, any reason for `but not azure`? And the file ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__amd64-nightly.yaml is only about 4.15, how about 4.14? And how about arm64?

let do it at 4.15 at first, and let's see check-pick to 4.14 or commit new pr later.
I think we don't put all egg in one basket, keep some non-fips test alive.

geliu2016:
> @geliu2016 please execute `make update` before submitting the PR and then it resolves the ci-operator-config-metadata job failure.
> 
> I am OK for the change after you correct it.
> 
> please ping qe-productivity team to approve it.

Updated

kuiwang02:
I am ok.

xingxingxia:
I see all other test jobs that enabled `FIPS_ENABLED: "true"` mostly have `fips` in the jobs' names. For example, `aws-ipi-imdsv2-fips-f14` enables `FIPS_ENABLED: "true"`, the job name `aws-ipi-imdsv2-fips-f14` reflects it. Is it better to follow same?

geliu2016:
> I see all other test jobs that enabled `FIPS_ENABLED: "true"` mostly have `fips` in the jobs' names. For example, `aws-ipi-imdsv2-fips-f14` enables `FIPS_ENABLED: "true"`, the job name `aws-ipi-imdsv2-fips-f14` reflects it. Is it better to follow same?

@kuiwang02 do you think is it necessary? and will job name change affect other part(need more change according to it)?

geliu2016:
@JianLi-RH @liangxia could you please take a review? thx

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__amd64-nightly.yaml
```diff
@@ -1093,6 +1093,19 @@ tests:
     - ref: idp-htpasswd
     - ref: openshift-extended-test-longduration
     workflow: cucushift-installer-rehearse-aws-ipi-private
+- as: aws-ipi-private-disasterrecovery-fips-f14
```
liangxia, line 1096: Suggest the name to `aws-ipi-private-fips-f14-disasterrecovery`.
The name style is `${workflow-name-without-prefix-cucushift-installer-rehearse}(-fips)?-fN-testtype`, existing examples are `xxx-fN-ui, xxx-fN-cloud`

xingxingxia:
Except above lxia's, LGTM from my side. Not sure if it needs at least one pass test?

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__amd64-nightly.yaml
geliu2016: Just done some update, is it expected? @liangxia

JianLi-RH: Just a reminder, there are many **disasterrecovery** tests which name is like **aws-ipi-private-disasterrecovery-fips-f14**, like aws-ipi-private-disasterrecovery-f360 in https://github.com/openshift/release/blob/master/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.10__amd64-nightly.yaml. If you put disasterrecovery after fN will cause the cases name not consistent.

liangxia: We have several jobs added already at the very beginning when we first adapt to Prow CI, before we have the rules for test names. I'd suggest for new jobs, all of us follow the test name rules to have the style consistent, while for old existing jobs, leave some time for sub-teams to decide whether/when they will update it or not.

xingxingxia:
@geliu2016 for openshift/release, no cherry-pick need

geliu2016:
@liangxia if we want to make this update on 4.14, we can only committed new pr to update files: openshift-openshift-tests-private-release-4.14__amd64-nightly.yaml, exact?

liangxia:
> @liangxia if we want to make this update on 4.14, we can only committed new pr to update files: openshift-openshift-tests-private-release-4.14__amd64-nightly.yaml, exact?

Yes, because openshift/release repository use the master branch for all OCP version tests with different file names.

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__amd64-nightly.yaml
xingxingxia: Agree

PR #22659, opened 2021-10-12
Annotate monitoring PVCs for deletion before running the upgrade tests

dgrisonnet:
This PR reverts #21564 which had the same objectives as this PR, to reschedule the monitoring pods across multiple nodes to prevent the monitoring operator to go `Upgradeable: false`, but which had 2 issues that the new implementation will cover.

We've noticed a lot of [CI failures](https://search.ci.openshift.org/?search=monitoring.*Upgradeable%3DFalse&maxAge=48h&context=1&type=bug%2Bjunit&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job) following the merge of https://github.com/openshift/cluster-monitoring-operator/pull/1330, which should have been prevented by #21564.  There were 2 reasons behind those:
1. Some CI jobs weren't executing the `ipi-monitoringpvc-annotate` step so the highly available monitoring workload couldn't be rescheduled resulting in the operator reporting `Upgradeable: false` and causing an early test to fail.
2. Workloads were rescheduled correctly after the `[sig-arch][Early] Managed cluster should start all core operators` test which was failing because the monitoring operator was reporting `Upgradeable: false`.

Issue 1 was fixed by adding the annotation directly to the step that setup persistent storage so we won't miss any job this time.
Issue 2 was fixed by creating the PVCs directly with the annotation so that the rescheduling is done in the same sync loop as the first evaluation of the `Upgradeable` status. So if the workflow are rescheduling properly on the first try, the operator will never report  `Upgradeable: false`.

/cc @simonpasquier 
/assign @deads2k

PR #43553, opened 2023-09-20
MON-3387: ci-operator/step-registry/ipi/install: Remove the openshift…

machine424:
….io/cluster-monitoring-drop-pvc annotation added in https://github.com/openshift/release/pull/22659 as it's no more needed/used

IIUC, this is what clusterbot uses as well.

/cc @simonpasquier

### DIFF
--- a/ci-operator/step-registry/ipi/install/monitoringpvc/ipi-install-monitoringpvc-commands.sh
+++ b/ci-operator/step-registry/ipi/install/monitoringpvc/ipi-install-monitoringpvc-commands.sh
@@ -43,17 +43,12 @@ fi
 # Alibaba requires PVCs to be >= 20GB
 STORAGE="20Gi"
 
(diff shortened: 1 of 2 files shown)


simonpasquier:
/lgtm

I think that the 4.8 > 4.9 upgrade path isn't tested anymore in the CI but to be confirmed. Otherwise I assume that we would need to keep the annotation.

machine424:
> I think that the 4.8 > 4.9 upgrade path isn't tested anymore in the CI but to be confirmed. Otherwise I assume that we would need to keep the annotation.

At the beginning I thought the repo uses versioning then I assumed 4.8 is not longer tested as it's maintenance support ends 4 months ago, but now I can see tests involving it here https://prow.ci.openshift.org/?type=periodic&job=*4.8*

I'll ask on testplatform

/hold

machine424:
Annotation still needed, see jira ticket.

PR #43556, opened 2023-09-20
[WIP] Test new imgae runner

musienko-maxim:


musienko-maxim:
Try to find other way to check runner changes

PR #43560, opened 2023-09-20
Allow force push to tempo and opentelemetry operator repos

pavolloffay:


core-services/prow/02_config/openshift/grafana-tempo-operator/_prowconfig.yaml
```diff
@@ -6,6 +6,7 @@ branch-protection:
           branches:
             main:
               protect: false
```
IshwarKanse, line 8: We can remove protect: false when unmanaged is defined.

frzifus (approved):

droslean:
/hold

Temporary holding to unblock the queue

PR #43628, opened 2023-09-22
Refine test frequency for 4.10

liangxia:
4.10 end of Maintenance was on 10-Sep-2023.


jianlinliu:
@liangxia sounds rebase against upstream would get the ovn-kubenates change disappeared, maybe you can have a try.

PR #43629, opened 2023-09-22
add cucushift-installer-check into azure-stack and azure-mag install chain

jinyunma:
as title


jinyunma:
cc @gpei @jianlinliu to review

jianlinliu:
@danwinship @JacobTanenbaum the PR did not change anything on ovn-kubernates part, but robot told us pre-check failed, have to run `make jobs`, after that, there is some change on ovn-kubernates jobs, can you help review it?

PR #43643, opened 2023-09-22
ensure cucushift-installer-check always run as the last step in provision chain

jianlinliu:


jianlinliu:
cc @jianli-wei @gpei to review.

PR #43656, opened 2023-09-22
stop collecting tcpdump and conntrackdump debugs

jluhrsen:
these only need to be enabled when actively debugging some issue. the files they collect consume a lot of storage and the processes themselves can be heavy
and affect jobs in odd ways, like this:

  https://issues.redhat.com/browse/OCPBUGS-18008


soltysh (approved):
/lgtm
/hold 
for rehearse

bertinatto:
/test all
It seems like tide is stuck

PR #43662, opened 2023-09-22
CNO: Update stable upgrade jobs

jcaamano:
We have branched to 4.15, we need to update the stable upgrade jobs for CNO repo


jluhrsen:
> We have branched to 4.15, we need to update the stable upgrade jobs for ovn-k repo

s/ovn-k/cno

PR #43676, opened 2023-09-25
update multiarch ci jobs to be more clear

jianlinliu:
1. remove `aws-ipi`, because it is being covered by `aws-ipi-ovn-ipsec`
2. rename multiarch ci jobs to prepare the coming migration of more jobs from amd and arm ci jobs, e.g: `amd-mixarch` means controlplane is running on amd arch, while computing is running on mixed of arm and amd nodes.
3. rename some baselinecaps ci jobs which is not the test point.

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__multi-nightly.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__multi-nightly.yaml
@@ -49,8 +49,8 @@ resources:
       cpu: "1"
       memory: 4Gi
 tests:
-- as: aws-ipi-disconnected-private-f14
-  cron: 36 3 11,26 * *
+- as: aws-ipi-disc-amd-mixarch-f14
+  cron: 43 1 3,18 * *
   steps:
     cluster_profile: aws-qe
     dependencies:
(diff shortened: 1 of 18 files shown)


jianlinliu:
cc @liangxia @shellyyang1989 @aleskandro @lwan-wanglin to review.

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__multi-nightly.yaml
```diff
@@ -66,8 +66,8 @@ tests:
     - ref: ipi-install-heterogeneous
     - chain: openshift-e2e-test-qe
     workflow: cucushift-installer-rehearse-aws-ipi-disconnected-private
-- as: aws-ipi-disconnected-private-f28-destructive
-  cron: 28 2 14 * *
+- as: aws-ipi-disc-amd-mixarch-f28-destructive
```
shellyyang1989, line 69: Do we intend to skip `private` in the name?

jianlinliu: Right, on aws, all the disconnected cluster is combining private feature, so I remove it directly to save some chars.

PR #42458, opened 2023-08-17
OTA-1016: CVO: Add hypershift conformance presubmit

petr-muller:
Hypershift is becoming increasingly important and with CVO being one of the major OCP components we may want to be sure we do not break it.

PR #43693, opened 2023-09-26
ci-operator/config/openshift/cluster-version-operator: skip_if_only_changed HyperShift

wking:
Catching the new HyperShift jobs from 5a03727d04 (#42458) up with the other jobs that only care about landed code.


petr-muller (approved):
/pj-rehearse ack

PR #43737, opened 2023-09-26
gcp: use standard size for arm64 control plane

Prashanth684:
#43677 made a change to have a custom 6vcpu,16G VM for masters. Custom sizes are not supported for arm64 VMs. Let's keep it at `standard-4` and if we see issues with high CPU alerts, we can change it to `standard-8`.


Prashanth684:
/pj-rehearse periodic-ci-openshift-multiarch-master-nightly-4.14-ocp-e2e-gcp-ovn-arm64

PR #43771, opened 2023-09-27
Stackrox: Add sensor profiling prow job

fredrb:
This job should call sensor profiling job, that requires OCP and kube-burner.

Not sure if it makes sense adding for all three OCP versions, maybe just one would be sufficient?


gavin-stackrox (commented):
Until you get the stackrox side working I would skip nightlies. Also set `always_run: false` on all configs. Whatabout GKE master.yaml?

PR #43775, opened 2023-09-27
Drop the kdump step from gcp-ovn-rt upgrade where it seems broken

dgoodwin:


dgoodwin:
/pj-rehearse ack

just removing steps

PR #43794, opened 2023-09-28
USHIFT-1588: Remove rebase-payload

pacevedom:
Remove the rebase-payload image by handling dependencies in the CI container where it runs, instead of preparing them upfront.
With this every test job will save around 5min on average, since there is no need to build an image that its not going to be used. Rebase job is also optimized as a byproduct.

### DIFF
--- a/ci-operator/config/openshift/microshift/openshift-microshift-main.yaml
+++ b/ci-operator/config/openshift/microshift/openshift-microshift-main.yaml
@@ -18,24 +18,6 @@ build_root:
     tag: rhel-9-release-golang-1.20-openshift-4.14
 images:
 - dockerfile_literal: |
-    FROM src
-    RUN dnf install -y python3-pip && \
-      pip3 install --upgrade pip && \
-      pip3 install setuptools-rust cryptography pyyaml pygithub gitpython
-    ENV YQ_URL=https://github.com/mikefarah/yq/releases/download/v4.26.1/yq_linux_amd64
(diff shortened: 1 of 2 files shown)


pacevedom:
/pj-rehearse pull-ci-openshift-microshift-main-test-rebase pull-ci-openshift-microshift-release-4.14-test-rebase

ci-operator/step-registry/openshift/microshift/rebase/openshift-microshift-rebase-ref.yaml
```diff
@@ -1,6 +1,6 @@
 ref:
   as: openshift-microshift-rebase
-  from: rebase-payload
+  from: src
```
pmtk, line 3: Is `rhel-9-release-golang-1.20-openshift-4.14` base for `src`?
I'm thinking ahead when we bump that and perhaps some day common code in `commands.sh` will need some tweaking (or some python libraries get updated so our code will work with older but not with newer), but we can deal with that in (far?) future

pmtk:
/lgtm
/hold

Unhold when convenient

ci-operator/step-registry/openshift/microshift/rebase/openshift-microshift-rebase-ref.yaml
pacevedom: Yes, `src` image comes from `root` image, which in turn comes from `build_root`, which is updated everytime we get a new go version (or ocp version, for that matter).

PR #43801, opened 2023-09-28
build terraform-provider periodic job with latest build from main branch

amalykhi:


amalykhi:
@nirarg @sagidayan @tshefi @AsherShoshan

PR #43804, opened 2023-09-28
NO ISSUE: Ushift should capature osbuild-worker/composer logs

copejon:
ISO and rpm-ostree images can fail to build during the metal host tests job.  Build logs are already captured, but do not usually contain useful information for failed builds. When builds fail, the build logs will only contain "The compose result is empty."

### DIFF
--- a/ci-operator/step-registry/openshift/microshift/infra/iso-build/openshift-microshift-infra-iso-build-commands.sh
+++ b/ci-operator/step-registry/openshift/microshift/infra/iso-build/openshift-microshift-infra-iso-build-commands.sh
@@ -41,12 +41,16 @@ cp /tmp/pull-secret "${PULL_SECRET}"
 
 cd ~/microshift
 
-./test/bin/ci_phase_iso_build.sh
+collect_osbuild_logs(){
+    mkdir /tmp/osbuilder-logs
+    workers_services=$(sudo systemctl list-units | awk '/osbuild-worker@/ {print $1} /osbuild-composer\.service/ {print $1}')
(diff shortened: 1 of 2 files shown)


pmtk:
/pj-rehearse pull-ci-openshift-microshift-main-microshift-metal-tests

ci-operator/step-registry/openshift/microshift/infra/iso-build/openshift-microshift-infra-iso-build-commands.sh
```diff
@@ -36,12 +36,19 @@ cp /tmp/ssh-privatekey ~/.ssh/id_rsa
...
 
 cd ~/microshift
 
 ./test/bin/ci_phase_iso_build.sh
+
+mkdir /tmp/osbuilder-logs
+workers_services=$(sudo systemctl list-units | awk '/osbuild-worker@/ {print $1} /osbuild-composer\.service/ {print $1}')
+while IFS= read -r service; do
+    journalctl -u "${service}" &> /tmp/osbuilder-logs/${service}.log
+done <<<"${workers_services}"
```
pacevedom, line 50: Can we add this to the ci_phase_iso_build in case it fails so that it drops it to the same dir where all output goes? That way it gets automatically picked up by the trap in line 68.

copejon: Sounds good. Will close and open the PR on openshift/microshift.

PR #43816, opened 2023-09-28
image-mirroring: Drop `driver-toolkit` from OKD 4.15

LorbusChris:


core-services/image-mirroring/openshift/mapping_origin_4_15
```diff
@@ -128,7 +128,6 @@ registry.ci.openshift.org/origin/4.15:descheduler quay.io/openshift/origin-desch
 registry.ci.openshift.org/origin/4.15:docker-builder quay.io/openshift/origin-docker-builder:4.15 quay.io/openshift/origin-docker-builder:4.15.0 quay.io/openshift/origin-docker-builder:latest
 registry.ci.openshift.org/origin/4.15:docker-registry quay.io/openshift/origin-docker-registry:4.15 quay.io/openshift/origin-docker-registry:4.15.0 quay.io/openshift/origin-docker-registry:latest
 registry.ci.openshift.org/origin/4.15:dpu-network-operator quay.io/openshift/origin-dpu-network-operator:4.15 quay.io/openshift/origin-dpu-network-operator:4.15.0 quay.io/openshift/origin-dpu-network-operator:latest
-registry.ci.openshift.org/origin/4.15:driver-toolkit quay.io/openshift/origin-driver-toolkit:4.15 quay.io/openshift/origin-driver-toolkit:4.15.0 quay.io/openshift/origin-driver-toolkit:latest
```
ybettan, line 131: Who is using those images?
DTK's customers are usually getting the image using `oc adm release info ... --image-for driver-toolkit` and getting an image from `quay.io/openshift-release-dev/...`

core-services/image-mirroring/openshift/mapping_origin_4_13
```diff
@@ -120,7 +120,6 @@ registry.ci.openshift.org/origin/4.13:descheduler quay.io/openshift/origin-desch
 registry.ci.openshift.org/origin/4.13:docker-builder quay.io/openshift/origin-docker-builder:4.13 quay.io/openshift/origin-docker-builder:4.13.0
 registry.ci.openshift.org/origin/4.13:docker-registry quay.io/openshift/origin-docker-registry:4.13 quay.io/openshift/origin-docker-registry:4.13.0
 registry.ci.openshift.org/origin/4.13:dpu-network-operator quay.io/openshift/origin-dpu-network-operator:4.13 quay.io/openshift/origin-dpu-network-operator:4.13.0
-registry.ci.openshift.org/origin/4.13:driver-toolkit quay.io/openshift/origin-driver-toolkit:4.13 quay.io/openshift/origin-driver-toolkit:4.13.0
```
jlebon, line 123: I don't know the specific details here on what's the right call, but high-level I'm thinking of this change to drop machine-os-content as affecting 4.15 and onwards only.

ybettan: I don't see what we would bother to remove those if `machine-os-content` is shipped in older versions anyway.

jlebon: Interested in what other stakeholders think, but IMO it doesn't seem worth the risk. Even if we did decide to do this in older releases, we should validate it in 4.15 first.

core-services/image-mirroring/openshift/mapping_origin_4_15
ybettan: Before merging this PR, we need to make sure that OKD's users running `oc adm release info ... --image-for driver-toolkit` won't get those images. If this is the case then we shouldn't remove them.

core-services/image-mirroring/openshift/mapping_origin_4_13
LorbusChris: The driver toolkit image that is in OKD right now is actually an RHCOS one from Prow CI. It shouldn't be there and if nothing's image-referencing it, removing it from all current OKD versions (4.13+) should be safe afaict.

Removing `driver-toolkit` from 4.15 is required before switching the image-reference from `machine-os-content` to `rhel-coreos`, as the release controllers for OKD (and required CI tests like `/test okd-images`) will otherwise fail due to the OKD ImageStreams not containing and image of that name.

To be extra cautious, I can split this and do only 4.15 first. Please let me know.

LorbusChris: On a related note, is it too late to switch to a more agnostic image name, like simply `coreos`? That would make separate builds for OCP and OKD for some components like the MCO superfluous.

LorbusChris: https://github.com/openshift/release/pull/43957

core-services/image-mirroring/openshift/mapping_origin_4_15
LorbusChris: Customers shouldn't be using any OKD images unless they specifically want to do that. OKD won't have the driver-toolkit going forward, unless OKD-variant specific builds of that image are set up.

core-services/image-mirroring/openshift/mapping_origin_4_13
ybettan: > To be extra cautious, I can split this and do only 4.15 first. Please let me know.

Can't we just remove it for 4.15? Is it an issue to keep machine-os-content in older versions?

> On a related note, is it too late to switch to a more agnostic image name, like simply coreos? That would make separate builds for OCP and OKD for some components like the MCO superfluous.

You mean instead of `rhel-coreos`? I don't mind as long as DTK's image-references is updated accordingly but I won't mix those 2 efforts anyway. Let's keep this PR simple IMO.

LorbusChris: Yes `rhel-coreos`. We're doing separate OKD / OCP builds for a couple of components, where the only difference is the base OS image-reference they contain (see e.g. https://github.com/openshift/machine-config-operator/blob/master/Dockerfile#L24-L35)

jlebon:
This looks sane to me. Thanks for working on this!
/approve

Will let @ybettan do the final stamping.

Re. naming, I think a better approach is probably to not care about the name at all and just focus on the `machine-os` label bit. But yeah, let's keep that discussion separate from this effort. :) (Probably worth a ticket somewhere... against the MCO maybe?)

core-services/image-mirroring/openshift/mapping_origin_4_13
ybettan: Then I am OK with it. Probably better to discuss it with the coreOS folks.
Is it related to this PR though?

PR #32573, opened 2022-09-23
adding more valid labels for openshift-docs

kalexand-rh:
I circulated our label-based workflow proposal (based on #31820 and #32326), and we need to add two more labels to the `openshift-docs` repo to allow more writers who don't have edit rights to label their PRs.

This set of label changes is part of an elaborate workaround for the lack of `triage` role in our GitHub org.

PR #43820, opened 2023-09-28
adding the cloud-experts label to openshift-docs

kalexand-rh:
As an elaborate workaround for the lack of `triage` role in our GitHub org, the `openshift-docs` repo uses a custom set of labels that can be applied with the `label` CI command to allow certain contributors who don't have edit rights to label their PRs.

We need to add another. This PR is similar to https://github.com/openshift/release/pull/32573.


kalexand-rh:
@psalajova, will you add a LGTM to this one so tide will merge it?

PR #43829, opened 2023-09-29
Stackrox: stop Sensor Profile from running in PRs that change sensor code

fredrb:
There was a mistake in the previous PR (#43771) which makes this job run in PRs that are changing sensor code. I assume this is because of the `run_if_changed` config that was left there.

I'm removing them until the job is stable.


tommartensen:
@fredrb You'd need to add the `rehearsals-ack` label to start the merge.

PR #43838, opened 2023-09-29
Revert "Mark build09 jobs as optional"

bear-redhat:
Reverts openshift/release#43818


smg247:
The last couple of runs are green!
/hold cancel

PR #43845, opened 2023-09-29
[LPTOCPCI-786] Remove branch protection for mtc-qe

calebevans:


smg247:
I believe you need to use `unmanaged: true` instead of `protect: false`

smg247:
Thank you!
/lgtm

PR #43836, opened 2023-09-29
OpenStack: Ensure workflows deprovision credentials

mandre:
For all workflows that include the `openstack-conf-clouds` ref in its `pre` section, also include the `openstack-deprovision-credentials` ref in its `post` section to ensure that created application credentials are cleaned up.

PR #43847, opened 2023-09-29
openstack: Replace OWNERS symlink

pierreprinetti:
We're symlinking too much for Prow.

see https://github.com/openshift/release/pull/43836


pierreprinetti:
dup of https://github.com/openshift/release/pull/43373

PR #43868, opened 2023-10-02
MGMT-15888: build for s390x at upstream images

osherdp:
Now that we're based on stream9, we can also build for this arch.


adriengentil (approved):
/cc @AmadeusPodvratnik

PR #43875, opened 2023-10-02
Split files for the oadp targets to include ocp versions

mpryc:
Such split will allow us easier compare jobs and create relevant new jobs for the new ocp releases.


mpryc:
@mateusoliveira43 @mrnold FYI

ci-operator/jobs/openshift/oadp-operator/openshift-oadp-operator-master-ocp-4.10-presubmits.yaml
```diff
@@ -0,0 +1,399 @@
+presubmits:
+  openshift/oadp-operator:
+  - agent: kubernetes
+    always_run: false
+    branches:
```
mateusoliveira43, line 5: Can we instead of having a file for each branch, add all branches needed for each ocp version file? (the periodics syntax is different, maybe just for pre and post submits)

mpryc: i closed this PR, it's wrong way to go as `make jobs` is generating the jobs, so we should not touch them

PR #43879, opened 2023-10-02
Experiment using INT TERM instead of EXIT for trap

xueqzhan:
Test Platform team suggests using INT TERM instead of EXIT to avoid observers existing with non-zero error codes.


xueqzhan:
/pj-rehearse periodic-ci-openshift-release-master-nightly-4.15-e2e-aws-sdn-upgrade

PR #43894, opened 2023-10-03
Change cron start time to trigger ocp-self-managed-lp-watcher-bot-message

dfrazzette:
Changing cron time from 20:00 UTC to 14:00 UTC on Mondays to trigger ocp-self-managed-lp-watcher-bot-message.


ascerra:
the generated-config job is failing here because it is expecting the job that is created by the config you changed to have the update cron change.
- if you run `make update` the prow job will be automatically updated.

ascerra:
Could you also include an update to the rosa-sts-hypershift watcher bots cron job?

[This line here](https://github.com/openshift/release/pull/43894/files#diff-c3a56393fda719977c318d3e43a3ca3fb148e0447d3a4fea5df3dd6e7149f388R55) can be changed from `0 20 * * 1` to `0 18 * * 1`

ascerra (approved):

PR #43897, opened 2023-10-03
Rename openstack-machine-controllers to openstack-cluster-api-controllers

mdbooth:
This image no longer provides a Machine API actuator. That role is now fulfilled by machine-api-provider-openstack. This repository now hosts an upstream CAPI provider and should be named consistently with:

* aws-cluster-api-controllers
* azure-cluster-api-controllers
* gcp-cluster-api-controllers
* ibmcloud-cluster-api-controllers
* vsphere-cluster-api-controllers


stephenfin:
Failure looks real. Have pushed a small follow-up patch to update the mappings also.

PR #43907, opened 2023-10-03
`crier` uses more memory than current request

smg247:
The `crier` pod keeps getting evicted due to using too much memory.
```
The node was low on resource: memory. Threshold quantity: 100Mi, available: 44080Ki. Container crier was using 5852220Ki, request is 4Gi, has larger consumption of memory.
```


smg247:
/close
We are going to use apps-auth again.

smg247:
/reopen
crier still uses more memory than we are requesting

PR #43916, opened 2023-10-03
Update cron and test name for conc tests.

svetsa-rh:
1. Both 4.13 and 4.14 are currently running on same day. So, updating cron for each version to run every other week instead.
2. Names are slightly inconsistent with our naming convention. Updating same.


svetsa-rh:
@lenahorsley fyi

PR #43950, opened 2023-10-04
MGMT-15888: remove multi-images publish for ACM branches

osherdp:
After https://github.com/openshift/release/pull/43868 got in, it would mean that we cannot use the same code in order to build for linux/s390x, as it's not supported for `quay.io/centos/centos:stream8`.

It seems unneeded to build those images if they are not being used (as for those branches we separately build the images, bundles and indices). So this removes those jobs from existance.


gamli75:
@filanov do you know if ACM are using the multi-arch images?

filanov:
does multi arch mean using ARM? if so then yes.

PR #43957, opened 2023-10-04
image-mirroring: Drop `driver-toolkit` from OKD 4.{13,14}

LorbusChris:


core-services/image-mirroring/openshift/mapping_origin_4_14
```diff
@@ -128,7 +128,6 @@ registry.ci.openshift.org/origin/4.14:descheduler quay.io/openshift/origin-desch
 registry.ci.openshift.org/origin/4.14:docker-builder quay.io/openshift/origin-docker-builder:4.14 quay.io/openshift/origin-docker-builder:4.14.0
 registry.ci.openshift.org/origin/4.14:docker-registry quay.io/openshift/origin-docker-registry:4.14 quay.io/openshift/origin-docker-registry:4.14.0
 registry.ci.openshift.org/origin/4.14:dpu-network-operator quay.io/openshift/origin-dpu-network-operator:4.14 quay.io/openshift/origin-dpu-network-operator:4.14.0
-registry.ci.openshift.org/origin/4.14:driver-toolkit quay.io/openshift/origin-driver-toolkit:4.14 quay.io/openshift/origin-driver-toolkit:4.14.0
```
ybettan, line 131: Why do we care about older versions? Can't we just keep machine-os-content for older version of OCP and only remove it in newer versions?

LorbusChris: machine-os-content is not touched here. This change removes an OCP CI image that has been leaking into OKD releases (driver-toolkit).

PR #43988, opened 2023-10-04
Set XDG_DATA_HOME for CAPO test target

mdbooth:
https://github.com/openshift/cluster-api-provider-openstack/pull/265 fails because it doesn't have a location to write envtests assets to. This provides the location without requiring a carry patch to the Makefile.


mdbooth:
Test failed successfully

/pj-rehearse skip

PR #43992, opened 2023-10-05
shiftstack/ci-configs: remove useless test

EmilienM:
provision test is useless and not working, it was a legacy that can be
removed now.


mandre (commented):
/lgtm
/approve

PR #44071, opened 2023-10-07
remove cucushift-installer-check ref step

jianlinliu:
`cucushift-installer-check` is almost the same as `cucushift-installer-check-info`, which is already added into most install workflow, so removing it.


jianlinliu:
cc @liangxia to review.

PR #44077, opened 2023-10-08
remove gather-installer-boostrap-logs-in-bastion step from non-private install

jianlinliu:


jianlinliu:
cc @jinyunma and @jianli-wei to review

PR #44080, opened 2023-10-08
Fix several fips jobs

liangxia:
The test names contains `fips`, but missing `FIPS_ENABLED` which is used to enable fips.

/cc @jianlinliu @JianLi-RH @pruan-rht @jhou1


ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.10__amd64-nightly.yaml
```diff
@@ -1419,6 +1419,7 @@ tests:
     cluster_profile: gcp-qe
     env:
       E2E_RUN_TAGS: '@amd64 and @gcp-ipi and @network-openshiftsdn'
+      FIPS_ENABLED: "true"
```
jianlinliu, line 1422: The workflow already enabled it, because rhel workers need some special steps for fips enabling.

PR #44105, opened 2023-10-09
prowconfig(osso): enable tide merging for 4.13 and 4.14

ingvagabund:
OSSO is built by CPaaS which is released async of OCP. There are different merging and releasing policies.


soltysh (approved):
/lgtm

PR #44015, opened 2023-10-05
OpenStack: Remove Kuryr jobs

dulek:
Kuryr is deprecated and we're removing it in 4.15. This removes all the jobs that test with `NETWORK_TYPE: Kuryr`.

PR #44117, opened 2023-10-09
OpenStack: Remove Kuryr jobs vol. 2

dulek:
PR #44015 needed to be reverted as it broke other jobs. We need to remove Kuryr from CNO before it's removed from the payload. But in order to remove it from CNO should remove the other Kuryr jobs. This commit does so.


dulek:
/cc @sdodson 
/cc @bparees

Can you take a look here folks? This is a replacement for the reverted #44015 and this time it doesn't remove the image yet, just the jobs in the other repositories.

soltysh (approved):
/approve

PR #44126, opened 2023-10-09
Add AAP to GitHub ignore check list

liswang89:
Since AAP doesn't run the automation from the repo, we want to add it to the ignore list to skip check-gh-automation.


smg247:
This repo will not be using prow.
/approve
/hold cancel

PR #44136, opened 2023-10-10
OCPQE-16721: Skip failing in the RP's overall CI test suite after pre-checks and before the extended-platform-tests run

aleskandro:
As of https://gitlab.cee.redhat.com/aosqe/prow-test-results-classfier/-/merge_requests/194, the data pipeline importing results from Prow to RP also creates a test suite to track the overall CI pipeline results and verify the test chain's steps are executing as expected.

The overall CI test suite processing creates a test case for each step in the `test` chain. The test case is considered successful if the related step exits with 0. However, some steps are processed more fine-grained and should be allowed to fail from a given point in their execution.

This PR lets the openshift-extended steps tolerate failures after all the pre-checks and preparation statements and before the extended-platform-tests suite starts running. Once it starts running, we consider further failures to be managed through the explicit JUnit XML files. This way, we avoid the need for specific JUnit files for each possible failure that is hard to predict, as they can depend on product bugs, automation issues, or other reasons.

The final goal is to prevent information loss when data move from Prow to ReportPortal, leading to watchers not being notified of failures that could need a fix in the products, in the automation, and such.

Refers [OCPQE-16721](https://issues.redhat.com//browse/OCPQE-16721) [OCPQE-16722](https://issues.redhat.com//browse/OCPQE-16722)

Finally, this PR proposes a few minor fixes in handling the `FORCE_SUCCESS_EXIT` variable as it can affect whether a step should be considered successful or not, and some test configs that set it to a "true" value with no need.

This PR depends on https://gitlab.cee.redhat.com/aosqe/prow-test-results-classfier/-/merge_requests/194

/hold

### DIFF
--- a/ci-operator/step-registry/openshift/e2e/test/qe/report/openshift-e2e-test-qe-report-commands.sh
+++ b/ci-operator/step-registry/openshift/e2e/test/qe/report/openshift-e2e-test-qe-report-commands.sh
@@ -6,6 +6,9 @@ set -o pipefail
 
 EXIT_STATUS=0
 
+# the exit code of this step is not expected to be caught from the overall test suite in RP. Excluding it
+touch "${ARTIFACT_DIR}/skip_overall_if_fail"
+
 # only exit 0 if junit result has no failures
 if [[ -f "${SHARED_DIR}/openshift-e2e-test-qe-report-openshift-extended-test-failures" ]]; then
     cat "${SHARED_DIR}/openshift-e2e-test-qe-report-openshift-extended-test-failures"


ci-operator/step-registry/openshift-extended/test/openshift-extended-test-commands.sh
```diff
@@ -322,6 +317,13 @@ function run {
     else
         echo "${failures} failures in openshift-extended-test" | tee -a "${SHARED_DIR}/openshift-e2e-test-qe-report-openshift-extended-test-failures"
     fi
+    # it ensure the the step after this step in test will be executed per https://docs.ci.openshift.org/docs/architecture/step-registry/#workflow
+    # please refer to the junit result for case result, not depends on step result.
+    if [ "W${FORCE_SUCCESS_EXIT}W" == "WnoW" ]; then
+        echo "do not force success exit"
+        exit $ret_value
+    fi
+    return 0
```
kuiwang02, line 326: @aleskandro it works, but if we align code with other step, could remove this line.
it is up to you to keep it or not.

kuiwang02:
I am ok for changing besides only one minor comment.

ci-operator/step-registry/openshift-extended/test/openshift-extended-test-commands.sh
aleskandro: yeah, let me do it in a while... There should be even a few others actually: I'd prefer keeping `return` statements vs `exit` in functions... But it could require some time to verify compliance with the current logic.

kuiwang02: ok to keep it. thanks for the information.

aleskandro: I meant that I prefer having the return statements vs exit within functions when possible... But to do that change consistently, we need a few more time to avoid unexpected behavior in the test configs. 

I removed the `return 0` for now, as you recommended.

aleskandro:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-arm64-nightly-gcp-ipi-sdn-f28

aleskandro:
I'm testing a job to verify that the files are correctly created and kept in the storage (although empty).

I'll add ack to rehearsals if it passes. In the meantime, this can be considered ready to merge as the dependent PR is now merged.

/unhold

aleskandro:
Hi @jhou1 @liangxia could you approve the change for the qe-report step too?

PR #44141, opened 2023-10-10
Remove the cnf-features-deploy periodic jobs

fedepaol:
These jobs are not notifying anywhere, removing them.


ffromani:
/lgtm

/hold
let's hear from @nishant-parekh if another team depends on these

PR #44153, opened 2023-10-10
[KubeVirt/NMO] Remove obselete jobs configuration and change cluster profile

orenc1:
The last supported CNV version that includes NMO is 4.11, and it is built from master branch of https://github.com/kubevirt/node-maintenance-operator.git. Removing all lanes besides that.
Updating the OCP used to 4.11.
Changing the cluster_profile used to aws-virtualization, so CI on that repo will be charged by CNV cost center, not openshift.


slintes:
Oh, the e2e error is a known issue actually, caused by a change in OCP 4.11, which we fixed a while ago in the medik8s/nmo repo: https://github.com/medik8s/node-maintenance-operator/pull/27

@orenc1 Does it need to be ported to kubevirt/nmo?
The issue is that currently on OCP 4.11 it's not prevented to put multiple control plane nodes into maintenance.

The CI config is ok though.

/pj-rehearse ack
/lgtm

PR #44167, opened 2023-10-10
Switch CAPO to use DOWNSTREAM_OWNERS

mdbooth:
TODO:
- [x] Merge https://github.com/openshift/cluster-api-provider-openstack/pull/269

/hold

### DIFF
--- a/core-services/prow/02_config/_plugins.yaml
+++ b/core-services/prow/02_config/_plugins.yaml
@@ -683,6 +683,9 @@ owners:
     openshift/aws-load-balancer-controller:
       owners: DOWNSTREAM_OWNERS
       owners_aliases: DOWNSTREAM_OWNERS_ALIASES
+    openshift/cluster-api-provider-openstack:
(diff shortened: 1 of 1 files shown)


mandre:
If I understand correctly, with this patch, we can no longer merge in CAPO, except for main (thanks to https://github.com/openshift/cluster-api-provider-openstack/pull/269). Shouldn't we have DOWNSTREAM_OWNERS in all branches before merging this patch?

mandre:
> If I understand correctly, with this patch, we can no longer merge in CAPO, except for main (thanks to [openshift/cluster-api-provider-openstack#269](https://github.com/openshift/cluster-api-provider-openstack/pull/269)). Shouldn't we have DOWNSTREAM_OWNERS in all branches before merging this patch?

Answering to myself: we switched to MAPO in 4.11, and OCP 4.10 is now EOL, which means we no longer have legacy CAPO branches to maintain. We're fine.
That being said, we'll have to be careful when we make the same change for cloud-provider-openstack.

PR #44171, opened 2023-10-10
Hive: Make e2e-{gcp|azure} mandatory if run

2uasimojo:
Previously: e2e-gcp and e2e-azure would only run if affected files contained substring `gcp` or `azure`, respectively. However, if they did run, they would not block merge if they failed.

Goal: e2e-gcp and e2e-azure only run if affected files contain substring `gcp` or `azure`, respectively; *and in that scenario they are mandatory* -- i.e. they block merge if they fail.

### DIFF
--- a/ci-operator/config/openshift/hive/openshift-hive-master.yaml
+++ b/ci-operator/config/openshift/hive/openshift-hive-master.yaml
@@ -166,7 +166,6 @@ tests:
           cpu: 100m
     workflow: generic-claim
 - as: e2e-gcp
-  optional: true
   run_if_changed: gcp
   steps:
     cluster_profile: gcp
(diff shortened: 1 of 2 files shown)


2uasimojo:
/pj-rehearse skip

We didn't actually change anything that should affect test success/failure, only the PR triggers that cause them to run. So no point rehearsing the jobs. If we pass overall release CI, we'll just have to merge this and test it out from the repo itself.

2uasimojo:
/assign @suhanime 

If this works, we'll need to push it back into active mce branches as well.

2uasimojo:
Validated via https://github.com/openshift/hive/pull/2129

PR #44185, opened 2023-10-10
update hypershift addon operator to use go 1.20

rokej:
update hypershift addon operator to use go 1.20


mikeshng (approved):
/approve

/lgtm

did a search on 18 and didn't find any other occurences.

PR #44199, opened 2023-10-11
test: check machinesets naming

pawelpaszki:
## WHAT
test: check machinesets naming

## LOCAL RUN OUTPUT
```
./machinesets.sh
Getting first available name of machineset to confirm that its valid
Confirming that the suffix of the machineset name: 'hs-mc-20bivna6g-wh8nq-serving-1-us-east-1a' doesn't include too many dashes - indicating double region in its name

Passed tests:

OCPQE-17159

Failed tests:
```

### DIFF
--- a/ci-operator/step-registry/osd-ccs/fleet-manager/test/osd-ccs-fleet-manager-test-commands.sh
+++ b/ci-operator/step-registry/osd-ccs/fleet-manager/test/osd-ccs-fleet-manager-test-commands.sh
@@ -496,7 +496,7 @@ function test_endpoints () {
 
 ##################################################################
 
-###### /audit endpoint tests (OCPQE-17072) ######
+###### /audit endpoint tests (OCP-67823) ######
 
 function test_audit_endpooint () {
 TEST_PASSED=true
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/osd-ccs/fleet-manager/test/osd-ccs-fleet-manager-test-commands.sh
```diff
@@ -624,6 +624,38 @@ TEST_PASSED=true
...
+    # if there are more than 4 dashes in the suffix, the name likely contains duplicated AZ in its name, e.g. 'us-east-2a-us-east-2a'
+    NUMBER_OF_DASHES=$(grep -o '-' <<<"$SUFFIX" | grep -c .)
+    if [ "$NUMBER_OF_DASHES" -gt 4 ]; then
+      echo "Incorrect machineset name detected: $MACHINE_SETS_OUTPUT"
+      TEST_PASSED=false
+    fi
+  else
+    echo "No machinesets found."
+  fi
+  update_results "OCPQE-17159" $TEST_PASSED
```
jianping-shu, line 654: Normally we use test case Id (like polarion case Id) instead of ticket Id, we'd better to use this consistently.

ci-operator/step-registry/osd-ccs/fleet-manager/test/osd-ccs-fleet-manager-test-commands.sh
```diff
@@ -624,6 +624,38 @@ TEST_PASSED=true
 
 ###### end of /audit endpoints tests (OCPQE-17072) ######
 
+##################################################################
+
+###### machinesets naming test (OCPQE-17159) ######
```
jianping-shu, line 629: Same comment

ci-operator/step-registry/osd-ccs/fleet-manager/test/osd-ccs-fleet-manager-test-commands.sh
pawelpaszki: thanks! I have addressed your comment. PTAL

PR #44205, opened 2023-10-11
feat(rhtap-load-test): Use dedicated scenarios for max-concurrency daily jobs

pmacik:
The daily jobs don't need to turn tekton performance profiling on so I separated the `max-concurrency` scenarios for daily jobs (without the tekton profiling) and kept the PR jobs with the original scenarios.


jhutar (approved):
/lgtm

PR #44232, opened 2023-10-11
create azurestack clusterbot job

rvanderp3:


elmiko (commented):
i'm not an expert with respect to azure stack, but this looks correct to me.

/lgtm

also, fwiw, we are currently experiencing an installation bug with azurestack on 4.14+. see https://issues.redhat.com/browse/OCPBUGS-20213

hoxhaeris (approved):

PR #44247, opened 2023-10-11
Use firewatch group feature to prevent cascading issue.

ascerra:
# Reason for change
Many Jira tickets have been created for this scenario calling out issues that were only due to an issue failing prior to the one in the ticket. With these changes we are going to lean on firewatch's group feature to hopefully prevent cascading issues from being created as their own separate issue.

# explanation of expected behavior

- Lines 47-48 both have the cluster group set to priority 1.
  - This means that if no matter what if those steps fail a firewatch ticket will be created.
    - If they both fail I'd expect 2 firewatch tickets to be created.
- Line 49 has the cluster group with priority 2 so this should mean that if a step with the cluster group priority 1 fails (line 48 for example) and the step on line 49 fails then we will not create a ticket for line 49 and only a ticket will be created to the failed step with the higher priority.


calebevans (approved):
/lgtm

PR #44258, opened 2023-10-11
okd scos periodic missing upi-installer

jcpowermac:


jcpowermac:
@vrutkovs @LorbusChris I keep seeing errors on the new vsphere build cluster about missing this image. I _think_ this fixes it.

/pj-rehearse periodic-ci-openshift-release-master-okd-scos-4.14-e2e-vsphere-ovn

ci-operator/config/openshift/release/openshift-release-master__okd-scos-4.12.yaml
```diff
@@ -1,3 +1,8 @@
+base_images:
+  upi-installer:
+    name: "4.12"
```
LorbusChris, line 3:
```suggestion
    name: "scos-4.12"
```
same change is needed on all of the others too

LorbusChris (changes requested):
@jcpowermac thank you for taking a look at this.
I had noticed that too, and discovered we weren't building the upi-installer image for SCOS.

Builds for 4.14 and 4.15 were added in https://github.com/openshift/release/pull/43624, so it'll only work with these versions.

You'll have to reference the `scos-4.1x` ImageStream though for it to point to the correct image.

I'd propose to drop the change for 4.13 and below. I'll open a separate PR to drop any vsphere tests on these OKD/SCOS versions.

LorbusChris:
@jcpowermac I hope you don't mind that I've folded this into https://github.com/openshift/release/pull/44273

/close

PR #44261, opened 2023-10-12
extend worker pool update wait time

jiajliu:
https://issues.redhat.com/browse/OCPQE-16745

### DIFF
--- a/ci-operator/step-registry/cucushift/upgrade/paused/unpause/cucushift-upgrade-paused-unpause-ref.yaml
+++ b/ci-operator/step-registry/cucushift/upgrade/paused/unpause/cucushift-upgrade-paused-unpause-ref.yaml
@@ -4,7 +4,7 @@ ref:
     namespace: ci
     name: verification-tests
     tag: latest
-  timeout: 1h
+  timeout: 1h10m
   grace_period: 10m
   commands: cucushift-upgrade-paused-unpause-commands.sh
   cli: latest


jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-eus-4.12-gcp-ipi-workload-identity-rt-fips-f14

jiajliu:
Extend time works for the failed job from the test result.

```
INFO[2023-10-12T10:04:50Z] Running step gcp-ipi-workload-identity-rt-fips-f14-cucushift-upgrade-paused-unpause. 
INFO[2023-10-12T11:02:16Z] Step gcp-ipi-workload-identity-rt-fips-f14-cucushift-upgrade-paused-unpause succeeded after 57m26s. 
```
cc @jianlinliu @shellyyang1989 for a review, thanks!

ci-operator/step-registry/cucushift/upgrade/paused/unpause/cucushift-upgrade-paused-unpause-commands.sh
```diff
@@ -41,7 +41,7 @@ function check_mcp() {
         if [[ ${updated} == "True" && ${updating} == "False" && ${degraded} == "False" ]]; then
             echo "Worker pool status check passed" && return 0
         fi  
-        sleep 60
+        sleep 120
```
jianlinliu, line 44: If that, the max time would be 1 hours, for this function, while seem like `cucushift-upgrade-paused-unpause`'s timeout is 1 hour, shall we reduce the max mcp check timeout a bit, or increase the `cucushift-upgrade-paused-unpause`'s timeout a bit?

jiajliu: yes, I extend the total time to be 1 hour intentionally. From the mco log I dig, the successful updated worker node last about 16min-18min(from cordon to completely apply), considering worker node is updated in serial, so i think leave 20min per node should be enough.

jianlinliu: > From the mco log I dig, the successful updated worker node last about 16min-18min(from cordon to completely apply), considering worker node is updated in serial, so i think leave 20min per node should be enough.

Agree.  Per my understanding, the `check_mcp` should be a part of the whole `cucushift-upgrade-paused-unpause` step, shall we extend the [timeout](https://github.com/openshift/release/blob/3f0647de22e3fbba8ff979907347fe2e18ce7637/ci-operator/step-registry/cucushift/upgrade/paused/unpause/cucushift-upgrade-paused-unpause-ref.yaml#L7) accordingly?

jiajliu: Ah, got u now, good catch! Let me extend the timeout accordingly. From the test log, this step cost `57m26s`, let's update it to 1h10m, wdyt?

jianlinliu: How about 2h directly?

jiajliu: I thought the timeout in ref should be final gate on some unexpected error, which may get the test hang there. Maybe more shorter more better to exit early?

jianlinliu: okay. agree~

PR #44265, opened 2023-10-12
[4.15] migrate some ovn/sdn migration jobs to multi-arch ci

jianlinliu:
Migrate ovn/sdn migration jobs to multi-arch ci
1. save cloud cost
2. increase multi-arch coverage
3. be easier to maintain the ci jobs in one central place

Also migrate `aws-ipi-preserve-ign` together, and remove e2e testing to save running time (the `preserve-ign` feature never has impact to e2e testing)

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__amd64-nightly.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__amd64-nightly.yaml
@@ -1069,19 +1069,6 @@ tests:
     test:
     - chain: openshift-upgrade-qe-test-heterogeneous
     workflow: cucushift-installer-rehearse-aws-ipi-ovn-ipsec
-- as: aws-ipi-ovn-migration-sdn-f14
-  cron: 2 14 14,29 * *
-  steps:
-    cluster_profile: aws-qe
-    env:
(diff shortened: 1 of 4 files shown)


jianlinliu:
cc @zhaozhanqi @anuragthehatter @yunjiang29 to review.

zhaozhanqi:
do we need to test on both arch platform?  for save resource, I think one is ok, eg, we can keep `aws-ipi-ovn-migration-sdn-arm-f28`

aws-ipi-ovn-migration-sdn-amd-f28
aws-ipi-ovn-migration-sdn-arm-f28

jianlinliu:
> Do we need to test on both arch platform? for save resource, I think one is ok

I added the two jobs here just for easily aligning with other jobs as a pair of job. Before it was `aws-ipi-ovn-migration-sdn-f14` running on amd, now it is split it into two jobs and f28, the total number did not change, also use more arm arch instances, which is cheaper than amd. So it is okay for now.

zhaozhanqi:
> > Do we need to test on both arch platform? for save resource, I think one is ok
> 
> I added the two jobs here just for easily aligning with other jobs as a pair. Before it was `aws-ipi-ovn-migration-sdn-f14` running on amd, now it is split it into two jobs and f28, the total number does not change, also use more arm arch instances, which is cheaper than amd. So it is okay for now.

ok, thanks. then LGTM.

jianlinliu:
cc @liangxia to review

PR #44268, opened 2023-10-12
WIP: test base images

jsafrane:
It should be taken from the repo, so we may not need to define it here.


jsafrane:
/pj-rehearse pull-ci-openshift-aws-ebs-csi-driver-operator-master-images

PR #44276, opened 2023-10-12
[makefile] Fix template-allowlist on ARM

black-dragon74:
Container engine is missing the platform flag and hence the step fails on ARM machines.

This PR adds the appropriate flags.


black-dragon74:
Can anyone take a look at this please?

PR #44289, opened 2023-10-12
Fix ci-operator-prowgen target when WHAT is specified

Tof1973:
Fix bug introduced by https://github.com/openshift/release/pull/43927


Tof1973:
@smg247, can you have a look ?
Thanks.

PR #44318, opened 2023-10-12
platform-operators: use repository configuration for build root 4.15+

joelanford:


joelanford:
/hold

I think golangci-lint may need to be bumped to run successfully with go 1.20.

ncdc (approved):

PR #44321, opened 2023-10-12
Remove compserv CI

rhmdnd:
We're no longer pursuing this PoC, and disabled maintaining dependencies a while ago https://github.com/rhmdnd/compserv/pull/324.

This commit removes it from the ci-config so it's not lingering.


rhmdnd:
Cleaned up conflict and rebased.

PR #44323, opened 2023-10-12
Execute ACS image-push

mclasmeier:
tbh, no idea if this fixes the issue:

```
No image-push entrypoint (.openshift-ci/image-push/entrypoint.sh) found.
```

which is currently being reported for this particular job.

### DIFF
--- a/ci-operator/config/stackrox/acs-fleet-manager/stackrox-acs-fleet-manager-main.yaml
+++ b/ci-operator/config/stackrox/acs-fleet-manager/stackrox-acs-fleet-manager-main.yaml
@@ -91,8 +91,9 @@ tests:
       cli: latest
       commands: |
         entrypoint=".openshift-ci/image-push/entrypoint.sh"
-        if [ -x "$entrypoint" ]; then
+        if [ -e "$entrypoint" ]; then
           echo "Executing image-push entrypoint."
+          chmod +x "$entrypoint"
           $entrypoint
         else
           echo "No image-push entrypoint ($entrypoint) found."


porridge:
I don't think this is it. The file **is** executable:
{noformat}
[mowsiany@mowsiany acs-fleet-manager]$ stat .openshift-ci/image-push/entrypoint.sh 
  File: .openshift-ci/image-push/entrypoint.sh
  Size: 2154      	Blocks: 8          IO Block: 4096   regular file
Device: 253,2	Inode: 132825996   Links: 1
Access: (0775/-rwxrwxr-x)  Uid: (4200660/ UNKNOWN)   Gid: (4200660/ UNKNOWN)
[...]
{noformat}

Also, you probably want to add your new github username to OWNERS since prow seems to think you're a stranger...

mclasmeier:
Well, I'm not convinced either. But, the file also *exists*, so, asking myself what I would consider more likely -- that a file in Git vanishes or that for some reason permissions are mangled -- I would say the latter is more likely. :-/

Regarding the `OWNERS` file, I had the PR open for a few days already, but it got merged with some delay (after opening this one, I believe).

porridge:
> Well, I'm not convinced either. But, the file also _exists_, so, asking myself what I would consider more likely -- that a file in Git vanishes or that for some reason permissions are mangled -- I would say the latter is more likely. :-/

I'd say the former is more likely - I suspect the `from: base` image is missing the mount that contains the checkout (unlike `from: e2e-runtime`).

Also, please keep `stackrox-acs-fleet-manager-stage.yaml` in sync.

PR #44340, opened 2023-10-13
Revert jsonpath in oc adm release info to json

shellyyang1989:
Observing an error in the long-chain [upgrade](https://qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/qe-private-deck/pr-logs/pull/openshift_release/44135/rehearse-44135-periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.3-aws-ipi-f30/1712296596718227456)

```
INFO[2023-10-12T17:14:55Z] Upgrade targets are registry.build04.ci.openshift.org/ci-op-ry24ntqx/release@sha256:eadcddba44e95b1e5d77c57a8f5eda35f3d94e5e0679f900bd5d8dd2cae573f8
error: --output only supports 'name', 'json', 'pullspec', or 'digest'
```

It's because old version `oc adm release info` doesn't support jsonpath

```
# ./oc version
Client Version: version.Info{Major:"4", Minor:"1", GitVersion:"4.1.37-202003021622-d372a16", GitCommit:"d372a16", GitTreeState:"clean", BuildDate:"2020-03-02T19:25:38Z", GoVersion:"go1.11.13", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.16+6981c04", GitCommit:"d328604e81077de3864c29da77f8a7f07e2b6a77", GitTreeState:"clean", BuildDate:"2023-09-12T19:36:27Z", GoVersion:"go1.18.10 X:strictfipsruntime", Compiler:"gc", Platform:"linux/amd64"}

# ./oc adm release info quay.io/openshift-release-dev/ocp-release:4.12.38-x86_64 -o jsonpath='{.metadata.version}'
error: --output only supports 'name', 'json', 'pullspec', or 'digest'
```

So reverting jsonpath to json.

### DIFF
--- a/ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
+++ b/ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
@@ -590,7 +590,7 @@ export PATH=${OC_DIR}:$PATH
 for target in "${TARGET_RELEASES[@]}"
 do
     export TARGET="${target}"
-    TARGET_VERSION="$(env "NO_PROXY=*" "no_proxy=*" oc adm release info "${TARGET}" -o jsonpath='{.metadata.version}')"
+    TARGET_VERSION="$(env "NO_PROXY=*" "no_proxy=*" oc adm release info "${TARGET}" --output=json | jq -r '.metadata.version')"
     extract_oc
 
     SOURCE_VERSION="$(oc get clusterversion --no-headers | awk '{print $2}')"


shellyyang1989:
cc @jianlinliu @jiajliu for review. Thank you.

shellyyang1989:
Refers [OCPQE-17028](https://issues.redhat.com/browse/OCPQE-17028)

jianlinliu:
LGTM

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -590,7 +590,7 @@ export PATH=${OC_DIR}:$PATH
 for target in "${TARGET_RELEASES[@]}"
 do
     export TARGET="${target}"
-    TARGET_VERSION="$(env "NO_PROXY=*" "no_proxy=*" oc adm release info "${TARGET}" -o jsonpath='{.metadata.version}')"
+    TARGET_VERSION="$(env "NO_PROXY=*" "no_proxy=*" oc adm release info "${TARGET}" --output=json | jq -r '.metadata.version')"
```
jiajliu, line 593: wondering what's the correct oc version here, 4.1?

shellyyang1989: The `oc` is extracted from the `latest` image. In upgrade test, it's the source version.

shellyyang1989:
The cli in the fix was what we were using before the change https://github.com/openshift/release/pull/43227/files#diff-a71e3444f4cc5e3d7ecb8ebd36fa8bd1fda181a0806d09523b318eadffaebc71L533-R593. So to save time, I don't run rehearsal.

/pj-rehearse ack

shellyyang1989:
@liangxia for approve

jianlinliu:
@shellyyang1989 shall we revert the change about `jsonpath` introduced in https://github.com/openshift/release/pull/43227/files#diff-a71e3444f4cc5e3d7ecb8ebd36fa8bd1fda181a0806d09523b318eadffaebc71L533-R593, especially in `check_clusteroperators`, `check_latest_machineconfig_applied`, `wait_machineconfig_applied`

shellyyang1989:
> @shellyyang1989 shall we revert the change about `jsonpath` introduced in https://github.com/openshift/release/pull/43227/files#diff-a71e3444f4cc5e3d7ecb8ebd36fa8bd1fda181a0806d09523b318eadffaebc71L533-R593, especially in `check_clusteroperators`, `check_latest_machineconfig_applied`, `wait_machineconfig_applied`

I've checked these command lines using 4.1 oc. Looks like they work

```
# ./oc get clusteroperator -o jsonpath='{.items[].status.conditions[?(@.type=="Degraded")].status}'
False

# ./oc get mcp/$role -o jsonpath='{.spec.configuration.name}'
rendered-worker-0f3f4aff1ba745c8330ccba32b799375

# ./oc get node $NODE -o jsonpath='{.metadata.annotations.machineconfiguration\.openshift\.io\/state}'
Done

# ./oc get -l node-role.kubernetes.io/$role -o jsonpath='{.items[*].meta.name}' nodes
ip-10-0-149-45.us-east-2.compute.internal ip-10-0-171-1.us-east-2.compute.internal ip-10-0-209-127.us-east-2.compute.internal

# ./oc get mcp/$role -o jsonpath='{.status.conditions[?(@.type=="Updated")].status}'
True

# ./oc -n openshift-config-managed get configmap admin-gates -o jsonpath='{.data}'
map[ack-4.11-kube-1.25-api-removals-in-4.12:Kubernetes 1.25 and therefore OpenShift 4.12 remove several APIs which require admin consideration. Please see the knowledge article https://access.redhat.com/articles/6955381 for details and instructions.]

# ./oc get clusterversion/version -o jsonpath='{.status.history[0].state}'
Completed
```

jianlinliu:
> I've checked these command lines using 4.1 oc. Looks like they work

Nice. 

/unhold

jiajliu:
> hecked these command lines using 4.1 oc. Looks like they work

Even 4.1 oc does not work with such commands, i think it will not break the following commands. We run such commands after `extract_oc`, which means the target version oc should be expected, so it should be high version oc, right？

jianlinliu:
> which means the target version oc should be expected

Good point, while in some edge test case, e.g: 4.1 -> 4.2 -> 4.3 -> 4.14, when running the 1st upgrade, the target version is 4.2, we maybe still hit `oc` command compatibility testing if oc 4.2 also does not support the options.

jiajliu:
> > which means the target version oc should be expected
> 
> Good point, while in some edge test case, e.g: 4.1 -> 4.2 -> 4.3 -> 4.14, when running the 1st upgrade, the target version is 4.2, we maybe still hit `oc` command compatibility testing if oc 4.2 also does not support the options.

for such path, 4.1-4.13 seems covered in `cucushift-chainupgrade-toimage`, which did not include the refactor. Only the last hop upgrade use `cucushift-upgrade-toimage`, which include the refactor. And only `on image info` is placed before `extract_oc`...

PR #44348, opened 2023-10-13
Update pelorus to trigger subset of current prow CI jobs.

mpryc:
This PR changes running jobs to limit AWS usage for the Pelorus project.

We do test now only for 4.13 for presubmit jobs and as a periodic only latest images are taken into consideration as this is main way of deploying project.


mpryc:
@mateusoliveira43 @weshayutin ^^

ci-operator/jobs/dora-metrics/pelorus/dora-metrics-pelorus-master-presubmits.yaml
```diff
@@ -1,251 +1,5 @@
...
-          secretName: manifest-tool-local-pusher
-      - name: pull-secret
-        secret:
-          secretName: registry-pull-credentials
-      - name: result-aggregator
-        secret:
-          secretName: result-aggregator
-    trigger: (?m)^/test( | .* )4.12-e2e-openshift,?($|\s.*)
   - agent: kubernetes
     always_run: true
```
mateusoliveira43, line 4: Lets address this issue here https://github.com/dora-metrics/pelorus/issues/1076

we need to change here to false https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift/oadp-operator/openshift-oadp-operator-master-presubmits.yaml#L979

and add this `run_if_changed` field https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift/oadp-operator/openshift-oadp-operator-master-presubmits.yaml#L996

I think this regex should be enough `"^exporters/|^charts/|^pelorus-operator/"`

mpryc: what about Makefile and files that are in fact changing e2e tests?

mpryc: ```^exporters/|^charts/|^pelorus-operator/|^scripts|Makefile```

?

mateusoliveira43: Good though, I agree

mateusoliveira43: wait, actually not

mateusoliveira43: If we add Makefile and script folder, we will still get some false positives

I suggest adding just `scripts/run-pelorus-e2e-tests.sh` file, which we use for running E2E tests

so, I would suggest something like `^exporters/|^charts/|^pelorus-operator/|^scripts/run-pelorus-e2e-tests.sh$`

ci-operator/config/dora-metrics/pelorus/dora-metrics-pelorus-master__4.13.yaml
```diff
@@ -30,7 +30,9 @@ resources:
       cpu: 500m
       memory: 1Gi
 tests:
-- as: e2e-openshift
+- always_run: false
+  as: e2e-openshift
+  run_if_changed: (^exporters/|^charts/|^pelorus-operator/|^scripts/run-pelorus-e2e-tests.sh$)
```
mateusoliveira43, line 35: To avoid duplication, can we remove this from `always_run` and `run_if_changed` here? checking here, does not seem necessary https://github.com/openshift/release/blob/master/ci-operator/config/openshift/oadp-operator/openshift-oadp-operator-master__4.13.yaml#L69

mpryc: But where would you create that?

The files under `ci-operator/jobs/` are generated so there should be no manual changes to those.

mateusoliveira43 (approved):

PR #44359, opened 2023-10-13
platform-operators: intermediate bump to go 1.19 to unwedge CI

joelanford:


ci-operator/config/openshift/platform-operators/openshift-platform-operators-main.yaml
```diff
@@ -11,7 +11,7 @@ build_root:
   image_stream_tag:
     name: release
     namespace: openshift
-    tag: golang-1.18
```
ncdc, line 14: Can you use this instead?
```
build_root:
  from_repository: true
```

tmshort: Presumably, this will keep up with the value in go.mod, and thus, never need to be modified again?

ncdc: Close; it uses https://github.com/openshift/platform-operators/blob/main/.ci-operator.yaml

ncdc: From whatever branch is appropriate

tmshort: Better than having to modify this repo...

joelanford: I can't. Go 1.20 from the repo fails CI. And 1.18 also fails CI.

I have another PR open to switch to from_repository.

ncdc: Gotcha.

ncdc:
What's going on with the rehearsal failures?

ci-operator/config/openshift/platform-operators/openshift-platform-operators-main.yaml
joelanford: Might need more surgery though. Maybe need to temporarily disable the lint test. Or force a go bump.

joelanford: I figured out at least one problem. platform-operators tries to use `setup-envtest@latest`, which apparently now requires a more recent version of Go to work.

I've updated @awgreene's PR (openshift/platform-operators#92) to pin setup-envtest. And I've also opened a separate PR to switch CI tooling over to bingo (openshift/platform-operators#95) to make sure build tools actually stay pinned.

We _may_ still need this PR because there is also an interplay between golangci-lint and the go compiler version that may require an intermediate bump to Go 1.19.

PR #44361, opened 2023-10-13
Hive: Skip periodic-images if unnecessary

2uasimojo:
Add our skip_if_only_changed to the periodic-images job so it doesn't run for e.g. docs-only changes.


2uasimojo:
/assign @suhanime 

It annoyed me that [this PR](https://github.com/openshift/hive/pull/2135) ran any CI at all.

PR #44366, opened 2023-10-14
wmco: mark deprecated bundle as optional to remove it from required pipeline contexts

jmguzik:
I implemented this possibility in conf generator.

### DIFF
--- a/ci-operator/config/openshift/windows-machine-config-operator/openshift-windows-machine-config-operator-master.yaml
+++ b/ci-operator/config/openshift/windows-machine-config-operator/openshift-windows-machine-config-operator-master.yaml
@@ -31,6 +31,7 @@ operator:
   - as: wmco-bundle
     context_dir: .
     dockerfile_path: bundle.Dockerfile
+    optional: true
   substitutions:
   - pullspec: REPLACE_IMAGE
     with: pipeline:windows-machine-config-operator-test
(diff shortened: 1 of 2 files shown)


ci-operator/config/openshift/windows-machine-config-operator/openshift-windows-machine-config-operator-master.yaml
```diff
@@ -31,6 +31,7 @@ operator:
   - as: wmco-bundle
     context_dir: .
     dockerfile_path: bundle.Dockerfile
+    optional: true
```
sebsoto, line 34: how is this deprecated?

sebsoto: This isn't building an index, but a bundle https://docs.ci.openshift.org/docs/how-tos/testing-operator-sdk-operators/#building-operator-bundles

aravindhp: @sebsoto If the bundle is not getting built now how are the tests that depend on it running and passing?

sebsoto: To be honest I'm a bit confused here. I'm not following where/what caused this to not be built.

PR #36771, opened 2023-02-27
ROX-15274: Add FIPS qa test to stackrox nightly

BradLugo:

PR #44427, opened 2023-10-16
stackrox: release-x.y: use automation-flavor for OCP 4

gavin-stackrox:
This  PR sets up the next release to use ACS automation for OCP clusters. And:
- corrects some `jobs/` always_run and `config/` and `jobs/` run_if_changed.

There will be followup PRs to handle master and nightlies and release*

## Testing

All rehearse jobs passed!

/hold
/uncc

### DIFF
--- a/ci-operator/config/stackrox/stackrox/stackrox-stackrox-release-x.y__ocp-4-10.yaml
+++ b/ci-operator/config/stackrox/stackrox/stackrox-stackrox-release-x.y__ocp-4-10.yaml
@@ -19,8 +19,7 @@ resources:
       memory: 4000Mi
 test_binary_build_commands: .openshift-ci/dispatch.sh test-binary-build-commands
 tests:
-- always_run: false
-  as: qa-e2e-tests
+- as: qa-e2e-tests
   optional: true
   run_if_changed: ^((generated|image|operator|pkg|make|deploy|scripts)/.*|Makefile|status\.sh|COLLECTOR_VERSION|EXPECTED_GO_VERSION)
   steps:
@@ -30,8 +29,7 @@ tests:
       TEST_SUITE: ocp-qa-e2e-tests
     workflow: stackrox-automation-flavors-ocp-4-e2e
   timeout: 5h0m0s
-- always_run: false
-  as: ebpf-qa-e2e-tests
+- as: ebpf-qa-e2e-tests
   optional: true
   steps:
     env:
@@ -40,8 +38,7 @@ tests:
       TEST_SUITE: ocp-qa-e2e-tests
     workflow: stackrox-automation-flavors-ocp-4-e2e
   timeout: 5h0m0s
-- always_run: false
-  as: operator-e2e-tests
+- as: operator-e2e-tests
   optional: true
   run_if_changed: ^((generated|image|operator|pkg|make|deploy|scripts)/.*|Makefile|status\.sh|COLLECTOR_VERSION|EXPECTED_GO_VERSION)
   steps:
@@ -50,8 +47,7 @@ tests:
       TEST_SUITE: ocp-operator-e2e-tests
     workflow: stackrox-automation-flavors-ocp-4-e2e
   timeout: 5h0m0s
-- always_run: false
-  as: sensor-integration-tests
+- as: sensor-integration-tests
   optional: true
   steps:
     env:
@@ -60,8 +56,7 @@ tests:
       TEST_SUITE: ocp-sensor-integration-tests
     workflow: stackrox-automation-flavors-ocp-4-e2e
   timeout: 5h0m0s
-- always_run: false
-  as: ui-e2e-tests
+- as: ui-e2e-tests
   optional: true
   steps:
     env:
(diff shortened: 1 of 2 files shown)


gavin-stackrox:
/pj-rehearse pull-ci-stackrox-stackrox-release-x.y-ocp-4-10-qa-e2e-tests

gavin-stackrox:
/pj-rehearse pull-ci-stackrox-stackrox-release-x.y-ocp-4-10-qa-e2e-tests

gavin-stackrox:
/pj-rehearse pull-ci-stackrox-stackrox-release-x.y-ocp-4-10-qa-e2e-tests

gavin-stackrox:
/pj-rehearse pull-ci-stackrox-stackrox-release-x.y-ocp-4-10-qa-e2e-tests

gavin-stackrox:
/pj-rehearse pull-ci-stackrox-stackrox-release-x.y-ocp-4-10-qa-e2e-tests

gavin-stackrox:
/pj-rehearse pull-ci-stackrox-stackrox-release-x.y-ocp-4-10-qa-e2e-tests

gavin-stackrox:
/pj-rehearse pull-ci-stackrox-stackrox-release-x.y-ocp-4-10-qa-e2e-tests

gavin-stackrox:
/pj-rehearse pull-ci-stackrox-stackrox-release-x.y-ocp-4-10-qa-e2e-tests

gavin-stackrox:
/pj-rehearse pull-ci-stackrox-stackrox-release-x.y-ocp-4-10-qa-e2e-tests

BradLugo (changes requested):
I don't think we typically define `always_run` in the `ci-operator/config`; it's one of the keys that are allowed to change between the config and the job (see https://docs.ci.openshift.org/docs/how-tos/contributing-openshift-release/#tolerated-changes-to-generated-jobs). I think you, Misha, and me had a convo in one of my first openshift PRs about this as well that I can dig up if we want more context and see where we collectively landed on it

This is also interesting:
>  **Note:** run_if_changed, skip_if_only_changed, and always_run: true are mutually exclusive.
\- https://docs.ci.openshift.org/docs/architecture/ci-operator/#pre-submit-tests

Which seems to make some of the `always_run` statements superfluous anyway

gavin-stackrox:
> I don't think we typically define `always_run` in the `ci-operator/config`; it's one of the keys that are allowed to change between the config and the job (see https://docs.ci.openshift.org/docs/how-tos/contributing-openshift-release/#tolerated-changes-to-generated-jobs). I think you, Misha, and me had a convo in one of my first openshift PRs about this as well that I can dig up if we want more context and see where we collectively landed on it
> 

I'd like to change that. There was a problem last week where jobs were failing on release PRs and confusing people resulting in https://github.com/openshift/release/pull/44360. Right now in the 4.2 presubmits always_run is true for most jobs which results in a lot of jobs for PRs that are pretty expensive e.g. ARO, ROSA, OSD*., EKS, etc. I think the intent was that those jobs should be by discretion of the author same as for master PRs. But somehow through a gap in automation or a missed step from instructions they were not set correctly. As a comparison these are always_run: false for the 4.0 and 4.1 presubmit jobs. Having `always_run:` in config/ is a new addition hence why we did not have it before.

I have another task to fixup the always_run for the main config once these OCP variants are done.

And an update to Mishas script to check that the always_run value in `config/` takes precedence over the value in `jobs/`.

> This is also interesting:
> 
> > **Note:** run_if_changed, skip_if_only_changed, and always_run: true are mutually exclusive.
> > - https://docs.ci.openshift.org/docs/architecture/ci-operator/#pre-submit-tests
> 
> Which seems to make some of the `always_run` statements superfluous anyway

Yeah. And they default to true. But I'd like to set them for presubmits to be explicit and avoid future problems if e.g. someone removes a `run_if_changed:`

BradLugo:
> Having always_run: in config/ is a new addition hence why we did not have it before.

I tried this a few months ago, and this was the outcome (copied from the convo mentioned earlier between you, Misha, and me: https://github.com/openshift/release/pull/36771#discussion_r1124444941):

> No it does not propagate. Changing a value of `always_run` in `config` does not change the value in `jobs` after `make update`.
> 
> It is coincidental that, if you set `always_run: false` in `config`, then delete current `always_run` in `jobs` and re-run `make update`, you will see `always_run: false`. However, try set `always_run: true` in `config` and do the same, you'll find `always_run: false` in `jobs`.

I think it would be more confusing (to both the author and reviewers) and more likely for a bug to slip through if someone changed `always_run: false` in `config` to `always_run: true` and that change was not actually picked up by the automation. If the behavior I quoted above ^ is fixed or works as you might expect (i.e., the `job` value _always_ matches the `config` value), then I would definitely much rather use the changes from this PR. But I don't think that's the case

gavin-stackrox:
True, it does not propagate. I think what I'll do is remove the `always_run` changes from this PR (except for the parts in `jobs/` where they are incorrect) and then propose the `always_run` change elsewhere and we can discuss.

PR #501, opened 2017-12-17
Track all dependencies centrally

smarterclayton:
It should be possible to track and manage a concrete list of "consistent commits" across all repos.  This is one particular example of how we might implement that.

All repos used by an origin release are git submodules under `dist/src` (which would be suitable for a GOPATH).  Each individual repo modeled here is expected to have a consistent tagging and release tracking pattern that follows from origin - `release-X.Y` and `vX.Y.Z[-prerelease]`.  The release repo itself would have a `release-X.Y` branch that would have different submodules and we would backport changes (it is assumed we will have to do this to keep our jobs from breaking).

A common `Makefile` has a set of actions to keep the repo in sync:

* `status` reports the commits and initializes the dir (clones the submodules)
* `sync` updates ALL submodules to the most recent commit on their release branch
* `tag` creates the same tag on the latest state of every module and then would push
* `branch` creates a release branch from the latest state and then would push 

The makefile could also track / link to the logic for building all images across the release in order, or executing specific actions across them.

PR #44431, opened 2023-10-17
fix wrong ack_data

jiajliu:
to fix https://issues.redhat.com/browse/OCPQE-17359 

```
$ oc -n openshift-config-managed get configmap admin-gates -o jsonpath='{.data}'|awk '{print $2}' | cut -f2 -d\"
1.26
$ oc -n openshift-config-managed get configmap admin-gates -o jsonpath='{.data}'|awk -F'"' '{print $2}'
ack-4.12-kube-1.26-api-removals-in-4.13
```

### DIFF
--- a/ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
+++ b/ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
@@ -490,16 +490,25 @@ function admin_ack() {
         echo "Admin ack is not required in either z-stream upgrade or 4.7 and earlier" && return
     fi
 
-    local out; out="$(oc -n openshift-config-managed get configmap admin-gates -o jsonpath='{.data}')"
+    local out; out="$(oc -n openshift-config-managed get configmap admin-gates -o json | jq -r ".data")"
     if [[ ${out} != *"ack-4.${SOURCE_MINOR_VERSION}"* ]]; then
-        echo "Admin ack not required" && return
+        echo "Admin ack not required: ${out}" && return
     fi
 
-    echo "Require admin ack"
+    echo "Require admin ack:\n ${out}"
     local wait_time_loop_var=0 ack_data
-    ack_data="$(echo ${out} | awk '{print $2}' | cut -f2 -d\")" && echo "Admin ack patch data is: ${ack_data}"
-    oc -n openshift-config patch configmap admin-acks --patch '{"data":{"'"${ack_data}"'": "true"}}' --type=merge
 
+    ack_data="$(echo "${out}" | jq -r "keys[]")"
+    for ack in ${ack_data};
+    do
+        # e.g.: ack-4.12-kube-1.26-api-removals-in-4.13
+        if [[ "${ack}" == *"ack-4.${SOURCE_MINOR_VERSION}"* ]]
+        then
+            echo "Admin ack patch data is: ${ack}"
+            oc -n openshift-config patch configmap admin-acks --patch '{"data":{"'"${ack}"'": "true"}}' --type=merge
+            break
+        fi
+    done
     echo "Admin-acks patch gets started"
 
     echo -e "sleep 5 min wait admin-acks patch to be valid...\n"
--- a/ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
+++ b/ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
@@ -416,11 +416,11 @@ function admin_ack() {
 
     echo "Require admin ack"
     local wait_time_loop_var=0 ack_data
-    ack_data="$(echo ${out} | jq -r "keys[]")"
+    ack_data="$(echo "${out}" | jq -r "keys[]")"
     for ack in ${ack_data};
     do
         # e.g.: ack-4.12-kube-1.26-api-removals-in-4.13
-        if [[ "${ack}" == *4\.$TARGET_MINOR_VERSION ]] 
+        if [[ "${ack}" == *"ack-4.${SOURCE_MINOR_VERSION}"* ]] 
         then
             echo "Admin ack patch data is: ${ack}"
             oc -n openshift-config patch configmap admin-acks --patch '{"data":{"'"${ack}"'": "true"}}' --type=merge


jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.13-amd64-stable-4.13-upgrade-from-stable-4.12-azure-ipi-fullyprivate-proxy-f28

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -497,7 +497,7 @@ function admin_ack() {
 
     echo "Require admin ack"
     local wait_time_loop_var=0 ack_data
-    ack_data="$(echo ${out} | awk '{print $2}' | cut -f2 -d\")" && echo "Admin ack patch data is: ${ack_data}"
+    ack_data="$(echo ${out} | awk -F'"' '{print $2}')" && echo "Admin ack patch data is: ${ack_data}"
```
jianlinliu, line 500: I'd like to use `jq` to handle the json output:
```
ack_data="$(echo ${out} | jq -r '. | to_entries[] | .key')"
```

jiajliu: Good suggestion. But I'm worried that we'll need to switch to an image without `jq` one day. So I try to keep jsonpath way as far as possible.

jiajliu:
```
INFO[2023-10-17T04:46:55Z] Running step azure-ipi-fullyprivate-proxy-f28-cucushift-upgrade-toimage. 
INFO[2023-10-17T06:31:46Z] Step azure-ipi-fullyprivate-proxy-f28-cucushift-upgrade-toimage succeeded after 1h44m51s. 
```
cucushift-upgrade-toimage PASS.

@jianlinliu @shellyyang1989 pls help review, thanks!

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -497,7 +497,7 @@ function admin_ack() {
...
     oc -n openshift-config patch configmap admin-acks --patch '{"data":{"'"${ack_data}"'": "true"}}' --type=merge
```
JianLi-RH, line 501: Can we output the command for debug?

jiajliu: You mean output for this command only or all the command in cucushift-upgrade-toimage? actually `echo "Admin ack patch data is: ${ack_data}"` is enough for debug if only for this command.

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
jianlinliu: Firstly, in this script - `cucushift-upgrade-toimage-commands.sh`, there are also other dependences, e.g: ansible, git, I do not think it is easy to be switched to a simple image.
Secondly, if that is really a concern, we can get it without `jq`.
```
oc -n openshift-config-managed get configmap admin-gates -o go-template='{{range $index,$value := .data}}{{printf "%s\n" $index}}{{end}}'
```

jiajliu: > Firstly, in this script - cucushift-upgrade-toimage-commands.sh, there are also other dependences, e.g: ansible, git, I do not think it is easy to be switched to a simple image.

ack, then `jq` should be better choice. i think we could get it in line 493 directly then.

```
$ oc -n openshift-config-managed get configmap admin-gates -o json | jq -r ".data|keys[]"
ack-4.12-kube-1.26-api-removals-in-4.13
```

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
JianLi-RH: look at https://gcsweb-qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/qe-private-deck/logs/periodic-ci-openshift-openshift-tests-private-release-4.14-arm64-stable-4.14-upgrade-from-stable-4.13-azure-ipi-disconnected-fullyprivate-f28/1713794938363187200/artifacts/azure-ipi-disconnected-fullyprivate-f28/cucushift-upgrade-toimage/build-log.txt

Seems `Admin ack patch data is: 1.27` is not enough.

JianLi-RH: I mean only this command, #501

jiajliu: Then you mean we should output ${out}, right?

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
shellyyang1989: I recall that we will get multiple keys in chain upgrade. But we only need to ack one of them in each round of upgrade. cc @JianLi-RH I recall you hit such issues before.

jianlinliu: > we only need to ack one of them in each round of upgrade

If that, sounds we need to filter out `ack-4.${SOURCE_MINOR_VERSION}` from `oc -n openshift-config-managed get configmap admin-gates -o json | jq -r ".data|keys[]"` output.

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
JianLi-RH: no, I want to know what we ran when debugging the failures.

jiajliu: > oc -n openshift-config patch configmap admin-acks --patch '{"data":{"'"${ack_data}"'": "true"}}' --type=merge

hmmm, only ${ack_data} is what we cared in above command. so i'm not quite getting u here, why ` echo "Admin ack patch data is: ${ack_data}"` is not enough for debug? could u explain more your requirement here:)

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
jianlinliu: Ignore my last codes, seem like the `cucushift-chainupgrade-toimage-commands` has the logic about "we only need to ack one of them in each round of upgrade", just need to sync the way of getting the keys there.

shellyyang1989: I meant even the data has been acked in previous upgrade, their keys are still present in the cm. So we need to pick the right one to ack.

jianlinliu: > even the data has been acked in previous upgrade,

But this case is only existing in `cucushift-chainupgrade-toimage`, right? Not related with this ref step, right?

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
JianLi-RH: All right, then that's enough for debugging.

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
shellyyang1989: The ref is the last step of chain upgrade. So we will hit the issue if the last upgrade image is stable where adminack is required. Now that seems like all the chain upgrade are on nightly. But I'm thinking to port some of them to stable builds to catch some bugs on upgrade w/o force option.

jiajliu: updated~

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -490,16 +490,25 @@ function admin_ack() {
...
+        echo "Admin ack not required: ${out}" && return
     fi
 
-    echo "Require admin ack"
+    echo "Require admin ack:\n ${out}"
     local wait_time_loop_var=0 ack_data
-    ack_data="$(echo ${out} | awk '{print $2}' | cut -f2 -d\")" && echo "Admin ack patch data is: ${ack_data}"
-    oc -n openshift-config patch configmap admin-acks --patch '{"data":{"'"${ack_data}"'": "true"}}' --type=merge
 
+    ack_data="$(echo ${out} | jq -r "keys[]")"
```
jianlinliu, line 501:
```suggestion
    ack_data="$(echo "${out}" | jq -r "keys[]")"
```

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -490,16 +490,25 @@ function admin_ack() {
...
+    for ack in ${ack_data};
+    do
+        # e.g.: ack-4.12-kube-1.26-api-removals-in-4.13
+        if [[ "${ack}" == *4\.$TARGET_MINOR_VERSION ]]
```
jianlinliu, line 505: Here should be `SOURCE_MINOR_VERSION` instead of `TARGET_MINOR_VERSION`, right?

jiajliu: target might be more correct? for example in a chain upgrade, 4.13 is the target version while 4.12 is the source version in both of keys.
```
ack-4.11-kube-1.25-api-removals-in-4.12
ack-4.12-kube-1.26-api-removals-in-4.13
```

jianlinliu: actually here I think it should be `[[ "${ack}" == *"ack-4.${TARGET_MINOR_VERSION}"* ]]` should be more accurate.

jiajliu: /s/ack-4.${TARGET_MINOR_VERSION}/ack-4.${SOURCE_MINOR_VERSION}? Agree to ack-***

jianlinliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-arm64-stable-4.14-upgrade-from-stable-4.13-azure-ipi-disconnected-fullyprivate-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-arm64-stable-4.14-upgrade-from-stable-4.13-azure-ipi-disconnected-fullyprivate-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.13-amd64-stable-4.13-upgrade-from-stable-4.12-azure-ipi-fullyprivate-proxy-f28

jiajliu:
Test pass. 
```
INFO[2023-10-18T04:54:53Z] Running step azure-ipi-fullyprivate-proxy-f28-cucushift-upgrade-toimage. 
INFO[2023-10-18T06:39:58Z] Step azure-ipi-fullyprivate-proxy-f28-cucushift-upgrade-toimage succeeded after 1h45m5s.
```

jiajliu:
@jianlinliu @JianLi-RH @shellyyang1989 PTAL again, thx!

PR #44433, opened 2023-10-17
Add multi-latest definition in hypershift-hosted job

dharaneeshvrd:


dharaneeshvrd:
Will remove dependancy on multi-latest release image on hypershift-hosted workflow, this PR is not required.

PR #43196, opened 2023-09-11
[azure] add post-check for private cluster

jinyunma:
Add post check for azure private/fullyprivate cluster to auto test cases:
1. OCP-25826 [ipi-on-azure] Create clusters with no public endpoints and access from internet
Note:  the checkpoint of bootstrap machine ( no public ip created and only access from machine in the vent) is not covered here, since export env `OPENSHIFT_INSTALL_PRESERVE_BOOTSTRAP` does not work well on Azure.

2. OCP-53233 [ipi-on-azure] Create fully private clusters using proxy as UserDefinedRouting outboundType

PR #44450, opened 2023-10-17
[azure] add post-check to all azure private/fullyprivate install chain

jinyunma:
post-check steps looks running well after https://github.com/openshift/release/pull/43196 is merged.

Extend the post-check step to other private/fullyprivate install chains on azure platform.


jinyunma:
cc @jianlinliu to review

PR #44452, opened 2023-10-17
Add INSTALLER_PROXY param to profile to make disconnected

jadhaj:
For the profile to be disconnected, INSTALLER_PROXY need to be set to true along with MIRROR_IMAGES


jadhaj:
cc @liangxia @YuLi517

PR #44453, opened 2023-10-17
prow: observability-operator wants to manage protected branches in gi…

jan--f:
…thub

### DIFF
--- a/core-services/prow/02_config/rhobs/observability-operator/_prowconfig.yaml
+++ b/core-services/prow/02_config/rhobs/observability-operator/_prowconfig.yaml
@@ -1,3 +1,9 @@
+branch-protection:
+  orgs:
+    rhobs:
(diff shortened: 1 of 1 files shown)


simonpasquier:
Reading https://docs.prow.k8s.io/docs/components/optional/branchprotector/ I don't see the `unmanaged` key but `protect`.
It might be good to add a comment explaining the reason why we want to keep control?

jan--f:
I went by https://docs.ci.openshift.org/docs/architecture/branch-protection/#is-it-possible-to-disable-the-branch-protection-for-my-repository-or-require-jobs-that-are-not-managed-by-prow

Added a comment as well.

simonpasquier:
> The comment was removed by the `make prow-config` target.

Nice :-/

/lgtm

PR #44462, opened 2023-10-17
grafana: disable auto branching

jan--f:
Grafana no longer ships in the payload, so we no longer need automatic release branches.


simonpasquier (commented):
/lgtm

jan--f:
/close
seems like this is not needed anymore

PR #44466, opened 2023-10-17
vsphere: boskos new devqe subnets

jcpowermac:


jcpowermac:
not sure where this error was coming from
> "Config validation failed" error=".83.max-count: must be >0"

PR #44474, opened 2023-10-17
Hive: Add snyk security scanning CI job

2uasimojo:
...per https://docs.ci.openshift.org/docs/how-tos/add-security-scanning/

TODO: Remove `optional: true` in a fup once we're happy with the job.

[HIVE-2336](https://issues.redhat.com//browse/HIVE-2336)


2uasimojo:
This is doing what it's supposed to. One set of vulns should be addressed by https://github.com/openshift/hive/pull/2139. Opened HIVE-2337 for the other.

/assign @suhanime 

Note: This will fail until vulns are addressed, but won't block merges since it's optional for now.

PR #44479, opened 2023-10-17
assisted-installer-agent: export GOCACHE and GOMODCACHE

rccrdpccl:
due to permissions error on cache files

/cc @danmanor @adriengentil


adriengentil:
don't we need to to the same things in the other repos (asssited-installer, assisted-service, ...)

adriengentil:
>Possibly: I'd suggest to see if this works properly or if we need other changes (by testing pending MR on the agent's repo), then I will propagate on other configs the full change - WDYT? 

Yes, I agree

adriengentil (approved):

PR #44490, opened 2023-10-17
extend the timeout for this ref from 2 to 3 hours.

ascerra:
# Problem
I've noticed that the ipi-install-install ref can sometimes fail due to the default timeout of 2hr for a ref in OCP CI.

Example is here: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-konveyor-tackle-ui-tests-mta_6.1.1-mta-ocp4.14-lp-interop-mta-interop-aws/1711260868357918720#1:build-log.txt%3A325

This failed due to the 2hr timeout while on its 2nd attempt to install a cluster.

# solution
Extend the timeout an additional hour to allow the ipi installs a chance to make it through all attempts
Example above failed during [Install attempt 2 of 3](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-konveyor-tackle-ui-tests-mta_6.1.1-mta-ocp4.14-lp-interop-mta-interop-aws/1711260868357918720#1:build-log.txt%3A197)

# Disclaimer
I'm aware that this change effects 12,000+ jobs so we need to consider if this extra hour will just waste time if the install is going to fail anyway or if we are already wasting time since maybe the extra hour could lead to more passing install steps.

[Looking at this line](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-konveyor-tackle-ui-tests-mta_6.1.1-mta-ocp4.14-lp-interop-mta-interop-aws/1711260868357918720#1:build-log.txt%3A362) makes me think that maybe the reason for the slowness, 2hrs to only be on install attempt 2, might be because of the OpenShift CI infra and not this install step itself.
- This is showing that the very next pod that tried to run in this job after the failure due to timeout failed to run due to a kubelet timeout issue, so maybe this problem is unavoidable since it might be an OCP CI intermittent infra issue.

### DIFF
--- a/ci-operator/step-registry/ipi/install/install/ipi-install-install-ref.yaml
+++ b/ci-operator/step-registry/ipi/install/install/ipi-install-install-ref.yaml
@@ -3,6 +3,7 @@ ref:
   from: installer
   grace_period: 10m
   commands: ipi-install-install-commands.sh
+  timeout: 3h
   cli: latest
   resources:
     requests:


yunjiang29:
Before increasing timeout, we need to find out the root cause of the failures. 
Generally, 2h is enough for the install process, this has been verified by tons of job builds. 
Holding this PR.
/hold

PR #44493, opened 2023-10-17
operator-marketplace: correct allowed teams

ncdc:
Replace the outdated openshift-patch-manager team with openshift-staff-engineers.

### DIFF
--- a/core-services/prow/02_config/operator-framework/operator-marketplace/_pluginconfig.yaml
+++ b/core-services/prow/02_config/operator-framework/operator-marketplace/_pluginconfig.yaml
@@ -17,7 +17,7 @@ label:
       - KeenonLee
       label: cherry-pick-approved
     - allowed_teams:
-      - openshift-patch-managers
+      - openshift-staff-engineers
       allowed_users:
       - kevinrizza
       - perdasilva


core-services/prow/02_config/operator-framework/operator-marketplace/_pluginconfig.yaml
```diff
@@ -17,7 +17,7 @@ label:
       - KeenonLee
       label: cherry-pick-approved
     - allowed_teams:
-      - openshift-patch-managers
+      - openshift-staff-engineers
```
bparees, line 20: before adding this, i'd like to know why this didn't get you the permission:
https://github.com/openshift/release/blob/bc2618b19cb79278ca12c02c6eb59d51cd2e804e/core-services/prow/02_config/_plugins.yaml#L642-L646

i don't think you should need this at a repo level.

bparees: (specifically because https://github.com/openshift/release/blob/bc2618b19cb79278ca12c02c6eb59d51cd2e804e/core-services/prow/02_config/_plugins.yaml#L642-L646 should already be getting you this permission)

i wonder if it's an org issue (openshift org vs operator-framework org)

bparees:
/hold cancel

discussed in slack

PR #44542, opened 2023-10-18
Pod isolation network host

pawelpaszki:
## WHAT
test: podisolation host-prefix check

## Local run
```
Getting list of management clusters in podisolation sector
Found 1 clusters
Getting network configuration for MC with cluster mgmt ID:<redacted>
Confirming that host_prefix of the MC is '24'

Passed tests:

OCPQE-17288

Failed tests:

```

### DIFF
--- a/ci-operator/step-registry/osd-ccs/fleet-manager/test/osd-ccs-fleet-manager-test-commands.sh
+++ b/ci-operator/step-registry/osd-ccs/fleet-manager/test/osd-ccs-fleet-manager-test-commands.sh
@@ -656,6 +656,40 @@ function test_machinesets_naming () {
 
 ###### end of machinesets naming test (OCP-68154) ######
 
+##################################################################
+
+###### host_prefix (podisolation) validation test (OCPQE-17288) ######
+
+function test_host_prefix_podisolation () {
+  TEST_PASSED=true
+  echo "Getting list of management clusters in podisolation sector"
+  CLUSTERS=$(ocm get /api/osd_fleet_mgmt/v1/management_clusters --parameter search="sector='podisolation'")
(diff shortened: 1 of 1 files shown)


pawelpaszki:
ping @jianping-shu

ci-operator/step-registry/osd-ccs/fleet-manager/test/osd-ccs-fleet-manager-test-commands.sh
```diff
@@ -656,6 +656,40 @@ function test_machinesets_naming () {
...
+        HOST_PREFIX=$(ocm get "$MGMT_CLUSTER_HREF" | jq -r .network.host_prefix)
+        echo "Confirming that host_prefix of the MC is '24'"
+        if [ "$HOST_PREFIX" -ne 24 ]; then
+          echo "Expected host_prefix of the MC to be '24'. Got '$HOST_PREFIX'"
+          TEST_PASSED=false
+        fi
+      fi
+    done
+  fi
+  update_results "OCPQE-17288" $TEST_PASSED
```
jianping-shu, line 688: I am thinking it might be a good idea to track such check with such OCPQE ticket instead of polarion test case as we talked before, since this doesn't worth a test case and maybe no appropriate test case to incorporating it.

pawelpaszki: ah yes. so are you suggesting to create a polarion test case and update the test case or keep the current code?

jianping-shu: I am ok with the current code.

PR #44547, opened 2023-10-18
sync okd builder with ocp to use go1.17

grokspawn:
As part of servicing [CVE-2023-44487](https://nvd.nist.gov/vuln/detail/CVE-2023-44487) we had to bump golang.org/x/net to 0.17.0, which requires go1.17.  OCP prow builder config is already at 1.17 but okd builder lags.

### DIFF
--- a/ci-operator/config/operator-framework/operator-marketplace/operator-framework-operator-marketplace-release-4.10__okd.yaml
+++ b/ci-operator/config/operator-framework/operator-marketplace/operator-framework-operator-marketplace-release-4.10__okd.yaml
@@ -3,19 +3,19 @@ base_images:
     name: "4.10"
     namespace: origin
     tag: base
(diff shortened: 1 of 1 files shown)


ncdc (approved):

ncdc:
e2e failed due to
```error: tag latest failed: Internal error occurred: registry.centos.org/dotnet/dotnet-31-centos7:latest: Get "https://registry.centos.org/v2/": dial tcp: lookup registry.centos.org on 172.30.0.10:53: no such host
```

That registry has been decommissioned: https://www.spinics.net/lists/centos-devel/msg22160.html

We won't be able to get the e2e to pass unless we make changes to the test code.

@joelanford given that the images rehearsal passed, I'm tempted to proceed to ack the rehearsals. WDYT?

PR #44554, opened 2023-10-18
cloud-credential-operator: Remove extra master and main branch config

sdodson:
There was duplicated master and main branch config that deviates from the norm. Remove it.


sdodson:
Observed in https://github.com/openshift/cloud-credential-operator/pull/607

PR #44557, opened 2023-10-18
Configuring SNYK for csi-driver-shared-resource-*

coreydaley:
It is expected that all of the `security` tests will fail.


coreydaley:
/assign @gabemontero 
it is ok that all of the `security` tests fail, we will fix them in the individual repositories at some point and then set the test to `required`

PR #44561, opened 2023-10-18
OTA-740: ci-operator/step-registry/ipi/conf/telemetry: Disable by default (again)

wking:
We'd tried this previously in 3c1da8eb20 (#32153), but had to roll it back with d61129c814 (#32249), to avoid failing:

    Prometheus when installed on the cluster should report telemetry if a cloud.openshift.com token is present

Subsequently, openshift/origin@76652fa4fa (openshift/origin#27422) taught that test-case about the config knob this step uses to disable Telemetry.  Those test-case changes are present in origin test suites starting in 4.12:

```console
$ for Y in $(seq 11 15); do git --no-pager grep 'should report telemetry' "origin/release-4.${Y}" -- test/extended/prometheus/prometheus.go; done
origin/release-4.11:test/extended/prometheus/prometheus.go:             g.It("should report telemetry if a cloud.openshift.com token is present [Late]", func() {
origin/release-4.12:test/extended/prometheus/prometheus.go:             g.It("should report telemetry [Late]", func() {
origin/release-4.13:test/extended/prometheus/prometheus.go:             g.It("should report telemetry [Late]", func() {
origin/release-4.14:test/extended/prometheus/prometheus.go:             g.It("should report telemetry [Serial] [Late]", func() {
origin/release-4.15:test/extended/prometheus/prometheus.go:             g.It("should report telemetry [Serial] [Late]", func() {
```

and [4.10 is end-of-life since 2023-09-10][1].  That leaves tests using 4.11 versions of the origin suite, and I'm addressing those via [the][2] [`JOB_NAME`][3] checks.  The checks are brittle, leaving out 4.9 and earlier, and possibly not matching some 4.11 jobs, but they will hopefully be sufficient to get us through until [4.11 goes end-of-life on 2024-02-10][3].  And when the defaulting logic breaks down, jobs that have an opinion can set `TELEMETRY_ENABLED` explicitly to match their needs.

[1]: https://access.redhat.com/support/policy/updates/openshift/#dates
[2]: https://docs.ci.openshift.org/docs/architecture/step-registry/#available-environment-variables
[3]: https://docs.prow.k8s.io/docs/jobs/#job-environment-variables

### DIFF
--- a/ci-operator/step-registry/ipi/conf/telemetry/ipi-conf-telemetry-commands.sh
+++ b/ci-operator/step-registry/ipi/conf/telemetry/ipi-conf-telemetry-commands.sh
@@ -4,6 +4,18 @@ set -o nounset
 set -o errexit
 set -o pipefail
 
+if test -z "${TELEMETRY_ENABLED}"
+then
+	if [[ "${JOB_NAME}" =~ 4[.]11 ]]
+	then
+		TELEMETRY_ENABLED=true
+		echo "TELEMETRY_ENABLED is empty, defaulting to '${TELEMETRY_ENABLED}' for the 4.11 ${JOB_NAME}"
+	else
+		TELEMETRY_ENABLED=false
+		echo "TELEMETRY_ENABLED is empty, defaulting to '${TELEMETRY_ENABLED}' for the not-4.11 ${JOB_NAME}"
+	fi
+fi
+
 if test true = "${TELEMETRY_ENABLED}"
 then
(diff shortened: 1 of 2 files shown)


wking:
Maybe the proxy is alive again:

/pj-rehearse

wking:
Not sure where Prow/Tide got off too.  Closing/reopening, to see if that unsticks them.

jupierce:
Trying to fix shellcheck anger here: https://github.com/openshift/release/pull/44591

wking:
Replaced by #44591 :bow:

PR #44588, opened 2023-10-19
Fix single arch to multi arch

shellyyang1989:


shellyyang1989:
cc @liangxia for approval. Thank you

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__amd64-stable.yaml
```diff
@@ -266,7 +266,7 @@ tests:
       UPGRADE_PRE_RUN_TAGS: '@upgrade-prepare and @4.14 and @azure-ipi and @heterogeneous
         and @network-ovnkubernetes and not @fips'
     test:
-    - chain: openshift-upgrade-qe-test-heterogeneous
+    - chain: openshift-upgrade-qe-test-arch-migration
```
liangxia, line 269: @shellyyang1989 Just to confirm, it means `xxx-to-multiarch` tests should use `arch-migration` test chain, or this is a special case here only ?

shellyyang1989: `xxx-to-multiarch` test uses `openshift-upgrade-qe-test-arch-migration` test chain on stable build, while it uses `openshift-upgrade-qe-test-heterogeneous` on nightly build.

liangxia: Good to know. Thanks for the info.

PR #44592, opened 2023-10-19
remove grafana branches and images for 4.12 and later

jan--f:


raptorsun:
It seems we should remove these lines, too:
https://github.com/openshift/release/blob/master/core-services/image-mirroring/openshift/mapping_origin_4_13#L143
https://github.com/openshift/release/blob/master/core-services/image-mirroring/openshift/mapping_origin_4_12#L141
https://github.com/openshift/release/blob/master/core-services/image-mirroring/openshift/mapping_origin_4_14#L151

PR #44595, opened 2023-10-19
aws-load-balancer-operator: use compacted credentials request for controller

alebedev87:
https://github.com/openshift/aws-load-balancer-operator/pull/114 changes the default controller's credentials request to unminified (equivalent to the upstream) version. For non-STS cluster it's still necessary to use the minified one as it results into an inline AWS role which is limited to 2048 chars.


alebedev87:
/hold

Holding this PR until github.com/openshift/aws-load-balancer-operator/pull/114 is ready to be merged.

alebedev87:
/pj-rehearse abort

The rehearsal won't pass until https://github.com/openshift/aws-load-balancer-operator/pull/114 is merged.

PR #44605, opened 2023-10-19
Snyk: un-ignore `snyk code` exit value

2uasimojo:
The `snyk_code` func as written would always succeed as the last command was an `echo`. Save the return code from the command and return it from the function.

[DPTP-2926](https://issues.redhat.com//browse/DPTP-2926)


2uasimojo:
This is working.

/pj-rehearse ack

The failure is expected, since `snyk code` failed. (Note that `snyk code` failed due to some problem with the service, not because it found vulns via the actual scan; but that's okay for purposes of this PR.)

PR #44623, opened 2023-10-19
stackrox: master: use automation-flavor for OCP 4

gavin-stackrox:
This PR sets up the main branch to use ACS automation for OCP clusters. Note: Sticking with cluster_profile for 4.14 until that reaches ocp/stable-4.14 (could use ocp/candidate-4.14 but nightlies from cluster_profile seems more useful).

Similar to https://github.com/openshift/release/pull/44427

## Testing

- [x] Rehearse jobs pass
  - sensor-profiling fails as that job is still a WIP (and always_run=false)
  - known flake in qa tests
- [x] Clusters created in the correct project and destroyed

/hold
/uncc

### DIFF
--- a/ci-operator/jobs/stackrox/stackrox/stackrox-stackrox-master-postsubmits.yaml
+++ b/ci-operator/jobs/stackrox/stackrox/stackrox-stackrox-master-postsubmits.yaml
@@ -471,7 +471,7 @@ postsubmits:
     always_run: true
     branches:
     - ^master$
-    cluster: build05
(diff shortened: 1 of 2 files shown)


gavin-stackrox:
/cc @davdhacs 

This is looking good in test ATM. If the remaining jobs pass (or fail with known flakes) *and* the clusters are cleaned up appropriately then I'd like to get this merged before another conflict comes along. Quotas have been updated to support a Monday morning PR stampede.

PR #44627, opened 2023-10-20
OVN IPI: enable IPsec after manifest is generated

rbbratta:
We need to enable IPsec after the manifests are generated in order to use the install-config CIDRs that were selected for each platform by previous steps.

This is to enable IPv6 IPsec testing


rbbratta:
First step of a refactor.  Next with change `ci-operator/step-registry/ovn/conf/ipsec-manifest/ovn-conf-ipsec-manifest-commands.sh`

rbbratta:
Looks like this is not needed, the installer re-generates the manifests.

PR #44667, opened 2023-10-20
fix: remove ubi8 based tf and pytorch images built from main branch

harshad16:
fix: remove ubi8 based tf and pytorch images built from main branch

As we only build the ubi8 tf and pytorch images on 2023a branch, they would be maintained only there.


atheo89 (approved):
/lgtm

atheo89:
Thanks Harshad, one small change request, we need to exclude them also from the e2e test job

/remove lgtm

harshad16:
Thanks for the review

PR #44670, opened 2023-10-20
core-services: Add openshift/kubernetes allowed_teams

aravindhp:
Add openshift-staff-engineers to allowed_teams for applying restricted labels.


soltysh (approved):
/lgtm
/approve

PR #44675, opened 2023-10-20
stackrox: fix always_run for powervs/ibmcloudz jobs

gavin-stackrox:
Missed on the introducing PR, jobs default to `always_run: true`.
/cc @davdhacs @BradLugo @mdafsanhossain


BradLugo (approved):
/lgtm

PR #44691, opened 2023-10-22
Disable all release jobs destined for `build05`

bradmwilliams:
This PR will disable all jobs destined for the `build05` cluster, as determined by this logic:
https://github.com/openshift/release-controller/blob/87e2390ec019de89f9201143543c8174184637b3/cmd/release-controller/sync_verify_prow.go#L88-L92

### DIFF
--- a/core-services/release-controller/_releases/release-ocp-4.10.json
+++ b/core-services/release-controller/_releases/release-ocp-4.10.json
@@ -113,18 +113,21 @@
       }
     },
     "metal-ipi": {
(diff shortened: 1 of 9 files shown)


bradmwilliams:
/hold
The impact of this change, on TRT's acceptance signal's, is pretty significant.  So, they need to be OK with this change before it gets merged.

bradmwilliams:
/close
They have re-enabled the `build05` cluster this morning and jobs appear to be running again.

PR #44705, opened 2023-10-23
[vsphere] optimize vsphere bastion host destroy script

jinyunma:


jinyunma:
cc @jianlinliu @sgaoshang @WenXinWei to reveiw

PR #44711, opened 2023-10-23
test: add checking obo machine pool in podisolation sector

pawelpaszki:
## WHAT
test: add checking obo machine pool in podisolation sector

## Local verification (against int:canary with podisolation enabled)
```
Getting list of management clusters in podisolation sector
Found 3 clusters
MC with ID: <redacted> is not ready
Confirming that 'obo' machine pool count is exactly 1 for cluster with ID: <redacted>
Confirming that the number of replicas and availability zones in the obo machine pool is 3
MC with ID: <redacted> is not ready

Passed tests:

OCPQE-17367

Failed tests:
```


pawelpaszki:
ping @jianping-shu

PR #44805, opened 2023-10-25
Request less cpu for cypress tests by default

liangxia:
Context in https://redhat-internal.slack.com/archives/CF8SMALS1/p1698116719896329

### DIFF
--- a/ci-operator/step-registry/openshift-extended/web-tests/openshift-extended-web-tests-ref.yaml
+++ b/ci-operator/step-registry/openshift-extended/web-tests/openshift-extended-web-tests-ref.yaml
@@ -6,9 +6,10 @@ ref:
   timeout: 8h0m0s
   resources:
     requests:
-      cpu: "2"
-      memory: 2Gi
+      cpu: "100m"
+      memory: 600Mi
     limits:
+      cpu: "2"
       memory: 4Gi
   env:
   - name: E2E_RUN_TAGS


ci-operator/step-registry/openshift-extended/web-tests/openshift-extended-web-tests-ref.yaml
```diff
@@ -6,7 +6,7 @@ ref:
   timeout: 8h0m0s
   resources:
     requests:
-      cpu: "2"
+      cpu: "500m"
```
jianlinliu, line 9: I saw most of vsphere step are limited to `100m`, can we also set it to `100m` here?

liangxia: I'm fine with it. Done

ci-operator/step-registry/openshift-extended/web-tests/openshift-extended-web-tests-ref.yaml
```diff
@@ -6,7 +6,7 @@ ref:
   timeout: 8h0m0s
   resources:
     requests:
-      cpu: "2"
+      cpu: "100m"
       memory: 2Gi
```
jianlinliu, line 10: Per the [thread](https://redhat-internal.slack.com/archives/C015H2WDJRY/p1698070785033039), seem like not only cpu hit bottleneck, also hit OOM, so shall we shrink the memory request to 600Mi?

To be safer, shall we set limits for `cpu`?

Another thing is there is `resources` setting in our ci job config level, e.g: openshift-openshift-tests-private-release-4.14__amd64-nightly.yaml, I am not sure the job config level setting will overwrite the step setting?

yanpzhan:
Web test consume much cpu/memory, I'm not sure if cypress auto could run normally after reduce cpu/memory. We can have this update for now. If it proves to affect cypress case running, we should fix then.

liangxia:
The actual resource usage is between requests and limits. E.g, request memory is 100Mi, and limits is 200Mi. The pod will allocate 100Mi at start phase, and can use more during test execution, but no more than 200Mi.

PR #44821, opened 2023-10-25
mco: don't always run jobs that are failing for a while in 4.14

sinnykumari:
This is to utilize resources better and skip noises as these jobs are mostly failing for a while in 4.14.
This includes e2e-gcp-rt-op, e2e-aws-ovn-fips-op and okd-scos-e2e-gcp-op.

If these jobs needs to run always, please make sure that they are passing majoirty of times.


yuqi-zhang (approved):
/lgtm

PR #44618, opened 2023-10-19
Cluster Logging - LOG-4449: Update to latest vector v0.33.0 for release-5.9

syedriko:
Add vector release-5.9 to Cluster Logging CI

JIRA: https://issues.redhat.com/browse/LOG-4449

PR #44838, opened 2023-10-25
LOG-4716: Add CI for logging vector 5.9

jcantrill:
https://issues.redhat.com/browse/LOG-4716

cc @syedriko @xperimental

### DIFF
--- a/ci-operator/config/ViaQ/vector/ViaQ-vector-release-5.9.yaml
+++ b/ci-operator/config/ViaQ/vector/ViaQ-vector-release-5.9.yaml
@@ -4,15 +4,15 @@ base_images:
     namespace: ocp
     tag: "9"
   cluster-logging-operator:
-    name: "5.8"
+    name: "5.9"
     namespace: logging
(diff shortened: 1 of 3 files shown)


ci-operator/config/ViaQ/vector/ViaQ-vector-release-5.8.yaml
```diff
@@ -28,7 +28,7 @@ base_images:
     namespace: logging
     tag: elasticsearch-proxy
   log-file-metric-exporter:
-    name: 5.y
+    name: "5.8"
```
xperimental, line 31: Does it make sense to keep this as `5.y` and build log-file-metric-exporter from master in our release?

syedriko:
@jcantrill we already have https://github.com/openshift/release/pull/44618

jcantrill:
> @jcantrill we already have https://github.com/openshift/release/pull/44618

We do but it is also missing the updates to the operator and other bits to properly test for 5.9

PR #44840, opened 2023-10-25
LOG-4716: Add CI for logging eventrouter 5.8

jcantrill:
https://issues.redhat.com/browse/LOG-4716

cc @xperimental

### DIFF
--- a/ci-operator/config/openshift/eventrouter/openshift-eventrouter-release-5.8.yaml
+++ b/ci-operator/config/openshift/eventrouter/openshift-eventrouter-release-5.8.yaml
@@ -1,0 +1,45 @@
+base_images:
+  base:
+    name: release
+    namespace: openshift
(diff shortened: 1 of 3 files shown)


xperimental:
It looks as if we have built eventrouter from master in the past releases. If there's no need to build different content for different versions, I think we can just keep using master instead of adding release branches.

jcantrill:
> It looks as if we have built eventrouter from master in the past releases. If there's no need to build different content for different versions, I think we can just keep using master instead of adding release branches.

This is one component that is not managed by the CLO and can really run on any release of OCP assuming the client is compatible.  I am fine doing that but it means we need to advise "older" releases to upgrade to "latest" if there are any bug fixes

xperimental:
Since the last release all releases (5.5+) use the `master` branch to build the product. Still need to check what's the case in the CI ...

PR #44845, opened 2023-10-25
OCPVE-722: fix: running snyk code and synk deps

jeff-roche:


jeff-roche:
/pj-rehearse ack

snyk-deps is an example of a failing test, snyk-code is an example of a passing one

PR #44850, opened 2023-10-25
[openshift/jenkins] Removing a couple label requirements

bradmwilliams:
Per request, I'm removing the requirements for:

1. jira/valid-bug
2. staff-eng-approved


bradmwilliams:
/hold
There are a couple on-going discussions about the nuances of making these kind of changes.  Just want to give folks a chance to fully understand and make an informed decision before proceeding.

PR #44851, opened 2023-10-25
stackrox: test less automatically on PRs

gavin-stackrox:
For OCP clusters:
- Only test EBPF on demand or if collector version changes.
- Only run sensor-integration on demand.

/cc @davdhacs @BradLugo


BradLugo (approved):
/lgtm

PR #44852, opened 2023-10-25
stackrox: test less automatically on merge

gavin-stackrox:
- When only `ui/` changes are made there is no need to run e2e tests
- For OCP 4 only run the ebpf variant of the qa tests when collector version changes
-  For OCP 4 only run operator e2e when operator files change
-  For OCP 4 only run compliance e2e when compliance files change


BradLugo (approved):
/lgtm

PR #44853, opened 2023-10-25
Update builder image to use golang 1.20

rhmdnd:
We need at least golang 1.20 to update controller-runtime to 0.16.3.


rhmdnd (commented):
/lgtm

PR #44863, opened 2023-10-26
[4.14/4.15] Opt out cypress tests for profile vsphere-ipi-disk-encryption-tang

liangxia:
Context in https://redhat-internal.slack.com/archives/C015H2WDJRY/p1698070785033039

/cc @yanpzhan @sgaoshang @jianlinliu @jhou1

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__amd64-nightly.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__amd64-nightly.yaml
@@ -3758,7 +3758,14 @@ tests:
       FIPS_ENABLED: "true"
       TAG_VERSION: '@4.14'
     test:
-    - chain: openshift-e2e-test-qe
+    - chain: cucushift-installer-check-cluster-health
+    - ref: idp-htpasswd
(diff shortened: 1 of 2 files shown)


sgaoshang:
@jianlinliu I think we should also remove it from vsphere-ipi-disk-encryption-tang-fips-f28-destructive job

liangxia:
> @jianlinliu I think we should also remove it from vsphere-ipi-disk-encryption-tang-fips-f28-destructive job

Nope. `openshift-e2e-test-qe-destructive` does not contain cypress tests.

jianlinliu:
@liangxia shall we also update the older versions on this job?

liangxia:
> @liangxia shall we also update the older versions on this job?

I think yes, but not sure.
As it exist in 4.11, and the profile in 4.11 was added on 2023-02-13. Does the issue occur since then or only happen recently ?

jianlinliu:
> > @liangxia shall we also update the older versions on this job?
> 
> I think yes, but not sure. As it exist in 4.11, and the profile in 4.11 was added on 2023-02-13. Does the issue occur since then or only happen recently ?

Recently a vsphere build farm migration happening, it affects all the existing ci jobs, so to be safer, I suggest to update the other versions too.

PR #44883, opened 2023-10-26
Trigger some tests when other *_VERSION files change too.

porridge:
In particular, operator e2e tests.

This should help catch earlier issues such as the one being reverted in https://github.com/stackrox/stackrox/pull/8371


porridge:
The rehearsal failures should now be fixed by the aforementioned revert. Let's see:

/pj-rehearse

gavin-stackrox (approved):

PR #44911, opened 2023-10-26
app.ci: Import SCOS as `machine-os-content` and `stream-coreos`

LorbusChris:
Also remove unused ISs.


LorbusChris:
ugh tide seems stuck
/lgtm

PR #44912, opened 2023-10-26
Snyk install jq if not installed

creydr:
Currently snyk tests fail, when jq is not installed
```
snyk installed to /tmp/snyk
1.1237.0
/bin/bash: line 18: jq: command not found
```
This PR addresses it and installs jq if not installed.

### DIFF
--- a/ci-operator/step-registry/openshift/ci-security/snyk/scan/openshift-ci-security-snyk-scan-commands.sh
+++ b/ci-operator/step-registry/openshift/ci-security/snyk/scan/openshift-ci-security-snyk-scan-commands.sh
@@ -15,7 +15,15 @@ ${SNYK_DIR}/snyk --version
 
 # install jq if not installed
 if ! [ -x "$(command -v jq)" ]; then
+    JQ_CHECKSUM="2f312b9587b1c1eddf3a53f9a0b7d276b9b7b94576c85bda22808ca950569716" # https://github.com/jqlang/jq/tree/master/sig
     curl -Lo ${SNYK_DIR}/jq "https://github.com/jqlang/jq/releases/download/jq-1.7/jq-linux-amd64"
+
+    actual_checksum=$(sha256sum ${SNYK_DIR}/jq | cut -d ' ' -f 1)
+    if [ "${actual_checksum}" != "${JQ_CHECKSUM}" ]; then
+        echo "Checksum of downloaded JQ didn't match expected checksum"
+        exit 1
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/openshift/ci-security/snyk/scan/openshift-ci-security-snyk-scan-commands.sh
```diff
@@ -13,6 +13,16 @@ chmod +x ${SNYK_DIR}/snyk
 echo snyk installed to ${SNYK_DIR}
 ${SNYK_DIR}/snyk --version
 
+# install jq if not installed
+if ! [ -x "$(command -v jq)" ]; then
+    JQ_DIR=/tmp/jq
```
bear-redhat, line 18: I think you can just put it to ${SNYK_DIR} for simplicity

creydr: updated

2uasimojo (commented):
:+1:

ci-operator/step-registry/openshift/ci-security/snyk/scan/openshift-ci-security-snyk-scan-commands.sh
```diff
@@ -13,6 +13,13 @@ chmod +x ${SNYK_DIR}/snyk
 echo snyk installed to ${SNYK_DIR}
 ${SNYK_DIR}/snyk --version
 
+# install jq if not installed
+if ! [ -x "$(command -v jq)" ]; then
+    curl -Lo ${SNYK_DIR}/jq "https://github.com/jqlang/jq/releases/download/jq-1.7/jq-linux-amd64"
```
cfergeau, line 18: For what it's worth, executing pre-built binaries downloaded from the net without knowing how/where they were built, if they were tampered with, ... could be problematic, especially when it's related to security.

creydr: @cfergeau good point. I added a check for the checksum. Or any better approaches?

cfergeau: Thanks! If this runs on fedora/RHEL, you can also `yum install jq`.

PR #44928, opened 2023-10-26
installer: bump golangci-lint to v1.53.1

vincepri:


patrickdillon:
Rehearsal shows that the 1.53 image is getting found. Although tests are failing due to (new?) lint errors, so I think we need to work out those failures before merging this change. 

/approve

vincepri:
I'll fix these as part of the other PR

PR #44930, opened 2023-10-26
fix 4.10 verify-deps jobs for azure disk and external provisioner

dobsonj:
- azure disk: ci must use golang 1.17, this is causing the `verify` and `verify-deps` jobs to fail. Fixes https://github.com/openshift/azure-disk-csi-driver/pull/58
- external provisioner: the unit test _also_ checks `go mod tidy` with no easy way to pass the compat flag. So this drops the compat flag from `verify-deps` to be consistent with the unit test. Fixes https://github.com/openshift/csi-external-provisioner/pull/76

cc @openshift/storage


dobsonj:
The verify-deps and unit jobs look okay now, but not certain why the others timed out.

dobsonj:
Wake up, Neo...
/retest

dobsonj:
all passed this time
/pj-rehearse ack

PR #43884, opened 2023-10-02
Bootstrap openshift-pipelines/console-plugin configuration.

vdemeester:
This repository is for the console dynamic plugin for OpenShift Pipelines.

/cc @jerolimov @concaf

PR #44935, opened 2023-10-26
Update openshift-pipelines/console-plugin base image

christoph-jerolimov:
This is a follow up on #43884

All CI jobs currently fail with this error:

> Some steps failed:                           
> could not resolve inputs: could not determine inputs for step [input:root]: could not resolve base image from openshift/release:console-pipelines-plugin-test-cypress: imagestreamtags.image.openshift.io "release:console-pipelines-plugin-test-cypress" not found 
> Reporting job state 'failed' with reason 'resolving_inputs' 

For example: https://prow.ci.openshift.org/pr-history/?org=openshift-pipelines&repo=console-plugin&pr=11

This changes the base image to `console-plugin-test-cypress` as used also in the console-plugin-template. We might change this in the future.


jerolimov:
Tests runs, but fails because we have linter issues in our code base.

/pj-rehearse ack

PR #44948, opened 2023-10-27
update aws 120 nodes cadence for 4.13 and 4.15 to avoid QE ES disk th…

qiliRedHat:
…roughput throttle
https://issues.redhat.com/browse/PERFSCALE-2592


qiliRedHat:
@svetsa-rh PTAL. As this is only cadence change, I did not run pj-rehearse.

PR #44955, opened 2023-10-27
METAL-752: Remove 4.10 ipv6 jobs

derekhiggins:
The 4.10 IPv6 jobs are broken (we can't mirror the images as the "oc adm release mirror" command doesn't support
"--keep-manifest-list=true"

4.10 is also EOL so delete these jobs.


iurygregory (approved):

JoelSpeed (commented):
/approve

dcbw:
@derekhiggins I was pretty sure this all worked on 4.10 at some point... do we know what broke?

derekhiggins:
closing in favour of fixing

PR #44959, opened 2023-10-27
Run openshift-pipelines/console-plugin main branch e2e tests against OCP 4.12 and 4.13

christoph-jerolimov:
This is a follow up on #43884 and #44935 for [openshift-pipelines/console-plugin](https://github.com/openshift-pipelines/console-plugin)

**Idea/changes:**

1. The plugin will be delivered as part of the Tekton CD / OpenShift Pipelines operator, not as part of OCP. So the version and branch names should be aligned with that releases, or have its own version schema. (As soon as we release a first version, we should have a release-1.0 branch, and not a release-4.15 branch.)
2. The main branch (and later these release branches) should run the e2e tests against multiple OCP versions because the operator will also support multiple versions (see files with `__4.12` suffix.).
3. The file without the OCP version suffix will run the build, unit tests (todo), and linter.

Changed are inspired by https://github.com/openshift/release/tree/master/ci-operator/config/openshift-knative/serverless-operator

This PR setups this test matrix for now:

| File | Branch | OCP version | Pipelines Operator version |
|---|---|---|---|
| `...main.yaml` | `main` | n/a | n/a - Not required. Just runs the build, unit tests and linter |
| `...main__4.12.yaml` | `main` | 4.12 | Latest/WIP, but not defined yet<sup>1</sup> |
| `...main__4.13.yaml` | `main` | 4.13 | Latest/WIP, but not defined yet<sup>1</sup>  |

<sup>1</sup> Currently, neither the job nor pipelines-plugin e2e tests install the Tekton CD / OpenShift Pipelines operator. We might need the WIP operator to get the latest Tekton Results API.


vikram-raj (commented):
/lgtm

bparees:
are you deliberately avoiding running it against 4.14 and 4.15?  i.e. non-GA versions?  Seems like you'd want some signal if the plugin is broken in an upcoming version of OCP

jerolimov:
> are you deliberately avoiding running it against 4.14 and 4.15? i.e. non-GA versions? Seems like you'd want some signal if the plugin is broken in an upcoming version of OCP

@bparees I would happily add it as soon as it's available. We're still setting up our project, that was the reason to focus on stable releases for the moment. But we will copy the 4.13 to have a 4.14 and 4.15 version as soon as we're running real tests on our side.

PR #44967, opened 2023-10-27
OCPVE-757: lvms: chore: add image replication

jakobmoellerdev:
Adds image replication to the repository that lvms uses in its public git repository for example deployments. Official Catalogs will still use internal builds. Secret was created in Vault to allow successful replication, but it would be great if someone can confirm the format.

Also excludes the `pre-commit` image from promotion as its only used during CI.

### DIFF
--- a/core-services/image-mirroring/lvms/OWNERS
+++ b/core-services/image-mirroring/lvms/OWNERS
@@ -1,0 +1,10 @@
+approvers:
+  - suleymanakbas91
+  - jeff-roche
+  - jakobmoellerdev
+reviewers:
+  - suleymanakbas91
+  - brandisher
+  - jeff-roche
+  - jerpeter1
+  - jakobmoellerdev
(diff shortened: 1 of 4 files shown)


jakobmoellerdev:
/pj-rehearse pull-ci-openshift-lvm-operator-main-lvm-operator-e2e-aws

jakobmoellerdev:
/pj-rehearse pull-ci-openshift-lvm-operator-main-lvm-operator-e2e-aws-sno

core-services/image-mirroring/lvms/mapping_lvms_latest
```diff
@@ -0,0 +1,2 @@
+registry.ci.openshift.org/ocp/lvm-operator:* quay.io/lvms_dev/lvms-operator
```
suleymanakbas91, line 1: Can we also mirror bundle and catalog images?

core-services/image-mirroring/lvms/OWNERS
```diff
@@ -0,0 +1,15 @@
+approvers:
```
suleymanakbas91, line 1: Did you create this manually? If so, could you make it the same as in https://github.com/openshift/lvm-operator/blob/main/OWNERS?

suleymanakbas91 (changes requested):

core-services/image-mirroring/lvms/mapping_lvms_latest
jakobmoellerdev: AFAIK we dont promote the bundle images because they contain the images from the bundle. We would need to introduce a new image for the bundle first

jmguzik (approved):

PR #44968, opened 2023-10-27
Add agent ha job for okd fcos variant

andfasano:
The following patch adds the latest missing agent job kind for the OKD FCOS builds (4.15  / 4.14 branches), so that it will be used the same base set of agent jobs (compact/sno/ha) already adopted to verify OCP releases and presubmit installer PRs.

The HA job has the following characteristics:
* Uses HA topology (3 masters, 2 workers)
* Uses ipv4/ipv6 dualstack networking
* Disconnected installation (using a local registry)
* Nodes use DHCP


LorbusChris:
Should the same tests be added to SCOS too?

PR #45010, opened 2023-10-30
Add post upgrade check on capability

shellyyang1989:
Refers [OCPQE-17567](https://issues.redhat.com/browse/OCPQE-17567)

### DIFF
--- a/ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
+++ b/ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
@@ -1,0 +1,451 @@
+#!/bin/bash
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+trap 'FRC=$?; create_postupgrade_junit' EXIT TERM
+
+# Generate the Junit for cap check
+function create_postupgrade_junit() {
+    echo "Generating the Junit for post upgrade check on capability"
+    filename="cluster upgrade"
+    testsuite="cluster upgrade"
+    subteam="OTA"
+    if (( FRC == 0 )); then
+        cat >"${ARTIFACT_DIR}/${filename}.xml" <<EOF
+<?xml version="1.0" encoding="UTF-8"?>
+<testsuite name="${testsuite}" failures="0" errors="0" skipped="0" tests="1" time="1">
+  <testcase name="${subteam}:Post upgrade check on capability should succeed or skip"/>
+</testsuite>
+EOF
+    else
+        cat >"${ARTIFACT_DIR}/${filename}.xml" <<EOF
+<?xml version="1.0" encoding="UTF-8"?>
+<testsuite name="${testsuite}" failures="1" errors="0" skipped="0" tests="1" time="1">
+  <testcase name="${subteam}:Post upgrade check on capability should succeed">
+    <failure message="">Post upgrade check on capability failed</failure>
+  </testcase>
+</testsuite>
+EOF
+    fi
+}
+
+baselinecaps_from_cluster=$(oc get clusterversion version -ojson | jq -r '.spec.capabilities.baselineCapabilitySet')
+if [[ "${baselinecaps_from_cluster}" == "" ]]; then
+    echo "spec.capabilities.baselineCapabilitySet is not set, skip the check!"
+    exit 0
+fi
+echo "baselinecaps_from_cluster: ${baselinecaps_from_cluster}"
+
+# Get capabilities string based on release version and capset
+# Need update when new cap set pops up
+function get_caps_for_version_capset() {
+    local version=$1
+    local cap_index=$2
+
+    if [ -z "${version}" ] || [ -z "${cap_index}" ]; then
+        echo "ERROR: Missing required arguments"
+        return 1
+    fi
+
+    major_version=$( echo "${version}" | cut -f1 -d. )
+    minor_version=$( echo "${version}" | cut -f2 -d. )
+    if [[ "${cap_index}" == "vCurrent" ]]; then
+        cap_index="v${major_version}.${minor_version}"
+    fi
+
+    local caps_string=""
+    case ${cap_index} in
+    "None")
+    ;;
+    "v4.11")
+    caps_string="baremetal marketplace openshift-samples"   
+    (( minor_version >=14 && major_version == 4 )) && caps_string="${caps_string} MachineAPI"
+    ;;
+    "v4.12")
+    caps_string="baremetal marketplace openshift-samples Console Insights Storage CSISnapshot"
+    (( minor_version >=14 && major_version == 4 )) && caps_string="${caps_string} MachineAPI"
+    ;;
+    "v4.13")
+    caps_string="baremetal marketplace openshift-samples Console Insights Storage CSISnapshot NodeTuning"
+    (( minor_version >=14 && major_version == 4 )) && caps_string="${caps_string} MachineAPI"
+    ;;
+    "v4.14")
+    caps_string="baremetal marketplace openshift-samples Console Insights Storage CSISnapshot NodeTuning MachineAPI Build DeploymentConfig ImageRegistry"
+    ;;
+    "v4.15")
+    caps_string="baremetal marketplace openshift-samples Console Insights Storage CSISnapshot NodeTuning MachineAPI Build DeploymentConfig ImageRegistry OperatorLifecycleManager"
+    ;;
+    *)
+    caps_string="baremetal marketplace openshift-samples Console Insights Storage CSISnapshot NodeTuning MachineAPI Build DeploymentConfig ImageRegistry OperatorLifecycleManager"
+    ;;
+    esac
+
+    echo $caps_string
+}
+
+# Extract manifests from release
+function extract_manifests() {
+    # echo "Extracing manifests..."
+    local work_dir="$1"
+    local release="$2"
+    #local tmp_known_caps
+
+    # Check if the directory exists
+    if [ ! -d "${work_dir}" ] || [ -z "${release}" ]; then
+        echo "ERROR: Missing required arguments"
+        return 1
+    fi
+
+    # Extract manifests
+    if ! oc adm release extract -a "${CLUSTER_PROFILE_DIR}/pull-secret" --to ${work_dir} --from ${release}; then
+        echo "ERROR: Extracting manifest failed"
+        return 1
+    fi
+}
+
+# Get implicitly enabled capabilities from cluster
+function get_actual_implicit_caps {
+    # echo "Getting actual implicitly enabled caps in cluster..."
+    local tmp_capabilities implicit_status implicit_message
+    # Obtain ImplicitlyEnabledCapabilities condition status from cluster
+    if ! implicit_status=$(oc get clusterversion version -o json | jq -r '.status.conditions[] | select(.type == "ImplicitlyEnabledCapabilities").status'); then
+        echo "ERROR: Get implicitly enabled caps failed"
+        return 1
+    fi
+    if [ "$implicit_status" = "False" ]; then
+        echo "INFO: No implicitly enabled capabilities in the cluster"
+        return
+    fi
+    # Obtain ImplicitlyEnabledCapabilities condition message from cluster
+    if ! implicit_message=$(oc get clusterversion version -o json | jq -r '.status.conditions[] | select(.type == "ImplicitlyEnabledCapabilities").message'); then
+        echo "ERROR: Get implicitly enabled caps failed"
+        return 1
+    fi
+
+    # Parse the message to extract implicit caps
+    # The condition message could be "The following capabilities could not be disabled: Console, Insights, Storage"
+    tmp_capabilities=$(echo "$implicit_message" | grep -oE 'could not be disabled: (.*)' | cut -d':' -f2)
+    actual_implicit_caps=$(echo "$tmp_capabilities" | awk -F', ' '{ for (i=1; i<=NF; i++) print $i }')
+    # echo "Actual implicitly enabled capabilities list is ${actual_implicit_caps}"
+}
+
+# Get new capabilities introduced in target release compared to source release
+# It doesn't take migrated caps into account
+function get_new_caps() {
+    # echo "Getting new caps introduced in target release as compared to source release..."
+    local source="$1"
+    local target="$2"
+    local array_source array_target extra_caps
+
+    # Check for missing environment variables
+    if [ -z "$source" ] || [ -z "$target" ]; then
+        echo "ERROR: Missing required arguments"
+        return 1
+    fi
+
+    # Split the elements into arrays
+    IFS=" " read -ra array_source <<< "${source}"
+    IFS=" " read -ra array_target <<< "${target}"
+
+    # Find the extra elements in $target compared to $source
+    extra_caps=()
+    for cap in "${array_target[@]}"; do
+        #shellcheck disable=SC2076
+        if [[ ! " ${array_source[*]} " =~ " ${cap} " ]]; then
+            extra_caps+=("$cap")
+        fi
+    done
+
+    # Combine the extra elements into a space-separated string
+    echo "${extra_caps[*]}"
+    #echo "Expected implicitly enabled capabilities list is ${expected_implicit_caps}"
+}
+
+# compares two payloads for capabilities difference at manifest level
+function extract_migrated_caps() {
+    echo "Extracting migrated caps between 2 releases..."
+    local source="$1"
+    local target="$2"
+    local source_dir target_dir source_cap_files
+
+    # Check for missing environment variables
+    if [ -z "$source" ] || [ -z "$target" ]; then
+        echo "ERROR: Missing required arguments"
+        return 1
+    fi
+
+    source_dir="/tmp/source"
+    target_dir="/tmp/target"
+    mkdir -p ${source_dir} ${target_dir}
+
+    if ! extract_manifests "${source_dir}" "${source}" || ! extract_manifests "${target_dir}" "${target}"; then
+        return 1
+    fi
+
+    source_cap_files=$(grep -rl "capability.openshift.io/name" ${source_dir})
+    for source_file in ${source_cap_files}; do
+        filename=$(basename "${source_file}")
+        target_file="$target_dir/$filename"
+
+        if [ -f "$target_file" ]; then
+            source_cap=$(grep 'capability.openshift.io/name:' "$source_file" | cut -d ':' -f2 | tr -d '"' | sort | uniq )
+            target_cap=$(grep 'capability.openshift.io/name:' "$target_file" | cut -d ':' -f2 | tr -d '"' | sort | uniq )
+
+            if [ "$source_cap" != "$target_cap" ]; then
+                echo "Migrated capability in $filename: $source_cap -> $target_cap"
+                migrated_caps["$source_cap"]=$target_cap
+            fi
+        fi
+    done
+}
+
+function check_cvo_cap() {
+
+    local result=0
+    local capability_set=$1
+    local expected_status=$2
+    local cvo_field=$3
+
+    cvo_caps=$(oc get clusterversion version -o json | jq -rc "${cvo_field}")
+    if [[ "${capability_set}" == "" ]] && [[ "${cvo_caps}" != "null" ]]; then
+        echo "ERROR: ${expected_status} capability set is empty, but find capabilities ${cvo_caps} in cvo ${cvo_field}"
+        result=1
+    fi
+    if [[ "${capability_set}" != "" ]]; then
+        if [[ "${cvo_caps}" == "null" ]]; then
+            echo "ERROR: ${expected_status} capability set are ${capability_set}, but it's empty in cvo ${cvo_field}"
+            result=1
+        else
+            cvo_caps_str=$(echo $cvo_caps | tr -d '["]' | tr "," " " | xargs -n1 | sort -u | xargs)
+            if [[ "${cvo_caps_str}" == "${capability_set}" ]]; then
+                echo "INFO: ${expected_status} capabilities matches with cvo ${cvo_field}!"
+                echo -e "cvo_caps: ${cvo_caps_str}\n${expected_status} capability set: ${capability_set}"
+            else
+                echo "ERROR: ${expected_status} capabilities does not match with cvo ${cvo_field}!"
+                echo -e "cvo_caps: ${cvo_caps_str}\n${expected_status} capability set: ${capability_set}"
+                echo "diff [cvo_caps] [${expected_status} capability set]"
+                diff <( echo $cvo_caps_str | tr " " "\n" | sort | uniq) <( echo $capability_set | tr " " "\n" | sort | uniq )
+                result=1
+            fi
+        fi
+    fi
+
+    return $result
+} >&2
+
+if [ -f "${SHARED_DIR}/kubeconfig" ] ; then
+    export KUBECONFIG=${SHARED_DIR}/kubeconfig
+else
+    echo "Unable to find kubeconfig under ${SHARED_DIR}!"
+    exit 1
+fi
+
+# Setup proxy if it's present in the shared dir
+if [[ -f "${SHARED_DIR}/proxy-conf.sh" ]]; then
+    # shellcheck disable=SC1091
+    source "${SHARED_DIR}/proxy-conf.sh"
+fi
+# shellcheck disable=SC2207
+version_set=($(oc get clusterversion version -ojson | jq -r .status.history[].version))
+
+# Mapping between optional capability and operators
+# Need update when new operator marks as optional
+declare -A caps_operator
+caps_operator[baremetal]="baremetal"
+caps_operator[marketplace]="marketplace"
+caps_operator[openshift-samples]="openshift-samples"
+caps_operator[CSISnapshot]="csi-snapshot-controller"
+caps_operator[Console]="console"
+caps_operator[Insights]="insights"
+caps_operator[Storage]="storage"
+caps_operator[NodeTuning]="node-tuning"
+caps_operator[MachineAPI]="machine-api control-plane-machine-set cluster-autoscaler"
+caps_operator[ImageRegistry]="image-registry"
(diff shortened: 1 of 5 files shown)


shellyyang1989:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-nightly-4.15-upgrade-from-stable-4.14-aws-ipi-baselinecaps-none-f28

shellyyang1989:
@jianlinliu @jinyunma @jiajliu @JianLi-RH Although it doesn't pass test yet, I need more eyes to make sure the logic makes sense. PTAL when you have time. Thank you!
cc @evakhoni

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
```diff
@@ -0,0 +1,439 @@
...
+# to make "oc registry login" interact with the build farm, set KUBECONFIG to empty,
+# so that the credentials of the build farm registry can be saved in docker client config file.
+# A direct connection is required while communicating with build-farm, instead of through proxy
+# KUBECONFIG="" oc --loglevel=8 registry login
+
+### why use RELEASE_IMAGE_LATEST_FROM_BUILD_FARM???
+#ocp_version=$(oc adm release info ${RELEASE_IMAGE_LATEST_FROM_BUILD_FARM} --output=json | jq -r '.metadata.version' | cut -d. -f 1,2)
+#echo "OCP Version: $ocp_version"
+
+source_major_version=$( echo "${RELEASE_IMAGE_LATEST}" | cut -f1 -d. )
```
jianlinliu, line 231: If the job is triggered from prow ci cron, the payload is using digest, maybe not tag name, so generally we are using `oc adm release info` to get version.

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
```diff
@@ -0,0 +1,439 @@
...
+# Append additional caps to enabled_capability_set
+additionalcaps_from_cluster=$(oc get clusterversion version -ojson | jq -r '.spec.capabilities.additionalEnabledCapabilities[*]')
+if [[ "${additionalcaps_from_cluster}" != "" ]]; then
+    enabled_capability_set="${enabled_capability_set} ${additionalcaps_from_cluster}"
+fi
+
+# Append new caps to enabled_capability_set
+expected_implicit_caps=""
+if ! get_new_caps "${source_cap_index}" "${target_cap_index}"; then
+    echo "ERROR: get_new_caps failed" && exit 1
```
jianlinliu, line 322: Shall we `exit` it as failure, if we are not very confident with that if the check is ready to move forward? Or should we always exit as passed, to check the log in early stage?

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
```diff
@@ -0,0 +1,439 @@
...
+    fi
+    done
+
+    # Combine the extra elements into a space-separated string
+    expected_implicit_caps="${extra_caps[*]}"
+    echo "Expected implicitly enabled capabilities list is ${expected_implicit_caps}"
+}
+
+# compares two payloads for capabilities difference at manifest level
+function extract_migrated_caps() {
```
jianlinliu, line 129: Shall we run `oc` commands to check the enabled resources instead of checking the extracted manifest?

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
shellyyang1989: Good catch.

shellyyang1989: Fixed by getting version from cv.status.history.

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
jianlinliu: Ignore the above comments, maybe I mis-understand the case.

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
```diff
@@ -0,0 +1,421 @@
...
+    echo "Unable to find kubeconfig under ${SHARED_DIR}!"
+    exit 1
+fi
+
+# Setup proxy if it's present in the shared dir
+if [[ -f "${SHARED_DIR}/proxy-conf.sh" ]]; then
+    # shellcheck disable=SC1091
+    source "${SHARED_DIR}/proxy-conf.sh"
+fi
+version_set=($(oc get clusterversion version -ojson | jq -r .status.history[].version))
```
jianlinliu, line 212: I wonder why we need one more `()` outside of `$()` here?

jianlinliu: ah, got it, you want it become an array.

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
```diff
@@ -0,0 +1,421 @@
...
+<testsuite name="${testsuite}" failures="1" errors="0" skipped="0" tests="1" time="1">
+  <testcase name="${subteam}:Post upgrade check on capability should succeed">
+    <failure message="">Post upgrade check on capability failed</failure>
+  </testcase>
+</testsuite>
+EOF
+    fi
+}
+
+baselinecaps_from_cluster=$(oc get clusterversion version -ojson | jq -r '.spec.capabilities.baselineCapabilitySet')
```
jianlinliu, line 34: Maybe we should get the baseline cap set from install.yaml, and ensure the baseline cap set equals to the value from cvo?

shellyyang1989: `cv.spec.capabilities` reflects the cap we defined in install-config.yaml. It looks like an install check point which I assume passed when we run the step.

jianlinliu: I checked `ci-operator/step-registry/cucushift/installer/check/capability/cucushift-installer-check-capability-commands.sh`, seem like there is no such check point, @jinyunma do you think we can add the check point in the  installer check script?

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
shellyyang1989: If we `exit 0`, we will not get alert about the error so we won't know it's failed until we check the log actively. Considering the test introduced critical bugs, I'm creating a Junit for it and thinking about to merge it to `upgrade` RP attribute. So we will be notified when the step fails

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
jianlinliu: @shellyyang1989 I am thinking another point, even we added installer check point, shall we also check the setting after the upgrade step completed.

jinyunma: Good point, we missed this checkpoint on installer side, it's okay to add it in installer check script.

shellyyang1989: Do you mean if we find the `cv.spec.capabilities` is different from the cap defined in install-config.yaml, we skip the check?

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
```diff
@@ -0,0 +1,422 @@
...
+fi
+
+# Setup proxy if it's present in the shared dir
+if [[ -f "${SHARED_DIR}/proxy-conf.sh" ]]; then
+    # shellcheck disable=SC1091
+    source "${SHARED_DIR}/proxy-conf.sh"
+fi
+# shellcheck disable=SC2207
+version_set=($(oc get clusterversion version -ojson | jq -r .status.history[].version))
+source_major_version=$( echo "${version_set[-1]}" | cut -f1 -d. )
```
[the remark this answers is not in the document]
shellyyang1989: This is a separate step which is out of chain upgrade. So it's expected to be run once in chain upgrade. It validates the caps on the upgraded cluster compared to the initial cluster.

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
```diff
@@ -0,0 +1,422 @@
...
+    extra_caps=()
+    for cap in "${array_target[@]}"; do
+    #shellcheck disable=SC2076
+    if [[ ! " ${array_source[*]} " =~ " ${cap} " ]]; then
+        extra_caps+=("$cap")
+    fi
+    done
+
+    # Combine the extra elements into a space-separated string
+    expected_implicit_caps="${extra_caps[*]}"
```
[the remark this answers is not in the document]
shellyyang1989: Good catch! Thank you.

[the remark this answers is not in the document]
shellyyang1989: Just realized that the extra_cap is introduced in the target version and it's unknown to source cluster. So it's impossible to be enabled in additional_enabled_cap

shellyyang1989:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-nightly-4.15-upgrade-from-stable-4.14-aws-ipi-baselinecaps-none-f28

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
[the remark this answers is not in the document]
jiajliu: do u mean, if it's 4.11-4.14 upgrade, it only do post check on cap for the last hop 4.14, while to compare with a 4.11 cluster? hmm, i wonder if the post check fail, is it easy for us to identify which hop got a wrong cap.

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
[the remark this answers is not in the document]
jiajliu: you're right, this scenario does not make sense. i'm thinking about if new_caps always implicit caps. do we distinguish which profile for the check? for example the initial cluster with vcurrent setting in baselineCapabilitySet, the new caps should be enabled, but not implicitly enabled, right?

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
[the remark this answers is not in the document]
shellyyang1989: Yeah, it's the drawback. But I don't prefer to move it to upgrade step as the step is complicated enough and it will be more complex with more caps introduced. It's not cvo only feature and all cap related components are supposed to ensure the functionality. I'm hoping other qe teams to treat it carefully in their e2e automation.

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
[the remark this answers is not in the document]
shellyyang1989: ah, got you. Good example. Thank you.

shellyyang1989:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-baselinecaps-none-f28

shellyyang1989:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-baselinecaps-none-f28

shellyyang1989:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-nightly-4.15-upgrade-from-stable-4.14-aws-ipi-baselinecaps-none-f28

shellyyang1989:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-nightly-4.15-upgrade-from-stable-4.14-aws-ipi-baselinecaps-none-f28

shellyyang1989:
@jiajliu @jianlinliu Thanks for your comments a lot! They are helpful. Please help re-review when you have time.

shellyyang1989:
The capability check step passed.

```
INFO[2023-11-02T10:09:36Z] Running step aws-ipi-baselinecaps-none-f28-cucushift-installer-check-capability. 
INFO[2023-11-02T10:09:46Z] Step aws-ipi-baselinecaps-none-f28-cucushift-installer-check-capability succeeded after 9s. 
```

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
jianlinliu: I meant is that possible, the installer checkpoint get passed, while after upgrade, the `cv.spec.capabilities` get different with the user setting from the fresh install.

shellyyang1989: IMO, it's a rule in CI to not change the cluster behavior through the test. Even if someone changes some resources but they must ensure those resources are reverted back after their test. It's a dangerous operation to day-2 update cap setting in CI because enabled resources can't be disabled any more. So I'm assuming the cap set is not changed.

jianlinliu: I did not mean some e2e testing changed the enabled caps, but maybe some CVO bugs cause the unexpected change.

shellyyang1989: ah, got you. `cv.spec.capabilities` reflects the cap set what user specifies, so CVO will set the field based on install-config.yaml in the beginning of installation. Later on, CVO will never update it. CVO only reflects cluster changes to `cv.status`

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
```diff
@@ -0,0 +1,451 @@
...
+    fi
+    # Obtain ImplicitlyEnabledCapabilities condition message from cluster
+    if ! implicit_message=$(oc get clusterversion version -o json | jq -r '.status.conditions[] | select(.type == "ImplicitlyEnabledCapabilities").message'); then
+        echo "ERROR: Get implicitly enabled caps failed"
+        return 1
+    fi
+
+    # Parse the message to extract implicit caps
+    # The condition message could be "The following capabilities could not be disabled: Console, Insights, Storage"
+    tmp_capabilities=$(echo "$implicit_message" | grep -oE 'could not be disabled: (.*)' | cut -d':' -f2)
```
jianlinliu, line 129:
```suggestion
    tmp_capabilities=$(echo "$implicit_message" | grep -oE 'could not be disabled: (.*)' | cut -d': ' -f2)
```

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
```diff
@@ -0,0 +1,451 @@
...
+
+    # Split the elements into arrays
+    IFS=" " read -ra array_source <<< "${source}"
+    IFS=" " read -ra array_target <<< "${target}"
+
+    # Find the extra elements in $target compared to $source
+    extra_caps=()
+    for cap in "${array_target[@]}"; do
+    #shellcheck disable=SC2076
+    if [[ ! " ${array_source[*]} " =~ " ${cap} " ]]; then
```
jianlinliu, line 156: Missing some leading whitespaces?

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
shellyyang1989: Updated.

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
shellyyang1989: Updated.

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
```diff
@@ -0,0 +1,451 @@
...
+    local expected_status=$2
+    local cvo_field=$3
+
+    cvo_caps=$(oc get clusterversion version -o json | jq -rc "${cvo_field}")
+    if [[ "${capability_set}" == "" ]] && [[ "${cvo_caps}" != "null" ]]; then
+        echo "ERROR: ${expected_status} capability set is empty, but find capabilities ${cvo_caps} in cvo ${cvo_field}"
+        result=1
+    fi
+    if [[ "${capability_set}" != "" ]]; then
+        if [[ "${cvo_caps}" == "null" ]]; then
```
jianlinliu, line 217: Can we extract the if condition to a separate block just like what we did on line 212 for easy reading?

jianlinliu: Okay, ignore it. I saw `cucushift-installer-check-capability` is also using the same syntax, if needed, we can update them together later.

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
```diff
@@ -0,0 +1,451 @@
...
+    ;;
+    "v4.15")
+    caps_string="baremetal marketplace openshift-samples Console Insights Storage CSISnapshot NodeTuning MachineAPI Build DeploymentConfig ImageRegistry OperatorLifecycleManager"
+    ;;
+    *)
+    caps_string="baremetal marketplace openshift-samples Console Insights Storage CSISnapshot NodeTuning MachineAPI Build DeploymentConfig ImageRegistry OperatorLifecycleManager"
+    ;;
+    esac
+
+    if (( minor_version >=14 && major_version == 4 )) && [[ $caps_string != *"MachineAPI"* ]]; then
```
jianlinliu, line 82: We already maintain a full list for each version, I think we do not need this if block.

jianlinliu, line 82: I think we need to exclude `cap_index == None` case here?

And for easy reading, I suggest to move the append operation to `v4.11` `v4.12` and `v4.13` branch case.

shellyyang1989: Good catch! Updated.

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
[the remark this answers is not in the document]
jiajliu: do we have a proper job for a test against chain upgrade on such profile?

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
```diff
@@ -0,0 +1,451 @@
...
+
+# v4.x baseline set should grow new caps in v4.x of target version explicitly after upgrade
+# v4.x baseline set should grow new caps in target version implicitly
+# v4.x baseline set should grow new caps implicitly after upgrade if cap migrates to a cap out of v4.x & additional caps of target version
+if [[ "${baselinecaps_from_cluster}" == "v4."* ]]; then
+    source_enabled_caps="${source_baseline_caps} ${additionalcaps_from_cluster}"
+    target_enabled_caps="${target_baseline_caps} ${additionalcaps_from_cluster}"
+    expected_enabled_caps="${target_enabled_caps}"
+    
+    expected_enabled_caps="${expected_enabled_caps} ${new_caps}"
```
jianlinliu, line 313: From my understanding, the expected enabled caps is the source enabled caps in  + new_caps, sounds the current logic would make several duplicated caps show up in the the expected enabled caps, right?

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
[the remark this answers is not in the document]
shellyyang1989: We don't have chain upgrade job on caps profile currently. If someday we decide to add such job, we can update the step as needed, but I'm afraid that would be complex

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
jianlinliu: Here `d` is the key point, generally I think v4.x would have no any change on both source and target version, of course, machine-api is an exception change. 

So let us take 4.13 + v4.11as an example source cluster, now it is going to be upgraded to 4.14.

In source cluster, baremetal, marketplace, openshift-sample is the enabled optional caps, the new cap is MachineAPI, Build, DeploymentConfig, ImageRegistry. 

So the target enabled should be baremetal, marketplace, openshift-sample + MachineAPI, Build, DeploymentConfig, ImageRegistry.

In your example, the key point is the same v4.x has different sets in source and target, and the different cap - `d` is also never show up in the new cap, but till now, I do not figure out such example yet.

shellyyang1989: Yeah, my example doesn't happen yet. I'm just in case it occurs some day and having `d` there twice won't hurt because I remove duplicates later

jianlinliu:
Looks good to me now.

Only one question, can we run some test job to check if we really can catch OCPBUGS-22718 by the PR? I think we can run pj-rehearse job on `gcp-ipi-baselinecaps-none-additionalcaps-f7` to catch it, but seem like the latest nightly build already include the fix, maybe we can specify a lower target nightly build in this PR temporarily just for reproduce it? or we can add a stable -> stable upgrade job for `gcp-ipi-baselinecaps-none-additionalcaps` install workflow in this PR to run the pj-rehearse testing (I saw quay build does not include the fix yet)?

shellyyang1989:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-stable-4.14-upgrade-from-stable-4.13-aws-ipi-baselinecaps-none-f28

shellyyang1989:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-stable-4.14-upgrade-from-stable-4.13-aws-ipi-baselinecaps-none-f28

shellyyang1989:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-stable-4.14-upgrade-from-stable-4.13-aws-ipi-baselinecaps-none-f28

shellyyang1989:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-stable-4.14-upgrade-from-stable-4.13-aws-ipi-baselinecaps-none-f28

shellyyang1989:
Rehearse [job](https://qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/qe-private-deck/pr-logs/pull/openshift_release/45010/rehearse-45010-periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-stable-4.14-upgrade-from-stable-4.13-aws-ipi-baselinecaps-none-f28/1720757956829515776) failed as expected.

```
actual_implicit_caps: Build Console DeploymentConfig ImageRegistry MachineAPI
 expected_implicit_caps: Build DeploymentConfig ImageRegistry MachineAPI
ERROR: Actual implicitly enabled capabilities don't match expected implicitly enabled capabilities
------check cvo status capabilities check-----
===check .status.capabilities.enabledCapabilities
ERROR: enabled capabilities does not match with cvo .status.capabilities.enabledCapabilities!
ERROR: capability Console: operator console -- found
```

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
shellyyang1989: Reverted back as cut only accepts one character of delimiter.

jianlinliu:
cc @liangxia to review.

jianlinliu:
@shellyyang1989 shall we also apply the checking to `openshift-upgrade-qe-test-disconnected` chain?

shellyyang1989:
> @shellyyang1989 shall we also apply the checking to `openshift-upgrade-qe-test-disconnected` chain?

Sure. Didn't realize we have cap test on disconnected env

PR #45031, opened 2023-10-30
openshift-logging: Remove not existing images from mirroring

xperimental:
During the configuration of CI for 5.8/5.9 we decided to keep these images at `5.y` and build them from `master` and not create a separate `5.8` branch. So they do not exist as an image.

PTAL @jcantrill @periklis


jcantrill (commented):
/lgtm

PR #45036, opened 2023-10-30
Provision/Deprovision the network firewall

yasun1:
[OCPQE-15265](https://issues.redhat.com//browse/OCPQE-15265)


ci-operator/step-registry/aws/provision/network-firewall/rule-group/aws-provision-network-firewall-rule-group-commands.sh
```diff
@@ -0,0 +1,76 @@
...
+{
+  "RulesSource": {
+    "RulesSourceList": {
+      "Targets": [
+        ".quay.io",
+        ".redhat.io",        
+        "api.openshift.com",
+        "access.redhat.com",
+        "sso.redhat.com",        
+        ".amazonaws.com"
```
yasun1, line 27: Just set some domains and need to check with the feature owners about what should be here.

yasun1:
@yunjiang29 Please help review. Thanks.

PR #45048, opened 2023-10-30
fix console logging image

jcantrill:
This PR:
* sets the logging console image using 2 vars to support the operator and the test pods


jcantrill:
cc @xperimental @cahartma

PR #45101, opened 2023-10-31
openshift/operator-framework-tooling: init

stevekuznetsov:


ci-operator/config/openshift/operator-framework-tooling/openshift-operator-framework-tooling-main.yaml
```diff
@@ -0,0 +1,63 @@
+base_images:
+  commitchecker:
+    name: commitchecker
+    namespace: ci
+    tag: "4.14"
```
ncdc, line 5: This can be 4.15 now I believe

ci-operator/config/openshift/operator-framework-tooling/openshift-operator-framework-tooling-main.yaml
```diff
@@ -0,0 +1,63 @@
...
+  ocp_builder_rhel-8-golang-1.20-openshift-4.15:
+    name: builder
+    namespace: ocp
+    tag: rhel-8-golang-1.20-openshift-4.15
+  os:
+    name: centos
+    namespace: origin
+    tag: stream8
+binary_build_commands: GO111MODULE=on GOPROXY=https://proxy.golang.org make build
```
ncdc, line 14: Do we need the 2 env vars - aren't they defaults now?

ci-operator/config/openshift/operator-framework-tooling/openshift-operator-framework-tooling-main.yaml
```diff
@@ -0,0 +1,63 @@
...
+build_root:
+  from_repository: true
+images:
+- dockerfile_path: v0.Dockerfile
+  from: os
+  inputs:
+    bin:
+      paths:
+      - destination_dir: .
+        source_path: /go/src/github.com/openshift/operator-framework-tooling/v0
```
ncdc, line 24: Is this right or should it be cmd/v0?

stevekuznetsov: The built binary lands at the repo root from `go build` so this is right

ci-operator/config/openshift/operator-framework-tooling/openshift-operator-framework-tooling-main.yaml
stevekuznetsov: The CI somewhere sets the default to vendor so I did this to set it back

ci-operator/config/openshift/operator-framework-tooling/openshift-operator-framework-tooling-main.yaml
```diff
@@ -0,0 +1,63 @@
...
+    ocp_builder_rhel-8-golang-1.20-openshift-4.15:
+      as:
+      - registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.20-openshift-4.15
+  to: olm-v0-bumper
```
tmshort, line 28: Noting that https://github.com/openshift/release/blob/3bf0b3ae011debaefefb564ad6f233c380d033f7/ci-operator/jobs/infra-periodics.yaml#L965 references `olm-bumper`. Does this matter?

stevekuznetsov: We're going to have both v0 and v1 bumpers. I'll update the job later.

PR #45150, opened 2023-11-01
OCPQE-17639 Using qe-private-deck for openshift/verification-tests

JianLi-RH:
/cc @nmoraitis @sgoeddel @wking @jianlinliu @yunjiang29


core-services/prow/02_config/_config.yaml
```diff
@@ -286,7 +287,12 @@ plank:
       timeout: 23h0m0s
     repo: openshift/distributed-tracing-qe
   - config:
-      timeout: 12h0m0s
+      gcs_configuration:
+        bucket: qe-private-deck
+        mediaTypes:
+          log: text/plain
+      gcs_credentials_secret: gce-sa-credentials-gcs-qe-private-deck
+      timeout: 23h0m0s
```
smg247, line 295: this part doesn't need to change

JianLi-RH: do you mean only **timeout** or the whole section? @smg247

smg247: just the timeout

smg247 (approved):

PR #45162, opened 2023-11-01
shiftstack/periodic: remove old & unsupported OCP versions

EmilienM:
From https://access.redhat.com/support/policy/updates/openshift/
Now we only support OCP 4.11 and onward.

Let's remove periodic jobs older than that, we don't take care of them
anyway.


stephenfin:
Holding while we wait for @mandre to take a look, but this looks reasonable from my perspective

pierreprinetti:
@EmilienM have you considered using our gitops scripts for bumping the release versions? https://github.com/shiftstack/shiftstack-ci/blob/main/bump-periodics.sh

Applied to 4.15, it should set a very long interval for releases that fell out of support scope.

EmilienM:
> https://github.com/shiftstack/shiftstack-ci/blob/main/bump-periodics.sh

why are we keeping the files if we don't support these releases?

mandre:
/hold
Pending resolution on whether or not it's fine to remove old periodic jobs.

pierreprinetti:
> > https://github.com/shiftstack/shiftstack-ci/blob/main/bump-periodics.sh
> 
> why are we keeping the files if we don't support these releases?

Because we’re asked to keep the CI ready to ramp up to validate old unsupported releases in case there is a support-exception case needing an exceptional change.

EmilienM:
Pierre, we were told on Slack that this is fine. If you're ok with that, let's clean this up.

ci-operator/config/shiftstack/shiftstack-ci/shiftstack-shiftstack-ci-main__periodic-4.11-upgrade-from-stable-4.10.yaml
mandre, line 1: Please keep this one. 4.11 is supported and we need to ensure we can upgrade from 4.10.

EmilienM: it's a waste of resources, the job has never passed and nobody is looking at it.

https://prow.ci.openshift.org/job-history/gs/origin-ci-test/logs/periodic-ci-shiftstack-shiftstack-ci-main-periodic-4.11-upgrade-from-stable-4.10-e2e-openstack-kuryr-upgrade

mandre: This is not exactly true, while the job isn't passing it doesn't mean there's no value in having it around.
Looking at [the last 4.10->4.11 upgrade job](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-shiftstack-shiftstack-ci-main-periodic-4.11-upgrade-from-stable-4.10-e2e-openstack-upgrade/1719990732082647040), we can see the upgrade was successful, but some tests failed because the cluster was not completely available during the upgrade. That is still good signal.

pierreprinetti:
> Pierre, we were told on Slack that this is fine. If you're ok with that, let's clean this up.

Ok. Please consider amending the job upgrade script.

ci-operator/config/shiftstack/shiftstack-ci/shiftstack-shiftstack-ci-main__periodic-4.11-upgrade-from-stable-4.10.yaml
EmilienM: done

EmilienM:
i've re-added the upgrade job.

mandre (commented):
/lgtm
/approve

PR #44835, opened 2023-10-25
Update quay-test-master to 3.9 in 4.14

oharan2:
Quay E2E tests with 3.8 is failing:
https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-quay-quay-tests-master-ocp414-quay38-quay-e2e-tests-quay38-ocp414-lp-interop/1713797633807486976

PR #45164, opened 2023-11-01
Rename file of Interop quay-e2e-tests and disable job on OCP 4.14

oharan2:
Following #44835 

1. Accidentally duplicated config files
quay-quay-tests-master__`ocp-414`-quay39.yaml
quay-quay-tests-master__`ocp414`-quay39.yaml
Fixed with uniquely specifying  `-lp-interop` in the new config file.

2. Disabled test on OCP 4.14 by setting cron value to `0 6 25 10 *`
**Note: This test stanza is getting placed here for reference to prepare for the interop programs OCP z-stream n-1 testing. The cron used is set to run a little less then a year from now on Oct 25th. This is essentially just a strategy to prevent this from running until we kick off the n-1 testing.**

### DIFF
--- a/ci-operator/jobs/quay/quay-tests/quay-quay-tests-master-periodics.yaml
+++ b/ci-operator/jobs/quay/quay-tests/quay-quay-tests-master-periodics.yaml
@@ -595,7 +595,7 @@ periodics:
         secretName: result-aggregator
 - agent: kubernetes
   cluster: build03
-  cron: 0 6 25 12 *
+  cron: 0 6 25 10 *
   decorate: true
   decoration_config:
     skip_cloning: true


ci-operator/config/quay/quay-tests/quay-quay-tests-master__ocp-414-quay39.yaml
```diff
@@ -43,6 +43,34 @@ tests:
     - ref: quay-tests-qbo-qe-test
     - ref: quay-tests-cso-qe-test
     workflow: cucushift-installer-rehearse-aws-ipi
+- as: quay-e2e-tests-quay39-ocp414-lp-interop
+  cron: 0 6 25 12 *
```
ascerra, line 47:
```suggestion
  cron: 0 6 25 10 *
```

PR #45173, opened 2023-11-01
ci-images-mirror: ignore syncing broken images (5)

deepsm007:
The image cannot be pulled because it was imported from [regstry.svc.ci.openshift.org](http://regstry.svc.ci.openshift.org/) which has been offline 2 years ago.


deepsm007:
Based on discussion with @hongkailiu we do not need this its a leftover temporary image

PR #45182, opened 2023-11-01
perf/scale: fix density-v2 vs cni test confusion in networking

dcbw:
The 120 node density-cni test actually ran the density-v2 workload; fix that to run density-cni.

ovnkube master branch didn't have any perf/scale tests at all, probably as a result of the confusion around branch date and merging.

@jluhrsen @jtaleric


jtaleric:
> ovnkube master branch didn't have any perf/scale tests at all, probably as a result of the confusion around branch date and merging.

Yeah - and a bit me learning the ropes... Thanks!

jtaleric:
@dcbw I think you might of ran too many large jobs all at once.

PR #45220, opened 2023-11-02
Sync Serverless CI

serverless-qe:


ReToCode (approved):
/lgtm
/approve

/pj-rehearse max

ReToCode:
@pierDipi not sure about the older eventing-kafka-broker CI. Do you want to re-run or merge and (maybe) fix later?

mgencur:
@ReToCode I think we should remove all CI configs for unsupported versions from all repositories. They're not required anymore. They only slow down CI updates like this one and they can be found in GIT history is required (but I doubt they will be required again).

PR #45221, opened 2023-11-02
CVO: Require `qe-approved` or `no-qe` for development branch

petr-muller:
We like our PRs tested before merge but we forget to hold them. Therefore, we will require either `qe-approved` or `no-qe` label as a merge requirement.

This should work based on https://github.com/kubernetes/test-infra/pull/27257 but it seems nobody else uses the feature in OpenShift so we need to keep an eye on what's happening.

### DIFF
--- a/core-services/prow/02_config/_labels.yaml
+++ b/core-services/prow/02_config/_labels.yaml
@@ -994,3 +994,10 @@ repos:
         name: regression_complete
         target: both
         addedBy: label
+  openshift/cluster-version-operator:
+    labels:
+      - color: c7def8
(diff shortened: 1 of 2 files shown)


petr-muller:
/hold

Let's discuss whether we want it or not.

/label qe-approved
Can I label random stuff?

wking (approved):
/lgtm

wking (approved):
Looks good to  me.  Holding in case anyone else wants to review, but also,  reverts are cheap so no worries from me if you want to lift the hold and experiment.

/lgtm
/hold

PR #25152, opened 2022-01-11
cincinnati-operator: test operator against released OSUS version and latest master

vrutkovs:
In order to prevent regressions in OSUS and the operator we should extend test matrix to include both latest released version and latest code from master

PR #17435, opened 2021-04-05
ci-operator/config/openshift/cincinnati-operator: Set RELATED_IMAGE_*

wking:
From [the OSBS docs][1]:

> Before OSBS can pin your pullspecs, it first needs to find them. Because it is practically impossible to tell if a string is a pullspec, atomic-reactor has a predefined set of locations where it will look for pullspecs.
>
> 1. metadata.annotations.containerImage anywhere in the file
>
>     ```yaml
>     jq: .. | .metadata?.annotations.containerImage | select(. != null)
>     ```
>
> 2. All containers in each deployment
>
>     ```yaml
>     jq: .spec.install.spec.deployments[].spec.template.spec.containers[]
>     ```
>
> 3. All initContainers in each deployment
>
>     ```yaml
>     jq: .spec.install.spec.deployments[].spec.template.spec.initContainers[]
>     ```
>
> 4. All RELATED_IMAGE_* variables for all containers and initContainers
>
>     ```yaml
>     jq: .env[] | select(.name | test("RELATED_IMAGE_")) for each of [2], [3]
>     ```
>
> 5. All pullspecs from all annotations. This is done heuristically (OSBS needs to guess what might be a pullspec). See heuristic annotations below...

This change allows us to pivot to the approach from (4) for 4.6 and
later, [rhbz#1945026][2] and openshift/cincinnati-operator#104.

[1]: https://osbs.readthedocs.io/en/latest/users.html#pullspec-locations
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=1945026

PR #12486, opened 2020-10-06
ci-operator/config/openshift/cincinnati-operator: Move e2e-operator to multi-step

wking:
Using the `openshift-e2e-gcp` workflow and overriding the `test` step per [these docs][1] to run our operator tests instead of the usual e2e suite.

I've dropped `cincinnati` from the job name, because this presubmit only runs in the cincinnati-operator repository.  The fact that it is `operator-e2e` is sufficient to distinguish from other presubmits in that repository.

I've dropped `aws` from the job name, because we are [platform-agnostic][2].

Generated by editing `ci-operator/config` and then running:

```console
$ make update
```

WIP because once we get a green rehearsal I'll extend this to cover 4.6+.

[1]: https://steps.ci.openshift.org/help#config
[2]: https://github.com/openshift/release/tree/e45d68dc29ab557cc887eb84ada146a857ae88f9/ci-operator/platform-balance#rebalancing-tests-among-platforms

PR #45245, opened 2023-11-02
ci-operator/config/openshift/cincinnati-operator: operator-e2e-old-ocp-published-graph-data, etc.

wking:
Moving to a recent Go builder, based on [these docs][1] and:

```console
$ oc -n ocp get -o json imagestream builder | jq -r '.status.tags[] | select(.items | length > 0) | .items[0].created + " " + .tag' | sort | grep golang
...
2023-11-02T19:53:15Z rhel-8-golang-1.18-openshift-4.11
2023-11-02T19:53:23Z rhel-8-golang-1.17-openshift-4.11
2023-11-02T20:49:19Z rhel-8-golang-1.19-openshift-4.13
2023-11-02T20:49:25Z rhel-9-golang-1.19-openshift-4.13
2023-11-02T21:54:25Z rhel-9-golang-1.20-openshift-4.14
2023-11-02T21:54:46Z rhel-8-golang-1.20-openshift-4.14
2023-11-02T21:55:24Z rhel-8-golang-1.19-openshift-4.14
2023-11-02T21:55:29Z rhel-9-golang-1.19-openshift-4.14
```

I'm dropping the `build_root` stanza, because we didn't seem to need [the functionality it delivers][2], and we need a UBI builder for the graph-data image and a Go builder for the operator image.

[The `operators` stanza][3] remains largely unchanged, although I did rename `cincinnati_operand_latest` to `cincinnati-operand`, because these tests use a single operand image, and there is no need to distinguish between multiple operand images with `latest`.

The image used for operator-sdk (which I bump to an OpenShift 4.14 base) and its use are [doc'ed here][4].  The 4.14 cluster-claim pool I'm transitioning to is [listed as healthy][5].

For the end-to-end tests, we install the operator via the test suite, so we do not need the SDK bits.  I've dropped `OPERATOR_IMAGE`, because we are well past the transition initiated by eae9d3813b (#17435) and
openshift/cincinnati-operator@799d18525b (openshift/cincinnati-operator#104).

I'm consistently using the current Cincinnati operand instead of the pinned one, because we ship the OpenShift Update Service Operator as a bundle with the operator and operand, and while it might be useful to grow update-between-OSUS-releases test coverage, we do not expect long durations of new operators coexisting with old-image operand pods. And we never expect new operators to touch Deployments with old operand images, except to bump them to new operand images.  We'd been using digest-pinned operand images here since efcafb613f (#12486), where I said:

> In a future pivot we'll pull the operand image out of CI too, instead of hard-coding.  But with this change we at least move the hard-coding into the CI repository.

4f46d7e84b (#25152) brought in that floating operand image, but neglected, for reasons that I am not clear on, did not drop the digest-pinned operand.  I'm dropping it now.

With "which operand image" removed as a differentiator, the remaining differentiators for the end-to-end tests are:

* Which host OpenShift?
  * To protect from "new operators require new platform capabilities not present in older OpenShift releases", we have an `old-ocp` job.  It's currently 4.11 for [the oldest supported release][6].
  * To protect from "new operators still use platform capabilities that have been removed from development branches of OpenShift", we have a `new-ocp` job.  It's currently 4.15.
  * To protect against "HyperShift does something the operator does not expect", we have a `hypershift` job.  This job currently defers "which version?" to the workflow, because we do not expect HyperShift-specific difference to evolve much between 4.y releases, while the APIs used by the operator (Deployments, Services, Routes, etc.) might.  We could revisit this and launch `old-hypershift` and `new-hypershift` flavors in the future if we see a need.
  * I'm not worrying about enumerating all the current 4.y options like we had done before.  That is more work to maintain, and renaming required jobs confuses Prow and requires an /override of the removed job.  It seems unlikely that we work on 4.old, break on some 4.middle, and work again on 4.dev.  Again, we can always revisit this if we change our minds about the exposure.

* Which graph-data?
  * To protect against "I updated my OSUS without changing the graph-data image, and it broke", we have `published-graph-data` jobs.  These consume images that were built by previous postsubmits in the cincinnati-graph-data repository.
  * We could theoretically also add coverage for older forms of graph-data images we suspect customers might be using.  I'm punting this kind of thing to possible future work, if we decide the exposure is significant enough to warrant ongoing CI coverage.
  * To allow testing new features like serving signatures, we have a `local-graph-data` job.  This consumes a graph-data image built from steps in the operator repository, allowing convenient testing of changes that simultaneously tweak the operator and how the graph-data image is built.  For example, openshift/cincinnati-operator#176 injects an image signature into graph-data, and updates graph-data to serve it. I'm setting a `GRAPH_DATA` environment variable to `local` to allow the test suite to easily distinguish this case.

[1]: https://docs.ci.openshift.org/docs/architecture/images/#ci-images
[2]: https://docs.ci.openshift.org/docs/architecture/ci-operator/#build-root-image
[3]: https://docs.ci.openshift.org/docs/how-tos/testing-operator-sdk-operators/#building-operator-bundles
[4]: https://docs.ci.openshift.org/docs/how-tos/testing-operator-sdk-operators/#simple-operator-installation
[5]: https://docs.ci.openshift.org/docs/how-tos/cluster-claim/#existing-cluster-pools
[6]: https://access.redhat.com/support/policy/updates/openshift/#dates

### DIFF
--- a/ci-operator/config/openshift/cincinnati-operator/openshift-cincinnati-operator-master.yaml
+++ b/ci-operator/config/openshift/cincinnati-operator/openshift-cincinnati-operator-master.yaml
@@ -1,32 +1,40 @@
 base_images:
-  base:
-    name: "4.12"
-    namespace: ocp
-    tag: base
-  cincinnati_operand_latest:
+  cincinnati-operand:
     name: cincinnati-build-root
     namespace: cincinnati-ci
     tag: deploy
-  graphdata:
+  golang:
+    name: builder
+    namespace: ocp
+    tag: rhel-9-golang-1.20-openshift-4.14
+  operator-sdk:
+    name: "4.14"
+    namespace: origin
+    tag: operator-sdk
+  published-cincinnati-graph-data:
     name: cincinnati-graph-data
     namespace: cincinnati-ci-public
     tag: stable
-  operator-sdk:
-    name: "4.13"
-    namespace: origin
-    tag: operator-sdk
+  ubi:
+    name: ubi
+    namespace: ocp
+    tag: "9"
 build_root:
   image_stream_tag:
-    name: release
-    namespace: openshift
-    tag: golang-1.18
+    name: builder
+    namespace: ocp
+    tag: rhel-9-golang-1.20-openshift-4.14
 images:
 - dockerfile_path: build/Dockerfile
-  from: base
+  from: ubi
+  inputs:
+    golang:
+      as:
+      - registry.ci.openshift.org/openshift/release:golang-1.18
   to: cincinnati-operator
 - dockerfile_path: dev/Dockerfile
-  from: base
-  to: cincinnati-graph-data-container
+  from: ubi
+  to: local-cincinnati-graph-data
 operator:
   bundles:
   - as: cincinnati-bundle
@@ -34,22 +42,14 @@ operator:
     skip_building_index: true
   substitutions:
   - pullspec: quay.io/cincinnati/cincinnati:latest
-    with: pipeline:cincinnati_operand_latest
+    with: pipeline:cincinnati-operand
   - pullspec: controller:latest
     with: pipeline:cincinnati-operator
 releases:
   latest:
     release:
       channel: candidate
-      version: "4.14"
-  thirteen:
-    release:
-      channel: candidate
       version: "4.13"
-  twelve:
-    release:
-      channel: candidate
-      version: "4.12"
 resources:
   '*':
     limits:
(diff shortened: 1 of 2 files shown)


wking:
All green :)

/pj-rehearse ack

petr-muller:
/hold

I'm very likely to LGMT this but I think we have opportunity to simplify a bit further, need to check out something (will not hold for long)

petr-muller (approved):
LGTM

I thought I can simplify this a bit but I cannot. I am leaving the hold in place b/c @LalatenduMohanty wanted to take a look, but feel free to lift the hold anytime.

PR #45248, opened 2023-11-03
Rename sanity test 2

svetsa-rh:
Sanity test for self managed and managed for renaming jobs. 
want to confirm dashboards look fine after renaming.

trying self-managed and managed, in case we run into issues with managed tests.


krishvoor:
Approving this for the tests to get running periodically.

krishvoor (approved):
/lgtm

PR #45255, opened 2023-11-03
serverless kitchensink, set more realistic test container reqs/limits

maschmid:


maschmid:
/pj-rehearse ack

failed on go test timeouts, which we need to fix in serverless-operator repo

PR #45268, opened 2023-11-03
fix: update slack alert

vprashar2929:
This PR adds `@powermon-ci-watchers` annotation in the slack alert message for fail test runs.


sthaha (approved):
/lgtm

PR #45270, opened 2023-11-03
OPRUN-3082: Remove verify-commits-shared test

awgreene:
The operator framework portfolio consists of a number of projects that have an upstream project. The verify-commits makefile target had been introduced in each of the downstream  projects to assist with the downstreaming efforts. The commit checking logic was eventually moved into a container and was rolled out as the verify-commits-shared test, which ran alongside the verify-commits test to ensure that it worked as expected.

We are now confident that the verify-commits-shared test is running as expected, and can replace the logic used in the verify-commits test with that of the verify-commits-shared test, which will allow us to remove the verify-commits Makefile targets from each of the repos, simplifying the code base and reducing duplication of code.


ncdc (approved):

ncdc:
Rehearsals on verify-commits jobs will never pass because the commit they use is coming from this repo/PR, not from the target repo.

/pj-rehearse ack

PR #45271, opened 2023-11-03
Require qe-approved in alibaba CSI driver + operator PRs

jsafrane:
Deleting rules that allow a PR to be merged to master without px-approved. docs-approved and qe-approved.

There already are rules above that will insist on these labels to be present.

A driver rebase has slipped into master without proper testing: https://github.com/openshift/alibaba-cloud-csi-driver/pull/41

cc @openshift/storage


bertinatto (commented):
/lgtm

PR #45284, opened 2023-11-03
chore(stackrox): move 4.1 to ACS owned clusters

gavin-stackrox:
/cc @davdhacs
Also removes prior 4.12 latest - should have been removed 1 month after 4.13 GA


gavin-stackrox:
/pj-rehearse ack
The failures are concerning - possibly OCP 4.11 related flakes or flakes not fixed on the ACS 4.1 branch. From the point of view of the version switch it looks good.

PR #45288, opened 2023-11-05
HyperShift-QE: fix hypershift-agent-mce-guest

LiangquanLi930:
For MCE part, if needed to enable
```
- ref: cucushift-hypershift-extended-idp-htpasswd
- ref: cucushift-hypershift-extended-enable-guest
```
we need to set env:`HYPERSHIFT_NAMESPACE` as `"local-cluster"`


LiangquanLi930:
https://issues.redhat.com/browse/OCPQE-17597

PR #45208, opened 2023-11-02
cover capability baselineCapabilitySet:none on upi

jinyunma:
1.  move feature aws customer-security-groups test from job `aws-ipi-baselinecaps-none-custom-sg-f14` to job `aws-ipi-usertags-custom-sg-f28`
2.  remove duplicated test coverage : job `aws-ipi-baselinecaps-none-f28` and `aws-ipi-disconnected-private-sno-baselinecaps-none-f28`
3.  add new job `aws-upi-baselinecaps-none-f28`

PR #45308, opened 2023-11-06
fix incorrect BASE_DOMAIN setting in capability aws upi job

jinyunma:
fix issues introduced in https://github.com/openshift/release/pull/45208


jinyunma:
cc @liangxia to review

PR #45335, opened 2023-11-06
MGMT-16073: Change test-infra jobs to use 4.15

danmanor:
Change test-infra jobs to use `4.15`


danmanor:
We should wait for https://github.com/openshift/assisted-test-infra/pull/2301 right ?

PR #45336, opened 2023-11-06
MCE hypershift addon 2.5 cutover

rokej:
MCE hypershift addon 2.5 cutover


mikeshng (approved):
/approve

/lgtm

did some sanity check on versions

PR #45340, opened 2023-11-06
[LPTOCPCI-1007] Fix interop firewatch configs - missing config values

calebevans:


ci-operator/config/stackrox/stackrox/stackrox-stackrox-master__ocp-4-15-lp-interop.yaml
```diff
@@ -33,6 +33,7 @@ tests:
          {"step": "gather-*", "failure_type": "pod_failure", "classification": "Other", "jira_project": "LPINTEROP", "group": {"name": "cluster", "priority": 2}, "jira_additional_labels": ["4.15-lp","self-managed-lp","acs-lp"]},
          {"step": "gather-*", "failure_type": "test_failure", "classification": "NONE", "jira_project": "NONE", "ignore": "true"},
          {"step": "openshift-*", "failure_type": "all", "classification": "Other", "jira_project": "LPINTEROP", "group": {"name": "cluster", "priority": 2}, "jira_additional_labels": ["4.15-lp","self-managed-lp","acs-lp"]},
+         {"step": "stackrox-*", "failure_type": "pod_failure", "classification": "ACS Test Execution", "jira_project": "LPINTEROP", "jira_additional_labels": ["4.15-lp","self-managed-lp","acs-lp","interop-tests"]},
```
ascerra, line 36:
```suggestion
         {"step": "stackrox-*", "failure_type": "pod_failure", "classification": "ACS Test Execution", "jira_project": "ROX", "jira_additional_labels": ["4.15-lp","self-managed-lp","acs-lp","interop-tests"]},
```

I'm leaning towards sending pod failures to PQE Jira projects from now on. Most pod failures that we see are not due to the build-clusters pod going down, it has been mostly due to the PQE test suite ending with exit 1

calebevans: We'll have to change it in every scenario then. I would just remove the two rules and change the failure type to all. Is this something you want to bring to Badre and the team first?

ascerra: Right changing it to all makes sense to me, each scenario might be different though.
- Some are test suite run commands are configured to never return exit 1, so in that case a pod failure would make sense to be sent to use
- For scenarios that don't force exit 0 on their test run command then the pod failure is probably coming from the test run exiting.

I think it makes sense to parking lot this tomorrow or even just chat on slack about it.

calebevans: Do you want to merge this change and talk about making a mass change in parking lot tomorrow or do you want to wait?

oharan2: I support this transition, and the logic @ascerra described here is something that can already be factored into scenarios that currently onboarding.

PR #45356, opened 2023-11-07
Remove unused field time from junit xml

liangxia:
The `time` field indicates the execution time of the test cases. Instead of hard code to `1`, we can omit it.

/cc @JianLi-RH @dis016 @pruan-rht @jhou1 @jianlinliu


JianLi-RH:
In my impression time attribute is required for ReportPortal, without it we can NOT import junit file to ReportPortal.

liangxia:
> In my impression time attribute is required to ReportPortal, without it we can NOT import junit file to ReportPortal.

It's not required. Example [log](https://gcsweb-qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/qe-private-deck/logs/periodic-ci-openshift-openshift-tests-private-release-4.11-amd64-nightly-aws-ipi-imdsv2-fips-f28/1721607943792627712/artifacts/aws-ipi-imdsv2-fips-f28/openshift-extended-web-tests/artifacts/gui_test_screenshots/console-cypress.xml).

JianLi-RH:
Yes, you are right.

/lgtm

PR #45373, opened 2023-11-07
`managed-clonerefs`: Multiarchitecture build

danilo-gemoli:
Add multiarch support to `managed-clonerefs`.

/cc @droslean

PR #45446, opened 2023-11-08
Add a periodic that triggers the managed-clonerefs mabc

danilo-gemoli:
Add a new periodic that deletes and recreates the `managed-clonerefs` mabc.
It needs [openshift/ci-tools#3756](https://github.com/openshift/ci-tools/pull/3756) and [openshift/release#45373](https://github.com/openshift/release/pull/45373)

/cc @droslean @openshift/test-platform

### DIFF
--- a/clusters/build-clusters/common/assets/admin_multi-arch-builder-controller_rbac.yaml
+++ b/clusters/build-clusters/common/assets/admin_multi-arch-builder-controller_rbac.yaml
@@ -23,6 +23,7 @@ rules:
   resources:
   - multiarchbuildconfigs
   - multiarchbuildconfigs/status
+  - multiarchbuildconfigs/finalizers
   verbs:
   - get
   - list


ci-operator/jobs/infra-periodics.yaml
```diff
@@ -2616,3 +2616,59 @@ periodics:
...
+      args:
+      - -c
+      - |
+        set -o errexit
+        set -o nounset
+        set -o pipefail
+
+        export KUBECONFIG='/etc/build-farm-credentials/kubeconfig'
+
+        curl -sL 'https://github.com/jqlang/jq/releases/download/jq-1.7/jq-linux-amd64' >/tmp/jq
```
droslean, line 2643: We shouldn't download the `jq` each time the job runs. Instead, we need to maintain a specific image and move the rest of the logic to a script.

ci-operator/jobs/infra-periodics.yaml
```diff
@@ -2616,3 +2616,59 @@ periodics:
...
+        managed_clonerefs=$(oc -n ci get mabc/managed-clonerefs -o json)
+        if [ -z "$(/tmp/jq -r '.status.state' <<<$managed_clonerefs)" ]; then
+            echo 'mabc/managed-clonerefs is running already'
+            exit 0
+        fi
+
+        oc -n ci delete --cascade=foreground --wait=true mabc/managed-clonerefs
+        echo "mabc/managed-clonerefs deleted"
+
+        oc -n ci apply --wait=true -f 'clusters/build-clusters/multi01/supplemental-ci-images/manage-clonerefs.yaml'
```
droslean, line 2655: This will failed. You need to take care of the `serviceAccount` as well.

ci-operator/jobs/infra-periodics.yaml
```diff
@@ -2616,3 +2616,56 @@ periodics:
       resources:
         requests:
           cpu: 500m
+- agent: kubernetes
+  cluster: app.ci
```
droslean, line 2620: This needs to run on `multi01` cluster.

ci-operator/jobs/infra-periodics.yaml
droslean: I am not sure. Let's config that the `config-update` has rbacs to manage the mabc resources.

ci-operator/jobs/infra-periodics.yaml
danilo-gemoli: `registry.ci.openshift.org/ocp/cli-jq:latest` is amd64 only. Should we add a node selector?

ci-operator/jobs/infra-periodics.yaml
danilo-gemoli: This is the output from a pod running `ocp/cli-jq:latest` on `multi01`:
```bash
sh-4.4# oc whoami
system:serviceaccount:ci:config-updater
sh-4.4# oc auth can-i '*' mabc
yes
```

ci-operator/jobs/infra-periodics.yaml
droslean: Yes the node selector would be required as well.

ci-operator/jobs/infra-periodics.yaml
droslean: I don't agree on using the config-updater account for that... I would prefer having a different service account just for managing the mabc for that specific case. @hongkailiu Can add more context here.

danilo-gemoli: In the end the only thing this periodic does is applying and deleting a resource. Isn't the `config-updater` SA meant to handle such a case? `applyconfig` does more or less the same and it leverages `configu-updater` capabilities.

droslean: This job is not related to the applyconfig. Instead this job is doing a specific task on specific resources. We should use more permissions that the job needs.

droslean (approved):

PR #45457, opened 2023-11-08
Adding CI config for openshift lightspeed-service

vishnuchalla:
Adding CI config to onboard openshift lightspeed-service on PROW.

Followed these on-boarding steps [here](https://docs.ci.openshift.org/docs/how-tos/onboarding-a-new-component/).

### DIFF
--- a/ci-operator/config/openshift/lightspeed-service/OWNERS
+++ b/ci-operator/config/openshift/lightspeed-service/OWNERS
@@ -1,0 +1,9 @@
+approvers:
+- bparees
+- thoraxe
+- vishnuchalla
+options: {}
+reviewers:
+- bparees
+- thoraxe
+- vishnuchalla
--- a/ci-operator/jobs/openshift/lightspeed-service/openshift-lightspeed-service-main-presubmits.yaml
+++ b/ci-operator/jobs/openshift/lightspeed-service/openshift-lightspeed-service-main-presubmits.yaml
@@ -1,0 +1,137 @@
+presubmits:
+  openshift/lightspeed-service:
+  - agent: kubernetes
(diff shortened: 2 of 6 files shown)


ci-operator/config/openshift/lightspeed-service/OWNERS
```diff
@@ -0,0 +1,15 @@
+# DO NOT EDIT; this file is auto-generated using https://github.com/openshift/ci-tools.
+# Fetched from https://github.com/openshift/library-go root OWNERS
```
bparees, line 2: this doesn't seem right....

vishnuchalla: Yes you are right. I just forgot to remove those comments from the template code. It wasn't auto-generated. I created these files manually.

ci-operator/jobs/openshift/lightspeed-service/openshift-lightspeed-service-main-presubmits.yaml
```diff
@@ -0,0 +1,137 @@
+presubmits:
+  openshift/lightspeed-service:
+  - agent: kubernetes
+    always_run: true
```
bparees, line 4: let's make this false for now, there's no reason to run an e2e job that's going to spin up a full cluster just for us to do nothing with it.  We can change it when we get to the point of actually wanting to deploy something in a cluster or otherwise need a cluster for something.

vishnuchalla: Yes sounds good to me. Changed as suggested.

PR #45458, opened 2023-11-08
Build stolostron/kube-state-metrics with Go 1.20

douglascamata:
This is supposed to match with https://github.com/stolostron/kube-state-metrics/pull/46.


douglascamata:
I figured another reason for the build failure. Closing this for now, but might reopen later if the upgrade will be required.

PR #45466, opened 2023-11-08
Move all self managed AWS tests to common workflow

svetsa-rh:
Move all existing self managed AWS tests to common workflow


jtaleric (approved):
/lgtm

PR #45475, opened 2023-11-09
AGENT-725: Add presubmit job for vsphere

rwsu:
The job runs the compact_ipv4 scenario for the vsphere platform.

The "agent wait-for install-complete" step is skipped because install-complete does not succeed for vsphere using libvirt virtual machines.

Depends on: https://github.com/openshift-metal3/dev-scripts/pull/1599

### DIFF
--- a/ci-operator/jobs/openshift/installer/openshift-installer-master-presubmits.yaml
+++ b/ci-operator/jobs/openshift/installer/openshift-installer-master-presubmits.yaml
@@ -2253,6 +2253,89 @@ presubmits:
     - ^master$
     - ^master-
     cluster: build05
+    context: ci/prow/e2e-agent-compact-ipv4-vmware
+    decorate: true
+    decoration_config:
+      timeout: 6h0m0s
+    labels:
+      ci-operator.openshift.io/cloud: equinix-ocp-metal
+      ci-operator.openshift.io/cloud-cluster-profile: equinix-ocp-metal
+      ci-operator.openshift.io/cluster: build05
+      ci.openshift.io/generator: prowgen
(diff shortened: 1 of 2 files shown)


zaneb:
I want to cut down the number of CI jobs we have already. Since this can't replace another CI job (because it doesn't run to install-complete) I don't think we should add this with `run_if_changed`. Let's keep it with only `run_always: false`.

rwsu:
/pj-rehearse pull-ci-openshift-installer-master-e2e-agent-compact-ipv4-vsphere

ci-operator/jobs/openshift/installer/openshift-installer-master-presubmits.yaml
```diff
@@ -2247,6 +2247,89 @@ presubmits:
...
+    labels:
+      ci-operator.openshift.io/cloud: equinix-ocp-metal
+      ci-operator.openshift.io/cloud-cluster-profile: equinix-ocp-metal
+      ci-operator.openshift.io/cluster: build05
+      ci.openshift.io/generator: prowgen
+      pj-rehearse.openshift.io/can-be-rehearsed: "true"
+    name: pull-ci-openshift-installer-master-e2e-agent-compact-ipv4-vmware
+    optional: true
+    rerun_command: /test e2e-agent-compact-ipv4-vmware
+    run_if_changed: /agent/
```
zaneb, line 2269: I think we decided to take this out, but once it is in the job is gets preserved even if you take it out of the config. (In the old days, you couldn't even put this stuff in the config, and many jobs still haven't so we have to continue preserving the data in the job because it's the only place it is recorded in many cases.) It pays to delete the job from this file and regenerate it to make sure it matches the config.

PR #45479, opened 2023-11-09
remove wrong distuptive filters in some jobs

sunzhaohua2:
@liangxia PTAL

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-nightly.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-nightly.yaml
@@ -2324,7 +2324,7 @@ tests:
       E2E_RUN_TAGS: '@amd64 and @nutanix-ipi and @disconnected and @network-ovnkubernetes
         and not @fips'
       TAG_VERSION: '@4.12'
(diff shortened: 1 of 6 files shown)


ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-nightly.yaml
```diff
@@ -2324,7 +2324,6 @@ tests:
       E2E_RUN_TAGS: '@amd64 and @nutanix-ipi and @disconnected and @network-ovnkubernetes
         and not @fips'
       TAG_VERSION: '@4.12'
-      TEST_FILTERS: ~ChkUpgrade&;~NonPreRelease&;~Serial&;~Disruptive&;~ConnectedOnly&;~HyperShiftMGMT&;~MicroShiftOnly&
```
kuiwang02, line 2327: @sunzhaohua2 please do not remove it, and it should be `TEST_FILTERS: ~ChkUpgrade&;~NonPreRelease&;Serial&;Disruptive&;~ConnectedOnly&;~HyperShiftMGMT&;~MicroShiftOnly&`

same rule for others.

liangxia, line 2327: For disconnected profiles, we need to overwrite `TEST_FILTERS`, because the default contains `~DisconnectedOnly&`.  So instead of removing this line, we should fix it.

PR #45503, opened 2023-11-09
telco5g: Increase job timeout for cnftests jobs

Sandeepyadav93:
cnftests jobs sometimes take more than 4 hours to deploy, with this patch we are increasing the job timeout to 6 hours, and also increase the cnf test step timeout to 4h15m.

In a followup, we plan to add deploy cluster in job itself instead of precreating OCP clusters.

[CNF-9805](https://issues.redhat.com//browse/CNF-9805)

### DIFF
--- a/ci-operator/config/openshift/release/openshift-release-master__nightly-4.13.yaml
+++ b/ci-operator/config/openshift/release/openshift-release-master__nightly-4.13.yaml
@@ -862,6 +862,7 @@ tests:
       enable:
       - observers-resource-watch
     workflow: telco5g-origin
+  timeout: 6h0m0s
 - as: e2e-telco5g-cnftests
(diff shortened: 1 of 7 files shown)


ci-operator/config/openshift/release/openshift-release-master__nightly-4.13.yaml
```diff
@@ -862,6 +862,7 @@ tests:
       enable:
       - observers-resource-watch
     workflow: telco5g-origin
+  timeout: 6h
```
Sandeepyadav93, line 865: fixing

Sandeepyadav93: done

PR #45531, opened 2023-11-10
update code to remove deploy dittybopper ref

liqcui:
It's not necessary we have to deploy dittybopper in all perfscale prow ci job, so we remove - ref: openshift-qe-deploy-dittybopper from create-infra-move-ingress-monitoring-registry-chain.yaml. so we can deploy dittybopper separatedly according to your requirements


liqcui:
@qiliRedHat @paigerube14 Please review my PR when you are available, thanks!

jtaleric (approved):
/lgtm

PR #45551, opened 2023-11-10
test: add two more regression test cases to fm healthcheck tests

pawelpaszki:
## WHAT
test: add two more regression test cases to fm healthcheck tests

## local verification
```
Checking MC MHC match expressions operator
Checking that MHC health check excludes 'master' and 'infra' machines
Checking that MHC health check includes 'worker' machines
Checking that compliance-monkey deployment is present and contains descheduler container

Passed tests:

OCPQE-17157
OCPQE-17578

Failed tests:
```


pawelpaszki:
@jianping-shu - can you take a look please?

PR #45566, opened 2023-11-10
stackrox: use credentialsMode image

gavin-stackrox:
`0.8.7` is the flavor with credentialsMode support to allow us to scale up clusters.

Given the challenges I've had onboarding this to infra, I'm suspicious that something might have changed in the 4.14 installer to break automation-flavor installs. We shall see.

/cc @davdhacs


gavin-stackrox:
Oooo all green.
/pj-rehearse ack
@davdhacs can I get an `/lgtm`
/honk

PR #45572, opened 2023-11-10
crio: fix crio build

rphillips:
add crio-gcloud-base to promoted images


ci-operator/config/cri-o/cri-o/cri-o-cri-o-main__ci.yaml
```diff
@@ -29,6 +29,8 @@ images:
   from: base
   to: crio-gcloud-base
 promotion:
+  additional_images:
+    crio-gcloud-base: crio-gcloud-base
   excluded_images:
   - crio-crio-base-src
   namespace: ci
```
sohankunkerkar, line 36: use `cri-o` here and see if that helps

PR #45583, opened 2023-11-12
Revert "Remove unused field time from junit xml"

pruan-rht:
Reverts openshift/release#45356

Taking this field out is causing the classifier job to fail
https://redhat-internal.slack.com/archives/C01UHADN91U/p1699750794514019?thread_ts=1699679495.124369&cid=C01UHADN91U

We can either fix the classifier script or revert the original PR


pruan-rht:
@liangxia PTAL and let us know if you are ok with this workaround

liangxia:
If RP does not require this field, why we need it in classifier job ?  Does it mean it's a bug in classifier job and we should fix that job, rather than introduce a pseudo time here ?

pruan-rht:
closing since this is addressed by https://gitlab.cee.redhat.com/aosqe/prow-test-results-classfier/-/merge_requests/213 now

PR #45618, opened 2023-11-13
[openstack-operators] suppress cleanup errors in manila tempest tests

gouthampacha:
We're seeing sporadic test failures where cleanups work, but produce intermediate errors. These are wasting CI cycles; while the team investigates and fixes bugs in the tests, we should temporarily suppress these errors for the sake of a more stable CI job.


fmount (approved):
/lgtm

viroel:
/pj-rehearse ack

change will only affect manila-build-deploy-tempest job, and it is also a temporary fix.

PR #45619, opened 2023-11-13
Add a new layer to claim to check if all external services used by rhtap e2e are alive

flacatus:
TSIA

### DIFF
--- a/ci-operator/step-registry/redhat-appstudio/external-services-health/redhat-appstudio-external-services-health-ref.metadata.json
+++ b/ci-operator/step-registry/redhat-appstudio/external-services-health/redhat-appstudio-external-services-health-ref.metadata.json
@@ -1,5 +1,5 @@
 {
-	"path": "redhat-appstudio/clean-external-resources/redhat-appstudio-external-services-health-ref.yaml",
+	"path": "redhat-appstudio/external-services-health/redhat-appstudio-external-services-health-ref.yaml",
 	"owners": {
 		"approvers": [
 			"flacatus",


flacatus:
/pj-rehearse pull-ci-redhat-appstudio-application-service-main-application-service-e2e

flacatus:
/pj-rehearse pull-ci-redhat-appstudio-application-service-main-application-service-e2e

ci-operator/step-registry/redhat-appstudio/external-services-health/redhat-appstudio-external-services-health-commands.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+# Check external services which are dealing RHTAP tests are alive. In case are not alive will fail and not claim a cluster to install
+
+set -e
+
+UNAOPERATIONAL_GITHUB_SERVICES="$(curl https://www.githubstatus.com/api/v2/summary.json | jq '.components[] | select(.status != "operational")')"
+UNAOPERATIONAL_QUAYIO_SERVICES="$(curl https://status.quay.io/api/v2/summary.json | jq '.components[] | select(.status != "operational")')"
```
psturc, line 8: not sure if we should fail the job in case the status of any of the service's component is **not equal** to `operational`

for example since yesterday quay is reporting this status
```
{
  "id": "72xlygjldsvt",
  "name": "Registry",
  "status": "degraded_performance",
  "created_at": "2014-06-10T14:58:44.723-04:00",
  "updated_at": "2023-11-13T17:04:51.796-05:00",
  "position": 1,
  "description": null,
  "showcase": false,
  "start_date": null,
  "group_id": null,
  "page_id": "8szqd6w4s277",
  "group": false,
  "only_show_if_degraded": false
}
```
Screenshot 2023-11-14 at 13 14 47

which doesn't seem to affect our CI very much at this moment. similar situation could be with e.g. github actions - if github actions doesn't work properly, that doesn't necessarily mean that it will affect our CI.

what I'm thinking we should start with is to keep this pre-step, but do not fail CI jobs yet. we should first figure out what are the key components that will definitely affect our CI when degraded.

wdyt @flacatus @Dannyb48 @rhopp?

Dannyb48: ack, you bring up a valid point without knowing which component impacts the actual environment it may tough to fail out right yet. So it a minimum record it as part of every job run and I wonder if we should save it as a json file to be collected and if this is something the post step can capture as part of summary?

That said for I would be fine comprising that on Github we just record and print until we get a better understanding of which components but in the case of quay.io we should actually go beyond record, print, fail. Since we've had 3 recent quay.io outages could we make a best guess. For example there is an outage right now and these 3 components are in not in a good state.

```
Registry [DEGRADED PERFORMANCE] 
API [DEGRADED PERFORMANCE] 
Build System [MAJOR OUTAGE] 
```

The quay.io outage from November 8th reported all of these components as major outage and I think the overall was set as major or critical and our jobs were affected as well I believe.

      Registry                        72xlygjldsvt    N/A          MAJOR OUTAGE     
        API                             29ttvthc3tc7    N/A          MAJOR OUTAGE     
        Build System                    dfmblnbcpjck    N/A          MAJOR OUTAGE     
        Frontend                        z0y8hl99w6ys    N/A          MAJOR OUTAGE     
        Security Scanning               gblcfv90cz4v    N/A          MAJOR OUTAGE   

So maybe in that case of quay.io we should check that none of these components are NOT in a major outage status, if any of these 5 are, fail out?

rhopp: +1 for not failing immediatelly, but it would be nice to somehow notify PR owner of service degradation immediately (to make him aware some dependncies are not healthy and there is possibility for this to affect tests). Maybe send comment to the PR?

psturc: ok failing the job in case of `major outage` status presence in critical components makes sense to me. 

I agree with failing the job in case any of quay components from `Registry`, `API`, `Build System` reports `major outage` (I'm not sure about `Frontend` and `Security Scanning`). and same for red hat components `registry.redhat.io` and `registry.access.redhat.com`

regarding [github](https://www.githubstatus.com/) - might be worth to consider failing the job in case of `major outage` status for following components:
- git operations
- api requests
- webhooks
- pull requests

psturc: @rhopp  regarding notifying PR owner - I think it's a good idea. but since the logic for analyzing the criticality of the services' outage won't be trivial and sending a PR comment using shell script is not ideal, I think it would make sense to implement it as [another command for qe-tools repo](https://github.com/redhat-appstudio/qe-tools) (we're already using qe-tools for the [report post step](https://github.com/openshift/release/blob/master/ci-operator/step-registry/redhat-appstudio/report/redhat-appstudio-report-commands.sh)), wdyt?

flacatus: totally agree. If we post artifacts in the post step then we can analyze how much external services are impacting us

PR #45653, opened 2023-11-14
aws-load-balancer: update ocp version

alebedev87:
Update the versions of the OpenShift clusters on which the tests are run.

### DIFF
--- a/ci-operator/config/openshift/aws-load-balancer-controller/openshift-aws-load-balancer-controller-main.yaml
+++ b/ci-operator/config/openshift/aws-load-balancer-controller/openshift-aws-load-balancer-controller-main.yaml
@@ -26,12 +26,12 @@ promotion:
 releases:
   initial:
     integration:
-      name: "4.12"
+      name: "4.15"
       namespace: ocp
   latest:
     integration:
       include_built_images: true
-      name: "4.12"
+      name: "4.15"
       namespace: ocp
 resources:
   '*':
--- a/ci-operator/config/openshift/aws-load-balancer-operator/openshift-aws-load-balancer-operator-release-1.0.yaml
+++ b/ci-operator/config/openshift/aws-load-balancer-operator/openshift-aws-load-balancer-operator-release-1.0.yaml
@@ -1,6 +1,6 @@
 base_images:
   base:
-    name: "4.12"
+    name: "4.13"
     namespace: ocp
     tag: base
   operator-index:
@@ -24,12 +24,12 @@ operator:
 releases:
   initial:
     integration:
-      name: "4.12"
+      name: "4.13"
       namespace: ocp
   latest:
     integration:
       include_built_images: true
-      name: "4.12"
+      name: "4.13"
       namespace: ocp
 resources:
   '*':
@@ -79,6 +79,7 @@ tests:
       OO_INSTALL_NAMESPACE: aws-load-balancer-operator
       OO_PACKAGE: aws-load-balancer-operator
       OO_TARGET_NAMESPACES: '!all'
+      OPENSHIFT_VERSION: "4.13"
     test:
     - as: test
       cli: latest
--- a/ci-operator/config/openshift/aws-load-balancer-controller/openshift-aws-load-balancer-controller-openshift-rel-1.0.yaml
+++ b/ci-operator/config/openshift/aws-load-balancer-controller/openshift-aws-load-balancer-controller-openshift-rel-1.0.yaml
@@ -26,12 +26,12 @@ promotion:
 releases:
   initial:
     integration:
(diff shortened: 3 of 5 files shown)


alebedev87:
/pj-rehearse pull-ci-openshift-aws-load-balancer-operator-main-e2e-aws-rosa-operator

alebedev87:
https://github.com/openshift/aws-load-balancer-operator/pull/120: to fix `release-1.0` jobs.

alebedev87:
/pj-rehearse pull-ci-openshift-aws-load-balancer-operator-release-1.0-e2e-aws-rosa-operator

ci-operator/config/openshift/aws-load-balancer-controller/openshift-aws-load-balancer-controller-main.yaml
```diff
@@ -25,12 +25,12 @@ promotion:
 releases:
   initial:
     integration:
-      name: "4.12"
+      name: "4.14"
```
gcs278, line 28: Should this be 4.15 now? Just like https://github.com/openshift/release/pull/45662

alebedev87:
/pj-rehearse pull-ci-openshift-aws-load-balancer-operator-main-e2e-aws-operator

ci-operator/config/openshift/aws-load-balancer-operator/openshift-aws-load-balancer-operator-release-1.0.yaml
```diff
@@ -79,6 +79,7 @@ tests:
       OO_INSTALL_NAMESPACE: aws-load-balancer-operator
       OO_PACKAGE: aws-load-balancer-operator
       OO_TARGET_NAMESPACES: '!all'
+      OPENSHIFT_VERSION: "4.13"
```
gcs278, line 82: This seems odd that you have `OPENSHIFT_VERSION` only on aws-load-balancer-operator-release-1.0.yaml. Is it needed? Why not the other files?

alebedev87: Good question, I think I ignored the other branches since they are for the TechPreview releases.

alebedev87: I figured out. The TechPreview branche cannot have this because they don't have the rosa test back ported. And the `1.1` release was using 4.14 at the time the PR was created. Let me update `1.1` branch too.

alebedev87:
/pj-rehearse

Re-rehearsing as new branches were impacted.

ci-operator/config/openshift/aws-load-balancer-operator/openshift-aws-load-balancer-operator-release-1.0.yaml
gcs278: Yea that makes sense.

I suppose the point of this is to "pin" to a specific version of OpenShift in ROSA; without this, it looks like https://steps.ci.openshift.org/reference/rosa-cluster-provision installs the latest version of OCP with ROSA, which I believe we don't want.

Shouldn't we add this to `ci-operator/config/openshift/aws-load-balancer-operator/openshift-aws-load-balancer-operator-main.yaml` too?

alebedev87: > I suppose the point of this is to "pin" to a specific version of OpenShift in ROSA; without this, it looks like https://steps.ci.openshift.org/reference/rosa-cluster-provision installs the latest version of OCP with ROSA, which I believe we don't want.

Right

> Shouldn't we add this to ci-operator/config/openshift/aws-load-balancer-operator/openshift-aws-load-balancer-operator-main.yaml too?

I think it's fine to keep the main branch as development one which runs on the latest OCP candidate available. This way we are sure we start new features on the latest OCP available on ROSA.

gcs278: Oh yea, that's a fair point. I think the tricky part is remembering to keep the OCP versioning up-to-date with our releases (hence the reason this PR was created).

alebedev87: Yes. Similar problem we have in the operator repository with [the `latest` doc links](https://github.com/openshift/aws-load-balancer-operator/blob/main/bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml#L111).

PR #45710, opened 2023-11-15
Bump coreos-assembler to Fedora 39

jlebon:
Matches https://github.com/coreos/coreos-assembler/pull/3653.

While we're here, use the Quay.io image instead since that's the future.

### DIFF
--- a/ci-operator/config/coreos/coreos-assembler/coreos-coreos-assembler-main.yaml
+++ b/ci-operator/config/coreos/coreos-assembler/coreos-coreos-assembler-main.yaml
@@ -1,7 +1,7 @@
 build_root:
   project_image:
     dockerfile_literal: |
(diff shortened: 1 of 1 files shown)


ci-operator/config/coreos/coreos-assembler/coreos-coreos-assembler-main.yaml
```diff
@@ -1,7 +1,7 @@
 build_root:
   project_image:
     dockerfile_literal: |
-      FROM registry.fedoraproject.org/fedora:38
+      FROM registry.fedoraproject.org/fedora:39
```
dustymabe, line 4: should we update this to `quay.io/fedora/fedora:39` ?

jlebon:
/pj-rehearse ack

We'll test this in https://github.com/coreos/coreos-assembler/pull/3653.

PR #45739, opened 2023-11-16
Restrict ODH cluster profile usage

riprasad:


riprasad:
Wouldn't it be nice if we could declare the profile as an array, thus eliminating the need to add multiple configurations for each cluster profile?

core-services/cluster-profiles/_config.yaml
```diff
@@ -1,6 +1,8 @@
-- profile: example-profile-name
+- profile: aws-opendatahub
```
riprasad, line 1: Wouldn't it be nice if we could declare the profile as an array, thus eliminating the need to add multiple configurations for each cluster profile?

PR #45748, opened 2023-11-16
automate OCP-24358 and some fix for OCP-66839

jianlinliu:
1. automate OCP-24358
2. resolve ccoctl extract from arm payload does not work issue - [OCPQE-17836](https://issues.redhat.com//browse/OCPQE-17836)
3. change preupgrade and postupgrade steps to use verification-test image to use `jq` binary
4. include `OCP-` prefix to unify OTA cases list
5. records all the enabled in `ota-test-result.txt` to easily identify what are tested whatever passed or failed to avoid case missing because of some auto script bug.
6. move `extract_oc` inside the required test case function.
7. Introduce `check_ota_case_enabled` to easily extend OTA test cases coverage


jianlinliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-arm64-nightly-4.15-upgrade-from-stable-4.14-aws-ipi-disconnected-sts-basecap-none-f28

jianlinliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.13-amd64-stable-4.13-upgrade-from-stable-4.12-gcp-ipi-proxy-private-f14

jianlinliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.13-amd64-stable-4.13-upgrade-from-stable-4.12-gcp-ipi-proxy-private-f14

jianlinliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.13-amd64-stable-4.13-upgrade-from-stable-4.12-gcp-ipi-proxy-private-f14

jianlinliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-arm64-nightly-4.15-upgrade-from-stable-4.14-aws-ipi-disconnected-sts-basecap-none-f28

jianlinliu:
ccoctl bin extracted from arm payload does not work issue is resolved, refer the [log](https://qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/qe-private-deck/pr-logs/pull/openshift_release/45748/rehearse-45748-periodic-ci-openshift-openshift-tests-private-release-4.15-arm64-nightly-4.15-upgrade-from-stable-4.14-aws-ipi-disconnected-sts-basecap-none-f28/1725319870226108416).

jianlinliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.13-amd64-stable-4.13-upgrade-from-stable-4.12-gcp-ipi-proxy-private-f14

jianlinliu:
@jiajliu I think this is ready for your review.

ci-operator/step-registry/cucushift/ota/preupgrade/cucushift-ota-preupgrade-commands.sh
```diff
@@ -96,24 +96,35 @@ function pre-ocp-66839(){
...
+    if [[ ! -s "${pre_proxy_spec}" ]]; then
+        echo "Fail to get proxy spec!"
+        return 1
+    fi
+    return 0
 }
 
-# Run single case through case ID
-function run_ota_single_case(){
-    if ! type pre-ocp-"${1}" &>/dev/null; then
```
jiajliu, line 108: suggest to keep this judgement before run the case to improve robustness

ci-operator/step-registry/cucushift/ota/preupgrade/cucushift-ota-preupgrade-commands.sh
```diff
@@ -124,13 +135,19 @@ if [ -f "${SHARED_DIR}/proxy-conf.sh" ] ; then
...
-elif [[ "${ENABLE_OTA_TEST}" == "true" ]]; then
-  extract_oc
-  run_ota_multi_test
-else
-  extract_oc
-  run_ota_single_case ${ENABLE_OTA_TEST}
-fi
-
+oc_extracted="false"
+for c in "OCP-66839" "OCP-24358"; do
```
jiajliu, line 139: hmm, do we hit any issue with original way by run_ota_single_case and run_ota_multi_test way?
case list in the for loop will be longer and longer.  Especially in run_ota_single_case scenario, the case num should be only 1 in ${ENABLE_OTA_TEST}, but it still travel over all cases. so i think original run_ota_single_case and run_ota_multi_test may distinguish it better.

jianlinliu: For `run_ota_single_case` scenario, when users enabled a case via ENABLE_OTA_TEST, `run_ota_single_case` is still going to run, then some dirty log are saved.

For `run_ota_multi_test` case, though we did not enable it yet, but I am thinking there would be some consequence issue between multiple cases are enabled. 

The new implementation would allow user to make cases in correct order.

For the introduced loop, it can work with whatever single cases or multiple cases.

jiajliu: > run_ota_single_case is still going to run it, then some dirty log are saved.

That's why there is following judgement to check if any func enabled actually. We could improve the log info here to "no case id func found or no need pre step for the case" sth like this.

```
if ! type pre-ocp-"${1}" &>/dev/null; then
        echo "Test Failed: pre-ocp-${1} due to no case id found!" >> "${report_file}"
```

or more better way is to leave a donothing func for such scenario?
```
function pre-OCP-24358(){
  echo "No checkpoints for the case in pre-upgrade step"
}
```

> For the introduced loop, it can work with whatever single cases or multiple cases.
It means we need maintain an extra case list in the code. And it will add unnecessary loop when the list become longer in single case more.

jianlinliu: Even we keep the original logic, users enabled a case in `ENABLE_OTA_TEST`, the user still need to write a function for the case testing in this step.

For the introduced loop, besides users need to write a function for it, users indeed also need to do one more step,  add the case id in the loop, this extra step is its disadvantage, but on the other side, this will our multiple cases in the future. 

If you prefer the original logic, I am going to revert the change here.

jiajliu: > the user still need to write a function for the case testing in this step.

Ah, user need one more step(but not a required step from my side) ,and reader/watcher is more clear to distinguish `no pre/post checkpoints for the case` or `miss checkpoint for some reason` here.

> this will our multiple cases in the future.

Per original discussion, run_ota_multi_test will not need pass trough case id in `ENABLE_OTA_TEST`. It only need run case func in serial and share common vars in run_ota_multi_test.

> I am going to revert the change here.

thanks, i'm not sure if original way will bring any issue in the future, we could improve it at that time.

ci-operator/step-registry/cucushift/ota/preupgrade/cucushift-ota-preupgrade-commands.sh
```diff
@@ -105,17 +116,26 @@ function run_ota_multi_test(){
...
+    fi
 }
+
+if [[ "${ENABLE_OTA_TEST}" == "false" ]]; then
+  exit 0
+fi
+
 report_file="${ARTIFACT_DIR}/ota-test-result.txt"
 export PATH=/tmp:${PATH}
+extract_oc
```
jianlinliu, line 138: @jiajliu generally we do not need extract oc in pre action, do your case - 66839 need the latest oc version?

ci-operator/step-registry/cucushift/ota/preupgrade/cucushift-ota-preupgrade-commands.sh
```diff
@@ -124,13 +144,8 @@ if [ -f "${SHARED_DIR}/proxy-conf.sh" ] ; then
...
 
 set +e
-if [[ "${ENABLE_OTA_TEST}" == "false" ]]; then
-  exit 0
-elif [[ "${ENABLE_OTA_TEST}" == "true" ]]; then
-  extract_oc
+if [[ "${ENABLE_OTA_TEST}" == "true" ]]; then
   run_ota_multi_test
 else
-  extract_oc
```
jianlinliu, line 133: @jiajliu generally we do not need extract oc in pre action, do your case - 66839 need the latest oc version?

jianlinliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-arm64-nightly-4.15-upgrade-from-stable-4.14-aws-ipi-disconnected-sts-basecap-none-f28

jianlinliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.13-amd64-stable-4.13-upgrade-from-stable-4.12-gcp-ipi-proxy-private-f14

ci-operator/step-registry/cucushift/ota/preupgrade/cucushift-ota-preupgrade-commands.sh
jiajliu: yes, the case need latest oc for `--included` option

jianlinliu: @jiajliu if that will move the extract_oc function to OCP-66839 function, is that okay for you?

jiajliu: yes, sure. 
btw, what's the expected oc version for other cases in preupgrade step?

jianlinliu: In my case - OCP-24358, it just get some proxy spec from cluster using oc command, it is a very basic operations, so the built oc command in the verification-test image should be enough. I am not sure the exact version in that image.

jianlinliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-arm64-nightly-4.15-upgrade-from-stable-4.14-aws-ipi-disconnected-sts-basecap-none-f28

jianlinliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.13-amd64-stable-4.13-upgrade-from-stable-4.12-gcp-ipi-proxy-private-f14

jianlinliu:
cc @liangxia to review

PR #45804, opened 2023-11-17
MULTIARCH-4025 Removing the 4.14 optional jobs from the P/Z release controllers

tvardema:


core-services/release-controller/_releases/priv/release-ocp-4.14-ppc64le.json
```diff
@@ -8,13 +8,5 @@
     "pullSecretName": "source",
     "referenceMode": "source",
     "to": "release-ppc64le-priv",
-    "verify": {
```
Prashanth684, line 11: do we even need this file if we are removing the only test?

tvardema: No, we need the file because it dictates what is visible on the release controller.

bradmwilliams (approved):

PR #45811, opened 2023-11-17
fix: OADP 1.3 mirror images

mateusoliveira43:
Before adding 1.3 operator image mirroring, it has to run at least once first. In other words, CI for that branch needs to added first.


mateusoliveira43:
we merged 1.3 CI before this, no need anymore for it

PR #45821, opened 2023-11-19
SPLAT-1241: WIP|aws/local-zones/profile: change the profile to default aws account

mtulio:
Tasks:
- [ ] [Blocker] Create a new profile including only regions that have Local or Wavelength zones, otherwise the jobs will fail when discovering the edge zones
- [ ] Validate if the Bug can be closed, removing the test skips in Local Zones: https://issues.redhat.com/browse/OCPBUGS-11609


mtulio:
/pj-rehearse pull-ci-openshift-installer-master-e2e-aws-ovn-shared-vpc-localzones

PR #45853, opened 2023-11-20
openshift: ptp-operator: enable security scanning

danimjaffe:
per https://docs.ci.openshift.org/docs/how-tos/add-security-scanning/


danimjaffe:
In order to ignore vulnerabilities you can follow the guidelines in the [following link](https://docs.ci.openshift.org/docs/how-tos/add-security-scanning/#how-to-ignore-vulnerabilities).
For reference you can see an example in the [following repo](https://github.com/openshift-kni/numaresources-operator/blob/main/.snyk).

aneeshkp (approved):
/lgtm

PR #45883, opened 2023-11-20
OSSM update boskos config

yxun:
This PR is adding another ossm profile `ossm-aws` boskos config. 
Regarding to the previous wrong lease name,
the previous profile is not working before and after updating a lease name.

ref: https://redhat-internal.slack.com/archives/CBN38N3MW/p1700499649811039


yxun:
this PR is related to https://github.com/openshift/ci-tools/pull/3794

PR #45886, opened 2023-11-20
Fix AWS EBS operator dockerfile

dobsonj:
It has been erroneously changed by auto-config-brancher job, changing it
back.

/cc @openshift/storage


dobsonj:
Fixed by the latest run: https://github.com/openshift/release/pull/45907

PR #45894, opened 2023-11-21
Add ipsec ns workflow

huiran0826:
Currently in https://github.com/openshift/release/blob/master/ci-operator/step-registry/ovn/conf/ipsec-manifest/ovn-conf-ipsec-manifest-commands.sh, it enabled both IPSEC E-W and N-S since 4.14.   But we also need to check upgrade  path `4.14 IPSEC N-S only -> 4.15`. 

Added a separate step for IPSEC N-S only and upgrade job. 

@anuragthehatter  PTAL, thanks!


huiran0826:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-nightly-4.15-upgrade-from-stable-4.14-gcp-ipi-ovn-ipsec-ns-f14

PR #45929, opened 2023-11-21
Fix test filter on baremetal-pxe-sno-agent-ipv4-static-connected-f14

kasturinarra:
Fix test filter on  baremetal-pxe-sno-agent-ipv4-static-connected-f14


kasturinarra:
@liangxia could you please help review ? thanks !!

PR #45999, opened 2023-11-23
OSSM-5422 run integration operator test on multiple ocp versions

fjglira:
According to the release matrix we want to run the integration operator test against:

1. Latest OCP version
2. Next OCP version (From now is 4.15)


ci-operator/config/maistra/istio-operator/maistra-istio-operator-maistra-3.0.yaml
```diff
@@ -153,13 +153,61 @@ tests:
           memory: 100Mi
       timeout: 2h0m0s
     workflow: servicemesh-istio-e2e-hypershift
-- as: integration-ocp
+- as: integration-ocp-latest-release
```
jwendell, line 156: Do you think it's better to name these jobs after the version?

e.g. `integration-ocp-4.14` and `integration-ocp-4.15`?

Benefits is that it's very straightforward to know which version if running just by looking at the GitHub PR. Plus, if we ever need to test on another version, using numbers is easier than `latest`, `next`, etc.

ci-operator/config/maistra/istio-operator/maistra-istio-operator-maistra-3.0.yaml
```diff
@@ -153,13 +153,61 @@ tests:
...
+      resources:
+        requests:
+          cpu: 100m
+          memory: 100Mi
+      timeout: 10m0s
+    - as: integration-ocp-latest-release
+      cli: latest
+      commands: |
+        oc rsh -n "${MAISTRA_NAMESPACE}" "${MAISTRA_SC_POD}" \
+          sh -c "\
```
jwendell, line 186: can you put an `oc version && \` before these jobs? That's just to make sure, when looking at the logs, if the version is the expected one.

fjglira:
This will fail until this is merged: https://github.com/openshift/release/pull/46379

jwendell (approved):

PR #46049, opened 2023-11-27
Increase the timeout value to 35m of test execution for hypershift-mgmt test profiles

wangke19:
Increase the timeout value to 35m of test execution for hypershift-mgmt test profiles, we have one case OCP-62093 always ran into failure `Interrupted by User`, because the case ran long time more than 30m, the default execution time is exceeded by some hypershift-mgmt test profiles with 30m test timeout setting.


heliubj18:
cc @LiangquanLi930

PR #46081, opened 2023-11-27
cluster-etcd-operator mark important jobs as non-optional

tjungblu:
This is to avoid running into the case where optionals are not even run by prow. We're marking important jobs as non-optional = required.


Elbehery:
/lgtm
/approve

Nice work 💪💪

PR #46089, opened 2023-11-27
TRT-1368: Move 4.15 gcp-ovn-rt-upgrade to new gcp account

dgoodwin:
Nightly Payloads failing due to a credentials related quota issue.
Trying to move to new account to get our payloads back asap while it's
worked out.

### DIFF
--- a/ci-operator/config/openshift/release/openshift-release-master__ci-4.15-upgrade-from-stable-4.14.yaml
+++ b/ci-operator/config/openshift/release/openshift-release-master__ci-4.15-upgrade-from-stable-4.14.yaml
@@ -105,7 +105,7 @@ tests:
   cluster: build02
   interval: 168h
   steps:
-    cluster_profile: gcp-openshift-gce-devel-ci-2
+    cluster_profile: gcp-3
     env:
       TEST_TYPE: upgrade-conformance
     observers:
--- a/ci-operator/jobs/openshift/release/openshift-release-master-periodics.yaml
+++ b/ci-operator/jobs/openshift/release/openshift-release-master-periodics.yaml
@@ -23605,7 +23605,7 @@ periodics:
   interval: 168h
   labels:
     ci-operator.openshift.io/cloud: gcp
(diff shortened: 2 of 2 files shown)


jstuever:
/hold
Let's not do this just yet. If we move to gcp-3 now and continue creating custom roles, then we will make that project uninhabitable as well. I recommend we wait until https://github.com/openshift/cloud-credential-operator/pull/631 lands in 4.14:latest. This will prevent a repeat of the 4.14-4.15 upgrade issue.

brenton:
Hi @jstuever, what's the timing on landing https://github.com/openshift/cloud-credential-operator/pull/631 in 4.14?  It's the last week of the sprint so we have until Wednesday to select a sprint payload.  History suggests if we don't have one by Tuesday we aren't going to get one without someone being a hero (and we'd like to avoid that :smile: ).

jstuever:
I should be more clear. TRT-1368 was triggered by a PR in openshift-image-registry-operator. [Reverting](https://github.com/openshift/cluster-image-registry-operator/pull/965) that PR should resolve the incident; the periodics will begin passing again in all projects. As a result, this PR is not urgent in that regard.

https://github.com/openshift/cloud-credential-operator/pull/631 is a long-term solution. The issue was that v4.14 CCO does the wrong thing when a operator presents v4.15 granular permissions, as happened in the cluster-image-registry-operator PR. We fixed 4.15 to do the right thing, and PR 631 is backporting that to 4.14. Once it lands, the v4.14 CCO will start doing the right thing, and not max out our quota. This will resolve the issue of 4.14-4.15 upgrades, and enable us to start merging the granular-permission PRs in all of the operators. At that point, we can move off of gcp-2 for 7-14 days to enable this work to continue.

jstuever:
I am hoping to land https://github.com/openshift/cloud-credential-operator/pull/631 today.

dgoodwin:
Thanks for the help sorting this out @jstuever and good call on the revert.

/close

jstuever:
/close
This PR is probably not enough to unblock granular permission work. I'll have to regroup.

PR #46093, opened 2023-11-27
USHIFT-1947: Narrow down regex for metal-periodic-test

pacevedom:
/cc @dhellmann @ggiguash

### DIFF
--- a/ci-operator/config/openshift/microshift/openshift-microshift-main.yaml
+++ b/ci-operator/config/openshift/microshift/openshift-microshift-main.yaml
@@ -123,7 +123,7 @@ tests:
     workflow: openshift-microshift-rebase
 - as: metal-periodic-test
   cluster: build02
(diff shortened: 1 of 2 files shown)


dhellmann:
/lgtm

We should consider which other branches might also need this update and we want to make sure it ends up on the 4.15 branch when that diverges from main.

pacevedom:
> /lgtm
> 
> We should consider which other branches might also need this update and we want to make sure it ends up on the 4.15 branch when that diverges from main.

4.14 has almost the same [regex](https://github.com/openshift/release/blob/master/ci-operator/config/openshift/microshift/openshift-microshift-release-4.14.yaml#L100) and there is only layer3 blueprints and one file from etcd compactor that would match the regex, besides the periodics.

As for 4.15 branch, this is guaranteed to be there if it merges before branch cut. It only takes a couple of hours to sync changes.

PR #46103, opened 2023-11-28
[4.12] update ci jobs frequency

jianlinliu:
As new versions get GAed, lower the 4.12 ci jobs frequency.


jianlinliu:
cc @liangxia to review.

PR #46115, opened 2023-11-28
local-storage-operator: remove unnecessary images

jsafrane:
base and ocp_builder_rhel is already defined in .ci-operator.yaml in the repository.

It should fix CI builds with RHEL9 [here](https://github.com/openshift/local-storage-operator/pull/446) - `ocp_builder_rhel-8-golang-1.20-openshift-4.15` defined in this CI job will not be used and RHEL9 from `.ci-operator.yaml` will be used.

### DIFF
--- a/ci-operator/config/openshift/local-storage-operator/openshift-local-storage-operator-master.yaml
+++ b/ci-operator/config/openshift/local-storage-operator/openshift-local-storage-operator-master.yaml
@@ -1,16 +1,8 @@
 base_images:
-  base:
-    name: "4.15"
-    namespace: ocp
-    tag: base
   ocp_4.14_cli:
     name: "4.15"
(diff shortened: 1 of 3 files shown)


jsafrane:
/pj-rehearse ack
Image builds just fine.

ci-operator/config/openshift/local-storage-operator/openshift-local-storage-operator-master.yaml
```diff
@@ -20,17 +16,9 @@ build_root:
 images:
 - dockerfile_path: Dockerfile.rhel7
   from: base
```
bertinatto, line 18: This line overwrites the Dockerfile's `FROM` directive with the `base` image. Please remove it as well

PR #46136, opened 2023-11-28
Allow m1kola to rerun jobs

m1kola:


core-services/prow/02_config/_config.yaml
```diff
@@ -23,6 +23,7 @@ deck:
       - bparees
       - AlexNPavel
       - jupierce
+      - m1kola
```
m1kola, line 26: It is a periodic job that I want to be able to rerun. It is defined [here](https://github.com/openshift/release/blob/b98c04fddbe6f0c68d4bdbd49d3eb6300db344c6/ci-operator/jobs/infra-periodics.yaml#L979-L1038).

Do I need to add `repo: openshift/release` into the list? Or what should it be?

m1kola: Got reply in Slack from:
> it should be enough adding a new entry for the sole openshift/operator-framework-operator-controller repository as it's the first one on [this list](https://github.com/openshift/release/blob/b98c04fddbe6f0c68d4bdbd49d3eb6300db344c6/ci-operator/jobs/infra-periodics.yaml#L984-L986)

PR #46143, opened 2023-11-28
cluster-olm-operator: rely on images from repo

ncdc:
Remove the base images and image overrides in the config here, so the images defined in the cluster-olm-operator repo's .ci-operator.yaml and Dockerfile are used. Having the images in here is causing problems because we can get a mismatch between RHEL versions, causing dynamic shared library symbol issues at runtime. The configurations in the repo become the source of truth, and they'll be automatically updated by ART bots as needed.

Needed for https://github.com/openshift/cluster-olm-operator/pull/34


stevekuznetsov (approved):
/lgtm
/approve

PR #46273, opened 2023-11-30
Remove sdn-specific o/k tests in 4.15+

danwinship:
openshift-sdn is deprecated, and will no longer be (easily) installable in 4.15.

o/k has some optional tests that use sdn instead of ovn-k, but there's not really much need for them; nobody is actually running those optional tests anyway (https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-openshift-kubernetes-master-e2e-aws-multitenant), and if some upstream change does somehow break openshift-sdn-but-not-ovn-kubernetes it will get caught by later tests anyway.

So this removes `e2e-aws-multitenant` and switches `e2e-metal-ipi` over to ovn-k, for 4.15 and later.

### DIFF
--- a/ci-operator/config/openshift/kubernetes/openshift-kubernetes-master.yaml
+++ b/ci-operator/config/openshift/kubernetes/openshift-kubernetes-master.yaml
@@ -374,12 +374,7 @@ tests:
         requests:
           cpu: 100m
     workflow: ipi-aws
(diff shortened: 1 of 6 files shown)


danwinship:
I'm going to let component owners fix their own tests; for now I'm just making sure that existing sdn-based tests keep working.

PR #46305, opened 2023-12-01
move csi driver to vsphere02 cluster

jcpowermac:
These jobs do not have a cluster profile
and were not migrated to the new vsphere build cluster

Manually setting the cluster until this can
be fixed in ci-tools.


jcpowermac:
https://github.com/openshift/ci-tools/pull/3819

jcpowermac:
/close
not needed

PR #46319, opened 2023-12-01
Explicitly specify NETWORK_TYPE in cucushift workflow rather than defaulting it

danwinship:
openshift-sdn is being deprecated. As a side effect of this, it's possible that dev-scripts may change its defaults. Once #46286 merges, there will only be one workflow left in openshift/release that sets `IP_STACK=v4` and then allows `NETWORK_TYPE` to just be defaulted by dev-scripts. This fixes it to set it explicitly.

I'm not really sure what this workflow is used for, but from grepping around, I'm pretty sure that nothing is using this workflow but overriding `NETWORK_TYPE=OVNKubernetes`...

### DIFF
--- a/ci-operator/step-registry/cucushift/installer/rehearse/baremetalds/assisted/cucushift-installer-rehearse-baremetalds-assisted-workflow.yaml
+++ b/ci-operator/step-registry/cucushift/installer/rehearse/baremetalds/assisted/cucushift-installer-rehearse-baremetalds-assisted-workflow.yaml
@@ -10,6 +10,7 @@ workflow:
     env:
       DEVSCRIPTS_CONFIG: |
         IP_STACK=v4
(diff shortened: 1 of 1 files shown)


LiangquanLi930:
@danwinship Thanks, AFAIK, NETWORK_TYPE's default is OVNKubernetes? For this workflow we hope NETWORK_TYPE keeps set OVN

danwinship:
`NETWORK_TYPE` defaults to `OVNKubernetes` for IPv6, but `OpenShiftSDN` for IPv4. So since this is setting `IP_STACK=v4`, it would currently be getting `OpenShiftSDN`.

However, I can't figure out where this actually gets used anyway... nothing else seems to reference this workflow, and [the bot seems to agree](https://github.com/openshift/release/pull/46319#issuecomment-1836317470)?

PR #46338, opened 2023-12-02
Switch virtualmedia workflows to use ovn-kubernetes

danwinship:
Splitting this out of #46286 because this part ends up requiring changes to `core-services` too...

This is part of a large set of changes around deprecating openshift-sdn. In this particular case, the `baremetalds-e2e-virtualmedia`, `baremetalds-e2e-serial-virtualmedia`,
and `baremetalds-e2e-sdn-serial-virtualmedia-bond` workflows all currently use openshift-sdn, but this seems like just a historical artifact, and in particular, there are no ovn equivalents of any of those workflows, so it really looks like everyone is just using them because they want to test virtualmedia/bonds, not because they want to test virtualmedia/bonds _on sdn_ specifically. So while there are some test jobs where there's an argument to made that we should keep testing sdn rather than ovn, that doesn't seem to apply to these workflows at all, so this PR switches them all over to ovn, including renaming the jobs that previously had "-sdn-" in their name (which then unfortunately causes the jobs files to get re-sorted making the diffs look broken).

As mentioned above, this ends up requiring changes to `core-services/release-controller/`, and I'm much less sure about how things work there...

/assign @stbenjam @derekhiggins


danwinship:
/assign @andfasano 
looks like you originally added these workflows...

ci-operator/jobs/openshift/installer/openshift-installer-release-4.16-presubmits.yaml
```diff
@@ -6857,7 +6857,7 @@ presubmits:
           secretName: result-aggregator
     trigger: (?m)^/test( | .* )e2e-vsphere-ovn,?($|\s.*)
   - agent: kubernetes
-    always_run: false
+    always_run: true
```
zaneb, line 6860: This looks unintentional.

danwinship: ugh, `make jobs` is so mysterious. fixed I think? (I had to rebase to current master first though to avoid undoing unrelated changes elsewhere.)

derekhiggins:
thanks,
/approve

droslean (approved):

PR #46344, opened 2023-12-03
Rename e2e-metal-assisted jobs that mistakenly claimed to be sdn-based

danwinship:
cluster-baremetal-operator has a job called `e2e-metal-sdn-assisted` that actually uses ovn-kubernetes. Presumably the name used to be correct but then the `baremetal-assisted` workflow changed out from under it...


danwinship:
(rebase for jobs conflicts)

PR #46368, opened 2023-12-04
cert-rotation: SNO jobs also need to be prepared for shutdown

vrutkovs:
Fixes Multus cert expiring on SNO cert rotation jobs, similar to HA


dinhxuanvu (approved):
/lgtm
/approve

PR #46380, opened 2023-12-04
Allow kube-burner version override

jtaleric:
In this change introduce `KUBE_BURNER_VERSION` env var, which will allow users to override the kube-burner version.


jtaleric:
> lgtm @jtaleric do you plan to add that variable to the tests in this same PR or a different one or later for future?

Not in this PR... but I do have one I am cooking up that will override this.

PR #46394, opened 2023-12-05
update oc-mirror cron set for 4.15

zhouying7780:
Since we have big feature change for oc-mirror , we need to do the regression test more frequently . 

@kasturinarra @liangxia  Please help review , thanks .


ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__amd64-nightly.yaml
```diff
@@ -592,7 +592,7 @@ tests:
     - chain: openshift-e2e-test-qe-longduration
     workflow: cucushift-installer-rehearse-aws-ipi
 - as: aws-ipi-f28-longduration-workload
-  cron: 44 0 12 * *
+  cron: 44 0 * * sun,wed
```
liangxia, line 595: Generally, we update the `fN` in the test name, then use `generate-cron-entry.sh` to generate the cron entries.
We also have tools to batch update cron entries. The change here might get reverted if someone run the batch update tool carelessly.

PR #46400, opened 2023-12-05
Disable agent okd scos jobs

andfasano:
Right now the agent okd scos jobs are perma-broken, and until a SCOS boot image will be available (see [here](https://github.com/openshift/installer/pull/7112#issuecomment-1794300674)) it's better to save CI resources - since right now for every PR they are usually triggered.
They could be readded once the SCOS support will be available/planned.


andfasano:
cc @LorbusChris

vrutkovs:
Oh, I assumed 'optional: true' would override 'run_if_changed: true'.

/lgtm

PR #46443, opened 2023-12-06
harden nw migration scripts

jluhrsen:
- increases timeout to check that migration status is removed from 'oc get network' from 60s to 120s
  - local testing with 4.15 saw this take longer than 60s

- modified the final check for each migration script to determine if all clusteroperators are healthy
  - there are example jobs that show clusteroperators move back to degraded after the migration script finishes because it happened to notice all operators in the proper state, but the following script's logs (in the case of rollback) end up showing some operators are still progressing
  - they need to all be Available=True, Progressing=False, Degraded=False
  - this needs to be the case for 3 consecutive checks spaced 30s apart
  - it will be retried for up to 2700s (45m)


jluhrsen:
/pj-rehearse ack

this nw migration jobs are definitely not super healthy, but this should help a little.

PR #46481, opened 2023-12-06
add back slack notification

vboulos:
Add back the slack notification to the ocp 4.15 job that seems to be remove somehow


ascerra:
When creating a new job which I believe was the case when we bumped to version 4.15 the slack notification information does not get automatically created as part of the auto-generated prow job.

Adding the slack notification is a manual addition once the job has been autogenerated.

This is just due to the ci-operator's lack of support for the Prow slack plugin.

gparvin (approved):

PR #46484, opened 2023-12-06
Interop - Update Firewatch config to match v2 [3sacle]

oharan2:
Following changes made in #46282

### DIFF
--- a/ci-operator/config/3scale-qe/3scale-deploy/3scale-qe-3scale-deploy-main__3scale-amp-ocp4.14-lp-rosa.yaml
+++ b/ci-operator/config/3scale-qe/3scale-deploy/3scale-qe-3scale-deploy-main__3scale-amp-ocp4.14-lp-rosa.yaml
@@ -39,25 +39,28 @@ tests:
       DEPL_BUCKET_NAME: 3scale-apimanager-s3-bucket
       DEPL_PROJECT_NAME: threescale
       FIREWATCH_CONFIG: |
+        {
+        "failure_rules":
         [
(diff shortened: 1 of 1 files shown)


calebevans (approved):
/lgtm

ci-operator/config/3scale-qe/3scale-deploy/3scale-qe-3scale-deploy-main__3scale-amp-ocp4.14-lp-rosa.yaml
```diff
@@ -39,25 +39,28 @@ tests:
       DEPL_BUCKET_NAME: 3scale-apimanager-s3-bucket
```
ascerra, line 39: Placing comment here as its the first line I can comment on but this is really meant to be for line 33

I believe this PR is putting in place testing for Rosa Classic but I don't see anything that tells us that explicitly.

We use unique identifiers to help us separate our work

For self-managed we use "-lp-interop"
For Rosa HyperShift we us "-lp-rosa-hypershift"

For Rosa Classic what should we use?

I'd reccomend "-lp-rosa-classic"
What do you think?
This unique identifier can either be in the filename or here in the test stanza name

oharan2: Since I want to raise a question on this topic, I can make this update in a separate PR; 
we also have ROSA with camel-k and AMQ streams, so a dedicated update will be made anyway :)

PR #46515, opened 2023-12-07
OCPQE-18245: Schedule baremetal testing in US timezone

liangxia:
/cc @aleskandro @jadhaj @mhanss @jhou1

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/tools/generate-cron-entry.sh
+++ b/ci-operator/config/openshift/openshift-tests-private/tools/generate-cron-entry.sh
@@ -21,7 +21,7 @@ fi
 TEST_NAME="$1"    # aws-c2s-ipi-disconnected-private-f7
 YAML_FILE="$2"    # openshift-openshift-tests-private-release-4.13__amd64-nightly.yaml
 
-if [[ "${TEST_NAME}" =~ (baremetal-|-disabled|-disasterrecovery|powervs-) ]] && [[ "$@" != *\ --force* ]]; then
+if [[ "${TEST_NAME}" =~ (-disabled|-disasterrecovery|powervs-) ]] && [[ "$@" != *\ --force* ]]; then
     echo "The test config ${TEST_NAME} should not get changes in the cron entry as
       the schedule rotation scheme is different than the other tests.
       Use --force to skip this check."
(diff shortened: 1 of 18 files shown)


ci-operator/config/openshift/openshift-tests-private/tools/generate-cron-entry.sh
```diff
@@ -52,6 +52,15 @@ let DAY_OF_MONTH=10#${NUMBERS:4:2}%30+1
 let MONTH=10#${NUMBERS:6:2}%12+1
 let DAY_OF_WEEK=10#${NUMBERS:8:1}%7
 
+if [[ "${TEST_NAME}" =~ baremetal- ]] ; then
```
aleskandro, line 55: Hi @liangxia , can you explain a bit how the calculation is done?

Is the calculated time independent across yaml files (versions, variant)?

For baremetal, we'd need something considering all of them and spreading all the jobs along the week/month. The less jobs run concurrently, the better.

Not sure of how it works for the others, just trying to define the requirement better.

liangxia: Generally, test name (contains test frequency `fN`) and config file name (contains OCP version) are used to generate a unique cron entry for a specific profile. So we can assume that the same profile in different versions should get different cron entry. We've verified it works that the jobs are spread across the hour/day/month.
And specific for baremetal, we reevaluate the HOUR to let it fall into the US timezone. So compared with the general profiles, the difference is that baremetal jobs use a specific range of the hours, but day of the week/month still spread across the whole range as the other profiles.

aleskandro: ok, so you essentially use fN and filename as an hash/seed... but there is no guarantee that the spread is done "properly" for a hardware constrained scenario...

can you dry run the script to show how the corn entries would come out for the baremetal jobs?

aleskandro: sorry @liangxia i did miss the reply notification.

It seems looking good, we can try and monitor how it goes in the time.

Just an additional thing. I did set a maintenance window of 1 hour each Saturday at 23:00 UTC. It should perform updates in the persistent nodes and a reboot (coreos-based). Can we avoid scheduling jobs on Saturday? Considering long-running jobs, I think we can fail if a job is scheduled even at 12PM (-5h from maintenance window)

liangxia: @aleskandro Since we are rescheduling the baremteal tests to run between 13:00 to 22:00 in UTC, perhaps we can reschedule the maintenance window to a time before 13:00 UTC. E.g, at 11:00 UTC (China time 19:00, Italy time 12:00).  If we do the maintenance on Saturday, we have a much larger time window  to choose, varies from 0:00 to 12:00 in UTC, and you can choose one that works best for you.

aleskandro: Ok. Let's move forward and follow-up later if needed

Thanks Liang

PR #46548, opened 2023-12-07
Add installer-aro and rename ARO-Installer to installer-aro-wrapper

lranjbar:
The repository ARO-Installer was renamed to installer-aro-wrapper. In addition the repository installer-aro was added. This updates the CI config for these two repositories.

To clarify we are not looking to add these images into the OCP payload. We are just looking to test this in Openshift CI.


dhellmann (commented):
Approving based just on the image name and the goal of having this testable in CI. Please get people who are more familiar with the prow config to help double check any of the job settings and provide the lgtm.

/approve

PR #46580, opened 2023-12-08
Update openshift-docs CI to run build and build_for_portal scripts

aireilly:
Updates openshift-docs CI to run build and build_for_portal scripts in parallel.


gabriel-rh (approved):
LGTM

Could we increase resources to improve performance? does that have a cost impact?

PR #46592, opened 2023-12-08
osc make check optional

cpmeadors:
This will make the unit tests (make check) optional, but it does this by just not running them.  Not sure if we want that or if we want them to continue to fail and override until we fix them.

Fixes: [KATA-2672](https://issues.redhat.com//browse/KATA-2672)


littlejawa (approved):
lgtm
Thanks @cpmeadors !

PR #46594, opened 2023-12-08
stackrox: fix the run regex for shell-unit

gavin-stackrox:
The current regex does not cause the test to run when *.sh or *.bat are changed. This change should work based on what I see from other configs.
/cc @davdhacs @BradLugo


gavin-stackrox:
/honk
@BradLugo @davdhacs

BradLugo (approved):
/lgtm

PR #46606, opened 2023-12-10
Interop - Add '-classic' identifier for ROSA scenarios

oharan2:
As for this moment, all Interop scenarios with ROSA classic testings filenames are ending with 
`-lp-rosa`

For self-managed we use "-lp-interop"
For Rosa HyperShift we us "-lp-rosa-hypershift"

We use unique identifiers to help us separate our work, and we need a strict specification for ROSA classic testings (against the -hypershift identifier).

This PR submits the update of all merged jobs with adding the "-lp-rosa-classic" into current filenames.


calebevans (approved):
/lgtm

PR #46629, opened 2023-12-11
ShiftStack: Update sync periodics for 4.15

dulek:
This commit makes sure that periodic jobs syncing upstream CPO with our downstream fork will sync 1.28 into 4.15, now that branching occurred.


MaysaMacedo:
@dulek looks like you need to run `make jobs`, after the fix it lgtm

PR #46634, opened 2023-12-11
extdns: setup mirroring to quay.io for 1.0 and 0.1 releases

alebedev87:
Add mirroring config files for `1.0` and `0.1` releases and promote `0.1` images to `external-dns-operator` namespace (like other branches do).


alebedev87:
/approve

Am I an approver?

PR #46680, opened 2023-12-12
Remove securesign from ci operator

JasonPowr:
Hello :)
Looking for some guidance on this one, I'm part of the securesign org and we are looking to remove all of our testing from the ci-operator, however we would like to keep prow enabled so we can use labels like lgtm, etc... 

Is what I've done here correct? or is there other producers to follow ?


sallyom:
placing hold on this as I will pick this commit into a new PR to add a nightly periodic & operator run

cooktheryan:
@sallyom you rock!

PR #46683, opened 2023-12-12
update hypershift image mirroring for 4.16

sjenning:


sjenning:
/close
already done in https://github.com/openshift/release/pull/46682

PR #46693, opened 2023-12-12
Increase timeout in prow

svetsa-rh:
Increasing timeout to see if we can get some successful runs in Prow.


paigerube14:
@svetsa-rh Can you post a failing job that you are hoping to have this fix? I wonder if we need to increase the timeouts on sub tests as well

PR #46708, opened 2023-12-13
NO-ISSUE: Make agent okd compact job as required

andfasano:
Now that we have a working agent okd implementations (for FCOS), it could be useful to set at least the compact job as required to avoid any silent future regression (and to sync with the ocp configuration)

### DIFF
--- a/ci-operator/config/openshift/installer/openshift-installer-master__okd.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-master__okd.yaml
@@ -220,7 +220,6 @@ tests:
     workflow: openshift-e2e-aws
 - as: e2e-agent-compact-ipv4
   cluster: build05
-  optional: true
   run_if_changed: /agent/
   steps:
     cluster_profile: equinix-ocp-metal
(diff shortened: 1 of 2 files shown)


andfasano:
cc @vrutkovs @LorbusChris

ci-operator/config/openshift/installer/openshift-installer-master__okd.yaml
```diff
@@ -220,7 +220,6 @@ tests:
     workflow: openshift-e2e-aws
 - as: e2e-agent-compact-ipv4
   cluster: build05
-  optional: true
```
vrutkovs, line 223: I'd rather leave them optional, but lets make them always_run? Installer team may introduce a breaking change and should not depend on OKD bugs to land it.

Another option is to add agent-compact jobs to OKD nightlies testing to make sure we catch regressions in the entire payload

andfasano: Putting `always_run: true` anyhow will consume an ofcir instance for all the PRs, even those ones that are not related at all for the agent. It could have more sense to have it always executed in an informing (periodic) job maybe, and leave the presubmits as they are now (not blocking)

vrutkovs: Right, informing job sounds like a better idea

PR #46731, opened 2023-12-13
telco: Remove GCP jobs for PAO repository

sshnaidm:
We don't need more e2e GCP jobs there because of EOL releases and all code sitting in NTO repo now.


yanirq:
I think we can actually remove the files completely for versions lower than 4.10

PR #46736, opened 2023-12-13
Regenerate nutanix jobs

danilo-gemoli:
Follow up of https://github.com/openshift/ci-tools/pull/3840 as some users are experiencing problems when checking out their code.


jianlinliu:
@danilo-gemoli thanks for the  update.

PR #46739, opened 2023-12-13
add lnxocp03 to CI

telpelt:
This commit will add a new LPAR for the OpenShift CI on IBM Z and refers to #46320

### DIFF
--- a/core-services/prow/02_config/_boskos.yaml
+++ b/core-services/prow/02_config/_boskos.yaml
@@ -1939,16 +1939,12 @@ resources:
 - names:
   - libvirt-s390x-0-0
   - libvirt-s390x-0-1
-  - libvirt-s390x-0-2
-  - libvirt-s390x-0-3
   - libvirt-s390x-1-0
   - libvirt-s390x-1-1
-  - libvirt-s390x-1-2
-  - libvirt-s390x-1-3
   - libvirt-s390x-2-0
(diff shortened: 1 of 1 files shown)


core-services/prow/02_config/generate-boskos.py
```diff
@@ -324,8 +324,8 @@
     },
 }
 
-for i in range(3):
-    for j in range(4):
+for i in range(1,4):
```
tvardema, line 327: Changing the range to `1,4` would effectively stop provisioning to `0-0` through `0-4`, is that intended in this PR?

telpelt: Yes, this will reduce the amount of parallel clusters to 6. It's the same amount, which we will have in the intermediate period.

jaypoulz:
I think you're going to need to run `make prow-config`

core-services/prow/02_config/generate-boskos.py
tvardema: As discussed in Slack, I would prefer to leave all clusters engaged until we've introduced all the new clusters, but removing one set would be fine for now given we'll still support a minimally appropriate number of clusters for CI.

telpelt:
> I think you're going to need to run `make prow-config`

I ran the command locally, but it changes nothing on the repository files.

PR #46745, opened 2023-12-13
NO-JIRA: step-registry: capi: remove upstream operator check

damdo:
We removed the upstream cluster-api-operator (for which the container is confusingly called `capi-operator-controller-manager`) so remove the check in the `apply-feature-gate` step.


JoelSpeed (commented):
/lgtm
/approve

PR #46763, opened 2023-12-13
Adding version overrides ENVs for e2e and individual tools

vishnuchalla:
Pre-requisite PR: https://github.com/cloud-bulldozer/e2e-benchmarking/pull/661 to run rehearsal tests and verify the results.

This PR helps us to override E2E and individual tool versions to experiment/test with them. If any of the tools versions are not specified, default versions will be used from e2e-benchmarking repo.

### DIFF
--- a/ci-operator/step-registry/openshift-qe/cluster-density-v2/openshift-qe-cluster-density-v2-commands.sh
+++ b/ci-operator/step-registry/openshift-qe/cluster-density-v2/openshift-qe-cluster-density-v2-commands.sh
@@ -14,8 +14,10 @@ source ./venv_qe/bin/activate
 ES_PASSWORD=$(cat "/secret/password")
 ES_USERNAME=$(cat "/secret/username")
 
-
-git clone https://github.com/cloud-bulldozer/e2e-benchmarking --depth=1
+REPO_URL="https://github.com/cloud-bulldozer/e2e-benchmarking";
+LATEST_TAG=$(curl -s "https://api.github.com/repos/cloud-bulldozer/e2e-benchmarking/releases/latest" | jq -r '.tag_name');
+TAG_OPTION="--branch $(if [ "$E2E_VERSION" == "default" ]; then echo "$LATEST_TAG"; else echo "$E2E_VERSION"; fi)";
+git clone $REPO_URL $TAG_OPTION --depth 1
 pushd e2e-benchmarking/workloads/kube-burner-ocp-wrapper
 export WORKLOAD=cluster-density-v2
 
--- a/ci-operator/step-registry/openshift-qe/cluster-density-v2/openshift-qe-cluster-density-v2-ref.yaml
+++ b/ci-operator/step-registry/openshift-qe/cluster-density-v2/openshift-qe-cluster-density-v2-ref.yaml
@@ -5,8 +5,12 @@ ref:
     name: ocp-qe-perfscale-ci
     tag: latest
   env:
+  - name: E2E_VERSION
+    default: "v2.1.2"
+    documentation: |-
+      Override the e2e version
   - name: KUBE_BURNER_VERSION
-    default: "1.7.12"
+    default: "default"
     documentation: |-
       Override the kube burner version
   - name: CHURN
--- a/ci-operator/step-registry/openshift-qe/ingress-perf/openshift-qe-ingress-perf-commands.sh
+++ b/ci-operator/step-registry/openshift-qe/ingress-perf/openshift-qe-ingress-perf-commands.sh
@@ -15,7 +15,10 @@ ES_PASSWORD=$(cat "/secret/password")
 ES_USERNAME=$(cat "/secret/username")
 
 # Clone the e2e repo
-git clone https://github.com/cloud-bulldozer/e2e-benchmarking
+REPO_URL="https://github.com/cloud-bulldozer/e2e-benchmarking";
+LATEST_TAG=$(curl -s "https://api.github.com/repos/cloud-bulldozer/e2e-benchmarking/releases/latest" | jq -r '.tag_name');
+TAG_OPTION="--branch $(if [ "$E2E_VERSION" == "default" ]; then echo "$LATEST_TAG"; else echo "$E2E_VERSION"; fi)";
+git clone $REPO_URL $TAG_OPTION --depth 1
 pushd e2e-benchmarking/workloads/ingress-perf
 
 # ES Configuration
--- a/ci-operator/step-registry/openshift-qe/conc-builds/openshift-qe-conc-builds-commands.sh
+++ b/ci-operator/step-registry/openshift-qe/conc-builds/openshift-qe-conc-builds-commands.sh
@@ -16,7 +16,10 @@ ES_USERNAME=$(cat "/secret/username")
 
 export KUBE_BURNER_URL="https://github.com/cloud-bulldozer/kube-burner/releases/download/v0.17.3/kube-burner-0.17.3-Linux-x86_64.tar.gz"
 
-git clone https://github.com/cloud-bulldozer/e2e-benchmarking/ --depth=1
+REPO_URL="https://github.com/cloud-bulldozer/e2e-benchmarking";
+LATEST_TAG=$(curl -s "https://api.github.com/repos/cloud-bulldozer/e2e-benchmarking/releases/latest" | jq -r '.tag_name');
+TAG_OPTION="--branch $(if [ "$E2E_VERSION" == "default" ]; then echo "$LATEST_TAG"; else echo "$E2E_VERSION"; fi)";
+git clone $REPO_URL $TAG_OPTION --depth 1
 pushd e2e-benchmarking/workloads/kube-burner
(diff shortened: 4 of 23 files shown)


ci-operator/step-registry/openshift-qe/cluster-density-v2/openshift-qe-cluster-density-v2-commands.sh
```diff
@@ -14,8 +14,10 @@ source ./venv_qe/bin/activate
 ES_PASSWORD=$(cat "/secret/password")
 ES_USERNAME=$(cat "/secret/username")
 
-
-git clone https://github.com/cloud-bulldozer/e2e-benchmarking --depth=1
+REPO_URL="https://github.com/cloud-bulldozer/e2e-benchmarking";
+LATEST_TAG=$(curl -s "https://api.github.com/repos/cloud-bulldozer/e2e-benchmarking/releases/latest" | jq -r '.tag_name');
+TAG_OPTION="--branch $(if [ "$E2E_VERSION" = "default" ]; then echo "$LATEST_TAG"; else echo "$E2E_VERSION"; fi)";
```
rsevilla87, line 19:
```suggestion
TAG_OPTION="--branch $(if [ "$E2E_VERSION" == "default" ]; then echo "$LATEST_TAG"; else echo "$E2E_VERSION"; fi)";
```

vishnuchalla: Changed as suggested for better readability.

vishnuchalla:
/pj-rehearse periodic-ci-openshift-qe-ocp-qe-perfscale-ci-main-gcp-4.13-nightly-x86-data-path-6nodes

ci-operator/step-registry/openshift-qe/ingress-perf/openshift-qe-ingress-perf-commands.sh
```diff
@@ -15,7 +15,10 @@ ES_PASSWORD=$(cat "/secret/password")
 ES_USERNAME=$(cat "/secret/username")
 
 # Clone the e2e repo
-git clone https://github.com/cloud-bulldozer/e2e-benchmarking
+REPO_URL="https://github.com/cloud-bulldozer/e2e-benchmarking";
+LATEST_TAG=$(curl -s "https://api.github.com/repos/cloud-bulldozer/e2e-benchmarking/releases/latest" | jq -r '.tag_name');
```
paigerube14, line 19: we had issues on jenins being able to consistently connect and get the latest tag. I wonder if we will hit that here as well. Might need another option

ci-operator/step-registry/openshift-qe/cluster-density-v2/openshift-qe-cluster-density-v2-commands.sh
```diff
@@ -14,8 +14,10 @@ source ./venv_qe/bin/activate
 ES_PASSWORD=$(cat "/secret/password")
 ES_USERNAME=$(cat "/secret/username")
 
-
-git clone https://github.com/cloud-bulldozer/e2e-benchmarking --depth=1
+REPO_URL="https://github.com/cloud-bulldozer/e2e-benchmarking";
```
paigerube14, line 17: should we add this repo url as an env variable as well so that in the future we can specify it in the config files if we ever want to test a pr/someone else repo/branch

ci-operator/step-registry/openshift-qe/cluster-density-v2/openshift-qe-cluster-density-v2-ref.yaml
```diff
@@ -5,8 +5,12 @@ ref:
     name: ocp-qe-perfscale-ci
     tag: latest
   env:
+  - name: E2E_VERSION
+    default: "default"
```
afcollins, line 9: I believe this is the version we want to hard-code to lock-in the specific e2e tag? "default" is the same as previous "latest", which leaves us open to unexpected changes.

ci-operator/step-registry/openshift-qe/ingress-perf/openshift-qe-ingress-perf-commands.sh
afcollins: +1 Also, do we even want a latest tag vs latest i.e. git clone of main branch?

vishnuchalla: That shouldn't ideally happen here as it is just hitting the endpoint and getting the tag_name. Where as previous command had some processing like below
```
$(curl -s https://api.github.com/repos/cloud-bulldozer/kube-burner/releases/latest | jq -r '.assets | map(select(.name | test("linux-x86_64"))) | .[0].browser_download_url')
```
And it used to run `tar` on top of it to extract binaries too.

vishnuchalla: That shouldn't ideally happen here as it is just hitting the endpoint and getting the tag_name. Where as previous command had some processing like below
```
$(curl -s https://api.github.com/repos/cloud-bulldozer/kube-burner/releases/latest | jq -r '.assets | map(select(.name | test("linux-x86_64"))) | .[0].browser_download_url')
```
And it used to run `tar` on top of it to extract binaries too.

vishnuchalla: So if someone wants a specific tag, they can override `E2E_VERSION` in their test or else `latest` would be picked by default.

ci-operator/step-registry/openshift-qe/cluster-density-v2/openshift-qe-cluster-density-v2-ref.yaml
vishnuchalla: Yes it picks `latest` tag by default. But before we even cut a release with `latest` tag, it is recommended to test and verify them using `pre-release` version and once everything works fine we can get it tagged as `latest`.
Example: https://github.com/vishnuchalla/e2e-benchmarking/releases/tag/v0.0.0

ci-operator/step-registry/openshift-qe/cluster-density-v2/openshift-qe-cluster-density-v2-commands.sh
vishnuchalla: * Testing PRs is out of scope for this change. It needs to be taken through this JIRA: https://issues.redhat.com/browse/PERFSCALE-2731
* And the idea is to use only tags for testing but not branches going forward as they are error prone. (because as soon as a PR gets merged branch gets updated with those changes whereas release tag does not)

ci-operator/step-registry/openshift-qe/cluster-density-v2/openshift-qe-cluster-density-v2-ref.yaml
afcollins: I don't think using latest solves the problem. Please hard-code the exact release version we've tested.

vishnuchalla: My idea was to use latest tag by default, and pass on a rule to the team as _"whenever some is cutting a release go with the pre-release first, verify it in PROW, and if it works well mark it as latest"_. If we think this plan doesn't work, we can hard code the tag number here. But one thing to note with hard-coding is, for every new release we will need to raise a PR here updating the value of `E2E_VERSION` env in multiple files, whereas it is not required in the above approach that I am suggesting.

afcollins: Ok. I think that's a process we can refine as needed.

> But one thing to note with hard-coding is, for every new release we will need to raise a PR here updating the value of E2E_VERSION env in multiple files,

This is very much the intent for now: To freeze the e2e version and require a PR and human eyes/hands on the upgrade step so nothing catches us by surprise.

Thanks!

vishnuchalla: Done.

vishnuchalla:
/pj-rehearse periodic-ci-openshift-qe-ocp-qe-perfscale-ci-main-gcp-4.13-nightly-x86-data-path-6nodes

afcollins:
/pj-rehearse periodic-ci-openshift-qe-ocp-qe-perfscale-ci-main-aws-4.15-nightly-x86-payload-control-plane-6nodes periodic-ci-openshift-qe-ocp-qe-perfscale-ci-main-aws-4.15-nightly-x86-payload-node-density-cni-6nodes periodic-ci-openshift-qe-ocp-qe-perfscale-ci-main-aws-4.15-nightly-x86-payload-node-density-heavy-6nodes

afcollins (approved):
/lgtm

PR #46777, opened 2023-12-14
Disable weekly job for 4.14 ppc64le

aishwaryabk:
FYI: @juliemathew @pravin-dsilva


aishwaryabk:
@liangxia @aleskandro Can you please review? Thanks

PR #46783, opened 2023-12-14
use openshift-ci-aws.rhaiseng.com as base domain

riprasad:


riprasad:
@Xaenalt We discovered an issue with the previously used base domain. Hence, updating to use the working one

riprasad:
@spolti @Xaenalt Could you please merge this? we have changed the base domain, the previous one had some issues

spolti:
hi @riprasad I am curious, why only on one specific version target?
aren't the other version affected too?

And, if possible, can you share more context about what is the issue?

riprasad:
@spolti We had missing NS record for the sub-domain at `odhdev.com`. I was unable to find where is that domain registered and managed.  As `rhaiseng.com` is managed on Route 53, I was able to add the necessary NS records for the new sub-domain.

> I am curious, why only on one specific version target? aren't the other version affected too?
I am not sure if I understand your point here, but as I explained above, this was a case of missing NS records and this is a DNS issue, completely unrelated to OpenShift.

spolti:
> I am curious, why only on one specific version target? aren't the other version affected too?

I meant, about the other jobs (kserve, mm, etc..) and for the other rest-proxy jogs, this one targets the 0.10 release.

spolti (approved):

PR #46800, opened 2023-12-14
Turn off review acts as lgtm and lgtm acts as approve for OLS

bparees:


eparis:
this means github review is meaningless?

bparees:
> this means github review is meaningless?

it's still useful as a way to review someone's code (provide a set of feedback/comments all tied to one round of reviewing, with a summary comment), but it will have no effect on actually getting the PR labeled for merge.

I don't want people accidentally lgtm'ing (let alone approving) PRs just by submitting a review

vishnuchalla:
looks like one of the PROW check is failing, so putting it on hold until it gets fixed.

bparees:
looks like it wants to drop the "false" fields entirely rather than setting them explicitly :roll_eyes: 

fixed+repushed, i'll reapply the lgtm.

PR #46811, opened 2023-12-14
OSSM-4914 Change istio operator jobs to ossm aws profile

fjglira:
Jira: https://issues.redhat.com/browse/OSSM-4914


fjglira:
It can't be used this ossm-aws profile because hyper shift cluster has domain hardcoded with aws-2 profile

PR #46834, opened 2023-12-15
PODAUTO-86: Run CMA tests from makefile targets in test image, migrate inline Dockerfile to keda repo where it belongs 

jkyros:
Right now if we want to fix tests and add or remove them from our run list (which is something we may be doing a lot more of as we start supporting new scalers) we have to put up a CI PR because of how I configured the targets.

This:

-  Uses makefile targets to run the tests instead of calling them directly in both the CMA operator and operand 
-  Uses the Dockerfile.tests dockerfile in the operand (keda) repo instead of the inlined Dockerfile

Which should let us get instant feedback on test changes when we put up a CMA PR instead of having to change it in CMA and then come back here to update the CI config to test it. 

This should merge after https://github.com/openshift/kedacore-keda/pull/19, as the targets we're switching to won't exist until that gets in.

### DIFF
--- a/ci-operator/config/openshift/custom-metrics-autoscaler-operator/openshift-custom-metrics-autoscaler-operator-main.yaml
+++ b/ci-operator/config/openshift/custom-metrics-autoscaler-operator/openshift-custom-metrics-autoscaler-operator-main.yaml
@@ -132,25 +132,9 @@ tests:
       resources:
         requests:
           cpu: 100m
-    - as: test-internals
+    - as: test-run
       cli: latest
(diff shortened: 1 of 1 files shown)


jkyros:
/pj-rehearse pull-ci-openshift-kedacore-keda-main-keda-e2e-aws-ovn

jkyros:
awesome, the keda one passed, let's try the CMA one: 
/pj-rehearse pull-ci-openshift-custom-metrics-autoscaler-operator-main-cma-e2e-aws-ovn

jkyros:
ok awesome, both of those worked, and I squashed out the fixup commit
/pj-rehearse ack

PR #46838, opened 2023-12-15
[Prow CI chain Enhancement]: add configurable 'DISCONNECTED' and 'ARCHITECTURE'

LiangquanLi930:
Adjust hypershift-conformance to support disconnected and different architectures, as there will be various combinations for the agent.
For example:

- dualstack + connected
- dualstack + disconnected
- v4 + connected
- v4 + disconnected
- v6 + disconnected

And for agent, We have amd, prow, ibmz and arm.
So adding the DISCONNECTED and ARCHITECTURE variable makes it easier for us to conduct conformance tests.


dharaneeshvrd:
@LiangquanLi930 can you please add a var to decide calling mirror_test_images func as well?
mirror_test_images func is more specific to amd64 and we don't want to call that func for mce power conformance for now.

LiangquanLi930:
> @LiangquanLi930 can you please add a var to decide calling mirror_test_images func as well?
> mirror_test_images func is more specific to amd64 and we don't want to call that func for mce power conformance for now.

We can solve it by adding ARCHITECTURE variable

PR #46859, opened 2023-12-15
MIXEDARCH-356: resolve temp state of the service instance

prb112:


jaypoulz (approved):
/lgtm

I'm approving these with the understanding that we are trying to fix a bug with a race condition where we check the state of a Cloud Resource before it has a readable state. This way we default to NOT_READY even when the cloud isn't unable to report a status to us yet.

prb112:
Thanks @jaypoulz I was holding on the PR until I had a good test in another branch... basically service-instances is paged. This works around the paging.

PR #46871, opened 2023-12-18
Fix cap CCO in post upgrade check

shellyyang1989:
Refers [OCPQE-18367](https://issues.redhat.com/browse/OCPQE-18367)

Prior to the change, cap CCO was added to v4.15 capset. With the change, cap CCO will be only added to v4.15 capset when cluster provider is baremetal.

### DIFF
--- a/ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
+++ b/ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
@@ -68,10 +68,11 @@ function get_caps_for_version_capset() {
     caps_string="baremetal marketplace openshift-samples Console Insights Storage CSISnapshot NodeTuning MachineAPI Build DeploymentConfig ImageRegistry"
     ;;
     "v4.15")
-    caps_string="baremetal marketplace openshift-samples Console Insights Storage CSISnapshot NodeTuning MachineAPI Build DeploymentConfig ImageRegistry OperatorLifecycleManager CloudCredential"
+    caps_string="baremetal marketplace openshift-samples Console Insights Storage CSISnapshot NodeTuning MachineAPI Build DeploymentConfig ImageRegistry OperatorLifecycleManager"
+    [[ "${CLUSTER_TYPE}" =~ ^packet.*$|^equinix.*$ ]] && caps_string="${caps_string} CloudCredential"
     ;;
     *)
-    caps_string="baremetal marketplace openshift-samples Console Insights Storage CSISnapshot NodeTuning MachineAPI Build DeploymentConfig ImageRegistry OperatorLifecycleManager CloudCredential"
+    caps_string="baremetal marketplace openshift-samples Console Insights Storage CSISnapshot NodeTuning MachineAPI Build DeploymentConfig ImageRegistry OperatorLifecycleManager"
     ;;
     esac
 


ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
```diff
@@ -68,10 +68,11 @@ function get_caps_for_version_capset() {
     caps_string="baremetal marketplace openshift-samples Console Insights Storage CSISnapshot NodeTuning MachineAPI Build DeploymentConfig ImageRegistry"
     ;;
     "v4.15")
-    caps_string="baremetal marketplace openshift-samples Console Insights Storage CSISnapshot NodeTuning MachineAPI Build DeploymentConfig ImageRegistry OperatorLifecycleManager CloudCredential"
+    caps_string="baremetal marketplace openshift-samples Console Insights Storage CSISnapshot NodeTuning MachineAPI Build DeploymentConfig ImageRegistry OperatorLifecycleManager"
+    [[ "${CLUSTER_TYPE}" =~ ^packet.*$|^equinix.*$ ]] && caps_string="${caps_string} CloudCredential"
     ;;
     *)
-    caps_string="baremetal marketplace openshift-samples Console Insights Storage CSISnapshot NodeTuning MachineAPI Build DeploymentConfig ImageRegistry OperatorLifecycleManager CloudCredential"
+    caps_string="baremetal marketplace openshift-samples Console Insights Storage CSISnapshot NodeTuning MachineAPI Build DeploymentConfig ImageRegistry OperatorLifecycleManager"
```
jianlinliu, line 75: Shall we also append the line 72 here, too?

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
```diff
@@ -68,10 +68,11 @@ function get_caps_for_version_capset() {
     caps_string="baremetal marketplace openshift-samples Console Insights Storage CSISnapshot NodeTuning MachineAPI Build DeploymentConfig ImageRegistry"
     ;;
     "v4.15")
-    caps_string="baremetal marketplace openshift-samples Console Insights Storage CSISnapshot NodeTuning MachineAPI Build DeploymentConfig ImageRegistry OperatorLifecycleManager CloudCredential"
+    caps_string="baremetal marketplace openshift-samples Console Insights Storage CSISnapshot NodeTuning MachineAPI Build DeploymentConfig ImageRegistry OperatorLifecycleManager"
+    [[ "${CLUSTER_TYPE}" =~ ^packet.*$|^equinix.*$ ]] && caps_string="${caps_string} CloudCredential"
```
jianlinliu, line 72: Hmm, actually for v4.15, `CloudCredential` indeed become an optional cap, but does not matter with platform, right? For non-baremetal platform, we have to configure `CloudCredential` as additional cap, and enable it, so that move forwards the testing, right? So I think we should get the caps from job configurations, but not reply on `CLUSTER_TYPE`

shellyyang1989: ah, you're right. Re-reviewing [OCPQE-18367](https://issues.redhat.com/browse/OCPQE-18367), looks like it was using 4.15.0-ec.3 for testing which doesn't include CCO cap yet. So the post check failed. It doesn't seem to be a CI issue.

PR #46891, opened 2023-12-18
MGMT-16419: assisted vsphere jobs are leaking passwords

adriengentil:
disable capture log in pytest to prevent vsphere credentials to be
leaked

### DIFF
--- a/ci-operator/step-registry/assisted/vsphere/setup/configure/assisted-vsphere-setup-configure-commands.sh
+++ b/ci-operator/step-registry/assisted/vsphere/setup/configure/assisted-vsphere-setup-configure-commands.sh
@@ -45,6 +45,9 @@ fi
 echo "$(date -u --rfc-3339=seconds) - Creating platform-conf.sh file..."
 cat >>"${SHARED_DIR}/platform-conf.sh" <<EOF
 export PLATFORM=vsphere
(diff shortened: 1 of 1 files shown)


adriengentil:
/pj-rehearse pull-ci-openshift-assisted-service-master-edge-e2e-vsphere-assisted

rccrdpccl:
Isn't there a better system? A wrapper for prow jobs or something?


/lgtm
/approve

adriengentil:
> Isn't there a better system? A wrapper for prow jobs or something?

Probably, I want to stop the bleeding first, and I'll open another ticket to follow-up. We have something is https://github.com/openshift/assisted-test-infra/blob/master/src/service_client/logger.py#L12 to redact the passwords, but for some reasons, it's bypassed by pytest :man_shrugging:

rccrdpccl:
> > Isn't there a better system? A wrapper for prow jobs or something?
> 
> Probably, I want to stop the bleeding first, and I'll open another ticket to follow-up. We have something is https://github.com/openshift/assisted-test-infra/blob/master/src/service_client/logger.py#L12 to redact the passwords, but for some reasons, it's bypassed by pytest 🤷‍♂️

I meant more generic, something closer to prow's output rather than our code, a bit like Jenkins or SaaS solutions do

PR #46895, opened 2023-12-18
Updates to devfile registry operator operator-ci config

michael-valdron:
# Changes

- Changes from https://github.com/openshift/release/pull/46094 for devfile registry operator release branch `release-v0` config
- Set target architecture build arg set to 'amd64' for devfile registry operator image builds https://github.com/openshift/release/pull/46896#issue-2047256257

PR #46896, opened 2023-12-18
Set target architecture build arg set to 'amd64' for devfile registry operator image builds

michael-valdron:
# Description

Changes to support multiple architectures for the devfile registry operator (devfile/registry-operator#68) is causing OpenShift CI testing to fail due to `TARGETARCH` build argument being overridden with a blank value.

This requires `TARGETARCH` to be manually set to the intended architecture for testing in OpenShift CI, currently only `amd64`.


michael-valdron:
Folded these changes into https://github.com/openshift/release/pull/46895.

PR #46901, opened 2023-12-18
WIP Update files to support z-stream testing 3scale.

ascerra:


ascerra:
/pj-rehearse periodic-ci-3scale-qe-3scale-deploy-main-3scale-amp-ocp4.13-lp-interop-3scale-amp-interop-aws

ascerra:
/pj-rehearse periodic-ci-3scale-qe-3scale-deploy-main-3scale-amp-ocp4.14-lp-interop-3scale-amp-interop-aws

ci-operator/config/3scale-qe/3scale-deploy/3scale-qe-3scale-deploy-main__3scale-amp-ocp4.13-lp-interop.yaml
```diff
@@ -43,17 +43,18 @@ tests:
         {
           "failure_rules":
             [
-              {"step": "ipi-*", "failure_type": "all", "classification": "Infrastructure Provisioning - Cluster"},
-              {"step": "install-operators", "failure_type": "all", "classification": "Infrastructure Provisioning - Operators"},
-              {"step": "gather-*", "failure_type": "pod_failure", "classification": "Other"},
+              {"step": "ipi-*", "failure_type": "all", "classification": "Infrastructure Provisioning - Cluster", "group": {"name": "cluster", "priority": 1}, "jira_additional_labels": ["!default"]},
+              {"step": "install-operators", "failure_type": "all", "classification": "Infrastructure Provisioning - Operators", "group": {"name": "lp-tests", "priority": 1}, "jira_additional_labels": ["!default"]},
+              {"step": "gather-*", "failure_type": "pod_failure", "classification": "Other", "group": {"name": "cluster", "priority": 2}, "jira_additional_labels": ["!default"]},
```
Simranpal, line 48: Should we add the z-stream test labels to these steps

Simranpal, line 48: Sorry i dont recall but what was the priority for?  Is it for jira tickets

Simranpal (commented):
I have also noticed you are still keeping the same file for z-stream tests?
Shouldn't we have seperate file names

ci-operator/config/3scale-qe/3scale-deploy/3scale-qe-3scale-deploy-main__3scale-amp-ocp4.13-lp-interop.yaml
ascerra: Yes good point, I've added the label "z-stream" to default vars using the variable FIREWATCH_DEFAULT_JIRA_ADDITIONAL_LABELS

ascerra: priority is for the [group feature in firewatch](https://github.com/CSPI-QE/firewatch/blob/main/docs/configuration_guide.md#group)

basically we group together certain steps so that we can control how many tickets get created by firewatch when one failed step leads to other cascading failures. For example if in the "cluster" group the step ipi-* fails then there is a good chance the must-gather step will fail since there was something wrong with cluster deployment. This group feature allows us to not create a new issue for a failed must-gather step in this example but just create a ticket for the real issue which would be cluster provisioning failure

ascerra: You've also made me notice that the firewatch rules here a bit to vague and we could benefit from making them a bit more specific. I'll update this here.

Simranpal: Got it. Thanks for the explanation. You mean the name of rule being vague?

Simranpal:
> > I have also noticed you are still keeping the same file for z-stream tests? Shouldn't we have seperate file names
> 
> This is a good question, the filename as it exists does not have any misleading information, which is good. But now the question is, do we want to have a unique identifier somewhere in the job name. Something like `-lp-interop-z-stream`
> 
> We could place this unique identifer in either the filename or in the test stanza name on line 33
> 
> What are your thoughts @Simranpal ?

I was thinking about about unique identifier for the prow job and therefore the reporting. In order to distinguish the z-stream it from self managed lp tests, having a seperate job makes sense.
We can also have a separate test stanza name and that would mean we run both z-stream and existing tests under 1 job. We can discuss this in our next meeting as well.

ascerra:
leaving this PR up for now as WIP but we are not currently targeting a merge of this PR at this time.

PR #46919, opened 2023-12-19
add minot tf-provider cases to optional presubmit run

amalykhi:
@shellymiron @xueli181114 @AsherShoshan  Please review.
I want to verify all the cases added in presubmit full run


ci-operator/config/terraform-redhat/terraform-provider-rhcs/OWNERS
```diff
@@ -25,3 +26,5 @@ reviewers:
 - robpblake
 - sagidayan
 - vkareh
+- shellymiron
```
nirarg, line 29: I think its generated from the the OWNERS file in the repository
I will approve it, lets see if it going to be override

PR #46945, opened 2023-12-19
[OCPBUGS-24691]: escape success_count var in nw migration rollback

jluhrsen:
this was accidentally added here:
https://github.com/openshift/release/pull/46686/commits/5d7b592ccd56583e2028a9a9bc3a59f061eba600


ci-operator/step-registry/ovn/sdn-migration-rollback/ovn-sdn-migration-rollback-commands.sh
```diff
@@ -77,7 +77,7 @@ oc wait mcp --all --for='condition=UPDATING=True' --timeout=300s
 # shellcheck disable=SC2154
```
kyrtapz, line 77: Is this disable still required?
https://www.shellcheck.net/wiki/SC2154

jluhrsen: nope. good catch. updated.

PR #46950, opened 2023-12-19
Set OKD releases 4.13 and earlier to EOL

sdodson:
Also bump the minimum build interval
Remove 4.13 test jobs


sdodson:
I believe the `endOfLife` is probably sufficient based on reading this but this brings the OKD releases inline with OCP EOL version minimum intervals.

https://github.com/openshift/release-controller/blob/e8b37875d348a4cd5e29b5450851fc1f4a04e9ae/pkg/release-controller/types.go#L82-L84

LorbusChris:
Thanks!
/lgtm

PR #46965, opened 2023-12-20
installer: Clean up agent e2e test jobs

zaneb:


ci-operator/config/openshift/installer/openshift-installer-master.yaml
```diff
@@ -772,7 +772,6 @@ tests:
     cluster_profile: equinix-ocp-metal
     env:
       DEVSCRIPTS_CONFIG: |
-        AGENT_E2E_TEST_SCENARIO=COMPACT_IPV4
```
andfasano, line 775: IINW partial override is not supported, the env var `DEVSCRIPTS_CONFIG` defined at the job level is going to replace fully the one defined at the workflow level. For such cases it would be probably better to use the generic workflow, which does not define any specific configuration

ci-operator/config/openshift/installer/openshift-installer-release-4.14__okd.yaml
```diff
@@ -201,30 +201,6 @@ tests:
     - ref: ipi-conf-aws-fcos-ami
     - chain: ipi-install
     workflow: openshift-e2e-aws
-- as: e2e-agent-compact-ipv4
```
andfasano, line 204: These jobs were meant to support backporting on 4.14

ci-operator/config/openshift/installer/openshift-installer-master.yaml
zaneb: Thanks, I definitely missed that. Should be fixed now.

ci-operator/config/openshift/installer/openshift-installer-release-4.14__okd.yaml
zaneb: I can't think of any reason why we would backport that.

zaneb:
/remove-lifecycle stale
I'd like to get this merged so we can stop burning money.

ci-operator/config/openshift/installer/openshift-installer-release-4.12.yaml
```diff
@@ -548,7 +548,7 @@ tests:
       DEVSCRIPTS_CONFIG: |
         AGENT_E2E_TEST_SCENARIO=COMPACT_IPV4
         AGENT_DEPLOY_MCE=true
-    workflow: agent-e2e-generic
+    workflow: agent-e2e-compact-ipv4
```
andfasano, line 551: This seems a typo, as the `DEVSCRIPTS_CONFIG` will override the one provided byu `agente2e-compact-ipv4` workflow

ci-operator/config/openshift/installer/openshift-installer-release-4.14__okd.yaml
andfasano: 4.14 backport is no more relevant, as ABI OKD was available in 4.15

andfasano:
> /remove-lifecycle stale I'd like to get this merged so we can stop burning money.

I have one general concern about hiding all the agent OKD jobs: now that ABI is supported (since 4.15), there could be a regression breaking it and it will pass unnoticed. At least the compact one could be useful to keep it.

zaneb:
> I have one general concern about hiding all the agent OKD jobs: now that ABI is supported (since 4.15), there could be a regression breaking it and it will pass unnoticed. At least the compact one could be useful to keep it.

From memory, that's what this patch does I thought?
In 4.15 OKD SCOS is removed; OKD compact runs as before; other OKD jobs become on-demand.

ci-operator/config/openshift/installer/openshift-installer-release-4.12.yaml
zaneb: Yeah, they both do the same thing today, since the compact workflow doesn't contain any config aside from DEVSCRIPTS_CONFIG.
It's arguable which one is more explicit.

andfasano: For the `generic` workflow is expected that the configuration will be (optionally) provided in the place where it's used, while for the specific workflows it's not expected to provided any additional config, so I'd keep the `generic` one

andfasano:
> > I have one general concern about hiding all the agent OKD jobs: now that ABI is supported (since 4.15), there could be a regression breaking it and it will pass unnoticed. At least the compact one could be useful to keep it.
> 
> From memory, that's what this patch does I thought? In 4.15 OKD SCOS is removed; OKD compact runs as before; other OKD jobs become on-demand.

oops, I guess I've misread this [line](https://github.com/openshift/release/pull/46965/files#diff-d3178f31c6eda5ba7966fe8a4329e9d46623667b886832fdb358c58d2db36ac0R287)

PR #46129, opened 2023-11-28
openshift/prometheus-operator: use base images from repo

simonpasquier:

PR #46969, opened 2023-12-20
ose-kube-metrics-server-container: use repository base images

rexagod:


rexagod:
I didn't see any `from: base` specifications below (for eg., [here](https://github.com/openshift/release/pull/46129/files#diff-d382d19cae13fc27b711cbe901d74951c66c2c1ff809b953e50745581d6763f8L15), so I've left that as is since this should be nothing more than specifying a couple of short-codes as inputs.

https://github.com/openshift/release/pull/46969/files#diff-fa01603cf1022418fc229bfc5d86f6c92daa749d3f409a365972b58145e34068R9-R18

simonpasquier:
I think that this change isn't needed. 6a9b406585a80f743f3e456aa4ad7b7038d3f70f was the appropriate change to use the `.ci-operator.yaml` from the repository.

PR #46990, opened 2023-12-20
CORS-2950: Use OVNKubernetes as default network type for agent installer

bfournie:
As OpenshiftSDN is being removed as an option for networkType, set the default network type for the agent-based installer to be OVNKubernetes. Older release will continue to use OpenshiftSDN.

### DIFF
--- a/ci-operator/config/openshift/installer/openshift-installer-master.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-master.yaml
@@ -762,6 +762,11 @@ tests:
   run_if_changed: /agent/
   steps:
     cluster_profile: equinix-ocp-metal
+    env:
+      DEVSCRIPTS_CONFIG: |
+        AGENT_E2E_TEST_SCENARIO=COMPACT_IPV4
+        FIPS_MODE=true
+        NETWORK_TYPE=OVNKubernetes
     workflow: agent-e2e-compact-ipv4
   timeout: 6h0m0s
 - as: e2e-agent-compact-ipv4-none-platform


ci-operator/config/openshift/installer/openshift-installer-master.yaml
```diff
@@ -762,6 +762,9 @@ tests:
   run_if_changed: /agent/
   steps:
     cluster_profile: equinix-ocp-metal
+    env:
+      DEVSCRIPTS_CONFIG: |
+        NETWORK_TYPE=OVNKubernetes
```
danwinship, line 767: I don't think this will work... either this setting of `DEVSCRIPTS_CONFIG` will override the one in `ci-operator/step-registry/agent/e2e/compact/ipv4/`, or that one will override this one. They definitely won't get merged.

Probably you should add `NETWORK_TYPE` there

zaneb: The problem with both this patch and changing it in the step registry is that it changes it for all versions, so pre-4.15 release-* branches won't test OpenShiftSDN any more.

bfournie: > The problem with both this patch and changing it in the step registry is that it changes it for all versions, so pre-4.15 release-* branches won't test OpenShiftSDN any more.

Won't earlier releases use the release specific file https://github.com/openshift/release/blob/master/ci-operator/config/openshift/installer/openshift-installer-release-4.14.yaml#L697-L701

bfournie:
/pj-rehearse pull-ci-openshift-installer-master-e2e-agent-compact-ipv4

ci-operator/config/openshift/installer/openshift-installer-master.yaml
bfournie: > I don't think this will work... either this setting of DEVSCRIPTS_CONFIG will override the one in ci-
> operator/step-registry/agent/e2e/compact/ipv4/, or that one will override this one. They definitely 
> won't get merged.
You're right, redoing this...

bfournie:
/pj-rehearse pull-ci-openshift-installer-master-e2e-agent-compact-ipv4

bfournie:
This looks to be working i.e. setting OVNKubernetes for > 4.15.

bfournie:
/pj-rehearse pull-ci-openshift-installer-master-e2e-agent-compact-ipv4

PR #46994, opened 2023-12-20
Scale down CVP resource for recharge week 

samvarankashyap:
$subject


bear-redhat:
/uncc
already approved

PR #47000, opened 2023-12-20
Disable spot instances for the cloud api chaos tests

chaitanyaenr:
Spot instances can't be stopped and started, so can't be used for the node disruption test cases.


paigerube14:
This is already here: https://github.com/openshift/release/blob/a2a50d062fd9a1cde2875dddd6c8bddc5a8b91b5/ci-operator/config/redhat-chaos/krkn-hub/redhat-chaos-krkn-hub-main__4.14-nightly.yaml#L51-L63

Had just moved it to be in one config instead of multiple

PR #47001, opened 2023-12-20
MGMT-16399: Pluralize VIP env vars in assisted vsphere and nutanix tests

carbonin:
The singular version of this is no longer in use as of https://github.com/openshift/assisted-test-infra/pull/2315

Requires https://github.com/openshift/assisted-test-infra/pull/2339

Before this change vsphere jobs were failing with errors like:

```
             # patch the aci with the vips. The cidr will be derived from the range
            access_vips = nodes.controller.get_ingress_and_api_vips()
>           api_vip = access_vips["api_vips"][0].get("ip", "") if len(access_vips["api_vips"]) > 0 else ""
E           TypeError: 'NoneType' object is not subscriptable

src/tests/test_kube_api.py:163: TypeError
```

An example failing job is https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-assisted-test-infra-master-e2e-vsphere-assisted-kube-api-periodic/1736174421904723968

https://issues.redhat.com/browse/MGMT-16399

cc @adriengentil

### DIFF
--- a/ci-operator/step-registry/assisted/nutanix/setup/configure/assisted-nutanix-setup-configure-commands.sh
+++ b/ci-operator/step-registry/assisted/nutanix/setup/configure/assisted-nutanix-setup-configure-commands.sh
@@ -55,8 +55,8 @@ export NUTANIX_ENDPOINT='${NUTANIX_HOST}'
 export NUTANIX_PORT="${NUTANIX_PORT}"
 export NUTANIX_CLUSTER_NAME="${pe_name}"
 export NUTANIX_SUBNET_NAME="${subnet_name}"
-export API_VIP="${API_VIP}"
-export INGRESS_VIP="${INGRESS_VIP}"
+export API_VIPS="[{\"ip\": \"${API_VIP}\"}]"
+export INGRESS_VIPS="[{\"ip\": \"${INGRESS_VIP}\"}]"
 export BASE_DOMAIN="${base_domain}"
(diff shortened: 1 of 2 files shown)


ci-operator/step-registry/assisted/nutanix/setup/configure/assisted-nutanix-setup-configure-commands.sh
```diff
@@ -55,8 +55,8 @@ export NUTANIX_ENDPOINT='${NUTANIX_HOST}'
 export NUTANIX_PORT="${NUTANIX_PORT}"
 export NUTANIX_CLUSTER_NAME="${pe_name}"
 export NUTANIX_SUBNET_NAME="${subnet_name}"
-export API_VIP="${API_VIP}"
-export INGRESS_VIP="${INGRESS_VIP}"
+export API_VIPS="[{\"ip\": ${API_VIP}}]"
+export INGRESS_VIPS="[{\"ip\": ${INGRESS_VIP}}]"
```
adriengentil, line 59:
```suggestion
export API_VIPS="[{\"ip\": \"${API_VIP}\"}]"
export INGRESS_VIPS="[{\"ip\": \"${INGRESS_VIP}\"}]"
```

I think the IPs need to be quoted

carbonin: Nice, catch. Done

adriengentil (approved):

PR #47004, opened 2023-12-20
upstream OLM: stop using Tide

ncdc:
Upstream OLM uses GitHub's actions and merge queue for CI and merging.


tmshort:
You're _removing_ excluded branches to stop using tide?

ncdc:
@tmshort yes, this covers `master`, which is what I care about right now

tmshort:
Ah, I see it now, seems a bit backwards.

PR #47034, opened 2023-12-21
Interop - Update Firewatch config to match V2 [ROSA classic]

oharan2:
Main updates:

1. Using `lp-tests` steps in the Firewatch group feature
2. Updating `interop-tests` Jira label in LP tests steps


calebevans (approved):
/lgtm

PR #47048, opened 2023-12-22
Skip OutOfServiceTaint remediation verification if out-of-service taint is not supported

k-keiichi-rh:
This RP is to set a env variable to skip OutOfServiceTaint remediation verification if out-of-service taint is not supported.


slintes:
you need to run `make ci-operator-config` and amend changes

k-keiichi-rh:
> you need to run make ci-operator-config and amend changes

Thanks for the suggestion. I fixed it.

PR #47064, opened 2023-12-22
BMQE: Minor fixes

aleskandro:
Some files are kept after deletion and some BMCs have some delay from the response of the ipmi command to the actual action. This commit should fix them and also temporarily disables concurrency for the ipmi/vmedia commands to allow further debugging.


aleskandro:
/pj-rehearse skip

(done some testing locally, now needs the CI to continue running)

PR #47071, opened 2023-12-24
Modifying DTK's tests.

ybettan:
* `check-commits-count` was bumped to use a more recent image.

    * `verify-imagestream` and `image-contents` were combined. Both of
      them are validating the existence of key-files inside the DTK's
      built container image.

    * `e2e-aws-driver-toolkit-presubmit` was completely removed. This
      test was extremely confusing since it was not testing the actual
      PR but instead it was testing a nightly payload (with a DTK image
      built in the nightly and not in the PR). In addition to that, it
      was easy to forget to keep in sync the nightly version used and the
      AWS cluster provisioned (which is bumped automatically unlike the
      nightly payload chosen for the job).
      Since we are running the `test-e2e` target anyway on every nightly as
      periodic jobs, there is no reason to also test it on PRs (especially
      when we are running a different nightly/installer versions).

---

/cc @qbarrand @yevgeny-shnaidman

### DIFF
--- a/ci-operator/config/openshift/driver-toolkit/openshift-driver-toolkit-master.yaml
+++ b/ci-operator/config/openshift/driver-toolkit/openshift-driver-toolkit-master.yaml
@@ -3,10 +3,10 @@ base_images:
     name: "4.15"
     namespace: ocp
     tag: base-rhel9
-  ocp_builder_rhel-8-golang-1.18-openshift-4.12:
+  ocp_builder_rhel-9-golang-1.20-openshift-4.15:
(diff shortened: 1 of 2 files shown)


ybettan:
/override ci/rehearse/openshift/driver-toolkit/master/check-commits-count

It is supposed to fail since there are no actual commits to DTK but only to the release repo.

ybettan:
/pj-rehearse ack
/unhold

It is supposed to fail since there are no actual commits to DTK but only to the release repo.

PR #47095, opened 2023-12-26
OCPBUGS-25863: Serialize nodes reboot in SDN migration script

pliurh:
Rebooting all the nodes at the same time makes the api-server unavaliable.

### DIFF
--- a/ci-operator/step-registry/ovn/sdn-migration-rollback/ovn-sdn-migration-rollback-commands.sh
+++ b/ci-operator/step-registry/ovn/sdn-migration-rollback/ovn-sdn-migration-rollback-commands.sh
@@ -13,7 +13,7 @@ oc patch MachineConfigPool worker --type='merge' --patch '{"spec":{"paused":true
 oc patch Network.operator.openshift.io cluster --type='merge' --patch '{"spec":{"migration":null}}'
 # Wait until CNO update the applied-cluster cm to the latest
 timeout 120s bash <<EOT
-until 
+until
   ! oc get network -o yaml | grep migration  > /dev/null
 do
   echo "migration field is not cleaned by CNO"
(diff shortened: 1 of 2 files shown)


pliurh:
> @pliurh can we just do that with the masters and halve the additional time required?

It looks like a CI-specific issue. I tested on clusters built by cluster-bot. The migration can always work without this node reboot serialization. I also tried to use 1 or 2 mins interval. Neither worked. 3 mins is the minimal interval that can resolve the issue, although I have no clue why the AWS LB keeps not responding when all the master reboots at the same time.

jcaamano:
> > @pliurh can we just do that with the masters and halve the additional time required?
> 
> It looks like a CI-specific issue. I tested on clusters built by cluster-bot. The migration can always work without this node reboot serialization. I also tried to use 1 or 2 mins interval. Neither worked. 3 mins is the minimal interval that can resolve the issue, although I have no clue why the AWS LB keeps not responding when all the master reboots at the same time.

What I mean is that you are serializing both workers and masters. Could only masters be serialized, and workers be rebooted all at the same time as the masters?

flavio-fernandes (commented):
This looks reasonable and good to me, but I thought of asking: is there an existing issue in the api server when all nodes are rebooted at once? It may be a good thing that the test exposed that after all.

pliurh:
> This looks reasonable and good to me, but I thought of asking: is there an existing issue in the api server when all nodes are rebooted at once? It may be a good thing that the test exposed that after all.

I cannot reproduce it with cluster-bot clusters. I'll try to investigate the issue with CI clusters later.

pliurh:
@jcaamano I've updated the reboot logic as you suggested. PTAL.

ci-operator/step-registry/ovn/sdn-migration-rollback/ovn-sdn-migration-rollback-commands.sh
```diff
@@ -39,16 +39,30 @@ oc wait co network --for='condition=PROGRESSING=True' --timeout=120s
...
+        until oc rsh -n openshift-machine-config-operator "$pods_on_node" chroot /rootfs shutdown -r +$delay; do echo "cannot reboot node $node, retry"&&sleep 3; done
+        delay=$((delay+3))
+    fi
+done
+# reboot all the worker nodes
+for node in $worker_nodes; do
+    pods_on_node=$(oc get pods -n openshift-machine-config-operator -o jsonpath='{.items[?(@.spec.nodeName=="'$node'")].metadata.name}' -l k8s-app=$daemonset_name)
+    if [[ -n $pods_on_node ]]; then
+        echo "reboot worker node $node in ${delay}m"
+        until oc rsh -n openshift-machine-config-operator "$pods_on_node" chroot /rootfs shutdown -r +$delay; do echo "cannot reboot node $node, retry"&&sleep 3; done
```
jcaamano, line 61: @pliurh is delaying the workers reboot until after the masters have rebooted intentional? Or could they be rebooted at the same time?

pliurh: Yes, that is intentional. I found that rebooting the workers at the same time as the first master caused flakes. I haven't looked that in-depth, as it doesn't always happen. We might need to update the migration steps in the document to make it more stable.

jcaamano:
/lgtm
/approve

@pliurh could you open a bug to track if we need to investigate this further or update the docs?

kyrtapz:
/lgtm

@jluhrsen @abhat @trozet please help approve the PR, looks like Jaime is not on the approver list:
https://github.com/openshift/release/blob/396f475a7ea33c4d5f16fb21f3201dbcb505c5f9/OWNERS_ALIASES#L210-L216

pliurh:
> @pliurh could you open a bug to track if we need to investigate this further or update the docs?

Bug https://issues.redhat.com/browse/OCPBUGS-26417 was filed for further investigation.

PR #47121, opened 2023-12-31
MGMT-15772: Moving Equinix to use Metros

eifrach:
- `facility` has been deprecated 
switching to metro, after this [PR](https://github.com/openshift/assisted-test-infra/pull/2347)


eifrach:
/hold
found an issue with this change

eifrach:
Reverting, cannot use `metro` value
/close

PR #47129, opened 2024-01-02
OCPQE-18510: Optimize the test filter for cucushift on fips

liangxia:
Got the idea of how to check if FIPS is enabled in OCP from [doc](https://access.redhat.com/solutions/6525331).

Verified with several jobs, e.g, [fips-job1](https://gcsweb-qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/qe-private-deck/pr-logs/pull/openshift_release/47129/rehearse-47129-periodic-ci-openshift-openshift-tests-private-release-4.11-amd64-nightly-aws-ipi-ovn-efs-fips-f28/1742099491743338496/artifacts/aws-ipi-ovn-efs-fips-f28/cucushift-e2e/build-log.txt), [fips-job2](https://gcsweb-qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/qe-private-deck/pr-logs/pull/openshift_release/47129/rehearse-47129-periodic-ci-openshift-openshift-tests-private-release-4.16-amd64-nightly-aws-ipi-wavelength-fips-f14/1742099491890139136/artifacts/aws-ipi-wavelength-fips-f14/cucushift-e2e/build-log.txt), [non-fips-job1](https://gcsweb-qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/qe-private-deck/pr-logs/pull/openshift_release/47129/rehearse-47129-periodic-ci-openshift-openshift-tests-private-release-4.13-amd64-nightly-azure-ipi-minimal-permission-f28/1742099491655258112/artifacts/azure-ipi-minimal-permission-f28/cucushift-e2e/build-log.txt), and [non-fips-job2](https://gcsweb-qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/qe-private-deck/pr-logs/pull/openshift_release/47129/rehearse-47129-periodic-ci-openshift-openshift-tests-private-release-4.16-multi-nightly-azure-ipi-oidc-arm-f14/1742099491823030272/artifacts/azure-ipi-oidc-arm-f14/cucushift-e2e/build-log.txt).

PR #47158, opened 2024-01-03
OCPQE-18510: Remove tag 'not @fips'

liangxia:
Follow up of https://github.com/openshift/release/pull/47129

When adding new jobs/profiles, one less test tag to worry about.
The tags have done in this dynamic way are: `@destructive`, `@long-duration`, `@4.Y` (Y=10, 11, 12, ...), `@network-openshiftsdn`, `@network-ovnkubernetes`, `@singlenode` and `@fips` (WIP)


liangxia:
When adding new jobs/profiles, we should take this into account.
/cc @jianlinliu @gpei @yunjiang29 @xiaojiey

PR #47162, opened 2024-01-03
stackrox: provide a step name env value

gavin-stackrox:
stackrox/stackrox uses the step name to access artifacts to provide links to e.g. [logs](https://github.com/stackrox/stackrox/blob/cd278ec51ed0dc673db2bfc6fab64f37fdd9e33a/qa-tests-backend/scripts/lib.sh#L109). This is hardcoded to `stackrox-e2e` but the name has changed. This PR creates a step name env var so that this can be reworked in a less brittle manner. (Note: Openshift CI does not provide anything we could use.)
/cc @davdhacs @tommartensen


tommartensen (approved):
/lgtm

PR #47163, opened 2024-01-03
installer-aro: update memory limits to match upstream for unit tests

lranjbar:
Updates the unit tests to match the memory limits of the [upstream installer.
](https://github.com/openshift/release/blob/master/ci-operator/config/openshift/installer/openshift-installer-master.yaml#L240-L243)


dofinn:
@lranjbar do we need to set `aro-unit` as well anywhere? https://github.com/openshift/release/blob/master/ci-operator/config/openshift/installer/openshift-installer-master.yaml#L208-L211

PR #47187, opened 2024-01-03
rename gcp workload-identity ci jobs to oidc on 4.14+

jianlinliu:
1. rename gcp workload-identity ci jobs to oidc on 4.14+ to save job name length and keep consistent with azure.
2. removing `gcp-ipi-workload-identity`, because it is already covered by `gcp-ipi-workload-identity-rt-fips`


jianlinliu:
cc @jianli-wei to review.

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__amd64-nightly-4.14-upgrade-from-eus-4.12.yaml
```diff
@@ -76,7 +76,7 @@ tests:
     test:
     - chain: openshift-upgrade-qe-test-paused
     workflow: cucushift-installer-rehearse-baremetalds-ipi-ovn
-- as: gcp-ipi-workload-identity-rt-fips-f14
+- as: gcp-ipi-oidc-rt-fips-f14
```
jianli-wei, line 79: Shall we update the cron entry as well?

jianli-wei (changes requested):

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__amd64-nightly-4.14-upgrade-from-eus-4.12.yaml
jianlinliu: Ideally, yes, but because I did not update its frequency, it make no much difference on cron,  maybe later we can update all the crons in a separate PR.

jianlinliu:
cc  @liangxia to review.

PR #47036, opened 2023-12-21
chore(RHIDP-885): Use new AWS account to run Backstage perf&scale tests

jhutar:

PR #47203, opened 2024-01-03
fix(RHIDP-885): set correct base domain to RHDH scalability workflow to match aws-rhdh-performance cluster profile

pmacik:
Follows up to https://github.com/openshift/release/pull/47036


jhutar (approved):
/lgtm

PR #47216, opened 2024-01-03
OCPBUGS-25400: Updated arm64 and multi release controller to upgrade from nightlies instead of PreviousMinor.

jaypoulz:
The goal of this patch is to have more consistency with the amd64 release controller, as well as to avoid our upgrade jobs from repeatedly testing against the same ec/rc which increases the risk of installer+release payload incompatibility.


djslavens (approved):
/lgtm

bradmwilliams (approved):

PR #47220, opened 2024-01-03
Update servicemesh config for zstream tests

smpsgill:
Z-Stream testing on OCP 4.14 and OCP 4.15 latest for servicemesh


Simranpal:
https://github.com/openshift/release/pull/48307

PR #47221, opened 2024-01-03
Increases perfscale 120node timeout to 7h

afcollins:


jtaleric (approved):
lgtm. Looking at the current run times, we are super close to the 6hr so this change should give us some breathing room. 
image

PR #47224, opened 2024-01-03
MIXEDARCH-356: resolved the service-instance retrieval

prb112:
Resolved issue with service-instance retrieval inside the if


prb112:
Hi @pkenchap please review when you can. Thank you, Paul

PR #47228, opened 2024-01-04
OCPQE-18091: Install cluster with valid disktype and disksize specified

jianli-wei:
- add step `ipi-conf-gcp-osdisk-disksize` to insert osDisk.diskSizeGB settings into install-config
- add post-installation check step `cucushift-installer-check-gcp-disksize` accordingly
- add the workflow `cucushift-installer-rehearse-gcp-ipi-disk-type-size` and the provision chain
- add verification test `gcp-ipi-osdisk-type-size-f28`


jianli-wei:
/pj-rehearse periodic-ci-openshift-verification-tests-master-installation-nightly-4.15-gcp-ipi-osdisk-type-size-f28

jianli-wei:
/pj-rehearse periodic-ci-openshift-verification-tests-master-installation-nightly-4.15-gcp-ipi-osdisk-type-size-f28

jianli-wei:
/pj-rehearse periodic-ci-openshift-verification-tests-master-installation-nightly-4.15-gcp-ipi-osdisk-type-size-f28

jianli-wei:
/pj-rehearse periodic-ci-openshift-verification-tests-master-installation-nightly-4.15-gcp-ipi-osdisk-type-size-f28

ci-operator/step-registry/ipi/conf/gcp/osdisk/disksize/ipi-conf-gcp-osdisk-disksize-commands.sh
```diff
@@ -0,0 +1,43 @@
...
+set -o nounset
+set -o errexit
+set -o pipefail
+
+CONFIG="${SHARED_DIR}/install-config.yaml"
+PATCH="${SHARED_DIR}/install-config-patch.yaml"
+
+BASE_SIZEGB=256
+if [ -z "${COMPUTE_DISK_SIZEGB}" ]; then
+  COMPUTE_DISK_SIZEGB=$(( $BASE_SIZEGB * (RANDOM % 3 + 1) ))
```
jianlinliu, line 12: Why need a RANDOM value here?

jianli-wei: To avoid using a fixed disk size for compute and control-plane nodes, if the env vars not specified. Your suggestion?

jianlinliu: Actually I do not suggest to do many silent customization in the common gcp-ipi install workflow, once something is breaking the install, it will affect other workflows too which reference this provision chain. 

Another point is we are trying our best to follow a rule, reflecting the setting in the install workflow, so if we think inserting this step in this chain is a good choice, then maybe we should exit it directly when `COMPUTE_DISK_SIZEGB` and `CONTROL_PLANE_DISK_SIZEGB` are empty, in order to avoid introducing unexpected change on this base install workflow.

jianlinliu: > To avoid using a fixed disk size for compute and control-plane nodes

I think we should leave that to users to decide to set them via `COMPUTE_DISK_SIZEGB` and `CONTROL_PLANE_DISK_SIZEGB`

ci-operator/step-registry/ipi/conf/gcp/osdisk/disksize/ipi-conf-gcp-osdisk-disksize-commands.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+if [ -z "${COMPUTE_DISK_SIZEGB}" ] || [ -z "${CONTROL_PLANE_DISK_SIZEGB}" ]; then
```
jianlinliu, line 7: here should be use `&&` instead of `||`, right?

jianli-wei: So you suggest to move on if any of them non-empty?

jianlinliu: right

PR #47232, opened 2024-01-04
update ocm version from 0.1.67 to 0.1.72

yasun1:
The 4.6 base image is too old so the built-in ocm version is 0.1.67.  Let us upgrade it to 4.15 to improve the ocm cli version to 0.1.72.

[OCPQE-18715](https://issues.redhat.com//browse/OCPQE-18715)


yasun1:
@bear-redhat Please help review. Thanks.

PR #47267, opened 2024-01-04
Remove obsolete Stolostron multicluster-global-hub jobs

skitt:
ACM 2.6 has reached EOL, pre-2.7 jobs are no longer needed.


skitt:
Closing since the only remaining jobs are 2.6 jobs and we still need those.

PR #47186, opened 2024-01-03
OCPQE-18625: Optimize the test filter for cucushift on different arch

liangxia:
For cucushift e2e tests in Prow CI, we are using `E2E_RUN_TAGS: '@xxx and @amd64 (or @arm64, or @heterogeneous)'` to filter test cases on different arch cluster

It works, but every job/profile need this, which cost extra efforts to maintain this.

We'd like to opt out this tag and detect arch dynamically, so that we have one less tag to worry about when adding new profiles.

Thanks @lwan-wanglin for clarifying the logic of how to determine whether to run amd64/arm64/heterogeneous tests on clusters installed via using multi arch payload.

Will use follow up PRs to remove the `@amd64/@arm64/@heterogeneous` from the test profiles.

Test logs,

- multi arch payload with arm64 nodes only, 

[log1](https://gcsweb-qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/qe-private-deck/pr-logs/pull/openshift_release/47186/rehearse-47186-periodic-ci-openshift-openshift-tests-private-release-4.15-multi-nightly-aws-upi-baselinecaps-none-arm-f14/1742744203273703424/artifacts/aws-upi-baselinecaps-none-arm-f14/cucushift-e2e/build-log.txt) (`@4.15 and @arm64` are added dynamically)
```
In function: filter_test_by_arch
E2E_RUN_TAGS: '@arm64 and @aws-upi and not @console and not @storage and not @workloads and @4.15 and @arm64'
```
[log2](https://gcsweb-qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/qe-private-deck/pr-logs/pull/openshift_release/47186/rehearse-47186-periodic-ci-openshift-openshift-tests-private-release-4.15-multi-nightly-gcp-upi-private-xpn-sdn-arm-f14/1742744203332423680/artifacts/gcp-upi-private-xpn-sdn-arm-f14/cucushift-e2e/build-log.txt) (`@4.15 and @arm64` are added dynamically)
```
In function: filter_test_by_arch
E2E_RUN_TAGS: '@arm64 and @gcp-upi and @4.15 and @arm64'
```
- multi arch payload with mixure arch nodes, 

[...]

PR #47333, opened 2024-01-05
OCPQE-18625: Apply test filter optimize for arch on the other type of tests

liangxia:
Follow up of https://github.com/openshift/release/pull/47186, to apply the same optimize to destructive/longduration/ui tests.


liangxia:
The logic have been verified in https://github.com/openshift/release/pull/47186, so skip tests here.

/pj-rehearse skip

PR #47335, opened 2024-01-05
OCPQE-18625: Remove tags amd64/arm64/heterogeneous

liangxia:
Follow up of https://github.com/openshift/release/pull/47186 and https://github.com/openshift/release/pull/47333

No longer need to worry about whether should filter amd64/arm64/heterogeneous/ppc64le/... test cases when adding new jobs/profiles.


liangxia:
Rebase and fix conflicts.

PR #47367, opened 2024-01-05
PowerVS: remove extra cleanups

hamzy:
Now that other projects are sharing CI, remove some hard coded cleanup code.


mjturek:
If we see a significant downtick in deploy successes, we should reevaluate and consider adding manual cleanups back in. For now let's merge this and see how it goes.

PR #47374, opened 2024-01-06
use branched COSA for 4.14 and 4.15

dustymabe:
When comparing to the openshift-os-release-4.13.yaml file there were clearly some spots in here where we should have been using the 4.14 of 4.15 instead of latest.

This should make it so the CI doesn't pull the latest COSA which has caused test failures [1] because an iscsi test was recently added that won't pass in these older branches.

[1] https://github.com/openshift/os/pull/1414#issuecomment-1879505611

### DIFF
--- a/ci-operator/config/openshift/os/openshift-os-release-4.14.yaml
+++ b/ci-operator/config/openshift/os/openshift-os-release-4.14.yaml
@@ -1,8 +1,8 @@
 base_images:
-  coreos_coreos-assembler_latest:
+  coreos_coreos-assembler_4.14:
     name: coreos-assembler
     namespace: coreos
-    tag: latest
+    tag: "4.14"
 build_root:
   image_stream_tag:
     name: fcos-buildroot
@@ -11,7 +11,7 @@ build_root:
 images:
 - dockerfile_literal: |
     # COSA + source code
-    FROM registry.ci.openshift.org/coreos/coreos-assembler:latest
+    FROM registry.ci.openshift.org/coreos/coreos-assembler:4.14
     WORKDIR /src
     # Tell the build scripts that we have setup a directory for unprivileged build
     ENV COSA_DIR=/tmp/cosa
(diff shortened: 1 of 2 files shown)


travier (approved):
/lgtm
Thanks for catching that Dusty!

jlebon:
I think we also need to update https://github.com/openshift/release/blob/036f9e148af3c4817f53b044d3c75ccaf7e2c8eb/core-services/image-mirroring/coreos/mapping_coreos_assembler_coreos_assembler to add 4.15. (Though honestly, I'm thinking we should drop that mirroring to make things simpler. But we need to make sure we're aware of all the users first though.)

dustymabe:
> I think we also need to update https://github.com/openshift/release/blob/036f9e148af3c4817f53b044d3c75ccaf7e2c8eb/core-services/image-mirroring/coreos/mapping_coreos_assembler_coreos_assembler to add 4.15. (Though honestly, I'm thinking we should drop that mirroring to make things simpler. But we need to make sure we're aware of all the users first though.)

I'm not sure I follow.. The latest version of that file has 4.15 in it:

https://github.com/openshift/release/blob/3c09ec71e99f13210ec1b8b5673c9d8fbd40389e/core-services/image-mirroring/coreos/mapping_coreos_assembler_coreos_assembler#L2

dustymabe:
From the tide status it looks like this PR needs the `rehearsals-ack` label?

jlebon:
> > I think we also need to update [`036f9e1`/core-services/image-mirroring/coreos/mapping_coreos_assembler_coreos_assembler](https://github.com/openshift/release/blob/036f9e148af3c4817f53b044d3c75ccaf7e2c8eb/core-services/image-mirroring/coreos/mapping_coreos_assembler_coreos_assembler) to add 4.15. (Though honestly, I'm thinking we should drop that mirroring to make things simpler. But we need to make sure we're aware of all the users first though.)
> 
> I'm not sure I follow.. The latest version of that file has 4.15 in it:

Ahhh indeed, sorry for the noise.

I think you're looking for:

/pj-rehearse ack

:)

PR #47410, opened 2024-01-08
[LPINTEROP-3893] Add if statement to check if directory exists prior to archiving.

calebevans:
This PR adds an if statement around copying a log folder into the artifacts directory. Apparently, the logs folder only exists if the scenario fails.


Simranpal (approved):
/lgtm

PR #47431, opened 2024-01-09
MGMT-16330: remove parametrization for SDN tests, move them to 4.14

rccrdpccl:
`TEST_FUNC=test_networking` parametrizes network type. SDN should not work for 4.15 and breaks some tests. The aim is to move such tests to 4.14 and/or below (thoughts? maybe unnecessary cost and only 4.14 is enough?)

/cc @adriengentil @danmanor

### DIFF
--- a/ci-operator/config/openshift/assisted-service/openshift-assisted-service-master__edge.yaml
+++ b/ci-operator/config/openshift/assisted-service/openshift-assisted-service-master__edge.yaml
@@ -753,8 +753,6 @@ tests:
     env:
       ASSISTED_CONFIG: |
         STATIC_IPS=true
-        TEST_FUNC=test_networking
-      PACKET_PLAN: m3.large.x86
     workflow: assisted-baremetal
 - as: e2e-metal-assisted-bond
   optional: true
@@ -764,8 +762,6 @@ tests:
     env:
       ASSISTED_CONFIG: |
         IS_BONDED=true
-        TEST_FUNC=test_networking
-      PACKET_PLAN: m3.large.x86
     workflow: assisted-baremetal
 - as: e2e-metal-assisted-kube-api-net-suite
   optional: true
@@ -819,6 +815,30 @@ tests:
       ASSISTED_CONFIG: |
         OPENSHIFT_VERSION=4.13
         OPENSHIFT_INSTALL_RELEASE_IMAGE=""
+    workflow: assisted-baremetal
+- always_run: false
+  as: e2e-metal-assisted-static-ip-suite-4-14
+  optional: true
+  steps:
+    cluster_profile: packet-assisted
+    env:
+      ASSISTED_CONFIG: |
+        STATIC_IPS=true
+        TEST_FUNC=test_networking
+        OPENSHIFT_VERSION=4.14
+      PACKET_PLAN: m3.large.x86
+    workflow: assisted-baremetal
+- always_run: false
+  as: e2e-metal-assisted-bond-4-14
+  optional: true
+  steps:
+    cluster_profile: packet-assisted
+    env:
+      ASSISTED_CONFIG: |
+        IS_BONDED=true
+        TEST_FUNC=test_networking
+        OPENSHIFT_VERSION=4.14
+      PACKET_PLAN: m3.large.x86
     workflow: assisted-baremetal
 - as: e2e-metal-assisted-4-14
   optional: true
--- a/ci-operator/config/openshift/assisted-test-infra/openshift-assisted-test-infra-master.yaml
+++ b/ci-operator/config/openshift/assisted-test-infra/openshift-assisted-test-infra-master.yaml
@@ -589,7 +589,6 @@ tests:
     env:
       ASSISTED_CONFIG: |
         STATIC_IPS=true
-        TEST_FUNC=test_networking
       PACKET_PLAN: m3.large.x86
     workflow: assisted-baremetal
 - as: e2e-metal-assisted-bond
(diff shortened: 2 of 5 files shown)


eifrach:
> `TEST_FUNC=test_networking` parametrizes network type. SDN should not work for 4.15 and breaks some tests. The aim is to move such tests to 4.14 and/or below (thoughts? maybe unnecessary cost and only 4.14 is enough?)
> 
> /cc @adriengentil @danmanor

If it breaks 4.15, I think this should be in the code - other team/ users are using this code so it applies to all of them not just running tests

ci-operator/config/openshift/assisted-service/openshift-assisted-service-release-4.14.yaml
```diff
@@ -78,6 +78,28 @@ tests:
         MIRROR_IMAGES=true
```
danmanor, line 78: the changes should be in master branch, otherwise they will not be triggered on master PRs (almost sure)

ci-operator/config/openshift/assisted-service/openshift-assisted-service-release-4.14.yaml
```diff
@@ -78,6 +78,28 @@ tests:
...
+  optional: true
+  run_if_changed: ^(data/.*)$
+  steps:
+    cluster_profile: packet-assisted
+    env:
+      ASSISTED_CONFIG: |
+        IS_BONDED=true
+        TEST_FUNC=test_networking
+      PACKET_PLAN: m3.large.x86
+    workflow: assisted-baremetal
```
adriengentil, line 102: we don't use directly these ocp branches. The job should be defined in the `__edge` file, in the same way we define `e2e-metal-assisted-4-14` for example.

ci-operator/config/openshift/assisted-service/openshift-assisted-service-release-4.14.yaml
rccrdpccl: Adding 4.14 to the tests would prevent testing OVN and 4.15 in those scenarios, but I will duplicate them in master with 4.14 specified

ci-operator/config/openshift/assisted-service/openshift-assisted-service-master__edge.yaml
```diff
@@ -753,7 +753,6 @@ tests:
     env:
       ASSISTED_CONFIG: |
         STATIC_IPS=true
-        TEST_FUNC=test_networking
```
danmanor, line 756: Why deleting this ?

danmanor: You just want to change version no ?

ci-operator/config/openshift/assisted-service/openshift-assisted-service-release-4.14.yaml
danmanor: Seems reasonable until SDN will be applicable for 4.15 no ?

ci-operator/config/openshift/assisted-service/openshift-assisted-service-master__edge.yaml
rccrdpccl: By deleting this we will test STATIC_IPS=true but only with default networking instead of both networking type. I think it makes sense in my mind, if not please let me know why :)

danmanor: Oh so it will run `test_install` by default, which will use the default network type with the default openshift versions

danmanor: sorry

adriengentil: That means we were creating 2 clusters? one with sdn and one with ovn?
Once this PR is merged, it would be worth to remove `PACKET_PLAN`, the medium machine would then be enough to run that test.

ci-operator/config/openshift/assisted-test-infra/openshift-assisted-test-infra-master.yaml
```diff
@@ -611,7 +609,6 @@ tests:
     env:
       ASSISTED_CONFIG: |
         STATIC_IPS=true
-        TEST_FUNC=test_networking
```
adriengentil, line 614: no need to keep a periodic test?

ci-operator/config/openshift/assisted-service/openshift-assisted-service-master__edge.yaml
```diff
@@ -820,6 +818,30 @@ tests:
...
+    env:
+      ASSISTED_CONFIG: |
+        STATIC_IPS=true
+        TEST_FUNC=test_networking
+        OPENSHIFT_VERSION=4.14
+      PACKET_PLAN: m3.large.x86
+    workflow: assisted-baremetal
+- as: e2e-metal-assisted-bond-4-14
+  optional: true
+  run_if_changed: ^(data/.*)$
```
adriengentil, line 835: should we really run it each time data is updated? Or should we just keep it as a manually triggered test (`always_run: false`)?

rccrdpccl: Not sure, in my mind it would make sense to have it periodic (but indeed maybe not every time `data` is changed). This way we can catch possible regressions. Is it worth the cost though? Are manual test run for regression each time?

rccrdpccl: As discussed we will keep presubmit on-demand tests plus periodic for testing 4.14 + both network types

adriengentil (approved):

rccrdpccl:
/pj-rehearse pull-ci-openshift-assisted-service-master-edge-e2e-metal-assisted-static-ip-suite-4-14

PR #47482, opened 2024-01-10
Using kubeconfig instead of usr/passwd which should work also on ROSA

FilipB:


FilipB:
Closing, the kubeconfig approach is not working

PR #47537, opened 2024-01-11
k-nmstate: enable security scanning

mkowalski:


mkowalski:
/pj-rehearse ack

I am happy with results. Released branches fail because the `.snyk` file did not merge yet, but we set them optional anyway so it's all good.

PR #47538, opened 2024-01-11
baremetal-runtimecfg: enable security scanning

mkowalski:


mkowalski:
/pj-rehearse ack

I am happy with results. Master branch fails for a good reason (will be fixed via https://github.com/openshift/baremetal-runtimecfg/pull/297). Released branches also fail for a valid reason but the test is optional so it will not block us.

cybertron:
/lgtm
/retest-required

I can't see how this would have broken the image mirror mappings so I can only assume that's a flake.

PR #45646, opened 2023-11-14
Update code to support rosa hcp loaded upgrade

liqcui:
/hold

PR #47553, opened 2024-01-11
Adding loaded upgrade for 4.16

paigerube14:


paigerube14:
/pj-rehearse periodic-ci-openshift-qe-ocp-qe-perfscale-ci-main-aws-4.16-nightly-x86-loaded-upgrade-from-4.15-perfscale-aws-24nodes-loaded-upgrade-415to416

svetsa-rh:
@paigerube14 Can you ptal https://github.com/openshift/release/pull/45646/files - I see @liqcui has added ITERATION_MULTIPLIER_ENV: "6" 
Do we need to add it here too?

paigerube14:
> https://github.com/openshift/release/pull/45646/files

Since its passing, I'm not sure why we would decrease the number of iterations. I think if we should test as much of a "fully loaded" cluster as possible. He also might have lowered it since the jobs were starting at a much older versions, it might not be able to upgrade properly but this test is passing and should keep the iterations as high as is passing I believe

svetsa-rh:
> > https://github.com/openshift/release/pull/45646/files
> 
> Since its passing, I'm not sure why we would decrease the number of iterations. I think if we should test as much of a "fully loaded" cluster as possible. He also might have lowered it since the jobs were starting at a much older versions, it might not be able to upgrade properly but this test is passing and should keep the iterations as high as is passing I believe

I agree with this. I have no objection here. I am fine with merging this and updating should the discussion with Liquan lead to new findings.

PR #47555, opened 2024-01-11
Update cron for 4.13 and 4.14 jobs

svetsa-rh:
@paigerube14 @jtaleric @afcollins @krishvoor if you want to take a quick look.
As discussed, 4.13 and 4.14 workloads will run every 2 weeks and 120 node test will run once a month.
4.13 - scheduled to run on 1st and 15th of every month
4.14 - scheduled to run on 7th and 21st of every month
Updated conc_builds tests to be non-periodic so they can be run on need basis only.

Ref: https://issues.redhat.com/browse/PERFSCALE-2765
P.S: make update in progress - will update files here once it finishes


paigerube14:
Seems good to me

jtaleric (commented):
/lgtm

jtaleric:
Sounds good to me @svetsa-rh !

PR #47564, opened 2024-01-12
Move 5 nodes from 4.12 cluster pool into 4.14

lewisdenny:
This patch moves the following 3 nodes from the 4.12 cluster pool into the 4.14 cluster pool.

- openstack-k8s-operators-cdc-vexx-06
- openstack-k8s-operators-cdc-vexx-07
- openstack-k8s-operators-cdc-vexx-08
- openstack-k8s-operators-cdc-vexx-09
- openstack-k8s-operators-cdc-vexx-10

JIRA: [OSPRH-2709](https://issues.redhat.com//browse/OSPRH-2709)


viroel:
I think that we are OK with runnigCount as 14, we should not increase it anymore in the last patch.

PR #47582, opened 2024-01-12
Add `openshift-kni` to label sync orgs

omertuc:
Otherwise label sync [fails](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-label-sync/1744871828112805888):

`"msg":"Repo isn't inside orgs"`


donpenney (commented):
/lgtm

PR #47587, opened 2024-01-12
Add go version to verify-deps

jsafrane:
For debugging purposes, add `go version` to `verify-deps` CI step, so it's clear what go was used to compute `go mod tidy`.

It would be useful for troubleshooting failures like [this](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_csi-operator/115/pull-ci-openshift-csi-operator-master-verify-deps/1745710217695334400).


bertinatto (commented):
/lgtm

PR #47605, opened 2024-01-12
Fix issue DPTP-3609 Use CI built image for ci-scheduling-webhook

bear-redhat:


clusters/build-clusters/common_ci_scheduling_webhook/deployment.yaml
```diff
@@ -42,7 +42,7 @@ spec:
           operator: Exists
       containers:
         # Temporary hosting location
```
jupierce, line 44: Still temporary?

PR #47615, opened 2024-01-12
OPRUN-3190: Reduce OLM Periodics Frequency

awgreene:


awgreene:
/hold
@dtfranz could you lead a conversation with the rest of the team to decide on how frequently we'd like to run these tests? There seems to be some process that reduces the frequency of these periodic tests once the release number gets older. I doubt we need to run all of these every 6 hours.

dtfranz:
Superseded by #47674
/close

PR #47625, opened 2024-01-15
OCPQE-18827: Set E2E_RUN_TAGS default to null

liangxia:
We are trying to optimize the test filters so that we select those tags dynamically. In the near future, it's possible that for some profiles, we might no longer need to specify the tags statically. And then, we can omit environment variable `E2E_RUN_TAGS` when adding jobs/profiles. Then it will use the default value `@aws-ipi`, which is not expected in most of the cases. That's why we need to set the default value to null.


yapei:
ack, thanks for the updates @liangxia

yapei (approved):
/approve

PR #47633, opened 2024-01-15
add TEST_FILTERS_SUPPLEMENTARY for rosa/osd jobs

sunzhaohua2:
This is to fix https://issues.redhat.com/browse/OCPQE-18940, as OCP-25436 is import, so was designed to run on any cluster, but it is not suitable for running in rosa osd clusters
@kuiwang02 @huali9 @miyadav @shellyyang1989 PTAL

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-nightly.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-nightly.yaml
@@ -471,6 +471,7 @@ tests:
       E2E_RUN_TAGS: '@aws-ipi and @osd_ccs'
       OPENSHIFT_VERSION: "4.12"
       TEST_FILTERS: ~ChkUpgrade&;~NonPreRelease&;~Serial&;~Disruptive&;~DisconnectedOnly&;~HyperShiftMGMT&;~MicroShiftOnly&;OSD_CCS&
+      TEST_FILTERS_SUPPLEMENTARY: ~ChkUpgrade&;~DisconnectedOnly&;~HyperShiftMGMT&;~MicroShiftOnly&;OSD_CCS&
     test:
     - chain: openshift-e2e-test-qe
     workflow: osd-ccs-aws
@@ -483,6 +484,7 @@ tests:
       E2E_RUN_TAGS: '@aws-ipi and @rosa'
       OPENSHIFT_VERSION: "4.12"
       TEST_FILTERS: ~ChkUpgrade&;~NonPreRelease&;~Serial&;~Disruptive&;~DisconnectedOnly&;~HyperShiftMGMT&;~MicroShiftOnly&;ROSA&
+      TEST_FILTERS_SUPPLEMENTARY: ~ChkUpgrade&;~DisconnectedOnly&;~HyperShiftMGMT&;~MicroShiftOnly&;ROSA&
     test:
     - chain: openshift-e2e-test-qe
     workflow: rosa-aws-sts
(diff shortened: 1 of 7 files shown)


ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__amd64-nightly.yaml
```diff
@@ -898,6 +898,7 @@ tests:
       E2E_RUN_TAGS: '@aws-ipi and @osd_ccs'
       OPENSHIFT_VERSION: "4.16"
       TEST_FILTERS: ~ChkUpgrade&;~NonPreRelease&;~Serial&;~Disruptive&;~DisconnectedOnly&;~HyperShiftMGMT&;~MicroShiftOnly&;OSD_CCS&
+      TEST_FILTERS_SUPPLEMENTARY: ~ChkUpgrade&;~DisconnectedOnly&;~HyperShiftMGMT&;~MicroShiftOnly&;OSD_CCS&
```
huali9, line 901: I found these conditions have been covered in the `TEST_FILTERS`, not sure why still need this `TEST_FILTERS_SUPPLEMENTARY`?

shellyyang1989:
Is it possible to skip the test-case when the cluster is rosa/osd at the case level instead of job level?

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__amd64-nightly.yaml
huali9: got the meaning after discussion with Zhaohua in slack, so no question from me. LGTM

sunzhaohua2:
> Is it possible to skip the test-case when the cluster is rosa/osd at the case level instead of job level?

Talked with @kuiwang02 , we have method to skip the case on rosa, not sure how to skip on osd

shellyyang1989:
> > Is it possible to skip the test-case when the cluster is rosa/osd at the case level instead of job level?
> 
> Talked with @kuiwang02 , we have method to skip the case on rosa, not sure how to skip on osd

So it's hard to determine whether a cluster is OSD or not?

sunzhaohua2:
> So it's hard to determine whether a cluster is OSD or not?

Yes, now the repo openshift-test-private only contains the rosa package, there is no ocm package. need use command to check `ocm get cluster <id>|jq -r .product.id`

kuiwang02:
> > So it's hard to determine whether a cluster is OSD or not?
> 
> Yes, now the repo openshift-test-private only contains the rosa package, there is no ocm package. need use command to check `ocm get cluster <id>|jq -r .product.id`

no. for ocp case, there is no ocm. ocm case is running in different env.

if you want to tell OSD cluster, need to follow similar way that tell ROSA cluster. you need to talk to Yanming Sun who create osd-ccs-aws workflow. maybe he could talk to you how to tell it, like IsROSA()

frankly label is also good, same to openshift-extended-test step.

miyadav:
LGTM too , based on comments.

kuiwang02:
i am ok.

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__amd64-nightly.yaml
```diff
@@ -507,6 +510,7 @@ tests:
       E2E_RUN_TAGS: '@aws-ipi and @disconnected and @rosa'
       OPENSHIFT_VERSION: "4.12"
       TEST_FILTERS: ~ChkUpgrade&;~NonPreRelease&;~Serial&;~Disruptive&;~ConnectedOnly&;~HyperShiftMGMT&;~MicroShiftOnly&;ROSA&
+      TEST_FILTERS_SUPPLEMENTARY: ~ChkUpgrade&;~DisconnectedOnly&;~HyperShiftMGMT&;~MicroShiftOnly&;ROSA&
```
liangxia, line 513: `TEST_FILTERS` use `~ConnectedOnly&;`, while `TEST_FILTERS_SUPPLEMENTARY` use `~DisconnectedOnly&;`. Is it expected ?

liangxia:
Just found that we only have one test case in `TEST_FILTERS_SUPPLEMENTARY`.  Perhaps we can use `TEST_FILTERS_SUPPLEMENTARY:  xxxIDxxx`, which is easier to understand IMO.
The current one looks good to me. Up to you to decide.

sunzhaohua2:
> Just found that we only have one test case in `TEST_FILTERS_SUPPLEMENTARY`. Perhaps we can use `TEST_FILTERS_SUPPLEMENTARY: xxxIDxxx`, which is easier to understand IMO. The current one looks good to me. Up to you to decide.

Thank you, let's keep it like this, so that if there are similar cases in future this can be used without modification.

PR #47668, opened 2024-01-15
Use Openshift local cluster profile

praveenkumar:


ci-operator/config/crc-org/crc/crc-org-crc-main.yaml
```diff
@@ -43,15 +43,15 @@ resources:
...
     workflow: code-ready-crc-e2e
 - as: integration-crc
   steps:
-    cluster_profile: gcp-3
+    cluster_profile: osl-gcp
     workflow: code-ready-crc-integration
 - as: e2e-microshift-crc
   steps:
-    cluster_profile: gcp-3
+    cluster_profile: osl-gcp
```
gbraad, line 54: `osl` stands for OpenShift Local?
Please, nly use this for technical reasons, as this is not an official abbreviation.

gbraad (approved):
/lgtm

Note: make sure `osl` is NOT used as an official acronym.

ci-operator/config/crc-org/crc/crc-org-crc-main.yaml
praveenkumar: yes, it is not offical abbreviation but need to use a short form for technical reason.

cfergeau: Why not `crc`?

PR #47674, opened 2024-01-15
OPRUN-3190: Reduce OLM Periodics Frequency

dtfranz:
Supersedes #47615


tmshort:
/lgtm

Eventually, we may want to cut down the older branches to longer period (e.g. weekly).

dtfranz:
> /lgtm
> 
> Eventually, we may want to cut down the older branches to longer period (e.g. weekly).

From @awgreene's comment [here](https://github.com/openshift/release/pull/47615#issuecomment-1889927354), it seems that happens automatically as the release's age increases. We can also probably do it manually.

PR #47701, opened 2024-01-16
GitOps | Remove unsupported branches and jobs

varshab1210:


anandrkskd:
Hey @varshab1210 can you add the info about which all jobs are getting removed as part of this PR in the description. It will be easier for us to get back in case we need to some time in future to track it.

otherwise, 
/lgtm

PR #47730, opened 2024-01-16
stackrox: test on ocp4.15 rc 2

davdhacs:
The OCP4.15 pod delays are fixed in 4.15rc2. We need to move our testing to use the ocp/candidate-4.15 to get the rc.2. The ocp-dev-preview/4.15 is still at the last preview/ec and not the rc's.
There are other test failures, but from a high-level scan those are product issues to be resolved in the stackrox repo instead of openshift-release.


BradLugo (approved):
/lgtm

PR #16861, opened 2021-03-16
OVN: add SDN migration tests

pliurh:
Add tests for the SDN migration and rollback. The workflow is:
1. deploy the OCP cluster with openshift-sdn.
2. migrate to ovn-kubernetes (Or migrate then rollback).
3. run e2e tests.

PR #35950, opened 2023-02-02
SDN-3681: Fix SDN migration jobs

pliurh:
Remove egressip configuration after finishing the egressip validation.

In AWS, the egressip might result in it appearing in the node status
as an additional internal IP. It will break the nodeip service tests.
Since the tests will use the egressip as the nodeip mistakenly.

PR #47733, opened 2024-01-16
update TEST_SKIPS for nw migration presubmits

jluhrsen:
Somehow the TEST_SKIPS env was not brought over when this job was added as a presubmit for ovnk and cno. The alerts are firing (and we expect them too on a network migration from sdn to ovn) and this test case fails nearly 100% of the time. the skips were original to the job and configured from day 1. here [0] is the original PR that brought this job in.

[0] https://github.com/openshift/release/pull/16861


jcaamano:
The test skips were removed by @pliurh here https://github.com/openshift/release/pull/35950
@pliurh PTAL

PR #47758, opened 2024-01-17
feat: set dependabot as trusted_apps for kubevirt repos

ksimon1:
add configuration for ssp-operator, vm-console-proxy, tekton-tasks repos to add dependabot as trusted app


0xFelix (approved):
/approve
/lgtm

Looks good to me!

PR #47759, opened 2024-01-17
SPLAT-1394: [4.16] remove alibaba tests

jianli-wei:
- [OCPSTRAT-1042 Remove IPI/UPI support of Alibaba Cloud from Openshift 4.16](https://issues.redhat.com/browse/OCPSTRAT-1042)
- Remove the tests on alibabacloud from 4.16


gpei:
cc @liangxia for review, it's for the coming Alibaba support removal in 4.16 - https://issues.redhat.com/browse/OCPSTRAT-1042, Alibaba is always in TP, so we don't have upgrade test.

PR #47760, opened 2024-01-17
[kubevirt-tekton-tasks] feat: disable external plugins kubevirt-tekton-tasks

ksimon1:
we are already using kubevirt-bot and having two bots for e.g. cherry-picking creates duplicite PRs.


0xFelix (approved):
/approve
/lgtm

PR #47792, opened 2024-01-17
Add support for sso testing on z-stream

smpsgill:


ci-operator/config/rhbk/interop-ocp-ci/rhbk-interop-ocp-ci-main__rhsso-product-ocp4.15-lp-interop.yaml
ascerra, line 1: 4.15 is still pre-release, these updates should be for files for 4.14 and 4.13.

This file will become z-stream once 4.16 is released.

ci-operator/config/rhbk/interop-ocp-ci/rhbk-interop-ocp-ci-main__rhsso-product-ocp4.14-lp-interop.yaml
```diff
@@ -0,0 +1,71 @@
...
+              {"step": "openshift-*", "failure_type": "all", "classification": "Other", "jira_project": "LPTOCPCI", "group": {"name": "cluster", "priority": 2}, "jira_additional_labels": ["!default"]},
+              {"step": "install-operators", "failure_type": "all", "classification": "Infrastructure Provisioning - Operators", "group": {"name": "lp-tests", "priority": 1}, "jira_additional_labels": ["!default"]},
+              {"step": "keycloak-qe-rhsso-tests", "failure_type": "pod_failure", "classification": "Test Execution", "jira_project": "LPTOCPCI",  "group": {"name": "lp-tests", "priority": 2}, "jira_additional_labels": ["!default","interop-tests"]},
+              {"step": "keycloak-qe-rhsso-tests", "failure_type": "test_failure", "classification": "Test Failure", "jira_project": "RHSSO", "group": {"name": "lp-tests", "priority": 2},"jira_additional_labels": ["rhsso","!default","interop-tests"],"jira_assignee":"ikhomyn@redhat.com"}
+            ]
+        }
+      FIREWATCH_DEFAULT_JIRA_ADDITIONAL_LABELS: '["4.14-lp","self-managed-lp","rhsso-lp","z-stream"]'
```
ascerra, line 53: on (12/21/23) in our canvas meeting notes we discussed the need for a unique identifier.

I don't believe we captured all resolutions there but to recap I think we agreed that unuqiue identifiers for zstream should be avoided and that we will need to rely on knowledge of the current OCP pre-release version to understand which versions are our zstream versions.

This philopsophy would lead to removing the label `z-stream` here as we would know that we need to filter on 4.14 as our N-1 version. Our Jira filters could be zstream specific if the watchers needed it, essentially search for all 4.14 and call it filter z-stream N-1. 

Removing z-stream here helps us when all of the current pre-release 4.15 jobs become the zstream n-1 version. If we add this label to z-stream now then when 4.15 becomes zstream we will need to add this label to every file manually.

Simranpal: Yes I agree, it makes sense to avoid adding the label for z-stream here

ascerra (approved):

PR #47799, opened 2024-01-18
Serverless: update Interop CI config to new style

mgencur:
Use base images from 4.14 even for 4.15 cluster for two reasons:
* The automatic "replacer" will enforce 4.14 because that version is
defined in the target dockerfile.
* The operator-registry image includes "opm" binary which works only
with RHEL9 for the base image and we currently use RHEL8 (more details in https://issues.redhat.com/browse/SRVCOM-2901)
This is a temporary workaround. We should use proper 4.15 variant as indicated in the JIRA once we target 4.15 with the product.

### DIFF
--- a/ci-operator/config/openshift-knative/serverless-operator/openshift-knative-serverless-operator-main__ocp4.15-lp-interop.yaml
+++ b/ci-operator/config/openshift-knative/serverless-operator/openshift-knative-serverless-operator-main__ocp4.15-lp-interop.yaml
@@ -51,7 +51,7 @@ images:
     ocp_4.14_base:
       as:
       - registry.ci.openshift.org/ocp/4.14:base
-    origin_4.15_operator-registry:
+    origin_4.14_operator-registry:
       as:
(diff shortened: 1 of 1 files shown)


mgencur:
/pj-rehearse periodic-ci-openshift-knative-serverless-operator-main-ocp4.15-lp-interop-operator-e2e-interop-aws-ocp415

dfrazzette:
If using base images from 4.14 for the 4.15 cluster is a valid configuration, then we can proceed with merge once the rehearsal job passes.

mgencur:
@dfrazzette Yes. I think it's a valid configuration to have base images from 4.14 temporarily as long as it works (and it does). After all, there's not much difference and the final product images are built differently. Not in CI, and with different base images anyway.

liswang89:
@mgencur @dfrazzette Thanks for the feedback. It appears the remaining issue is in the last test suite, 'eventing-kafka-broker.' While we have the option to close the current ticket, I'm concerned we won't be allowed to merge the PR with existing failures. I hope we can get the issue resolved soon.

PR #47813, opened 2024-01-18
skip must-gather for successful test

kpouget:
Skip must-gather, gather-extra and gather-audit-logs if the test is succesful as collecting logs is taking longer time and resulting in timeouts.


ccamacho (approved):
LGTM

PR #47819, opened 2024-01-18
Fix a (harmless) bug and an incorrect comment in ovn-sdn-migration-commands.sh

danwinship:
`success_count` was being initialized in the outer script but then referenced from a nested script where it wasn't set yet. The context results in it not really mattering but it leaves a spurious error in the logs:
```
  bash: line 1: [: -eq: unary operator expected
```

Also, fix a comment that didn't get updated for the new timeouts. (And some typos.)

/assign @jluhrsen


danwinship:
/pj-rehearse
(want to make sure it really works)

jluhrsen:
thank you
/lgtm
/pj-rehearse ack

PR #47849, opened 2024-01-19
OCPQE-18827: Add more info for platform none/external/unknown

liangxia:
For platform that show as `none`, `external` and other unexpected values ( other than aws / azure / baremetal / gcp / ibmcloud / nutanix / vsphere ), I'd like to have more info to double confirm the dynamic test case selection logic.

Test with [log](https://gcsweb-qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/qe-private-deck/pr-logs/pull/openshift_release/47849/rehearse-47849-periodic-ci-openshift-openshift-tests-private-release-4.13-amd64-nightly-vsphere-upi-platform-none-f14/1748299417548492800/artifacts/vsphere-upi-platform-none-f14/cucushift-e2e/build-log.txt),
```
Expected, got platform as 'external'
...
platform: External
platformStatus:
  external: {}
  type: External
```

### DIFF
--- a/ci-operator/step-registry/cucushift/e2e/cucushift-e2e-commands.sh
+++ b/ci-operator/step-registry/cucushift/e2e/cucushift-e2e-commands.sh
@@ -69,13 +69,16 @@ function filter_test_by_platform() {
         ipixupi='ipi'
     fi
     platform="$(oc get infrastructure cluster -o yaml | yq '.status.platform' | tr 'A-Z' 'a-z')"
+    extrainfoCmd="oc get infrastructure cluster -o yaml | yq '.status'"
     if [[ -n "$platform" ]] ; then
         case "$platform" in
             none)
                 export E2E_RUN_TAGS="@baremetal-upi and ${E2E_RUN_TAGS}"
+                eval "$extrainfoCmd"
                 ;;
             external)
                 echo "Expected, got platform as '$platform'"
+                eval "$extrainfoCmd"
                 ;;
             alibabacloud)
                 export E2E_RUN_TAGS="@alicloud-${ipixupi} and ${E2E_RUN_TAGS}"
@@ -88,6 +91,7 @@ function filter_test_by_platform() {
                 ;;
             *)
                 echo "Unexpected, got platform as '$platform'"
+                eval "$extrainfoCmd"
                 ;;
         esac
     fi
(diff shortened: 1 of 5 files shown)


ci-operator/step-registry/cucushift/e2e/cucushift-e2e-commands.sh
```diff
@@ -73,9 +73,11 @@ function filter_test_by_platform() {
         case "$platform" in
             none)
                 export E2E_RUN_TAGS="@baremetal-upi and ${E2E_RUN_TAGS}"
+                oc get infrastructure cluster -o yaml | yq '.status.platform'
```
JianLi-RH, line 76: Why not output platform before if block? I don't think it can take more effort but it can reduces amount of code.

liangxia: Several known platforms does not need this info.
And in the long term, only the `*` one will need this line of code.

JianLi-RH: Yes, then if there is only one place output platform, it is more maintainable.

liangxia: Moving the code before if block, means all platforms will output this info, which is unexpected, because for those known platforms, I don't want this info.

JianLi-RH: OK, I don't have more comments now, let's wait others to review.

liangxia:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-nightly-vsphere-upi-platform-external-f28 periodic-ci-openshift-openshift-tests-private-release-4.13-amd64-nightly-vsphere-upi-platform-none-f14

ci-operator/step-registry/cucushift/e2e/cucushift-e2e-commands.sh
JianLi-RH: Alright, this is another way to make our script simple.

liangxia:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-nightly-vsphere-upi-platform-external-f28 periodic-ci-openshift-openshift-tests-private-release-4.13-amd64-nightly-vsphere-upi-platform-none-f14

PR #47878, opened 2024-01-19
olm downstreaming: set $GOROOT

stevekuznetsov:


grokspawn:
ref: https://github.com/golang/go/issues/61921
 ... watching rehearsal on tenterhooks

PR #47902, opened 2024-01-22
[PERFSCALE-2796]: Allow passing e2e repo in tests

venkataanil:
Currently we can't test e2e or kube-burner changes without them part of main releases.
This change allow user to test e2e or kube-burner on their branches. User can create a release on their branches and pass that as a test env variable.


venkataanil:
User can directly change the repo in the shell script instead of passing that as an environment variable.

PR #47907, opened 2024-01-22
feat: add remotesecrets resources to appstudio-gather

rsoaresd:
Add remotesecrets resources to appstudio-gather


flacatus:
ok-to-test

PR #47914, opened 2024-01-22
Cleanup openshift-docs config - base_images is not required

aireilly:
Cleanup openshift-docs config - base_images is not required


gabriel-rh (approved):
LGTM

PR #47930, opened 2024-01-22
hypershift: initial image should match release

sjenning:
hypershift `main` presub is currently do a y-stream upgrade instead of a z-stream upgrade.

This is problematic for the NodePool upgrade tests.


sjenning:
confirmed `release:initial` is 4.16 ci stream now :+1: 

/pj-rehearse ack
/hold cancel

PR #47942, opened 2024-01-23
remove deprecated StackRox Scanner versions

RTann:
I kept 3.74 around because I'm not sure if we still support it or not... Otherwise just kept 4.2 and 4.3.

/cc @BradLugo @gavin-stackrox


BradLugo:
@RTann, I think we should match the versions for stackrox/stackrox that's configured in OSCI, i.e., only delete scanner 2.26 and 2.27 (my understanding is stackrox 3.74 uses scanner 2.28 and 4.0 uses 2.29)

RTann:
running into merge conflicts I don't want to deal with so I will replace this with a different PR

RTann:
https://github.com/openshift/release/pull/47992

PR #47947, opened 2024-01-23
Make the aws-ovn-arm64 install job blocking

sdodson:
We have only an upgrade job as blocking currently, we want an install job to block too so that we catch install time defects in this branch


core-services/release-controller/_releases/release-ocp-4.16-arm64.json
```diff
@@ -24,7 +24,6 @@
       }
     },
     "aws-ovn-arm64": {
-      "optional": true,
       "prowJob": {
```
jeffdyoung, line 27: @sdodson can you add `"maxRetries": 3,` here?

PR #47963, opened 2024-01-23
AWS IPI install: add tags for jobs resources on AWS

sshnaidm:
Add common tag for all IPI jobs to separate Prow CI jobs resources from any other on AWS account (jobsource: prowci)
Add tags which will indicate build ID and Prow job name in resources


sshnaidm:
done in https://github.com/openshift/release/pull/47967

PR #47974, opened 2024-01-23
telco: Add myself to approvers in repos with AWS managed jobs

sshnaidm:
Because of various changes of AWS jobs and their config, I suggest to add myself to the approvers in all repos which use AWS IPI jobs.


donpenney:
These files being modified are all generated based on the OWNERS files from those repos, aren't they? So these changes you're trying to make would be overridden, I think.

sshnaidm:
> These files being modified are all generated based on the OWNERS files from those repos, aren't they? So these changes you're trying to make would be overridden, I think.

Let me check this.

PR #47989, opened 2024-01-23
remove periodic jobs for cluster autoscaler operator

elmiko:
after an internal review from the cloud infrastructure team, we do not believe these tests should be performed on a periodic basis and are removing them. This operator does not have platform specific behavior and its functionality is also transitively tested through our cluster autoscaler testing.


droslean:
@elmiko You need also to delete:
https://github.com/openshift/release/blob/master/core-services/prow/02_config/_plugins.yaml#L403-L406

elmiko:
updated with suggestion

elmiko:
rebased

elmiko:
rebased

PR #48002, opened 2024-01-24
remove most OpenShiftSDN jobs from CNO for 4.15+

jluhrsen:
JIRA: https://issues.redhat.com/browse/SDN-4410

in 4.15, OpenShiftSDN is no longer accepted as a valid CNI for new cluster installations. There is less of a reason to run these jobs now.

the remaining sdn jobs will be e2e-aws-sdn-upgrade and the job to test going from sdn to ovn via live migration and the job to rollback to sdn after live migration.


ci-operator/config/openshift/cluster-network-operator/openshift-cluster-network-operator-master.yaml
```diff
@@ -256,39 +233,6 @@ tests:
     - ref: ovn-sdn-migration-feature-migration-validate-multicast
     - ref: openshift-e2e-test
     workflow: openshift-e2e-aws-sdn
-- as: e2e-aws-sdn-network-migration-rollback
```
danwinship, line 259: So we're keeping sdn-to-ovn "old" migration, and sdn-to-ovn live migration, and sdn-to-ovn live migration rollback, but _not_ sdn-to-ovn "old" migration rollback? Any reason? It seems inconsistent...

ci-operator/config/openshift/cluster-network-operator/openshift-cluster-network-operator-master.yaml
```diff
@@ -339,20 +283,6 @@ tests:
     - chain: baremetalds-ipi-test
     workflow: baremetalds-e2e-ovn-ipv6
   timeout: 5h0m0s
-- as: e2e-network-mtu-migration-sdn-ipv4
```
danwinship, line 342: This is redundant with the ovn migration anyway isn't it? (Since sdn-to-ovn requires decreasing the MTU for Geneve.)

ci-operator/config/openshift/cluster-network-operator/openshift-cluster-network-operator-master.yaml
```diff
@@ -81,12 +75,6 @@ tests:
     env:
       OVERRIDE_OPENSHIFT_SDN_DEPRECATION: "true"
     workflow: openshift-upgrade-aws-sdn
-- as: e2e-aws-sdn-multi
```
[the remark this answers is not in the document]
jluhrsen: curious to why it's important to keep this job for 4.16? but yes, we do have a periodic.

ci-operator/config/openshift/cluster-network-operator/openshift-cluster-network-operator-master.yaml
jluhrsen: my intention was to remove as many jobs that do a fresh install with OpenShiftSDN. I missed removing
sdn-to-ovn "old" migration rollback. will update this PR for that.

ci-operator/config/openshift/cluster-network-operator/openshift-cluster-network-operator-master.yaml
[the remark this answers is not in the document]
danwinship: In theory, there are still customers using multitenant mode. And multitenant mode occasionally breaks in new releases and needs an update to https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/openshift-sdn/004-multitenant.yaml. (OK, I guess it hasn't needed an update since 4.10... but it still could end up needing an update in 4.16 if someone adds a new core operator.)

ci-operator/config/openshift/cluster-network-operator/openshift-cluster-network-operator-master.yaml
danwinship: I don't know the answer for sure, but I think it is. Anyway, redundant or not, it seems fine to kill.

danwinship:
/approve
but you should remove _everything_ (ie `e2e-aws-sdn-upgrade` and `e2e-aws-live-migration-sdn-ovn` too) from `openshift-cluster-network-operator-release-4.17.yaml` since they (eventually) won't work there at all.

danwinship:
though I guess we'll have to remove them from the "master" file at that point too, so removing them from 4.17 could wait for that
/lgtm
/hold
feel free to update 4.17 or `/hold cancel` to merge as is

jluhrsen:
> though I guess we'll have to remove them from the "master" file at that point too, so removing them from 4.17 could wait for that /lgtm /hold feel free to update 4.17 or `/hold cancel` to merge as is

good point. I had to fix merge conflicts and I also updated 4.17 to not have any jobs that would initially install
with OpenShiftSDN.

@danwinship, PTAL

PR #48019, opened 2024-01-24
bump must-gather to rhel9

nunnatsa:


soltysh (changes requested):
/hold
Those changes are not necessary. must-gather builder is explicitly configured in https://github.com/openshift/must-gather/blob/master/.ci-operator.yaml so once we land https://github.com/openshift/must-gather/pull/402 it will automatically migrate to this file as well.

PR #48031, opened 2024-01-24
drop vsphere-windows jobs in CNO and OVN-K

danwinship:
In `openshift/ovn-kubernetes` and `openshift/cluster-network-operator`, we currently run a basic hybrid overlay test job on AWS, and a hybrid-overlay-with-custom-vxlan-port job on vsphere. But the vsphere jobs fail all the time due to vsphere infra flakes.

This makes the vsphere jobs optional/not-always-run, and makes the AWS jobs use a custom vxlan port.

I was originally going to make the vsphere jobs _not_ use a custom vxlan port, but there is no workflow for that...

Also, the ovn-kubernetes version of the aws job does "install a cluster and then enable hybrid overlay" (as opposed to the CNO AWS job and both vsphere jobs, which do "install a cluster with hybrid overlay enabled from the start") and there was no workflow to do "enable hybrid overlay post install with a custom port" and rather than adding one I just fixed the existing workflows to let you specify whether to use a custom port via an environment variable rather than needing a separate flow. I did not bother to clean up existing chains/workflows/jobs to use the new feature though because then I'd have to figure out which of the vsphere jobs actually cared about custom vxlan port and which ones didn't...

/assign @JacobTanenbaum @jluhrsen

### DIFF
--- a/ci-operator/config/openshift/cluster-network-operator/openshift-cluster-network-operator-master.yaml
+++ b/ci-operator/config/openshift/cluster-network-operator/openshift-cluster-network-operator-master.yaml
@@ -130,7 +130,7 @@ tests:
   steps:
     cluster_profile: aws
     env:
-      WINDOWS_SERVER_VERSION: "2019"
+      CUSTOM_VXLAN_PORT: "true"
     test:
     - ref: windows-e2e-operator-test
     workflow: ipi-aws-ovn-hybrid
@@ -178,7 +178,9 @@ tests:
   steps:
     cluster_profile: gcp
     workflow: openshift-e2e-gcp-ovn
-- as: e2e-vsphere-ovn-windows
+- always_run: false
+  as: e2e-vsphere-ovn-windows
+  optional: true
   steps:
     cluster_profile: vsphere-2
     test:
--- a/ci-operator/config/openshift/cluster-network-operator/openshift-cluster-network-operator-release-4.12.yaml
+++ b/ci-operator/config/openshift/cluster-network-operator/openshift-cluster-network-operator-release-4.12.yaml
@@ -115,7 +115,7 @@ tests:
   steps:
     cluster_profile: aws
     env:
-      WINDOWS_SERVER_VERSION: "2019"
+      CUSTOM_VXLAN_PORT: "true"
     test:
     - ref: windows-e2e-operator-test
     workflow: ipi-aws-ovn-hybrid
@@ -145,7 +145,9 @@ tests:
   steps:
     cluster_profile: gcp-3
     workflow: openshift-e2e-gcp-ovn
-- as: e2e-vsphere-ovn-windows
+- always_run: false
+  as: e2e-vsphere-ovn-windows
+  optional: true
   steps:
     cluster_profile: vsphere-2
     test:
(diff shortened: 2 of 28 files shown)


jluhrsen:
no problem if this is the route we want to go. I had some plans [here](https://docs.google.com/document/d/1COReXStxMMgeENWLFl6CPbVQhnQk5wg83O4T0Jl7g4Q/edit#bookmark=id.rrzcn2amyy1m) to disable the aws-ovn-windows and just fight
(fix the flakes) to keep vSphere. But if we don't care to have the specific combo of vSphere+windows then this should
be fine. I don't know who would know if we care or not though.

JacobTanenbaum:
@jluhrsen 
I don't think there is a reason to test specifically on vSphere. We need to ensure that the custom port code is tested, which is a feature that vSphere specifically needed. It would be nice to test it on vSphere but given the flakes on vSphere I think it does make sense to move that job over to aws.

jluhrsen:
why didn't the rehearsals run?

/pj-rehearse

danwinship:
They don't automatically re-run, and I've had problems in the past if you say `/pj-rehearse` before the other tests run so I hadn't said it yet... (Maybe it's only a problem if you say it and then the validation tests fail?)

danwinship:
/pj-rehearse ack

Had to remove `WINDOWS_SERVER_VERSION: "2019"` because the older version of Windows server doesn't support the custom vxlan port.

That was put there as part of WINC-1037 to make sure there were some jobs testing 2019 and some testing 2022, but for our repos we don't really need to care about the older version.

jluhrsen:
4.10 and 4.11 rehearsals failed. I did not dig very deep. do we need to figure those out?

danwinship:
/pj-rehearse

```
error obtaining aws interface object: unable to get latest Windows AMI: found zero images matching given filter: {
  Name: "name",
  Values: ["Windows_Server-2019-English-Full-ContainersLatest-????.??.??"]
}
```

that doesn't seem like it should be version-specific? maybe just a flake? let's see if they fail again...

danwinship:
/pj-rehearse pull-ci-openshift-cluster-network-operator-release-4.10-e2e-aws-ovn-windows /pj-rehearse pull-ci-openshift-cluster-network-operator-release-4.10-e2e-aws-ovn-windows

danwinship:
/pj-rehearse pull-ci-openshift-cluster-network-operator-release-4.10-e2e-aws-ovn-windows pull-ci-openshift-cluster-network-operator-release-4.11-e2e-aws-ovn-windows

danwinship:
/pj-rehearse ack

`e2e-aws-ovn-windows` seems to be broken in `release-4.10`; the AMI name was updated in 4.11 and later but not in 4.10.

jluhrsen:
> /pj-rehearse ack
> 
> `e2e-aws-ovn-windows` seems to be broken (due to `windows-machine-config-operator`) in `release-4.10`; the AMI name was updated in 4.11 and later but not in 4.10.

/lgtm

think we should just remove the 4.10 job entirely? if so, I'll throw up a PR to do that. LMK

danwinship:
> think we should just remove the 4.10 job entirely?

:man_shrugging:

PR #48061, opened 2024-01-25
OSASINFRA-3357: openstack: CAPI jobs

pierreprinetti:
Add an optional OpenStack CAPI job in Installer, and a periodic.

These jobs are to be removed once the CAPI feature gate is removed and the CAPI flow becomes the default.

### DIFF
--- a/ci-operator/config/openshift/installer/openshift-installer-master.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-master.yaml
@@ -541,6 +541,19 @@ tests:
       TEST_SUITE: experimental/reliability/minimal
     workflow: openshift-e2e-openstack-ipi
   timeout: 6h0m0s
+- as: e2e-openstack-ovn-capi
+  optional: true
+  run_if_changed: openstack|(^data/.*/rhcos\.json)
+  steps:
+    cluster_profile: openstack-vexxhost
+    env:
+      BASE_DOMAIN: shiftstack.devcluster.openshift.com
+      CONFIG_TYPE: minimal
+      FEATURE_GATES: ClusterAPIInstall=true
+      FEATURE_SET: CustomNoUpgrade
+      TEST_SUITE: experimental/reliability/minimal
+    workflow: openshift-e2e-openstack-ipi
+  timeout: 6h0m0s
 - as: e2e-openstack-upi
   optional: true
   run_if_changed: (^images/openstack/)|(^docs/user/openstack/install_upi\.md$)|(^upi/openstack/)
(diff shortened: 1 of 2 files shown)


pierreprinetti:
/pj-rehearse ack
The jobs seem to be doing what they're asked to exactly.

ci-operator/config/shiftstack/shiftstack-ci/shiftstack-shiftstack-ci-main__periodic-4.16.yaml
```diff
@@ -107,6 +107,16 @@ tests:
       ROTATE_CLOUD_CREDENTIALS: "true"
       TEST_SUITE: openshift/conformance/parallel
     workflow: openshift-e2e-openstack-ipi
+- as: e2e-openstack-ovn-parallel-capi
+  minimum_interval: 24h
```
MaysaMacedo, line 111: isn't this too frequent? Every 2 or 3 days should be ok specially as is tech preview

MaysaMacedo (commented):
Are those jobs testing integration of CAPO in CAPI?

ci-operator/config/shiftstack/shiftstack-ci/shiftstack-shiftstack-ci-main__periodic-4.16.yaml
pierreprinetti: Why do you say "tech preview"? As far as I know, this is 4.16 GA material.

However, given that we already know it doesn't work yet, I am removing the periodic job and only leaving the presubmit. We may switch all our periodics to CAPI when we have a successful run.

ci-operator/config/openshift/installer/openshift-installer-master.yaml
```diff
@@ -541,6 +541,19 @@ tests:
       TEST_SUITE: experimental/reliability/minimal
     workflow: openshift-e2e-openstack-ipi
   timeout: 6h0m0s
+- as: e2e-openstack-ovn-capi
```
mdbooth, line 544: Does this job need `ovn` in the name? Seems confusing as it's simply the default.

pierreprinetti: It does, that's just an OpenShift-wide convention semi-enforced with automation... which means that if you don't add `ovn`, something will break somewhere and you won't know why

pierreprinetti:
/close
superseded by https://github.com/openshift/release/pull/48135

PR #48063, opened 2024-01-25
Don't use root image for SDN

jcaamano:


jcaamano:
So it looks like ci-operator requires a root image build even if it is not used.

PR #48071, opened 2024-01-25
openshift-loki: Use loki v2.9.4 in operator release-5.6

periklis:
Ref:
- [LOG-5011](https://issues.redhat.com//browse/LOG-5011)

/cc @xperimental @JoaoBraveCoding

### DIFF
--- a/ci-operator/config/openshift/loki/openshift-loki-release-5.6.yaml
+++ b/ci-operator/config/openshift/loki/openshift-loki-release-5.6.yaml
@@ -4,7 +4,7 @@ base_images:
     namespace: ocp
     tag: base
   loki:
-    name: v2.9.3
+    name: v2.9.4
     namespace: logging
     tag: loki
   lokistack-gateway:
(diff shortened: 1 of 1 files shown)


ci-operator/config/openshift/loki/openshift-loki-release-5.6.yaml
```diff
@@ -33,7 +33,7 @@ operator:
   substitutions:
   - pullspec: quay.io/openshift-logging/loki-operator:v0.0.1
     with: pipeline:loki-operator
-  - pullspec: quay.io/openshift-logging/loki:v2.9.3
+  - pullspec: quay.io/openshift-logging/loki:v2.9.4
```
xperimental, line 36: I realized after the lgtm that this won't work ... but I have an idea: What if we always have the "last two pullspecs" in the configuration file? So, in this case that would be 2.9.3 and 2.9.4. That way both the old and new configuration can build. And in the next update we just increment both (e.g. 2.9.4 and 2.9.5).

```suggestion
  - pullspec: quay.io/openshift-logging/loki:v2.9.3
    with: pipeline:loki
  - pullspec: quay.io/openshift-logging/loki:v2.9.4
```

PR #48072, opened 2024-01-25
StackRox: Set COLLECTION_METHOD=core_bpf for scanner v4 tests

mclasmeier:
Within StackRox (aka RHACS) at least one of the Scanner V4 tests (running on OCP 4.14) is failing because the collection method still defaults to `EBPF` within the definition of the test steps.
Hence, I am setting an explicit collection method here.


vladbologa (approved):

janisz (approved):

mclasmeier:
There are still some problems with the scanner v4 tests, but this PR does what it is supposed to be doing. Hence I will proceed with merging.

PR #48080, opened 2024-01-25
OSPRH-1949: openstack: set ironic tempest to enforce rbac scope

juliakreger:
This change sets the tests to "enforce_scope" mode for ironic's tempest plugin to execute, which is necessary to be in place if the overall service and container config is going to enforce the newer role based access controls, otherwise the test fails as the setting controls which access clients are used and how the tests respond.


juliakreger:
/pj-rehearse pull-ci-openstack-k8s-operators-ironic-operator-main-ironic-operator-build-deploy-tempest

PR #48089, opened 2024-01-25
Add periodic compliance jobs for OpenShift 4.14

rhmdnd:
Let's start running all compliance jobs on OpenShift 4.14 so we have a
better understanding of what the compliance posture is on that version.


rhmdnd:
I'm good with the testing here since it adds coverage for all supported profiles on 4.14. We can chase down the handful of failures in separate bugs targeted to those profiles.

PR #48091, opened 2024-01-25
Add periodic compliance jobs for OpenShift 4.16

rhmdnd:
Let's start running all compliance jobs on OpenShift 4.16 so we have a
better understanding of what the compliance posture is on that version.


rhmdnd:
Rehearsal failures are already being worked in CaC/content.

Vincent056 (approved):
/lgtm

PR #48108, opened 2024-01-26
Remove tox job from ci-framework

pablintino:
Since all the testing we do using tox is entirely covered by ansible-test let's remove this duplication that wastes resources and make maintenance harder.


marios (approved):
lgtm

PR #48114, opened 2024-01-26
Add CI security scanning jobs to cloud-network-config-controller, sdn and cluster-network-operator

pperiyasamy:
This change adds security scanning jobs to cloud-network-config-controller, sdn and cluster-network-operator repos. The jobs are currently marked optional which means they won't gate the merging of PRs.

### DIFF
--- a/ci-operator/config/openshift/cluster-network-operator/openshift-cluster-network-operator-master.yaml
+++ b/ci-operator/config/openshift/cluster-network-operator/openshift-cluster-network-operator-master.yaml
@@ -438,6 +438,13 @@ tests:
     - chain: create-infra-move-ingress-monitoring-registry
     workflow: openshift-qe-node-density-cni
   timeout: 5h0m0s
+- as: security
+  optional: true
+  steps:
+    env:
+      ALL_PROJECTS: "true"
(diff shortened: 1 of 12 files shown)


tssurya:
as agreed let's add SDN to the same PR (I think CNO/OVNK perhaps have it already because they are "mandatory components" and not optional)

pperiyasamy:
> as agreed let's add SDN to the same PR (I think CNO/OVNK perhaps have it already because they are "mandatory components" and not optional)

yes @tssurya , added security ci lane for the remaining networking components.

jluhrsen:
I think the ci/prow/generated-config job has something we need to clean up

pperiyasamy:
> I think the ci/prow/generated-config job has something we need to clean up

@jluhrsen It's fixed now, can you take a look ?

jluhrsen:
we expect all of these jobs to fail right now right?

the PR itself seems right though.

/lgtm

jcaamano:
> we expect all of these jobs to fail right now right?
> 
> the PR itself seems right though.
> 
> /lgtm

yup!

/lgtm
/approve

trozet (approved):
/lgtm

PR #48120, opened 2024-01-26
update vsphere metrics link

elmiko:
This change makes it so that the link on the "vSphere Metrics" accordion entry will link to the file that generates it.

flattered to have the camgi html copied, but i think we should update the github link.


elmiko:
cc @rvanderp3

elmiko:
cc @vr4manta

PR #48125, opened 2024-01-26
Enable branch_cleaner for openshift-eng/art-tools

ashwindasr:
To automatically clean branches after PR merge


core-services/prow/02_config/openshift-eng/art-tools/_pluginconfig.yaml
```diff
@@ -55,3 +55,9 @@ plugins:
     - wip
     - yuks
     - approve
+branch_cleaner:
```
sosiouxme, line 58: pretty sure we want to limit the plugin to our repo, and i don't think this does?

PR #48127, opened 2024-01-26
NO-JIRA: feat: added aliases for edge-enablement

jeff-roche:
Adding in the reviewers and approvers lists for the edge enablement team

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -390,3 +390,18 @@ aliases:
   - pmtk
   - eggfoobar
   - pliurh
+  edge-enablement-reviewers:
+  - jerpeter1
+  - eggfoobar
+  - jeff-roche
+  - suleymanakbas91
+  - jakobmoellerdev
+  - qJkee
+  edge-enablement-approvers:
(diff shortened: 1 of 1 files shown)


OWNERS_ALIASES
```diff
@@ -390,3 +390,17 @@ aliases:
   - pmtk
   - eggfoobar
   - pliurh
+edge-enablement-reviewers:
```
suleymanakbas91, line 393: Indentation seems to be off: 

Screenshot 2024-01-29 at 09 31 14

suleymanakbas91 (changes requested):

OWNERS_ALIASES
suleymanakbas91: Is there a reason why Bulat is not in the reviewers?

jeff-roche: No reason other than I just missed it. I fixed the spacing and added Bulat to reviewers

jeff-roche:
@danilo-gemoli how do I retrigger the tests?

PR #48130, opened 2024-01-26
Add branchcleaner for openshift-eng/art-tools

ashwindasr:


ashwindasr:
/close

Since this feature works only for branches from the same repo, and not forks.

PR #48134, opened 2024-01-26
Add a periodic job for ols e2e

bparees:


bparees:
pending a resolution to https://issues.redhat.com/browse/DPTP-3845 this is as good as it gets for now

PR #48161, opened 2024-01-29
[CVP-3974] Bugfix: Reset values of size and maxsize parameters for CVP OCP 4.16 clusterpools

samvarankashyap:
$subject


maneeshmehra:
LGTM

PR #48168, opened 2024-01-29
fix: OADP CI Standardize cron

mateusoliveira43:
Standardize cron definition across branches:
- master should be daily
- release branches should be weekly

To test crons, you can use something like https://crontab.guru/

To make the changes, `grep -Inr 'cron:' ci-operator/config/openshift/oadp-operator/` was run in the repo to get all OADP crons and then `make jobs`.


mpryc:
As per [Slack Thread](https://redhat-internal.slack.com/archives/CBN38N3MW/p1706535872520109) on the forum-ocp-testplatform, we should not change the cron times.

mpryc:
I think what we could do here is to see if frequency can be reduced?

PR #48173, opened 2024-01-29
MGMT-16656: remove e2e-metal-assisted-prod job

adriengentil:
remove e2e-metal-assisted-prod jobs:
- test-infra use mismatching assisted client, which sometimes causes flakes
- QE performs sanity tests in production, which makes this job redundant


ci-operator/config/openshift/assisted-test-infra/openshift-assisted-test-infra-master.yaml
```diff
@@ -82,14 +82,6 @@ tests:
     env:
       PACKET_PLAN: m3.small.x86
     workflow: assisted-baremetal
-- as: e2e-metal-assisted-prod
-  optional: true
-  steps:
-    cluster_profile: packet-assisted
-    env:
-      ENVIRONMENT: production
```
osherdp, line 90: there's all that logic in the steps that is not really needed now:
https://github.com/openshift/release/blob/ec1d33dafa0c181c67d8f4feb969018a7f745d27/ci-operator/step-registry/assisted/common/setup/prepare/assisted-common-setup-prepare-commands.sh#L111

can we clean that up as well?

osherdp (approved):

PR #48179, opened 2024-01-29
OSSM-4947, OSSM-4948: Jobs for envoy & proxy 2.5

jwendell:


ci-operator/config/maistra/proxy/maistra-proxy-maistra-2.5.yaml
```diff
@@ -0,0 +1,166 @@
...
+  timeout: 8h0m0s
+- as: copy-artifacts-gcs-2-5
+  postsubmit: true
+  steps:
+    cluster_profile: ossm-aws
+    env:
+      ARTIFACTS_GCS_PATH: gs://maistra-prow-testing/proxy
+      CI: "true"
+      COMPUTE_NODE_REPLICAS: "1"
+      COMPUTE_NODE_TYPE: m5d.8xlarge
```
yxun, line 84: Could you choose a smaller EC2 type for this copy artifacts step ? I don't think copying artifacts or post submit job need a lot amount of CPU resources.

m6i.large should be good.

ci-operator/config/maistra/proxy/maistra-proxy-maistra-2.5.yaml
```diff
@@ -0,0 +1,166 @@
...
+      GCS_PROJECT: maistra-prow-testing
+      LOCAL_CPU_RESOURCES: "30"
```
yxun, line 86: Do we still need to assign large resource amount values in those env when running copy artifacts post submit job ?

ci-operator/config/maistra/proxy/maistra-proxy-maistra-2.5.yaml
jwendell: the copy artifacts job also builds proxy. The only difference between it and the unit tests job is that the unit tests also run tests after building it.

yxun: ok, sounds like we have a duplicate part of building proxy between proxy unit job and copy artifacts job.
We can follow the current setting and try to remove duplicate costs in the future.

yxun (approved):

PR #48181, opened 2024-01-29
stackrox: add nightlies__ocp-4-15

davdhacs:
/hold
/uncc
candidate-4.15 is at rc.4 now: https://mirror.openshift.com/pub/openshift-v4/clients/ocp/candidate-4.15/

Using "candidate-4.15" to get the latest RC, and accepting changes (expecting fixes, accepting regressions may appear as we need to know about them asap).
4.15 is not GA yet and so there is no "ocp/stable-4.15" yet.


gavin-stackrox (approved):
/lgtm

PR #48186, opened 2024-01-30
update test script for managed service

yapei:
updated test script for managed service with a new name `console-test-managed-service.sh`


yapei:
@yanpzhan could you help review this change?

PR #48225, opened 2024-01-30
No need to install pdm, the make targets handle it

bparees:


bparees:
/hold

needs https://github.com/openshift/lightspeed-service/pull/297 first

PR #48233, opened 2024-01-30
Add image build config for ols rag image

bparees:


bparees:
@thoraxe fyi.  Note that the image is only going to get rebuild/published when PRs merge to the ols repo, so if the rag content is updated that won't trigger a new image build per se.

but this should at least keep it relatively current.

/hold

needs to wait until https://github.com/openshift/lightspeed-service/pull/293 is merged

PR #48251, opened 2024-01-31
add optional e2e-aws-ovn-upi to origin

lihongan:


lihongan:
/pj-rehearse ack
two jobs failed at installation because only 2 AZ in region "us-west-1" but the job needs 3 AZ.

lihongan:
/assign @bparees

set `always_run: false` and `optional: true`

PR #48258, opened 2024-01-31
METAL-832: Update base image of image-customization-controller to rhel-9

elfosardo:


elfosardo:
failures in 4.17 are expected

PR #48269, opened 2024-01-31
MGMT-15888: remove multi-images publish for ACM branches

osherdp:
After https://github.com/openshift/release/pull/43868 got in, it would mean that we cannot use the same code in order to build for linux/s390x, as it's not supported for `quay.io/centos/centos:stream8`.

It seems unneeded to build those images if they are not being used (as for those branches we separately build the images, bundles and indices). So this removes those jobs from existence.

This is a re-open of https://github.com/openshift/release/pull/43950 that got garbage-collected over time.


osherdp:
/hold
/cc @gamli75 

we do have multi-arch support in ACM/MCE, but it's not really using the upstream images that we produce here

adriengentil (approved):

PR #48288, opened 2024-01-31
Test openshift/imagebuilder with go 1.20

nalind:
Run our unit tests with a not-archived-yet version of Go.  Version 1.22 is about to be released, but more importantly, one of our dependencies is about to start using functions that were added to the standard library in version 1.20.


nalind:
This blocks https://github.com/openshift/imagebuilder/pull/276.

PR #48298, opened 2024-01-31
vSphere02: remove common except hosted

jcpowermac:
The configurations defined in common_except_hosted should not be applied to vsphere02.


jcpowermac:
cc: @rvanderp3 @vr4manta

PR #48338, opened 2024-02-01
NO-ISSUE: MicroShift clusterbot: Install all RPMs, optional included

pmtk:
Install all `microshift*` RPMs (including optional) so using microshift via clusterbot is more pleasant experience as installing, for example, `microshift-olm` requires several steps to get to the OS level as a root, install the RPM, restart microshift (which will terminate the `oc debug/node` session).
By installing everything, VM is more suitable to play with all capabilities of MicroShift.


dhellmann:
Since there's no jira link, could you add some details about why we want/need to do this?

pmtk:
> Since there's no jira link, could you add some details about why we want/need to do this?

Updated PR description. One reason I didn't mention is to help docs team with testing OLM on MicroShift.
In the future, when we have many more optional components, a parameter provided to clusterbot (like `MICROSHIFT_OPTIONAL=olm,multus,git-ops`) might be more suitable to not put too much load on the ec2 instance

PR #48362, opened 2024-02-02
Enhance cron entry for f360 jobs

liangxia:
Issue reported in https://redhat-internal.slack.com/archives/CJARLA942/p1703833815764369
For f360 jobs, they are regarded as disabled jobs, should not be executed in CI automatically. So hard-code the cron entry for those f360 jobs to `Feb 1, 0:00` ( today is Feb 2), so that those jobs will not be executed in the next 360 days.

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.10__amd64-nightly-4.10-upgrade-from-eus-4.8.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.10__amd64-nightly-4.10-upgrade-from-eus-4.8.yaml
@@ -45,7 +45,7 @@ resources:
       memory: 200Mi
 tests:
 - as: aws-ipi-f360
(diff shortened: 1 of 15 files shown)


jianlinliu:
> So hard-code the cron entry for those f360 jobs to Feb 1, 0:00 ( today is Feb 2), so that those jobs will not be executed in the next 360 days.

So when it comes to 2025, if we still keep them, then in a short time, a bunch of jobs would be triggered on the same comment?

liangxia:
> > So hard-code the cron entry for those f360 jobs to Feb 1, 0:00 ( today is Feb 2), so that those jobs will not be executed in the next 360 days.
> 
> So when it comes to 2025, if we still keep them, then in a short time, a bunch of jobs would be triggered on the same comment?

Most of the jobs should have been removed till then. And for the others, we can move them to Jan (e.g, maybe move them to Jan 30 on Jan 31, 2025).

JianLi-RH:
so far, lgtm

PR #48379, opened 2024-02-02
Adding codecov token secret for ols

vishnuchalla:
Followed these [instructions](https://docs.ci.openshift.org/docs/how-tos/adding-a-new-secret-to-ci/) to create a secret that has codecov bot token for OLS.

Code changes to point out the secret to PROW jobs which is required for uploading coverage reports.


tisnik (commented):
it looks ok!

PR #48384, opened 2024-02-02
Generate ols e2e periodic from common job config

bparees:


bparees:
/close

covered by https://github.com/openshift/release/pull/48244

thanks @hongkailiu !

PR #48385, opened 2024-02-02
cvo: add techpreview jobs

petr-muller:
https://github.com/openshift/cluster-version-operator/pull/1032 and https://github.com/openshift/cluster-version-operator/pull/1033 add TP-sensitive behavior to CVO, so having optional, manually triggered jobs does not hurt


wking (approved):
No opinion on the rehersal output yet, but that's a separate label-set anyway.  Diff  inspection looks good to me:

/lgtm

PR #48416, opened 2024-02-04
Add ocm cli in the rosa-aws-cli image

yasun1:
When we execute the rosa steps in Prow CI, sometimes we need to use the API endpoints to manage the OCM resources. It is inconvenient to make an individual step with the cli-ocm image for some tiny requirements. Let us add the ocm cli to the image for easy usage.

[OCPQE-19285](https://issues.redhat.com//browse/OCPQE-19285)


yasun1:
@hunterkepley Please help review. Thanks.

PR #48451, opened 2024-02-05
stackrox: fix max_concurrency for ocp-4-14 jobs

gavin-stackrox:
This should drain the queue and fix job backup going forward.
/uncc


BradLugo (approved):
I'm concerned we'll hit our quotas. We might want to do the monster math to see if that concern is justified, but I'm also down to experiment

/lgtm

PR #48463, opened 2024-02-05
Fix cvp-ocp-4-16 cluster pool

nmars:
Fixed region name and configured running count the same as other regions.


asergienk:
lgtm

PR #48491, opened 2024-02-06
Workaround pbr issue for openstack sushy

elfosardo:


elfosardo:
used a different workaround

PR #48505, opened 2024-02-06
Add firewatch fail on test failure to scenario.

ascerra:
I noticed a false positive in the devspaces scenario from the job result. The issue was still filed in Jira but the job showed green. making this change will allow for the job to fail when there is test failures for a more obvious approach to identifying jobs with failures.


calebevans (approved):
/lgtm

PR #48510, opened 2024-02-06
assisted installer: use pinned version of skipper

rccrdpccl:
Latest skipper has breaking changes that might break some jobs, this PR avoids that from happening


eifrach:
again
/lgtm

PR #48518, opened 2024-02-06
[ART-8873] Synk for openshift-eng/art-tools

ashwindasr:
Enable synk scanning on PRs for openshift-eng/art-tools


locriandev (commented):
/lgtm

PR #48522, opened 2024-02-06
jobs/openshift/os: run tests against default cosa variant

jlebon:
Instead of hardcoding a specific minor RHEL version here, just have the test run against whatever the default variant is in openshift/os. That way we can change it there without having to touch here.

Works together with https://github.com/openshift/os/pull/1424.


jlebon:
/pj-rehearse skip

We'll test this as part of https://github.com/openshift/os/pull/1424.

PR #48526, opened 2024-02-06
Drop a bunch of machine-os-content references

jlebon:
It's long been deprecated now and nothing should be relying on it
anymore.

See individual commits for details.

See also https://issues.redhat.com/browse/MCO-392.


jlebon:
OK, so I know this is kinda cutting across lots of different teams, and I can certainly split it into multiple PRs if preferred. Though at the same time, doing it this way allows us to keep the conversation in one place, which is helpful as this effort is honestly 90% about coordination.

There's also this reference: https://github.com/openshift/release/blob/4782116c8cf3ba6cf21fb499d7edb1a57654ce98/clusters/app.ci/assets/dptp-controller-manager.yaml#L160

which I didn't touch because I'm not sure exactly what that is. Maybe @hongkailiu can chime in for that one.

sinnykumari:
@soltysh Can we run any test here to ensure that this doesn't break O/K ci ?

jlebon:
OK, split this into separate PRs to make this easier to go through.
https://github.com/openshift/release/pull/49155
https://github.com/openshift/release/pull/49156
https://github.com/openshift/release/pull/49157

PR #48527, opened 2024-02-06
Stagger periodic compliance CI by 30 minutes

rhmdnd:
Instead of running 60 jobs at the same time, we can stagger the runs by
OpenShift version, which makes the results a little easier to parse
since they'll be grouped by version.


rhmdnd:
@Vincent056 @yuumasato this should be ready for review.

yuumasato (approved):
/lgtm

PR #48528, opened 2024-02-06
Remove redundant periodic compliance CI jobs

rhmdnd:
Previously, we were only testing periodic compliance jobs against a
single OpenShift version. A recent move to a different CI profile allows
us to run more jobs, meaning we can expand our coverage.

Another patch updated the CI to include periodic jobs for 4.14, which is
what the `master` jobs were testing against.

  https://github.com/openshift/release/pull/48527

Since we already have 4.14 test coverage, we can remove the legacy
periodic jobs.


rhmdnd:
@Vincent056 @yuumasato this should be ready for review.

yuumasato:
/lgtm

But @rhmdnd and @Vincent056 I don't have approver's permisison, :sweat_smile:

PR #48543, opened 2024-02-07
add baseimage 'assisted-service-scripts' to support hypershift agent

LiangquanLi930:


LiangquanLi930:
@jhou1 @liangxia Could you help to review this PR? Thanks

jhou1:
Does this also need to be applied to 4.16 files and other versions?

LiangquanLi930:
> Does this also need to be applied to 4.16 files and other versions?

We don't need to update to 4.16. For MCE-Agent 4.16, we need to wait for MCE2.6, which will require 2 weeks or even more after OCP 4.15 GA to have the build for MCE 2.6.

PR #48568, opened 2024-02-07
New jobs for the oadp-kubevirt

mpryc:
New set of jobs for the kubevirt within OADP.


mrnold:
This can be closed, I pulled it in with https://github.com/openshift/release/pull/48623

PR #48607, opened 2024-02-08
check if job is triggered from PR

bmanzari:
This PR copies the environment variables set in the Prow Pod running the reservation of nodes for the BM QE jobs to the bastion's /var/builds/$CLUSTER_NAME dir. It will improve debugging when resources are left because the deprovisioning steps don't run.

Use PR number if exists


aleskandro (approved):
Thanks @bmanzari

PR #48644, opened 2024-02-09
Fix some issues for rosacli testing

radtriste:
- Increase test timeout
    - According to https://github.com/openshift/openshift-tests-private/pull/14383#discussion_r1467152221    
- Avoid "\n" into `create_cluster.sh` command

### DIFF
--- a/ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
+++ b/ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
@@ -500,8 +500,7 @@ if [[ ${ENABLE_SHARED_VPC} == "yes" ]]; then
   echo "    SAHRED_VPC_BASE_DOMAIN: ${SAHRED_VPC_BASE_DOMAIN}"
 fi
 
-echo -e "
-rosa create cluster -y \
+echo -e "rosa create cluster -y \
 ${STS_SWITCH} \
 --mode auto \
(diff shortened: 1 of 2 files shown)


radtriste:
@yasun1 To you for approval

ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
```diff
@@ -500,8 +500,7 @@ if [[ ${ENABLE_SHARED_VPC} == "yes" ]]; then
   echo "    SAHRED_VPC_BASE_DOMAIN: ${SAHRED_VPC_BASE_DOMAIN}"
 fi
 
-echo -e "
-rosa create cluster -y \
+echo -e "rosa create cluster -y \
```
yasun1, line 503: Do we need to update this?

yasun1 (commented):
/lgtm

ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
radtriste: we may not anymore. But may be better to avoid the newline at the beginning of the file. wdyt ?

yasun1: No need. Both are ok for me.

PR #48649, opened 2024-02-09
Use the latest CLI when creating nested mgmt clusters

csrwng:
Uses the latest CLI (that corresponds to the version of hypershift on the CI cluster) to create a management cluster.


sjenning:
actually this is straight forward enough for me
/pj-rehearse ack

PR #48655, opened 2024-02-09
Remove rosa config for Kiali

smpsgill:
Removing the ROSA config for Kiali operator, as we don't use this and there is already config under [servicemesh](https://github.com/openshift/release/tree/master/ci-operator/config/maistra/maistra-test-tool)


calebevans (approved):
/lgtm

PR #48720, opened 2024-02-12
Power VS: Disable siguid job for 4.16

mjturek:
We're seeing some instability with creating and deleting workspaces.

In order to run more reliably, we are disabling the 4.16 job that creates a power vs workspace in favor of the job that uses a pre-created workspace.


vrutkovs (approved):
/approve
/lgtm

PR #48731, opened 2024-02-13
openshift-loki: Upgrade golang builder to 1.21 from openshift-4.16

periklis:
Finally moving the openshift/loki main branch that builds also Loki from main to Golang 1.21.

/cc @xperimental


xperimental:
fwiw: The `ci-operator-config` failure seems to affect more PRs and seems unrelated to this change.

PR #48509, opened 2024-02-06
STOR-1354, STOR-1611: Remove Alibaba CSI driver and IBM node labeler

dobsonj:
https://issues.redhat.com/browse/STOR-1611
https://issues.redhat.com/browse/STOR-1354

- [STOR-1611](https://issues.redhat.com//browse/STOR-1611): remove ibm-vpc-node-label-updater from release payload
- [STOR-1354](https://issues.redhat.com//browse/STOR-1354): remove alibaba-disk-csi-driver-operator from release payload
- [STOR-1354](https://issues.redhat.com//browse/STOR-1354): remove alibaba-cloud-csi-driver from release payload

/cc @openshift/storage @joepvd @jcpowermac

PR #48733, opened 2024-02-13
STOR-1354: Remove periodic e2e-alibaba-csi from 4.16

jsafrane:
Alibaba is being phased out. This is continuation of https://github.com/openshift/release/pull/48509, remove all 4.16 jobs that use `alibaba-disk-csi-driver-operator-test` image.


vrutkovs (approved):
/approve
/lgtm

PR #48738, opened 2024-02-13
Add autoscaler to build09

bear-redhat:


clusters/build-clusters/build09/openshift-machine-api/autoscaler-worker.yaml
```diff
@@ -0,0 +1,38 @@
...
+metadata:
+  name: build09-x99t2-worker-us-east-2a
+  namespace: openshift-machine-api
+spec:
+  maxReplicas: 10
+  minReplicas: 1
+  scaleTargetRef:
+    apiVersion: machine.openshift.io/v1beta1
+    kind: MachineSet
+    name: build09-x99t2-worker-us-east-2a
```
danilo-gemoli, line 12: Where do the names `build09-x99t2-worker-us-east-*` come from? Are they auto-generated?

bear-redhat: yes thats the default naming

PR #48753, opened 2024-02-13
Change the name of a slack channel for quarkus-qe team

jedla97:
The #47716 only change one slack channel and forgot change 2nd channel. So this is follow up on it.


fedinskiy (approved):
Thanks!

PR #48754, opened 2024-02-13
ROX-21665: [stackrox] Disable nightly gke e2e ebpf

ovalenti:
The COS image version is broken with ebpf drivers 2.9.0.

stackrox/collector is testing ebpf+COS.


tommartensen (approved):
/lgtm

PR #48766, opened 2024-02-13
Use builder root from repo operator-framework-rukpak

kevinrizza:
Update rukpak 4.12 and 4.13 to match config for rest of builds Don't pull releease tag from image stream


ankitathomas (approved):
/lgtm

kevinrizza:
The one test failure is because we're already pointing to a go 1.19 image in the repo and that is failing our linter, which is expected. We can fix that up when we bump the rest of the repo to use 1.19 to build

/pj-rehearse ack

grokspawn (changes requested):
If lint is never going to succeed until/unless we correct code in the rukpak repo, then shouldn't you remove these lines to (temporarily) disable the linter?

grokspawn (approved):

PR #48776, opened 2024-02-14
telco5g: Run cnftest BM job on sriov-network-operator repo

danimjaffe:
With this patch we enable BM job on sriov-network-operator repo. It is important to note that this job is optional and is triggered only when commented `/test e2e-telco5g-cnftests` in PR.

### DIFF
--- a/ci-operator/config/openshift/sriov-network-operator/openshift-sriov-network-operator-master.yaml
+++ b/ci-operator/config/openshift/sriov-network-operator/openshift-sriov-network-operator-master.yaml
@@ -156,6 +156,19 @@ tests:
       ADDITIONAL_WORKERS_NETWORKS: ""
       OPENSTACK_SRIOV_NETWORK: mellanox-hwoffload
     workflow: openshift-e2e-openstack-sriov-cgroupsv1
+- always_run: false
+  as: e2e-telco5g-cnftests
+  cluster: build05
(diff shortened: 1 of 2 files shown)


ci-operator/config/openshift/sriov-network-operator/openshift-sriov-network-operator-master.yaml
```diff
@@ -156,6 +156,19 @@ tests:
...
+- always_run: false
+  as: e2e-telco5g-cnftests
+  cluster: build05
+  optional: true
+  steps:
+    env:
+      E2E_TESTS_CONFIG: |
+        CNF_E2E_TESTS=true
+        CNF_ORIGIN_TESTS=false
+      T5CI_JOB_TYPE: cnftests
```
SchSeba, line 168: should we have a focus here on stuff that we want?
for sriov I will like 
dpdk, sriov, multinetworkpolicy, sctp,security, vrf

danimjaffe: For sure.
Currently working on this but it requires us to add some variable that we could use to specify which features we would want to run tests on:
https://github.com/openshift/release/pull/48949

SchSeba (commented):
/lgtm
/approve

PR #48783, opened 2024-02-14
Bump equinix-ocp-metal-quota-slice up to 70

tdomnesc:
We have 72 CIRs available in OFCIR right now but jobs are waiting (and some timing out) on the old quota of 50.


derekhiggins:
/lgtm
Note that this may end up with a CI job failing due to lack of available host but it avoids the situation where a CI job is held up for upto an hour only to fail because the hour waiting on the lease consumed to much of the time allocated to the job.

PR #48784, opened 2024-02-14
ofcir acquire: also display body on error

tdomnesc:
Display body on error. Most useful when no hosts are available where " No available resource found" will be displayed after the possibly misleading "curl: (22) The requested URL returned error: 404"


andfasano:
Can you paste a sample output please?

tdomnesc:
@andfasano - here is the string https://github.com/openshift-eng/ofcir/blob/6eb99825b946ca34bba6818a9dd08fd1bb703b6f/pkg/server/commands/acquire.go#L95 but curl has output into the file that I "cat" in this PR

derekhiggins:
/pj-rehearse ack
no need for all of these to pass, once one got passed the ofcir step we know the change is good

PR #48786, opened 2024-02-14
Add rule tests to telemeter job

simonpasquier:


ci-operator/config/openshift/telemeter/openshift-telemeter-master.yaml
```diff
@@ -44,7 +44,7 @@ tests:
   container:
     from: src
 - as: unit
```
machine424, line 46: nit: we could rename it to `test` then?

machine424:
/lgtm
/hold 
until the linked PR is merged.

ci-operator/config/openshift/telemeter/openshift-telemeter-master.yaml
simonpasquier: I didn't change the name because 1) rule tests are also unit tests and 2) (more important) I'm lazy :)

machine424: Then let’s keep it for the next person who might get confused by it :)

PR #48811, opened 2024-02-14
Increase the time between periodic compliance CI tests by 30 minutes

rhmdnd:
Even with 30 minutes in-between tests, some results are getting reported
together. This commit increases the time between tests across
versions giving each version a 1 hour head start before the next version
starts running, which should be ample time for the cluster to come up
and the test to run before the next suite starts.


Vincent056 (approved):
/lgtm

rhmdnd:
/pj-rehearse ack

No need to run rehearsals since this is only adjusting run times.

Vincent056 (approved):
/lgtm

PR #48819, opened 2024-02-15
[LPTOCPCI-1123] Add interop tests for MTA v7

calebevans:


Sau1506mya:
Just a quick question, Why you haven't made the change in the existing file for OCP 4.15 and made other file for v7?

PR #48825, opened 2024-02-15
telco5g: Remove baremetal job from numaresources-operator repo, this is a full revert of #48774

danimjaffe:
As of now we don't have e2e tests from numaresources operator repo in cnf-features-deploy: https://github.com/openshift-kni/cnf-features-deploy/tree/master/cnf-tests/submodules So we don't run them in periodic and can't run them in baremetal PR jobs that we are working on them. Once it will be added, we can add this job back.


shajmakh:
I understand this is a full revert of https://github.com/openshift/release/pull/48774, if so I think it would be more clear to mention this in the commit title to avoid any confusion

ffromani (commented):
/approve
/lgtm

PR #48853, opened 2024-02-15
feat: Run tests just one a week

jhutar:
Once we have a proven process for looking at the results, we can move the frequency back if it is useful.


concaf (approved):

khrm (commented):
/test generated-config

PR #48855, opened 2024-02-15
Update upstream podman required status checks

cevich:
Third-party CI systems are in use by upstream podman.  Some of which
will post status that is required for PRs to merge safely, other
statuses should be left up to human-operators to pontificate.

Prow is already desensitized to "unknown contexts", by
`skip-unknown-contexts: true` setting in
`core-services/prow/02_config/_config.yaml`
as per:
https://docs.prow.k8s.io/docs/components/core/tide/config/#context-policy-options

For all podman branches, prow should not manage branch-protection,
but it should block PR merging based on the Cirrus-CI success-aggregation
task "Total Success" as per:
https://docs.ci.openshift.org/docs/architecture/branch-protection

### DIFF
--- a/core-services/prow/02_config/containers/podman/_prowconfig.yaml
+++ b/core-services/prow/02_config/containers/podman/_prowconfig.yaml
@@ -1,11 +1,3 @@
-branch-protection:
-  orgs:
-    containers:
-      repos:
-        podman:
-          branches:
-            v1.6:
-              protect: false
 tide:
   queries:
   - labels:
--- a/core-services/prow/02_config/_config.yaml
+++ b/core-services/prow/02_config/_config.yaml
@@ -414,6 +414,8 @@ tide:
           netavark:
             skip-unknown-contexts: true
           podman:
+            required-contexts:
+            - Total Success
             skip-unknown-contexts: true
           podman-py:
             skip-unknown-contexts: true


core-services/prow/02_config/containers/podman/_prowconfig.yaml
```diff
@@ -3,9 +3,11 @@ branch-protection:
     containers:
       repos:
         podman:
-          branches:
-            v1.6:
-              protect: false
+          protect: false  # Do not manage branch-protection settings.
+        required_status_checks:
+          contexts:
+            - success  # "Total Success" Cirrus-CI task must pass.
```
edsantiago, line 9: I've been impossibly confused the last few days when reading comments like "branch protection goes by NAME", because there are many names. If the following comment is true (and it might not be, because I'm still confused), I think it might be more helpful for future maintainers:
```
 # Require the cirrus task with this ALIAS to pass
```

cevich: There's different Human-names (for display) and machine-names (for use/reference).  I get your point though.  And this reminds me, I need to use the github API to actually check what prow/tide will see...

cevich: ...glad I checked, looks like it's using the human name.  For Example podman PR 21657, I find the check-run ID is directly in the 'details' url, so 21579329450:

curl -L -H "Accept: application/vnd.github+json" \
  -H "X-GitHub-Api-Version: 2022-11-28" \
  https://api.github.com/repos/containers/podman/check-runs/21579329450

mentions:

`"name": "Total Success"`

So that's probably what needs to be referenced...force-push incoming...

cevich:
Ref: [Link to discussion re: comment and what name to reference](https://github.com/openshift/release/pull/48855#discussion_r1491458479)

hongkailiu (approved):

edsantiago:
Hard to believe that it could be this simple, but it sure seems worth a try! Thank you!

cevich:
Thanks for following along Ed, as always, the extra pair of sharp eyes is much appreciated.

We'll run a test to confirm this actually does what we want.  But I agree it does seem simple.  Then you realize we have an (excellent) staffed helpdesk for it, and go look at the size of both the openshift-CI docs for prow, and realize those are on top of the (HUGE) set of docs from google about it.  Prow and it's configuration is the opposite of a simple tool :disappointed_relieved: 

Anyway, if the test passes, I intend to setup something similar for the other prow-managed containers-org repos.  I don't believe any other repo. (except buildah) has a success aggregator, and clearly prow is not sensitized to any other Cirrus-CI task in them.

**Note for future maintainers**: After changing files here, you have to run `make prow-config` which requires `docker-podman` to be installed, and you must `export CONTAINER_ENGINE=podman VOLUME_MOUNT_FLAGS=''` for it to work.

PR #48912, opened 2024-02-16
stackrox: OCP 4.15 tests on master, add merge and alwaysrun

davdhacs:
/hold
/uncc


BradLugo (approved):
Gavin already approved this; the only change since then was the build server

/lgtm

PR #48914, opened 2024-02-16
route-monitor-operator: remove integration presubmit test

tnierman:
Removes the route-monitor-operator's `test-integration` presubmit, as it ~~doesn't provide value~~ provides reduced value to the maintainers & contributors compared to our other, well-established, testing systems for the following reasons:

- Stage is our test bed. This is why we don't SLA it and have a pretty formal ceremony around promotions. The inflexibility of these tests make it significantly more difficult to prototype changes we want to experiment with in a casual setting
- The tests are flaky/difficult to reproduce: prow consistently reports failures that I cannot recreate
- The tests degrade the cluster they run on: since the default namespace used to test in is openshift-route-monitor-operator, and we end up deleting the test namespace [during cleanup](https://github.com/openshift/route-monitor-operator/blob/master/hack/test-integration.sh#L131)
- The goal of the tests (ensure the operator is behaving as expected) is duplicated by [osde2e](https://github.com/openshift/osde2e/blob/main/pkg/e2e/operators/routemonitor.go#L1). In the other operators that I've worked with, tests ensure specific components/packages/logic within the codebase continue to function as expected, not the operator as a whole.
- Not as impactful as above, but each test-integration run needs at least 20 minutes to run, many times more (ymmv based on prows status). It's annoying when trying to make quick hotfixes or revert production-breaking changes


mjlshen (commented):
> doesn't provide value to the maintainers or contributors.

Is a bit strong, I think it normally does provide an additional sanity check - and I am in favor of moving this check to osde2e, but I think that the osde2e test should preferably exist first before removing this one (afaik it doesn't exist yet).

Otherwise, removing specific parts of this integration test that are flaking for now if something is hindering development.

NautiluX:
Did you consider the failing test might have actually revealed a problem with your change?

tnierman:
Yes, of course I have considered that the changes I'm trying to introduce to the codebase are causing the test to fail. My reasoning can be summarized as follows: 
- Stage is our test bed. This is why we don't SLA it and have a pretty formal ceremony around promotions. The inflexibility of these tests make it significantly more difficult to prototype changes we want to experiment with in a casual setting
- The tests are flaky/difficult to reproduce: prow consistently reports failures that I cannot recreate
- The tests degrade the cluster they run on: since the default namespace used to test in is openshift-route-monitor-operator, and we end up deleting the test namespace [during cleanup](https://github.com/openshift/route-monitor-operator/blob/master/hack/test-integration.sh#L131)
- The goal of the tests (ensure the operator is behaving as expected) is duplicated by [osde2e](https://github.com/openshift/osde2e/blob/main/pkg/e2e/operators/routemonitor.go#L1). In the other operators that I've worked with, tests ensure specific components/packages/logic within the codebase continue to function as expected, not the operator as a whole.
- Not as impactful as above, but each test-integration run needs at least 20 minutes to run, many times more (ymmv based on prows status). It's annoying when trying to make quick hotfixes or revert production-breaking changes

mjlshen:
/reopen
/remove-lifecycle rotten

Can be removed now that this is captured by osde2e https://github.com/openshift/route-monitor-operator/pull/300

mjlshen (approved):
/lgtm

PR #48934, opened 2024-02-19
auto ocp-53921

jiajliu:


jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-stable-azure-ipi-to-multiarch-f14

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-stable-azure-ipi-to-multiarch-f14

jiajliu:
cc @jianlinliu @shellyyang1989 for a review, thanks!

ci-operator/step-registry/cucushift/ota/postupgrade/cucushift-ota-postupgrade-commands.sh
```diff
@@ -159,6 +159,51 @@ function post-OCP-47197(){
...
+        return 1
+    fi
+    recommends=$(oc get clusterversion version -o json|jq -r '.status.availableUpdates')
+    if [[ "${recommends}" == "null" ]]; then
+        echo "No recommended update available!"
+        return 1
+    fi
+    mapfile -t images < <(echo ${recommends}|jq -r '.[].image')
+    for image in ${images[*]}; do
+        if [ -z "${image}" ] ; then
```
jianlinliu, line 190: if `${images[*]}` is empty, the for loop will be skipped, right? Sounds need to check the length of `images` outside of the for loop?

jiajliu: right, usually a build should include metadata, so i think if any issue came from the payload itself(build error?), we could ignore the metadata info check for this test.

jianlinliu: Hmm, my comment is not about image metadata checking, but about `image` is empty or not, if the `image` is an empty string, the for loop would directly skip it, would never come to this if condition.

jiajliu: Got u now:) good point!

`availableUpdates` is a node like following(including image and version)
```
$ oc get clusterversion -o json|jq '.items[0].status.availableUpdates'
[
  {
    "image": "registry.ci.openshift.org/ocp/release@sha256:c7c3d2b39681c37817fb99c6f05140ee92a52920b7909d6e973d0025e6b60568",
    "version": "4.15.0-rc.1"
  }
]
```
there is a non-null check for availableupdate in https://github.com/openshift/release/pull/48934/files#diff-bcec571809d92fb0e609baf8907886244e300b0f2aa08759cc9f6cac2154a225R184, so if availableupdate is not null, images[] should not be null?

jianlinliu: If that, the line 190 - 193 sounds can be removed?

jiajliu: Hmm, agree that in theory. `availableupdate!=null` is to ensure recommends list is not empty. But if there is one bad node join in the list unluckily, line 190 - 193 is for extra guarantee. 

In the extreme case that there is only one node in recommends list and it’s a bad node, then it will skip. I did not consider such a bad case, let me update it to handle it.

jiajliu: > If that, images would be an empty array,

If there are more than one node in the array, then it's not an empty array but with a bad node with empty `image`. The loop will be skipped in the extreme case that there is only one bad node in the array.

jiajliu:
```
[
  {
    "image": "registry.ci.openshift.org/ocp/release@sha256:c7c3d2b39681c37817fb99c6f05140ee92a52920b7909d6e973d0025e6b60568",
    "version": "4.15.0-rc.1"
  },
 {
    "image": "",
    "version": "4.15.0-rc.2"
  },
 {
    "image": "registry.ci.openshift.org/ocp/release@sha256:xxxx",
    "version": ""
  },
]
```

jiajliu:
cc @liangxia for a review, thanks!

PR #48973, opened 2024-02-19
TRT-1521: Add dummy log to debug missing artifact folder for openshift-tests step

xueqzhan:
We are debugging a case where no artifact is generated at all for the openshift-test step. This adds a dummy file before running the test to make sure it is not the openshift-tests binary that is misbehaving.


hongkailiu (commented):
/appove

hongkailiu (commented):
/approve

PR #48987, opened 2024-02-20
Fix taint label for Outposts node

yunjiang29:
Fix label issue in https://github.com/openshift/release/pull/48950


yunjiang29:
@gpei PTAL, thanks

PR #49014, opened 2024-02-20
Enhancement: add node-role label for Outposts node

Phaow:
### Enhancement: add node-role label for Outposts node
- The outposts machineset template better also add the role label to keep consistent with official doc. I just launch a cluster and check the node doesn't have the `outpost` role, add the role will make the configuration more obviously.
https://github.com/openshift/openshift-docs/pull/71263/files#diff-5352612c7759b518002f58019ba9d4560a5534418e5edc300b9f80c9dca566c4R121-R125


Phaow:
@yunjiang29 @gpei Could you please help take a look when you get a chance?

PR #49017, opened 2024-02-20
Interop - Add support for minimal OPERATORS dict

oharan2:
We see that our operator-dicts are usually duplicated with the same values ​​for `channel`, `source`, and installation `namespace`.

Therefore, instead of exiting if not mentioned, we provide support for the minimal operator dict, which will mention for example only the operator name value.

Ref: https://redhat-internal.slack.com/archives/C04PK4QPSR1/p1707923448438239?thread_ts=1707836882.233399&cid=C04PK4QPSR1

### DIFF
--- a/ci-operator/config/oadp-qe/oadp-qe-automation/oadp-qe-oadp-qe-automation-main__oadp1.3-ocp4.15-lp-interop.yaml
+++ b/ci-operator/config/oadp-qe/oadp-qe-automation/oadp-qe-oadp-qe-automation-main__oadp1.3-ocp4.15-lp-interop.yaml
@@ -78,8 +78,8 @@ tests:
       FIREWATCH_DEFAULT_JIRA_PROJECT: LPINTEROP
       OPERATORS: |
         [
-          {"name": "redhat-oadp-operator", "channel": "stable-1.3", "install_namespace": "openshift-adp", "target_namespaces": "openshift-adp", "operator_group":"oadp-operator-group"},
-          {"name": "volsync-product", "operator_group":"global-operators"}
+          {"name": "redhat-oadp-operator", "source": "redhat-operators", "channel": "stable-1.3", "install_namespace": "openshift-adp", "target_namespaces": "openshift-adp", "operator_group":"oadp-operator-group"},
+          {"name": "volsync-product", "source": "redhat-operators", "channel": "!default", "install_namespace": "openshift-operators", "target_namespaces": "", "operator_group":"global-operators"}
         ]
       USER_TAGS: |
         scenario oadp


oharan2:
/pj-rehearse periodic-ci-oadp-qe-oadp-qe-automation-main-oadp1.3-ocp4.15-lp-interop-oadp-interop-aws

oharan2:
/pj-rehearse periodic-ci-oadp-qe-oadp-qe-automation-main-oadp1.3-ocp4.15-lp-interop-oadp-interop-aws

ci-operator/step-registry/install-operators/install-operators-commands.sh
```diff
@@ -26,22 +33,20 @@ for operator_obj in "${OPERATOR_ARRAY[@]}"; do
...
         exit 1
     fi
 
-    # If channel is not defined, exit.
+    # If install_namespace not defined, use DEFAULT_OPERATOR_INSTALL_NAMESPACE.
+    if [[ -z "${operator_install_namespace}" ]]; then
+        operator_install_namespace="${DEFAULT_OPERATOR_INSTALL_NAMESPACE}"
+    fi
+
+    # If channel is not defined, use DEFAULT_OPERATOR_CHANNEL.
```
Simranpal, line 47: Do you want to handle the situation when channel is not defined, but some other channel(eg stable-x.y like [here](https://github.com/Simranpal/release/blob/5fad04cec9ef18fa08b65993d331fe4118929457/ci-operator/config/ansible/tower-qa/ansible-tower-qa-devel__ocp4.15-lp-interop.yaml#L48)) is being used other than default? or will using default set that channel

oharan2: I want to provide the user the option to not specify the channel, so when it's not mentioned, it'll automatically set with `!default`. Than our script is picking the default channel from the marketplace (might be stable-x.y or anything else).

If channel is mentioned, like in your ref- sets to `"channel": "stable-2.4"`, it'll not use the default value "!default". Hope that answered your question :)

Simranpal: okay yes, I wanted to know if it will pick the default channel (like example above) from marketplace.
Thanks!

oharan2:
/pj-rehearse periodic-ci-oadp-qe-oadp-qe-automation-main-oadp1.3-ocp4.15-lp-interop-oadp-interop-aws

oharan2:
/pj-rehearse periodic-ci-oadp-qe-oadp-qe-automation-main-oadp1.3-ocp4.15-lp-interop-oadp-interop-aws

calebevans (approved):
/lgtm

PR #49049, opened 2024-02-21
auto ocp-69948

jiajliu:


jiajliu:
cc @jianlinliu @shellyyang1989 for a review first, due to there might be bug for z-stream upgrade on tp, so the post-upgrade test could not be ran in this pr.

jiajliu:
The issue has been confirm to be a bug, tracked in https://issues.redhat.com/browse/OCPBUGS-29815. So i think that's okey to get the pr merged without a pass job. 

/pj-rehearse ack

jiajliu:
cc @liangxia for a review, thanks!

PR #49057, opened 2024-02-21
Sync Serverless CI

serverless-qe:


mgencur:
Note: All the jobs on 4.15 run in a cloud account owned by Serverless Engineering, except for serverless-operator repo which runs with the shared "aws" cluster profile.

mgencur:
There's a problem in the always_run part which should be "false" for all jobs running on OCP 4.15. Will send a fix.

PR #49060, opened 2024-02-21
NO-ISSUE: Change OCM 2.7 periodic jobs to run once a week

danmanor:
Change OCM 2.7 periodic jobs to run once a week


osherdp (approved):

danmanor:
@adriengentil @osherdp @gamli75 Do we want it to run in a specific day in the week (like Saturday) ?

osherdp:
> @adriengentil @osherdp @gamli75 Do we want it to run in a specific day in the week (like Saturday) ?

I think it's not too critical

PR #49084, opened 2024-02-21
release-controller: Stop mirroring OCP CI builds to OKD

LorbusChris:
/cc @ausil @lmzuccarelli


lmzuccarelli:
@LorbusChris - as mentioned this PR has high priority so from my perspective the sooner it gets merged the better

vrutkovs:
/hold

This breaks https://github.com/openshift/enhancements/blob/master/enhancements/okd-on-fedora-coreos-in-prow.md. It needs to be updated first along with a discussion on its effects at Arch meeting

ausil:
We can not share ELS RHEL content in OKD; this is a necessary step to rectify the issue and to set up builds for OKD with CentOS Stream content.

vrutkovs:
We don't even know which exactly content is being shared, it may or may not be exactly the same as UBI RPMs. Lets hold off potentially destructive actions until the plan for further actions is discussed in enhancements, agreed by all parties (OKD community, 3rd party operators using these images and OpenShift architects).

jupierce:
Hold to be discussed in Slack, but:
/approve

LorbusChris:
The enhancement will be updated in due time to reflect the new OKD image build process
/hold cancel

PR #49157, opened 2024-02-22
core-services: drop machine-os-content references for 4.16+

jlebon:
That image isn't actually used in the cluster and will be removed from the release payload soon.

See also https://issues.redhat.com/browse/MCO-392.


jlebon:
IIUC, I think we will actually want to merge this at some point _after_ we flip the switch again to not publish a machine-os-content image in the payload. This PR is removing the m-o-c from the list of tags to exclude from the OKD/origin releases. We want to keep excluding it while it still exists. Doing it after the m-o-c no longer exists would just be an aesthetic cleanup.

/hold

LorbusChris:
This won't be necessary once we stop mirroring images from ocp to origin altogether: https://github.com/openshift/release/pull/49084

PR #49165, opened 2024-02-23
Fix tests for aws-ipi-wavelength-fips

liangxia:


yunjiang29:
@liangxia thanks for the fix!

PR #49204, opened 2024-02-26
Moving Vale script logic to prow step to try to fix 404 response from curl github API when the step calls a sub-script

rohennes:
Troubleshooting a 404 response for curl requests in this script. Instead of calling a script with the vale review logic, this moves the vale review logic into the step. Seeing if this makes a difference.


aireilly:
/pj-rehearse pull-ci-openshift-openshift-docs-prow-test-2-validate-asciidoc

PR #49227, opened 2024-02-27
Add fips in test name for fips enabled jobs and update cron entry

liangxia:
Several jobs have `FIPS_ENABLED: "true"`, but the names do not indicate that, so add `-fips` to the names.


JianLi-RH:
Sorry, I can see only a few of your changes are related to `fips`, most of the changes are about cron settings, I don't think they should be in one PR. at least the pr title does not describe we will update cron setting.

liangxia:
> Sorry, I can see only a few of your changes are related to `fips`, most of the changes are about cron settings, I don't think they should be in one PR. at least the pr title does not describe we will update cron setting.

Yes, only a few job renames. Since there are job renames, so the related cron entry needs update, too. So I used the batch update tool to do that. The other jobs that do not have rename in this PR while have cron entry update might be because there are renames earlier but no cron change at that time.

PR #49267, opened 2024-02-27
provide existing kubeconfig to osde2e test step if available from provision step

ritmun:
currently, if cluster provisioning is done in a separate step registry step, the test step does not have access to it by default.

Load it using osde2e env var "TEST_KUBECONFIG"

### DIFF
--- a/ci-operator/step-registry/osde2e/test/osde2e-test-commands.sh
+++ b/ci-operator/step-registry/osde2e/test/osde2e-test-commands.sh
@@ -16,7 +16,7 @@ then
     exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/osde2e/test/osde2e-test-commands.sh
```diff
@@ -16,6 +16,11 @@ then
     exit 1
 fi
 
+if [[ ! -f "${SHARED_DIR}/kubeconfig" ]];
```
jbpratt, line 19:
```suggestion
if [ -f "${SHARED_DIR}/kubeconfig" ];
```
if the file exists

PR #49292, opened 2024-02-28
Add cidr-expansion job as presubmit

jluhrsen:
the cidr-expansion already runs as a periodic, but we need it as a presubmit if we want to trigger it on CNO PRs. This adds it as optional:trey and always_run:false so it will only run if done with a GH comment.


abhat:
/lgtm
/approve
The rehearse job failing will be resolved by a CNO change that @jluhrsen is working on.

PR #49302, opened 2024-02-28
Check CV arch while checking arch migration status

shellyyang1989:
Closes [OCPQE-19299](https://issues.redhat.com/browse/OCPQE-19299). Prior to the change, the [step](https://github.com/openshift/release/blob/master/ci-operator/step-registry/cucushift/upgrade/arch-migration/cucushift-upgrade-arch-migration-commands.sh) might return migrate succeed while migrate is not started.

### DIFF
--- a/ci-operator/step-registry/cucushift/upgrade/arch-migration/cucushift-upgrade-arch-migration-commands.sh
+++ b/ci-operator/step-registry/cucushift/upgrade/arch-migration/cucushift-upgrade-arch-migration-commands.sh
@@ -294,7 +294,7 @@ function health_check() {
 
 # Monitor the migration status
 function check_migrate_status() {
-    local wait_migrate="${TIMEOUT}" out avail progress
+    local wait_migrate="${TIMEOUT}" out avail progress arch
     while (( wait_migrate > 0 )); do
         sleep 5m
         (( wait_migrate -= 5 ))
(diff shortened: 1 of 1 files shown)


shellyyang1989:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.13-arm64-stable-aws-ipi-ovn-ipsec-to-multiarch-f14

shellyyang1989:
@jianlinliu @jiajliu @JianLi-RH @dis016 for review. Thanks!

jiajliu:
> the [step](https://github.com/openshift/release/blob/master/ci-operator/step-registry/cucushift/upgrade/arch-migration/cucushift-upgrade-arch-migration-commands.sh) might return migrate succeed while migrate is not started.

hmm, why the migrate did not start? no available update?

ci-operator/step-registry/cucushift/upgrade/arch-migration/cucushift-upgrade-arch-migration-commands.sh
```diff
@@ -294,7 +294,7 @@ function health_check() {
 
 # Monitor the migration status
 function check_migrate_status() {
-    local wait_migrate="${TIMEOUT}" out avail progress
+    local wait_migrate="${TIMEOUT}" out avail progress arch
     while (( wait_migrate > 0 )); do
```
aleskandro, line 298: > the [step](https://github.com/openshift/release/blob/master/ci-operator/step-registry/cucushift/upgrade/arch-migration/cucushift-upgrade-arch-migration-commands.sh) might return migrate succeed while migrate is not started.

If I understand the issue here is about "we check that the CV is available and not in progress" when the migration/upgrade is not started yet, right?

shellyyang1989: Right

shellyyang1989:
> > the [step](https://github.com/openshift/release/blob/master/ci-operator/step-registry/cucushift/upgrade/arch-migration/cucushift-upgrade-arch-migration-commands.sh) might return migrate succeed while migrate is not started.
> 
> hmm, why the migrate did not start? no available update?

Yes. Seems like we need to check if VersionNotFound is present, if yes, skip the test.

```
2024-02-02T18:57:13.787629383Z I0202 18:57:13.787570       1 cincinnati.go:114] Using a root CA pool with 0 root CA subjects to request updates from https://api.openshift.com/api/upgrades_info/v1/graph?arch=multi&channel=candidate-4.13&id=64011054-0c99-4334-8595-76e5a84ec08f&version=4.13.32
2024-02-02T18:57:14.826668385Z I0202 18:57:14.826617       1 availableupdates.go:348] Upstream server https://api.openshift.com/api/upgrades_info/v1/graph could not return available updates: VersionNotFound: currently reconciling cluster version 4.13.32 not found in the "candidate-4.13" channel
```

aleskandro:
I wonder if it makes sense to add a check earlier than the "migration completed" one, that wait until the migration start (at an higher frequency than 1/5m).

jiajliu:
> Yes. Seems like we need to check if VersionNotFound is present, if yes, skip the test.

Yes, that's my concern. Sounds like the migrate need some precondition to start. +1 to add the check before run migrate.

shellyyang1989:
the job was triggered by someone manually and the image was not selected appropriately. So VersionNotFound (in candidate channel) was present and the test failed.
By default, the latest version in the candidate channel will be picked and such issue will not occur.

/close

PR #49316, opened 2024-02-28
Temp disable all loki due to outage and failing tests as a result (again)

dgoodwin:
Reverts openshift/release#49277

Loki is busted again for different reasons.

We will fix the test that fails on loki outage in 4.16 at least but likely not all releases. (https://github.com/openshift/origin/pull/28627)

For now, shut it all down again.


dgoodwin:
/hold

we might be fixed.

PR #49317, opened 2024-02-28
Adds optional flag to doctoc

iamkirkbater:
If links go stale in other parts of the repo, this PR check will block the current merge.

The goal here will be to make this optional to allow PR reviewers to make the decision whether the links that are dead need to be updated as part of this PR or if that can be done in a separate PR to not muddy the commit history for unrelated changes.

### DIFF
--- a/ci-operator/jobs/openshift/ops-sop/openshift-ops-sop-master-presubmits.yaml
+++ b/ci-operator/jobs/openshift/ops-sop/openshift-ops-sop-master-presubmits.yaml
@@ -14,6 +14,7 @@ presubmits:
       ci.openshift.io/generator: prowgen
       pj-rehearse.openshift.io/can-be-rehearsed: "true"
     name: pull-ci-openshift-ops-sop-master-doctoc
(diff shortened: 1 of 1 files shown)


iamkirkbater:
Sure, I'll grab the link in a bit when I get to my computer today. The specific links were links to ROSA docs, external to the repo. 

I'd also be happy to consider an alternative where we block the PR on any links that are dead that are within the diff but have a separate optional job that signals other links in the repo too.

Tof1973:
I'm a bit lost, because I expect doctoc to just generate toc on md documents, and do not see why doctoc would check any external links in the documents... Did I miss something in doctoc [documentation](https://www.npmjs.com/package/doctoc) ?

iamkirkbater:
I think I'm lost too - I could have sworn that yesterday it was complaining about two unrelated ROSA docs links that were 404s in that make target, but when I look at the builds they're just complaining about the doctoc like you said.

Primary must be killing my brain.

/close

PR #49318, opened 2024-02-28
[opendatahub-io/notebooks] Update image stream tags

jstourac:
* sync golang version with kubeflow
* update OCP version used since the 4.10 is EOL already [1]

[1] https://access.redhat.com/support/policy/updates/openshift#dates

@harshad16 , please let me know whether we really want something like this :upside_down_face:


harshad16 (commented):
/lgtm
/approve


thanks for the update.
this may help with the build system.

PR #49321, opened 2024-02-28
Add optional recert job to openshift-cluster-kube-apiserver-operator master, 4.15, 4.16 and 4.17

eranco74:


eranco74:
We need to get the periodic job blocking first.

PR #49327, opened 2024-02-28
MULTIARCH-4242 - Deprecating sdn network deployment jobs on ARM in 4.15+

tvardema:


clusters/app.ci/release-reimport-controller/deployment.yaml
```diff
@@ -29,6 +29,7 @@ spec:
       - command:
         - /usr/bin/release-reimport-controller
         - start
+        - --dry-run
```
bradmwilliams, line 32: You'll need to rebase against the latest/greatest changes.  This parameter was removed recently.

bradmwilliams (changes requested):

clusters/app.ci/release-reimport-controller/deployment.yaml
tvardema: I literally just generated that an hour ago... talk about unfortunate timing, ha ha ha

bradmwilliams: This isn't your fault actually...  It needs to be corrected in the correct place...
https://github.com/openshift/release/pull/49343
That will need to merge and then this file should drop out after you rebase and do a:
`make release-controllers`

bradmwilliams: Sorry for the churn!

PR #49356, opened 2024-02-28
Run hypershift-aws-conformance, not hypershift-aws-e2e-external in CNO

danwinship:
`hypershift-aws-e2e-external` is an "install a hypershift cluster and then run hypershift-related tests" job, while `hypershift-aws-conformance` is "install a hypershift cluster and then run the ocp e2e test suite", which sounds more like what we want. More importantly, it doesn't have a crazy flake rate.

/assign @jluhrsen

### DIFF
--- a/ci-operator/config/openshift/cluster-network-operator/openshift-cluster-network-operator-master.yaml
+++ b/ci-operator/config/openshift/cluster-network-operator/openshift-cluster-network-operator-master.yaml
@@ -248,7 +248,7 @@ tests:
 - as: e2e-hypershift-ovn
   steps:
     cluster_profile: hypershift
-    workflow: hypershift-aws-e2e-external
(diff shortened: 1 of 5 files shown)


jluhrsen:
i like it
/lgtm

if the rehearsals are good, just ack them and it'll land.

danwinship:
The 4.17 run is failing because... something seems to be missing? Presumably it will get fixed at some point...

AFAICT, the 4.16/master runs are failing because a particular `oc` invocation is printing a warning (presumably something new in 4.16) which an e2e test then tries to parse as json and fails. But it doesn't print enough info to debug and I can't reproduce it in a normal hypershift cluster...

danwinship:
/pj-rehearse

https://github.com/openshift/origin/pull/28631 should fix the 4.16/master problem

PR #49378, opened 2024-02-29
Remove OKD jobs from old releases in ironic-agent

elfosardo:
OKD is not going to ship any more builds older than 4.15


elfosardo:
@derekhiggins @LorbusChris can you please approve/lgtm again?
also please @bparees @dhellmann @jupierce this is kind of urgent as the CI for all the modified releases is broken until we merge this
thanks!

PR #49384, opened 2024-02-29
Revert "'log' file compression isn't working properly, temporarily remove to avoid corrupted data"

smg247:
To be merged after https://github.com/openshift/ci-tools/pull/3997

Reverts openshift/release#49376

/hold


smg247:
let's not compress log files (at least for now), the vast majority are already compressed, and this gains us nothing.
/close

PR #49388, opened 2024-02-29
PERFSCALE-2870 - Reduce payload jobs to once per week

afcollins:
We were running these payload jobs M-F to gain confidence to submit them as payload jobs. Now that they are run with payloads, we don't need them as frequently.

control-plane is running multiple times on most days, so we do not need to run on our own.
cni and heavy can run once/week so we can keep a pulse on development.


afcollins:
/pj-rehearse periodic-ci-openshift-qe-ocp-qe-perfscale-ci-main-aws-4.16-nightly-x86-payload-control-plane-6nodes

afcollins:
Good to go now. Please check it

jtaleric (approved):
/lgtm

PR #49406, opened 2024-03-01
Changed the cluster profile of 4.16 ppc64le

aishwaryabk:
Changed the cluster profile to enable the 4.16 ppc64le job on dal10.
cc: @juliemathew @aleskandro


aishwaryabk:
Updated the job schedule

PR #49422, opened 2024-03-01
enable the multi-arch building for the ovn-kubernetes repo

droslean:


droslean:
@smg247 I am not sure if this is a bug, but I was expecting to rehearse those jobs. Is this expected?

smg247:
> @smg247 I am not sure if this is a bug, but I was expecting to rehearse those jobs. Is this expected?

They are all `postsubmits`, we don't rehearse those.

PR #49431, opened 2024-03-01
MGMT-17099: Fix recert rename containers not stopping

mresvanis:
This PR fixes the recert script waiting on all crio pods to be forcefully removed, which on OCP 4.15 fails indefinitely. Instead, we stop all crio containers, stop crio and then change the hostname of the VM and run recert.

It also increases the OCP stability period from 2 min to 5 min after running recert, in order to avoid false positives. Specifically, it can be that for more that 2 min, etcd and the kube-apiserver are up and running, serving the last CVO status (e.g. `all cluster operators are stable`), but no cluster operator is up yet. This would lead to the job indicating that recert was successful, but actually the cluster could be broken (e.g. because after those 2 min MCO was up and indicated that its status is not healthy).

### DIFF
--- a/ci-operator/step-registry/baremetalds/sno/rename/baremetalds-sno-rename-commands.sh
+++ b/ci-operator/step-registry/baremetalds/sno/rename/baremetalds-sno-rename-commands.sh
@@ -198,7 +198,7 @@ then
 
   delete_crts_keys
 
-  stable_period_minutes=2
+  stable_period_minutes=5
   start=\$(date +%s)
   start_containers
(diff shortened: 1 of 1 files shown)


mresvanis:
/pj-rehearse periodic-ci-rh-ecosystem-edge-recert-main-4.14-baremetalds-sno-recert-cluster-rename periodic-ci-rh-ecosystem-edge-recert-main-4.15-baremetalds-sno-recert-cluster-rename periodic-ci-rh-ecosystem-edge-recert-main-4.16-baremetalds-sno-recert-cluster-rename

PR #49448, opened 2024-03-01
add kubeconfig to test command if provided by pre-step

ritmun:
Step jobs with separate provision steps need kubeconfig passed into the test step, so added TEST_KUBECONFIG var when file is present in shared dir.

https://issues.redhat.com/browse/SDCICD-1192


YiqinZhang (commented):
/lgtm

PR #49451, opened 2024-03-01
[LPTOCPCI-1153] Add re-trigger functionality to MTR and MTA 4.16 interop tests

calebevans:
- Adds post step to `firewatch-ipi-aws` workflow that will re-trigger jobs if the jobs have `RE_TRIGGER_ON_FAILURE: "true"` (eventually, the default will be true in this workflow, after the role-out is complete).
- Enables auto re-triggering for MTR and MTA 4.16 scenarios


Simranpal (approved):
/lgtm

PR #49503, opened 2024-03-04
Restrict tests run for cinder-operator

Akrog:
This patch restricts the tempests tests to be run on the cinder-operator-build-deploy-tempest job using a regex.

The regex limits it to the volume tests.


Akrog:
/pj-rehearse pull-ci-openstack-k8s-operators-cinder-operator-main-cinder-operator-build-deploy-tempest

Akrog:
Closing because this would still need an skip list, and now that the job has been fixed with the `tempest-skiplist` in the tempest container it's no longer that important.

PR #49540, opened 2024-03-05
OCPQE-19762 Skip upgrade when target is not recommended in stable-stable path

JianLi-RH:
[OCPQE-19762](https://issues.redhat.com/browse/OCPQE-19762)

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__amd64-nightly-4.16-upgrade-from-stable-4.1.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__amd64-nightly-4.16-upgrade-from-stable-4.1.yaml
@@ -1,0 +1,192 @@
+base_images:
+  ansible:
+    name: "4.16"
+    namespace: ocp
+    tag: ansible
+  cli:
+    name: "4.16"
+    namespace: ocp
+    tag: cli
+  dev-scripts:
+    name: test
+    namespace: ocp-kni
+    tag: dev-scripts
+  tests-private-postupg:
+    name: tests-private
+    namespace: ci
+    tag: "4.16"
+  tests-private-preupg:
+    name: tests-private
+    namespace: ci
+    tag: "4.15"
+  tools:
+    name: "4.16"
+    namespace: ocp
+    tag: tools
+  verification-tests:
+    name: verification-tests
+    namespace: ci
+    tag: latest
+releases:
+  initial:
+    release:
+      architecture: amd64
+      channel: stable
+      version: "4.1"
+  intermediate42:
+    release:
+      architecture: amd64
+      channel: fast
+      version: "4.2"
+  intermediate43:
+    release:
+      architecture: amd64
+      channel: fast
+      version: "4.3"
+  intermediate44:
+    release:
+      architecture: amd64
+      channel: fast
+      version: "4.4"
+  intermediate45:
+    release:
+      architecture: amd64
+      channel: fast
+      version: "4.5"
+  intermediate46:
+    release:
+      architecture: amd64
+      channel: fast
+      version: "4.6"
+  intermediate47:
+    release:
+      architecture: amd64
+      channel: fast
+      version: "4.7"
+  intermediate48:
+    release:
+      architecture: amd64
+      channel: fast
+      version: "4.8"
+  intermediate49:
+    release:
+      architecture: amd64
+      channel: fast
+      version: "4.9"
+  intermediate410:
+    release:
+      architecture: amd64
+      channel: fast
+      version: "4.10"
+  intermediate411:
+    release:
+      architecture: amd64
+      channel: fast
+      version: "4.11"
+  intermediate412:
+    candidate:
+      architecture: amd64
+      product: ocp
+      stream: nightly
+      version: "4.12"
+  intermediate413:
+    candidate:
+      architecture: amd64
+      product: ocp
+      stream: nightly
+      version: "4.13"
+  intermediate414:
+    candidate:
+      architecture: amd64
+      product: ocp
+      stream: nightly
+      version: "4.14"
+  intermediate415:
+    candidate:
+      architecture: amd64
+      product: ocp
+      stream: nightly
+      version: "4.15"
+  latest:
+    release:
+      architecture: amd64
+      channel: candidate
+      version: "4.14"
+  target:
+    candidate:
+      architecture: amd64
+      product: ocp
+      stream: nightly
+      version: "4.16"
+resources:
+  '*':
+    requests:
+      cpu: 100m
+      memory: 200Mi
+tests:
+- as: aws-ipi-f30
+  cluster: build04
+  cron: 43 0 2 * *
+  steps:
+    cluster_profile: aws-qe
+    dependencies:
+      OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE: release:initial
+    env:
+      BASE_DOMAIN: qe.devcluster.openshift.com
+      LOKI_ENABLED: "false"
+    post:
+    - chain: cucushift-installer-rehearse-aws-ipi-deprovision
+    pre:
+    - ref: ipi-conf
+    - ref: ipi-conf-telemetry
+    - ref: ipi-conf-aws
+    - chain: ipi-install-stableinitial
+    test:
+    - as: set-upgrade-releases
+      commands: echo "${RELEASE_IMAGE_INTERMEDIATE42},${RELEASE_IMAGE_INTERMEDIATE43},${RELEASE_IMAGE_INTERMEDIATE44},${RELEASE_IMAGE_INTERMEDIATE45},${RELEASE_IMAGE_INTERMEDIATE46},${RELEASE_IMAGE_INTERMEDIATE47},${RELEASE_IMAGE_INTERMEDIATE48},${RELEASE_IMAGE_INTERMEDIATE49},${RELEASE_IMAGE_INTERMEDIATE410},${RELEASE_IMAGE_INTERMEDIATE411},${RELEASE_IMAGE_INTERMEDIATE412},${RELEASE_IMAGE_INTERMEDIATE413},${RELEASE_IMAGE_INTERMEDIATE414},${RELEASE_IMAGE_INTERMEDIATE415},${RELEASE_IMAGE_TARGET}"
+        > ${SHARED_DIR}/upgrade-edge
+      dependencies:
+      - env: RELEASE_IMAGE_INTERMEDIATE415
+        name: release:intermediate415
+      - env: RELEASE_IMAGE_INTERMEDIATE414
+        name: release:intermediate414
+      - env: RELEASE_IMAGE_INTERMEDIATE413
+        name: release:intermediate413
+      - env: RELEASE_IMAGE_INTERMEDIATE412
+        name: release:intermediate412
+      - env: RELEASE_IMAGE_INTERMEDIATE411
+        name: release:intermediate411
+      - env: RELEASE_IMAGE_INTERMEDIATE410
+        name: release:intermediate410
+      - env: RELEASE_IMAGE_INTERMEDIATE49
+        name: release:intermediate49
+      - env: RELEASE_IMAGE_INTERMEDIATE48
+        name: release:intermediate48
+      - env: RELEASE_IMAGE_INTERMEDIATE47
+        name: release:intermediate47
+      - env: RELEASE_IMAGE_INTERMEDIATE46
+        name: release:intermediate46
+      - env: RELEASE_IMAGE_INTERMEDIATE45
+        name: release:intermediate45
+      - env: RELEASE_IMAGE_INTERMEDIATE44
+        name: release:intermediate44
+      - env: RELEASE_IMAGE_INTERMEDIATE43
+        name: release:intermediate43
+      - env: RELEASE_IMAGE_INTERMEDIATE42
+        name: release:intermediate42
+      - env: RELEASE_IMAGE_TARGET
+        name: release:target
+      from: cli
+      resources:
+        requests:
+          cpu: 100m
+          memory: 200Mi
+    - ref: cucushift-upgrade-drop-last-hop
+    - ref: cucushift-chainupgrade-toimage
+    - ref: enable-qe-catalogsource
+    - chain: openshift-upgrade-qe-test
+zz_generated_metadata:
+  branch: release-4.16
+  org: openshift
+  repo: openshift-tests-private
+  variant: amd64-nightly-4.16-upgrade-from-stable-4.1
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__amd64-nightly-4.16-upgrade-from-stable-4.10.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__amd64-nightly-4.16-upgrade-from-stable-4.10.yaml
@@ -1,0 +1,114 @@
+base_images:
+  ansible:
+    name: "4.16"
+    namespace: ocp
+    tag: ansible
+  cli:
+    name: "4.16"
+    namespace: ocp
+    tag: cli
+  dev-scripts:
+    name: test
+    namespace: ocp-kni
+    tag: dev-scripts
+  tests-private-postupg:
+    name: tests-private
+    namespace: ci
+    tag: "4.16"
+  tests-private-preupg:
+    name: tests-private
+    namespace: ci
+    tag: "4.15"
+  tools:
+    name: "4.16"
+    namespace: ocp
+    tag: tools
+  upi-installer:
+    name: "4.10"
+    namespace: ocp
+    tag: upi-installer
+  verification-tests:
+    name: verification-tests
+    namespace: ci
+    tag: latest
+releases:
+  intermediate411:
+    release:
+      architecture: amd64
+      channel: fast
+      version: "4.11"
+  intermediate412:
+    candidate:
+      architecture: amd64
+      product: ocp
+      stream: nightly
+      version: "4.12"
+  intermediate413:
+    candidate:
+      architecture: amd64
+      product: ocp
+      stream: nightly
+      version: "4.13"
+  intermediate414:
+    candidate:
+      architecture: amd64
+      product: ocp
+      stream: nightly
+      version: "4.14"
+  intermediate415:
+    candidate:
+      architecture: amd64
+      product: ocp
+      stream: nightly
+      version: "4.15"
+  latest:
+    release:
+      architecture: amd64
+      channel: fast
+      version: "4.10"
+  target:
+    candidate:
+      architecture: amd64
+      product: ocp
+      stream: nightly
+      version: "4.16"
+resources:
+  '*':
+    requests:
+      cpu: 100m
+      memory: 200Mi
+tests:
+- as: vsphere-ipi-ovn-ipsec-f30
+  cron: 33 8 22 * *
+  steps:
+    cluster_profile: vsphere-2
+    test:
+    - as: set-upgrade-releases
+      commands: echo "${RELEASE_IMAGE_INTERMEDIATE411},${RELEASE_IMAGE_INTERMEDIATE412},${RELEASE_IMAGE_INTERMEDIATE413},${RELEASE_IMAGE_INTERMEDIATE414},${RELEASE_IMAGE_INTERMEDIATE415},${RELEASE_IMAGE_TARGET}"
+        > ${SHARED_DIR}/upgrade-edge
+      dependencies:
+      - env: RELEASE_IMAGE_INTERMEDIATE415
+        name: release:intermediate415
+      - env: RELEASE_IMAGE_INTERMEDIATE414
+        name: release:intermediate414
+      - env: RELEASE_IMAGE_INTERMEDIATE413
+        name: release:intermediate413
+      - env: RELEASE_IMAGE_INTERMEDIATE412
+        name: release:intermediate412
+      - env: RELEASE_IMAGE_INTERMEDIATE411
+        name: release:intermediate411
+      - env: RELEASE_IMAGE_TARGET
+        name: release:target
+      from: cli
+      resources:
+        requests:
+          cpu: 100m
+          memory: 200Mi
+    - ref: cucushift-chainupgrade-toimage
+    - chain: openshift-upgrade-qe-test
+    workflow: cucushift-installer-rehearse-vsphere-ipi-ovn-ipsec
+zz_generated_metadata:
+  branch: release-4.16
+  org: openshift
+  repo: openshift-tests-private
+  variant: amd64-nightly-4.16-upgrade-from-stable-4.10
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__amd64-nightly-4.16-upgrade-from-stable-4.11.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__amd64-nightly-4.16-upgrade-from-stable-4.11.yaml
@@ -1,0 +1,109 @@
+base_images:
+  ansible:
+    name: "4.15"
+    namespace: ocp
+    tag: ansible
+  cli:
+    name: "4.15"
+    namespace: ocp
+    tag: cli
+  dev-scripts:
+    name: test
+    namespace: ocp-kni
+    tag: dev-scripts
+  tests-private-postupg:
+    name: tests-private
+    namespace: ci
+    tag: "4.15"
+  tests-private-preupg:
+    name: tests-private
+    namespace: ci
+    tag: "4.14"
+  tools:
+    name: "4.15"
+    namespace: ocp
+    tag: tools
+  upi-installer:
+    name: "4.11"
+    namespace: ocp
+    tag: upi-installer
+  verification-tests:
+    name: verification-tests
+    namespace: ci
+    tag: latest
+releases:
+  intermediate412:
+    candidate:
+      architecture: amd64
+      product: ocp
+      stream: nightly
+      version: "4.12"
+  intermediate413:
+    candidate:
+      architecture: amd64
+      product: ocp
+      stream: nightly
+      version: "4.13"
+  intermediate414:
+    candidate:
+      architecture: amd64
+      product: ocp
+      stream: nightly
+      version: "4.14"
+  intermediate415:
+    candidate:
+      architecture: amd64
+      product: ocp
+      stream: nightly
+      version: "4.15"
+  latest:
+    release:
+      architecture: amd64
+      channel: fast
+      version: "4.11"
+  target:
+    candidate:
+      architecture: amd64
+      product: ocp
+      stream: nightly
+      version: "4.16"
+resources:
+  '*':
+    requests:
+      cpu: 100m
+      memory: 200Mi
+tests:
+- as: azure-ipi-ovn-ipsec-f30
+  cron: 33 0 6 * *
+  steps:
+    cluster_profile: azure-qe
+    env:
+      BASE_DOMAIN: qe.azure.devcluster.openshift.com
+    test:
+    - as: set-upgrade-releases
+      commands: echo "${RELEASE_IMAGE_INTERMEDIATE412},${RELEASE_IMAGE_INTERMEDIATE413},${RELEASE_IMAGE_INTERMEDIATE414},${RELEASE_IMAGE_INTERMEDIATE415},${RELEASE_IMAGE_TARGET}"
+        > ${SHARED_DIR}/upgrade-edge
+      dependencies:
+      - env: RELEASE_IMAGE_INTERMEDIATE415
+        name: release:intermediate415
+      - env: RELEASE_IMAGE_INTERMEDIATE414
+        name: release:intermediate414
+      - env: RELEASE_IMAGE_INTERMEDIATE413
+        name: release:intermediate413
+      - env: RELEASE_IMAGE_INTERMEDIATE412
+        name: release:intermediate412
+      - env: RELEASE_IMAGE_TARGET
+        name: release:target
+      from: cli
+      resources:
+        requests:
+          cpu: 100m
+          memory: 200Mi
+    - ref: cucushift-chainupgrade-toimage
+    - chain: openshift-upgrade-qe-test
+    workflow: cucushift-installer-rehearse-azure-ipi-ovn-ipsec
+zz_generated_metadata:
+  branch: release-4.16
+  org: openshift
+  repo: openshift-tests-private
+  variant: amd64-nightly-4.16-upgrade-from-stable-4.11
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__amd64-nightly-4.16-upgrade-from-stable-4.12.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__amd64-nightly-4.16-upgrade-from-stable-4.12.yaml
@@ -1,0 +1,213 @@
+base_images:
+  ansible:
+    name: "4.12"
+    namespace: ocp
+    tag: ansible
+  cli:
+    name: "4.12"
+    namespace: ocp
+    tag: cli
+  dev-scripts:
+    name: test
+    namespace: ocp-kni
+    tag: dev-scripts
+  tests-private-postupg:
+    name: tests-private
+    namespace: ci
+    tag: "4.16"
+  tests-private-preupg:
+    name: tests-private
+    namespace: ci
+    tag: "4.15"
+  tools:
+    name: "4.12"
+    namespace: ocp
+    tag: tools
+  upi-installer:
+    name: "4.12"
+    namespace: ocp
+    tag: upi-installer
+  verification-tests:
+    name: verification-tests
+    namespace: ci
+    tag: latest
+releases:
+  intermediate413:
+    candidate:
+      architecture: amd64
+      product: ocp
+      stream: nightly
+      version: "4.13"
+  intermediate414:
+    candidate:
+      architecture: amd64
+      product: ocp
+      stream: nightly
+      version: "4.14"
+  intermediate415:
+    candidate:
+      architecture: amd64
+      product: ocp
+      stream: nightly
+      version: "4.15"
+  latest:
+    release:
+      architecture: amd64
+      channel: fast
+      version: "4.12"
+  target:
+    candidate:
+      architecture: amd64
+      product: ocp
+      stream: nightly
+      version: "4.16"
+resources:
+  '*':
+    requests:
+      cpu: 100m
+      memory: 200Mi
+tests:
+- as: baremetalds-ipi-ovn-dualstack-fips-f14
+  cron: 2 23 10,24 * *
+  steps:
+    cluster_profile: equinix-ocp-metal-qe
+    env:
+      DEVSCRIPTS_CONFIG: |
+        FIPS_MODE=true
+        IP_STACK=v4v6
+        NETWORK_TYPE=OVNKubernetes
+    test:
+    - as: set-upgrade-releases
+      commands: echo "${RELEASE_IMAGE_INTERMEDIATE413},${RELEASE_IMAGE_INTERMEDIATE414},${RELEASE_IMAGE_TARGET}"
+        > ${SHARED_DIR}/upgrade-edge
+      dependencies:
+      - env: RELEASE_IMAGE_INTERMEDIATE414
+        name: release:intermediate414
+      - env: RELEASE_IMAGE_INTERMEDIATE413
+        name: release:intermediate413
+      - env: RELEASE_IMAGE_TARGET
+        name: release:target
+      from: cli
+      resources:
+        requests:
+          cpu: 100m
+          memory: 200Mi
+    - chain: openshift-upgrade-qe-test-chainupgrade
+    workflow: cucushift-installer-rehearse-baremetalds-ipi-ovn
+- as: baremetalds-ipi-ovn-ipv6-fips-f14
+  cron: 9 20 1,16 * *
+  steps:
+    cluster_profile: equinix-ocp-metal-qe
+    env:
+      DEVSCRIPTS_CONFIG: |
+        FIPS_MODE=true
+        IP_STACK=v6
+        MIRROR_IMAGES=true
+        INSTALLER_PROXY=true
+        NETWORK_TYPE=OVNKubernetes
+    test:
+    - as: set-upgrade-releases
+      commands: echo "${RELEASE_IMAGE_INTERMEDIATE413},${RELEASE_IMAGE_INTERMEDIATE414},${RELEASE_IMAGE_INTERMEDIATE415},${RELEASE_IMAGE_TARGET}"
+        > ${SHARED_DIR}/upgrade-edge
+      dependencies:
+      - env: RELEASE_IMAGE_INTERMEDIATE415
+        name: release:intermediate415
+      - env: RELEASE_IMAGE_INTERMEDIATE414
+        name: release:intermediate414
+      - env: RELEASE_IMAGE_INTERMEDIATE413
+        name: release:intermediate413
+      - env: RELEASE_IMAGE_TARGET
+        name: release:target
+      from: cli
+      resources:
+        requests:
+          cpu: 100m
+          memory: 200Mi
+    - chain: openshift-upgrade-qe-test-chainupgrade
+    workflow: cucushift-installer-rehearse-baremetalds-ipi-ovn
+- as: gcp-ipi-basecap-none-f14
+  cron: 17 7 12,26 * *
+  steps:
+    cluster_profile: gcp-qe
+    env:
+      BASELINE_CAPABILITY_SET: None
+      PERSISTENT_MONITORING: "false"
+      UPGRADE_CHECK_RUN_TAGS: not @console and not @storage
+      UPGRADE_PRE_RUN_TAGS: not @console and not @storage
+    test:
+    - as: set-upgrade-releases
+      commands: echo "${RELEASE_IMAGE_INTERMEDIATE413},${RELEASE_IMAGE_INTERMEDIATE414},${RELEASE_IMAGE_INTERMEDIATE415},${RELEASE_IMAGE_TARGET}"
+        > ${SHARED_DIR}/upgrade-edge
+      dependencies:
+      - env: RELEASE_IMAGE_INTERMEDIATE415
+        name: release:intermediate415
+      - env: RELEASE_IMAGE_INTERMEDIATE414
+        name: release:intermediate414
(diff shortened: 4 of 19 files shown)


JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.16-amd64-nightly-4.16-upgrade-from-stable-4.15-aws-ipi-disc-priv-f28

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-stable-4.15-upgrade-from-stable-4.14-azure-ipi-fullyprivate-proxy-f28

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-stable-4.15-upgrade-from-stable-4.15-aws-ipi-disc-priv-tp-f28

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-multi-stable-4.15-upgrade-from-stable-4.15-aws-ipi-ovn-ipsec-arm-mixarch-f28

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-automated-release-stable-4.15-upgrade-from-stable-4.15-aws-ipi-fips-arm-mixarch-f1

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-automated-release-stable-4.15-upgrade-from-stable-4.15-aws-ipi-fips-arm-mixarch-f1

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-nightly-4.15-upgrade-from-stable-4.14-aws-ipi-shared-vpc-phz-f14

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-nightly-4.15-upgrade-from-stable-4.14-aws-ipi-byo-iam-role-fips-f28

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.16-amd64-nightly-4.16-upgrade-from-stable-4.15-vsphere-upi-f28

JianLi-RH:
/cc @shellyyang1989 @jiajliu @jianlinliu @dis016 
Both stable-stable and stable-nightly upgrade are working normal.

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -685,7 +685,23 @@ function upgrade() {
         clear_upgrade
         check_upgrade_status "${cluster_src_ver}"
     fi
-    run_command "oc adm upgrade --to-image=${TARGET} --allow-explicit-upgrade --force=${FORCE_UPDATE}"
+
+    cmd="oc adm upgrade --to-image=${TARGET} --allow-explicit-upgrade"
+    if [[ "$SOURCE_VERSION" != *"nightly"* ]] && [[ "$TARGET_VERSION" != *"nightly"* ]]; then
+        if [[ "$ALLOW_NOT_RECOMMENDED" == 'true' ]]; then
+            run_command "$cmd --allow-not-recommended"
```
jiajliu, line 692: If there is not Recommended update, but ALLOW_NOT_RECOMMENDED is set to true, it will fail or not? For ci, we never know if there will be recommended update when setting a job. 

There is a Recommended condition in conditionalUpdates. Maybe we could check if that condition exist/match before upgrade func.

JianLi-RH: from the help:
```
    --allow-not-recommended=false:
	Allows upgrade to a version when it is supported but not recommended
	for updates.
```
I think `--allow-not-recommended=false` will not affect a normal upgrade, I add this parameter is just for future testing that we want to check unsafe upgrade.

jiajliu: > run_command "$cmd --allow-not-recommended"

Here means --allow-not-recommended=true, right? Could we check if it  affect a normal upgrade?

JianLi-RH: We don't have profiles that enabled `ALLOW_NOT_RECOMMENDED`, but anyway I can check it locally.

JianLi-RH: Here is the test without and with --allow-not-recommended
```
[jianl@localhost 415]$ oc adm upgrade --to-image quay.io/openshift-release-dev/ocp-release@sha256:0da6316466d60a3a4535d5fed3589feb0391989982fba59d47d4c729912d6363 --allow-explicit-upgrade
warning: The requested upgrade image is not one of the available updates. You have used --allow-explicit-upgrade for the update to proceed anyway
Requested update to release image quay.io/openshift-release-dev/ocp-release@sha256:0da6316466d60a3a4535d5fed3589feb0391989982fba59d47d4c729912d6363
[jianl@localhost 415]$ 
[jianl@localhost 415]$ 
[jianl@localhost 415]$ oc adm upgrade --clear
Cancelled requested upgrade to quay.io/openshift-release-dev/ocp-release@sha256:0da6316466d60a3a4535d5fed3589feb0391989982fba59d47d4c729912d6363
[jianl@localhost 415]$ 
[jianl@localhost 415]$ oc adm upgrade --to-image quay.io/openshift-release-dev/ocp-release@sha256:0da6316466d60a3a4535d5fed3589feb0391989982fba59d47d4c729912d6363 --allow-explicit-upgrade --allow-not-recommended
warning: The requested upgrade image is not one of the available updates. You have used --allow-explicit-upgrade for the update to proceed anyway
Requested update to release image quay.io/openshift-release-dev/ocp-release@sha256:0da6316466d60a3a4535d5fed3589feb0391989982fba59d47d4c729912d6363
[jianl@localhost 415]$ 

```

JianLi-RH:
```
[jianl@localhost 415]$ oc get clusterversion
NAME      VERSION                              AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.15.0-0.nightly-2024-03-06-142724   True        True          61s     Working towards 4.15.0: 108 of 874 done (12% complete), waiting on etcd, kube-apiserver
[jianl@localhost 415]$ 
```

jianlinliu: I think we should add a separate step in `openshift-upgrade-qe-sanity` as the 1st step, once the target payload is not recommended, we can exit immediately to save the running for the following step in the chain.

jiajliu: from the test result, it looks like a no harm var when there is not recommended update, so i'm ok to leave ALLOW_NOT_RECOMMENDED there.

jiajliu: > once the target payload is not recommended, we can exit immediately to save the running for the following step in the chain.

+1 for this way, i also suggest to check it before upgrade with the recommended condition in my 1st comment.

JianLi-RH:
@shellyyang1989 Because 'not recommended' not means not allowed, so clients will also choose that version to upgrade to, so I want to add `--allow-not-recommended` in `oc adm upgrade` when the version is a conditionalUpdates.

JianLi-RH:
Admin can make decision if they will upgrade or not upgrade to a risk version against the risk link:
[https://docs.openshift.com/container-platform/4.15/updating/understanding_updates/how-updates-work.html#update-evaluate-avai[…]lity_how-updates-work](https://docs.openshift.com/container-platform/4.15/updating/understanding_updates/how-updates-work.html#update-evaluate-availability_how-updates-work)
```
The user interface, either the web console or the OpenShift CLI (oc), presents this information in sectioned headings to the administrator. Each supported but not recommended update recommendation contains a link to further resources about the risk so that the administrator can make an informed decision about the update.
```

shellyyang1989:
> @shellyyang1989 Because 'not recommended' not means not allowed, so clients will also choose that version to upgrade to, so I want to add `--allow-not-recommended` in `oc adm upgrade` when the version is a conditionalUpdates.

Yes, we should cover not recommended upgrade. But I don't suggest we cover it in common upgrade workflow. Instead, we can cover it in OTA test cases.

JianLi-RH:
@shellyyang1989 How to cover it in OTA test cases? By manually run v4-install job?

shellyyang1989:
> @shellyyang1989 How to cover it in OTA test cases? By manually run v4-install job?

I meant explicitly enabling it in a Prow job by using `ENABLE_OTA_TEST` when we want to run the test-case, instead of implicitly enabling it in all possible Prow jobs.

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
shellyyang1989: Agree with Jianlin, the intention is to exit as early as possible when we find the target image is a not recommended update.

JianLi-RH: I think we need a call to discuss it

shellyyang1989: Sure.

JianLi-RH:
OK, since no custom will choose a not-recommended payload, I will add a step to skip that version.

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.16-arm64-nightly-4.16-upgrade-from-stable-4.15-aws-ipi-disc-priv-f28

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-stable-4.15-upgrade-from-stable-4.14-ibmcloud-ipi-proxy-private-fips-rt-f14

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.16-amd64-nightly-4.16-upgrade-from-stable-4.15-aws-ipi-proxy-sts-f14

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.16-amd64-nightly-4.16-upgrade-from-stable-4.15-aws-ipi-imdsv2-fips-f28

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-multi-stable-4.15-upgrade-from-stable-4.15-aws-ipi-ovn-ipsec-arm-mixarch-f28

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-4.14-upgrade-from-stable-4.13-aws-ipi-ovn-winc-f7

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-nightly-4.15-upgrade-from-stable-4.14-azure-ipi-disc-fullyprivate-f14

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-arm64-stable-4.15-upgrade-from-stable-4.14-aws-ipi-proxy-sts-f28

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-nightly-4.15-upgrade-from-stable-4.2-aws-ipi-f30

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.16-amd64-nightly-4.16-upgrade-from-stable-4.15-vsphere-ipi-disc-f28

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-stable-4.15-upgrade-from-stable-4.14-aws-ipi-ovn-fips-f28

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.13-amd64-stable-4.13-upgrade-from-stable-4.12-vsphere-ipi-disc-f14

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -685,7 +685,19 @@ function upgrade() {
         clear_upgrade
         check_upgrade_status "${cluster_src_ver}"
     fi
-    run_command "oc adm upgrade --to-image=${TARGET} --allow-explicit-upgrade --force=${FORCE_UPDATE}"
+
+    cmd="oc adm upgrade --to-image=${TARGET} --allow-explicit-upgrade"
+    if [[ "$TARGET_VERSION" != *"nightly"* ]]; then
+        conditionalUpdates=$(oc get clusterversion version -o json | jq -r '.status.conditionalUpdates[]?.version')
```
jianlinliu, line 691:
```suggestion
        conditionalUpdates=$(oc get clusterversion version -o json | jq -r '.status.conditionalUpdates[]?.release.version')
```
Missing `release` ?

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -685,7 +685,19 @@ function upgrade() {
...
+        if [[ "$conditionalUpdates" == *"$TARGET_VERSION"* ]]; then
+            echo "Error: $TARGET_VERSION is not recommended, for details please refer:"
+            oc get clusterversion version -o json | jq -r '.status.conditionalUpdates'
+        else
+            run_command "$cmd"
+        fi
+    else
+        run_command "$cmd --force=${FORCE_UPDATE}"
```
jianlinliu, line 699: Actually we do not need to interfere `FORCE_UPDATE`, just keep the origin line, just add a pre-check, then exit before run this upgrade command line.

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
JianLi-RH: no, it is as expected

jianlinliu, line 691: Do you have an example output of `oc get clusterversion version -o json | jq -r '.status.conditionalUpdates` ?

jianlinliu: also cc @jiajliu to help confirm.

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -685,7 +685,19 @@ function upgrade() {
         clear_upgrade
         check_upgrade_status "${cluster_src_ver}"
     fi
-    run_command "oc adm upgrade --to-image=${TARGET} --allow-explicit-upgrade --force=${FORCE_UPDATE}"
+
+    cmd="oc adm upgrade --to-image=${TARGET} --allow-explicit-upgrade"
+    if [[ "$TARGET_VERSION" != *"nightly"* ]]; then
+        conditionalUpdates=$(oc get clusterversion version -o json | jq -r '.status.conditionalUpdates[]?.version')
+        if [[ "$conditionalUpdates" == *"$TARGET_VERSION"* ]]; then
```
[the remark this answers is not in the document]
JianLi-RH: Yes, you are right, missing `release`.

```
[jianl@localhost 413]$ cat clusterversion.json | jq -r '.status.conditionalUpdates[]?.release.version' | xargs
4.15.0
[jianl@localhost 413]$ conditionalUpdates=$(cat clusterversion.json | jq -r '.status.conditionalUpdates[]?.release.version' | xargs)
[jianl@localhost 413]$ TARGET_VERSION=4.15
[jianl@localhost 413]$ if [[ " $conditionalUpdates " =~ " $TARGET_VERSION " ]]; then
> echo 'Yes'
> fi
[jianl@localhost 413]$ if [[ "$conditionalUpdates" == *"$TARGET_VERSION"* ]]; then
> echo 'Yes'
> fi
Yes
[jianl@localhost 413]$ 
```

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
JianLi-RH: sorry, you are right

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
JianLi-RH: Sorry, i don't understand you

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-arm64-stable-4.15-upgrade-from-stable-4.14-aws-ipi-proxy-sts-f28

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
jianlinliu: Something like:
```
<check if the target payload is not recommended, if yes, exit>
run_command "oc adm upgrade --to-image=${TARGET} --allow-explicit-upgrade --force=${FORCE_UPDATE}"
```

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-arm64-stable-4.15-upgrade-from-stable-4.14-azure-ipi-disc-fullyprivate-f28

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -652,7 +652,7 @@ function clear_upgrade() {
 
 # Upgrade the cluster to target release
 function upgrade() {
-    local log_file history_len cluster_src_ver
+    local log_file history_len cluster_src_ver cmd
```
jianlinliu, line 655: we do not need the `cmd` var any more, right?

jianlinliu:
```suggestion
    local log_file history_len cluster_src_ver conditionalUpdates
```

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -685,6 +685,14 @@ function upgrade() {
         clear_upgrade
         check_upgrade_status "${cluster_src_ver}"
     fi
+
+    conditionalUpdates=$(oc get clusterversion version -o json | jq -r '.status.conditionalUpdates[]?.release.version' | xargs)
+    if [[ "$conditionalUpdates" == *"$TARGET_VERSION"* ]]; then
```
jianlinliu, line 690:
```suggestion
    if [[ " $conditionalUpdates " =~ " $TARGET_VERSION " ]]; then
```
That will be more accurate.

I explained it a bit in https://github.com/openshift/release/pull/49540#discussion_r1522866207

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -685,6 +685,14 @@ function upgrade() {
         clear_upgrade
         check_upgrade_status "${cluster_src_ver}"
     fi
+
+    conditional_updates=$(oc get clusterversion version -o json | jq -r '.status.conditionalUpdates[]?.release.version' | xargs)
```
jianlinliu, line 689: Just like what we discussed in slack, I think we can do more:
1. introduce a dedicated step to do the check, opt it into `openshift-upgrade-qe-sanity`, so that we can exit the whole job in advance.
2. sounds like `cucushift-chainupgrade-toimage` also need this check.

ci-operator/step-registry/cucushift/upgrade/checktarget/cucushift-upgrade-checktarget-ref.metadata.json
```diff
@@ -0,0 +1,23 @@
+{
+	"path": "cucushift/upgrade/checktarget/cucushift-upgrade-checktarget-ref.yaml",
```
jianlinliu, line 2:
```suggestion
	"path": "cucushift/upgrade/check-not-recommended-upgrade/check-not-recommended-upgrade-ref.yaml",
```

How about this new name? @jiajliu @shellyyang1989 if you have more good idea, pls comment here.

ci-operator/step-registry/cucushift/upgrade/checktarget/cucushift-upgrade-checktarget-commands.sh
```diff
@@ -0,0 +1,68 @@
...
+mkdir -p /tmp/client
+export OC_DIR="/tmp/client"
+export PATH=${OC_DIR}:$PATH
+
+
+export TARGET="${TARGET_RELEASES[0]}"
+TARGET_VERSION="$(env "NO_PROXY=*" "no_proxy=*" oc adm release info "${TARGET}" --output=json | jq -r '.metadata.version')"
+extract_oc
+
+check_target_version
```
jianlinliu, line 68: A minor suggestion, how about extracting the code in the function here to avoid introducing the function.

JianLi-RH: if the step will not do other works in the future, I am ok to remove to function.

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
```diff
@@ -499,6 +499,12 @@ function admin_ack() {
 
 # Upgrade the cluster to target release
 function upgrade() {
+    local conditional_updates=$(oc get clusterversion version -o json | jq -r '.status.conditionalUpdates[]?.release.version' | xargs)
+    if [[ " $conditionalUpdates " =~ " $TARGET_VERSION " ]]; then
```
jianlinliu, line 503:
```suggestion
    #shellcheck disable=SC2076
    if [[ " $conditionalUpdates " =~ " $TARGET_VERSION " ]]; then
```

ci-operator/step-registry/cucushift/upgrade/checktarget/cucushift-upgrade-checktarget-ref.metadata.json
jiajliu: check-conditional-update? and both of above also sgtm. 

I'm a bit conflicted now, shall we drop condition update? I don't remember why `--allow-not-recommended` var came out? At least, there should be users need it and then it's into our roadmap, right? As it is a supported var and documented in the official doc, then user might use it pending on their assessment. Of course, we could cover it in a dedicated ota case, but condition update is rare compared with regular upgrade, only regular ci could help cover most of condition updates in more paths.

ci-operator/step-registry/openshift/upgrade/qe/sanity/openshift-upgrade-qe-sanity-chain.yaml
```diff
@@ -1,6 +1,7 @@
 chain:
   as: openshift-upgrade-qe-sanity
   steps:
+  - ref: cucushift-upgrade-checktarget
```
jiajliu, line 4: Hmm, if it's exit here, then no following steps. where to add dedicated OTA case for the test?

JianLi-RH: Do we really need to run OTA cases in case this is a not recommended payload?

jiajliu: yes, i think so. there is test case for this checkpoint.

JianLi-RH: cc @shellyyang1989 @jianlinliu for confirm

ci-operator/step-registry/cucushift/upgrade/checktarget/cucushift-upgrade-checktarget-ref.metadata.json
JianLi-RH: @jiajliu I used to have the same concern, but as @shellyyang1989 said there is no user choose a not recommended version.
I read below doc again, it just mentioned `administrator can waive the safety guards and proceed the update`:  
https://docs.openshift.com/container-platform/4.15/updating/updating_a_cluster/updating-cluster-cli.html#update-conditional-upgrade-pathupdating-cluster-cli

jiajliu: so the doc means there will be users update with --allow-not-recommended, right?

shellyyang1989: I meant updating to a not recommended image is supported but I don't think it's a common option that's worth to be covered in common upgrade step. We do have test case to cover the scenario. We can cover it in test case level.

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
```diff
@@ -499,6 +499,13 @@ function admin_ack() {
 
 # Upgrade the cluster to target release
 function upgrade() {
+    local conditional_updates=$(oc get clusterversion version -o json | jq -r '.status.conditionalUpdates[]?.release.version' | xargs)
+    #shellcheck disable=SC2076
+    if [[ " $conditionalUpdates " =~ " $TARGET_VERSION " ]]; then
```
jianlinliu, line 504:
```suggestion
    if [[ " $conditional_updates " =~ " $TARGET_VERSION " ]]; then
```

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -685,6 +685,15 @@ function upgrade() {
...
+    #shellcheck disable=SC2076
+    if [[ " $conditionalUpdates " =~ " $TARGET_VERSION " ]]; then
```
jianlinliu, line 691:
```suggestion
    if [[ " $conditional_updates " =~ " $TARGET_VERSION " ]]; then
```

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-arm64-stable-4.15-upgrade-from-stable-4.14-aws-ipi-disc-priv-sts-ep-f28

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-arm64-stable-4.15-upgrade-from-stable-4.14-aws-ipi-disc-priv-sts-ep-f28

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-arm64-stable-4.15-upgrade-from-stable-4.14-aws-ipi-proxy-sts-f28

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-arm64-stable-4.15-upgrade-from-stable-4.14-aws-ipi-proxy-sts-f28

ci-operator/step-registry/openshift/upgrade/qe/sanity/openshift-upgrade-qe-sanity-chain.yaml
jianlinliu: This is to resolve the regular upgrade ci issue to exit in advance. Now we have a dedicated OTA ci yaml, when we need to automate the OTA test case, we can organize a dedicated step chains in a test job in OTA dedicated ci yaml.

JianLi-RH: Yes, agree. so we can keep current step orders

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-stable-4.15-upgrade-from-stable-4.14-aws-ipi-proxy-sts-f28

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-stable-4.15-upgrade-from-stable-4.14-aws-ipi-proxy-sts-f28

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-arm64-stable-4.15-upgrade-from-stable-4.14-aws-ipi-proxy-sts-f28

ci-operator/step-registry/cucushift/upgrade/checktarget/cucushift-upgrade-checktarget-ref.metadata.json
jianlinliu: Back to the topic of the step naming, because `checktarget` is a bit common word, not easy to understanding what purpose for this step from its name, so that is why I suggest to rename it to `check-not-recommended`.

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
```diff
@@ -499,6 +499,14 @@ function admin_ack() {
 
 # Upgrade the cluster to target release
 function upgrade() {
+    local conditional_updates
+    conditional_updates=$(oc get clusterversion version -o json | jq -r '.status.conditionalUpdates[]?.release.version' | xargs)
+    #shellcheck disable=SC2076
+    if [[ " $conditional_updates " =~ " $TARGET_VERSION " ]]; then
+        echo "Error: $TARGET_VERSION is not recommended, for details please refer:"
+        oc get clusterversion version -o json | jq -r '.status.conditionalUpdates'
+        exit 1
```
shellyyang1989, line 508: Is exit 0 better?

shellyyang1989: Ah, I get it. We want to interrupt the test so exiting 1.

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-stable-4.15-upgrade-from-stable-4.14-aws-ipi-proxy-sts-f28

jianlinliu:
Most looks good to me, is there a pj-rehearse job link to help me confirm if this PR works as our expectation?

JianLi-RH:
> Most looks good to me, is there a pj-rehearse job link to help me confirm if this PR works as our expectation?

No, its running, will ping you when it's done

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-nightly-4.15-upgrade-from-stable-4.7-gcp-ipi-f30

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-stable-4.15-upgrade-from-stable-4.14-vsphere-ipi-proxy-fips-f14

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-stable-4.15-upgrade-from-stable-4.14-vsphere-ipi-proxy-fips-f14

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-stable-4.15-upgrade-from-stable-4.14-nutanix-ipi-proxy-fips-f14

JianLi-RH:
Job does not fail when the target payload is not recommended: https://qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/qe-private-deck/pr-logs/pull/openshift_release/49540/rehearse-49540-periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-stable-4.15-upgrade-from-stable-4.14-aws-ipi-proxy-sts-f28/1770763208676610048

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-stable-4.15-upgrade-from-stable-4.14-aws-ipi-proxy-sts-f28

jianlinliu:
> Job does not fail when the target payload is not recommended: https://qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/qe-private-deck/pr-logs/pull/openshift_release/49540/rehearse-49540-periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-stable-4.15-upgrade-from-stable-4.14-aws-ipi-proxy-sts-f28/1770763208676610048

```
conditional_updates=$(oc get clusterversion version -o json | jq -r '.status.conditionalUpdates[]?.release.version' | xargs)
```

Maybe only when a right target channel is set, the `conditionalUpdates` can be retrieved? cc @jiajliu @shellyyang1989

shellyyang1989:
> > Job does not fail when the target payload is not recommended: https://qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/qe-private-deck/pr-logs/pull/openshift_release/49540/rehearse-49540-periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-stable-4.15-upgrade-from-stable-4.14-aws-ipi-proxy-sts-f28/1770763208676610048
> 
> ```
> conditional_updates=$(oc get clusterversion version -o json | jq -r '.status.conditionalUpdates[]?.release.version' | xargs)
> ```
> 
> Maybe only when a right target channel is set, the `conditionalUpdates` can be retrieved? cc @jiajliu @shellyyang1989

Right. We should use candidate-target-version channel.

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-stable-4.15-upgrade-from-stable-4.14-gcp-ipi-proxy-private-f14

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__amd64-stable-4.15-upgrade-from-stable-4.14.yaml
```diff
@@ -153,7 +147,12 @@ tests:
     env:
       ENABLE_OTA_TEST: OCP-24358
     test:
-    - chain: openshift-upgrade-qe-sanity
+    - ref: cucushift-upgrade-healthcheck
+    - ref: cucushift-upgrade-setchannel
```
jianlinliu, line 151: Once this step is confirmed that it can be fixed the issue, maybe also need some update on the step, remove the [lines](https://github.com/openshift/release/blob/master/ci-operator/step-registry/cucushift/upgrade/setchannel/cucushift-upgrade-setchannel-commands.sh#L75-L90), to avoid unexpected exit when the target does not show up in `availableUpdates` yet. Just make the step only do one thing - set channel. @jiajliu WDYT?

jiajliu: why we need such steps for case OCP-24358? i mean cucushift-upgrade-setchannel and others?

jiajliu: only `cucushift-upgrade-toversion` need `cucushift-upgrade-setchannel`?

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-stable-4.15-upgrade-from-stable-4.14-gcp-ipi-proxy-private-f14

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__amd64-stable-4.15-upgrade-from-stable-4.14.yaml
jianlinliu: https://github.com/openshift/release/pull/49540#issuecomment-2014315633

I think Jian is using this job to run a test to confirm if the solution works.

jiajliu: Ah, for one-time test, then sounds like no need any change, right? Just do some hack in this pr for the test.

jiajliu: or do u mean, we always need setchannel for this test, it sounds not necessary. 
about https://github.com/openshift/release/pull/49540#issuecomment-2014315633

I think, if it's for a dedicated upgrade test with conditional update only, then we need setchannel, and also need do the upgrade with toversion.
if it's for a status check on live cluster, no need to set channel. since it's a workaround for the issue we hit in regular ci without any channel update in upgrade steps. we only need check if condition update available on the live cluster even do upgrade with toimage and then skip it if matched.

jiajliu: I thought `cucushift-upgrade-check-not-recommended` only resolve the issue that the upgrade is blocked due to error `the update 4.15.0 is not one of the recommended updates` which is already present on the live cluster. After discussion, the scope is more than that, agree to move the check out of setchannel for more scenarios. 

Update it in https://github.com/openshift/release/pull/50128. cc @JianLi-RH

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
```diff
@@ -499,6 +499,16 @@ function admin_ack() {
 
 # Upgrade the cluster to target release
 function upgrade() {
+    local conditional_updates
```
jiajliu, line 502: we also need add set_channal step here since `cucushift-upgrade-setchannel` step could not be call in the chain loop

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -685,6 +685,17 @@ function upgrade() {
         clear_upgrade
         check_upgrade_status "${cluster_src_ver}"
     fi
+
+    conditional_updates=$(oc get clusterversion version -o json | jq -r '.status.conditionalUpdates[]?.release.version' | xargs)
```
jiajliu, line 689: Since line689-line697 is already split out to be a dedicated step, i would suggest to remove it in `cucushift-upgrade-toimage`.

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__amd64-stable-4.15-upgrade-from-stable-4.14.yaml
JianLi-RH: @jiajliu @jianlinliu  this is one time test, before merging this pr, I will remove all changes in ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__amd64-stable-4.15-upgrade-from-stable-4.14.yaml

ci-operator/step-registry/cucushift/upgrade/check-not-recommended/cucushift-upgrade-check-not-recommended-commands.sh
```diff
@@ -0,0 +1,66 @@
...
+# Setup proxy if it's present in the shared dir
+if [[ -f "${SHARED_DIR}/proxy-conf.sh" ]]; then
+    # shellcheck disable=SC1091
+    source "${SHARED_DIR}/proxy-conf.sh"
+fi
+
+# Get the target upgrades release, by default, OPENSHIFT_UPGRADE_RELEASE_IMAGE_OVERRIDE is the target release
+# If it's serial upgrades then override-upgrade file will store the release and overrides OPENSHIFT_UPGRADE_RELEASE_IMAGE_OVERRIDE
+# upgrade-edge file expects a comma separated releases list like target_release1,target_release2,...
+export TARGET_RELEASES=("${OPENSHIFT_UPGRADE_RELEASE_IMAGE_OVERRIDE}")
```
jianlinliu, line 39: I am tidying up the similar code in https://github.com/openshift/release/pull/50219/files#diff-a71e3444f4cc5e3d7ecb8ebd36fa8bd1fda181a0806d09523b318eadffaebc71L756-L765, so I think you also can do the same thing here to keep consistent. I think you can directly drop these lines, and use `export TARGET="${OPENSHIFT_UPGRADE_RELEASE_IMAGE_OVERRIDE}"` on line 54.

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-stable-4.15-upgrade-from-stable-4.14-vsphere-ipi-proxy-fips-f14

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-stable-4.15-upgrade-from-stable-4.14-gcp-ipi-proxy-private-f14

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-stable-4.15-upgrade-from-stable-4.14-gcp-ipi-proxy-private-f14

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-stable-4.15-upgrade-from-stable-4.14-gcp-ipi-proxy-private-f14

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-stable-4.15-upgrade-from-stable-4.14-vsphere-ipi-proxy-fips-f14

JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-stable-4.15-upgrade-from-stable-4.14-nutanix-ipi-proxy-fips-f14

PR #49589, opened 2024-03-06
Enable 4.15 ppc64le job on a different region

aishwaryabk:
cc: @juliemathew @aleskandro


liangxia:
> @liangxia this is still another case where we'd need the cron rotation to happen across all the YAML files and more deterministically, instead of using a random approach: the cluster profile only allows 1 concurrent cluster, so when two jobs overlap with each other, the second (in terms of start time) will always fail.

@aleskandro Since we have only 1 quota for this, i guess we do not have too much profiles (Currently 4 profiles) on it.  So perhaps we can manage it manually currently ?

aleskandro:
yes @liangxia , just an fyi, that eventually we should  think another algorithm for such kind of profiles (bm, and p/z ones currently)

PR #49599, opened 2024-03-06
OCPQE-18644: Wipe disk always

sgoveas:
If jobs running from PR or in other ways is interrupted may result in the file `"${SHARED_DIR}/CLUSTER_INSTALL_START_TIME"` not getting created. This causes the wipe to fail as it checked for the existence of the file to decide how to proceed. Removing this check will enforce wipe always.


aleskandro (approved):
Thanks @sgoveas

PR #49658, opened 2024-03-07
Revert "OCPBUGS-23957: Remove permafailing local-gateway jobs"

jluhrsen:
Reverts openshift/release#48688

we have a bug to track these permafailing jobs and sounds like we need to address it and get
it fixed. Will want these jobs back.


dtantsur:
Please only re-add the jobs when you have at least an idea how to fix them. The last thing we should do is to re-add permafailing jobs and then hope that the bug gets fixed one day.

PR #49699, opened 2024-03-11
OCPQE-18638: Add slack notification for automated release jobs

liangxia:


rioliu-rh:
@liangxia  only one concern. Do we need to send notification for successful job?

liangxia:
> @liangxia only one concern. Do we need to send notification for successful job?

To make sure we can get notification for those jobs. E.g, if we do not send the notification for the success jobs, and the job success in the coming 10 days. It's hard to know whether the notification works as expected or not (the notification not working or due to no failure jobs).
We can remove the notification for those success jobs later after we confirm the notification work as expected.

PR #49728, opened 2024-03-11
Implement check requirements test for ironic-image

elfosardo:


elfosardo:
/hold
need to add release variable

elfosardo:
/pj-rehearse ack
looks like it's working as it should

PR #49739, opened 2024-03-11
add cidr-expansion job to 4.16/4.17 for CNO

jluhrsen:
these were unintentionally removed here:
https://github.com/openshift/release/pull/49568/commits/d13018f90d0e46152a481f6e149568c382c0f048


jluhrsen:
not needed. master==4.16 right now and there is a periodic for master. I've learned that when the next branch cuts the periodic will stay in the master file which will then be 4.17 and a 4.16 job will be created for that file.

PR #49741, opened 2024-03-12
Use specific cluster profile for automated release

liangxia:
We are heavily using the original cluster profile `aws-1-qe`, sometimes it take up to hours to acquiring leases from this profile, so using a specific profile for automated release to speed up lease acquiring.

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__automated-release-stable-4.15-upgrade-from-stable-4.15.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__automated-release-stable-4.15-upgrade-from-stable-4.15.yaml
@@ -65,7 +65,7 @@ tests:
 - as: aws-ipi-fips-arm-mixarch-f1
   cron: 0 10 * * *
   steps:
-    cluster_profile: aws-qe
+    cluster_profile: aws-autorelease-qe
     env:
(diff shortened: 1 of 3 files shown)


ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__automated-release-stable-4.15-upgrade-from-stable-4.15.yaml
```diff
@@ -65,7 +65,7 @@ tests:
 - as: aws-ipi-fips-arm-mixarch-f1
   cron: 0 10 * * *
```
yunjiang29, line 66: @liangxia, we need to complete these in working hour, right?

yunjiang29: I'm thinking if we can manually update cron job settings, as we currently have 2 slices, if more jobs run in the daytime, we will hit quota issues.

liangxia: For automated release, we mainly use api to trigger this job.  And currently, we just let this job run everyday, to evaluate the pass ratio of the job and the test cases in it.

PR #49743, opened 2024-03-12
auto ocp-53907

jiajliu:
1. create dedicate ota job to avoid noise in regular ci, refer to https://issues.redhat.com/browse/OCPQE-20055
2. add upgrade-toversion to include both --to and --to-latest
3. auto ocp-53907
4. split set_channel as a dedicated step upgrade-setchannel

### DIFF
--- a/ci-operator/config/openshift/verification-tests/openshift-verification-tests-master__ota-multi-stable-4.15-upgrade-from-stable-4.14.yaml
+++ b/ci-operator/config/openshift/verification-tests/openshift-verification-tests-master__ota-multi-stable-4.15-upgrade-from-stable-4.14.yaml
@@ -60,7 +60,7 @@ resources:
       memory: 200Mi
 tests:
 - as: azure-ipi-amd-mixarch-f28
-  cron: 54 6 14,28 * *
+  cron: 36 20 11 * *
   steps:
     cluster_profile: azure-qe
     env:
@@ -71,10 +71,11 @@ tests:
       ENABLE_OTA_TEST: OCP-53907
     test:
     - ref: ipi-install-heterogeneous
-    - ref: cucushift-upgrade-healthcheck
+    - chain: cucushift-installer-check-cluster-health
+    - ref: cucushift-upgrade-setchannel
     - ref: cucushift-ota-preupgrade
     - ref: cucushift-upgrade-toversion
-    - chain: cucushift-installer-check-cluster-health
+    - ref: cucushift-upgrade-healthcheck
     - ref: cucushift-ota-postupgrade
     workflow: cucushift-installer-rehearse-azure-ipi
 zz_generated_metadata:
--- a/ci-operator/config/openshift/verification-tests/openshift-verification-tests-master__ota-multi-stable-4.16-upgrade-from-stable-4.15.yaml
+++ b/ci-operator/config/openshift/verification-tests/openshift-verification-tests-master__ota-multi-stable-4.16-upgrade-from-stable-4.15.yaml
@@ -47,12 +47,11 @@ releases:
       channel: fast
       version: "4.15"
   target:
-    prerelease:
+    candidate:
       architecture: multi
       product: ocp
-      version_bounds:
-        lower: 4.16.0-0
-        upper: 4.17.0-0
+      stream: nightly
+      version: "4.16"
 resources:
   '*':
     requests:
@@ -60,7 +59,7 @@ resources:
       memory: 200Mi
 tests:
 - as: azure-ipi-amd-mixarch-f28
-  cron: 54 6 14,28 * *
+  cron: 26 14 16 * *
   steps:
     cluster_profile: azure-qe
     env:
@@ -71,10 +70,11 @@ tests:
       ENABLE_OTA_TEST: OCP-53907
     test:
     - ref: ipi-install-heterogeneous
-    - ref: cucushift-upgrade-healthcheck
+    - chain: cucushift-installer-check-cluster-health
+    - ref: cucushift-upgrade-setchannel
     - ref: cucushift-ota-preupgrade
     - ref: cucushift-upgrade-toversion
-    - chain: cucushift-installer-check-cluster-health
+    - ref: cucushift-upgrade-healthcheck
     - ref: cucushift-ota-postupgrade
     workflow: cucushift-installer-rehearse-azure-ipi
 zz_generated_metadata:
--- a/ci-operator/step-registry/cucushift/ota/preupgrade/cucushift-ota-preupgrade-commands.sh
+++ b/ci-operator/step-registry/cucushift/ota/preupgrade/cucushift-ota-preupgrade-commands.sh
@@ -241,36 +241,7 @@ function pre-OCP-53921(){
 
 function pre-OCP-53907(){
     echo "Test Start: ${FUNCNAME[0]}"
-    local version
-    version="$(oc get clusterversion --no-headers | awk '{print $2}')"
-    if [ -z "${version}" ] ; then
-        echo "Fail to get cluster version!"
-        return 1
-    fi
-    x_ver=$( echo "${version}" | cut -f1 -d. )
-    y_ver=$( echo "${version}" | cut -f2 -d. )
-    y_ver=$((y_ver+1))
-    ver="${x_ver}.${y_ver}"
-    if ! oc adm upgrade channel candidate-${ver}; then
-        echo "Fail to change channel to candidate-${ver}!"
-        return 1
-    fi
-    local retry=3
-    while (( retry > 0 ));do
-        recommends=$(oc get clusterversion version -o json|jq -r '.status.availableUpdates')
-        if [[ "${recommends}" == "null" ]]; then
-            (( retry -= 1 ))
-            sleep 60
-            echo "No recommended update available! Retry..."
-        else
-            echo "Recommencded update: ${recommends}"
-            break
-        fi
-    done
-    if (( retry == 0 )); then
-        echo "Timeout to get recommended update!" 
-        return 1
-    fi
+    recommends=$(oc get clusterversion version -o json|jq -r '.status.availableUpdates')
     mapfile -t images < <(echo ${recommends}|jq -r '.[].image')
     if [ -z "${images[*]}" ]; then
         echo "No image extracted from recommended update!"
--- a/ci-operator/step-registry/cucushift/upgrade/setchannel/cucushift-upgrade-setchannel-commands.sh
+++ b/ci-operator/step-registry/cucushift/upgrade/setchannel/cucushift-upgrade-setchannel-commands.sh
@@ -1,0 +1,91 @@
+#!/bin/bash
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+export HOME="${HOME:-/tmp/home}"
+export XDG_RUNTIME_DIR="${HOME}/run"
+export REGISTRY_AUTH_PREFERENCE=podman # TODO: remove later, used for migrating oc from docker to podman
+mkdir -p "${XDG_RUNTIME_DIR}"
+# After cluster is set up, ci-operator make KUBECONFIG pointing to the installed cluster,
+# to make "oc registry login" interact with the build farm, set KUBECONFIG to empty,
+# so that the credentials of the build farm registry can be saved in docker client config file.
+# A direct connection is required while communicating with build-farm, instead of through proxy
+KUBECONFIG="" oc --loglevel=8 registry login
+
+# Extract oc binary which is supposed to be identical with target release
+function extract_oc(){
+    echo -e "Extracting oc\n"
+    local retry=5 tmp_oc="/tmp/client-2"
+    mkdir -p ${tmp_oc}
+    while ! (env "NO_PROXY=*" "no_proxy=*" oc adm release extract -a "${CLUSTER_PROFILE_DIR}/pull-secret" --command=oc --to=${tmp_oc} ${DUMMY_TARGET});
+    do
+        echo >&2 "Failed to extract oc binary, retry..."
+        (( retry -= 1 ))
+        if (( retry < 0 )); then return 1; fi
+        sleep 60
+    done
+    mv ${tmp_oc}/oc ${OC_DIR} -f
+    which oc
+    oc version --client
+    return 0
+}
+
+valid_channels=("fast" "stable" "candidate" "eus")
+valid_channel="false"
+for channel in ${valid_channels[*]}; do
+    if [[ "${UPGRADE_CHANNEL}" == "${channel}" ]]; then
+        valid_channel="true" 
+        break
+    fi
+done
+if [[ "${valid_channel}" == "false" ]]; then
+    echo "Specified channel ${UPGRADE_CHANNEL} is not valid, upgrade will not start due to fail to set correct channel!"
+    exit 1
+fi
+
+if [[ -f "${SHARED_DIR}/kubeconfig" ]] ; then
+    export KUBECONFIG=${SHARED_DIR}/kubeconfig
+fi
+
+# Setup proxy if it's present in the shared dir
+if [[ -f "${SHARED_DIR}/proxy-conf.sh" ]]; then
+    # shellcheck disable=SC1091
+    source "${SHARED_DIR}/proxy-conf.sh"
+fi
+
+mkdir -p /tmp/client
+export OC_DIR="/tmp/client"
+export PATH=${OC_DIR}:$PATH
+export DUMMY_TARGET="${OPENSHIFT_UPGRADE_RELEASE_IMAGE_OVERRIDE}"
+# we need this DUMMY_TARGET_VERSION from ci config to download oc
+DUMMY_TARGET_VERSION="$(env "NO_PROXY=*" "no_proxy=*" oc adm release info "${DUMMY_TARGET}" --output=json | jq -r '.metadata.version')"
+echo "Target version of ci config is: ${DUMMY_TARGET_VERSION}"
+extract_oc
+
+x_ver=$( echo "${DUMMY_TARGET_VERSION}" | cut -f1 -d. )
+y_ver=$( echo "${DUMMY_TARGET_VERSION}" | cut -f2 -d. )
+ver="${x_ver}.${y_ver}"
+target_channel="${UPGRADE_CHANNEL}-${ver}"
+if ! oc adm upgrade channel ${target_channel}; then
+    echo "Fail to change channel to ${target_channel}!"
+    exit 1
+fi
+retry=3
+while (( retry > 0 ));do
+    recommends=$(oc get clusterversion version -o json|jq -r '.status.availableUpdates[]?.version'| xargs)
+    if [[ "${recommends}" == "null" ]] || [[ "${recommends}" != *"${ver}"* ]]; then
+        (( retry -= 1 ))
+        sleep 60
+        echo "No recommended update available! Retry..."
+    else
+        echo "Recommencded update: ${recommends}"
+        break
+    fi
+done
+if (( retry == 0 )); then
+    echo "Timeout to get recommended update!" 
+    exit 1
+fi
+
--- a/ci-operator/step-registry/cucushift/upgrade/toversion/cucushift-upgrade-toversion-commands.sh
+++ b/ci-operator/step-registry/cucushift/upgrade/toversion/cucushift-upgrade-toversion-commands.sh
@@ -186,7 +186,7 @@ function extract_oc(){
     echo -e "Extracting oc\n"
     local retry=5 tmp_oc="/tmp/client-2"
     mkdir -p ${tmp_oc}
-    while ! (env "NO_PROXY=*" "no_proxy=*" oc adm release extract -a "${CLUSTER_PROFILE_DIR}/pull-secret" --command=oc --to=${tmp_oc} ${TARGET});
+    while ! (env "NO_PROXY=*" "no_proxy=*" oc adm release extract -a "${CLUSTER_PROFILE_DIR}/pull-secret" --command=oc --to=${tmp_oc} ${DUMMY_TARGET});
     do
         echo >&2 "Failed to extract oc binary, retry..."
         (( retry -= 1 ))
@@ -302,46 +302,6 @@ function admin_ack() {
     fi
 }
 
-# check if any of cases is enabled via ENABLE_OTA_TEST
-function check_ota_case_enabled() {
-    local case_id
-    local cases_array=("$@")
-    for case_id in "${cases_array[@]}"; do
-        # shellcheck disable=SC2076
-        if [[ " ${ENABLE_OTA_TEST} " =~ " ${case_id} " ]]; then
-            echo "${case_id} is enabled via ENABLE_OTA_TEST on this job."
-            return 0
-        fi
-    done
-    return 1
-}
-
-function set_channel(){
-    x_ver=$( echo "${DUMMY_TARGET_VERSION}" | cut -f1 -d. )
-    y_ver=$( echo "${DUMMY_TARGET_VERSION}" | cut -f2 -d. )
-    ver="${x_ver}.${y_ver}"
-    if ! oc adm upgrade channel candidate-${ver}; then
-        echo "Fail to change channel to candidate-${ver}!"
-        return 1
-    fi
-    local retry=3
-    while (( retry > 0 ));do
-        recommends=$(oc get clusterversion version -o json|jq -r '.status.availableUpdates')
-        if [[ "${recommends}" == "null" ]]; then
-            (( retry -= 1 ))
-            sleep 60
-            echo "No recommended update available! Retry..."
-        else
-            echo "Recommencded update: ${recommends}"
-            break
-        fi
-    done
-    if (( retry == 0 )); then
-        echo "Timeout to get recommended update!" 
-        return 1
-    fi
-}
-
 # judge if upgrade with --to or --to-latest cmd
 # set correct target_version for the upgrade
 function set_target_and_upgrade_cmd(){
@@ -350,21 +310,20 @@ function set_target_and_upgrade_cmd(){
         echo "No recommended update available!"
         exit 1
     fi
-    # for "oc adm upgrade --to x.y.z" and "oc adm upgrade --to $latest_version_in_available_update"
-    upgrade_cmd="oc adm upgrade --to ${TARGET_VERSION}"
     # x.y.z: run upgrade with "oc adm upgrade --to x.y.z";
-    if [[ -n "${UPGRAGE_TO_VERSION}" ]] && [[ "${UPGRAGE_TO_VERSION}" != "latest" ]]; then
+    if [[ -n "${UPGRADE_TO_VERSION}" ]] && [[ "${UPGRADE_TO_VERSION}" != "latest" ]]; then
         valid_target="false"
         # check if x.y.z is in available_versions, if not, it's invalid target with --to.
         for version in ${available_versions[*]}; do
-            if [[ "${UPGRAGE_TO_VERSION}" == "${version}" ]]; then
-                TARGET_VERSION="${UPGRAGE_TO_VERSION}"
+            if [[ "${UPGRADE_TO_VERSION}" == "${version}" ]]; then
+                TARGET_VERSION="${UPGRADE_TO_VERSION}"
                 valid_target="true"
+                upgrade_cmd="oc adm upgrade --to ${TARGET_VERSION}"
                 break
             fi
         done
         if [[ "${valid_target}" == "false" ]]; then
-            echo "Specified target version ${UPGRAGE_TO_VERSION} is not in recommended update, upgrade will not continue!"
+            echo "Specified target version ${UPGRADE_TO_VERSION} is not in recommended update, upgrade will not continue!"
             exit 1
         fi
     # get target_version from available updates
@@ -376,7 +335,7 @@ function set_target_and_upgrade_cmd(){
         # when there is both 4.15.0-rc1 and 4.15.0-ec1, sort will return 4.15.0-rc1 as latest(expected),
         # there will not be 4.15.0 and 4.15.0-ec1 only, ignore
         # so here we need correct unexpected latest with 4.15.0
-        if [[ "${TARGET_VERSION}" =~ "-rc" ]]; then
+        if [[ "${TARGET_VERSION}" =~ "-" ]]; then
             expected_target="${TARGET_VERSION%-*}"
             for version in ${available_versions[*]}; do
                 if [[ "${version}" == "${expected_target}" ]]; then
@@ -386,14 +345,24 @@ function set_target_and_upgrade_cmd(){
             done
         fi
         # latest: run upgrade with "oc adm upgrade --to-latest";
-        if [[ "${UPGRAGE_TO_VERSION}" == "latest" ]]; then
+        if [[ "${UPGRADE_TO_VERSION}" == "latest" ]]; then
             upgrade_cmd="oc adm upgrade --to-latest"
+        # "": run upgrade with "oc adm upgrade --to $latest_version_in_available_update"
+        else
+            upgrade_cmd="oc adm upgrade --to ${TARGET_VERSION}"
         fi
     fi
+
     TARGET_MINOR_VERSION="$(echo "${TARGET_VERSION}" | cut -f2 -d.)"
+    TARGET=$(oc get clusterversion version -o json| jq -r --arg z "${TARGET_VERSION}" '.status.availableUpdates[]? | select(.version==$z).image')
+    if [[ "${TARGET}" == null ]] || [[ -z "${TARGET}" ]]; then
+        echo "The target version ${TARGET_VERSION} is a bad update without image!"
+        exit 1
+    fi
     export TARGET_MINOR_VERSION
     export TARGET_VERSION
-    echo "Target release version is: ${TARGET_VERSION}\n"
+    export TARGET
+    echo "Target release version is: ${TARGET_VERSION}"
 }
 
 # Monitor the upgrade status
@@ -449,21 +418,18 @@ mkdir -p /tmp/client
 export OC_DIR="/tmp/client"
 export PATH=${OC_DIR}:$PATH
 
-export TARGET="${OPENSHIFT_UPGRADE_RELEASE_IMAGE_OVERRIDE}"
+export DUMMY_TARGET="${OPENSHIFT_UPGRADE_RELEASE_IMAGE_OVERRIDE}"
 # we need this DUMMY_TARGET_VERSION from ci config to download oc and do some pre-check
-DUMMY_TARGET_VERSION="$(env "NO_PROXY=*" "no_proxy=*" oc adm release info "${TARGET}" --output=json | jq -r '.metadata.version')"
-echo "Target version of ci config is: ${DUMMY_TARGET_VERSION}\n"
+DUMMY_TARGET_VERSION="$(env "NO_PROXY=*" "no_proxy=*" oc adm release info "${DUMMY_TARGET}" --output=json | jq -r '.metadata.version')"
+echo "Target version of ci config is: ${DUMMY_TARGET_VERSION}"
 extract_oc
 
 SOURCE_VERSION="$(oc get clusterversion --no-headers | awk '{print $2}')"
 SOURCE_MINOR_VERSION="$(echo "${SOURCE_VERSION}" | cut -f2 -d.)"
 export SOURCE_VERSION
 export SOURCE_MINOR_VERSION
-echo "Source release version is: ${SOURCE_VERSION}\n"
+echo "Source release version is: ${SOURCE_VERSION}"
 
-if ! check_ota_case_enabled "OCP-53907"; then
-    set_channel
-fi
 set_target_and_upgrade_cmd
 export FORCE_UPDATE="false"
 if ! check_signed; then
--- a/ci-operator/jobs/openshift/verification-tests/openshift-verification-tests-master-periodics.yaml
+++ b/ci-operator/jobs/openshift/verification-tests/openshift-verification-tests-master-periodics.yaml
@@ -8835,7 +8835,7 @@ periodics:
         secretName: result-aggregator
 - agent: kubernetes
   cluster: build04
-  cron: 54 6 14,28 * *
+  cron: 36 20 11 * *
   decorate: true
   decoration_config:
     skip_cloning: true
@@ -8915,7 +8915,7 @@ periodics:
         secretName: result-aggregator
 - agent: kubernetes
   cluster: build04
-  cron: 54 6 14,28 * *
+  cron: 26 14 16 * *
   decorate: true
   decoration_config:
     skip_cloning: true
--- a/ci-operator/step-registry/cucushift/upgrade/setchannel/OWNERS
+++ b/ci-operator/step-registry/cucushift/upgrade/setchannel/OWNERS
@@ -1,0 +1,10 @@
+approvers:
+  - jianlinliu
+  - jiajliu
+  - shellyyang1989
+  - jhou1
+reviewers:
+  - jiajliu
+  - jianlinliu
+  - shellyyang1989
+  - jhou1
--- a/ci-operator/step-registry/cucushift/upgrade/setchannel/cucushift-upgrade-setchannel-ref.metadata.json
+++ b/ci-operator/step-registry/cucushift/upgrade/setchannel/cucushift-upgrade-setchannel-ref.metadata.json
@@ -1,0 +1,17 @@
+{
+	"path": "cucushift/upgrade/setchannel/cucushift-upgrade-setchannel-ref.yaml",
+	"owners": {
+		"approvers": [
+			"jianlinliu",
+			"jiajliu",
+			"shellyyang1989",
+			"jhou1"
+		],
+		"reviewers": [
+			"jiajliu",
+			"jianlinliu",
+			"shellyyang1989",
+			"jhou1"
+		]
+	}
+}
--- a/ci-operator/step-registry/cucushift/upgrade/setchannel/cucushift-upgrade-setchannel-ref.yaml
+++ b/ci-operator/step-registry/cucushift/upgrade/setchannel/cucushift-upgrade-setchannel-ref.yaml
@@ -1,0 +1,23 @@
+ref:
+  as: cucushift-upgrade-setchannel
+  from_image:
+    namespace: ci
+    name: verification-tests
+    tag: latest
+  cli: latest
+  commands: cucushift-upgrade-setchannel-commands.sh
+  resources:
+    limits:
+      cpu: "1"
+    requests:
+      cpu: 100m
+      memory: 200Mi
+  env:
+  - name: UPGRADE_CHANNEL
+    default: "candidate"
+    documentation: |-
+      This env is used to set cv.spec.channel, which will be used to get avialable update from upstream.
+      There are 4 valid channels, refer to https://docs.openshift.com/container-platform/4.15/updating/understanding_updates/understanding-update-channels-release.html
+  dependencies:
+  - name: "release:target"
+    env: "OPENSHIFT_UPGRADE_RELEASE_IMAGE_OVERRIDE"
--- a/ci-operator/step-registry/cucushift/upgrade/toversion/cucushift-upgrade-toversion-ref.yaml
+++ b/ci-operator/step-registry/cucushift/upgrade/toversion/cucushift-upgrade-toversion-ref.yaml
@@ -19,7 +19,7 @@ ref:
     default: "false"
     documentation: |-
       Detemine if OTA test will be included
-  - name: UPGRAGE_TO_VERSION
+  - name: UPGRADE_TO_VERSION
     default: "latest"
     documentation: |-
       latest: run upgrade with "oc adm upgrade --to-latest";


jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-multi-stable-4.15-upgrade-from-stable-4.14-azure-ipi-amd-mixarch-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-multi-stable-4.15-upgrade-from-stable-4.14-azure-ipi-amd-mixarch-f28

jiajliu:
cc @jianlinliu @shellyyang1989 review, thanks!

ci-operator/step-registry/cucushift/ota/postupgrade/cucushift-ota-postupgrade-commands.sh
```diff
@@ -133,6 +133,17 @@ function post-OCP-21588(){
     return $ret
 }
 
+function post-OCP-53907(){
+    echo "Test Start: ${FUNCNAME[0]}"
+    local arch 
+    arch=$(oc get clusterversion version -ojson|jq -r '.status.conditions[]|select(.type == "ReleaseAccepted")|.message')
+    if [[ "${arch}" != *"Multi"* ]]; then
```
jianlinliu, line 140:
```suggestion
    if [[ "${arch}" != *architecture="Multi"* ]]; then
```
To be more accurate?

jianlinliu: btw, can we also check `ReleaseAccepted`'s status to be True for one more guard.

ci-operator/step-registry/cucushift/ota/preupgrade/cucushift-ota-preupgrade-commands.sh
```diff
@@ -239,6 +239,61 @@ function pre-OCP-53921(){
...
+    if [ -z "${images[*]}" ]; then
+        echo "No image extracted from recommended update!"
+        return 1
+    fi
+    for image in ${images[*]}; do
+        if [[ "${image}" == "null" ]] ; then
+            echo "No image info!"
+            return 1
+        fi
+        metadata=$(oc adm release info ${image} -ojson|jq .metadata.metadata)
```
jianlinliu, line 284: can we directly get `.release.openshift.io/architecture` here to be more accurate?

ci-operator/step-registry/cucushift/ota/preupgrade/cucushift-ota-preupgrade-commands.sh
```diff
@@ -239,6 +239,61 @@ function pre-OCP-53921(){
...
+    version="$(oc get clusterversion --no-headers | awk '{print $2}')"
+    if [ -z "${version}" ] ; then
+        echo "Fail to get cluster version!"
+        return 1
+    fi
+    x_ver=$( echo "${version}" | cut -f1 -d. )
+    y_ver=$( echo "${version}" | cut -f2 -d. )
+    y_ver=$((y_ver+1))
+    ver="${x_ver}.${y_ver}"
+    if ! oc adm upgrade channel candidate-${ver}; then
```
jianlinliu, line 254: I have some concern here, generally the target payload would be provided by ci config level, while in this testing, it depends on official candidate channel, is there a unmatched issue?

ci-operator/step-registry/cucushift/ota/postupgrade/cucushift-ota-postupgrade-commands.sh
jiajliu: Sounds like no need. ReleaseAccepted's status means no failures blocks an upgrade happen(for example, image verification or precondition check,etc). It's a post check about arch after upgrade succeed, which ensured in toimage step.

ci-operator/step-registry/cucushift/ota/preupgrade/cucushift-ota-preupgrade-commands.sh
jiajliu: yes, we can. but seems not here, but add one more line after the judgement of metadata=null.

ci-operator/step-registry/cucushift/ota/preupgrade/cucushift-ota-preupgrade-commands.sh
jiajliu: Yes, possible.  The ci config is choosing the latest one, and the upgrade is using --tolatest, it should be the same.  There might be some edge case caused they are unmatched. But even it's unmatched, the upgrade path is also valid for this case's test.

ci-operator/step-registry/cucushift/ota/preupgrade/cucushift-ota-preupgrade-commands.sh
jianlinliu:
```
a=$(oc adm release info quay.io/openshift-release-dev/ocp-release:4.15.2-x86_64 -ojson|jq -r '.metadata.metadata."release.openshift.io/architecture"'); echo $a
null
```
Once that, we can just need one condition to determinate the testing failed or not.

jiajliu: It can not distinguish if it's a null metadata or a null architecture, right? when metadata=null, it should continue to next one.

```
# ./oc adm release info quay.io/openshift-release-dev/ocp-release@sha256:0ed362d46466541178ad24c849e9aeacb0c622b5a2cab85cba820330e01a29a1 -ojson|jq -r '.metadata.metadata."release.openshift.io/architecture"'
null
# ./oc adm release info quay.io/openshift-release-dev/ocp-release@sha256:0ed362d46466541178ad24c849e9aeacb0c622b5a2cab85cba820330e01a29a1 -ojson|jq .metadata
{
  "kind": "cincinnati-metadata-v0",
  "version": "4.16.0-ec.1",
  "previous": [
    "4.15.0-ec.0",
    "4.15.0-ec.1",
    "4.15.0-ec.2",
    "4.15.0-ec.3",
    "4.15.0-rc.0",
    "4.15.0-rc.1",
    "4.16.0-ec.0"
  ]
}
```

ci-operator/step-registry/cucushift/ota/postupgrade/cucushift-ota-postupgrade-commands.sh
jiajliu: updated the 1st one in [d585db5](https://github.com/openshift/release/pull/49743/commits/d585db51ead597b2083f1ac79b5a4b22346d47cf)

ci-operator/step-registry/cucushift/ota/preupgrade/cucushift-ota-preupgrade-commands.sh
jiajliu: updated the 1st one in [d585db5](https://github.com/openshift/release/pull/49743/commits/d585db51ead597b2083f1ac79b5a4b22346d47cf)

jiajliu: --to-latest will pick the latest one in available update and don't care about arch. But cincinnati should send correct arch back when cvo request this info with arch, so there is pre-upgrade step to check if all update in available update has correct arch

jianlinliu: From my understanding, we can rely on cincinnati returning a correct arch of payload image, but run post-check to ensure the upgrade get successful upgrade, but not have to check each returned image from cincinnati, after in the end, we are expecting cvo and cincinnati should work together correctly, and get the cluster upgraded.

ci-operator/step-registry/cucushift/ota/preupgrade/cucushift-ota-preupgrade-commands.sh
jianlinliu: Once it is unmatched, the cucushift-upgrade-toimage step will fail, because there are several guard to check the target version, while the target version is from ci config.

ci-operator/step-registry/cucushift/ota/preupgrade/cucushift-ota-preupgrade-commands.sh
jiajliu: > rely on cincinnati returning a correct arch of payload image

Yes, there are two checkpoints here. One of the checkpoint is to check if Cincinnati return a correct arch.  And then pick one to upgrade to check if upgrade succeed. --tolatest is picking one of the available update to upgrade, even upgrade to A succeed, it could not guide all other available updates with correct arch.

ci-operator/step-registry/cucushift/ota/preupgrade/cucushift-ota-preupgrade-commands.sh
jiajliu: good catch, let me update to do post check with correct target version when unmatched.

ci-operator/step-registry/cucushift/ota/preupgrade/cucushift-ota-preupgrade-commands.sh
```diff
@@ -239,6 +239,62 @@ function pre-OCP-53921(){
...
+        echo "No image extracted from recommended update!"
+        return 1
+    fi
+    for image in ${images[*]}; do
+        if [[ "${image}" == "null" ]] ; then
+            echo "No image info!"
+            return 1
+        fi
+        metadata=$(oc adm release info ${image} -ojson|jq .metadata.metadata)
+        if [[ "${metadata}" == "null" ]]; then
```
jianlinliu, line 285: If all images have no metadata, the loop will exit as success, shall we do one more check after the loop?

ci-operator/step-registry/cucushift/ota/preupgrade/cucushift-ota-preupgrade-commands.sh
jianlinliu: I am thinking if we should create a dedicated ci job (only for OTA testing) to run such testing to avoid bring some noise when running together with regular ci.

jiajliu:
/pj-rehearse periodic-ci-openshift-verification-tests-release-4.15-ota-multi-stable-4.15-upgrade-from-stable-4.14-azure-ipi-amd-mixarch-f28

jiajliu:
@jianlinliu @shellyyang1989  PTAL again for https://issues.redhat.com/browse/OCPQE-20055. thanks! I will add 4.15-4.16 ota job if lgtu.

jiajliu:
also cc @JianLi-RH @dis016 for awareness about the dedicated ota job in case there will be other cases not suitable to be hosed in regular ci job.

ci-operator/step-registry/cucushift/upgrade/tolatest/cucushift-upgrade-tolatest-commands.sh
```diff
@@ -0,0 +1,681 @@
...
+    fi
+    if [[ "${FORCE_UPDATE}" == "false" ]]; then
+        admin_ack
+        cco_annotation
+    fi
+    if [[ "${UPGRADE_CCO_MANUAL_MODE}" == "oidc" ]]; then
+        update_cloud_credentials_oidc
+    fi
+    if ! check_ota_case_enabled "OCP-53907"; then
+        set_channel
```
jianlinliu, line 674: We already set channel in preupgrade step, we do not need this if block, right?

jiajliu: not yet, only case ocp-53907 will do some checks depending on channel update. so skip set_channel for ocp-53907. for other cases in the future, they just call tolatest step for a `--to-latest` upgrade if they don't need prechecks depending on channel.

jianlinliu: We are setting channel in preupgrade step only when `ENABLE_OTA_TEST: OCP-53907` is setting in config level, so that means if `OCP-53907` is not enabled in other ci jobs, `set_channel` will never happen, sounds we do not need this if block, is my understanding right?

jiajliu: set_channel is a required step if we want do Y-upgrade with --to-latest. if ocp-530907 is not enabled in other ci jobs, we need it to get available update due to default channel is on X.

ci-operator/step-registry/cucushift/upgrade/tolatest/cucushift-upgrade-tolatest-commands.sh
```diff
@@ -0,0 +1,681 @@
...
+        recommends=$(oc get clusterversion version -o json|jq -r '.status.availableUpdates')
+        if [[ "${recommends}" == "null" ]]; then
+            (( retry -= 1 ))
+            sleep 60
+            echo "No recommended update available! Retry..."
+        else
+            echo "Recommencded update: ${recommends}"
+            break
+        fi
+    done
```
jiajliu, line 560: looks like i left out the timeout check here, will add it together with 4.16 job later

ci-operator/step-registry/cucushift/upgrade/tolatest/cucushift-upgrade-tolatest-commands.sh
jianlinliu: Ah, got it, sounds like we always need `set_channel` whatever running `OCP-53907` once tolatest step is referenced, if that, I think we remove the if condition to always run it and remove the step of setting channel in preupgrade. WDTY?

jiajliu: yes, we always need set_channel. But for ocp-53907, we need set_channel in advance, which needed to be ran in preupgrade for the checkpoint on available update. It's an exception here.

ci-operator/step-registry/cucushift/upgrade/tolatest/cucushift-upgrade-tolatest-commands.sh
```diff
@@ -0,0 +1,681 @@
...
+function upgrade_tolatest(){
+    echo "Upgrade cluster to the latest recommended update with --to-latest"
+    recommends=$(oc get clusterversion version -o json|jq -r '.status.availableUpdates')
+    if [[ "${recommends}" == "null" ]]; then
+        echo "No recommended update available!"
+        exit 1
+    fi
+    tmp_log=$(mktemp)
+    run_command "oc adm upgrade --to-latest 2>&1 | tee ${tmp_log}"
+    TARGET_VERSION=$(cat "${tmp_log}"|awk '{print $NF}')
```
jianlinliu, line 571: Is it possible to get the `TARGET_VERSION` after `set_channel` but without running `admin_ack` and `cco_annotation`?

ci-operator/step-registry/cucushift/upgrade/tolatest/cucushift-upgrade-tolatest-commands.sh
```diff
@@ -0,0 +1,681 @@
...
+    SOURCE_MINOR_VERSION="$(echo "${SOURCE_VERSION}" | cut -f2 -d.)"
+    export SOURCE_VERSION
+    export SOURCE_MINOR_VERSION
+    echo -e "Source release version is: ${SOURCE_VERSION}\n"
+
+    TARGET_MINOR_VERSION="$(echo "${DUMMY_TARGET_VERSION}" | cut -f2 -d.)"
+    export TARGET_MINOR_VERSION
+
+    export FORCE_UPDATE="false"
+    if ! check_signed; then
```
jianlinliu, line 660: Sounds like we do not need to run `check_signed` any more, because we are running a recommended upgrade from upstream channel.

ci-operator/step-registry/cucushift/upgrade/tolatest/cucushift-upgrade-tolatest-commands.sh
jiajliu: we need run --to-latest, then it will return the target version, so sounds not.

ci-operator/step-registry/cucushift/upgrade/tolatest/cucushift-upgrade-tolatest-commands.sh
jiajliu: we also hit some recommended update in candidate channel without signature yet during test, then --force is still needed.

ci-operator/step-registry/cucushift/upgrade/tolatest/cucushift-upgrade-tolatest-commands.sh
jiajliu: seems not dry-run option to trigger a dummy --to-latest in advance.
```
# ./oc adm upgrade --to-latest --help|grep -E "dry|dry-run|dryrun"
#
```

ci-operator/step-registry/cucushift/upgrade/tolatest/cucushift-upgrade-tolatest-commands.sh
```diff
@@ -0,0 +1,681 @@
...
+        if [[ " ${ENABLE_OTA_TEST} " =~ " ${case_id} " ]]; then
+            echo "${case_id} is enabled via ENABLE_OTA_TEST on this job."
+            return 0
+        fi
+    done
+    return 1
+}
+
+function set_channel(){
+    x_ver=$( echo "${DUMMY_TARGET_VERSION}" | cut -f1 -d. )
```
jianlinliu, line 542: On https://github.com/openshift/release/pull/49743/files#diff-c587fa8e293fdade6e68f947f6b3b4c7f27c3967a4ad98c6c7409f3a3d50ef6dR245, it is using source cluster version to determinate the target channel, while here is using dummy target payload imag version, can we keep them consistent in the two places?

ci-operator/step-registry/cucushift/upgrade/tolatest/cucushift-upgrade-tolatest-commands.sh
jianlinliu: I am thinking if we can move the checkpoint in preupgrade to this step, once we always run `set_channel`, so that easily understand the case design idea.

jiajliu: I also thinking about it. But finally, i prefer to keep case checkpoint in pre/post upgrade steps to get tolatest step more simple and clear. and cases specific checkpoints in pre/post upgrade steps is the initial purpose for it to toimage step.

ci-operator/step-registry/cucushift/upgrade/tolatest/cucushift-upgrade-tolatest-commands.sh
jiajliu: ah, because ${DUMMY_TARGET_VERSION} is ready-made which does not need `+1` on y_ver here, so i just use it directly. i'm ok to keep it consistent to use source cluster version.

jiajliu: oh, just realize we could not use source cluster version +1 because it might be a z-stream upgrade with --tolatest? so  ${DUMMY_TARGET_VERSION} should be the correct one here.

As for source cluster version in prestep, it's for ocp-53907 case only, we enable it in a y-version upgrade job explicitly, so that's ok to use source cluster version, and we don't need import env OPENSHIFT_UPGRADE_RELEASE_IMAGE_OVERRIDE for it.

ci-operator/config/openshift/verification-tests/openshift-verification-tests-release-4.15__ota-multi-stable-4.15-upgrade-from-stable-4.14.yaml
```diff
@@ -0,0 +1,86 @@
...
+    test:
+    - ref: ipi-install-heterogeneous
+    - ref: cucushift-upgrade-healthcheck
+    - ref: cucushift-ota-preupgrade
+    - ref: cucushift-upgrade-tolatest
+    - ref: cucushift-ota-postupgrade
+    - chain: cucushift-installer-check-cluster-health
+    workflow: cucushift-installer-rehearse-azure-ipi
+zz_generated_metadata:
+  branch: release-4.15
```
jianlinliu, line 83: To keep consistent with installation dedicated ci, I suggest to use `master` branch here, we already indicated release version in `variant`.

ci-operator/step-registry/cucushift/upgrade/tolatest/cucushift-upgrade-tolatest-commands.sh
jianlinliu: > it might be a z-stream upgrade with --tolatest

Ah, I did not consider the z-stream upgrade, agree with you.

ci-operator/step-registry/cucushift/upgrade/tolatest/cucushift-upgrade-tolatest-commands.sh
jianlinliu: good to know.

ci-operator/step-registry/cucushift/upgrade/tolatest/cucushift-upgrade-tolatest-commands.sh
jianlinliu: Okay, in this step, it is using `DUMMY_TARGET_VERSION` as target channel, while in your case, you are using source version + 1 to set target channel, though we also can  `DUMMY_TARGET_VERSION` to do that, but that would introduce more codes in preupgrade, so I am okay with the current code now.

ci-operator/step-registry/cucushift/upgrade/tolatest/cucushift-upgrade-tolatest-commands.sh
```diff
@@ -0,0 +1,681 @@
...
+echo "Upgrade targets are ${TARGET_RELEASES[*]}"
+
+export OC="run_command_oc"
+
+# Target version oc will be extract in the /tmp/client directory, use it first
+mkdir -p /tmp/client
+export OC_DIR="/tmp/client"
+export PATH=${OC_DIR}:$PATH
+
+for target in "${TARGET_RELEASES[@]}"
```
jianlinliu, line 643: For this step - `cucushift-upgrade-to` step, it should be only used in one-hop upgrade, so I think we can remove the loop, once that, we can also can remove the related codes about health check in this script, leave the health check to `cucushift-installer-check-cluster-health`, once that, we can largely shorten the code in this script.

ci-operator/step-registry/cucushift/upgrade/tolatest/cucushift-upgrade-tolatest-commands.sh
```diff
@@ -0,0 +1,681 @@
...
+    extract_oc
+
+    SOURCE_VERSION="$(oc get clusterversion --no-headers | awk '{print $2}')"
+    SOURCE_MINOR_VERSION="$(echo "${SOURCE_VERSION}" | cut -f2 -d.)"
+    export SOURCE_VERSION
+    export SOURCE_MINOR_VERSION
+    echo -e "Source release version is: ${SOURCE_VERSION}\n"
+
+    TARGET_MINOR_VERSION="$(echo "${DUMMY_TARGET_VERSION}" | cut -f2 -d.)"
+    export TARGET_MINOR_VERSION
```
jianlinliu, line 657: Just like what we discussed on slack, we might can determinate the exact `TARGET_VERSION` and `TARGET` before running `oc adm upgrade` command. The code is something like:
```
if ! check_ota_case_enabled "OCP-53907"; then
    set_channel
fi

SOURCE_VERSION="$(oc get clusterversion --no-headers | awk '{print $2}')"
SOURCE_MINOR_VERSION="$(echo "${SOURCE_VERSION}" | cut -f2 -d.)"
export SOURCE_VERSION
export SOURCE_MINOR_VERSION
echo -e "Source release version is: ${SOURCE_VERSION}\nSource minor version is: ${SOURCE_MINOR_VERSION}"

availableUpdates_file=$(mktemp)
echo "Available updates from CV:"
oc get clusterversion version -o json|jq -r '.status.availableUpdates' | tee "${availableUpdates_file}"
if [[ ! -s "${availableUpdates_file}" ]]; then
    echo "No recommended update available! exit..."
    exit 1
fi
availableUpdates_version_file=$(mktemp)
echo "Available update versions from CV:"
cat "${availableUpdates_file}" | jq -r "sort_by(.version) | reverse | .[].version" | tee "${availableUpdates_version_file}"
TARGET_VERSION=$(head -1 "${availableUpdates_version_file}")
if [[ -z "${TARGET_VERSION}" ]]; then
    echo "No recommended update version found! exit..."
    exit 1
fi
# for OCP version management, 4.x.y > 4.x.y-rc.n > 4.x.y-ec.n
# while sort only place the versions in alphabetical order
# so need this if block to get the correct target version
if [[ "$TARGET_VERSION" == *"-"* ]]; then
    temp_target_version="${TARGET_VERSION%-*}"
    if grep -q "^${temp_target_version}$" "${availableUpdates_version_file}"; then
        TARGET_VERSION="${temp_target_version}"
    fi
fi
TARGET_MINOR_VERSION="$(echo "${TARGET_VERSION}" | cut -f2 -d.)"
export TARGET_VERSION
export TARGET_MINOR_VERSION
echo -e "Target release version is: ${TARGET_VERSION}\nTarget minor version is: ${TARGET_MINOR_VERSION}"

target=$(cat "${availableUpdates_file}" | jq -r --arg z "${TARGET_VERSION}" '.[] | select(.version==$z) | .image')
export TARGET="${target}"
echo -e "Target release image is: ${TARGET}"
```

ci-operator/step-registry/cucushift/upgrade/tolatest/cucushift-upgrade-tolatest-commands.sh
jianlinliu: This issue should be able to fix in https://github.com/openshift/release/pull/49743#discussion_r1526174170

ci-operator/step-registry/cucushift/upgrade/tolatest/cucushift-upgrade-tolatest-commands.sh
```diff
@@ -0,0 +1,681 @@
...
+}
+function upgrade_tolatest(){
+    echo "Upgrade cluster to the latest recommended update with --to-latest"
+    recommends=$(oc get clusterversion version -o json|jq -r '.status.availableUpdates')
+    if [[ "${recommends}" == "null" ]]; then
+        echo "No recommended update available!"
+        exit 1
+    fi
+    tmp_log=$(mktemp)
+    run_command "oc adm upgrade --to-latest 2>&1 | tee ${tmp_log}"
```
jianlinliu, line 570: Per our discussion, we are on the same page now to get `--to` and `--tolatest` supported in this step, we can introduce a ENV called `TO_VERSION` for that, something like:
```
local upgrade_to_option=""
if [[ "$TO_VERSION" == "latest" ]]; then
    upgrade_to_option="--tolatest"
else
    if [[ -n "$TO_VERSION" ]]; then
        upgrade_to_option="--to $TO_VERSION"
    else
        upgrade_to_option="--to $TARGET_VERSION"
    fi
fi
run_command "oc adm upgrade ${upgrade_to_option}"
```

ci-operator/step-registry/cucushift/upgrade/tolatest/cucushift-upgrade-tolatest-commands.sh
```diff
@@ -0,0 +1,681 @@
...
+            echo "CCO annotation change is not required in default or Mint mode on 4.${SOURCE_MINOR_VERSION} ${platform} cluster"
+            return 0
+        fi
+    else
+        echo "CCO annotation change is not required in ${cco_mode} mode"
+        return 0
+    fi
+
+    echo "Require CCO annotation change"
+    local wait_time_loop_var=0; to_version="$(echo "${DUMMY_TARGET_VERSION}" | cut -f1 -d-)"
```
jianlinliu, line 162: Once we can determinate the exact target version just like https://github.com/openshift/release/pull/49743/files#r1526174170, we can use `TARGET_VERSION` instead of `DUMMY_TARGET_VERSION`

ci-operator/config/openshift/verification-tests/openshift-verification-tests-release-4.15__ota-multi-stable-4.15-upgrade-from-stable-4.14.yaml
jiajliu: sure

ci-operator/step-registry/cucushift/upgrade/tolatest/cucushift-upgrade-tolatest-commands.sh
jiajliu: ok, let me update and limit it the one-hop upgrade only.

ci-operator/step-registry/cucushift/upgrade/tolatest/cucushift-upgrade-tolatest-commands.sh
jiajliu: ok, let me merge the two options in one step

jiajliu:
/pj-rehearse periodic-ci-openshift-verification-tests-master-ota-multi-stable-4.15-upgrade-from-stable-4.14-azure-ipi-amd-mixarch-f28

ci-operator/step-registry/cucushift/upgrade/toversion/cucushift-upgrade-toversion-commands.sh
```diff
@@ -0,0 +1,487 @@
...
+            exit 1
+        fi
+    # get target_version from available updates
+    else
+        TARGET_VERSION=$(oc get clusterversion version -o json|jq -r '.status.availableUpdates[]?.version'|sort|tail -n1)
+        # there will be three types version in candidate channel:4.15.0-ec1,4.15.0-rc1,4.15.0
+        # when there are all above, sort will return 4.15.0-rc1 as latest(unexpected, should be 4.15.0),
+        # when there is both 4.15.0 and 4.15.0-rc1, sort will return 4.15.0-rc1 as latest(unexpected, should be 4.15.0),
+        # when there is both 4.15.0-rc1 and 4.15.0-ec1, sort will return 4.15.0-rc1 as latest(expected),
+        # there will not be 4.15.0 and 4.15.0-ec1 only, ignore
```
jianlinliu, line 377: If there is 4.15.0, 4.16-ec.1 in `availableUpdates`, the logic will get wrong results.

ci-operator/step-registry/cucushift/upgrade/toversion/cucushift-upgrade-toversion-commands.sh
```diff
@@ -0,0 +1,487 @@
...
+# judge if upgrade with --to or --to-latest cmd
+# set correct target_version for the upgrade
+function set_target_and_upgrade_cmd(){
+    mapfile -t available_versions < <(echo "$(oc get clusterversion version -o json|jq -r '.status.availableUpdates[]?.version')")
+    if [[ -z "${available_versions[*]}" ]]; then
+        echo "No recommended update available!"
+        exit 1
+    fi
+    # for "oc adm upgrade --to x.y.z" and "oc adm upgrade --to $latest_version_in_available_update"
+    upgrade_cmd="oc adm upgrade --to ${TARGET_VERSION}"
```
jianlinliu, line 354: I did not not see where defined the `TARGET_VERSION` before referenced it in this command line.

ci-operator/step-registry/cucushift/upgrade/toversion/cucushift-upgrade-toversion-commands.sh
```diff
@@ -0,0 +1,487 @@
...
+
+# This step slim down for one-hop upgrade only. 
+# If we will call it in chanin upgrade in the future, we need add target loop and health_check back. 
+export OC="run_command_oc"
+# Target version oc will be extract in the /tmp/client directory, use it first
+mkdir -p /tmp/client
+export OC_DIR="/tmp/client"
+export PATH=${OC_DIR}:$PATH
+
+export TARGET="${OPENSHIFT_UPGRADE_RELEASE_IMAGE_OVERRIDE}"
```
jianlinliu, line 452: To be more accurate, I think after we determinate `TARGET_VERSION`, we should overwrite the `TARGET` var with the real limage, too.

ci-operator/step-registry/cucushift/upgrade/tolatest/cucushift-upgrade-tolatest-commands.sh
jianlinliu: After checked the latest codes, I saw change the codes a bit and move it in `set_target_and_upgrade_cmd` function. I want to explain my idea for the above codes.

Most of upgrade steps are developed based on `cucushift-upgrade-toimage` step, so here I almost copied the same code lines and layout from `cucushift-upgrade-toimage`, just for easily comparing code difference between steps in the future, in order to help locate potential script issues and code change sync. In the above code, the biggest change (compared with `cucushift-upgrade-toimage`) is to identify `TARGET_VERSION`, once the real `TARGET_VERSION` is confirmed, the real image also can be confirmed, save it in `TARGET`, then all other codes is almost the same as `cucushift-upgrade-toimage`. Once that, in the future, when comparing the difference, I just need to focus on the difference in the main part in this shell script.

ci-operator/step-registry/cucushift/upgrade/toversion/cucushift-upgrade-toversion-commands.sh
jiajliu: 4.16 should always > 4.15 in sort, right?

ci-operator/step-registry/cucushift/upgrade/toversion/cucushift-upgrade-toversion-commands.sh
jiajliu: yeah, updated

ci-operator/step-registry/cucushift/upgrade/toversion/cucushift-upgrade-toversion-commands.sh
jiajliu: updated

ci-operator/step-registry/cucushift/upgrade/tolatest/cucushift-upgrade-tolatest-commands.sh
jiajliu: Hmm, i think here are two key points to change:
1. add the judgement of env ${UPGRADE_TO_VERSION}
2. determine the TARGET_VERSION
And the two points are cross-lap, so I add them into one func `set_target_and_upgrade_cmd` to make it more clear.

jiajliu:
/pj-rehearse periodic-ci-openshift-verification-tests-master-ota-multi-stable-4.15-upgrade-from-stable-4.14-azure-ipi-amd-mixarch-f28

ci-operator/config/openshift/verification-tests/openshift-verification-tests-master__ota-multi-stable-4.15-upgrade-from-stable-4.14.yaml
```diff
@@ -0,0 +1,84 @@
...
+      version_bounds:
+        lower: 4.15.0-0
+        upper: 4.16.0-0
+resources:
+  '*':
+    requests:
+      cpu: 100m
+      memory: 200Mi
+tests:
+- as: azure-ipi-amd-mixarch-f28
```
jianlinliu, line 62: `f28` means running once per 28 days, while the following cron sounds running on f14, suggest to run `bash ci-operator/config/openshift/openshift-tests-private/tools/generate-cron-entry.sh azure-ipi-amd-mixarch-f28 <ci-yaml-file>` to generate an uniq one.

ci-operator/config/openshift/verification-tests/openshift-verification-tests-master__ota-multi-stable-4.16-upgrade-from-stable-4.15.yaml
```diff
@@ -0,0 +1,84 @@
...
+        lower: 4.16.0-0
+        upper: 4.17.0-0
+resources:
+  '*':
+    requests:
+      cpu: 100m
+      memory: 200Mi
+tests:
+- as: azure-ipi-amd-mixarch-f28
+  cron: 54 6 14,28 * *
```
jianlinliu, line 63: The same comment as https://github.com/openshift/release/pull/49743/files#r1528207747

ci-operator/step-registry/cucushift/upgrade/toversion/cucushift-upgrade-toversion-commands.sh
```diff
@@ -0,0 +1,492 @@
...
+        if [[ "${UPGRAGE_TO_VERSION}" == "latest" ]]; then
+            upgrade_cmd="oc adm upgrade --to-latest"
+        # "": run upgrade with "oc adm upgrade --to $latest_version_in_available_update"
+        else
+            upgrade_cmd="oc adm upgrade --to ${TARGET_VERSION}"
+        fi
+    fi
+    TARGET_MINOR_VERSION="$(echo "${TARGET_VERSION}" | cut -f2 -d.)"
+    target_repo="${DUMMY_TARGET%:*}"
+    TARGET="${target_repo}:${TARGET_VERSION}"
```
jianlinliu, line 397:
```suggestion
    TARGET=$(oc get clusterversion version -o json| jq -r --arg z "${TARGET_VERSION}" '.status.availableUpdates[]? | select(.version==$z) | .image')
```
Based on we already got `TARGET_VERSION` from CV's availableUpdates, I would suggest to get the target image from CV, too. Reduce interact with `DUMMY_TARGET`.

jiajliu:
/pj-rehearse periodic-ci-openshift-verification-tests-master-ota-multi-stable-4.16-upgrade-from-stable-4.15-azure-ipi-amd-mixarch-f28

ci-operator/config/openshift/verification-tests/openshift-verification-tests-master__ota-multi-stable-4.16-upgrade-from-stable-4.15.yaml
```diff
@@ -0,0 +1,85 @@
...
+      BASE_DOMAIN: qe.azure.devcluster.openshift.com
+      COMPUTE_NODE_REPLICAS: "2"
+      ENABLE_OTA_TEST: OCP-53907
+    test:
+    - ref: ipi-install-heterogeneous
+    - ref: cucushift-upgrade-healthcheck
+    - ref: cucushift-upgrade-setchannel
+    - ref: cucushift-ota-preupgrade
+    - ref: cucushift-upgrade-toversion
+    - chain: cucushift-installer-check-cluster-health
```
jianlinliu, line 78: cucushift-upgrade-healthcheck```suggestion
    - ref: cucushift-upgrade-healthcheck
```
Last time, I pointed to a wrong health check step name, maybe `cucushift-upgrade-healthcheck` is more proper here. Sorry for that.

File: ci-operator/step-registry/cucushift/upgrade/setchannel/cucushift-upgrade-setchannel-commands.sh [hunk 22]
```diff
@@ -0,0 +1,91 @@
...
+    source "${SHARED_DIR}/proxy-conf.sh"
+fi
+
+mkdir -p /tmp/client
+export OC_DIR="/tmp/client"
+export PATH=${OC_DIR}:$PATH
+export DUMMY_TARGET="${OPENSHIFT_UPGRADE_RELEASE_IMAGE_OVERRIDE}"
+# we need this DUMMY_TARGET_VERSION from ci config to download oc
+DUMMY_TARGET_VERSION="$(env "NO_PROXY=*" "no_proxy=*" oc adm release info "${DUMMY_TARGET}" --output=json | jq -r '.metadata.version')"
+echo "Target version of ci config is: ${DUMMY_TARGET_VERSION}\n"
```
jianlinliu, line 64:
```suggestion
echo -e "Target version of ci config is: ${DUMMY_TARGET_VERSION}\n"
```

ci-operator/step-registry/cucushift/upgrade/toversion/cucushift-upgrade-toversion-commands.sh
```diff
@@ -0,0 +1,453 @@
...
+export OC="run_command_oc"
+# Target version oc will be extract in the /tmp/client directory, use it first
+mkdir -p /tmp/client
+export OC_DIR="/tmp/client"
+export PATH=${OC_DIR}:$PATH
+
+export DUMMY_TARGET="${OPENSHIFT_UPGRADE_RELEASE_IMAGE_OVERRIDE}"
+# we need this DUMMY_TARGET_VERSION from ci config to download oc and do some pre-check
+DUMMY_TARGET_VERSION="$(env "NO_PROXY=*" "no_proxy=*" oc adm release info "${DUMMY_TARGET}" --output=json | jq -r '.metadata.version')"
+echo "Target version of ci config is: ${DUMMY_TARGET_VERSION}\n"
```
jianlinliu, line 424:
```suggestion
echo -e "Target version of ci config is: ${DUMMY_TARGET_VERSION}\n"
```

ci-operator/step-registry/cucushift/upgrade/toversion/cucushift-upgrade-toversion-commands.sh
```diff
@@ -0,0 +1,453 @@
...
+extract_oc
+
+SOURCE_VERSION="$(oc get clusterversion --no-headers | awk '{print $2}')"
+SOURCE_MINOR_VERSION="$(echo "${SOURCE_VERSION}" | cut -f2 -d.)"
+export SOURCE_VERSION
+export SOURCE_MINOR_VERSION
+echo "Source release version is: ${SOURCE_VERSION}\n"
```
jianlinliu, line 431:
```suggestion
echo -e "Source release version is: ${SOURCE_VERSION}\n"
```

ci-operator/step-registry/cucushift/upgrade/toversion/cucushift-upgrade-toversion-commands.sh
```diff
@@ -0,0 +1,453 @@
...
+    mapfile -t available_versions < <(echo "$(oc get clusterversion version -o json|jq -r '.status.availableUpdates[]?.version')")
+    if [[ -z "${available_versions[*]}" ]]; then
+        echo "No recommended update available!"
+        exit 1
+    fi
+    # x.y.z: run upgrade with "oc adm upgrade --to x.y.z";
+    if [[ -n "${UPGRADE_TO_VERSION}" ]] && [[ "${UPGRADE_TO_VERSION}" != "latest" ]]; then
+        valid_target="false"
+        # check if x.y.z is in available_versions, if not, it's invalid target with --to.
+        for version in ${available_versions[*]}; do
```
jianlinliu, line 317: The checking loop looks like similar to the one on line 340, I think maybe we can extract it as a function named something like `validate_version_in_available_updates`, wdyt?

jiajliu: do we really need to add one more func here for 3 lines while it is indeed in a dedicated func? and there are two difference logics here. it seems no benefit here:)

ci-operator/config/openshift/verification-tests/openshift-verification-tests-master__ota-multi-stable-4.16-upgrade-from-stable-4.15.yaml
jiajliu: there is already `cucushift-upgrade-healthcheck` in line74, dup steps name seems not accepted here.

ci-operator/step-registry/cucushift/upgrade/toversion/cucushift-upgrade-toversion-commands.sh
jianlinliu: whatever, on your call.

ci-operator/config/openshift/verification-tests/openshift-verification-tests-master__ota-multi-stable-4.16-upgrade-from-stable-4.15.yaml
jianlinliu: Ah, I did not noticed the line 74, we also have `cucushift-upgrade-prehealthcheck` to use, I guess we can use it on line 74. For now, I saw you switch the place of `cucushift-upgrade-healthcheck` and `cucushift-installer-check-cluster-health`, that is also a solution for me.

jiajliu:
/pj-rehearse periodic-ci-openshift-verification-tests-master-ota-multi-stable-4.16-upgrade-from-stable-4.15-azure-ipi-amd-mixarch-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-verification-tests-master-ota-multi-stable-4.16-upgrade-from-stable-4.15-azure-ipi-amd-mixarch-f28

ci-operator/step-registry/cucushift/upgrade/setchannel/cucushift-upgrade-setchannel-commands.sh
```diff
@@ -0,0 +1,91 @@
...
+echo "Target version of ci config is: ${DUMMY_TARGET_VERSION}"
+extract_oc
+
+x_ver=$( echo "${DUMMY_TARGET_VERSION}" | cut -f1 -d. )
+y_ver=$( echo "${DUMMY_TARGET_VERSION}" | cut -f2 -d. )
+ver="${x_ver}.${y_ver}"
+target_channel="${UPGRADE_CHANNEL}-${ver}"
+if ! oc adm upgrade channel ${target_channel}; then
+    echo "Fail to change channel to ${target_channel}!"
+    return 1
```
jianlinliu, line 73:
```suggestion
    exit 1
```

ci-operator/step-registry/cucushift/upgrade/setchannel/cucushift-upgrade-setchannel-commands.sh
```diff
@@ -0,0 +1,91 @@
...
+        sleep 60
+        echo "No recommended update available! Retry..."
+    else
+        echo "Recommencded update: ${recommends}"
+        break
+    fi
+done
+if (( retry == 0 )); then
+    echo "Timeout to get recommended update!" 
+    return 1
```
jianlinliu, line 89:
```suggestion
    exit 1
```

jiajliu: good catch

jiajliu:
/pj-rehearse periodic-ci-openshift-verification-tests-master-ota-multi-stable-4.16-upgrade-from-stable-4.15-azure-ipi-amd-mixarch-f28

jiajliu:
cc @liangxia to review, thanks

PR #49748, opened 2024-03-12
Increase CMO e2e-agnostic-cmo timeout

danielmellado:
This commit increases e2e-agnostic-cmo timeout to 2h 30m in order to
avoid test runs when running CMO CI.

Signed-off-by: Daniel Mellado <dmellado@redhat.com>


danielmellado:
After discussing, we'll for now keep up with the default timeout to keep a simpler configuration.

PR #49765, opened 2024-03-12
Copy provisioning_network file in AUX_HOST

sgoveas:


sgoveas:
This is not needed. It is already covered here https://github.com/openshift-qe/baremetal-qe-infra/blob/master/images/prow-image/root/usr/bin/deprovision.sh#L27

PR #49769, opened 2024-03-12
Sync Serverless CI

serverless-qe:


pierDipi:
I see the change here https://github.com/openshift-knative/hack/pull/130

/cc @maschmid 

/ok-to-test

PR #49776, opened 2024-03-12
GitOps | Remove clusters from cluster pool

varshab1210:
Disabling cluster pool until we have CI jobs set up in the CI


mehabhalodiya (approved):
/lgtm

PR #49788, opened 2024-03-12
If vale-review step fails to execute successfullly, ensure it doesn't fail the build

rohennes:
Ensuring that the vale-review step doesn't fail the build if it doesn't execute successfully - e.g. errors out because of a bad github token


aireilly (approved):
lgtm

PR #49846, opened 2024-03-14
Fix automated release upgrade job

liangxia:
Got failure in the last two installation,
```
INFO[2024-03-13T10:03:08Z] Logs for container test in pod aws-ipi-fips-arm-mixarch-f1-ipi-conf-aws: 
INFO[2024-03-13T10:03:08Z] AWS region: us-east-1 (zones: [ us-east-1d,us-east-1f ])
Using control plane instance type: m6g.xlarge
Using compute instance type: m6g.xlarge
Using compute node replicas: 2
Using controlPlane node replicas: 3
W0313 10:03:06.208921      78 loader.go:222] Config not found: /tmp/kubeconfig-2397551570
info: Using registry public hostname registry.build05.ci.openshift.org
Saved credentials for registry.build05.ci.openshift.org into /tmp/pull-secret
error: image "registry.build05.ci.openshift.org/ci-op-7fq7ybcb/release@sha256:4121547fa7eb5caac749cdcd56553c998e9824411476c700e06d8c1231bffb01" not found: manifest unknown: manifest unknown
```
To unblock the automated-release process, moving to amd64 single architecture.

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__automated-release-stable-4.15-upgrade-from-stable-4.15.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__automated-release-stable-4.15-upgrade-from-stable-4.15.yaml
@@ -43,7 +43,7 @@ build_root:
 releases:
   latest:
     prerelease:
-      architecture: multi
+      architecture: amd64
       product: ocp
       relative: 1
       version_bounds:
@@ -51,7 +51,7 @@ releases:
         upper: 4.16.0-0
   target:
     prerelease:
-      architecture: multi
+      architecture: amd64
       product: ocp
       version_bounds:
         lower: 4.15.0-0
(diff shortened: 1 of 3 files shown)


liangxia:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-automated-release-stable-4.15-upgrade-from-stable-4.15-aws-ipi-fips-amd-f1

jianlinliu:
Does that mean we will only take amd into consideration in automated release?

liangxia:
> Does that mean we will only take amd into consideration in automated release?

I think we will use several e2e and upgrade jobs to evaluate a payload, but currently we only use one e2e and one upgrade to get through the process ( and verify the codes ).

jianlinliu:
> currently we only use one e2e and one upgrade to get through the process ( and verify the codes ).

If that, looks good to me. But in the end, I suggest to pick up some job from the existing ci jobs to run the automated release testing to keep consistent.

liangxia:
> > currently we only use one e2e and one upgrade to get through the process ( and verify the codes ).
> 
> If that, looks good to me. But in the end, I suggest to pick up some job from the existing ci jobs to run the automated release testing to keep consistent.

Yes, that's the plan.
Currently we run a specific job for automated release, because existing jobs can not met the high pass ratio requirement of the job and test cases in it. So we a use specific job with subset of test cases from existing ci jobs.

liangxia:
for aware of the job name changes,
/cc @rioliu-rh @jianzhangbjz

PR #49859, opened 2024-03-14
prow-scheduler: Add scheduler in prow-controller-manager

danilo-gemoli:
Add the scheduler in `prow-controller-manager`.
The flag `--enable-controller=plank` is also needed since `plank` is enabled by default only when no controller is passed as an argument.
This PR doesn't enable scheduling because it doesn't set:
```yaml
scheduler:
  enabled: true
```
therefore ProwJobs are still being created in `triggered` state by each component in Prow.

I just want to make sure the scheduler controller runs and doesn't cause any trouble at least.
I will watch the rolling and, if everything goes smoothly, I am going to enable `scheduling` as well.

### DIFF
--- a/clusters/app.ci/prow/03_deployment/prow-controller-manager.yaml
+++ b/clusters/app.ci/prow/03_deployment/prow-controller-manager.yaml
@@ -92,6 +92,8 @@ items:
           - --projected-token-file=/var/sa-token/token
           - --kubeconfig-dir=/etc/build-farm-credentials
           - --kubeconfig-suffix=config
+          - --enable-controller=plank
+          - --enable-controller=scheduler
           ports:
           - name: metrics
             containerPort: 9090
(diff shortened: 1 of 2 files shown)


danilo-gemoli:
```
time="2024-03-14T08:46:06Z" level=fatal msg="could not update Prow configuration" error="failed to load Prow config in strict mode: error unmarshaling /tmp/tmp.L3vC3R6cCc/_config.yaml: error unmarshaling JSON: while decoding JSON: json: unknown field \"scheduler\""
```
Ok we have some inter-repository dependencies. I can't use a new Prow configuration in `o/release` if Prow isn't up to date in `o/ci-tools` and I can't update Prow in `o/ci-tools` if it hasn't been updated in `o/ci-chat-bot` and `o/release-controller` too.

danilo-gemoli:
/hold
it needs https://github.com/openshift/ci-tools/pull/4047

PR #49865, opened 2024-03-14
Add trusted apps for cluster-proxy components.

xuezhaojun:


xuezhaojun:
/assign @Kyl-Bempah 

Looking for approving! thanks

core-services/prow/02_config/stolostron/cluster-proxy-addon/_pluginconfig.yaml
```diff
@@ -0,0 +1,10 @@
+plugins:
```
zhujian7, line 1: Not sure if an `OWNERS` file is required?

xuezhaojun: I thought OWNERS are auto-genrated by prow, I will follow up, if it's not created, I will create it.

PR #49870, opened 2024-03-14
OSSM-4824 Add jobs for sail-operator

jwendell:


ci-operator/config/openshift-service-mesh/sail-operator/openshift-service-mesh-sail-operator-pre-main.yaml
```diff
@@ -0,0 +1,476 @@
...
+      - name: MAISTRA_SC_POD
+      - name: XDG_CACHE_HOME
+      from: src
+      resources:
+        requests:
+          cpu: 100m
+          memory: 100Mi
+      timeout: 2h0m0s
+    workflow: servicemesh-istio-e2e-profile
+- as: integration-kind
```
fjglira, line 306: Is needed to run here again the kind job? If we are going to run this in the istio-ecosystem ci job maybe we can skip this one to save some resources. In the end the behavior should be the same and we are not going to make change in the fork so running on OCP should be the priority I think

jwendell: We should run everything, because we might have manual changes (PR's) that only will happen in the fork. We should make sure all tests pass.

core-services/prow/02_config/openshift-service-mesh/sail-operator/_prowconfig.yaml
```diff
@@ -0,0 +1,33 @@
...
+            pre-main:
+              protect: true
+              required_pull_request_reviews:
+                required_approving_review_count: 1
+              unmanaged: false
+tide:
+  merge_method:
+    openshift-service-mesh/sail-operator: squash
+  queries:
+  - author: maistra-bot
```
fjglira, line 16: This can be renamed to another name that not include maistra? I mean, something like sail-bot for example

fjglira (approved):

core-services/prow/02_config/openshift-service-mesh/sail-operator/_prowconfig.yaml
jwendell: Not renamed. We'd need to create another account for this. Not sure if it's high priority. We can do this in a follow up if necessary.

PR #49888, opened 2024-03-14
Clean up the Habana build for 1.9 and 1.11 as they are not served

harshad16:
Clean up the Habana build for 1.9 and 1.11 as they are not served
we don't need to build them
they take space of quay.io, even though we don't serve them


atheo89 (commented):
Thanks Harshad, seems that needs a rebase.  
/lgtm 
/approve

PR #49904, opened 2024-03-15
Serverless cluster pool

mgencur:
Create a new cluster pool that will be used by jobs under github.com/openshift-knative organization.

https://issues.redhat.com/browse/SRVCOM-3056

Followed the instructions from https://docs.ci.openshift.org/docs/how-tos/cluster-claim/#existing-cluster-pools
The platform credentials were prepared in Vault and this PR references them, specifically:
```
{
  "aws_access_key_id": "...",
  "aws_secret_access_key": "...",
  "secretsync/target-clusters": "hive",
  "secretsync/target-name": "serverless-aws-credentials",
  "secretsync/target-namespace": "serverless-cluster-pool"
}
```

### DIFF
--- a/clusters/hive/pools/serverless/serverless-ocp-4-15-amd64-aws-us-east-1_clusterpool.yaml
+++ b/clusters/hive/pools/serverless/serverless-ocp-4-15-amd64-aws-us-east-1_clusterpool.yaml
@@ -23,7 +23,7 @@ spec:
     name: install-config-aws-us-east-1
   labels:
     tp.openshift.io/owner: serverless-ci
-  maxSize: 10
+  maxSize: 12
   platform:
     aws:
       credentialsSecretRef:


mgencur:
Hmm, do we actually want OCP 4.15? Probably yes...

clusters/hive/pools/serverless/serverless-ocp-4-15-amd64-aws-us-east-1_clusterpool.yaml
cardil, line 1: Do we want only 4.15? Shouldn't we have here a pool for the lower band, now: 4.12?

clusters/hive/pools/serverless/install-config-aws-us-east-1_secret.yaml
```diff
@@ -0,0 +1,40 @@
...
+  install-config.yaml: |
+    apiVersion: v1
+    baseDomain: serverless.devcluster.openshift.com
+    compute:
+    - architecture: amd64
+      hyperthreading: Enabled
+      name: worker
+      platform:
+        aws:
+          type: m5.xlarge
```
cardil, line 16: Does are [spot instances](https://aws.amazon.com/ec2/spot)?

clusters/hive/pools/serverless/serverless-ocp-4-15-amd64-aws-us-east-1_clusterpool.yaml
```diff
@@ -0,0 +1,35 @@
...
+  hibernationConfig:
+    resumeTimeout: 20m0s
+  imageSetRef:
+    name: ocp-release-4.15.3-x86-64-for-4.15.0-0-to-4.16.0-0
+  installAttemptsLimit: 1
+  installConfigSecretTemplateRef:
+    name: install-config-aws-us-east-1
+  labels:
+    tp.openshift.io/owner: serverless-ci
+  maxSize: 10
```
cardil, line 26: Maybe we should have here some multiple of the number of jobs that usually run per PR. You were saying 4 is the usual number (thus the size), so I think `12` or `8` seams like a better value.

cardil (commented):
Looks great, but I got some questions.

clusters/hive/pools/serverless/install-config-aws-us-east-1_secret.yaml
mgencur: AFAIK, these are NOT spot instances. I think using spot instances in this PR now is probably not a good idea (if it's possible at all). We should first test it somewhere else and see how reliable it is. But it's an interesting idea. Worth trying.

clusters/hive/pools/serverless/serverless-ocp-4-15-amd64-aws-us-east-1_clusterpool.yaml
mgencur: We could set e.g. 12. Some repositories have only two parallel checks (e.g. Serving/Eventing), and some have 4 (s-o, EKB).

clusters/hive/pools/serverless/serverless-ocp-4-15-amd64-aws-us-east-1_clusterpool.yaml
mgencur: AFAIK, we only want 4.15. We don't run on 4.12 for every pull request anymore so this cluster pool would not be utilized. And 4.12 can still be run on demand by using `/test <test_name>`

clusters/hive/pools/serverless/serverless-ocp-4-15-amd64-aws-us-east-1_clusterpool.yaml
mgencur: I've set this to 12

clusters/hive/pools/serverless/install-config-aws-us-east-1_secret.yaml
cardil: Using spot instances for testing would bring massive savings. Also, they are actually targeting CI/CD.

UI agrees we should start with Spot instances, but we definitely should migrate to it. That would probably require fixing some tests, so they may tolerate K8s' node failure. That most likely needs to be fixed upstream. In fact, I think we could use [GKE spot instances](https://cloud.google.com/kubernetes-engine/docs/concepts/spot-vms) on upstream, too. I just opened an issue for this: https://github.com/knative/infra/issues/384

When using spot instances, we should have more, but smaller nodes.

mgencur:
/hold

We want to have everything ready on Serverless side (prepare pull requests for using the cluster pool) before merging this.

mgencur:
@deepsm007 Thanks for review!

PR #49922, opened 2024-03-15
Add not workloads tag

kasturinarra:
Add e2e_test_run and not @workloads tag as the testcases that are tagged with @workloads are not supposed to run on this profile as it has build and dc capability disabled.

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__multi-nightly.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__multi-nightly.yaml
@@ -768,6 +768,7 @@ tests:
     env:
       BASE_DOMAIN: qe.azure.devcluster.openshift.com
       COMPUTE_NODE_TYPE: Standard_D4ps_v5
+      E2E_RUN_TAGS: not @workloads
       OCP_ARCH: arm64
       PERSISTENT_MONITORING: "false"
     test:
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__amd64-nightly.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__amd64-nightly.yaml
@@ -649,7 +649,6 @@ tests:
       TEST_IMPORTANCE: Critical
       TEST_TIMEOUT: "35"
     test:
-    - ref: wait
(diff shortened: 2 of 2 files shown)


kasturinarra:

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__multi-nightly.yaml
```diff
@@ -768,6 +768,7 @@ tests:
     env:
       BASE_DOMAIN: qe.azure.devcluster.openshift.com
       COMPUTE_NODE_TYPE: Standard_D4ps_v5
+      E2E_RUN_TAGS: not @workloads
```
liangxia, line 771: Do we need this for 4.16, too ?

BTW, could you explain more details ?  I'd like to know which capability cause this test filter requirement, so that I can check if it can be auto detected.

kasturinarra: @liangxia i do not see build and dc capability enabled due to which tests dealing with build, buildconfig & deploymentconfig are failing, so added it.

liangxia: For this profile, the build and dc capability are not always disabled, as the capability set to disabled are choose randomly. Always skip workloads tests are not expected as we have this capability on this profile sometimes.
/cc @jinyunma

liangxia: BTW, I've implemented a dynamic way to include/exclude workloads tests based on the capability on the cluster in https://github.com/openshift/release/pull/49946

jinyunma: yes, this profile is not always disable one specific capability, but choose one randomly to disable. 
Thanks @liangxia, dynamic way should fix the issue.

kasturinarra: @liangxia @jinyunma thank you !!

kasturinarra:
closing the PR as @liang has implemented https://github.com/openshift/release/pull/49946

PR #49965, opened 2024-03-18
[RHOAIENG-4565]Configure /approve on ocp-ci to work independently of …

spolti:
…/lgtm

- kserve


Jooho:
@spolti could you please update what this pr is for? moreover, if this is needed for kserve, then please update other repositories as well

PR #50004, opened 2024-03-19
MGMT-16883: Update external jobs to use ocm-2.11 image-sets

adriengentil:
Update assisted to use ocm-2.11 for jobs on master, 4.16 and 4.17
branches.

### DIFF
--- a/ci-operator/config/openshift/installer/openshift-installer-release-4.17.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-release-4.17.yaml
@@ -1,26 +1,26 @@
 base_images:
   assisted-image-service:
-    name: ocm-2.10
+    name: ocm-2.11
     namespace: edge-infrastructure
     tag: assisted-image-service
   assisted-installer:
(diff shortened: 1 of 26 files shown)


ci-operator/config/openshift/installer/openshift-installer-release-4.17.yaml
```diff
@@ -1,26 +1,26 @@
...
     namespace: edge-infrastructure
     tag: assisted-installer-controller
   assisted-service:
-    name: ocm-2.10
+    name: ocm-2.11
     namespace: edge-infrastructure
     tag: assisted-service
   assisted-test-infra:
-    name: ocm-2.10
+    name: ""
```
gamli75, line 23: Is there a reason why you removed it?

adriengentil: nope, good catch, thanks!

gamli75 (approved):

PR #50025, opened 2024-03-20
Adding leftover e2e CI tests

vishnuchalla:
### Description
Adding leftover CI tests for cloud-bulldozer/e2e-benchmarking.
JIRA: https://issues.redhat.com/browse/PERFSCALE-2883

### Testing
Needs to be tested and verified through e2e PRs.


ci-operator/step-registry/cloud-bulldozer/OWNERS
```diff
@@ -1,4 +1,4 @@
-pprovers:
+approvers:
```
paigerube14, line 1: here, instead of needing to edit each of the files you can use a symbolic link to have this main one and then reference it in each of the subfolders. This way we wont have to constantly edit the owners in multiple places or if there are issues like this

vishnuchalla: Yes, makes sense when we have same group of owners for the entire repo. Updated as suggested.

krishvoor:
/pj-rehearse pull-ci-cloud-bulldozer-e2e-benchmarking-master-e2e-conc-builds-single-node

krishvoor (approved):
/lgtm

PR #50031, opened 2024-03-20
drop dup health check

jiajliu:
There is already health check in step `cucushift-upgrade-toimage`. A quick fix to drop dup healthy check.

### DIFF
--- a/ci-operator/step-registry/openshift/upgrade/qe/sanity/disconnected/openshift-upgrade-qe-sanity-disconnected-chain.yaml
+++ b/ci-operator/step-registry/openshift/upgrade/qe/sanity/disconnected/openshift-upgrade-qe-sanity-disconnected-chain.yaml
@@ -6,7 +6,6 @@ chain:
   - ref: cucushift-ota-preupgrade
   - ref: cucushift-upgrade-toimage
   - ref: cucushift-ota-postupgrade  
-  - chain: cucushift-installer-check-cluster-health
   documentation: |-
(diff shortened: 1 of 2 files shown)


jiajliu:
cc @jianlinliu for a review, seems no need a test.

jianlinliu:
Actually the original purpose of `chain: cucushift-installer-check-cluster-health` is run a sanity testing instead of a suite of e2e testing (because we have no dedicated smoke suite in e2e framework), so that means, the purpose of `chain: cucushift-installer-check-cluster-health` actually is not used to check if upgrade success health check.

jiajliu:
> So that means, the purpose of chain: cucushift-installer-check-cluster-health actually is not used to check if upgrade success health check.

Hmm, so we could drop it, right? There is already health check in upgrade step, which called a sanity test for upgrade it self. And we have pre-post upgrade test for other scenarios. If we have other test tobe included in upgrade, we could add them later case by case. 

Now the two steps is duplicated to extend the test time, sounds not reasonable and unmatched to our purpose for enhancement.

jiajliu:
> just for a placeholder to easily opt-in real e2e smoke testing once we have it.

placeholder means doing nothing, right? but now it indeed re-do cluster health check for upgrade job. so i think we should drop it. and if one day it extend the usage scope, we should design it with upgrade into the consideration or not.

jianlinliu:
>  so we could drop it, right?
I am okay with it. cc @shellyyang1989 for more comment.

PR #50050, opened 2024-03-20
Update make e2e target

fjglira:


jwendell (approved):

jwendell (changes requested):
actually I think you need to change the job names too?

fjglira:
I just added: `/pj-rehearse ack` because this need to be merged first to be able to run the job properly again after this change: https://github.com/maistra/istio-operator/pull/1709

fjglira:
> actually I think you need to change the job names too?

Job names updated

jwendell (approved):

PR #50060, opened 2024-03-20
MetalLB CI: decommission the sdn lane and add a frr-k8s lane

fedepaol:
SDN is going to be deprecated, we remove the sdn lane and we add a lane with frr-k8s as a backend.


zeeke (commented):
Relates to:
- https://github.com/openshift/metallb/pull/167

/lgtm

PR #50102, opened 2024-03-21
HOSTEDCP-1462: Add Hypershift Azure presubmits

Patryk-Stefanski:


ci-operator/step-registry/hypershift/azure/setup-nested-management-cluster/hypershift-azure-setup-nested-management-cluster-chain.yaml
```diff
@@ -0,0 +1,88 @@
...
+          documentation: "The Namespace where to create the HostedCluster and NodePools"
+        - name: HYPERSHIFT_BASE_DOMAIN
+          default: ""
+          documentation: "The cluster's FQDN will be a subdomain of the base domain."
+      commands: |-
+        DEFAULT_BASE_DOMAIN="hypershift.azure.devcluster.openshift.com"
+        
+        DOMAIN=${HYPERSHIFT_BASE_DOMAIN:-$DEFAULT_BASE_DOMAIN}
+        
+        DEFAULT_RELEASE_STREAM=4.16.0-0.nightly
```
muraee, line 24: why nightly?

ci-operator/step-registry/hypershift/install/hypershift-install-commands.sh
```diff
@@ -21,17 +21,27 @@ if [ "${ENABLE_HYPERSHIFT_CERT_ROTATION_SCALE}" = "true" ]; then
...
+  --aws-private-creds=/etc/hypershift-pool-aws-credentials/credentials \
+  --aws-private-region="${HYPERSHIFT_AWS_REGION}" \
+  --external-dns-provider=aws \
+  --external-dns-credentials=/etc/hypershift-pool-aws-credentials/credentials \
+  --external-dns-domain-filter=service.ci.hypershift.devcluster.openshift.com \
+  --wait-until-available \
+  ${EXTRA_ARGS}
+fi
+
+if [ "${CLOUD_PROVIDER}" = "AZURE" ]; then
```
jparrill, line 41: don't you need "==" here?

Patryk-Stefanski: can use both "=" and "==" for comparison in this case

ci-operator/step-registry/hypershift/azure/setup-nested-management-cluster/hypershift-azure-setup-nested-management-cluster-chain.yaml
muraee: this is for the management cluster, right? why would we want to catch errors related to the mgmt cluster in e2e?

Patryk-Stefanski: My thought while setting this up was catch out issues with creating hc but thinking of it now the e2e covers that. I will change it to use 4-dev-preview

Patryk-Stefanski: Although we do use 4.16 nightlies in our AWS e2e aswell so don't see any harm in leaving it as is.

bryan-cox: I think we should do whatever we do for AWS e2e to be consistent

muraee: I think there was an  issue in AWS that made us switch to nightly temporary. I think we should be using `4-stable`.
@sjenning wdyt?

sjenning: In general, I would rather not copy/paste this rather large chain.  How does it differ from the AWS version?  Could we do the `CLOUD_PROVIDER` if/else thing you do on the hypershift-install chain?

ci-operator/jobs/openshift/hypershift/openshift-hypershift-main-presubmits.yaml
```diff
@@ -316,6 +316,85 @@ presubmits:
...
+    - ^main-
+    cluster: build05
+    context: ci/prow/e2e-azure
+    decorate: true
+    labels:
+      ci-operator.openshift.io/cloud: hypershift
+      ci-operator.openshift.io/cloud-cluster-profile: hypershift
+      ci.openshift.io/generator: prowgen
+      pj-rehearse.openshift.io/can-be-rehearsed: "true"
+    name: pull-ci-openshift-hypershift-main-e2e-azure
```
csrwng, line 332: I think we want `optional: true` here as in line 412

csrwng: And for all the new jobs in the other branches

PR #50104, opened 2024-03-21
red-hat-data-services: get rid of useless cron jobs

dtrifiro:
remove useless cron jobs


heyselbi:
@dtrifiro @Xaenalt I think [releases](https://github.com/openshift/release/blob/master/ci-operator/config/red-hat-data-services/caikit-tgis-serving/red-hat-data-services-caikit-tgis-serving-rhoai-2.6.yaml#L19) needs to be added and/or golang version updated. But I would say we just completely remove 1.31 CI, instead of debugging it, no?

Xaenalt:
Yeah, agreed, 1.31 is dead

PR #50106, opened 2024-03-21
Update openshift-cli-installer tag

liswang89:
CI testing necessitates a stable tool version to prevent updates from disrupting the pipeline.

### DIFF
--- a/core-services/image-mirroring/supplemental-ci-images/mapping_supplemental_ci_images_ci
+++ b/core-services/image-mirroring/supplemental-ci-images/mapping_supplemental_ci_images_ci
@@ -121,7 +121,7 @@ quay.io/rh_integration/xpaas-qe:7.12.x-openjdk1.8_local registry.ci.openshift.or
 quay.io/coreos/tectonic-console-builder:v25 registry.ci.openshift.org/ci/tectonic-console-builder:v25
 quay.io/coreos/tectonic-console-builder:v26 registry.ci.openshift.org/ci/tectonic-console-builder:v26
 quay.io/kubevirt/prom-metrics-linter:v0.0.2 registry.ci.openshift.org/ci/prom-metrics-linter:v0.0.2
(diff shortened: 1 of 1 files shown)


core-services/image-mirroring/supplemental-ci-images/mapping_supplemental_ci_images_ci
```diff
@@ -121,7 +121,7 @@ quay.io/rh_integration/xpaas-qe:7.12.x-openjdk1.8_local registry.ci.openshift.or
 quay.io/coreos/tectonic-console-builder:v25 registry.ci.openshift.org/ci/tectonic-console-builder:v25
 quay.io/coreos/tectonic-console-builder:v26 registry.ci.openshift.org/ci/tectonic-console-builder:v26
 quay.io/kubevirt/prom-metrics-linter:v0.0.2 registry.ci.openshift.org/ci/prom-metrics-linter:v0.0.2
-quay.io/redhat_msi/openshift-cli-installer:latest registry.ci.openshift.org/ci/openshift-cli-installer:latest
+quay.io/redhat_msi/openshift-cli-installer:stable registry.ci.openshift.org/ci/openshift-cli-installer:latest
```
rnetser, line 124: if you need a specific version; please add a new entry. your change affects anyone using the tool and is using latest

rnetser (changes requested):

core-services/image-mirroring/supplemental-ci-images/mapping_supplemental_ci_images_ci
liswang89: What we require is a stable version that remains consistent regardless of updates. By employing this stable tag, we can avoid the necessity of updating the CI mappings whenever there is a version change. Therefore, whenever there is an update, please allow us time to verify its stability, and then I can retag the stable version accordingly.

PR #50122, opened 2024-03-22
Add optional capabilities for 4.16 e2e/upgrade

liangxia:
Add two more optional capabilities for 4.16. Also pre-add 4.17.

### DIFF
--- a/ci-operator/step-registry/cucushift/e2e/cucushift-e2e-commands.sh
+++ b/ci-operator/step-registry/cucushift/e2e/cucushift-e2e-commands.sh
@@ -149,13 +149,14 @@ function filter_test_by_capability() {
     local enabledcaps xversion yversion
     enabledcaps="$(oc get clusterversion version -o yaml | yq '.status.capabilities.enabledCapabilities[]')"
     IFS='.' read xversion yversion _ < <(oc version -o yaml | yq '.openshiftVersion')
-    local v411 v412 v413 v414 v415 v416
+    local v411 v412 v413 v414 v415 v416 v417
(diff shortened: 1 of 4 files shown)


ci-operator/step-registry/cucushift/e2e/cucushift-e2e-commands.sh
```diff
@@ -149,13 +149,14 @@ function filter_test_by_capability() {
...
     IFS='.' read xversion yversion _ < <(oc version -o yaml | yq '.openshiftVersion')
-    local v411 v412 v413 v414 v415 v416
+    local v411 v412 v413 v414 v415 v416 v417
     v411="baremetal marketplace openshift-samples"
     v412="${v411} Console Insights Storage CSISnapshot"
     v413="${v412} NodeTuning"
     v414="${v413} MachineAPI Build DeploymentConfig ImageRegistry"
     v415="${v414} OperatorLifecycleManager CloudCredential"
-    v416="${v415}"
+    v416="${v415} CloudControllerManager Ingress"
```
jianlinliu, line 158: As far as I  know, `Ingress` is a bit special, it is always enabled in OCP, while it becomes optional in hypershift cluster.

liangxia: This is a list, we use this list to compare with the caps on the cluster to determine which feature to test.

jianlinliu: nice.

PR #50128, opened 2024-03-22
move out check recommended update

jiajliu:
Due to setchannel might be used in other scenarios, which is not only for `cucushift-upgrade-toversion`, so move it out .


jiajliu:
cc @jianlinliu for a review, thanks!

PR #50137, opened 2024-03-22
[CVP-3985] Update the image from ec to rc tag for ocp 4.16 clusters

samvarankashyap:
Do not merge until 4.16 is released.


asergienk:
cvp-fips v4.16 clusterpool has to be updated as well:
https://github.com/openshift/release/blob/7eb4019c692399173b5260480e344ce1962d5f82/clusters/hosted-mgmt/hive/pools/cvp/cvp-fips-ocp-4-16-amd64-aws-us-west-1_clusterpool.yaml

PR #50140, opened 2024-03-22
OSSM-6154: skip tests for insignificant changes

FilipB:


ci-operator/jobs/maistra/istio-operator/maistra-istio-operator-maistra-3.0-presubmits.yaml
```diff
@@ -1,7 +1,7 @@
 presubmits:
   maistra/istio-operator:
   - agent: kubernetes
-    always_run: true
+    always_run: false
```
jwendell, line 4: Not sure about this. Do we need to change it? 
Wouldn't this mean we'd need to run all jobs manually?

I'd revert these changes, and keep only the `skip_if_only_changed` ones and experiment with it.

FilipB: When using `skip_if_only_changed` the `always_run` must be false, otherwise it's failing with: "job pull-ci-maistra-istio-operator-maistra-3.0-unit is set to always run but also declares skip_if_only_changed targets, which are mutually exclusive"

jwendell: I see. I hope prow is smart enough to run the jobs automatically when they don't fit in the `skip` rules :)

jwendell (approved):

PR #50155, opened 2024-03-23
OLS-81: run e2e standalone tests against RAG from the lightspeed-rag-content image

syedriko:
Run OLS e2e standalone tests against RAG from the lightspeed-rag-content image.
Depends on https://github.com/openshift/lightspeed-service/pull/632

### DIFF
--- a/ci-operator/config/openshift/lightspeed-service/openshift-lightspeed-service-main.yaml
+++ b/ci-operator/config/openshift/lightspeed-service/openshift-lightspeed-service-main.yaml
@@ -9,6 +9,9 @@ images:
 - dockerfile_path: Containerfile.rag
   from: lightspeed-service-api
   to: lightspeed-service-api-rag
+- dockerfile_path: Containerfile.test
+  from: lightspeed-service-api-rag
+  to: lightspeed-service-api-test
 - dockerfile_path: Containerfile.redis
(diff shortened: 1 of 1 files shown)


ci-operator/config/openshift/lightspeed-service/openshift-lightspeed-service-main.yaml
```diff
@@ -9,6 +9,9 @@ images:
 - dockerfile_path: Containerfile.rag
   from: lightspeed-service-api
   to: lightspeed-service-api-rag
+- dockerfile_path: Containerfile.rag.test
+  from: lightspeed-service-api-rag
+  to: lightspeed-service-api-rag-test
```
bparees, line 14: why do we need to introduce this image instead of continuing to use the source image?

at some point we might need it if it we want to run these tests against PRs in another repo, but you're not doing that here

syedriko: Because the e2e tests need both RAG and the tests. I looked for a way to mount the rag-content image to the src image, but didn't find a way to use it.

bparees:
https://github.com/openshift/release/pull/50207 should make this PR unnecessary

PR #50163, opened 2024-03-25
OCPEDGE-1018: feat: add upgrade-rt jobs for 4.15 and 4.16

eggfoobar:
Add missing `update-rt` jobs for 4.15 and 4.16

/assign @jeff-roche


eggfoobar:
/unhold

Payloads contain new origin change

PR #50181, opened 2024-03-25
Removing the e2e-openstack-sdn job from 4.15 and 4.16 OpenShift releases

MiguelCarpio:
Removing the e2e-openstack-sdn job because, although OpenShift supports OpenShiftSDN in its 4.15 version, it is not supported as an option in the 4.15 and 4.16 installers, the installer will not accept it in its install-config.yaml.


MaysaMacedo:
@MiguelCarpio the command `make jobs` needs to be run locally on your PC and then pushed for the `shiftstack-shiftstack-ci-main-periodics.yaml` to be updated with the removed jobs.

PR #50196, opened 2024-03-25
TRT-1511: add p-interop-rhsso-interop-aws for 4.14 nightly informing

DennisPeriquet:


DennisPeriquet:
we an revisit this when ready

PR #50208, opened 2024-03-25
Use ci-operator config from backplane-tools repo

rendhalver:


rendhalver:
/hold

It seems the .ci-operator.yaml file in backplane-tools isn't correct.
Going to open a branch we can test that in to get it working first.
(Should have done that to start with but lesson re-learnt)

PR #50259, opened 2024-03-26
release-controllers: add oc-prepare init container

AlexNPavel:
This adds an initContainer for the release-controller and release-controller-api pods along with the new volumes and volumeMounts needed by the local use of the `oc` command.


AlexNPavel:
/hold

Requires the git-credentials secret to be in the `ci` namespace.

AlexNPavel:
The version of `git` in the RHEL8 base of `4.16:tools` is too old for the recent changes to `oc adm release info --changelog/--bugs`. We need to wait until a RHEL9 based image is available.

bradmwilliams (approved):

bradmwilliams (approved):

PR #50324, opened 2024-03-28
OCPQE-20550: Summarize test result only when ARTIFACT_DIR exist

liangxia:
When there are no test cases executed, the ARTIFACT_DIR is not created.
Make sure ARTIFACT_DIR exist before searching in it, so that we can move on to the next tests.

### DIFF
--- a/ci-operator/step-registry/openshift-extended/test/openshift-extended-test-commands.sh
+++ b/ci-operator/step-registry/openshift-extended/test/openshift-extended-test-commands.sh
@@ -320,6 +320,7 @@ function run {
     # summarize test results
     echo "Summarizing test results..."
     failures=0 errors=0 skipped=0 tests=0
+    [[ -e "${ARTIFACT_DIR}" ]] || exit 0
     grep -r -E -h -o 'testsuite.*tests="[0-9]+"' "${ARTIFACT_DIR}" | tr -d '[A-Za-z=\"_]' > /tmp/zzz-tmp.log
     while read -a row ; do
         # if the last ARG of command `let` evaluates to 0, `let` returns 1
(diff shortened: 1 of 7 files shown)


ci-operator/step-registry/openshift-extended/test/openshift-extended-test-commands.sh
```diff
@@ -320,6 +320,7 @@ function run {
     # summarize test results
     echo "Summarizing test results..."
     failures=0 errors=0 skipped=0 tests=0
+    [[ -e "${ARTIFACT_DIR}" ]] || mkdir --parents "${ARTIFACT_DIR}"
```
kuiwang02, line 323: @liangxia I take this step as example to raise my concern (same concern for other step.):
when step run here, if there is no artifacts directory, it means no case is executed.
if we create this directory here, the follow code does not create junit xml. and then trap function warn_0_case_executed also does not create junit xml because this directory is already existing (created with this line.), so finally still no junit xml.

so,  `[[ -e "${ARTIFACT_DIR}" ]] || mkdir --parents "${ARTIFACT_DIR}"` to `[[ -e "${ARTIFACT_DIR}" ]] || exit 0` to skip to grep this directory and finally run into trap function to generate junit xml.

liangxia: warn_0_case_executed checks whether there are files in `${ARTIFACT_DIR}`, so we do not care where is the folder created.

liangxia: Rethink the issue, I agree that `exit 0` is much better.

PR #50346, opened 2024-03-28
[nvidia-ci] Adding a `lint` pre-commit job.

ybettan:
---

/cc @wabouhamad @empovit


ybettan:
/pj-rehearse ack
The job is failing because the pod is run as a user but with the `/` as the home directory, therefore, the linter won't have the permission to create `/.cache`.

This will be fixed in https://github.com/rh-ecosystem-edge/nvidia-ci/pull/12

PR #50148, opened 2024-03-22
install google-chrome in UI images

memodi:
install google-chrome for cypress tests

PR #50370, opened 2024-03-29
[CIGolang]rebuild baseui

kuiwang02:
/cc @Xia-Zhao-rh 

cc @memodi @yapei

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__images.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.12__images.yaml
@@ -35,7 +35,9 @@ images:
 - dockerfile_literal: |
     FROM src AS builder
     FROM registry.ci.openshift.org/ci/tests-private-baseui:4.12
-    RUN mkdir -p /go/src/github.com/openshift
+    RUN wget https://dl.google.com/linux/direct/google-chrome-stable_current_x86_64.rpm && \
+        yum install -y ./google-chrome-stable_current_*.rpm && \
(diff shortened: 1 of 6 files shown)


memodi:
Thanks @kuiwang02, could you also change other release files changed in this PR: https://github.com/openshift/release/pull/50148

kuiwang02:
> Thanks @kuiwang02, could you also change other release files changed in this PR: #50148

OK. thanks to reminder it.

kuiwang02:
@memodi 
suppose https://dl.google.com/linux/direct/google-chrome-stable_current_x86_64.rpm is updated repeatedly to point to the current stable version, so maybe it will be updated few months(1~2, I am not sure).

if we put it into baseui, it means we will check if it is updated and rebuild baseui image with PR. 
if we put it into ui, it could be rebuilt every time when we submit PR in tests-private repo. although it has few overlap, I think it is better than the above. so, I prefer it.

is it ok for you? thanks

memodi:
Yeah, no worries. I had thought about including in test-private-ui but I thought it would be too much to download chrome every time this image is built, so I included in baseui but I didn't know the limitation of baseui at the time that it would not built/promoted routinely.

kuiwang02:
> Yeah, no worries. I had thought about including in test-private-ui but I thought it would be too much to download chrome every time this image is built, so I included in baseui but I didn't know the limitation of baseui at the time that it would not built/promoted routinely.

Thanks. lets go in this way currently.

kuiwang02:
@memodi the tests pass. could you please add lgtm? thanks

memodi (approved):
/lgtm

PR #50416, opened 2024-04-01
update(go): update go version for ODH and RHOAI operator

zdtsw:
- we are using 1.19 for v2, update from 1.18
- keep 1.18 for "master" which is for v1
- update ocp4.10 to 4.14 for must-gather


zdtsw:
@VaishnaviHire FYI

VaishnaviHire (approved):
/lgtm

PR #50443, opened 2024-04-01
fix: add OADP non admin image mirroring job

mateusoliveira43:
## Why the changes were made

OADP non admin images are needed to start testing OADP with non admin controller. This PR adds CD job that will update/upload image to `quay.io`

## How the changes were made

Added new image to mirroring job

## How to test the changes made

Check quay if image is being uploaded.


kaovilai (approved):

shubham-pampattiwar (commented):
/lgtm

PR #50446, opened 2024-04-02
OCPQE-19762 Skip upgrade when target is not recommended in stable-stable path

JianLi-RH:
[OCPQE-19762](https://issues.redhat.com/browse/OCPQE-19762)

/cc @shellyyang1989 @jiajliu @jianlinliu @liangxia @dis016

### DIFF
--- a/ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
+++ b/ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
@@ -618,6 +618,25 @@ function admin_ack() {
 
 # Upgrade the cluster to target release
 function upgrade() {
+    set_channel $TARGET_VERSION
+    local retry=3 conditional_updates
+    while (( retry > 0 )); do
+        conditional_updates=$(oc get clusterversion version -o json | jq -r '.status.conditionalUpdates[]?.release.version' | xargs)
+        if [[ -z "${conditional_updates}" ]]; then
+            retry=$((retry - 1))
+            sleep 60
+            echo "No conditionalUpdates update available! Retry..."
+        else
+            #shellcheck disable=SC2076
+            if [[ " $conditional_updates " =~ " $TARGET_VERSION " ]]; then
+                echo "Error: $TARGET_VERSION is not recommended, for details please refer:"
+                oc get clusterversion version -o json | jq -r '.status.conditionalUpdates'
+                exit 1
+            fi
+            break
+        fi
+    done
+
     run_command "oc adm upgrade --to-image=${TARGET} --allow-explicit-upgrade --force=${FORCE_UPDATE}"
     echo "Upgrading cluster to ${TARGET} gets started..."
 }
@@ -867,6 +886,18 @@ function run_upgrade_e2e() {
     echo "Ending the upgrade e2e tests on $(date "+%F %T")"
     e2e_end_time=$(date +%s)
     echo "e2e test take $(( ($e2e_end_time - $e2e_start_time)/60 )) minutes"
+}
+
+function set_channel(){
+    local x_ver y_ver version="$1"
+    x_ver=$( echo "${version}" | cut -f1 -d. )
+    y_ver=$( echo "${version}" | cut -f2 -d. )
+    ver="${x_ver}.${y_ver}"
+    target_channel="${UPGRADE_CHANNEL}-${ver}"
+    if ! oc adm upgrade channel ${target_channel}; then
+        echo "Fail to change channel to ${target_channel}!"
+        exit 1
+    fi
 }
 
 if [[ -f "${SHARED_DIR}/kubeconfig" ]] ; then
--- a/ci-operator/step-registry/cucushift/upgrade/check-not-recommended/cucushift-upgrade-check-not-recommended-commands.sh
+++ b/ci-operator/step-registry/cucushift/upgrade/check-not-recommended/cucushift-upgrade-check-not-recommended-commands.sh
@@ -1,0 +1,62 @@
+#!/bin/bash
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+
+# Extract oc binary which is supposed to be identical with target release
+function extract_oc(){
+    echo -e "Extracting oc\n"
+    local retry=5 tmp_oc="/tmp/client-2"
+    mkdir -p ${tmp_oc}
+    while ! (env "NO_PROXY=*" "no_proxy=*" oc adm release extract -a "${CLUSTER_PROFILE_DIR}/pull-secret" --command=oc --to=${tmp_oc} ${TARGET});
+    do
+        echo >&2 "Failed to extract oc binary, retry..."
+        (( retry -= 1 ))
+        if (( retry < 0 )); then return 1; fi
+        sleep 60
+    done
+    mv ${tmp_oc}/oc ${OC_DIR} -f
+    which oc
+    oc version --client
+    return 0
+}
+
+if [[ -f "${SHARED_DIR}/kubeconfig" ]] ; then
+    export KUBECONFIG=${SHARED_DIR}/kubeconfig
+fi
+
+# Setup proxy if it's present in the shared dir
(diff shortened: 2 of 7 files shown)


ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
```diff
@@ -869,6 +880,30 @@ function run_upgrade_e2e() {
...
+    target_channel="${UPGRADE_CHANNEL}-${ver}"
+    if ! oc adm upgrade channel ${target_channel}; then
+        echo "Fail to change channel to ${target_channel}!"
+        exit 1
+    fi
+    local retry=3
+    while (( retry > 0 )); do
+        versions=$(oc get clusterversion version -o json|jq -r '.status.availableUpdates[]?.version'| xargs)
+        if [[ "${versions}" == "null" ]]; then
+	        retry=$((retry - 1))
```
jianlinliu, line 897:
```suggestion
	    retry=$((retry - 1))
```

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
```diff
@@ -869,6 +880,30 @@ function run_upgrade_e2e() {
...
+    x_ver=$( echo "${version}" | cut -f1 -d. )
+    y_ver=$( echo "${version}" | cut -f2 -d. )
+    ver="${x_ver}.${y_ver}"
+    target_channel="${UPGRADE_CHANNEL}-${ver}"
+    if ! oc adm upgrade channel ${target_channel}; then
+        echo "Fail to change channel to ${target_channel}!"
+        exit 1
+    fi
+    local retry=3
+    while (( retry > 0 )); do
```
jianlinliu, line 894: I guess we also need to add the while loop to `cucushift-upgrade-setchannel` ?

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
```diff
@@ -869,6 +880,30 @@ function run_upgrade_e2e() {
...
+        versions=$(oc get clusterversion version -o json|jq -r '.status.availableUpdates[]?.version'| xargs)
```
jiajliu, line 895: we should check `conditionalUpdates` instead of `availableUpdates` here, right? Since availableUpdates existed does not mean end of conditionupdate sync. And i would suggest to move it out of set_channel func and hosted in line623.

JianLi-RH: done

jiajliu: and this one miss?

>  i would suggest to move it out of set_channel func and hosted in line623.

1. to align with https://github.com/openshift/release/pull/49540#discussion_r1535044082, make set_channel only do one thing `set channel` in case there will be other scenarios which need set_channel step but not wait for a conditional update check. 
2. it would exit earlier if `[[ " $conditional_updates " =~ " $TARGET_VERSION " ]]`

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
jianlinliu: That move-out is because on that time, it is checking image available or not, if no any available image, directly exit as failure, that is I suggest to move it out. But in later Jian Li's testing, introduced this loop is waiting for cvo can get available update for upstream server because of some delay, if no any available update, this would not cause any failure, but just a warning, the purpose of this loop is ensure cvo get connected with Cincinnati server, they are different.

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
JianLi-RH: done

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
jiajliu: so you want it always wait for conditionupdate after set_channel, right? but there is some test does not care about conditionupdate, it will also wait in this scenario. 
personally +1 for your original thought on `set channel is only doing set channel`, and leaving wait or not and waiting for what to the following steps depending on the specified test scenario. 
considering 3min sounds not too bad, so leave it to you and jian

jianlinliu: Always waiting less than 3 mins should be harmless, so I think it is acceptable.

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
```diff
@@ -618,6 +618,25 @@ function admin_ack() {
 
 # Upgrade the cluster to target release
 function upgrade() {
+    set_channel $TARGET_VERSION
+    local retry=3 conditional_updates
+    while (( retry > 0 )); do
```
jianlinliu, line 623: The similar comments to https://github.com/openshift/release/pull/50446#discussion_r1547164336, if here we are using while loop, I guess https://github.com/openshift/release/pull/50446/files#diff-b1c231163eb97f0cd23233ca9867c3a719e46d494dc44546166ac0074128579dR46 also need a similar loop.

jiajliu: +1, except for this one, lgtm

jianlinliu: for `cucushift-upgrade-check-not-recommended`, we definitely need to exit it as failure, not just copy-paste the same codes.

jianlinliu: Wait, let me try to understand your comment.

JianLi-RH: I think you are right, this step is to check `not recommended`, so if the target version is not recommended we should exit 1

jianlinliu: Yeah, the purpose of the not-recommended step is for exit in advance, so I think it should be work just copying the same code. But when you said how to add loop, I was confused, I thought I missed something.

Now it looks good to me.

ci-operator/step-registry/cucushift/upgrade/check-not-recommended/cucushift-upgrade-check-not-recommended-commands.sh
```diff
@@ -0,0 +1,62 @@
...
+mkdir -p /tmp/client
+export OC_DIR="/tmp/client"
+export PATH=${OC_DIR}:$PATH
+
+echo "Upgrade target is ${OPENSHIFT_UPGRADE_RELEASE_IMAGE_OVERRIDE}"
+export TARGET="${OPENSHIFT_UPGRADE_RELEASE_IMAGE_OVERRIDE}"
+TARGET_VERSION="$(env "NO_PROXY=*" "no_proxy=*" oc adm release info "${TARGET}" --output=json | jq -r '.metadata.version')"
+extract_oc
+
+local retry=3 conditional_updates
```
jianlinliu, line 46:
```suggestion
retry=3
```

jianlinliu:
Let us merge it now. I will open a new PR to fix some other issues soon, in that PR, I can trigger some rehearse job to confirm no regression issue happened.

PR #50455, opened 2024-04-02
OCM-6387 Add notification to `rosa-hcp-ad` nightlies

radtriste:


radtriste:
@xueli181114 @yasun1 Can you please review ?

PR #50462, opened 2024-04-02
Revert skip_if_only_changed and enroll origin to pipelines

jmguzik:


jmguzik:
/cc @dgoodwin @stbenjam 
This is follow-up of what we discussed 2 weeks ago.

PR #50463, opened 2024-04-02
Use the common install-operators step for Distributed Tracing jobs

IshwarKanse:
* Simplify the Distributed Tracing jobs by using the common install-operators step and deleting the Tracing specific steps to install operators.


vprashar2929 (approved):
/lgtm

PR #49421, opened 2024-03-01
OCPBUGS-20097: Set default image as RHEL9 for oc

ardaguclu:
This PR defines base RHEL9 images whose will be used by oc.

PR #50468, opened 2024-04-02
OCPBUGS-20097: Update oc master configuration to RHEL9

ardaguclu:
Opened this https://github.com/openshift/release/pull/49421  to support both RHEL9 and RHEL8 base images. However, bot reverted it https://github.com/openshift/release/commit/da07fcedbaf633593431b4d2e9703604326b6a54#diff-a3aabdc6d9228089f6578c4e764fc60a076217ec3a31ca778e4df1f652432de1 by removing the unused tags. 

This PR reintroduces this but directly changing the base images to RHEL9, so that bot will not remove them again. 

Problem in previous PR was that `tests` image was using RHEL8 base image which seems to be coming from https://github.com/openshift/origin/blob/master/images/tests/Dockerfile.rhel. I'm not sure how we can fix this chicken-egg problem. Because we are doing this to bump `tests` image to RHEL9.

### DIFF
--- a/ci-operator/config/openshift/oc/openshift-oc-master.yaml
+++ b/ci-operator/config/openshift/oc/openshift-oc-master.yaml
@@ -1,8 +1,8 @@
 base_images:
-  base:
+  base-rhel9:
     name: "4.16"
     namespace: ocp
-    tag: base
+    tag: base-rhel9
   dev-scripts:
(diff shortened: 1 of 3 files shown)


soltysh (approved):
/lgtm

PR #50484, opened 2024-04-02
stackrox: add dev-preview to nightlies

davdhacs:


gavin-stackrox:
> /cc @msugakov @gavin-stackrox Is this useful? Or would it be better as a periodic test run weekly like the qe-interop tests? Or, should it be reduced to a single run of qa-e2e?

@davdhacs I think it would be useful to get an early signal for the 'next' OCP release but there are some challenges:

- JIRAs will be created for failures which can trigger triage and investigation which may be premature for early releases.
- These results will show up in the #acs-nightly-ci-runs which can also trigger triage and investigation.

Maybe doing this as a weekly periodic for the attention of OCP compat/interop folks makes the most sense. I think periodics and already excluded from junit2jira and they can send results to slack if that is helpful.

PR #50497, opened 2024-04-02
[ART-9247] payload fips check step

ashwindasr:
Ref https://issues.redhat.com/browse/ART-9247

### DIFF
--- a/ci-operator/step-registry/openshift/release/fips/openshift-release-fips-commands.sh
+++ b/ci-operator/step-registry/openshift/release/fips/openshift-release-fips-commands.sh
@@ -4,6 +4,11 @@ set -o nounset
 set -o errexit
 set -o pipefail
 
+if [[ "${RELEASE_IMAGE_LATEST}" == *"@sha256"* ]]; then
+    echo "digest based pullspecs is not supported"
+    exit 1
+fi
+
 # Get major.minor from payload url
 major_minor=$(echo "${RELEASE_IMAGE_LATEST}" | awk -F':' '{print $2}' | awk -F'.' '{print $1"."$2}')
 


ci-operator/step-registry/openshift/release/fips/OWNERS
```diff
@@ -0,0 +1,9 @@
+owners:
+- jupierce
+- bradmwilliams
+- deads2k
```
bradmwilliams, line 4: You can drop David as he'll most likely have no idea what this is all about.

ci-operator/step-registry/openshift/release/fips/openshift-release-fips-ref.metadata.json
```diff
@@ -0,0 +1,10 @@
+{
+	"path": "openshift/release/fips/openshift-release-fips-ref.yaml",
+	"owners": {
+		"approvers": [
+			"jupierce",
+			"bradmwilliams",
+			"deads2k"
```
bradmwilliams, line 7: Same

ci-operator/step-registry/openshift/release/fips/openshift-release-fips-commands.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+# Get major.minor from payload tag
+major_minor=$(echo "${PAYLOAD_TAG}" | awk -F'.' '{print $1"."$2}')
```
bradmwilliams, line 8: A word of caution about this PAYLOAD_TAG variable.  The release-controller has special logic to populate it only for "analysis" jobs.  So, you'll most likely need to rely on one of the variables passed into the steps:
https://docs.ci.openshift.org/docs/architecture/step-registry/#available-environment-variables

bradmwilliams: Yes, it'll contain the full PullSpec

ci-operator/step-registry/openshift/release/fips/openshift-release-fips-commands.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+# Get major.minor from payload url
+major_minor=$(echo "${RELEASE_IMAGE_LATEST}" | awk -F':' '{print $2}' | awk -F'.' '{print $1"."$2}')
```
bradmwilliams, line 8: You should probably defend against digest based pullspecs.  Technically, these are both valid values and point to the same release:

`registry.ci.openshift.org/ocp/release:4.16.0-0.nightly-2024-04-01-213440`
and
`registry.ci.openshift.org/ocp/release@sha256:2ff463cd4c32392e4318a37892934d4aaee6475393c9f3c3afff14b55bb94632`

bradmwilliams (changes requested):

ci-operator/step-registry/openshift/release/fips/openshift-release-fips-commands.sh
ashwindasr: So when a nightly or promoted payload is created, which value will be passed? Is there a different scenario where the digest will be passed?

bradmwilliams: I can only assure you that what comes out of the `release-controller` is a `Tag.Name` pullSpec.  However, all we are really doing is handing a Prowjob definition to Kubernetes.  What happens from that point until it gets to your step could literally be anything.  (Not saying it will, but it could and it's better to have an error that states exactly why it failed vs some crazy error from multiple piped `awk` commands).

ashwindasr: added in https://github.com/openshift/release/pull/50497/commits/8e84f777308a0ce05012f088440fcbfcd1178982

bradmwilliams (approved):

PR #50521, opened 2024-04-03
HyperShift QE: remove 4.13 jobs

LiangquanLi930:
Version 4.13 is not supported by HyperShift, so remove any unsupported jobs.


LiangquanLi930:
@liangxia PTAL, Thanks

PR #50595, opened 2024-04-04
Echo distro to stdout for openshift-docs portal build

aireilly:
Add an echo to the portal script so that we know what distro is being built when the build fails...


aireilly:
/pj-rehearse pull-ci-openshift-openshift-docs-enterprise-4.14-validate-asciidoc

gabriel-rh (approved):
LGTM

PR #50598, opened 2024-04-04
Kubevirt/HyperShift/Azure - Don't run the lane automatically - fix

orenc1:
It turned out that if the job's config specifies `skip_if_only_changed`, `always_run: false` is ignored.


davidvossel (commented):
/lgtm
/approve

PR #50604, opened 2024-04-04
[CVP-4073]: Slightly scale back cluster pool sizes and running count

maneeshmehra:


maneeshmehra:
@nmars: I think you need to also apply an LGTM label

PR #50614, opened 2024-04-04
installer: altinfra: make aws jobs truly optional

r4f4:
The jobs kept the previous `run_if_changed` in the generated config and were being triggered every time even though we intended for them to be only triggered manually.


mtulio (commented):
/lgtm
/approve

PR #50628, opened 2024-04-05
FIX: Adding  $JOB_NAME to openshift-docs comments step

aireilly:
Previously this value was hard-coded to a single job name


gabriel-rh (approved):
LGTM

PR #50632, opened 2024-04-05
Use ofcir for hypershift jobs

derekhiggins:
The equinix steps wont work when we remove the credentials.


derekhiggins:
@stbenjam @dgoodwin would you mind taking a look

PR #50642, opened 2024-04-05
OLS: added bundle image mirror

xrajesh:
Add bundle image mirror for catalog images.


bparees:
i don't think this bundle is going to be usable, but we can mirror it and check
/approve
/lgtm

bparees:
/hold

pull registry.ci.openshift.org/ols/lightspeed-operator-bundle:latest  and see what is in there before we mirror it

PR #50723, opened 2024-04-09
Add offline token for e2e cluster test

onmete:


bparees:
i used a slightly different name in vault.  good thing we rehearsed.  

i've updated the name in vault.

/pj-rehearse auto-ack

bparees:
nothing further we can do on this until CI comes back and we can successfully rehearse, but it looks like the secret was found this time, anyway.

bparees:
@onmete for rehearsals i believe you need to explicitly request the rehearsal, it's not a standard test/retest.

/pj-rehearse auto-ack

PR #50734, opened 2024-04-09
add sdn profile for 4.15 and 4.16

zhaozhanqi:
Due to this SE https://issues.redhat.com/browse/SUPPORTEX-19752, we still have customer want to use openshift-sdn because some reasons. and also sdn should still be supported upgrade from previous version to 4.15 and 4.16. so we need to cover sdn regression testing until sdn is deprecated in 4.17. 

Here setup the sdn cluster hack way in CI by setting the env `OVERRIDE_OPENSHIFT_SDN_DEPRECATION: "true"`

current plan is user cannot do upgrade from 4.16 to 4.17 for sdn.  https://issues.redhat.com/browse/SDN-4682
so we can remove this from 4.17 e2e profile and 4.16 to 4.17 upgrade profile

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__amd64-nightly.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__amd64-nightly.yaml
@@ -704,6 +704,27 @@ tests:
     test:
     - ref: fips-check-optional-operator-images-scan
     workflow: cucushift-installer-rehearse-aws-ipi-proxy
+- as: aws-ipi-proxy-sdn-workers-rhel8-fips-f14
+  cron: 21 17 10,25 * *
+  steps:
+    cluster_profile: aws-qe
+    env:
+      BASE_DOMAIN: qe.devcluster.openshift.com
+      OVERRIDE_OPENSHIFT_SDN_DEPRECATION: "true"
+    test:
+    - chain: openshift-e2e-test-qe
+    workflow: cucushift-installer-rehearse-aws-ipi-proxy-sdn-workers-rhel8-fips
+- as: aws-ipi-proxy-sdn-workers-rhel8-fips-f28-destructive
+  cron: 8 5 7 * *
+  steps:
+    cluster_profile: aws-qe
+    env:
+      BASE_DOMAIN: qe.devcluster.openshift.com
+      OVERRIDE_OPENSHIFT_SDN_DEPRECATION: "true"
+      TEST_ADDITIONAL: Network_Observability
+    test:
+    - chain: openshift-e2e-test-qe-destructive
+    workflow: cucushift-installer-rehearse-aws-ipi-proxy-sdn-workers-rhel8-fips
 - as: aws-ipi-registry-s3-cloudfront-f5
   cron: 13 2 5,10,15,20,25,30 * *
   steps:
@@ -2847,6 +2868,15 @@ tests:
     test:
     - chain: openshift-e2e-test-qe-destructive
     workflow: cucushift-installer-rehearse-vsphere-ipi-remote-worker
+- as: vsphere-ipi-sdn-migration-ovn-f14
+  cron: 36 19 13,28 * *
+  steps:
+    cluster_profile: vsphere-2
+    env:
+      OVERRIDE_OPENSHIFT_SDN_DEPRECATION: "true"
+    test:
+    - chain: cucushift-sdn-migration-ovn
+    workflow: cucushift-installer-rehearse-vsphere-ipi-sdn
 - as: vsphere-ipi-static-ip-tp-f28
   cron: 26 11 21 * *
   steps:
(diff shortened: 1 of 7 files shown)


ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__amd64-nightly.yaml
```diff
@@ -2847,6 +2868,19 @@ tests:
     test:
     - chain: openshift-e2e-test-qe-destructive
     workflow: cucushift-installer-rehearse-vsphere-ipi-remote-worker
+- as: vsphere-ipi-sdn-migration-ovn-f14
+  cron: 36 19 13,28 * *
+  steps:
+    cluster_profile: vsphere-2
+    env:
+      OVERRIDE_OPENSHIFT_SDN_DEPRECATION: "true"
+      UPGRADE_CHECK_RUN_TAGS: '@4.15 and @vsphere-ipi and @upgrade-check and @network-ovnkubernetes
```
jianlinliu, line 2877: I think we do not need the TAG setting now.

zhaozhanqi: yes,  I removed them. thanks.

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__amd64-nightly.yaml
```diff
@@ -704,6 +704,27 @@ tests:
     test:
     - ref: fips-check-optional-operator-images-scan
     workflow: cucushift-installer-rehearse-aws-ipi-proxy
+- as: aws-ipi-proxy-sdn-workers-rhel8-fips-f14
+  cron: 21 17 10,25 * *
```
jianlinliu, line 708: suggest to lower the frequency to f14.

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__amd64-nightly.yaml
jianlinliu: cc @liangxia to confirm if I am wrong.

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__amd64-nightly.yaml
zhaozhanqi: Hmm, can we keep 14?  because I do not add more sdn profile this time, only this one on aws

jianlinliu: we also have a destructive beside this one, if it is on f28, that means at least 2 jobs in one month, of course, if you think that is not enough, on your call.

liangxia: @zhaozhanqi The test name has fips but there are no env to enable fips.

liangxia: @zhaozhanqi The test name has fips but there are no env to enable fips.

jianlinliu: this is a fips cluster with rhel worker node, so `FIPS_ENABLED` is set in the install workflow, but not set it in the job config level.

liangxia: Ack

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__amd64-nightly.yaml
liangxia: Ack

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15__amd64-nightly.yaml
zhaozhanqi: > we also have a destructive beside this one, if it is on f28, that means at least 2 jobs in one month, of course, if you think that is not enough, on your call.

yes, hmm, (2 normal and 1 destructive) per 28 days. should be ok I think.

liangxia: cucushift check `.data` in `configmap/cluster-config-v1`( in namespace `kube-system`) to determine whether or not to include fips test cases. So it should be fine.

PR #50793, opened 2024-04-10
OLS-434: Add weekly response evaluation

asamal4:
New weekly job to do sanity check for responses.


bparees:
@asamal4 how about making the job report the results to our slack channel?

see: https://docs.ci.openshift.org/docs/how-tos/notification/

PR #50802, opened 2024-04-10
add openshift-ci config for konflux workspaces

sadlerap:
This adds the configuration to run openshift-ci on https://github.com/konflux-workspaces/workspaces.  For now, we only need to run e2e tests here.

### DIFF
--- a/ci-operator/config/konflux-workspaces/workspaces/konflux-workspaces-workspaces-main.yaml
+++ b/ci-operator/config/konflux-workspaces/workspaces/konflux-workspaces-workspaces-main.yaml
@@ -27,56 +27,62 @@ tests:
     product: ocp
     timeout: 1h0m0s
     version: "4.14"
-  run_if_changed: ^.github/workflows/push.yaml$|^server|^operator|^ci
+  run_if_changed: ^.github/workflows/push.yaml$|^server|^operator|^ci|^e2e
   steps:
     test:
     - as: e2e
       cli: latest
       commands: |
+        # TODO(sadlerap): take this out once we unify tooling
         make -C server kustomize
+
+        # set up environment variables
         export KUSTOMIZE="$(realpath server/bin/kustomize)"
         export TAG="pr-${PULL_NUMBER}-${PULL_PULL_SHA:0:8}"
-        export OPERATOR_IMG="quay.io/konflux-workspaces/workspaces-operator:${TAG}"
-        export SERVER_IMG="quay.io/konflux-workspaces/workspaces-server:${TAG}"
+        export QUAY_REPO="quay.io/konflux-workspaces"
+        export OPERATOR_IMG="${QUAY_REPO}/workspaces-operator:${TAG}"
+        export SERVER_IMG="${QUAY_REPO}/workspaces-server:${TAG}"
         export NAMESPACE=workspaces-system
         export USE_INSECURE_TLS=true
-        unset REPO_NAME
+        export SERVER_LOG_LEVEL=-4
+        export E2E_POLL_TIMEOUT=5m
         echo "${TAG}"
 
+        # needed because the build system for kubesaw deploys the latest
+        # version whenever we make a change to openshift/release, which
+        # makes testing very difficult
+        unset REPO_NAME
+
+        # wait for kubesaw images to become available
         while [[ -z "$(skopeo list-tags docker://quay.io/konflux-workspaces/host-operator | grep ${TAG})" ]]; do
             sleep 5
         done
+
+        # deploy kubesaw
         ci/toolchain_manager.sh deploy "${TAG}"
 
+        # wait for workspace images to become available
         while [[ -z "$(skopeo list-tags docker://quay.io/konflux-workspaces/workspaces-operator | grep ${TAG})" ]]; do
             sleep 5
         done
 
-        export SERVER_LOG_LEVEL=-4
-        export E2E_POLL_TIMEOUT=5m
+        # deploy workspaces
         make -C e2e deploy-operator deploy-server
 
-        kubectl get -n "${NAMESPACE}" routes workspaces-rest-api-server -o yaml
-        echo "+++"
-        kubectl get -n "${NAMESPACE}" services workspaces-rest-api-server -o yaml
-
+        # run tests
         if make -C e2e test; then
             exit 0
-        else
-            echo "operator logs:"
-            echo "+++"
-            kubectl logs -n "${NAMESPACE}" deployments/workspaces-controller-manager manager
-            echo "+++"
-            echo "server logs:"
-            echo "+++"
-            kubectl logs -n "${NAMESPACE}" deployments/workspaces-rest-api-server rest-api
-            echo "+++"
-            echo "proxy logs:"
-            echo "+++"
-            kubectl logs -n "${NAMESPACE}" deployments/workspaces-rest-api-server proxy
-            echo "+++"
-            exit 1
         fi
+
+        # tests failed, dump logs
+        printf "operator logs:\n+++\n"
+        kubectl logs -n "${NAMESPACE}" deployments/workspaces-controller-manager manager
+        printf "+++\nserver logs:\n+++\n"
+        kubectl logs -n "${NAMESPACE}" deployments/workspaces-rest-api-server rest-api
+        printf "+++\nproxy logs:\n+++\n"
+        kubectl logs -n "${NAMESPACE}" deployments/workspaces-rest-api-server proxy
+        echo "+++\n"
+        exit 1
       from: src
       resources:
         requests:
--- a/ci-operator/jobs/konflux-workspaces/workspaces/konflux-workspaces-workspaces-main-presubmits.yaml
+++ b/ci-operator/jobs/konflux-workspaces/workspaces/konflux-workspaces-workspaces-main-presubmits.yaml
@@ -15,7 +15,7 @@ presubmits:
       pj-rehearse.openshift.io/can-be-rehearsed: "true"
     name: pull-ci-konflux-workspaces-workspaces-main-e2e
     rerun_command: /test e2e
-    run_if_changed: ^.github/workflows/push.yaml$|^server|^operator|^ci
+    run_if_changed: ^.github/workflows/push.yaml$|^server|^operator|^ci|^e2e
     spec:
       containers:
       - args:


sadlerap:
See https://github.com/konflux-workspaces/workspaces/pull/69 for the container environment.

ci-operator/config/konflux-workspaces/workspaces/konflux-workspaces-workspaces-main.yaml
```diff
@@ -0,0 +1,39 @@
...
+    requests:
+      cpu: 100m
+      memory: 250Mi
+tests:
+- as: e2e
+  run_if_changed: ^.github/workflows/push.yaml$|^server|^operator|^ci
+  steps:
+    cluster_profile: aws-rhtap-qe
+    test:
+    - as: e2e
```
ascerra, line 25: Similar to my first comment. As of right now this works for a single test/job that you want to run but if you plan to leverage what this step is doing in any other config you'll need to copy and paste this code in each new config. Example: Some other team might want to run this test as a part of an integration test between a stack of many deployed products.

We can use the step-registry to hold repeatable tests that can be easily leveraged. This allows others to use our tests if ever needed and prevents us from ever needing to copy/paste test steps in our own configs.

Sticking with the same example in the first comment which promoted the clc-ui-e2e image, we see that image being used in [this step in the step-registry](https://github.com/openshift/release/tree/93352c979a01cb17dcdcc30dd1dfb7463e56325e/ci-operator/step-registry/acm/tests/clc-create).

2 files are needed
- .yaml file to tell this step instructions about the execution environment (image to use, variables, secrets to mount, ..etc)
- commands.sh script to hold the execution steps. lines 28-30 that  you have below would go here.

Then you can just call the step/ref that you created in your config file [like this](https://github.com/openshift/release/blob/93352c979a01cb17dcdcc30dd1dfb7463e56325e/ci-operator/config/stolostron/acmqe-autotest/stolostron-acmqe-autotest-main__acm-ocp4.16-lp-interop.yaml#L87)

ci-operator/config/konflux-workspaces/workspaces/konflux-workspaces-workspaces-main.yaml
```diff
@@ -0,0 +1,39 @@
...
+resources:
+  '*':
+    requests:
+      cpu: 100m
+      memory: 250Mi
+tests:
+- as: e2e
+  run_if_changed: ^.github/workflows/push.yaml$|^server|^operator|^ci
+  steps:
+    cluster_profile: aws-rhtap-qe
```
ascerra, line 23: I don't believe this is how the aws-rhtap-qe cluster_profile is being used by our team. @psturc @flacatus you might need to help me here but what I understand is that we exclusively use cluster_claims.

I think this because doing a [search for our cluster_profile](https://github.com/search?q=repo%3Aopenshift%2Frelease%20aws-rhtap-qe&type=code) shows that its not used in any config.

We are using [cluster_claim like this to get a cluster for the job](https://github.com/openshift/release/blob/93352c979a01cb17dcdcc30dd1dfb7463e56325e/ci-operator/config/konflux-ci/remote-secret/konflux-ci-remote-secret-main.yaml#L29) @sadlerap this is much different then how the interop team got clusters for testing that you've already seen.

ci-operator/config/konflux-workspaces/workspaces/konflux-workspaces-workspaces-main.yaml
```diff
@@ -0,0 +1,39 @@
...
+      cli: latest
+      commands: |
+        export OPERATOR_IMAGE="quay.io/konflux-workspaces/workspaces-operator:pr-${PULL_NUMBER}-${PULL_PULL_SHA:0:8}"
+        export SERVER_IMAGE="quay.io/konflux-workspaces/workspaces-server:pr-${PULL_NUMBER}-${PULL_PULL_SHA:0:8}"
+        make -C e2e deploy-operator deploy-server test
+      from: src
+      resources:
+        requests:
+          cpu: 100m
+    workflow: ipi-aws
```
ascerra, line 35: Using cluster_claim to get a cluster would remove the need for using this workflow.

ci-operator/config/konflux-workspaces/workspaces/konflux-workspaces-workspaces-main.yaml
sadlerap: Fixed in a2e513179b8.

ci-operator/config/konflux-workspaces/workspaces/konflux-workspaces-workspaces-main.yaml
```diff
@@ -0,0 +1,39 @@
+build_root:
+  project_image:
+    dockerfile_path: ci/Dockerfile
```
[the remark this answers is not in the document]
sadlerap: I'm of the opinion that we can add this once we start needing more jobs.  For the short-term future (i.e. the next few months), I think we're going to only be running e2e tests here.

ci-operator/config/konflux-workspaces/workspaces/konflux-workspaces-workspaces-main.yaml
sadlerap: I'm of a similar mindset here; I think this would be nice to have, but I'm thinking I'll punt this to a future PR if you're okay with that.  As much as I want to make this change, I think I'd like to get end-to-end tests working in CI first.

sadlerap:
/cc @filariow

I'd like to make sure I'm running tests correctly, especially the changes in e42bf0b3fcd.

ci-operator/config/konflux-workspaces/workspaces/konflux-workspaces-workspaces-main.yaml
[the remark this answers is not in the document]
ascerra: That makes sense to me, I agree this can be added in the future no problem if we have a need for it.

ci-operator/config/konflux-workspaces/workspaces/konflux-workspaces-workspaces-main.yaml
ascerra: Again I agree, it can be nice to get things running first then decide how to optimize if the need arises later.

ci-operator/config/konflux-workspaces/workspaces/konflux-workspaces-workspaces-main.yaml
```diff
@@ -0,0 +1,48 @@
...
+  run_if_changed: ^.github/workflows/push.yaml$|^server|^operator|^ci
+  cluster_claim:
+    architecture: amd64
+    cloud: aws
+    labels:
+      region: us-west-2
+    owner: rhtap
+    product: ocp
+    timeout: 1h0m0s
+    version: "4.15"
```
ascerra, line 30: It looks like we only have clusterpools setup for 4.14 as the latest right now
https://github.com/openshift/release/tree/d83b0661d19b9f97c9a2a16118a51bd31fc8f6df/clusters/hive/pools/rhtap

I'm not sure if there is a reason not to have 4.15 besides maybe just the maintenance of making the switch. @psturc @flacatus ?

psturc: Konflux staging clusters [were migrated to 4.14 very recently](https://redhat-internal.slack.com/archives/C02CTEB3MMF/p1712776304010559), so we did not need to test with 4.15 on PR basis yet. But now creating 4.15 cluster pool makes sense.

@flacatus on which version of ocp the RHTAP is tested (and is supported)?

sadlerap: I had a chat with @dperaza4dustbit, sounds like we should be targeting 4.14 for now.  We can add 4.15 to the mix in a future PR once 4.15 clusters are available in the cluster pool (which will likely be a separate PR).

ci-operator/config/konflux-workspaces/workspaces/konflux-workspaces-workspaces-main.yaml
```diff
@@ -0,0 +1,48 @@
...
+    requests:
+      cpu: 100m
+      memory: 250Mi
+tests:
+- as: e2e
+  cluster_claim:
+    architecture: amd64
+    cloud: aws
+    labels:
+      region: us-west-2
```
ascerra, line 25: Try changing this to us-east-2 that is how I'm seeing it used for the [RHTAP installer job](https://github.com/openshift/release/blob/3f303c3dec138a39d9938a5aa9c3c86c5d4023e8/ci-operator/config/redhat-appstudio/rhtap-installer/redhat-appstudio-rhtap-installer-release.yaml#L54) which is succeeding in claiming a cluster.

sadlerap: Fixed in c152115.

psturc: yeah that's because [we don't have 4.14 cluster pool in us-west-2 region yet](https://github.com/openshift/release/tree/2114421a38f67d38e64e4a8b0ef86e3ee528e51e/clusters/hosted-mgmt/hive/pools/rhtap)

we will probably bump the version soon

psturc:
@sadlerap so the [gh-automation check is failing](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/50802/pull-ci-openshift-release-check-gh-automation-tide/1780959010279657472#1:build-log.txt%3A29-30) because it requires your repo (or org) to have proper github apps installed - [see the docs](https://docs.ci.openshift.org/docs/how-tos/onboarding-a-new-component/#granting-robots-privileges-and-installing-the-github-app)

sadlerap:
/test check-gh-automation-tide

Looks like I was missing `https://github.com/apps/openshift-merge-bot`

sadlerap:
Once https://github.com/konflux-workspaces/workspaces/pull/94 bakes and merges upstream, I should be able to fix the issues we're running into during rehearsal properly.

ci-operator/config/konflux-workspaces/workspaces/konflux-workspaces-workspaces-main.yaml
```diff
@@ -0,0 +1,87 @@
...
+  cluster_claim:
+    architecture: amd64
+    cloud: aws
+    labels:
+      region: us-east-2
+    owner: rhtap
+    product: ocp
+    timeout: 1h0m0s
+    version: "4.14"
+  run_if_changed: ^.github/workflows/push.yaml$|^server|^operator|^ci
```
filariow, line 30: should we also look for changes in `^e2e`?

ci-operator/jobs/konflux-workspaces/workspaces/konflux-workspaces-workspaces-main-presubmits.yaml
```diff
@@ -0,0 +1,72 @@
...
+    context: ci/prow/e2e
+    decorate: true
+    decoration_config:
+      skip_cloning: true
+    labels:
+      ci.openshift.io/generator: prowgen
+      pj-rehearse.openshift.io/can-be-rehearsed: "true"
+    name: pull-ci-konflux-workspaces-workspaces-main-e2e
+    rerun_command: /test e2e
+    run_if_changed: ^.github/workflows/push.yaml$|^server|^operator|^ci
```
filariow, line 18: should we also look for changes in `^e2e`?

ci-operator/config/konflux-workspaces/workspaces/konflux-workspaces-workspaces-main.yaml
```diff
@@ -0,0 +1,87 @@
...
+            echo "server logs:"
+            echo "+++"
+            kubectl logs -n "${NAMESPACE}" deployments/workspaces-rest-api-server rest-api
+            echo "+++"
+            echo "proxy logs:"
+            echo "+++"
+            kubectl logs -n "${NAMESPACE}" deployments/workspaces-rest-api-server proxy
+            echo "+++"
+            exit 1
+        fi
```
filariow, line 79: I'd suggest adding some comments to make the script easier to read and introducing some constants.

Also, would it make sense to have this code in a file inside the repo?

```suggestion
        # constants
        QUAY_REPO="quay.io/konflux-workspaces"
        OPERATOR_IMG_REPO="${QUAY_REPO}/workspaces-operator"
        SERVER_IMG_REPO="${QUAY_REPO}/workspaces-server"
        
        # build local kustomize and export to enforce its usage instead of local once
        # TODO(sadlerap): remove this workaround once build tools are unified
        make -C server kustomize
        export KUSTOMIZE="$(realpath server/bin/kustomize)"
        
        # override make targets variables for deploying the PR's images
        export TAG="pr-${PULL_NUMBER}-${PULL_PULL_SHA:0:8}"
        export OPERATOR_IMG="${OPERATOR_IMG_REPO}:${TAG}"
        export SERVER_IMG="${SERVER_IMG_REPO}:${TAG}"
        export NAMESPACE=workspaces-system

        # ...
        unset REPO_NAME
        echo "${TAG}"

        # waiting for repackaged codeready-toolchain images to be pushed
        while [[ -z "$(skopeo list-tags docker://${QUAY_REPO}/host-operator | grep ${TAG})" ]]; do
            sleep 5
        done

        # deploying codeready-toolchain
        ci/toolchain_manager.sh deploy "${TAG}"

        # waiting for konflux-workspaces images to be pushed
        while [[ -z "$(skopeo list-tags docker://${OPERATOR_IMG_REPO} | grep ${TAG})" ]]; do
            sleep 5
        done

        #  deploy operator and server
        export USE_INSECURE_TLS=true
        export SERVER_LOG_LEVEL=-4
        make -C e2e deploy-operator deploy-server

        # log konflux-workspaces routes and services
        kubectl get -n "${NAMESPACE}" routes workspaces-rest-api-server -o yaml
        echo "+++"
        kubectl get -n "${NAMESPACE}" services workspaces-rest-api-server -o yaml

        # run e2e tests
        export E2E_POLL_TIMEOUT=5m
        if make -C e2e test; then
            exit 0
        fi
        
        # if e2e test failed, dump logs
        echo "operator logs:"
        echo "+++"
        kubectl logs -n "${NAMESPACE}" deployments/workspaces-controller-manager manager
        echo "+++"
        echo "server logs:"
        echo "+++"
        kubectl logs -n "${NAMESPACE}" deployments/workspaces-rest-api-server rest-api
        echo "+++"
        echo "proxy logs:"
        echo "+++"
        kubectl logs -n "${NAMESPACE}" deployments/workspaces-rest-api-server proxy
        echo "+++"
        exit 1
```

ci-operator/config/konflux-workspaces/workspaces/konflux-workspaces-workspaces-main.yaml
```diff
@@ -0,0 +1,87 @@
...
+        export KUSTOMIZE="$(realpath server/bin/kustomize)"
+        export TAG="pr-${PULL_NUMBER}-${PULL_PULL_SHA:0:8}"
+        export OPERATOR_IMG="quay.io/konflux-workspaces/workspaces-operator:${TAG}"
+        export SERVER_IMG="quay.io/konflux-workspaces/workspaces-server:${TAG}"
+        export NAMESPACE=workspaces-system
+        export USE_INSECURE_TLS=true
+        unset REPO_NAME
+        echo "${TAG}"
+
+        while [[ -z "$(skopeo list-tags docker://quay.io/konflux-workspaces/host-operator | grep ${TAG})" ]]; do
```
filariow, line 46: would it make sense to wrap these `while` with `timeout`s? The pipeline would timeout after 1h, but waiting for the images 1h probably just don't make sense

ci-operator/jobs/konflux-workspaces/workspaces/konflux-workspaces-workspaces-main-presubmits.yaml
```diff
@@ -0,0 +1,72 @@
+presubmits:
+  konflux-workspaces/workspaces:
+  - agent: kubernetes
+    always_run: false
+    branches:
+    - ^main$
+    - ^main-
```
filariow, line 7: is this needed?

```suggestion
```

filariow (changes requested):
Great job! Mostly LGTM, I just have some questions and suggestions

ci-operator/jobs/konflux-workspaces/workspaces/konflux-workspaces-workspaces-main-presubmits.yaml
sadlerap: Most of this file is autogenerated from the config above, and taking this out has it get replaced the next time it's refreshed.

ci-operator/config/konflux-workspaces/workspaces/konflux-workspaces-workspaces-main.yaml
sadlerap: `timeout` doesn't work great with shell builtins.  I could wrap it with a call to `bash -c`, but considering I'm going to want to revisit some other things after this lands (see previous comments from @ascerra), I'm going to punt on this one until then.

ci-operator/config/konflux-workspaces/workspaces/konflux-workspaces-workspaces-main.yaml
sadlerap: Fixed in da9f87d5.

ci-operator/jobs/konflux-workspaces/workspaces/konflux-workspaces-workspaces-main-presubmits.yaml
sadlerap: Fixed in da9f87d5.

ci-operator/config/konflux-workspaces/workspaces/konflux-workspaces-workspaces-main.yaml
sadlerap: Fixed in da9f87d5.

ascerra (approved):

PR #50825, opened 2024-04-11
OpenStack: Drop kuryr jobs running in mecha central

mandre:
Mecha central runs OSP 17.1, which is not a supported platform for kuryr. In consequence, we've seen the jobs failing for a long time due to neutron issues. We can safely remove the kuryr jobs running on this environment as QE are still testing kuryr on supported platforms, and we're leaving the kuryr-nfv jobs around so we still have coverage.


ci-operator/config/shiftstack/shiftstack-ci/shiftstack-shiftstack-ci-main__periodic-4.14.yaml
```diff
@@ -218,20 +205,6 @@ tests:
     - ref: openstack-test-openstack
     workflow: openshift-e2e-openstack-ipi
   timeout: 7h0m0s
-- as: openstack-kuryr-migration
```
MaysaMacedo, line 221: The only place we run kuryr-migration is 4.14, I wonder if it's not worth to keep running it, but on the nfv cloud.

If we think it's fine to stop running it, [the workflow code](https://github.com/openshift/release/blob/master/ci-operator/step-registry/openshift/e2e/openstack/kuryr-migration/openshift-e2e-openstack-kuryr-migration-workflow.yaml) needs to be removed.

MaysaMacedo: @gryf @mandre @dulek Thoughts?

MaysaMacedo: This job has been failing for a while
https://prow.ci.openshift.org/job-history/origin-ci-test/logs/periodic-ci-shiftstack-shiftstack-ci-main-periodic-4.14-openstack-kuryr-migration

dulek: Ah, better to keep that probably, it's our escape-route from Kuryr, it should work.

mandre: I think we can remove it. I've added a second commit to remove the kuryr-migration workflow and step.

PR #50835, opened 2024-04-11
add feature-promotion-summary failures to the spyglass page

deads2k:
https://github.com/openshift/api/pull/1847 demonstrates the addition of a new `feature-promotion-summary.html` in `ARTIFACT_DIR` that is produced when a FeatureGate is moving to accessible-by-default.  This summary page should be displayed because it contains a summary of the test criteria that explains the most common causes of automation failures.


deads2k:
huh, maybe not needed?  it seems to be showing up.

PR #50839, opened 2024-04-11
Fix issue DPTP-3943 Add environment var for OIDC

bear-redhat:


bear-redhat:
All tests passed. Bot merge since no one here.

PR #50872, opened 2024-04-12
[OKO] Disable remaining clusterpool in hive cluster

viroel:
Since we are moving our clusterpools to the new Hive cluster we need to set all sizes to 0 in the remaining clusterpool. All files can be deleted afterwards.


viroel:
closing in favor of https://github.com/openshift/release/pull/50877

PR #50890, opened 2024-04-12
rhecoedge pool: fix install-config secret namespace

2uasimojo:
Missing trailing `s` puts the secret in the wrong ns and prevents the pool from deploying.


2uasimojo:
/hold

/assign @fabiendupont 

Hi Fabien. It looks like this clusterpool has never worked. This would fix it (though you'll also need to curate creds). But it looks like this has been sitting busted for a couple of years. If you don't need this pool anymore, please submit a PR to remove its configs from the release repo; and notify DPTP so they can make sure the stale artifacts are removed from the CI cluster.

/cc @bear-redhat

bear-redhat (approved):

PR #50920, opened 2024-04-15
Update Vale test step for main

aireilly:
1. Vale test step should be in the validate-asciidoc test job, not the validate portal test job.
2. Updates prow-test-2 to match main


gabriel-rh (approved):
LGTM

PR #50921, opened 2024-04-15
kubevirt: vm-console-proxy: Add CI for branch "release-v0.5"

akrejcir:
Branch `release-v0.5` will be used for CNV version 4.16.


akrejcir:
@ksimon1 do you think it is ok to override the `ci/prow/check-gh-automation` test?
It fails because the bot does not have `admin` privileges in the `vm-console-proxy` repo.

I've checked our other repos, and it has only `write` permissions everywhere:
https://github.com/kubevirt/project-infra/blob/92c7ce0d93266972d66ae9bd4693e296f25afbbf/github/ci/prow-deploy/kustom/base/configs/current/orgs/orgs.yaml#L664-L670

I assume it is safe to ignore it for now, and we can increase the permissions if some commands will not work.

PR #50930, opened 2024-04-15
Sync CRI-O owners

saschagrunert:
Sync the owners with the main CRI-O repository: 
- https://github.com/cri-o/cri-o/blob/main/OWNERS_ALIASES
- https://github.com/cri-o/cri-o/blob/main/OWNERS

cc @giuseppe @haircommander @kolyshkin @nailnd @umohnani8 @kwilczynski @hasan4791 @QiWang19 @sohankunkerkar @wgahnagl

### DIFF
--- a/ci-operator/step-registry/cri-o/cgroupv2/OWNERS
+++ b/ci-operator/step-registry/cri-o/cgroupv2/OWNERS
@@ -1,5 +1,15 @@
 approvers:
-- wgahnagl
-- rphillips
-- sohankunkerkar 
+- giuseppe
 - haircommander
+- kolyshkin
+- mrunalp
+- nalind
+- rphillips
+- saschagrunert
+- umohnani8
+reviewers:
+- hasan4791
+- kwilczynski
+- qiwang19
+- sohankunkerkar
+- wgahnagl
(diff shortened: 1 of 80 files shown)


ci-operator/step-registry/cri-o/cgroupv2/OWNERS
```diff
@@ -1,5 +1,14 @@
 approvers:
-- wgahnagl
-- rphillips
```
sohankunkerkar, line 3: can we add @rphillips to the list of OWNERS as well?

sohankunkerkar: Does it make sense to add active cri-o developers to this list rather than syncing the cri-o OWNERS list?

saschagrunert: > can we add @rphillips to the list of OWNERS as well?

Yes, sure

> Does it make sense to add active cri-o developers to this list here rather than syncing the cri-o OWNERS list?
I think we should keep the CRI-O OWNERS in sync with related repositories. We also can't add non Red Hat folks, but that's given by nature.

saschagrunert:
@sohankunkerkar PTAL again

PR #50948, opened 2024-04-15
Try a 4.12 cluster-pool cluster

DennisPeriquet:
sippy-e2e has been using 4.11 but there's an AWS cred issue on that one; maybe 4.12 might work.


DennisPeriquet:
not needed

PR #50950, opened 2024-04-15
install clang to be able to run clang-format during lint task

msherif1234:
to be able to run ebpf c code linter we need to install clang


msherif1234:
hmm not the right way to install dependency

PR #50962, opened 2024-04-16
installer: Fine tune `run_if_changed` for agent jobs

r4f4:
Only `/agent/` can match some vendored code with `agent` folders in them. Let's make the regex a bit more specific.

### DIFF
--- a/ci-operator/config/openshift/installer/openshift-installer-master.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-master.yaml
@@ -689,10 +689,10 @@ tests:
   commands: hack/go-integration-test.sh
   container:
     from: src-oc
-  run_if_changed: /agent/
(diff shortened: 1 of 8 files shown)


r4f4:
/pj-rehearse ack
Running the jobs here won't validate the triggering regex.

andfasano:
Any specific reason to limit this change to version >= 4.16? Otherwise it lgtm

r4f4:
> Any specific reason to limit this change to version >= 4.16? Otherwise it lgtm

Not really. How far back do we want to go?

andfasano:
> > Any specific reason to limit this change to version >= 4.16? Otherwise it lgtm
> 
> Not really. How far back do we want to go?

I'd include at least 4.15, as it could be the target of some backporting activity in future. Lower than that it's probably not worth the game

r4f4:
Update: include 4.15 config in the changes.

PR #50993, opened 2024-04-16
chore: remove stackrox/collector from step-registry

gavin-stackrox:
This will allow removal of the 'stackrox-push' OSCI vault.
/uncc


Molter73 (approved):
/lgtm

AFAICT and from what I can remember, we are no longer doing any patch releases on OSCI, even the collector version used with 3.74 is built on GHA, so we should be fine to remove all of this.

Mxdon123 (approved):

PR #51005, opened 2024-04-17
MULTIARCH-4563: Disable branch protection

AnnaZivkovic:
Disable branch protection in the prow config for the MTO operator so that non-prow, non-required jobs do not block tide from merging Pull Requests

### DIFF
--- a/core-services/prow/02_config/openshift/multiarch-tuning-operator/_prowconfig.yaml
+++ b/core-services/prow/02_config/openshift/multiarch-tuning-operator/_prowconfig.yaml
@@ -1,3 +1,12 @@
+branch-protection:
+  orgs:
+    openshift:
+      repos:
+        multiarch-tuning-operator:
+          branches:
+            main:
+              protect: false
+              restrictions: {}
 tide:
   queries:
   - labels:


core-services/prow/02_config/openshift/multiarch-tuning-operator/_prowconfig.yaml
```diff
@@ -1,3 +1,12 @@
+branch-protection:
+  orgs:
+    openshift:
+      repos:
+        multiarch-tuning-operator:
+          branches:
+            main:
+              protect: true
```
aleskandro, line 8: shouldn't we set this to false?

AnnaZivkovic: That would make scene, but the task said set it to true

Prashanth684: i assume this is a temporary measure?

aleskandro: Ahah my bad then 😅


Yes, we temporarily need to be able to push on main and overcome the failures in Konflux as they block us from merging and we always need to override

aleskandro (approved):
Thanks @AnnaZivkovic

PR #51009, opened 2024-04-17
Skip host FIPS validation in installer

zaneb:
Since CI does not run with FIPS mode enabled, disable checking for it in openshift-install so that https://github.com/openshift/installer/pull/7566 doesn't break jobs.

This is a follow-up to #44188 to capture more jobs.

### DIFF
--- a/ci-operator/step-registry/ipi/install/install/aws/ipi-install-install-aws-commands.sh
+++ b/ci-operator/step-registry/ipi/install/install/aws/ipi-install-install-aws-commands.sh
@@ -187,6 +187,10 @@ echo "install-config.yaml"
 echo "-------------------"
 cat ${SHARED_DIR}/install-config.yaml | grep -v "password\|username\|pullSecret\|auth" | tee ${ARTIFACT_DIR}/install-config.yaml
 
+if [ "${FIPS_ENABLED:-false}" = "true" ]; then
+    export OPENSHIFT_INSTALL_SKIP_HOSTCRYPT_VALIDATION=true
+fi
+
 # move private key to ~/.ssh/ so that installer can use it to gather logs on
 # bootstrap failure
 mkdir -p ~/.ssh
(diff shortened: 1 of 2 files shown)


gpei:
@zaneb would you mind helping to update https://github.com/openshift/release/blob/master/ci-operator/step-registry/ipi/install/install/aws/ipi-install-install-aws-commands.sh along with this PR, it's also being used for many AWS install workflows? thanks!  cc @yunjiang29 @jianlinliu

jianlinliu:
@zaneb some more question:
1. The FIPS validation would not happen when destroying a FIPS cluster using installer, right?
2. Sounds `ipi-install-install` also need the similar update.
3. Shall we also update upi-install steps in this PR? (Or you can leave that to QE, we can do similar update for UPI installs.)

gpei:
> @zaneb some more question:
> 
> 1. The FIPS validation would not happen when destroying a FIPS cluster using installer, right?
> 2. Sounds `ipi-install-install` also need the similar update.
> 3. Shall we also update upi-install steps in this PR? (Or you can leave that to QE, we can do similar update for UPI installs.)

For Question 2
it's already there -> https://github.com/openshift/release/blob/master/ci-operator/step-registry/ipi/install/install/ipi-install-install-commands.sh#L598

jianlinliu:
> it's already there

Ah, good to know, I missed that.

zaneb:
Thanks for taking a look @jianlinliu!

> The FIPS validation would not happen when destroying a FIPS cluster using installer, right?

Correct, it runs when validating the install-config. So in practice, only the first time you run `create manifests|ignition-configs|cluster` after modifying the install-config.

> Sounds ipi-install-install also need the similar update.

Yep, that was done in #44188

> Shall we also update upi-install steps for now in this PR for safe guard? (Or you can leave that to QE, we can do similar update for UPI installs.)

I couldn't find any evidence of UPI jobs that enable FIPS. I don't think there's any harm in adding it to be safe, but the list of required approvals for this PR is long already so I think if QE are able to update it in a separate PR that would be ideal.

jianlinliu:
> I couldn't find any evidence of UPI jobs that enable FIPS. I don't think there's any harm in adding it to be safe, but the list of required approvals for this PR is long already so I think if QE are able to update it in a separate PR that would be ideal.

Sounds a plan, thanks for the info.

zaneb:
/assign @sdodson
Need an approval for the installer templates, whatever those are.

PR #51017, opened 2024-04-17
add support for WRS testing

yapei:
cypress cases tagged with `@wrs` changes already in place


yapei:
@liangxia could you help review?

PR #51030, opened 2024-04-17
rebasebot: sync k-nmstate every day

mkowalski:
Rebasebot proven to be excellent for the k-nmstate use case. Now we make it run every day to get even more frequent upstream updates.


mkowalski:
/pj-rehearse ack

It works. Changing cron schedule won't make it fail

damdo (commented):
/approve

PR #51034, opened 2024-04-17
chore: change LVMS image promotion from wildcard to latest only due to DPTP-3932

jakobmoellerdev:
…


jakobmoellerdev:
closing because requested workaround was implemented

PR #51041, opened 2024-04-17
should use metallb L2Advertisement

lihongan:
We should use L2Advertisement instead of BGPAdvertisement here.

see also https://github.com/openshift/release/blob/2c96a46329c850ba7f300d4236544b1357ec6909/ci-operator/step-registry/hypershift/kubevirt/baremetalds/metallb/hypershift-kubevirt-baremetalds-metallb-commands.sh#L27-L35


lihongan:
@LiangquanLi930 @heliubj18 Please help take a look. thanks

LiangquanLi930:
https://issues.redhat.com/browse/OCPBUGS-36665

PR #50199, opened 2024-03-25
Vboulos acm210 ocp4.16

vboulos:
- Create OPP new job to run OCP 4.16 with ACM 2.10
- Update firewatch priorities to fix the cascading failures from being reported
- Update the component name in firewatch to become "OPPQE" instead of just "QE"

PR #51043, opened 2024-04-17
Don't build policy-collection images in PRs

dhaiducek:
Reset configuration changed in:
- https://github.com/openshift/release/pull/50199

FYI @gparvin -- something to watch for when jobs are updated (since this manual setting likely got overridden when the jobs were regenerated)

### DIFF
--- a/ci-operator/jobs/stolostron/policy-collection/stolostron-policy-collection-main-presubmits.yaml
+++ b/ci-operator/jobs/stolostron/policy-collection/stolostron-policy-collection-main-presubmits.yaml
@@ -1,7 +1,7 @@
 presubmits:
   stolostron/policy-collection:
   - agent: kubernetes
-    always_run: true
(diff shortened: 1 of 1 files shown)


gparvin (approved):
Thanks, I definitely didn't catch that, but will try to keep an eye out next time!

dhaiducek:
> Thanks, I definitely didn't catch that, but will try to keep an eye out next time!

Actually, this should have been set in the `config` so that it's not overwritten. I'll submit a followup PR.

dhaiducek:
> > Thanks, I definitely didn't catch that, but will try to keep an eye out next time!
> 
> Actually, this should have been set in the `config` so that it's not overwritten. I'll submit a followup PR.

Actually, that doesn't apply to the `images` jobs--we'll just have to keep an eye out for it.

PR #51050, opened 2024-04-17
Update tempest allow/skip list link

fmount:
ci-framework recently changed the link pointing to the allow/skip list, and in case a regex is not defined, tempest is going to fail. This patch updates the reference to the missing files.


viroel:
/ph-rehearse ack

The current link is invalid, since it was moved to a new location. There is no need to run a test for this change.

PR #51068, opened 2024-04-17
Make `policy-collection` image optional

dhaiducek:
Followup to:
- https://github.com/openshift/release/pull/51043


dhaiducek:
Adding `optional` to the image I think causes the image build to run on-demand rather than automatically, which is what we want.

JustinKuli (approved):

PR #51090, opened 2024-04-18
openshift-service-mesh: set VERBOSE to true in e2e tests

dgn:


luksa:
No. We said that we don't want verbose to be enabled by default, because we only want the logs of failed tests to be printed, not everything.

dgn:
> No. We said that we don't want verbose to be enabled by default, because we only want the logs of failed tests to be printed, not everything.

And how do we enable that? In the current case, there were no verbose logs for the failed test

luksa:
That's supposed to be the default behavior. What's the PR?

dgn:
> That's supposed to be the default behavior. What's the PR?

https://github.com/openshift-service-mesh/sail-operator/pull/9

and the logs are here:

https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift-service-mesh_sail-operator/9/pull-ci-openshift-service-mesh-sail-operator-main-e2e-ocp-4-15/1780901233561702400

Maybe this is actually outputting the logs for the failing test, not sure. But it sure doesn't output the log of the previous test, where "deleting operator namespace" was the cause of the failure. I'm fine with keeping it disabled though if that's the agreement

dgn:
Closing as per team agreement

PR #51113, opened 2024-04-18
[release-controller] Adding `multi-2` for multi-arch investigation

bradmwilliams:
We are standing up a new release-controller environment to investigate the necessary changes to support multi-arch payloads as similar as possible to the "other" non-multi-arch environments.


bradmwilliams:
All tags specified in the inbound imagestream will need to have the following `importPolicy` specified:
```
  importPolicy:
    importMode: PreserveOriginal
```

PR #51120, opened 2024-04-19
Remove sonarcloud-post-submit which is always failed

clyang82:
the sonarcloud-post-submit is always failed. it is useless. just remove it.


morvencao (commented):
/lgtm

PR #51127, opened 2024-04-19
installer: bump linter for go 1.21 support

r4f4:


r4f4:
Depends on https://github.com/openshift/installer/pull/8291

PR #51137, opened 2024-04-19
Implement sso for rosa hcp workflow

rhmdnd:

PR #51154, opened 2024-04-19
Implement support for using SSO authentication for ROSA

rhmdnd:
Requesting ROSA environments through CI requires using an OCM token,
which works but depending on how the account is setup, the token may be
short-lived.

This commit makes using the workflow a little easier for these cases
by incorporating support for SSO authentication, instead of relying on
an offline OCM tokens.

### DIFF
--- a/ci-operator/step-registry/rosa/dnsdomain/create/rosa-dnsdomain-create-commands.sh
+++ b/ci-operator/step-registry/rosa/dnsdomain/create/rosa-dnsdomain-create-commands.sh
@@ -6,9 +6,25 @@ set -o pipefail
 
 trap 'CHILDREN=$(jobs -p); if test -n "${CHILDREN}"; then kill ${CHILDREN} && wait; fi' TERM
 
+read_profile_file() {
+  local file="${1}"
+  if [[ -f "${CLUSTER_PROFILE_DIR}/${file}" ]]; then
+    cat "${CLUSTER_PROFILE_DIR}/${file}"
+  fi
+}
+
 # Log in
-ROSA_TOKEN=$(cat "${CLUSTER_PROFILE_DIR}/ocm-token")
-if [[ ! -z "${ROSA_TOKEN}" ]]; then
+SSO_CLIENT_ID=$(read_profile_file "sso-client-id")
+SSO_CLIENT_SECRET=$(read_profile_file "sso-client-secret")
+ROSA_TOKEN=$(read_profile_file "ocm-token")
+if [[ -n "${SSO_CLIENT_ID}" && -n "${SSO_CLIENT_SECRET}" ]]; then
+  echo "Logging into ${OCM_LOGIN_ENV} with with SSO credentials"
+  rosa login --env "${OCM_LOGIN_ENV}" --client-id "${SSO_CLIENT_ID}" --client-secret "${SSO_CLIENT_SECRET}"
+  if [ $? -ne 0 ]; then
+    echo "Login failed"
+    exit 1
+  fi
+elif [[ -n "${ROSA_TOKEN}" ]]; then
   echo "Logging into ${OCM_LOGIN_ENV} with offline token"
   rosa login --env "${OCM_LOGIN_ENV}" --token "${ROSA_TOKEN}"
   if [ $? -ne 0 ]; then
@@ -16,7 +32,7 @@ if [[ ! -z "${ROSA_TOKEN}" ]]; then
     exit 1
   fi
 else
-  echo "Cannot login! You need to specify the offline token ROSA_TOKEN!"
+  echo "Cannot login! You need to securely supply SSO credentials or an ocm-token!"
   exit 1
 fi
 


yasun1:
@rhmdnd The token stored in `ocm-token` is the offline token which is a forever lived token. The token is gotten from https://console.redhat.com/openshift/token -> Load token. 

In the future, the offline token will be deprecated, and we will use the service account to do login, like what you do in the PR
`rosa login --client-id xxxxx --client-secret xxxx`. For the migration, we need to apply the SA, apply role/permission on the SA, and updating the vault, updating all of the steps.

rhmdnd:
> @rhmdnd The token stored in `ocm-token` is the offline token which is a forever lived token. The token is gotten from https://console.redhat.com/openshift/token -> Load token.
> 
> In the future, the offline token will be deprecated, and we will use the service account to do login, like what you do in the PR `rosa login --client-id xxxxx --client-secret xxxx`. For the migration, we need to apply the SA, apply role/permission on the SA, and updating the vault, updating all of the steps.

Thanks @yasun1 - that makes sense. We are in that process now.

We have a service account, but we can't use it to get offline tokens. We must use the SSO client ID and secret to authenticate.

Based on my understanding this is the only path forward for newly provisioned service accounts to access ROSA in CI.

rhmdnd:
After looking at the rehearsal failures, I don't have a reason to believe they were caused by this change.

yasun1 (commented):
There are 17 files under [rosa](https://github.com/openshift/release/tree/master/ci-operator/step-registry/rosa) have the keywords ` with offline token`, could you update the left?

afcollins:
/pj-rehearse pull-ci-openshift-qe-ocp-qe-perfscale-ci-main-rosa_hcp-4.15-nightly-x86-control-plane-3nodes

afcollins:
/pj-rehearse pull-ci-openshift-qe-ocp-qe-perfscale-ci-main-rosa_hcp-4.15-ec-x86-data-path-9nodes

svetsa-rh:
@rhmdnd Looking through the logs of the runs executed, I see that we are still using offline token. Examples below:

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_release/51154/rehearse-51154-pull-ci-openshift-qe-ocp-qe-perfscale-ci-main-rosa_hcp-4.15-ec-x86-data-path-9nodes/1788285269439418368/artifacts/data-path-9nodes/rosa-cluster-provision/build-log.txt

Logging into staging with offline token
INFO: Logged in as 'ocm-prow-ci' on 'https://api.stage.openshift.com'
Available cluster versions:

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_release/51154/rehearse-51154-pull-ci-openshift-qe-ocp-qe-perfscale-ci-main-rosa_hcp-4.15-nightly-x86-control-plane-3nodes/1788281818223480832/artifacts/control-plane-3nodes/rosa-cluster-provision/build-log.txt

Logging into staging with offline token
INFO: Logged in as 'ocm-prow-ci' on 'https://api.stage.openshift.com'
Available cluster versions:

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_release/51154/rehearse-51154-pull-ci-openshift-qe-ocp-qe-perfscale-ci-main-rosa_hcp-4.15-nightly-x86-control-plane-3nodes/1788281818223480832/artifacts/control-plane-3nodes/rosa-cluster-wait-ready-cluster/build-log.txt

Logging into staging with offline token
INFO: Logged in as 'ocm-prow-ci' on 'https://api.stage.openshift.com'

svetsa-rh:
@rhmdnd I suggest that we validate that SA is in effect and offline tokens are not being used by running one test first and then once we confirm it all then we can run additional tests to confirm for different tests/versions etc.. if we want to make sure.

This is a straightforward one if you want to go with:
`/pj-rehearse pull-ci-openshift-qe-ocp-qe-perfscale-ci-main-rosa_hcp-4.15-nightly-x86-control-plane-3nodes`

rhmdnd:
> @rhmdnd Looking through the logs of the runs executed, I see that we are still using offline token.

I believe this is due to how your credentials are configured for CI, which is handled on a per-team basis. This patch is implemented to be backwards compatible with offline tokens since teams will likely adopt SSO authentication at different times and it's what most teams are using.

rhmdnd:
> @rhmdnd I suggest that we validate that SA is in effect and offline tokens are not being used by running one test first and then once we confirm it all then we can run additional tests to confirm for different tests/versions etc.. if we want to make sure.
> 
> This is a straightforward one if you want to go with: `/pj-rehearse pull-ci-openshift-qe-ocp-qe-perfscale-ci-main-rosa_hcp-4.15-nightly-x86-control-plane-3nodes`

For additional context, I was able to verify this approach in a separate PR using a service account: https://github.com/openshift/release/pull/51137

ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
```diff
@@ -116,14 +116,26 @@ else
...
 # Log in
-ROSA_TOKEN=$(cat "${CLUSTER_PROFILE_DIR}/ocm-token")
-if [[ ! -z "${ROSA_TOKEN}" ]]; then
-  echo "Logging into ${OCM_LOGIN_ENV} with offline token using rosa cli"
+SSO_CLIENT_ID=$(read_profile_file "sso-client-id")
+SSO_CLIENT_SECRET=$(read_profile_file "sso-client-secret")
+ROSA_TOKEN=$(read_profile_file "ocm-token")
+if [[ -n "${SSO_CLIENT_ID}" && -n "${SSO_CLIENT_SECRET}" ]]; then
+  echo "Logging into ${OCM_LOGIN_ENV} with SSO credentials"
+  rosa login --env "${OCM_LOGIN_ENV}" --client-id "${SSO_CLIENT_ID}" --client-secret "${SSO_CLIENT_SECRET}"
```
yasun1, line 132: We also need to add `ocm login --url "${OCM_LOGIN_ENV}" --client-id "${SSO_CLIENT_ID}" --client-secret "${SSO_CLIENT_SECRET}"` here to keep the login logic as same as offline token.

rhmdnd: Nice catch, thanks.

svetsa-rh:
> > @rhmdnd Looking through the logs of the runs executed, I see that we are still using offline token.
> 
> I believe this is due to how your credentials are configured for CI, which is handled on a per-team basis. This patch is implemented to be backwards compatible with offline tokens since teams will likely adopt SSO authentication at different times and it's what most teams are using.

@rhmdnd In that case, your changes to https://github.com/openshift/release/pull/51154/files#diff-bd4255bf5f115420773e17473bf3b750d1c537523741283bfc96c91168d50fac and deprovision will break our tests unless ours is also updated right? And same for other teams that are using the common workflow?

rhmdnd:
> > > @rhmdnd Looking through the logs of the runs executed, I see that we are still using offline token.
> > 
> > 
> > I believe this is due to how your credentials are configured for CI, which is handled on a per-team basis. This patch is implemented to be backwards compatible with offline tokens since teams will likely adopt SSO authentication at different times and it's what most teams are using.
> 
> @rhmdnd In that case, your changes to https://github.com/openshift/release/pull/51154/files#diff-bd4255bf5f115420773e17473bf3b750d1c537523741283bfc96c91168d50fac and deprovision will break our tests unless ours is also updated right? And same for other teams that are using the common workflow?

It should be backwards compatible by checking for the SSO credentials first, and if those variables are not set, it will fall back to using an offline token. So if you haven't switched over to using SSO authentication yet, all your jobs should still work.

Do you have a rehearsal job we can run to test it out on this PR?

rhmdnd:
Failure for pull-ci-openshift-qe-ocp-qe-perfscale-ci-main-rosa_hcp-4.15-ec-x86-data-path-9nodes look unrelated, but the job is still using offline tokens so that's good.

PR #51172, opened 2024-04-21
DNM: Remove failing periodic-image-mirroring-addon-operator

deepsm007:
Removing the failing periodic-image-mirroring-addon-operator along with the mappings file. Fix the periodic to keep the mirroring job.
https://prow.ci.openshift.org/job-history/gs/test-platform-results/logs/periodic-image-mirroring-addon-operator?buildId= 

**_Note_: Fixing the credentials may fix the job and we wouldn't need to remove them.**

Tagging @aditya-konarde @apahim @erdii @thetechnick as owners of the addon operator mappings for awareness https://github.com/openshift/release/blob/master/core-services/image-mirroring/addon-operator/OWNERS 

/hold for response from the above owners

### DIFF
--- a/ci-operator/jobs/infra-image-mirroring.yaml
+++ b/ci-operator/jobs/infra-image-mirroring.yaml
@@ -748,54 +748,6 @@ periodics:
   cron: '@hourly'
   decorate: true
   labels:
-    ci.openshift.io/area: addon-operator-dev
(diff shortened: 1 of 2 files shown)


deepsm007:
Updating the credentials and permission of the image to fix the failing periodic

PR #51179, opened 2024-04-22
Set test case timeout to 35 for WRS test

liangxia:
Context in https://redhat-internal.slack.com/archives/GH9JHQFKN/p1713769957919089?thread_ts=1710981765.784209&cid=GH9JHQFKN


kuiwang02:
I am ok

PR #51245, opened 2024-04-23
Simplify enable-stage-catalogsource step

IshwarKanse:
The PR simplifies the enable-stage-catalogsource step by using a configmap for trusted CA certs instead of propagating the certs to OCP nodes using machine config which is slow.


vprashar2929 (approved):
/lgtm

PR #51253, opened 2024-04-23
Fix dockerfile path in oc tools image

soltysh:


ardaguclu:
It seems that bot aligned with the changes in tools by converting the base image to base-rhel9 but couldn't incorporate the fix in here.

soltysh:
This worked as expected, no need to test more. 
/pj-rehearse skip

PR #51258, opened 2024-04-23
Update ibmcloudz automation flavor to support OCP 4.15

dale-fu:
Adding an additional automation flavor image for OCP 4.15 deployment on s390x for Stackrox upstream CI.

### DIFF
--- a/core-services/image-mirroring/_config.yaml
+++ b/core-services/image-mirroring/_config.yaml
@@ -525,6 +525,8 @@ supplementalCIImages:
     image: quay.io/stackrox-io/ci:automation-flavors-rosa-0.9.1
   stackrox/automation-flavors:ibmcloudz-0.9.1:
     image: quay.io/stackrox-io/ci:automation-flavors-ibmcloudz-0.9.1
+  stackrox/automation-flavors:ibmcloudz-0.10.8:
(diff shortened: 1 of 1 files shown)


core-services/image-mirroring/_config.yaml
```diff
@@ -523,8 +523,8 @@ supplementalCIImages:
     image: quay.io/stackrox-io/ci:automation-flavors-aro-0.9.1
   stackrox/automation-flavors:rosa-0.9.1:
     image: quay.io/stackrox-io/ci:automation-flavors-rosa-0.9.1
-  stackrox/automation-flavors:ibmcloudz-0.9.1:
-    image: quay.io/stackrox-io/ci:automation-flavors-ibmcloudz-0.9.1
```
gavin-stackrox, line 527: You have to keep the old ones for existing dependencies not to break during the switch.

gavin-stackrox (approved):

gavin-stackrox (changes requested):
You will need to keep 0.9.1

core-services/image-mirroring/_config.yaml
dale-fu: Understood. I kept the old one.

PR #51259, opened 2024-04-23
Add bastion ping

jtaleric:
Test connectivity between Prow and our bastion host


ci-operator/step-registry/openshift-qe/installer/bm/ping/openshift-qe-installer-bm-ping-commands.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -o errexit
+set -o nounset
+set -o pipefail
+set -x
+
+bastion=$(cat "/secret/hostname")
+
+ping -c 5 $bastion
```
vishnuchalla, line 9: Looks like just `hostname` is being used to do a ping. Are we using ssh key as well?

jtaleric: Why do we need to use the sshkey to ping?

krishvoor (approved):
/lgtm

svetsa-rh:
@jtaleric can we kick off a test prior to merging?

PR #51287, opened 2024-04-24
Support oc adm upgrade rollback in CI

shellyyang1989:
The change adds support for oc adm upgrade rollback cli.

Refers [OTA-1071](https://issues.redhat.com/browse/OTA-1071)

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__amd64-rollback-ec.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__amd64-rollback-ec.yaml
@@ -62,10 +62,14 @@ build_root:
     tag: golang-1.18
 releases:
   latest:
-    release:
+    prerelease:
       architecture: amd64
-      channel: candidate
-      version: 4.16.0-ec.0
+      product: ocp
+      relative: 1
+      version_bounds:
+        lower: 4.16.0-0
+        stream: 4-dev-preview
+        upper: 4.17.0-0
   target:
     prerelease:
       architecture: amd64
@@ -89,19 +93,7 @@ tests:
     env:
       BASE_DOMAIN: qe.devcluster.openshift.com
     test:
-    - as: set-upgrade-releases
-      commands: echo "${RELEASE_IMAGE_TARGET},${RELEASE_IMAGE_LATEST}" > ${SHARED_DIR}/upgrade-edge
-      dependencies:
-      - env: RELEASE_IMAGE_TARGET
-        name: release:target
-      - env: RELEASE_IMAGE_LATEST
-        name: release:latest
-      from: cli
-      resources:
-        requests:
-          cpu: 100m
-          memory: 200Mi
-    - ref: cucushift-chainupgrade-toimage
+    - chain: openshift-upgrade-qe-sanity-rollback
     workflow: cucushift-installer-rehearse-aws-ipi-byo-route53
 zz_generated_metadata:
   branch: release-4.16
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__amd64-rollback-nightly.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__amd64-rollback-nightly.yaml
@@ -58,12 +58,10 @@ build_root:
     tag: golang-1.18
 releases:
   latest:
-    candidate:
+    release:
       architecture: amd64
-      product: ocp
-      relative: 1
-      stream: nightly
-      version: "4.16"
+      channel: candidate
+      version: 4.16.0-rc.0
   target:
     candidate:
       architecture: amd64
@@ -85,19 +83,7 @@ tests:
     env:
       BASE_DOMAIN: qe.devcluster.openshift.com
     test:
-    - as: set-upgrade-releases
-      commands: echo "${RELEASE_IMAGE_TARGET},${RELEASE_IMAGE_LATEST}" > ${SHARED_DIR}/upgrade-edge
-      dependencies:
-      - env: RELEASE_IMAGE_TARGET
-        name: release:target
-      - env: RELEASE_IMAGE_LATEST
-        name: release:latest
-      from: cli
-      resources:
-        requests:
-          cpu: 100m
-          memory: 200Mi
-    - ref: cucushift-chainupgrade-toimage
+    - chain: openshift-upgrade-qe-sanity-rollback
     workflow: cucushift-installer-rehearse-aws-ipi-byo-route53
 - as: aws-ipi-ovn-hypershift-inplace-f7
   cron: 34 23 7,14,23,30 * *
@@ -110,26 +96,14 @@ tests:
     test:
     - chain: openshift-upgrade-qe-test-hypershift-rollback
     workflow: cucushift-installer-rehearse-aws-ipi-ovn-hypershift-guest
-- as: azure-upi-f14
-  cron: 13 18 15,29 * *
+- as: azure-upi-f28
+  cron: 13 18 10 * *
   steps:
     cluster_profile: azure-qe
     env:
       BASE_DOMAIN: qe.azure.devcluster.openshift.com
     test:
-    - as: set-upgrade-releases
-      commands: echo "${RELEASE_IMAGE_TARGET},${RELEASE_IMAGE_LATEST}" > ${SHARED_DIR}/upgrade-edge
-      dependencies:
-      - env: RELEASE_IMAGE_TARGET
-        name: release:target
-      - env: RELEASE_IMAGE_LATEST
-        name: release:latest
-      from: cli
-      resources:
-        requests:
-          cpu: 100m
-          memory: 200Mi
-    - ref: cucushift-chainupgrade-toimage
+    - chain: openshift-upgrade-qe-sanity-rollback
     workflow: cucushift-installer-rehearse-azure-upi
 - as: baremetal-ipi-ovn-ipv4-f28
   cluster: build05
@@ -163,78 +137,30 @@ tests:
   steps:
     cluster_profile: gcp-qe
     test:
-    - as: set-upgrade-releases
-      commands: echo "${RELEASE_IMAGE_TARGET},${RELEASE_IMAGE_LATEST}" > ${SHARED_DIR}/upgrade-edge
-      dependencies:
-      - env: RELEASE_IMAGE_TARGET
-        name: release:target
-      - env: RELEASE_IMAGE_LATEST
-        name: release:latest
-      from: cli
-      resources:
-        requests:
-          cpu: 100m
-          memory: 200Mi
-    - ref: cucushift-chainupgrade-toimage
+    - chain: openshift-upgrade-qe-sanity-rollback
     workflow: cucushift-installer-rehearse-gcp-ipi-ovn-ipsec
 - as: ibmcloud-ipi-f28
   cron: 27 0 6 * *
   steps:
     cluster_profile: ibmcloud-qe
     test:
-    - as: set-upgrade-releases
-      commands: echo "${RELEASE_IMAGE_TARGET},${RELEASE_IMAGE_LATEST}" > ${SHARED_DIR}/upgrade-edge
-      dependencies:
-      - env: RELEASE_IMAGE_TARGET
-        name: release:target
-      - env: RELEASE_IMAGE_LATEST
-        name: release:latest
-      from: cli
-      resources:
-        requests:
-          cpu: 100m
-          memory: 200Mi
-    - ref: cucushift-chainupgrade-toimage
+    - chain: openshift-upgrade-qe-sanity-rollback
     workflow: cucushift-installer-rehearse-ibmcloud-ipi
-- as: nutanix-ipi-f28
+- as: nutanix-ipi-boot-categories-project-f28
   cluster: build01
   cron: 2 14 3 * *
   steps:
     cluster_profile: nutanix-qe
     test:
-    - as: set-upgrade-releases
-      commands: echo "${RELEASE_IMAGE_TARGET},${RELEASE_IMAGE_LATEST}" > ${SHARED_DIR}/upgrade-edge
-      dependencies:
-      - env: RELEASE_IMAGE_TARGET
-        name: release:target
-      - env: RELEASE_IMAGE_LATEST
-        name: release:latest
-      from: cli
-      resources:
-        requests:
-          cpu: 100m
-          memory: 200Mi
     - ref: nutanix-upgrade-pre-config
-    - ref: cucushift-chainupgrade-toimage
+    - chain: openshift-upgrade-qe-sanity-rollback
     workflow: cucushift-installer-rehearse-nutanix-ipi-boot-categories-project
 - as: vsphere-upi-f28
   cron: 32 5 28 * *
   steps:
     cluster_profile: vsphere-2
     test:
-    - as: set-upgrade-releases
-      commands: echo "${RELEASE_IMAGE_TARGET},${RELEASE_IMAGE_LATEST}" > ${SHARED_DIR}/upgrade-edge
-      dependencies:
-      - env: RELEASE_IMAGE_TARGET
-        name: release:target
-      - env: RELEASE_IMAGE_LATEST
-        name: release:latest
-      from: cli
-      resources:
-        requests:
-          cpu: 100m
-          memory: 200Mi
-    - ref: cucushift-chainupgrade-toimage
+    - chain: openshift-upgrade-qe-sanity-rollback
     workflow: cucushift-installer-rehearse-vsphere-upi
 zz_generated_metadata:
   branch: release-4.16
--- a/ci-operator/jobs/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16-periodics.yaml
+++ b/ci-operator/jobs/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16-periodics.yaml
@@ -38208,7 +38208,6 @@ periodics:
     ci-operator.openshift.io/cloud-cluster-profile: aws-qe
     ci-operator.openshift.io/variant: amd64-rollback-nightly
     ci.openshift.io/generator: prowgen
-    job-release: "4.16"
     pj-rehearse.openshift.io/can-be-rehearsed: "true"
   name: periodic-ci-openshift-openshift-tests-private-release-4.16-amd64-rollback-nightly-aws-ipi-byo-route53-f28
   spec:
@@ -38296,7 +38295,6 @@ periodics:
     ci-operator.openshift.io/cloud-cluster-profile: aws-qe
     ci-operator.openshift.io/variant: amd64-rollback-nightly
     ci.openshift.io/generator: prowgen
-    job-release: "4.16"
     pj-rehearse.openshift.io/can-be-rehearsed: "true"
   name: periodic-ci-openshift-openshift-tests-private-release-4.16-amd64-rollback-nightly-aws-ipi-ovn-hypershift-inplace-f7
   reporter_config:
@@ -38382,7 +38380,7 @@ periodics:
         secretName: result-aggregator
 - agent: kubernetes
   cluster: build02
-  cron: 13 18 15,29 * *
+  cron: 13 18 10 * *
   decorate: true
   decoration_config:
     skip_cloning: true
@@ -38395,9 +38393,8 @@ periodics:
     ci-operator.openshift.io/cloud-cluster-profile: azure-qe
     ci-operator.openshift.io/variant: amd64-rollback-nightly
     ci.openshift.io/generator: prowgen
-    job-release: "4.16"
     pj-rehearse.openshift.io/can-be-rehearsed: "true"
-  name: periodic-ci-openshift-openshift-tests-private-release-4.16-amd64-rollback-nightly-azure-upi-f14
+  name: periodic-ci-openshift-openshift-tests-private-release-4.16-amd64-rollback-nightly-azure-upi-f28
   spec:
     containers:
     - args:
@@ -38407,8 +38404,8 @@ periodics:
       - --oauth-token-path=/usr/local/github-credentials/oauth
       - --report-credentials-file=/etc/report/credentials
       - --secret-dir=/secrets/ci-pull-credentials
-      - --secret-dir=/usr/local/azure-upi-f14-cluster-profile
-      - --target=azure-upi-f14
+      - --secret-dir=/usr/local/azure-upi-f28-cluster-profile
+      - --target=azure-upi-f28
       - --variant=amd64-rollback-nightly
       command:
       - ci-operator
@@ -38425,7 +38422,7 @@ periodics:
       - mountPath: /secrets/ci-pull-credentials
         name: ci-pull-credentials
         readOnly: true
-      - mountPath: /usr/local/azure-upi-f14-cluster-profile
+      - mountPath: /usr/local/azure-upi-f28-cluster-profile
         name: cluster-profile
       - mountPath: /secrets/gcs
         name: gcs-credentials
@@ -38484,7 +38481,6 @@ periodics:
     ci-operator.openshift.io/cluster: build05
     ci-operator.openshift.io/variant: amd64-rollback-nightly
     ci.openshift.io/generator: prowgen
-    job-release: "4.16"
     pj-rehearse.openshift.io/can-be-rehearsed: "true"
   name: periodic-ci-openshift-openshift-tests-private-release-4.16-amd64-rollback-nightly-baremetal-ipi-ovn-ipv4-f28
   spec:
@@ -38572,7 +38568,6 @@ periodics:
     ci-operator.openshift.io/cloud-cluster-profile: gcp-qe
     ci-operator.openshift.io/variant: amd64-rollback-nightly
     ci.openshift.io/generator: prowgen
-    job-release: "4.16"
     pj-rehearse.openshift.io/can-be-rehearsed: "true"
   name: periodic-ci-openshift-openshift-tests-private-release-4.16-amd64-rollback-nightly-gcp-ipi-ovn-ipsec-f28
   spec:
@@ -38660,7 +38655,6 @@ periodics:
     ci-operator.openshift.io/cloud-cluster-profile: ibmcloud-qe
     ci-operator.openshift.io/variant: amd64-rollback-nightly
     ci.openshift.io/generator: prowgen
-    job-release: "4.16"
     pj-rehearse.openshift.io/can-be-rehearsed: "true"
   name: periodic-ci-openshift-openshift-tests-private-release-4.16-amd64-rollback-nightly-ibmcloud-ipi-f28
   spec:
@@ -38749,9 +38743,8 @@ periodics:
     ci-operator.openshift.io/cluster: build01
     ci-operator.openshift.io/variant: amd64-rollback-nightly
     ci.openshift.io/generator: prowgen
-    job-release: "4.16"
     pj-rehearse.openshift.io/can-be-rehearsed: "true"
-  name: periodic-ci-openshift-openshift-tests-private-release-4.16-amd64-rollback-nightly-nutanix-ipi-f28
+  name: periodic-ci-openshift-openshift-tests-private-release-4.16-amd64-rollback-nightly-nutanix-ipi-boot-categories-project-f28
   spec:
     containers:
     - args:
@@ -38761,8 +38754,8 @@ periodics:
       - --oauth-token-path=/usr/local/github-credentials/oauth
       - --report-credentials-file=/etc/report/credentials
       - --secret-dir=/secrets/ci-pull-credentials
-      - --secret-dir=/usr/local/nutanix-ipi-f28-cluster-profile
-      - --target=nutanix-ipi-f28
+      - --secret-dir=/usr/local/nutanix-ipi-boot-categories-project-f28-cluster-profile
+      - --target=nutanix-ipi-boot-categories-project-f28
       - --variant=amd64-rollback-nightly
       command:
       - ci-operator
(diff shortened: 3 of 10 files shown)


shellyyang1989:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.16-amd64-rollback-ec-aws-ipi-byo-route53-f28

ci-operator/step-registry/cucushift/upgrade/rollback/cucushift-upgrade-rollback-commands.sh
```diff
@@ -0,0 +1,159 @@
...
+
+function run_command() {
+    local CMD="$1"
+    echo "Running command: ${CMD}"
+    eval "${CMD}"
+}
+
+# Rollback the cluster to target release
+function rollback() {
+    run_command "oc adm upgrade rollback"
```
evakhoni, line 82: I wonder, if the command is unable to rollback for some reason, will it fail the script with return code 1? if not, maybe worth an additional check on its output, checking it indeed accepted and starting

ci-operator/step-registry/cucushift/upgrade/rollback/cucushift-upgrade-rollback-commands.sh
```diff
@@ -0,0 +1,159 @@
...
+        if ! out="$(oc get clusterversion --no-headers || false)"; then
+            echo "Error occurred when getting clusterversion"
+            continue 
+        fi
+        avail="$(echo "${out}" | awk '{print $3}')"
+        progress="$(echo "${out}" | awk '{print $4}')"
+        if [[ ${avail} == "True" && ${progress} == "False" && ${out} == *"Cluster version is ${cluster_version}" ]]; then
+            echo -e "Rollback succeed on $(date "+%F %T")\n\n"
+            return 0
+        fi
```
evakhoni, line 106: so we're sleeping for 5 minutes before we do anything else, so rollback will have enough time to start, and report Progressing=True? I seen `cluster_version="${TARGET_VERSION}"` few lines above, so in case of a rollback, this is the original version right?
asking because I recall a situation where an upgrade automation failed because Progressing went temporarily to false,  and back to true where upgrade is not yet fully complete, however the script already exited

ci-operator/step-registry/cucushift/upgrade/rollback/cucushift-upgrade-rollback-commands.sh
```diff
@@ -0,0 +1,159 @@
...
+
+# Check version, state in history
+function check_history() {
+    local version state
+    version=$(oc get clusterversion/version -o jsonpath='{.status.history[0].version}')
+    state=$(oc get clusterversion/version -o jsonpath='{.status.history[0].state}')
+    if [[ ${version} == "${TARGET_VERSION}" && ${state} == "Completed" ]]; then
+        echo "History check PASSED, cluster is now rollbacked to ${TARGET_VERSION}" && return 0
+    else
+        echo >&2 "History check FAILED, cluster rollbacked to ${TARGET_VERSION} failed, current version is ${version}, exiting" && return 1
```
evakhoni, line 121: perhaps worth adding `${state}` to this print, so it'll make sense in case where 
`${version} == "${TARGET_VERSION}"` but `${state} != "Completed"`

ci-operator/step-registry/cucushift/upgrade/rollback/cucushift-upgrade-rollback-commands.sh
```diff
@@ -0,0 +1,159 @@
...
+# Target version oc will be extract in the /tmp/client directory, use it first
+mkdir -p /tmp/client
+export OC_DIR="/tmp/client"
+export PATH=${OC_DIR}:$PATH
+
+run_command "oc get machineconfig"
+
+export TARGET="${OPENSHIFT_UPGRADE_RELEASE_IMAGE_OVERRIDE}"
+TARGET_VERSION="$(env "NO_PROXY=*" "no_proxy=*" oc adm release info "${TARGET}" --output=json | jq -r '.metadata.version')"
+extract_oc
```
evakhoni, line 144: so in this case, extract_oc will extract ${TARGET} version, which is older than the current cluster. wonder if this is intended behavior

ci-operator/step-registry/cucushift/upgrade/rollback/cucushift-upgrade-rollback-commands.sh
```diff
@@ -0,0 +1,159 @@
...
+<testsuite name="cluster upgrade" tests="1" failures="0">
+  <testcase classname="cluster upgrade" name="upgrade should succeed"/>
+</testsuite>
+EOF
+    else
+      cat >"${ARTIFACT_DIR}/junit_upgrade.xml" <<EOF
+<?xml version="1.0" encoding="UTF-8"?>
+<testsuite name="cluster upgrade" tests="1" failures="1">
+  <testcase classname="cluster upgrade" name="upgrade should succeed">
+    <failure message="">openshift cluster upgrade failed</failure>
```
evakhoni, line 48: shouldn't we mention "rollback failed" instead of "upgrade" :thinking:

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__amd64-rollback-ec.yaml
```diff
@@ -89,19 +89,7 @@ tests:
...
-        name: release:target
-      - env: RELEASE_IMAGE_LATEST
-        name: release:latest
-      from: cli
-      resources:
-        requests:
-          cpu: 100m
-          memory: 200Mi
-    - ref: cucushift-chainupgrade-toimage
+    - chain: openshift-upgrade-qe-sanity-rollback-stable
```
jiajliu, line 92: do we intently to drop the support of `rollback` in nightly job?

ci-operator/step-registry/openshift/upgrade/qe/sanity/rollback-stable/openshift-upgrade-qe-sanity-rollback-stable-chain.yaml
```diff
@@ -0,0 +1,13 @@
+chain:
+  as: openshift-upgrade-qe-sanity-rollback-stable
+  steps:
+  - ref: cucushift-upgrade-setchannel
+  - ref: cucushift-upgrade-check-not-recommended
+  - ref: cucushift-upgrade-prehealthcheck
+  - ref: cucushift-upgrade-toimage
+  - ref: cucushift-upgrade-healthcheck
+  - ref: cucushift-upgrade-rollback
```
jiajliu, line 9: Could we re-use `cucushift-chainupgrade-toimage`? It looks like the only difference is `rollback` and `upgrade` sub command here, other checks almost the same.

ci-operator/step-registry/cucushift/upgrade/rollback/cucushift-upgrade-rollback-commands.sh
shellyyang1989: Do you suggest to check "Requested rollback from 4.16.0-ec.4 to 4.16.0-ec.3"? The progress of rollback is covered in the [function](https://github.com/openshift/release/pull/51287/files#diff-a003e2c46766a9d595447c0888d984d36de90873971555e2962f2e7276b4dd06R87).

ci-operator/step-registry/cucushift/upgrade/rollback/cucushift-upgrade-rollback-commands.sh
shellyyang1989: > so in case of a rollback, this is the original version
Right

>asking because I recall a situation where an upgrade automation failed because Progressing went temporarily to false

Good point. So we not only check `progressing`, but also check the `version` and `available`. Does it sound safe enough for you?

ci-operator/step-registry/cucushift/upgrade/rollback/cucushift-upgrade-rollback-commands.sh
shellyyang1989: Yes, I have the same concern. Let me confirm with Trevor.

ci-operator/step-registry/cucushift/upgrade/rollback/cucushift-upgrade-rollback-commands.sh
shellyyang1989: Yeah, mentioning rollback here is better. But changing the junit will lead to the follow-up work in RP. We need to balance between the workload and criteria of acceptance. cc @JianLi-RH for correcting me if I'm wrong.

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__amd64-rollback-ec.yaml
shellyyang1989: `rollback` cli doesn't work for nightly because it doesn't accept any options, even `--force`

ci-operator/step-registry/cucushift/upgrade/rollback/cucushift-upgrade-rollback-commands.sh
evakhoni: yeah I think it won't hurt to have a safeguard

ci-operator/step-registry/cucushift/upgrade/rollback/cucushift-upgrade-rollback-commands.sh
evakhoni: its ok I believe the combination of those three should be enough

ci-operator/step-registry/cucushift/upgrade/rollback/cucushift-upgrade-rollback-commands.sh
evakhoni: makes sense :ok:

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__amd64-rollback-ec.yaml
jiajliu: do u mean `nightly` build is not signed and `rollback` cli does not support `--force` option, so it doesn't work for nightly, but `rollback` command would not prevent it against a cluster with nightly in `cv.history` from oc cli, right?

shellyyang1989: Right, `oc` client will not error out, instead, it will send correct rollback request. But the rollback will be stuck on the signature verification as `force: false`

ci-operator/step-registry/cucushift/upgrade/rollback/cucushift-upgrade-rollback-commands.sh
shellyyang1989: Trevor guesses newer oc would be better. So I've switched to use target-versioned oc

ci-operator/step-registry/cucushift/upgrade/rollback/cucushift-upgrade-rollback-commands.sh
shellyyang1989: Updated. Thanks!

ci-operator/step-registry/cucushift/upgrade/rollback/cucushift-upgrade-rollback-commands.sh
shellyyang1989: Updated. Thanks

ci-operator/step-registry/openshift/upgrade/qe/sanity/rollback-stable/openshift-upgrade-qe-sanity-rollback-stable-chain.yaml
shellyyang1989: I dropped a mount of functions in `rollback` step, like extract_oc, admin_ack, cco related functions which are included in `chainupgrade-toimage`. They are not required in z-rollback for now. We can involve @jianlinliu @evakhoni @JianLi-RH to discuss which is better.

evakhoni: good question.. 
tbh, I'm not a big fan of the current status quo of how our steps organized in terms of a large amount of code is being copy-pasted into many different scripts (which I consider an anti-pattern) and a lack of a central utils in contrast to what we used to have before prow, **_which is quite a step back_** imo. ideally, **all** the functions code (unless _definitely_ a one-time only) should reside in a central utils, and imported by each step as needed, so then we **don't have to reuse** steps unless they do exactly as their name suggests 

as for the current **far from being ideal** state of things, it is quite a dilemma, considering that a reuse will _temporary?_ save us from yet another script that require being updated at each little change to the ci, while on the other hand a reuse will complicate things where each step now be requited to fulfill many different uses at once, some of which aren't self explanatory and contradicts the step naming.

lastly I do think a _proper_ reuse of steps, as in where it simplifies things by decreasing the overall amount of different steps to maintain, instead of complicating them, is in fact beneficial, but only as long as we create those steps to be generic enough to begin with, and not trying to force a specific step into doing what it doesn't meant to do, only because of perceivable similarities

evakhoni: perhaps if we break those large complex steps that have lots of _ifs_ inside because all the use-cases, into smaller, more generic, simple and clean steps, then we'll not only be able to reuse, but encouraged doing so. the only limiting factor may be the amount of steps each chain is able to contain without stretching the limitations of the platform, and the amount of work be required for such a paradigm change

ci-operator/step-registry/cucushift/upgrade/rollback/cucushift-upgrade-rollback-commands.sh
```diff
@@ -0,0 +1,140 @@
...
+}
+
+# Rollback the cluster to target release
+function rollback() {
+    res=$(run_command "oc adm upgrade rollback")
+    out="Requested rollback from ${TARGET_VERSION} to ${SOURCE_VERSION}"
+    if [[ ${res} == *"${out}"* ]]; then
+        echo "Rolling back cluster gets started..."
+    else
+        echo "Rolling back cluster doesn't start..."
```
evakhoni, line 68: perhaps we should fail (return 1) if we're getting unexpected output, as well as capture stderr (2>&1) into res and print the error vs the expected? wdyt?
```suggestion
    res=$(oc adm upgrade rollback 2>&1 || true)
    out="Requested rollback from ${TARGET_VERSION} to ${SOURCE_VERSION}"
    if [[ ${res} == *"${out}"* ]]; then
        echo "Rolling back cluster from ${TARGET_VERSION} to ${SOURCE_VERSION} started..."
    else
        echo "Rolling back cluster returned unexpected:\n${res}\nexpecting: ${out}"
        return 1
```

ci-operator/step-registry/cucushift/upgrade/rollback/cucushift-upgrade-rollback-commands.sh
evakhoni: looks good thanks

ci-operator/step-registry/cucushift/upgrade/rollback/cucushift-upgrade-rollback-commands.sh
evakhoni: left a [comment](https://github.com/openshift/release/pull/51287#discussion_r1586034196) below

ci-operator/step-registry/openshift/upgrade/qe/sanity/rollback-stable/openshift-upgrade-qe-sanity-rollback-stable-chain.yaml
jiajliu: > I dropped a mount of functions in rollback step, like extract_oc, admin_ack, cco related functions which are included in chainupgrade-toimage. 

Hmm, if the existed chainupgrade step could be reused here without too much change, then `upgrade+healthcheck` and `rollback+healthcheck` step could be done in one loop, just be the same with rollback nightly job.  Or else, there will be one more health check step `cucushift-upgrade-posthealthcheck` needed.

ci-operator/step-registry/cucushift/upgrade/rollback/cucushift-upgrade-rollback-commands.sh
shellyyang1989: ah, good catch!

ci-operator/step-registry/openshift/upgrade/qe/sanity/rollback-stable/openshift-upgrade-qe-sanity-rollback-stable-chain.yaml
shellyyang1989: > perhaps if we break those large complex steps that have lots of ifs inside because all the use-cases, into smaller, more generic, simple and clean steps, then we'll not only be able to reuse, but encouraged doing so.

There is limitation in Prow that a step can't be used more than once in a workflow. But we do need to re-use steps in chain upgrade. That's why we write upgrade in one step by iterating on target images. We know that it's not that readable and manageable in long run so I've opened https://issues.redhat.com/browse/OCPQE-19479 and I hope we can collaborate with the dev team to improve it.

shellyyang1989: > if the existed chainupgrade step could be reused here without too much change

If we re-use chainupgrade step, we need to have an ENV to identify the stable rollback test. And we need a `if...else` for the rollback to not extract oc and use `oc adm upgrade rollback` subcommand. Is it your suggestion?

evakhoni: oh I wasn't aware of the once per workflow step limitation. such a pity should I say.  
well, I guess other option would be spending some research into implementing a proper Utils, so we can keep our code nice and readable, and easy to maintain in one place, instead of copy-pasting over and over again. not in the current PR scope obviously.  
perhaps it is something I can spend some time on, however considering my limited experience on the subject of prow, I'll definitely benefit from some help.

shellyyang1989: Thanks Evgeni for your willingness to improve it. I'd suggest you comment in [OCPQE-19479](https://issues.redhat.com/browse/OCPQE-19479) and involve more folks in the dev and qe teams into discussion.

shellyyang1989: @jianlinliu @JianLi-RH May I have your opionion on either keeping the rollback sub-command in a separate step or integrating into chainupgrade step?

shellyyang1989: > I guess other option would be spending some research into implementing a proper Utils

Per my understanding, it's impossible to call functions outside of a step in Prow. If we want to leverage Utils model, we might need to write upgrade in a separate place. That's how ota dev does. They write upgrade in [Origin](https://github.com/openshift/origin).

evakhoni: for simplicity sake, it may worth trying `source ${SHARED_DIR}/ota/utils.sh` or something similar :thinking: 
however +1 on the dev collaboration, probably that ticket is a good place for further discussion on that subject..

shellyyang1989: Good point! But there is also limitation in the functions that `source` can execute. Agree that let's move the discussion to the jira ticket.

ci-operator/step-registry/cucushift/upgrade/rollback/cucushift-upgrade-rollback-commands.sh
```diff
@@ -0,0 +1,139 @@
...
+    fi
+}
+
+# Monitor the rollback status
+function check_rollback_status() {
+    local wait_rollback="${TIMEOUT}" out avail progress cluster_version
+    cluster_version="${TARGET_VERSION}"
```
jiajliu, line 75: should be SOURCE_VERSION here?

ci-operator/step-registry/cucushift/upgrade/rollback/cucushift-upgrade-rollback-commands.sh
```diff
@@ -0,0 +1,139 @@
...
+        echo -e "Rollback timeout on $(date "+%F %T"), exiting\n" && return 1
+    fi
+}
+
+# Check version, state in history
+function check_history() {
+    local version state
+    version=$(oc get clusterversion/version -o jsonpath='{.status.history[0].version}')
+    state=$(oc get clusterversion/version -o jsonpath='{.status.history[0].state}')
+    if [[ ${version} == "${TARGET_VERSION}" && ${state} == "Completed" ]]; then
```
jiajliu, line 104: the same here with https://github.com/openshift/release/pull/51287/files#r1593479250

ci-operator/step-registry/cucushift/upgrade/rollback/cucushift-upgrade-rollback-commands.sh
```diff
@@ -0,0 +1,139 @@
...
+        echo "History check PASSED, cluster is now rollbacked to ${TARGET_VERSION}" && return 0
```
jiajliu, line 105: and here

ci-operator/step-registry/cucushift/upgrade/rollback/cucushift-upgrade-rollback-commands.sh
```diff
@@ -0,0 +1,139 @@
...
+    source "${SHARED_DIR}/proxy-conf.sh"
+fi
+
+run_command "oc get machineconfig"
+
+export TARGET="${OPENSHIFT_UPGRADE_RELEASE_IMAGE_OVERRIDE}"
+TARGET_VERSION="$(env "NO_PROXY=*" "no_proxy=*" oc adm release info "${TARGET}" --output=json | jq -r '.metadata.version')"
+
+SOURCE_VERSION="$(oc get clusterversion --no-headers | awk '{print $2}')"
+SOURCE_MINOR_VERSION="$(echo "${SOURCE_VERSION}" | cut -f2 -d.)"
```
jiajliu, line 127: It looks like `SOURCE_MINOR_VERSION` and `TARGET_MINOR_VERSION` are not used in any other places except in log, so could be dropped, right?

ci-operator/step-registry/cucushift/upgrade/rollback/cucushift-upgrade-rollback-commands.sh
jiajliu: ignore, sounds like `"release:latest"` is the target

ci-operator/step-registry/cucushift/upgrade/rollback/cucushift-upgrade-rollback-commands.sh
```diff
@@ -0,0 +1,139 @@
...
+function run_command() {
+    local CMD="$1"
+    echo "Running command: ${CMD}"
+    eval "${CMD}"
+}
+
+# Rollback the cluster to target release
+function rollback() {
+    res=$(run_command "oc adm upgrade rollback")
+    out="Requested rollback from ${TARGET_VERSION} to ${SOURCE_VERSION}"
```
jiajliu, line 64: refer to https://github.com/openshift/release/pull/51287/files#r1593479250, should adjust ${TARGET_VERSION} with ${SOURCE_VERSION}, right?

shellyyang1989:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.16-amd64-rollback-ec-aws-ipi-byo-route53-f28 periodic-ci-openshift-openshift-tests-private-release-4.16-amd64-rollback-nightly-azure-upi-f14

ci-operator/step-registry/cucushift/upgrade/rollback/cucushift-upgrade-rollback-commands.sh
shellyyang1989: Updated.

ci-operator/step-registry/cucushift/upgrade/rollback/cucushift-upgrade-rollback-commands.sh
shellyyang1989: Updated.

ci-operator/step-registry/cucushift/upgrade/rollback/cucushift-upgrade-rollback-commands.sh
shellyyang1989: Updated.

shellyyang1989:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.16-amd64-rollback-ec-aws-ipi-byo-route53-f28

evakhoni:
maybe worth adding status dump such as https://github.com/openshift/release/blob/dc31e8e43e95a750ccf6887da247664edae0444d/ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh#L22-L25
sounds like rollback failures may be a valuable insight 
however this may go into another PR as well, just an idea

evakhoni:
all in all lgtm, leaving to @jiajliu for a final review.

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__amd64-rollback-nightly.yaml
```diff
@@ -163,78 +139,30 @@ tests:
...
-        name: release:latest
-      from: cli
-      resources:
-        requests:
-          cpu: 100m
-          memory: 200Mi
-    - ref: cucushift-chainupgrade-toimage
+    - chain: openshift-upgrade-qe-sanity-rollback
     workflow: cucushift-installer-rehearse-ibmcloud-ipi
 - as: nutanix-ipi-f28
```
jianlinliu, line 151: Shall we update the nutanix job per the install workflow together in this PR?

ci-operator/jobs/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16-periodics.yaml
```diff
@@ -38397,7 +38397,7 @@ periodics:
     ci.openshift.io/generator: prowgen
     job-release: "4.16"
     pj-rehearse.openshift.io/can-be-rehearsed: "true"
-  name: periodic-ci-openshift-openshift-tests-private-release-4.16-amd64-rollback-nightly-azure-upi-f14
+  name: periodic-ci-openshift-openshift-tests-private-release-4.16-amd64-rollback-nightly-azure-upi-f28
```
jiajliu, line 38400: we had a discussion around nightly job in https://github.com/openshift/release/pull/51287#discussion_r1578760798, wondering the signature problem is not an issue now?

ci-operator/step-registry/cucushift/upgrade/rollback/cucushift-upgrade-rollback-ref.yaml
```diff
@@ -0,0 +1,23 @@
...
+  resources:
+    limits:
+      cpu: "1"
+    requests:
+      cpu: 100m
+      memory: 200Mi
+  env:
+  - name: TIMEOUT
+    default: "130"
+    documentation: Time to wait for upgrade finish
```
jiajliu, line 20: nit: upgrade->rollback

ci-operator/jobs/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16-periodics.yaml
shellyyang1989: ah, good catch! I should have updated the latest image to rc.0 but I forgot. Rolling back from nightly to rc should be okay.

shellyyang1989:
Rollback step passed in the [reheasal](https://qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/qe-private-deck/pr-logs/pull/openshift_release/51287/rehearse-51287-periodic-ci-openshift-openshift-tests-private-release-4.16-amd64-rollback-ec-aws-ipi-byo-route53-f28/1788423847632965632).
```
INFO[2024-05-09T06:30:58Z] Running step aws-ipi-byo-route53-f28-cucushift-upgrade-rollback. 
INFO[2024-05-09T07:36:32Z] Step aws-ipi-byo-route53-f28-cucushift-upgrade-rollback succeeded after 1h5m34s. 
```

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__amd64-rollback-nightly.yaml
shellyyang1989: Updated.

ci-operator/jobs/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16-periodics.yaml
shellyyang1989: Updated.

ci-operator/step-registry/cucushift/upgrade/rollback/cucushift-upgrade-rollback-ref.yaml
shellyyang1989: Updated.

jiajliu:
lgtm

shellyyang1989:
@liangxia PTAL. Thank you!

PR #51302, opened 2024-04-24
Switch k8s tests and origin tests to use rhel9

soltysh:


ci-operator/config/openshift/origin/openshift-origin-master.yaml
```diff
@@ -15,10 +15,22 @@ base_images:
     name: "4.14"
     namespace: ocp
     tag: base
+  ocp_4.16_base:
```
ardaguclu, line 18: This is not used.

ardaguclu (commented):
/lgtm

PR #51336, opened 2024-04-25
Exclude ChkUpgrade test cases in WRS tests

liangxia:


ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__amd64-nightly.yaml
```diff
@@ -1688,7 +1688,7 @@ tests:
         IP_STACK=v4
         NETWORK_TYPE=OVNKubernetes
       E2E_RUN_TAGS: '@wrs'
-      FILTERS_ADDITIONAL: ""
```
kuiwang02, line 1691: suggest change to
```yaml
      FILTERS_ADDITIONAL: ""
      TEST_FILTERS: ~ChkUpgrade&;~MicroShiftOnly&
```
1, PreChkUpgrade only exclude pre case and ChkUpgrade exclude both pre and pst case.
2, try to set TEST_FILTERS firstly if you do not take default value of TEST_FILTERS. When to use FILTERS_ADDITIONAL, do not set TEST_FILTERS in definition to take default, but want to more filter,  so use FILTERS_ADDITIONAL

for MicroShiftOnly, if you do not need it you could not add it, but I prefer to add it.

PR #51338, opened 2024-04-25
OCPQE-21220: [4.16] move more ci jobs to multi arch ci

jianlinliu:
1. per our agreement on multi-arch payload coverage, if the job can be run in multi-arch ci, then multi arch ci has the priority, so move more gcp/aws/azure ci jobs to there.
2. remove some duplicated ci jobs, e.g:
`azure-ipi-disk-encryption-set` is already covered by `azure-ipi-des-minimal-permission-amd-mixarch`
`azure-ipi-ovn-ipsec` is already covered by `azure-ipi-ovn-ipsec-arm-mixarch`
3. rename `minimal-permission` to `mini-perm` for a shorter job name and keep consistent across all platforms.
4. decrease `ipi-ovn-ipsec-arm-mixarch` test frequency a bit, because more ci jobs are added in multi-arch ci.

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__amd64-ec.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__amd64-ec.yaml
@@ -629,7 +629,7 @@ tests:
     test:
     - chain: cucushift-installer-check-cluster-health
     workflow: cucushift-installer-rehearse-nutanix-ipi-disconnected
-- as: nutanix-ipi-proxy-fips-minimal-permission-f14
+- as: nutanix-ipi-proxy-fips-mini-perm-f14
   cluster: build01
   cron: 14 12 4,18 * *
   steps:
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__amd64-nightly-4.16-upgrade-from-stable-4.15.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__amd64-nightly-4.16-upgrade-from-stable-4.15.yaml
@@ -76,26 +76,6 @@ tests:
     test:
     - chain: openshift-upgrade-qe-test-disconnected
     workflow: cucushift-installer-rehearse-aws-c2s-ipi-disconnected-private
-- as: aws-ipi-byo-iam-role-fips-f28
-  cron: 44 8 15 * *
-  steps:
-    cluster_profile: aws-qe
-    env:
-      BASE_DOMAIN: qe.devcluster.openshift.com
-      FIPS_ENABLED: "true"
-    test:
-    - chain: openshift-upgrade-qe-test
-    workflow: cucushift-installer-rehearse-aws-ipi-byo-iam-role
-- as: aws-ipi-byo-kms-etcd-encryption-fips-f28
-  cron: 35 22 12 * *
-  steps:
-    cluster_profile: aws-qe
-    env:
-      BASE_DOMAIN: qe.devcluster.openshift.com
-      FIPS_ENABLED: "true"
-    test:
-    - chain: openshift-upgrade-qe-test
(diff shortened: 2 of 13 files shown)


jianlinliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.16-multi-nightly-4.16-upgrade-from-stable-4.15-gcp-upi-arm-f28 periodic-ci-openshift-openshift-tests-private-release-4.16-multi-nightly-4.16-upgrade-from-stable-4.15-gcp-ipi-oidc-rt-fips-amd-f14

jianlinliu:
@yunjiang29 @jinyunma @jianli-wei pls help review, also cc @JianLi-RH to be noticed.

jinyunma:
> azure-ipi-disk-encryption-set is already covered by azure-ipi-des-minimal-permission-amd-mixarch

seems that profile `azure-ipi-des-minimal-permission-amd-mixarch` is only in upgrade ci job, and profile `azure-ipi-disk-encryption-set` is removed from openshift-openshift-tests-private-release-4.16__amd64-nightly.yaml, didn't find either of them in openshift-openshift-tests-private-release-4.16__multi-nightly.yaml, but both are in 4.16 upgrade ci job, is it on purpose?

jianlinliu:
> > azure-ipi-disk-encryption-set is already covered by azure-ipi-des-minimal-permission-amd-mixarch
> 
> seems that profile `azure-ipi-des-minimal-permission-amd-mixarch` is only in upgrade ci job, and profile `azure-ipi-disk-encryption-set` is removed from openshift-openshift-tests-private-release-4.16__amd64-nightly.yaml, didn't find either of them in openshift-openshift-tests-private-release-4.16__multi-nightly.yaml, but both are in 4.16 upgrade ci job, is it on purpose?

Maybe I can update `azure-ipi-des-fips-amd-f28` in openshift-openshift-tests-private-release-4.16__multi-nightly-4.16-upgrade-from-stable-4.15.yaml (and another z-stream upgrade ci) to `azure-ipi-des-mini-perm-fips-amd-f28` to keep consistent, wdyt?

jinyunma:
> Maybe I can update `azure-ipi-des-fips-amd-f28` in openshift-openshift-tests-private-release-4.16__multi-nightly-4.16-upgrade-from-stable-4.15.yaml (and another z-stream upgrade ci) to `azure-ipi-des-mini-perm-fips-amd-f28` to keep consistent, the reason of not removing the one in upgrade ci is because we are combining fips on the install profile, wdyt?

I'm okay for the change in upgrade CI. 
But my concern is that if so, only one profile `azure-des-min-perm` is in 4.16 e2e CI, which configures DES by setting `ENABLE_DES_COMPUTE` and `ENABLE_DES_CONTROL_PLANE`. And another scenario to configure DES via ENABLE_DES_DEFAULT_MACHINE is missing in e2e CI, and I think this has high priority to test because default sc will also configure des for data disk encryption. so maybe we also need to add profile (via ENABLE_DES_DEFAULT_MACHINE)  into 4.16 e2e CI ?

jinyunma:
> or you want to switch over to use non-destructive job to cover ENABLE_DES_DEFAULT_MACHINE` ?

Checked with @Phaow , their BYOK cases are all non-destructive, so better to use non-destructive job to cover `ENABLE_DES_DEFAULT_MACHINE`.

jianlinliu:
> > or you want to switch over to use non-destructive job to cover ENABLE_DES_DEFAULT_MACHINE` ?
> 
> Checked with @Phaow , their BYOK cases are all non-destructive, so better to use non-destructive job to cover `ENABLE_DES_DEFAULT_MACHINE`.
> 
> Edit: seems that current job is as expected, using non-destructive job to cover `ENABLE_DES_DEFAULT_MACHINE`, no need to change any more. @jianlinliu thanks for checking, didn't notice this before.

Right, my last message confused you, the current jobs are there as our expectation.

jinyunma:
LGTM for azure part

jianlinliu:
cc @liangxia to review.

PR #51347, opened 2024-04-25
NO-ISSUE: Adjust late-binding job

adriengentil:
Looks like late-binding job requires more resources to run. Use a larger machine to  avoid flakiness.


adriengentil:
/pj-rehearse periodic-ci-openshift-assisted-service-master-edge-e2e-metal-assisted-kube-api-late-binding-single-node-periodic

PR #50084, opened 2024-03-21
STOR-1692: CI implementation: Test the Vault provider with Secret Store CSI driver operator

ropatil010:
Work in progress, to check if this works fine with CI implementation.
/hold

PR #51367, opened 2024-04-25
Add secrets-store-csi-driver-src image

dobsonj:
This is required by https://github.com/openshift/release/pull/50084

We're adding a new workflow that needs to execute `make e2e-vault` from the secrets-store-csi-driver source, and to do that we need to build a source image for the driver.

This has to be done as a separate PR first though, before the rehearsal for the operator will pass. Otherwise the rehearsal job fails with `"4.16:secrets-store-csi-driver-src" not found` because running the operator rehearsal does not trigger a rehearsal for the driver to actually build the source image.

/cc @openshift/storage @ropatil010


dobsonj:
/pj-rehearse pull-ci-openshift-secrets-store-csi-driver-main-images pull-ci-openshift-secrets-store-csi-driver-release-4.16-images pull-ci-openshift-secrets-store-csi-driver-release-4.17-images

PR #51386, opened 2024-04-26
CI config for ocm-container cli

rendhalver:


clcollins:
@rendhalver Looks like one of these tests is failing

rendhalver:
> @rendhalver Looks like one of these tests is failing
Yeah it seems that I needed to use a make job to build these but I can't get that to work locally.
I will revisit when I get some time.

PR #51391, opened 2024-04-26
RHOAIENG-4731 - Test gitops-update-pipeline as a part of PR check

MarianMacik:
Include github credentials also into the periodic job.

/cc @grdryn In the end, the periodic job was failing because I forgot to inject the credentials into it. I injected them just into the presubmit job. It would be good if one job was able to be configured as both presubmit and periodic.


grdryn (approved):
/lgtm

PR #51452, opened 2024-04-29
OpenStack: Make CAPO e2e jobs voting again

mdbooth:
Making these jobs optional was a temporary measure to avoid a TechPreview feature blocking a GA feature. This is no longer required.

This reverts commit 45578716852c81a28cd02b0ef199d1063e3402a9.


mdbooth:
@MaysaMacedo I think this is good now.

PR #51458, opened 2024-04-29
[CI Framework] remove pre-commit from prow jobs

cjeanner:
We moved that job to zuul pod, and are about to switch it to voting via
https://github.com/openstack-k8s-operators/ci-framework/pull/1585


cjeanner:
Now that https://github.com/openstack-k8s-operators/ci-framework/pull/1585 is merged, we can move on with this one: ready for inclusion! cc @viroel maybe? :)

PR #51465, opened 2024-04-29
Get the correct presub PR sha for kubevirt-csi conformance tests

awels:
Was using OS_BUILD_COMMIT which is not the correct sha for the PR. Instead use JOB_SPEC to get the correct SHA


awels:
/pj-rehearse pull-ci-openshift-kubevirt-csi-driver-master-e2e-kubevirt-aws-csi

PR #51472, opened 2024-04-29
PowerVS: Temporarily fix missing pip3

hamzy:
pip was removed from the installer image.  So we are switching to the upi-installer image.


hamzy:
/pj-rehearse pull-ci-openshift-installer-master-altinfra-e2e-powervs-capi-ovn

PR #51502, opened 2024-04-30
openstack: optional installer capi-manifest test

pierreprinetti:
The interface for enabling CAPI is not stable and this test may break. Don't make it voting.


r4f4 (commented):
/approve

PR #51046, opened 2024-04-17
use adm upgrade status in upgrade jobs

evakhoni:
follow up some experimentation done by @JianLi-RH and me in #50773 
adding oc adm release status to upgrade jobs to:
a. provide additional log in CI during upgrade failures
b. collect feedback from ci watchers that will be forwarded to dev for evaluation of status command usability

PR #51518, opened 2024-04-30
safeguard against OCPBUGS-32682

evakhoni:
recently added to ci `upgrade status` command via #51046 turned out to be susceptible to [OCPBUGS-32682](https://issues.redhat.com/browse/OCPBUGS-32682)
* adding a safeguard against code 1 returned by the command from interfering with upgrade-toimage step.
* there is a new `--details` flag added to `status`, which added as well to provide additional beneficial log in case of upgrade failure.
* added version >= 4.16 check in `debug()` since `status` output is not useful in older oc clients

/cc @jiajliu @jianlinliu @shellyyang1989 
you're the only reviewers according to OWNERS

edit: if you folks think it worth extracting 4.16+ oc to collect insights on an older (target<15) cluster, instead of skipping (valid use case according to dev) let me know so I can update the PR accordingly 

on a side note, how about we update OWNERS? afaik jhou1 is not really involved in ota prs review, however JianLi-RH is, and we can add me as well.. wdyt?

### DIFF
--- a/ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
+++ b/ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
@@ -19,7 +19,10 @@ KUBECONFIG="" oc --loglevel=8 registry login
 # Print cv, failed node, co, mcp information for debug purpose
 function debug() {
     if (( FRC != 0 )); then
-        echo -e "\n# oc adm upgrade status\n$(env OC_ENABLE_CMD_UPGRADE_STATUS='true' oc adm upgrade status)"
+        if [[ -n "${TARGET_MINOR_VERSION}" ]] && [[ "${TARGET_MINOR_VERSION}" -ge "16" ]] ; then
+            echo -e "\n# oc adm upgrade status\n"
+            env OC_ENABLE_CMD_UPGRADE_STATUS='true' oc adm upgrade status --details=all || true 
+        fi
         echo -e "\n# oc get clusterversion/version -oyaml\n$(oc get clusterversion/version -oyaml)"
         echo -e "\n# oc get machineconfig\n$(oc get machineconfig)"
         echo -e "\n# Describing abnormal nodes...\n"
@@ -648,9 +651,10 @@ function upgrade() {
 function dump_status_if_unexpected() {
     # expecting oc to equal TARGET_MINOR_VERSION, skip if less than .16
         if [[ -n "${TARGET_MINOR_VERSION}" ]] && [[ "${TARGET_MINOR_VERSION}" -ge "16" ]] ; then
-            local out; out="$(env OC_ENABLE_CMD_UPGRADE_STATUS=true oc adm upgrade status)"
+            local out; out="$(env OC_ENABLE_CMD_UPGRADE_STATUS=true oc adm upgrade status 2>&1 || true)"
             # if upgrading, and not progressing well, dump status to log
-            if ! grep -qE 'The cluster version is not updating|Upgrade is proceeding well' <<< "${out}" ; then
+            # "resource name may not be empty" is a known issue, remove once OCPBUGS-32682 is fixed
+            if ! grep -qE 'The cluster version is not updating|Upgrade is proceeding well|resource name may not be empty' <<< "${out}" ; then
                 echo "${out}"
             fi
         fi
--- a/ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
+++ b/ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
@@ -19,7 +19,10 @@ KUBECONFIG="" oc --loglevel=8 registry login
 # Print cv, failed node, co, mcp information for debug purpose
 function debug() {
     if (( FRC != 0 )); then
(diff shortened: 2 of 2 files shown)


evakhoni:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.16-amd64-nightly-4.16-upgrade-from-stable-4.15-aws-ipi-ovn-winc-f7
this one were observed to produce the error

jianlinliu:
>  if you folks think it worth extracting 4.16+ oc to collect insights on an older (target<15) cluster, instead of skipping (valid use case according to dev) let me know so I can update the PR accordingly

I think it should be skipped when running against an older target cluster, per our official doc, we always suggest users to run oc client with the same version as the target cluster.


> on a side note, how about we update OWNERS? afaik jhou1 is not really involved in ota prs review, however JianLi-RH is, and we can add me as well.. wdyt?

I am okay with it.

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
```diff
@@ -648,9 +651,10 @@ function upgrade() {
 function dump_status_if_unexpected() {
     # expecting oc to equal TARGET_MINOR_VERSION, skip if less than .16
         if [[ -n "${TARGET_MINOR_VERSION}" ]] && [[ "${TARGET_MINOR_VERSION}" -ge "16" ]] ; then
-            local out; out="$(env OC_ENABLE_CMD_UPGRADE_STATUS=true oc adm upgrade status)"
+            local out; out="$(env OC_ENABLE_CMD_UPGRADE_STATUS=true oc adm upgrade status 2>&1 || true)"
             # if upgrading, and not progressing well, dump status to log
-            if ! grep -qE 'The cluster version is not updating|Upgrade is proceeding well' <<< "${out}" ; then
+            # "resource name may not be empty" is a known issue, remove once OCPBUGS-32682 is fixed
+            if ! grep -qE 'The cluster version is not updating|Upgrade is proceeding well|resource name may not be empty' <<< "${out}" ; then
```
jiajliu, line 657: It sounds not a 100% reproduced bug, we will remove "resource name may not be empty" to watch it for a while in ci to help verify the issue, right?

evakhoni: oh it have been reproduced multiple times since then. [according to dev](https://redhat-internal.slack.com/archives/C03G7REB4JV/p1714478028731939?thread_ts=1713272766.656219&cid=C03G7REB4JV) it have to do with auto-scaling. I"m seeing it quite predictably on windows container clusters. 

if we leave it out of the filter here in the loop, it'll spam the logs pretty badly. 

lastly, I left it unfiltered in `function debug()` so in case of an upgrade failure we'll still going to log it once.

jiajliu: > lastly, I left it unfiltered in function debug() so in case of an upgrade failure we'll still going to log it once.

do u mean the issue also happen with `--details=all`?

evakhoni:
> I think it should be skipped when running against an older target cluster, per our official doc, we always suggest users to run oc client with the same version as the target cluster.

that's a bit of exception in this case. 
according to dev, upgrade status in its current client-side implementation, can be executed on older versions of openshift as well. dev use it (afaik with latest oc) in their CI all the way back to 4.13 I think. 
also according to @ming1013 the logs generated by upgrade status are helpful in CI watching. 

on the other hand, extracting a separate oc only for upgrade status collection complicates the implementation quite a bit, and since this PR is a quick fix for an issue that causes CI failures, and since in your opinion it can be left out as well, I'm ok with merging this one as-is, and if we later decide otherwise, I can fire another PR specifically for that enhancement. 

---
as for the owners file
> >   on a side note, how about we update OWNERS? afaik jhou1 is not really involved in ota prs review, however JianLi-RH is, and we can add me as well.. wdyt?

> I am okay with it.

just checked quickly, it appears that there is plenty of other OWNERS files for ota staff that we can update, so since its not just two lines of change, I'll do it in a separate PR then of you don't mind.

evakhoni:
> @evakhoni: all tests **passed!**
> 
> [Full PR test history](https://prow.ci.openshift.org/pr-history?org=openshift&repo=release&pr=51518). [Your PR dashboard](https://prow.ci.openshift.org/pr?query=is:pr+state:open+author:evakhoni).
> 
> Instructions for interacting with me using PR comments are available [here](https://git.k8s.io/community/contributors/guide/pull-requests.md). If you have questions or suggestions related to my behavior, please file an issue against the [kubernetes/test-infra](https://github.com/kubernetes/test-infra/issues/new?title=Prow%20issue:) repository. I understand the commands that are listed [here](https://go.k8s.io/bot-commands).

/pj-rehearse ack

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
evakhoni: positive. I checked, it happens with or without the flag, and with or without the cluster being upgraded.

jiajliu: ack, lgtm

PR #51533, opened 2024-04-30
Fix cli, cli-artifacts image generation for okd scos

Prashanth684:
In an effort to build OKD on top of centos stream(https://issues.redhat.com/browse/OKD-210), OKD-scos configs were added in https://github.com/openshift/release/pull/50118. We are going through all images and identifying the ones which do not build and are working to fix them. Changed the golang-builder to use 1.21.

### DIFF
--- a/ci-operator/config/openshift/oc/openshift-oc-master__okd-scos.yaml
+++ b/ci-operator/config/openshift/oc/openshift-oc-master__okd-scos.yaml
@@ -1,39 +1,24 @@
 base_images:
-  ocp_builder_rhel-8-golang-1.20-openshift-4.16:
-    name: builder
-    namespace: ocp
-    tag: rhel-8-golang-1.20-openshift-4.16
-  ocp_builder_rhel-8-golang-1.21-openshift-4.16:
-    name: builder
-    namespace: ocp
-    tag: rhel-8-golang-1.21-openshift-4.16
-  ocp_builder_rhel-9-golang-1.20-openshift-4.16:
-    name: builder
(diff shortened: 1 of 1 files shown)


ci-operator/config/openshift/oc/openshift-oc-master__okd-scos.yaml
```diff
@@ -11,6 +11,10 @@ base_images:
     name: builder
     namespace: ocp
     tag: rhel-9-golang-1.20-openshift-4.16
+  ocp_builder_rhel-9-golang-1.21-openshift-4.16:
+    name: builder
```
Prashanth684, line 15: @jupierce is this something that needs to be corrected in the automation?

ci-operator/config/openshift/oc/openshift-oc-master__okd-scos.yaml
```diff
@@ -43,10 +47,10 @@ images:
   dockerfile_path: images/cli-artifacts/Dockerfile.rhel
   from: cli
   inputs:
-    ocp_builder_rhel-8-golang-1.20-openshift-4.16:
+    ocp_builder_rhel-8-golang-1.21-openshift-4.16:
```
ardaguclu, line 50: Bumping to go 1.21 makes sense. However, we have switched to RHEL9 entirely https://github.com/openshift/oc/pull/1752 and I think, this configuration files should be simplified by only having rhel9-1-21 builder and removing the rest. Something similar to https://github.com/openshift/release/pull/51610

PR #51536, opened 2024-04-30
installer: remove ovirt-installer image building

r4f4:
This image is only used in the release payload so there is no reason for us to build it in CI with every installer PR. Especially because this image cross-builds the installer 4 times in a row thus slowing down the time it takes for the actual tests to run.
The ovirt-installer image is identical to the `installer` image and it is not being used anymore since ovirt/rhv was retired.

### DIFF
--- a/core-services/image-mirroring/openshift/mapping_origin_4_6
+++ b/core-services/image-mirroring/openshift/mapping_origin_4_6
@@ -7,5 +7,4 @@ registry.ci.openshift.org/origin/4.6:installer-artifacts quay.io/openshift/origi
 registry.ci.openshift.org/origin/4.6:libvirt-installer quay.io/openshift/origin-libvirt-installer:4.6 quay.io/openshift/origin-libvirt-installer:4.6.0
 registry.ci.openshift.org/origin/4.6:machine-os-content quay.io/openshift/origin-machine-os-content:4.6 quay.io/openshift/origin-machine-os-content:4.6.0
 registry.ci.openshift.org/origin/4.6:openstack-installer quay.io/openshift/origin-openstack-installer:4.6 quay.io/openshift/origin-openstack-installer:4.6.0
-registry.ci.openshift.org/origin/4.6:ovirt-installer quay.io/openshift/origin-ovirt-installer:4.6 quay.io/openshift/origin-ovirt-installer:4.6.0
 registry.ci.openshift.org/origin/4.6:upi-installer quay.io/openshift/origin-upi-installer:4.6 quay.io/openshift/origin-upi-installer:4.6.0
--- a/ci-operator/config/openshift/installer/openshift-installer-fcos.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-fcos.yaml
@@ -90,16 +90,6 @@ images:
       as:
       - registry.svc.ci.openshift.org/openshift/release:golang-1.14
   to: baremetal-installer
-- dockerfile_path: images/installer/Dockerfile.ci
-  from: base
-  inputs:
-    ocp_builder_rhel-8-golang-1.15-openshift-4.6:
-      as:
-      - registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.6
-    openshift_release_golang-1.14:
-      as:
-      - registry.svc.ci.openshift.org/openshift/release:golang-1.14
-  to: ovirt-installer
 promotion:
   to:
   - name: "4.6"
@@ -140,10 +130,6 @@ resources:
       cpu: "3"
       memory: 5Gi
   openstack-installer:
-    requests:
-      cpu: "3"
-      memory: 5Gi
-  ovirt-installer:
     requests:
       cpu: "3"
       memory: 5Gi
--- a/ci-operator/config/openshift/installer/openshift-installer-master.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-master.yaml
@@ -120,18 +120,6 @@ images:
       as:
       - registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.21-openshift-4.16
   to: baremetal-installer
-- dockerfile_path: images/installer/Dockerfile.ci
-  inputs:
-    ocp_4.16_base-rhel9:
-      as:
-      - registry.ci.openshift.org/ocp/4.16:base-rhel9
-    ocp_4.16_installer-terraform-providers:
-      as:
-      - registry.ci.openshift.org/ocp/4.16:installer-terraform-providers
-    ocp_builder_rhel-9-golang-1.21-openshift-4.16:
-      as:
-      - registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.21-openshift-4.16
-  to: ovirt-installer
 - dockerfile_literal: |
     FROM src
     COPY oc /usr/bin/oc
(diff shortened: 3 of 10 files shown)


patrickdillon:
I put a hold on to review mappings more thoroughly

patrickdillon:
/approve cancel
/lgtm cancel

I really like this idea, but 2 concerns:

1. This removes our ability to test changes to the installer-artifacts Dockerfile. That is less than like .01% of our changes, but still we need it for when we do make them.
2. Removes support for mac os in ci builds--probably not too many people using that, but still. 

Correct me if I'm wrong on any of this. 

Perhaps we want the installer-artifacts image job to just run separately. Is a variant the best way to do that?

r4f4:
/pj-rehearse pull-ci-openshift-installer-master-artifacts-images

r4f4:
/pj-rehearse pull-ci-openshift-installer-master-artifacts-verify

core-services/image-mirroring/openshift/mapping_origin_4_6
```diff
@@ -3,9 +3,7 @@ registry.ci.openshift.org/origin/4.6:branding quay.io/openshift/origin-branding:
 registry.ci.openshift.org/origin/4.6:cluster-samples-operator quay.io/openshift/origin-cluster-samples-operator:4.6 quay.io/openshift/origin-cluster-samples-operator:4.6.0
 registry.ci.openshift.org/origin/4.6:cluster-update-keys quay.io/openshift/origin-cluster-update-keys:4.6 quay.io/openshift/origin-cluster-update-keys:4.6.0
 registry.ci.openshift.org/origin/4.6:installer quay.io/openshift/origin-installer:4.6 quay.io/openshift/origin-installer:4.6.0
-registry.ci.openshift.org/origin/4.6:installer-artifacts quay.io/openshift/origin-installer-artifacts:4.6 quay.io/openshift/origin-installer-artifacts:4.6.0
```
sdodson, line 6: I'm assuming you didn't actually mean to touch 4.6 given you skipped over 4.15 - 4.7?

sdodson:
/approve
Fine with the approach and I believe that ART will tell us pretty quickly if we broke any builds whenever we don't touch Dockerfiles.

As commented, I believe you have an unintended change to 4.6 files.

patrickdillon:
> Can you explain how that works/how it's used? Are mac users getting the installer from the `installer-artifacts` image from CI instead of the release payload?

IIUC, the macos installer would no longer be present in CI builds like this: https://amd64.ocp.releases.ci.openshift.org/releasestream/4.16.0-0.ci/release/4.16.0-0.ci-2024-05-01-114754 

If you click on the link at the top that says "Download the Installer" it will take you to a page where you can directly download the mac installer binary, which is the equivalent of:

```
oc adm release extract --command-os mac --tools registry.ci.openshift.org/ocp/release:4.16.0-0.ci-2024-05-01-114754
```

I think this might also be the case if you are using clusterbot as well. So I am doubting that we want to remove macos support from the CI release image.

> A variant would still build it with every PR unless we can use a trick similar to the `terraform-providers` image. That would solve item 1. Let me play with that first.

I think it's ok--perhaps even a necessary evil--if the artifacts image would build on every PR, but we don't want it to be part of the `ci/prow/images` job which all the ci jobs depend on. If artifacts builds as a separate job that doesn't block other jobs, I think that would get us to where we want to be, right?

sdodson:
@bradmwilliams Is the downloading of Installer from the release controller something that you said may go away in the near future?

core-services/image-mirroring/openshift/mapping_origin_4_6
r4f4: This was done automatically by `make openshift-image-mirror-mappings` I had no control over it

r4f4:
> > Can you explain how that works/how it's used? Are mac users getting the installer from the `installer-artifacts` image from CI instead of the release payload?
> 
> IIUC, the macos installer would no longer be present in CI builds like this: https://amd64.ocp.releases.ci.openshift.org/releasestream/4.16.0-0.ci/release/4.16.0-0.ci-2024-05-01-114754
> 
> If you click on the link at the top that says "Download the Installer" it will take you to a page where you can directly download the mac installer binary, which is the equivalent of:
> 
> ```
> oc adm release extract --command-os mac --tools registry.ci.openshift.org/ocp/release:4.16.0-0.ci-2024-05-01-114754
> ```
> 
> I think this might also be the case if you are using clusterbot as well. So I am doubting that we want to remove macos support from the CI release image.

Aren't those release payloads built by the ART configs? In there we still have the artifacts image: https://github.com/openshift-eng/ocp-build-data/blob/openshift-4.16/images/ose-installer-artifacts.yml


> > A variant would still build it with every PR unless we can use a trick similar to the `terraform-providers` image. That would solve item 1. Let me play with that first.
> 
> I think it's ok--perhaps even a necessary evil--if the artifacts image would build on every PR, but we don't want it to be part of the `ci/prow/images` job which all the ci jobs depend on. If artifacts builds as a separate job that doesn't block other jobs, I think that would get us to where we want to be, right?

Then I don't think the gain is worth maintaining another job config.

bradmwilliams:
> @bradmwilliams Is the downloading of Installer from the release controller something that you said may go away in the near future?

Hey @sdodson,
Currently, we have been soliciting input from users of this functionality while we investigate alternatives to the way we deliver it.  There are no plans to drop this functionality outright.

r4f4:
Talking to ART, they are not familiar with how CI payloads are assembled which probably means it's using the images built from post-submits. Just to be safe, I'm going to drop the `installer-artifacts` changes from this PR and change it to only remove `ovirt-installer`.

patrickdillon:
> Talking to ART, they are not familiar with how CI payloads are assembled which probably means it's using the images built from post-submits. Just to be safe, I'm going to drop the `installer-artifacts` changes from this PR and change it to only remove `ovirt-installer`.

I extracted an installer from the altinfra image like this:
```
altinfra_image=$(oc adm release info --image-for='installer-altinfra' registry.ci.openshift.org/ocp/release:4.16.0-0.ci-2024-05-02-054924) oc image extract "$altinfra_image" --file="/usr/bin/openshift-install"
```
and it can run capi, so that would indicate it is built from ci images.

> Then I don't think the gain is worth maintaining another job config.

Hm maybe. Let's definitely take this to another PR, but I think a variant would make sense here and would be low overhead.

core-services/image-mirroring/openshift/mapping_origin_4_6
```diff
@@ -7,5 +7,4 @@ registry.ci.openshift.org/origin/4.6:installer-artifacts quay.io/openshift/origi
 registry.ci.openshift.org/origin/4.6:libvirt-installer quay.io/openshift/origin-libvirt-installer:4.6 quay.io/openshift/origin-libvirt-installer:4.6.0
 registry.ci.openshift.org/origin/4.6:machine-os-content quay.io/openshift/origin-machine-os-content:4.6 quay.io/openshift/origin-machine-os-content:4.6.0
 registry.ci.openshift.org/origin/4.6:openstack-installer quay.io/openshift/origin-openstack-installer:4.6 quay.io/openshift/origin-openstack-installer:4.6.0
-registry.ci.openshift.org/origin/4.6:ovirt-installer quay.io/openshift/origin-ovirt-installer:4.6 quay.io/openshift/origin-ovirt-installer:4.6.0
```
sadasu, line 10: Is this change necessary?

r4f4: See https://github.com/openshift/release/pull/51536#discussion_r1586851487

PR #51544, opened 2024-05-01
OCPBUGS-33152: fix: use upi-installer image for sno config update

eggfoobar:
The `installer` image was recently updated to use rhel9 which does not include `pip3`, this exposed that we are incorrectly using the installer image, we should be using the `upi-installer` image which does include tooling for CI.


eggfoobar:
/pj-rehearse ack

The `single-node-conf` step passes on all runs

PR #51554, opened 2024-05-01
OCPBUGS-33165: fix: change metal os to fix SNO metal tests running

eggfoobar:
With the change to rhel9 as the base image on 4.16 pipelines, we noticed a failure to execute sno metal tests. This is due to a mismatch of glibc versions from the rhel8 version that the test is running on and the rhel9 version that the openshift-tests binary was built with. We need to update the CI to use `rocky_9`


eggfoobar:
This PR relies on a fix to assisted-test-infra to support RHEL9, see https://github.com/openshift/assisted-test-infra/pull/2416

eggfoobar:
We're waiting on the fix from https://github.com/openshift/assisted-test-infra/pull/2419 to resolve this

PR #51572, opened 2024-05-02
CentOS 8 containers were end of lifed quite awhile ago replace them

ausil:
As CentOS linux went EOL December 31 2021 https://www.centos.org/centos-linux-eol/ replace the centos:8 references with centos:stream8 as it is still being updated.


LorbusChris:
Stream 8 is going EOL at the end of this month. Can't we migrate these to Stream 9 now?

PR #51582, opened 2024-05-02
ironic-lib: use ci-operator config from repo

elfosardo:
This is a workaround for a known issue where the git metadata are not present in the source image causing the unit and pep8 tests to fail
Updating the root image will fix that


derekhiggins:
could do with a bit of context about why but
/lgtm

PR #51591, opened 2024-05-02
Update cron for e2e-powervs for 4.16 and 4.17 release jobs

dharaneeshvrd:
Running at the same time on 4.15, 4.16 and 4.17 release causes a collision in infra creation and deletion since we are reusing the resources. Want to keep the jobs on different time.


mkumatag (approved):
/lgtm

PR #51595, opened 2024-05-02
NOJIRA: fix: fixing the grep so 4.17 isn't pulled in from CI

jeff-roche:
Fixing the grep so that `.` is not evaluated as a regex


jeff-roche:
This was tested via manual testing to avoid delay getting this fix out

jakobmoellerdev (approved):
/lgtm

PR #51603, opened 2024-05-02
NO-ISSUE: setup mirroring to new openshift-lightspeed quay org

bparees:


bparees:
rehearsal won't pass since the configmap containing the mirror config won't exist until after this merges.  i don't think we really need to worry about rehearsing it though.

PR #51626, opened 2024-05-03
Disable LCA ibu-e2e-flow presubmit on non-code PRs

omertuc:
Similar to the integration job, if only non-code files have changed,
running the `ibu-e2e-flow` job is unnecessary


omertuc:
@donpenney

javipolo (approved):

PR #51662, opened 2024-05-03
remove OpenShiftSDN jobs from 4.16+ in k8s-nmstate

jluhrsen:
4.16 is no longer allowing sdn to be used at install time without overrides. removing these jobs to save on cloud spend/resources.


mkowalski (approved):

qinqon:
Happy with this change

PR #51663, opened 2024-05-03
remove OpenShiftSDN jobs in 4.16+ for CBO

jluhrsen:
sdn is no longer allowed to be installed in 4.16+ without overrides. remove these jobs


zaneb:
Is it worth leaving these as on-demand jobs (i.e. remove `skip_if_only_changed` and add `always_run: false`)? On the theory that there may be people with support exceptions overriding this.
I'm happy either way.

/approve

jluhrsen:
> Is it worth leaving these as on-demand jobs (i.e. remove `skip_if_only_changed` and add `always_run: false`)? On the theory that there may be people with support exceptions overriding this. I'm happy either way.
> 
> /approve

up to you. I can re-do the PR to leave the jobs and just disable them entirely making them available only if you issue a `/test` comment in github. If CBO thinks they may need to deal with this down the road then it makes sense. If it's more of an openshift wide concern I think we still will have these jobs available in other projects.

let me know which way to go :)

jluhrsen:
@zaneb , wanna give this lgtm?

PR #51704, opened 2024-05-06
[WIP] OCPQE-21095 Move 2 long duration jobs to Graviton(arm) platform

JianLi-RH:
https://issues.redhat.com/browse/OCPQE-21095: Graviton is AWS's implementation of the AArch64 64-bit architecture that could save cost between 20%-40% without degrading performance. This is appealing for us to move part of our CI workload from x86 instances to Graviton.

This is a test PR for migrate long duration jobs to ARM (Graviton) AWS instance. 

/cc @jhou1 @jianlinliu

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__arm64-nightly.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__arm64-nightly.yaml
@@ -83,6 +83,35 @@ resources:
       cpu: "1"
       memory: 1Gi
 tests:
+- as: aws-ipi-f9-longduration-apiserver
+  cron: 4 14 9,21,30 * *
+  steps:
+    cluster_profile: aws-qe
+    dependencies:
+      OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE: release:arm64-latest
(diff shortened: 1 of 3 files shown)


ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__arm64-nightly.yaml
```diff
@@ -83,6 +83,31 @@ resources:
...
+    cluster_profile: aws-qe
+    env:
+      BASE_DOMAIN: qe.devcluster.openshift.com
+      OCP_ARCH: arm64
+      TEST_SCENARIOS: API_Server
+      TEST_TIMEOUT: "600"
+    test:
+    - ref: openshift-extended-test-longduration
+    workflow: cucushift-installer-rehearse-aws-ipi
+- as: aws-ipi-f9-longduration-hive-remaining
```
jianlinliu, line 98: If I remember right, hive test cases can not executed in a arm cluster, cc @jianping-shu to confirm.

jianping-shu: You are right! Hive operator can be installed on amd cluster only so far.

PR #51705, opened 2024-05-06
chore: update jobs for PowerMon Operator

vprashar2929:
This commit updates the following in PowerMon Operator jobs:
* Increase the timeout from 5m to 15m while installing operator
* Increase the requested resources for PowerMon jobs
* suppress the output of must gather info while running e2e for Operator and Kepler


IshwarKanse (approved):
/lgtm

PR #51729, opened 2024-05-06
CAPI altinfra periodics for OpenStack & Nutanix

patrickdillon:


patrickdillon:
/close
in favor of testing in tech preview periodics

PR #51733, opened 2024-05-06
trt-1478: add ovn based jobs for aws-cgroupsv2 and gcp-serial

neisw:
Updating coverage to remove / migrate sdn jobs.


neisw:
Thanks @jluhrsen, looks like I missed an sdn->ovn in my copy pasting...  make registry-metadata fixed it right up!

PR #51744, opened 2024-05-07
always create install junit even no kubeconfig is generated

jinyunma:
when installer or previous steps failed before creating ignition-configs/cluster, there is no kubeconfig file generated, the step `gather-must-gather` exit directly.
In such case, install junit file should be generated based on `${SHARED_DIR}/install-status.txt`, so that those failed jobs can be shown in sippy/component readiness.


jinyunma:
cc @jianlinliu to review

jianlinliu:
Good to see the enhancement for sippy/component readiness.

/lgtm

jinyunma:
@vrutkovs  could you help to review this PR when you have time? thanks.

PR #51745, opened 2024-05-07
Update rhdh nightly job : add Slack Notifications for Nightly Job

subhashkhileri:
Implement Slack Notifications for Nightly Job Completion with Status and Direct Links to Artifacts, Logs and Report
https://issues.redhat.com/browse/RHIDP-2245

### DIFF
--- a/ci-operator/step-registry/janus-idp/backstage-showcase/janus-idp-backstage-showcase-commands.sh
+++ b/ci-operator/step-registry/janus-idp/backstage-showcase/janus-idp-backstage-showcase-commands.sh
@@ -4,12 +4,11 @@ HOME=/tmp
 WORKSPACE=$(pwd)
 cd /tmp || exit
 
-export GIT_PR_NUMBER GITHUB_ORG_NAME GITHUB_REPOSITORY_NAME NAME_SPACE TAG_NAME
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/janus-idp/backstage-showcase/janus-idp-backstage-showcase-commands.sh
```diff
@@ -9,7 +9,7 @@ GIT_PR_NUMBER=$(echo "${JOB_SPEC}" | jq -r '.refs.pulls[0].number')
 echo "GIT_PR_NUMBER : $GIT_PR_NUMBER"
 GITHUB_ORG_NAME="janus-idp"
 GITHUB_REPOSITORY_NAME="backstage-showcase"
-NAME_SPACE=showcase-ci
+NAME_SPACE=showcase
```
josephca, line 12: @subhashkhileri do we need to set this variable? it would be overridden from 'openshift-ci-tests.sh' call anyways.

subhashkhileri: true it would be overridden. it was there so I just kept it as it for reference and just matched the value `showcase`

subhashkhileri: I will remove it.

PR #51756, opened 2024-05-07
NO-JIRA: Revert fix: use upi-installer image for sno config update

droslean:
Reverts openshift/release#51544


eggfoobar:
/retitle NO-JIRA: Revert fix: use upi-installer image for sno config update

I'll reopen a fix to account for the SNO failures in 4.16 due to the rhel9 changes.

PR #51762, opened 2024-05-07
Only notify on failed compliance periodic jobs

rhmdnd:
We have sippy queries to view positive patterns on jobs, but notify only
on failed jobs reduces the overall amount of traffic posted to the
channel, and still provides incentive to look at failures as soon as
possible.


Vincent056 (approved):
/lgtm

PR #51763, opened 2024-05-07
Run compliance periodics twice a week

rhmdnd:
Now that we have sippy queries for tracking compliance jobs, let's
run the jobs twice a week which gives us more information and insight in
failures and patterns.


Vincent056 (approved):
/lgtm

Vincent056:
it looks like we need to change in https://github.com/openshift/release/blob/master/ci-operator/config/ComplianceAsCode/content/ComplianceAsCode-content-master__4.13.yaml and for other versions too

rhmdnd:
These are just cron job changes, so I don't anticipate any testing/job issues, but running a few rehearsals anyway.

Vincent056 (approved):
/lgtm

Vincent056 (approved):
/lgtm

PR #51771, opened 2024-05-07
move ovn config helper functions to util lib

jluhrsen:
planning to use this for work on [SDN-4776](https://issues.redhat.com//browse/SDN-4776) [0] which covers more day-2 configs


ci-operator/step-registry/ovn/utils/ovn-utils-ref.yaml
```diff
@@ -0,0 +1,15 @@
+ref:
+  as: ovn-utils
+  from_image:
+    name: cli-yq
```
danwinship, line 4: not sure it matters, but it could use a smaller/simpler from_image, since it just needs `bash` and `cat`

ci-operator/step-registry/ovn/utils/ovn-utils-ref.yaml
```diff
@@ -0,0 +1,15 @@
...
+    tag: latest
+  commands: ovn-utils-commands.sh
+  resources:
+    requests:
+      cpu: 10m
+      memory: 100Mi
+  documentation: |-
+    This step will test that new nodes cannot be added because the existing clusterNetwork CIDR does
+    not provide enough IP space. Then it will configure the CIDR such that more IP space is added and
+    verify the new nodes are added.
```
danwinship, line 15: this was copied from somewhere...

ci-operator/step-registry/ovn/utils/ovn-utils-ref.yaml
jluhrsen: I don't need `yq` at the moment, but in my latest update to this PR I'm using `oc` so I think I need at least cli. That's what I have it at now.

ci-operator/step-registry/ovn/utils/ovn-utils-ref.yaml
jluhrsen: fixed.

jluhrsen:
@danwinship , thank you for reviewing this PR :)

ci-operator/step-registry/ovn/utils/ovn-utils-ref.yaml
jluhrsen: actually, just `cli` didn't work. going back to `cli-yq`

wking (approved):
As long as the content is [small enough](https://kubernetes.io/docs/concepts/configuration/secret/#restriction-data-size) to fit into the Secret backing `SHARED_DIR`, I guess this works.  And reverts or other pivots should be fairly straightforward if we do run into trouble.

/lgtm

PR #51818, opened 2024-05-08
Introduce conditional for cerberus

jtaleric:
Allow us to skip cerberus for some jobs.


paigerube14 (approved):
/lgtm

paigerube14:
@jtaleric I think we still need to set the environment variable in the main Cerberus ref, if we call Cerberus from the cluster density v2 path it won't currently have that variable and will fail

PR #51852, opened 2024-05-09
NO_JIRA: Moved a working vlan into multi vcenter pool.

vr4manta:
NO_JIRA

### Changes
- Moved a lease from existing vsphere lease pool into multi vcenter pool due to vlan 956 having issues.


danilo-gemoli:
renaming, I hope boskos will handle it well...
/lgtm

PR #51940, opened 2024-05-13
CNF-10016: Mark ran latency performance as informing jobs

fontivan:
[CNF-10016](https://issues.redhat.com//browse/CNF-10016): Mark ran latency performance as informing jobs for 4.14, 4.15, and 4.16

- Cyclictest and Oslat latency is considered for a PASS/FAIL grade based on internal criteria
- Also updated test job names so that they did not contain duplicate information and were more helpful in general

### DIFF
--- a/hack/validate-release-jobs.py
+++ b/hack/validate-release-jobs.py
@@ -19,7 +19,8 @@ job_definitions_paths = [
     'ci-operator/jobs/openshift/microshift',
     'ci-operator/jobs/openshift/cluster-control-plane-machine-set-operator',
     'ci-operator/jobs/openshift/osde2e',
-    'ci-operator/jobs/openshift-qe/ocp-qe-perfscale-ci'
+    'ci-operator/jobs/openshift-qe/ocp-qe-perfscale-ci',
+    'ci-operator/jobs/ran-telco5g/kpi-results'
 ]
 
 


bradmwilliams:
@fontivan From the root of your `openshift/release` repo, you'll need to run a `make release-controllers`.  That should clear the failing test.

fontivan:
@bradmwilliams I did that, but I think its upset with me because I renamed the jobs I had created to remove the redundant information. I think to make it happy I'll need to do this in 2 PRs then, one to rename and one to add informing?

deepsm007:
@fontivan Execute `make update` which will update all the jobs and then run `make release-controllers`

bradmwilliams:
You're adding jobs to a location that our automation knows nothing about.  You'll need to add your path here:
https://github.com/openshift/release/blob/79909095eb1b55c5ab62c9429ce18fa8f28509a4/hack/validate-release-jobs.py#L15-L23

bradmwilliams:
You can reproduce/test the failure by running:
`$ ./hack/validate-release-jobs.py -r .`

fontivan:
@bradmwilliams  @deepsm007  is there any way to configure rehearsals for performance metric based jobs? E.g. I'd like to be able to properly run rehearsals, but I know it will fail since its a performance based test and the results were already below threshold over the weekend

bradmwilliams:
> @bradmwilliams @deepsm007 is there any way to configure rehearsals for performance metric based jobs? E.g. I'd like to be able to properly run rehearsals, but I know it will fail since its a performance based test and the results were already below threshold over the weekend

I'm pretty sure I cannot answer that question.  I'm not even sure what a `performance metric based job` would be.
AFAIK, a rehearsal is just a job that runs the specified workflow (`telco5g-kpi-results`).  If it returns a 0 then the job is successful.  If not, then it's a failure.  Any special handling of those failures would have to be written into the workflow.

PR #51946, opened 2024-05-13
ROX-11558: Build acs-fleetshard-operator for E2E tests

kovayur:
In order to test the dp-terraform helm chart and acs-fleetshard operator we need to build the image on Openshift CI


GrimmiMeloni (commented):
LGTM

PR #51948, opened 2024-05-13
Temporarliy disable the hypershift addon e2e test

rokej:
The current hypershift addon e2e test uses cluster pool from OpenShift CI. Before re-enabling the e2e test, we need to:

- create a new cluster pool for the stolostron hypershift addon e2e test
- update the e2e test to run more functional tests


mikeshng (approved):
/approve

/lgtm

PR #51952, opened 2024-05-13
Update installer shellcheck job trigger

rwsu:
Changed the trigger to

run_if_changed: \.sh$

so that all shell script changes, even in the docs directory, will be checked. Previously scripts under docs were being skipped.


barbacbd (commented):
/lgtm

PR #51810, opened 2024-05-08
Add observability cluster pool configuration

cahartma:
Following hive documentation to create our first cluster pool for the new observability AWS account.    Will be adding more configurations once I can verify this is working:
*  https://docs.ci.openshift.org/docs/how-tos/cluster-claim/
*  https://issues.redhat.com/browse/LOG-5375

PR #51969, opened 2024-05-13
Migrating openshift-observability to correct dir under hosted-mgmt

deepsm007:
Renaming the dir for -> https://github.com/openshift/release/pull/51810


2uasimojo:
I didn't vet the contents of these files, but the path is now correct. We can iterate on the specs if necessary once this lands.

/lgtm

PR #51982, opened 2024-05-14
Reduce managed cluster claim for ACM MCO

coleenquadros:
Reducing count on managed cluster for MCO as it is exhausting the Claim and not running the e2e tests to completion


philipgough (approved):
/lgtm

PR #51985, opened 2024-05-14
prometheus-alertmanager: test target requires npm

jan--f:
Install in the same way like we do for prometheus.


jan--f:
/close
duplicate of #51237

PR #51994, opened 2024-05-14
prometheus: run go mod vendor for vendor check

jan--f:
The make target `make unused` has dropped the vendoring bits, relying fully on go modules.


machine424:
/lgtm
thanks!

PR #52004, opened 2024-05-14
Updating krkn to prow scripts repo

paigerube14:
We have moved to using prow-scripts repo, need to be able to rerun jobs


paigerube14:
CC: @chaitanyaenr

PR #52044, opened 2024-05-15
ComplianceAsCode: Add BSI profiles to e2e

yuumasato:
Upstream is working on Bundesamt für Sicherheit in der Informationstechnik (BSI) security compliance profiles for OCP4.
Let's enable testing on those profiles so that we keep watch on any issue with the profile.


yuumasato:
CC @rhmdnd @Vincent056

rhmdnd (approved):
/lgtm

Are we going to add periodic testing for these profiles at some point?

yuumasato:
> Are we going to add periodic testing for these profiles at some point?

@rhmdnd Yeah, we can add periodic testing when the profile gets to a more mature state.

PR #52052, opened 2024-05-15
Add Go 1.22 builder

tpantelis:


skitt:
@tpantelis could you split out the `image-builder` part into a separate PR? It needs to go in first, then https://github.com/stolostron/image-builder/pull/138, and only after that will we be able to update the addon configuration.

tpantelis:
> @tpantelis could you split out the `image-builder` part into a separate PR? It needs to go in first, then [stolostron/image-builder#138](https://github.com/stolostron/image-builder/pull/138), and only after that will we be able to update the addon configuration.

yeah I was wondering about that. I thought I'd first try to combine them 😄

gurnben (approved):
/lgtm
/approve

PR #52055, opened 2024-05-15
NO-ISSUE: Disable SNO jobs for ocm-2.10 branch

danmanor:


ci-operator/jobs/openshift/assisted-service/openshift-assisted-service-release-ocm-2.10-presubmits.yaml
```diff
@@ -1015,7 +1015,6 @@ presubmits:
       job-release: "4.15"
       pj-rehearse.openshift.io/can-be-rehearsed: "true"
     name: pull-ci-openshift-assisted-service-release-ocm-2.10-e2e-ai-operator-ztp-sno-day2-masters
-    optional: true
```
adriengentil, line 1018: these jobs still need to be optional, no?

danmanor: What is the point of having an optional job that is not running unless you trigger it ?

adriengentil: without optional=true, you'll need to trigger the job manually to make the check pass

danmanor: You mean without `optional=true`, if you choose to trigger the job and it failed you will have to override it in order to merge ?

danmanor:
@adriengentil I changed it

adriengentil:
> @adriengentil I changed it
but is this what we want? To trigger it manually to make the PR check pass and merge?

danmanor:
@adriengentil Are you sure this is the case here? I was under the impression that job not triggered automatically will never be required

PR #51537, opened 2024-04-30
NETOBSERV-1593: Add ARM periodics for NetObserv

Amoghrd:
PR has following updates:
- Removed some more SKIPPED profiles for NetObserv
- Added NetObserv periodic jobs for ARM architecture across OCP 4.12-4.16

PR #52056, opened 2024-05-15
NETOBSERV-1593: Move NetObserv jobs to multi

Amoghrd:


Amoghrd:
Followup to https://github.com/openshift/release/pull/51537 since missed moving some jobs to multi

ci-operator/jobs/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16-periodics.yaml
```diff
@@ -21941,18 +21743,7 @@ periodics:
...
-      channel: '#forum-ocp-mco-qe'
-      job_states_to_report:
-      - failure
-      - error
-      - success
-      report_template: '{{if eq .Status.State "success"}} :rainbow: Job *{{.Spec.Job}}*
-        ended with *{{.Status.State}}*. <{{.Status.URL}}|View logs> :rainbow: {{else}}
-        :volcano: Job *{{.Spec.Job}}* ended with *{{.Status.State}}*. <{{.Status.URL}}|View
-        logs> :volcano: {{end}}'
+  name: periodic-ci-openshift-openshift-tests-private-release-4.16-amd64-nightly-gcp-ipi-f9-longduration-mco-g1
```
memodi, line 21746: @Amoghrd mco jobs shouldn't have changed, do you need a rebase?

liangxia: I think that's a job resort, so github think there are changes, but actually there are no changes for mco jobs

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__multi-nightly.yaml
```diff
@@ -2212,6 +2247,15 @@ tests:
     test:
     - chain: openshift-e2e-test-qe
     workflow: baremetal-lab-ipi-virtual-media
+- as: vsphere-ipi-amd-f7-netobserv
```
jianlinliu, line 2250: vsphere does not support multi payload yet, so we should keep it in amd ci yaml file.

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__multi-nightly.yaml
```diff
@@ -57,6 +57,18 @@ resources:
       cpu: "1"
       memory: 1Gi
 tests:
+- as: aws-ipi-amd-f7-netobserv
```
jianlinliu, line 60: I wonder is it on purpose to cover both arm and arm arches in two netobserv jobs? If netobserv jobs is not arch sensitive, we suggest to run the testing on arm arch to save cost. The same comment on gcp and azure jobs.

memodi:
@Amoghrd feel free to add `pj-rehearse ack` label if you're all set with this one.

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__multi-nightly.yaml
jianlinliu: > ARM jobs are run only on AWS once a month

The `f7` in `aws-ipi-arm-f7-netobserv` means once per 7 days, so the job is running weekly, but not monthly, right?

And in theory, there is no much difference between arm and amd jobs, if you want to cover both arches on purpose, one is once per week, another is once per month, I suggest to run the `f7` job on arm, another is named with `f28` running on amd to save cost. About the frequency setting about `f` field, and how to generate an uniq cron, you can refer to https://github.com/openshift/release/blob/efd66d3ae22fd32abed4f3b2c7921196ea4d02ff/ci-operator/config/openshift/openshift-tests-private/README.md#job-frequency

Amoghrd: Done

memodi: @Amoghrd - m6i.2xlarge machine type doesn't exist with ARM arch. Closest available ARM machine type to m6i.2xlarge are m6gd.2xlarge and m7g.2xlarge - were you able to successfully run ARM rehearsal?

@jianlinliu - Based on which ARM machine we go with to match closely for resources we require it's just a difference of $0.02-$0.05/hour (m6gd.2xlarge ~ 0.49, m7g.2xlarge ~ 0.45 vs m6i.2xlarge ~ 0.51 for on-demand instances), I don't think if it would be significant impact on our budget. If it's okay I'd like to keep AMD runs every week and ARM run once per month to catch issues on AMD platform sooner as I presume most of our current customers are on AMD.

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__multi-nightly.yaml
```diff
@@ -850,6 +862,18 @@ tests:
     - ref: ipi-install-heterogeneous
     - chain: openshift-e2e-test-qe-destructive
     workflow: cucushift-installer-rehearse-azure-ipi
+- as: azure-ipi-amd-f7-netobserv
```
jianlinliu, line 865: per the alphabet order, I think the job should be placed prior to `azure-ipi-amd-mixarch-f9-destructive`, right?

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__multi-nightly.yaml
Amoghrd: I can bump up the machines to `m6gd.2xlarge` if that's going to help our tests

Amoghrd: Also reverted trigger updates, i.e ARM runs monthly and AMD weekly

nathan-weinberg: +1 to weekly AMD and monthly ARM

PR #52068, opened 2024-05-15
Updated installer config for dev release-9.9 and backplane-9.9

dislbenn:
Adding and updating release branches for dev `9.9`. This will allow us to begin testing for AKS work.

/cc @cameronmwall @ngraham20


ngraham20 (approved):
/lgtm

PR #52077, opened 2024-05-15
fix(konflux): integration-service image name

MartinBasti:
Reverting name back to redhat-appstudio prefix, as the test suite still expects this image name


dirgim (approved):
/lgtm
/approve

PR #52087, opened 2024-05-16
Drop 4.12 rollback

shellyyang1989:
We decided not to support 4.12 rollback because we won't fix 4.12 rollback bugs.

cc @jiajliu @jianlinliu @JianLi-RH PTAL, thanks!


jianlinliu:
shall we directly remove the yaml file?

shellyyang1989:
> shall we directly remove the yaml file?

I'd like to leave it for a while in case PM will change his mind.

shellyyang1989:
@liangxia could you please help approve? Thanks!

PR #52089, opened 2024-05-16
WIP: upkeep: update sno config to use upi-installer

eggfoobar:


eggfoobar:
/pj-rehearse periodic-ci-rh-ecosystem-edge-nvidia-ci-main-4.14-nvidia-gpu-operator-e2e-23-6-x

eggfoobar:
/pj-rehearse pull-ci-openshift-cluster-authentication-operator-master-e2e-aws-single-node

empovit:
Also ci-operator/config/rh-ecosystem-edge/nvidia-ci/rh-ecosystem-edge-nvidia-ci-main__4.17.yaml.

There are also other jobs that use `single-node-conf-aws`. A quick search brings ~7 workflows and one chain that include this step. Many jobs in turn rely on those. E.g. the `openshift-e2e-aws-single-node` workflow is part of ~300 job definition files :/ I'm not sure how this should be fixed, but adding a upi-image to each file looks like a lot of work. 

Maybe this can be mitigated by adding the image at the workflow/chain level, if that's possible. Then, all one-off jobs need to be addressed. For NVIDIA, for instance, we don't need the entire workflow and therefore include individual steps. I don't know how common it is, though.

eggfoobar:
Yeah, I was curious if this was going to work as intended. Like you pointed out there are a lot of jobs that use this script, looking for ways to keep things simple and maintainable in the long run but cause very little disruption to teams in the process.

empovit:
> Yeah, I was curious if this was going to work as intended. Like you pointed out there are a lot of jobs that use this script, looking for ways to keep things simple and maintainable in the long run but cause very little disruption to teams in the process.

Can you override the installer image just where you need it to be different from the default?
E.g.
```
  installer:
    name: "4.16"
    namespace: ocp
    tag: upi-installer
```
I think this was one of the suggestions, but not sure about the right syntax/section.

PR #52093, opened 2024-05-16
Update recert 4.16 job to use 4.15 image for openshift-tests

eranco74:
This should mitigate:
openshift-tests: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by openshift-tests) Which started since OCP 4.16 binaries started to be built against EL9 libraries rather than EL8.


mresvanis:
/pj-rehearse periodic-ci-rh-ecosystem-edge-recert-main-4.16-baremetalds-sno-recert-cluster-rename

mresvanis:
@eranco74 instead of this change her, [this](https://github.com/openshift/release/pull/52149) seemed to work as expected :+1:

PR #52112, opened 2024-05-16
Upgrade konflux compatibility tasks to 4.14

thepetk:
Upgrades the `registry` tests to point to ocp 4.14. It also removes the `konflux-nightly-devfile-compatibility` as the current component/application creation process will be deprecated and this test will be not needed.

fixes https://github.com/devfile/api/issues/1535


thepetk:
/pj-rehearse periodic-ci-devfile-registry-main-v4.14-konflux-nightly-devfile-compatibility

thepetk:
I've updated the PR to remove the `konflux-nightly-devfile-compatibility` test as the process that the test is following will be deprecated.

maysunfaisal (approved):

PR #52155, opened 2024-05-17
HyperShift: adjust mce install and lvm install

LiangquanLi930:
disable hive to fix memory resource exceeded issue


heliubj18:
disable hive to fix memory resource exceeded issue

PR #52181, opened 2024-05-17
HIVE-2495: Undo fast-forwarding

celebdor:
ci-tools does not currently support the branch naming scheme nor the usage for non openshift release-x.y repositories. Thus, it is better to disable this until things change.


2uasimojo:
Thanks.

/lgtm

Will need DPTP nod IIUC.

PR #52187, opened 2024-05-19
OTA-1069: adding new test point to OCP-23799

evakhoni:
as part of [OTA-1069](https://issues.redhat.com/browse/OTA-1069) effort, [new step](https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitem?id=OCP-23799&view=historyOnce) has been added to OCP-23799
this PR updates the automation to include the aforementioned step

### DIFF
--- a/ci-operator/step-registry/cucushift/upgrade/toversion/cucushift-upgrade-toversion-commands.sh
+++ b/ci-operator/step-registry/cucushift/upgrade/toversion/cucushift-upgrade-toversion-commands.sh
@@ -384,21 +384,6 @@ function check_history() {
     fi
 }
 
-# for OCP-23799 OTA-455 - only minor upgrade to-latest, 4.16+
-function check_rollback_denied() {
-    if [[ "${UPGRADE_TO_VERSION}" == "latest" ]] && [[ ${TARGET_MINOR_VERSION} -ge "16" ]] && [[  ${TARGET_MINOR_VERSION} -gt ${SOURCE_MINOR_VERSION} ]]; then
-        export OC_ENABLE_CMD_UPGRADE_ROLLBACK=true #OCPBUGS-33905, rollback is protected by env feature gate now
-        out="$(oc adm upgrade rollback 2>&1 || true)" # expecting an error, capture and don't fail
-        expected="error: ${SOURCE_VERSION} is less than the current target ${TARGET_VERSION} and matches the cluster's previous version, but rollbacks that change major or minor versions are not recommended."
-        if [[ ${out} != "${expected}" ]]; then
-            echo -e "to-latest rollback reject step failed. \nexpecting: \"${expected}\" \nreceived: \"${out}\""
-            return 1
-        else
-            echo "to-latest rollback reject step passed."
-        fi
-    fi
-}
-
 if [[ -f "${SHARED_DIR}/kubeconfig" ]] ; then
     export KUBECONFIG=${SHARED_DIR}/kubeconfig
 fi
@@ -463,4 +448,3 @@ run_command "${upgrade_cmd}"
 echo "Upgrading cluster to ${TARGET_VERSION} gets started..."
 check_upgrade_status
 check_history
-check_rollback_denied
--- a/ci-operator/config/openshift/verification-tests/openshift-verification-tests-master__ota-multi-stable-4.16-upgrade-from-stable-4.15.yaml
+++ b/ci-operator/config/openshift/verification-tests/openshift-verification-tests-master__ota-multi-stable-4.16-upgrade-from-stable-4.15.yaml
@@ -64,6 +64,7 @@ tests:
     cluster_profile: aws-qe
     env:
       BASE_DOMAIN: qe.devcluster.openshift.com
+      ENABLE_OTA_TEST: OCP-23799
     test:
     - ref: cucushift-upgrade-setchannel
     - ref: cucushift-ota-preupgrade
--- a/ci-operator/step-registry/cucushift/ota/postupgrade/cucushift-ota-postupgrade-commands.sh
+++ b/ci-operator/step-registry/cucushift/ota/postupgrade/cucushift-ota-postupgrade-commands.sh
@@ -248,6 +248,20 @@ function post-OCP-56083(){
     return 1
 }
 
+function post-OCP-23799(){
+        export OC_ENABLE_CMD_UPGRADE_ROLLBACK="true" #OCPBUGS-33905, rollback is protected by env feature gate now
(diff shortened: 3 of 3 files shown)


evakhoni:
/pj-rehearse periodic-ci-openshift-verification-tests-master-ota-multi-stable-4.16-upgrade-from-stable-4.15-aws-ipi-amd-tolatest-f28

evakhoni:
/pj-rehearse periodic-ci-openshift-verification-tests-master-ota-multi-stable-4.16-upgrade-from-stable-4.15-aws-ipi-amd-tolatest-f28

evakhoni:
/cc @jiajliu @JianLi-RH
rehearsal passed.
I have [enabled](https://github.com/openshift/release/pull/52187/commits/088e037443f81e078ca9bc399ef08affcf3b1fd3) blocking on failure for this step
I think this PR is ready for review.

evakhoni:
@jiajliu @JianLi-RH @jianlinliu @shellyyang1989 @dis016 please help review

ci-operator/step-registry/cucushift/upgrade/toversion/cucushift-upgrade-toversion-commands.sh
```diff
@@ -448,3 +462,4 @@ run_command "${upgrade_cmd}"
 echo "Upgrading cluster to ${TARGET_VERSION} gets started..."
 check_upgrade_status
 check_history
+check_rollback_denied
```
jiajliu, line 465: This check sounds not a basic checkpoint for all the upgrade with `--to` option, right? so I would suggest to enable it through `ENABLE_OTA_TEST: OCP-23799` in [job config](https://github.com/openshift/release/blob/master/ci-operator/config/openshift/verification-tests/openshift-verification-tests-master__ota-multi-stable-4.16-upgrade-from-stable-4.15.yaml#L61), which will not interrupt other cases and then we don't need to check if it's minor upgrade.

There might be two options:
1. if holding the check in `cucushift-upgrade-toversion` step, then we need add check_ota_case_enabled func and judge the case enabled or not, sth like https://github.com/openshift/release/blob/master/ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh#L419
2. it sounds like a post check after upgrade, so how about just moving the check into postupgrade step, https://github.com/openshift/release/blob/master/ci-operator/step-registry/cucushift/ota/postupgrade/cucushift-ota-postupgrade-commands.sh, it's more clear and no touch current `cucushift`

evakhoni: thanks Jia!
as for `sounds not a basic checkpoint for all the upgrade with --to option`,   
in this case there is no impact on any upgrade with `--to`  option: 
my `function check_rollback_denied()` have condition 
```
[[ "${UPGRADE_TO_VERSION}" == "latest" ]] && 
[[ ${TARGET_MINOR_VERSION} -ge "16" ]] && 
[[  ${TARGET_MINOR_VERSION} -gt ${SOURCE_MINOR_VERSION} ]]
```
where also `function set_target_and_upgrade_cmd()` have
```
if [[ "${UPGRADE_TO_VERSION}" == "latest" ]]; then
upgrade_cmd="oc adm upgrade --to-latest"
```
so check_rollback_denied runs only on minor upgrade, and only `--to-latest`. this basically already narrows down to a few selected profiles. do you think any more narrowing than that is required?

evakhoni: as for chain: `openshift-upgrade-qe-sanity-rollback` it appears that step: `cucushift-ota-postupgrade` is not part of that chain. do you suggest adding it?   
+1 on having a post-upgrade check separate from the main upgrade step, however is it worth the additional effort? or am I mistaken something and this step is in fact being loaded already?

jiajliu: > in this case there is no impact on any upgrade with --to option

hmm, take another [test case](https://github.com/openshift/release/blob/master/ci-operator/config/openshift/verification-tests/openshift-verification-tests-master__ota-multi-stable-4.16-upgrade-from-stable-4.15.yaml#L83) using `cucushift-upgrade-toversion` for example, this is a 4.15-4.16 minor upgrade and upgrade with `--to-latest`, then it will also call `check_rollback_denied` checkpoint, which is not a required step for case OCP-53907, right? But if the checkpoint failed at  `check_rollback_denied`, the postupgrade for case ocp-53907 will not be executed. That's why we have `ENABLE_OTA_TEST` to help split different test cases.

>  so you suggest relocating my check_rollback_denied() over there? 
yes, that's one of the solutions to move the check into cucushift-ota-postupgrade, then only if OCP-23799 enabled, it will be executed.

> does it have all the variables needed, such as UPGRADE_TO_VERSION TARGET_MINOR_VERSION and SOURCE_MINOR_VERSION available? 
if we setting `ENABLE_OTA_TEST: OCP-23799` in this [job](https://github.com/openshift/release/blob/master/ci-operator/config/openshift/verification-tests/openshift-verification-tests-master__ota-multi-stable-4.16-upgrade-from-stable-4.15.yaml#L61), then no need to judge above three variables. The job config `stable-4.16-upgrade-from-stable-4.15` will ensure it's a minor upgrade. The [default value of `UPGRADE_TO_VERSION` ](https://github.com/openshift/release/blob/master/ci-operator/step-registry/cucushift/upgrade/toversion/cucushift-upgrade-toversion-ref.yaml#L23)is latest, so no need to judge it again in the post upgrade func since we will not setting `UPGRADE_TO_VERSION` to any other value in the job config.

evakhoni:
/pj-rehearse periodic-ci-openshift-verification-tests-master-ota-multi-stable-4.16-upgrade-from-stable-4.15-aws-ipi-amd-tolatest-f28

ci-operator/step-registry/cucushift/upgrade/toversion/cucushift-upgrade-toversion-commands.sh
evakhoni: ok I gave it a go and implemented in `cucushift-ota-postupgrade` b77449506e9422fb89eeca55da402be47e1884fd
@jiajliu ptal

evakhoni:
yep passed indeed 
```
Running command: which oc
/cli/oc
Running command: oc version --client
Client Version: 4.16.0-202405212036.p0.g2574bc7.assembly.stream.el9-2574bc7
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
------> OCP-23799
to-latest rollback reject step passed.
```
/pj-rehearse ack

PR #52190, opened 2024-05-20
CAPI is the default install method on AWS/vSphere/Nutanix now

yunjiang29:
Remove CAPI flags in 4.16 jobs.

https://github.com/openshift/installer/pull/8428


yunjiang29:
@gpei PTAL

yunjiang29:
@liangxia PTAL, thanks

PR #52209, opened 2024-05-20
Updating openshift-docs CI to only use main for build scripts

aireilly:
Updating openshift-docs CI to only use origin/main for build scripts. This update means that we don't have to update scripts across branches going forward, just in main


aireilly:
/pj-rehearse pull-ci-openshift-openshift-docs-pipelines-docs-1.10-validate-asciidoc

PR #52221, opened 2024-05-20
Updating re-trigger as false for CNV due to its over pricing

Sau1506mya:
CNV operator is installed on AWS cluster profile via bare metal machine, they are more pricey then a simple VM even with same "spec" cpu/memory.
So, updating the `RE_TRIGGER_ON_FAILURE` as false: https://github.com/openshift/release/blob/master/ci-operator/config/RedHatQE/interop-testing/RedHatQE-interop-testing-cnv-4.16__cnv-odf-ocp4.16-lp-interop.yaml#L73


calebevans (approved):
/lgtm

rabin-io (approved):
/lgtm

PR #52253, opened 2024-05-20
TRT-487: add ovn to installer vsphere job names

jluhrsen:


jluhrsen:
@zaneb, @sdodson , could you PTAL at this one?
looks like I'll need MCO approval too (CC @cheesesashimi , @dkhater-redhat )

PR #52254, opened 2024-05-20
TRT-487: add ovn to hypershift kubevirt job names

jluhrsen:


muraee:
/approve
cc @davidvossel

jluhrsen:
need `/lgtm` @muraee

PR #52257, opened 2024-05-20
OCPBUGS-28690: rm sdn jobs from MCO 4.17+

jluhrsen:


dkhater-redhat:
ACK that the need to remove those jobs in 4.15 and/or 4.16 in a different commit. Does the secret (e.g. cluster-secrets-equinix-ocp-metal), configMaps, or any similar resources specifically used by these jobs need to be cleaned up as well? not sure if they are exclusive to this or not? probably not?

cheesesashimi:
@dkhater-redhat Those are managed by `ci-operator`, so there's nothing to clean up.

@jluhrsen Overall, this looks reasonable. Just a quick question before I approve this: Do you plan to handle removing the jobs from 4.15 / 4.16 in this PR or would in a separate PR? Either way is fine with us.

jluhrsen:
> @jluhrsen Overall, this looks reasonable. Just a quick question before I approve this: Do you plan to handle removing the jobs from 4.15 / 4.16 in this PR or in a separate PR? Either way is fine with us.

sorry for the delay (was on PTO). I just updated this PR to include 4.15 and 4.16. I also noticed we have a 4.18 file to handle too so took care of that.

PR #52282, opened 2024-05-21
adjust azure stack ci jobs frequency

jianlinliu:
Recently we saw several azure stack ci jobs hit quota limit, while azure stack is a small deployment, do not have capacity to run many concurrent ci jobs, so decrease the test frequency back to f28.


jianlinliu:
cc @liangxia to review.

PR #52285, opened 2024-05-21
Add 4.16 separate job for cnf-feature-deploy repo

sshnaidm:
Change master job to 4.17


nishant-parekh:
@sshnaidm need to add quay mapping for the new (4.17) version here:
https://github.com/openshift/release/blob/master/core-services/image-mirroring/openshift-kni/mapping_openshift-kni_quay

PR #52290, opened 2024-05-21
add a new clusterpool for the observability intelligence DVO

tremes:


ci-operator/config/app-sre/deployment-validation-operator/app-sre-deployment-validation-operator-master.yaml
```diff
@@ -68,7 +68,7 @@ tests:
   cluster_claim:
     architecture: amd64
     cloud: aws
-    owner: openshift-ci
+    owner: obsint-dvo
```
cahartma, line 71: I would also make it a habit of specifying your region here as well, even if you only have one.   I suggest: 
```suggestion
    owner: obsint-dvo
    labels:
      region: us-east-1
```

PR #52300, opened 2024-05-21
step-registry: hypershift: forward interrupts to test process

stevekuznetsov:
It's not clear to me why we have so many different test steps that all do the same thing under the hood, but we can clean that up in the future.


sjenning:
/pj-rehearse pull-ci-openshift-hypershift-main-e2e-conformance

PR #52303, opened 2024-05-21
vault: deploy v1.16.2

danilo-gemoli:
Follow up of https://github.com/openshift/release/pull/52294.

GCS bucket `vault-ci-openshift-clone` is a clone of `vault-ci-openshift`, we are going to use that and see what happens before switching back to the production one.


danilo-gemoli:
Thinking about whether using production bucket instead of the `-clone`:
```diff
-            bucket        = "vault-ci-openshift"
+            bucket        = "vault-ci-openshift-clone"
```

PR #52304, opened 2024-05-21
Remove stolostron hypershift addon e2e

rokej:
Temporarily remove e2e test from stolostron hypershift addon to stop using openshift-ci cluster pool.


mikeshng (approved):
/approve

/lgtm

PR #52327, opened 2024-05-21
OCPBUGS-28678: rm sdn jobs from baremetal-runtimecfg in 4.17+

jluhrsen:

### DIFF
--- a/ci-operator/config/openshift/baremetal-runtimecfg/openshift-baremetal-runtimecfg-master.yaml
+++ b/ci-operator/config/openshift/baremetal-runtimecfg/openshift-baremetal-runtimecfg-master.yaml
@@ -62,12 +62,12 @@ tests:
   commands: make test
   container:
     from: src
-- as: e2e-metal-ipi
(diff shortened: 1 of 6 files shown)


jluhrsen:
@mkolesnik could you please look at this one?

ci-operator/config/openshift/baremetal-runtimecfg/openshift-baremetal-runtimecfg-master.yaml
```diff
@@ -62,12 +62,6 @@ tests:
   commands: make test
   container:
     from: src
-- as: e2e-metal-ipi
```
cybertron, line 65: I don't think we want to remove these jobs completely. They provide our ipv4 coverage in addition to SDN. I believe we can just switch them to the baremetalds-e2e-ovn-ipv4 workflow.

jluhrsen: that sounds right. I've updated the PR for this. thanks for the comment.

PR #52328, opened 2024-05-21
OCPBUGS-28688: rm sdn jobs from openshift/kubernetes in 4.17+

jluhrsen:


jluhrsen:
@deads2k @soltysh @mfojtik @tkashem can one of you please look at this?

PR #52216, opened 2024-05-20
add telco cnf images to image mirroring

SchSeba:
now that all the metallb sriov and ptp repos are private we can enable the image mirroring again

ticket: [DPTP-4036](https://issues.redhat.com//browse/DPTP-4036)

PR #52340, opened 2024-05-22
add sriov-cni okd mirroring

SchSeba:


sshnaidm:
thanks @SchSeba

core-services/image-mirroring/openshift/_config.yaml
```diff
@@ -40,9 +40,25 @@ images:
   - "4.17"
   - "4.17.0"
   - latest
+  "sriov-4.17":
```
zeeke, line 43: why do we need these `sriov` specific tags?

SchSeba: because we can't have this as part of the 4.17 ImageStream as it's core OCP payload

SchSeba:
/close
https://github.com/openshift/release/pull/52216

PR #52345, opened 2024-05-22
Remove 'day1-post,day2' from the default label filters of the presubmit jobs

yasun1:
[OCM-8223](https://issues.redhat.com//browse/OCM-8223)

### DIFF
--- a/ci-operator/config/openshift/rosa/openshift-rosa-master__e2e-presubmits.yaml
+++ b/ci-operator/config/openshift/rosa/openshift-rosa-master__e2e-presubmits.yaml
@@ -44,7 +44,7 @@ tests:
   steps:
     cluster_profile: aws-sd-qe
     env:
-      TEST_LABEL_FILTERS: (day1-post,day2)&&!NonHCPCluster&&!Exclude
+      TEST_LABEL_FILTERS: '!NonHCPCluster&&!Exclude'
     test:
     - ref: rosa-test-e2e-pull-request
     workflow: rosa-aws-sts-hcp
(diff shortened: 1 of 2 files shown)


yasun1:
/pj-rehearse pull-ci-openshift-rosa-master-e2e-presubmits-pr-rosa-hcp

yasun1:
@ciaranRoche  Could you help review the pull request? Thanks.

radtriste:
@yasun1 I am not sure about that change. Other labels will be tested in Nightlies. We don't need to test everything in PR check.
@xueli181114 wdyt ?

yasun1:
@radtriste We do not test anything. After the change, the presubmit job will test:
1. if `/rosa/tests/ci/data/commit-focus`(generated when build PR image) is not empty, the cases list for each configuration is like `--ginkgo.focus ${FOCUS} --ginkgo.label-filter '!NonHCPCluster&&!Exclude'`.
2.  if `/rosa/tests/ci/data/commit-focus` is empty, which means there are no case IDs recorded, the cases list for each configuration is like `--ginkgo.label-filter 'Critical&&!NonHCPCluster&&!Exclude'`.

In the past, we defined the label filter for each job as `(day1-post,day2)&&!NonHCPCluster&&!Exclude`, which means that even if the case IDs list is not empty, if the case does not belong to `(day1-post,day2)`, it will not be executed. The `(day1-post,day2)&&!NonHCPCluster&&!Exclude` works at beginning. With the team members start automating other runtime features, we need to do this update

yasun1:
@ciaranRoche Could you also help review the PR? Thanks.

radtriste (approved):
/lgtm

PR #52454, opened 2024-05-27
Update golangci-lint to correctly verify go version in go.mod

dinhxuanvu:
Go version in go.mod can be formatted as major.minor.patch starting go 1.22


soltysh (approved):
/lgtm
/approve

PR #52468, opened 2024-05-27
Fix Dockerfile path in ci-operator config, ironic-image master okd-scos

MahnoorAsghar:
Dockerfile.scos should be referenced instead of Dockerfile.ocp.


MahnoorAsghar:
/pj-rehearse pull-ci-openshift-ironic-image-master-okd-scos-images

PR #52495, opened 2024-05-28
remove useless PLATFORM_VERSION evn from ci job configs

jianlinliu:
`PLATFORM_VERSION` is not needed, because its default value in upgrade steps is `8.6`, and actually scripts referencing the ENV only care about the major version, so it should be okay to ensure the major version is `8` for the default value.


jianlinliu:
cc @gpei to review.

jianlinliu:
cc @liangxia to review.

PR #52504, opened 2024-05-28
fix: use correct ocp version in image_stream_tag for powermon

vprashar2929:
This PR fixes the issue where wrong OCP version was used in image stream tags for golang-1.21


IshwarKanse (approved):
/lgtm

PR #52518, opened 2024-05-28
openstack: Remove broken altinfra manifest-test Installer presubmit

pierreprinetti:
This test is running by default on all OpenStack pull requests in Installer, and it's not bringing any useful signal in its current form.

/cc patrickdillon

We might as well remove all OpenStack altinfra periodics, since we run CAPI in the default featureSet?


pierreprinetti:
/cc patrickdillon 

We might as well remove all OpenStack altinfra periodics, since we run CAPI in the default featureSet?

mandre (commented):
/lgtm

PR #52527, opened 2024-05-28
NOJIRA: Update cluster-etcd-operator e2e tests to use TechPreviewNoUp…

dusk125:
…grade

This should fix the issues with the installer on AWS

### DIFF
--- a/ci-operator/config/openshift/cluster-etcd-operator/openshift-cluster-etcd-operator-master.yaml
+++ b/ci-operator/config/openshift/cluster-etcd-operator/openshift-cluster-etcd-operator-master.yaml
@@ -120,8 +120,7 @@ tests:
   steps:
     cluster_profile: aws
     env:
-      FEATURE_GATES: '[EtcdBackendQuota=true,AutomatedEtcdBackup=true]'
-      FEATURE_SET: CustomNoUpgrade
+      FEATURE_SET: TechPreviewNoUpgrade
     test:
     - as: test
       cli: latest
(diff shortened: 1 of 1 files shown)


hasbro17:
/lgtm

Although do we also need to update for 4.16 (and prior)?
https://github.com/openshift/release/blob/1b79448ab755d34655bb5e65ec4f2fca4cea2fd9/ci-operator/config/openshift/cluster-etcd-operator/openshift-cluster-etcd-operator-release-4.16.yaml#L124-L141

I can't tell if our installs for e2e-operator are broken there too.

ci-operator/config/openshift/cluster-etcd-operator/openshift-cluster-etcd-operator-master.yaml
```diff
@@ -138,7 +138,7 @@ tests:
     cluster_profile: aws
     env:
       FEATURE_GATES: '[EtcdBackendQuota=true,AutomatedEtcdBackup=true]'
```
tjungblu, line 140:
```suggestion
      - FEATURE_GATES: '[EtcdBackendQuota=true,AutomatedEtcdBackup=true]'
```

ci-operator/config/openshift/cluster-etcd-operator/openshift-cluster-etcd-operator-master.yaml
```diff
@@ -121,7 +121,7 @@ tests:
...
```
tjungblu, line 123:
```suggestion
```

PR #52528, opened 2024-05-28
Output curl to file

fxiang1:
https://issues.redhat.com/browse/ACM-10683

Curl is writing to stdout, need to write to file instead.

- Output curl to file

### DIFF
--- a/ci-operator/config/stolostron/cluster-curator-controller/stolostron-cluster-curator-controller-main.yaml
+++ b/ci-operator/config/stolostron/cluster-curator-controller/stolostron-cluster-curator-controller-main.yaml
@@ -37,7 +37,7 @@ tests:
     export os=$(go env GOOS)
     export arch=$(go env GOARCH)
     export KB_TOOLS_ARCHIVE_NAME="kubebuilder-tools-${KUBEBUILDER_VERSION}-${os}-${arch}.tar.gz"
-    curl -s -f -L https://storage.googleapis.com/kubebuilder-tools/${KB_TOOLS_ARCHIVE_NAME}
+    curl -s -f -L https://storage.googleapis.com/kubebuilder-tools/${KB_TOOLS_ARCHIVE_NAME} -o ${KB_TOOLS_ARCHIVE_NAME}
     tar -zxvf ${KB_TOOLS_ARCHIVE_NAME}
     sudo mv -f ./kubebuilder /usr/local/kubebuilder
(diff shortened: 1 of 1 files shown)


mikeshng (approved):
/approve

/lgtm

ci-operator/config/stolostron/cluster-curator-controller/stolostron-cluster-curator-controller-main.yaml
```diff
@@ -37,7 +37,7 @@ tests:
     export os=$(go env GOOS)
     export arch=$(go env GOARCH)
     export KB_TOOLS_ARCHIVE_NAME="kubebuilder-tools-${KUBEBUILDER_VERSION}-${os}-${arch}.tar.gz"
-    curl -s -f -L https://storage.googleapis.com/kubebuilder-tools/${KB_TOOLS_ARCHIVE_NAME}
+    curl -s -f -L https://storage.googleapis.com/kubebuilder-tools/${KB_TOOLS_ARCHIVE_NAME} -o ${KB_TOOLS_ARCHIVE_NAME}
     tar -zxvf ${KB_TOOLS_ARCHIVE_NAME}
     sudo mv -f ./kubebuilder /usr/local/kubebuilder
```
mikeshng, line 42: this sudo usually prompt for a password. your best bet is just to use mv without sudo if this doesn't work

mikeshng (approved):

ci-operator/config/stolostron/cluster-curator-controller/stolostron-cluster-curator-controller-main.yaml
fxiang1: Good point Mike. The other issue could be a read-only file system in which case I'm tempted to use kubebuilder from /tmp. Thanks!

xiangjingli: @fxiang1  may try to set the PATH to ./kubebuilder before running sonar cloud and reset PATH to original after sonar cloud scan.

fxiang1: Thanks @xiangjingli I'll try to set the PATH and also KUBEBUILDER_ASSETS which overwrites the default kubebuilder path.

PR #52149, opened 2024-05-17
Fix openshift-tests glibc errors for recert 4.16 baremetal jobs

mresvanis:
This PR mitigates the following errors in recert baremetal jobs on OCP 4.16:

```
openshift-tests: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by openshift-tests)
``` 

This error has showed up since OCP 4.16 binaries started to be built against RHEL9 libraries (rather than RHEL8) . 
> This is a temporary solution until `assisted-test-infra` supports RHEL9/rocky 9 - https://github.com/openshift/assisted-test-infra/pull/2419.

PR #52545, opened 2024-05-29
Revert "Fix openshift-tests glibc errors for recert 4.16 baremetal jobs"

eranco74:
Reverts openshift/release#52149


mresvanis:
@eranco74  IIUC I think we need to also set the `PACKET_OS` env var of the SNO packet setup CI step, as it [defaults](https://github.com/openshift/release/blob/master/ci-operator/step-registry/baremetalds/packet/setup/baremetalds-packet-setup-ref.yaml#L25-L29) to `rocky_8`.

We could the following to both recert manifests for 4.16 for the `baremetalds-sno-recert-cluster-rename` test definition:

```yaml
env:
    PACKET_OS: rocky_9
```

mresvanis:
/pj-rehearse pull-ci-rh-ecosystem-edge-recert-main-baremetalds-sno-recert-cluster-rename

PR #52565, opened 2024-05-29
ComplianceAsCode: Add PCI-DSS v4 profiles to e2e

yuumasato:
Upstream is working on PCI-DSS v4 security compliance profiles for OCP4.
Let's enable testing on those profiles so that we keep watch on any issue with the profile.


yuumasato:
CC: @rhmdnd @Vincent056

PR #52567, opened 2024-05-29
Add openshift-abi-qe Rover group to QCI RBAC

bmanzari:
As per documentation at https://docs.ci.openshift.org/docs/how-tos/use-registries-in-build-farm/#how-do-i-gain-access-to-qci

Adding OCP Agent Based Installer QE Rover Group to QCI RBAC

https://rover.redhat.com/groups/group/openshift-abi-qe


danilo-gemoli:
/lgtm
again

PR #52579, opened 2024-05-29
Add fips-check-image-scan to steps registry

swghosh:
**Usage**: Allow scanning of built images by specifying `${SCAN_IMAGE}` to use [check-payload](https://github.com/openshift/check-payload) tool to verify binaries for FIPS compliance. 

To use this step in your CI worflow or chain, ensure to have the following base image:

```yaml
base_images:
  check-payload:
    name: check-payload
    namespace: ci
    tag: latest
...
```
because the step uses `from: check-payload` image.

### DIFF
--- a/ci-operator/step-registry/fips-check/OWNERS
+++ b/ci-operator/step-registry/fips-check/OWNERS
@@ -1,2 +1,7 @@
 approvers:
+- ashwindasr
+- kolyshkin
+- mrunalp
+- rhmdnd
+- rphillips
 - sallyom
--- a/ci-operator/step-registry/fips-check/fips-check-ref.metadata.json
+++ b/ci-operator/step-registry/fips-check/fips-check-ref.metadata.json
@@ -2,6 +2,11 @@
 	"path": "fips-check/fips-check-ref.yaml",
 	"owners": {
 		"approvers": [
+			"ashwindasr",
+			"kolyshkin",
+			"mrunalp",
+			"rhmdnd",
+			"rphillips",
 			"sallyom"
 		]
 	}


ci-operator/step-registry/fips-check/image-scan/fips-check-image-scan-ref.metadata.json
```diff
@@ -0,0 +1,10 @@
+{
+	"path": "fips-check/image-scan/fips-check-image-scan-ref.yaml",
+	"owners": {
+		"approvers": [
+			"mrunalp",
+			"swghosh",
+			"xingxingxia"
```
mrunalp, line 7: Please add @rphillips here.

rphillips: I believe this is autogenerated from the OWNERS file. It probably needs to be added there, then regenerated.

swghosh: Added.

ci-operator/step-registry/fips-check/OWNERS
```diff
@@ -1,2 +1,7 @@
 approvers:
+- ashwindasr
+- kolyshkin
+- mrunalp
+- rhmdnd
+- rphillips
```
sallyom, line 6: I'm fine if you remove me from this list, it's been awhile since I had any interactions with openshift ci, it's probably best. Or, I can do this as a follow-up.

swghosh: I missed your comment in time, before ti's got merged.
We already have the new owners added who can now approve it,
so guess it's still fine to have yours around :) feel free to follow-up a removal in case otherwise, thanks!

PR #52605, opened 2024-05-30
Keep Node Js version to v16.13.1

coleenquadros:
Keep the node js version in sync with https://github.com/openshift/release/blob/4954423e496c35b31782861027709668d6c07623/ci-operator/config/openshift/prometheus/openshift-prometheus-master.yaml#L52


saswatamcode (approved):
/lgtm

PR #52614, opened 2024-05-30
NOJIRA: pinning ibi

jeff-roche:


javipolo:
IMHO Let's better hold https://github.com/rh-ecosystem-edge/ib-orchestrate-vm/pull/74 until we can verify that it wont break CI, so pinning ib-orchestrate-vm is not needed :)

PR #52713, opened 2024-06-03
MGMT-17960: Get Assisted-Installer team authorization to pull from QCI

danmanor:
see the issue for a detailed explanation


hongkailiu (approved):

danmanor:
@hongkailiu I think we need /approve

PR #52716, opened 2024-06-03
use squash merging for konflux-workspaces

sadlerap:


sadlerap:
@filariow I need your approval to merge this.

PR #52766, opened 2024-06-04
stackrox: never skip nongroovy tests

janisz:
In order to make them required on Github level as we do not use prow merge features we need to make it mandatory for all PRs. This PR remove condition to not run nongroovy on ui only changes.

### DIFF
--- a/ci-operator/jobs/stackrox/stackrox/stackrox-stackrox-master-presubmits.yaml
+++ b/ci-operator/jobs/stackrox/stackrox/stackrox-stackrox-master-presubmits.yaml
@@ -193,7 +193,7 @@ presubmits:
           secretName: result-aggregator
     trigger: (?m)^/test( | .* )eks-qa-e2e-tests,?($|\s.*)
   - agent: kubernetes
-    always_run: false
+    always_run: true
     branches:
     - ^master$
     - ^master-


ci-operator/config/stackrox/stackrox/stackrox-stackrox-master.yaml
```diff
@@ -75,9 +75,9 @@ tests:
       COLLECTION_METHOD: core_bpf
     workflow: stackrox-stackrox-e2e-job
   timeout: 5h0m0s
-- as: gke-nongroovy-e2e-tests
+- always_run: false
+  as: gke-nongroovy-e2e-tests
```
gavin-stackrox, line 79: in order to make it always run:
```suggestion
- as: gke-nongroovy-e2e-tests
```

gavin-stackrox (changes requested):

ci-operator/config/stackrox/stackrox/stackrox-stackrox-master.yaml
janisz: @gavin-stackrox It results with `ERROR: One of always_run, run_if_changed or skip_if_only_changed should be set in config`
Should always run be true?

ci-operator/config/stackrox/stackrox/stackrox-stackrox-master.yaml
```diff
@@ -77,7 +77,6 @@ tests:
   timeout: 5h0m0s
 - as: gke-nongroovy-e2e-tests
```
janisz, line 78:
```suggestion
- always_run: true
  as: gke-nongroovy-e2e-tests
```

ci-operator/config/stackrox/stackrox/stackrox-stackrox-master.yaml
gavin-stackrox: Yeah, you are right.

PR #52774, opened 2024-06-04
Move 4 nodes from oko-op-05(OCP4.14) to oko-op-06(OCP4.15)

lewisdenny:


viroel:
/hold

so you can release when ready.

PR #52775, opened 2024-06-04
OCPEDGE-1070: fix: updating main to use 4.17 images and targets

jeff-roche:


jakobmoellerdev (changes requested):
I think this is missing replicating the build configuration to have a set of pipelines for the 4.16 branch

jeff-roche:
> I think this is missing replicating the build configuration to have a set of pipelines for the 4.16 branch

I'm tackling that on a separate PR

ci-operator/config/openshift/lvm-operator/openshift-lvm-operator-main.yaml
```diff
@@ -16,7 +16,7 @@ build_root:
   image_stream_tag:
     name: builder
     namespace: ocp
-    tag: rhel-9-golang-1.21-openshift-4.16
+    tag: rhel-9-golang-1.21-openshift-4.17
```
jakobmoellerdev, line 19: I believe 4.17 will run with go 1.22, so we should probably use rhel-9-golang-1.22-openshift-4.17

jeff-roche: I stuck with 1.21 since that's what is in the gomod https://github.com/openshift/lvm-operator/blob/main/go.mod#L3

jakobmoellerdev: thats only the minimum compiled version for language features, go1.22 also uses the toolchain. We cannot bump to 1.22 before the image_stream here changes, thats why I think we should go here first. Then we can switch in the repo when were ready (probably immediately)

PR #52781, opened 2024-06-04
Add version_upper and version_lower to pool config

cahartma:
Update to doc to add `version_upper` and `version_lower` so that our imageRef will be updated automatically.   @deepsm007  will be updating all cluster pool configs to include the new params in:  
https://github.com/openshift/release/tree/master/clusters/hosted-mgmt/hive/pools/openshift-observability


deepsm007:
https://github.com/openshift/release/pull/52780

PR #52782, opened 2024-06-04
stackrox: remove "-4" from ocp-4-(preview|candidate) test names

davdhacs:
Remove "-4" in the ocp test(config) names. We do not need to track the OCP major version in the test name. We have ocp-stable tests (added for scanner-v4 tests), and this matches that with "ocp-next-candidate"(RC's) and "ocp-dev-preview"(EC's).
"next-candidate" and "dev-preview" are used instead of "candidate" and "preview" to make the names more descriptive matching how these targets are named in other contexts ("next" for upcoming ocp releases targetted by testing, and "dev-preview" to match ocp's "ocp-dev-preview" labeling for versions in EC state, https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp-dev-preview/)

Related but not required (automation-flavor openshift-4 default size and count are large enough, https://github.com/stackrox/automation-flavors/blob/master/openshift-4/entrypoint.sh#L94): removing "-4" from job name matches to set worker size+count in the tested repo: https://github.com/stackrox/stackrox/pull/11406

### DIFF
--- a/ci-operator/jobs/stackrox/stackrox/stackrox-stackrox-master-presubmits.yaml
+++ b/ci-operator/jobs/stackrox/stackrox/stackrox-stackrox-master-presubmits.yaml
@@ -2226,6 +2226,7 @@ presubmits:
     branches:
     - ^master$
     - ^master-
+    cluster: build02
     context: ci/prow/ocp-dev-preview-compliance-e2e-tests
     decorate: true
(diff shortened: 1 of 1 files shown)


gavin-stackrox (approved):
/lgtm

PR #52829, opened 2024-06-05
opendatahub-io/vllm: increase memory limits

dtrifiro:
Each nvcc thread requires up to 3.5 GB of memory, possibly more.


dtrifiro:
/hold 

failures might be unrelated to high memory usage

dtrifiro:
It seems like builds are succeeding for now, meaning we don't need this memory increase, for now.

PR #52848, opened 2024-06-05
Update trustyai-explainability-trustyai-explainability-main.yaml

spolti:


ruivieira:
@spolti Do we also need to change ci-operator/config/trustyai-explainability/trustyai-service-operator/trustyai-explainability-trustyai-service-operator-main.yaml?

ruivieira (changes requested):
LGTM, just needing the same change on the operator.

thanks @spolti

spolti:
> @spolti Do we also need to change ci-operator/config/trustyai-explainability/trustyai-service-operator/trustyai-explainability-trustyai-service-operator-main.yaml?

Hi @ruivieira, before applying there, would be good to merge this first to test, wdyt?

RobGeada:
@spolti https://github.com/openshift/release/pull/52874 replaces this PR: it adds the config to the operator CI as well as uses ubi instead of ubi-minimal

spolti:
okay

PR #52860, opened 2024-06-06
Lower test frequency for 4.17 destructive jobs temporarily

liangxia:


ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.17__amd64-nightly.yaml
```diff
@@ -444,7 +444,7 @@ tests:
     - chain: openshift-e2e-test-hypershift-qe-mgmt
     workflow: cucushift-installer-rehearse-aws-ipi-ovn-hypershift
 - as: aws-ipi-ovn-hypershift-guest-ext-oidc-f7
```
jianlinliu, line 446: Missed this one?

liangxia: I checked when got the msg "11 jobs are running 5 are waiting", and found there are 9 destructive jobs (due to its long running time) among the 11 running jobs. So I'd like to only change the destructive jobs first. For the other jobs, we can do it in a follow up PR if needed.

jianlinliu: okay, let us see how it goes. btw, can you update the title a bit to reflect only update the destructive ci jobs.

liangxia: Title updated.

PR #52864, opened 2024-06-06
Stop 4.14 rollback testing

shellyyang1989:
I was informed that we won't support 4.14 rollback. So stopping the 4.14 testing.


shellyyang1989:
cc @jiajliu @jianlinliu @JianLi-RH @dis016 for review.

shellyyang1989:
@liangxia PTAL. Thank you!

PR #52874, opened 2024-06-06
Update TrustyAI OS image

RobGeada:


ruivieira (approved):

spolti:
Why is it using epel? 
`RUN dnf -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm`

PR #52877, opened 2024-06-06
shiftstack: Promote tooling images

pierreprinetti:
This is the first step in an effort to reuse images rather than building tools from scratch at each execution.

Implements [OSASINFRA-3506](https://issues.redhat.com//browse/OSASINFRA-3506)


mandre (commented):
/lgtm
/approve

PR #52917, opened 2024-06-07
[azure] adjust azure des job frequency

jinyunma:
Reduce the frequency of testing on azure disk encryption set to avoid azure vault key conflict issue.
Because azure vault key is just soft deleted when deleting it or its resource group, and permanently deleted after 90days. High testing frequency might get issue of "keyvault name xxx is already in used" and fail to create it.


jinyunma:
cc @jianlinliu @liangxia to review

PR #52922, opened 2024-06-07
Small enhancement to test result format

liangxia:
Using test step name as the keyword for test report, so that the keywords are unique in the report.

Context for the changes, https://redhat-internal.slack.com/archives/CH76YSYSC/p1717738365976539


rioliu-rh:
LGTM
/assign @kuiwang02

PR #52924, opened 2024-06-07
fix: update pool name to use right OCP version

vprashar2929:
This PR updates the obs-powermon cluster pool name to point to right OCP version for OCP 4.15


IshwarKanse (approved):
/lgtm

PR #52928, opened 2024-06-07
Remove unused aws-centos-40 config map

psalajova:
This config map is not used in any test. A part of the initiative to clean up possibly deprecated derivative cluster profiles (profiles which share both secret and leases with the original aws/gcp/vsphere cluster profile).

/cc @jmguzik


core-services/prow/02_config/_plugins.yaml
```diff
@@ -411,10 +411,6 @@ config_updater:
       cluster_groups:
       - build_farm_ci
       name: prow-job-cluster-launch-installer-openstack-upi-e2e
-    cluster/test-deploy/aws-4.0/*:
```
droslean, line 414: If this is not being used, we should assume that it is safe to delete those files as well.

psalajova: good point

droslean (approved):

PR #52933, opened 2024-06-07
OCPQE-22684: Switch base image from centos:stream8 to rhel-8

liangxia:
centos:stream8 reach EOL as of May 31, and our image stopped building on May 30. Switch to RHEL 8 based images.

### DIFF
--- a/ci-operator/config/openshift/cucushift/openshift-cucushift-master.yaml
+++ b/ci-operator/config/openshift/cucushift/openshift-cucushift-master.yaml
@@ -1,8 +1,8 @@
 base_images:
   os:
-    name: centos
-    namespace: origin
-    tag: stream8
+    name: builder
+    namespace: ocp
+    tag: rhel-8-base-openshift-4.12
 build_root:
   image_stream_tag:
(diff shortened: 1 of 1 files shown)


JianLi-RH:
Sorry, I am not familiar with base images, could you explain why choose  below image especially the tag rhel-8-base-openshift-4.12? 
```
    name: builder
    namespace: ocp
    tag: rhel-8-base-openshift-4.12
```

JianLi-RH:
Because rhel-8-base-openshift-4.12 is the oldest supported OCP version?

liangxia:
> Because rhel-8-base-openshift-4.12 is the oldest supported OCP version?

I just want a rhel 8 image, any one should be fine.

liangxia:
> > Because rhel-8-base-openshift-4.12 is the oldest supported OCP version?
> 
> I just want a rhel 8 image, any one should be fine.

I just randomly pick up one rhel8 images, and verified the image has the repositories we use. Then move on with the PR.

JianLi-RH:
ok, got it.

/lgtm

PR #52988, opened 2024-06-10
chore: update the stage job for PowerMon Operator

vprashar2929:
Since we have a new release of PowerMon Operator, this updates the existing stage job to:

* Point to `tech-preview-0.2` branch for running the e2e

* Renames the config file to be compatible with the `tech-preview-0.2` branch

* Update the config to make use of OCP 4.15 cluster


IshwarKanse (approved):
/lgtm

PR #52996, opened 2024-06-10
release-controller: Allow creation of a new OKD/SCOS release after one hour

LorbusChris:


bradmwilliams:
@LorbusChris Just want to make sure that you're aware of the implications of this change.  Basically, if when a release is Accepted/Rejected, and there are new changes, a new release will almost certainly be fired off.

LorbusChris:
Yes, I'm aware. We need to iterate more quickly while fixing up remaining issues on OKD.
The interval will probable need to be increased again once we have a successful build. 
/cc @dgilmore @Prashanth684

bradmwilliams (approved):

PR #53007, opened 2024-06-10
prow/openshift/os: Allow jira & pipeline failure labels

travier:
Allow setting those two labels freely for Jira tracking and pipeline failures tracking.


travier:
Hum, I'm using https://github.com/kubernetes/test-infra/blob/master/config/prow/plugins.yaml as an example.

PR #53014, opened 2024-06-10
ci-operator/openshift/os: Remove C9S jobs from 4.16

travier:
Those test jobs only make sense on the main development branch.

### DIFF
--- a/ci-operator/jobs/openshift/os/openshift-os-release-4.16-presubmits.yaml
+++ b/ci-operator/jobs/openshift/os/openshift-os-release-4.16-presubmits.yaml
@@ -1,7 +1,7 @@
 presubmits:
   openshift/os:
   - agent: kubernetes
-    always_run: false
+    always_run: true
     branches:
     - ^release-4\.16$
     - ^release-4\.16-
@@ -15,7 +15,6 @@ presubmits:
       pj-rehearse.openshift.io/can-be-rehearsed: "true"
     name: pull-ci-openshift-os-release-4.16-images
     rerun_command: /test images
-    skip_if_only_changed: ^docs/|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|PROJECT|LICENSE)$
     spec:
       containers:
       - args:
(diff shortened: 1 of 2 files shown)


HuijingHei (approved):

LorbusChris:
OKD has two release streams: stable (== 4.16) and next (== 4.17). So I think these tests should still be run somewhere for 4.16

travier:
> OKD has two release streams: stable (== 4.16) and next (== 4.17). So I think these tests should still be run somewhere for 4.16

SCOS should be built only once from the master branch and then specialized for OKD versions. It does not make sense to build SCOS based on older OCP.

/pj-rehearse skip

HuijingHei (approved):
/lgtm

ci-operator/jobs/openshift/os/openshift-os-release-4.16-presubmits.yaml
```diff
@@ -1,7 +1,7 @@
 presubmits:
   openshift/os:
   - agent: kubernetes
-    always_run: false
+    always_run: true
```
c4rt0, line 4: I'm just curious, what's the logic behind this boolean?

c4rt0 (approved):
lgtm

ci-operator/jobs/openshift/os/openshift-os-release-4.16-presubmits.yaml
c4rt0: Ok, NVM.
```    rerun_command: /test images```

travier: This part is automatically generated so I never looked at it :)

jlebon:
> > OKD has two release streams: stable (== 4.16) and next (== 4.17). So I think these tests should still be run somewhere for 4.16
> 
> SCOS should be built only once from the master branch and then specialized for OKD versions. It does not make sense to build SCOS based on older OCP.

I think the new `c9s` variant added in https://github.com/openshift/os/pull/1445 will make this really clear because you would then take this OCP/OKD-agnostic bootable container image and derive from it multiple times (once per supported OKD version). Those base composes don't consume the `c9s-sig-cloud-okd` repo.

PR #53015, opened 2024-06-10
Adding sa, Role and Rolebinding for autoscaler

deepsm007:
/cc @bear-redhat


clusters/build-clusters/build04/assets/auto-scaler-cron.yaml
```diff
@@ -26,3 +27,36 @@ apiVersion: v1
...
+metadata:
+  name: delete-pods
+  namespace: openshift-machine-api
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  verbs:
+  - delete
```
bear-redhat, line 48: add `list` here

PR #53029, opened 2024-06-11
OCPQE-22684: nodejs is required

liangxia:
`centos:stream8` reach EOL as of May 31, and our image stopped building after May 30. So we switch to RHEL 8 based image `rhel-8-base-openshift-4.12` in https://github.com/openshift/release/pull/52933
However that image does not provide nodejs which is required for UI automation, so switch the image to `rhel-8-base-nodejs-openshift-4.12`

### DIFF
--- a/ci-operator/config/openshift/cucushift/openshift-cucushift-master.yaml
+++ b/ci-operator/config/openshift/cucushift/openshift-cucushift-master.yaml
@@ -2,7 +2,7 @@ base_images:
   os:
     name: builder
     namespace: ocp
-    tag: rhel-8-base-openshift-4.12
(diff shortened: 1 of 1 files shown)


JianLi-RH:
Why I can not find rhel-8-base-nodejs-openshift-4.12 in release repo? where is it defined?

liangxia:
> Why I can not find rhel-8-base-nodejs-openshift-4.12 in release repo? where is it defined?

For images used in release repo, most of the dockefiles are defined outside of release repo (and lives in one of the repos under openshift org). I've downloaded rhel-8-base-nodejs-openshift-4.12 locally for tests and can confirm the image exists.

PR #53035, opened 2024-06-11
disable app.ci cluster login due to expired token

rioliu-rh:
job is failed due to expired token, disable login action temporarily,  because we already disabled the resourcepayload update, so it won't impact current logic

### DIFF
--- a/ci-operator/step-registry/release-qe-tests/release-qe-tests-commands.sh
+++ b/ci-operator/step-registry/release-qe-tests/release-qe-tests-commands.sh
@@ -16,7 +16,7 @@ export RELEASE_PAYLOAD_MODIFIER_TOKEN=$release_payload_modifier_token
 export GCS_CRED_FILE=/var/run/vault/release-tests-sa/gcs_sa_for_qe_artifact_access.json
 
 echo "Login cluster app.ci"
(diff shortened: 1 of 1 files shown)


rioliu-rh:
/pj-rehearse periodic-ci-openshift-release-tests-master-qe-release-gate-test

ci-operator/step-registry/release-qe-tests/release-qe-tests-commands.sh
```diff
@@ -16,7 +16,7 @@ export RELEASE_PAYLOAD_MODIFIER_TOKEN=$release_payload_modifier_token
 export GCS_CRED_FILE=/var/run/vault/release-tests-sa/gcs_sa_for_qe_artifact_access.json
 
 echo "Login cluster app.ci"
```
liangxia, line 18: Shall we comment out this line, too ?

rioliu-rh: I think we don't have to do this, when the new token is available. I will revert this change

rioliu-rh: @liangxia this is a temp w/a, the change will be reverted when the token issue is fixed.

rioliu-rh:
/close
token issue is resolved by CRT team, latest job can be succeed

PR #53043, opened 2024-06-11
`helpdesk-faq`: service account and RBAC to get configmaps

smg247:
For: https://issues.redhat.com/browse/DPTP-3521

### DIFF
--- a/clusters/app.ci/assets/admin_helpdesk-faq.yaml
+++ b/clusters/app.ci/assets/admin_helpdesk-faq.yaml
@@ -1,26 +1,24 @@
-kind: List
-apiVersion: v1
-items:
-- kind: ClusterRole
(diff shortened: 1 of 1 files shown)


clusters/app.ci/assets/admin_helpdesk-faq.yaml
```diff
@@ -0,0 +1,26 @@
+kind: List
```
bear-redhat, line 1: We probably don't need the List here

bear-redhat (changes requested):

clusters/app.ci/assets/admin_helpdesk-faq.yaml
deepsm007: Do we?

PR #53051, opened 2024-06-11
Add upgrade and heavy build in 4.16 and 4.17 ppc64le release controller-trigger jobs

KeerthanaAP:


tvardema:
Are we sure we have the resources to support these additions?  Remember, the resources for every job is shared with every other job, so we're talking about going from using 1 lease per detected build, to 3.  Meaning, at any given time, we might be consuming up to 5 total leases.  I will be happily proven incorrect, but I'm pretty sure this will cause more job failures...

tvardema:
After a discussion with Jeremy P, we're okay with merging these additions to the informing jobs, but we need to pay close attention to these jobs and any time they fail gathering the "why".

KeerthanaAP:
> After a discussion with Jeremy P, we're okay with merging these additions to the informing jobs, but we need to pay close attention to these jobs and any time they fail gathering the "why".

Agreed.

KeerthanaAP:
@xueqzhan @sosiouxme Could you please review this PR.

sosiouxme (approved):
/lgtm

PR #53052, opened 2024-06-11
openstack: Move cleanup jobs to shiftstack/ci

pierreprinetti:
and reuse images instead of building a binary at each execution.

Implements [OSASINFRA-3506](https://issues.redhat.com//browse/OSASINFRA-3506)


mandre (commented):
/lgtm

PR #53065, opened 2024-06-11
fix: OADP default branches with go1.22

mateusoliveira43:
Add go 1.22.2 to all remaining OADP repos' default branches

### DIFF
--- a/ci-operator/config/openshift/velero/openshift-velero-konveyor-dev.yaml
+++ b/ci-operator/config/openshift/velero/openshift-velero-konveyor-dev.yaml
@@ -33,7 +33,7 @@ tests:
   steps:
     test:
     - as: unit
-      commands: GOPATH=/opt/app-root/src/go make -f Makefile.prow ci
+      commands: PATH="/go/bin:$PATH" make -f Makefile.prow ci
       from: src
       resources:
         requests:


mpryc (approved):

mpryc (approved):

ci-operator/config/openshift/velero/openshift-velero-konveyor-dev.yaml
```diff
@@ -33,7 +33,7 @@ tests:
   steps:
     test:
     - as: unit
-      commands: make -f Makefile.prow ci
+      commands: PATH="/go/bin:$PATH" make -f Makefile.prow ci
```
mateusoliveira43, line 36: had to do this because GOPATH is set to `/go` (there is a `from src` right below)

To test this will work with velero fork, create Dockerfile in root of that repo
```dockerfile
FROM quay.io/konveyor/builder:ubi9-v1.22.2

WORKDIR /go/src/github.com/openshift/velero

ENV GOPATH=/go

COPY . .
```
In Makefile.prow change line 10 `CONGEN_VERSION=0.14.0`

In Makefie change line 87 `else ifeq ($(shell which docker && cat $(shell which docker) | grep -c "exec podman"), 1)`

Build image `docker image build --tag oadp-development -f test.Dockerfile .` and run it
`docker container run -it --rm oadp-development PATH="/go/bin:$PATH" make -f Makefile.prow ci`

mateusoliveira43: rhel 8 1.22 would probably work, but preferred to use same image across branches

shubham-pampattiwar (approved):

PR #53100, opened 2024-06-12
Update for removal of sdn image

danwinship:
openshift-sdn has been removed from the release image, update for that.

In 4.17, we are still building kube-proxy from the sdn repo, so this doesn't entirely remove the repo from the CI config.

(Yes, there is no CI that checks that the generated kube-proxy image actually works. This has always been the case...)


tssurya (approved):
lgtm

ci-operator/jobs/openshift/sdn/openshift-sdn-master-presubmits.yaml
```diff
@@ -1,321 +1,5 @@
 presubmits:
   openshift/sdn:
```
tssurya, line 2: when will this file go away? given there are no more jobs in sdn repo moving forward?

danwinship: When we move the kube-proxy build somewhere else. (Or possibly just stop building it.)

PR #53101, opened 2024-06-12
OCPQE-20152 disable baremetalds-ipi-ovn-dualstack-primaryv6 and barem…

ming1013:
Disable upgrade job with baremetalds-ipi-ovn-dualstack-primaryv6 and baremetalds-ipi-ovn-ipv6-fips profile
 cc @jiajliu @jianlinliu @shellyyang1989 @JianLi-RH PTAL, thank you!


shellyyang1989:
Hi Ming, we need to drop it from config (ci-operator/config/openshift/openshift-tests-private/) and then run make update.

ming1013:
> Hi Ming, we need to drop it from config (ci-operator/config/openshift/openshift-tests-private/) and then run make update.

Thanks. @shellyyang1989 I added config files update, but met error when make update, will check more

jianlinliu:
cc @liangxia to review.

PR #53105, opened 2024-06-12
Randomizing and repeating functional tests to detect flakyness

jpodivin:
Depends-On: https://github.com/openstack-k8s-operators/openstack-operator/pull/931

### DIFF
--- a/ci-operator/config/openstack-k8s-operators/openstack-operator/openstack-k8s-operators-openstack-operator-main.yaml
+++ b/ci-operator/config/openstack-k8s-operators/openstack-operator/openstack-k8s-operators-openstack-operator-main.yaml
@@ -61,7 +61,7 @@ tests:
     mkdir -p ../operator && cp -r . ../operator
     cd ../operator
     export GOFLAGS=
-    make test GINKGO_ARGS='--no-color'
+    make test-ci
   container:
     from: src
 - as: precommit-check


bogdando (commented):
lgtm, good idea

ci-operator/config/openstack-k8s-operators/dataplane-operator/openstack-k8s-operators-dataplane-operator-main.yaml
```diff
@@ -64,7 +64,7 @@ tests:
     mkdir -p ../operator && cp -r . ../operator
     cd ../operator
     export GOFLAGS=
-    make test GINKGO_ARGS='--no-color'
+    make test GINKGO_ARGS='--no-color --repeat=5 --randomize-all'
```
SeanMooney, line 67: i woudl not add --repeat=5 in ci

if you want to support repeate for local execution i would add a new ENV var and default it to 1
--randomize-all i do agree with

jpodivin: Running a single randomized test wouldn't give us as much benefit. Purpose of this PR is to prevent merging of flaky code, which can slip by thanks to a "lucky" sequence of test cases. If run multiple random tests, the probability of something like this happening drops considerably.

Although for obvious reasons I can't say by how much since I don't know how many "lucky" combinations are there. But with successive tests the probability of only hitting "lucky" combinations drops pretty fast to zero.

fmount: -1: I think that value hardcoded to 5 is not a good default (unless you have good reasons to think that 5 is the right default).
I agree with @SeanMooney and use a `ENV`  var that we can override in the future without requiring a new change here.
Also, are you going to propose this for the service operators or this change applies to openstack-operator only? Also, considering that dataplane operator has been merged into openstack operator, shouldn't just remove and cleanup this part?

fmount (changes requested):

ci-operator/config/openstack-k8s-operators/dataplane-operator/openstack-k8s-operators-dataplane-operator-main.yaml
jpodivin: Adding var to makefile can help, but it would force repeated runs locally as well. That would quickly become annoying and possibly lead to overriding of that var. I want to avoid that by setting it here. Where it will only affect CI and nothing else.

Furthermore, setting repetitions to `1`, as @SeanMooney  proposed won't do us any good, unless we get very lucky.

jpodivin: Turned the setting into variable in openstack-operator repo. This means that the PR now has a dependency it has to pull in order to execute. But it's the only way to satisfy conditions laid out.

The leftover from dataplane repo was removded some time ago

bogdando:
lgtm

PR #53128, opened 2024-06-12
OSSM-6653: adding jobs for new release-3.0-tp1 branch

FilipB:


ci-operator/config/openshift-service-mesh/sail-operator/openshift-service-mesh-sail-operator-release-3.0-tp1.yaml
```diff
@@ -0,0 +1,68 @@
+base_images:
+  cli:
+    name: "4.15"
+    namespace: ocp
+    tag: cli
+build_root:
+  image_stream_tag:
+    name: maistra-builder
+    namespace: ci
+    tag: upstream-master
```
jwendell, line 10: We are using the `upstream_master` image to run jobs in the main branch, which is fine because operator master is supposed to work with istio master.

When we create a branch that will track a particular istio release, e.g. 1.22, then we should use the 1.22 image.

I think this is safe now, but we might want to change this in the future.

jwendell (approved):

ci-operator/config/openshift-service-mesh/sail-operator/openshift-service-mesh-sail-operator-release-3.0-tp1.yaml
FilipB: I will change this, thanks.

FilipB: Actually 1.22 tag is not available yet - https://console.cloud.google.com/gcr/images/istio-testing/GLOBAL/build-tools?pli=1
So I will keep the `upstream_master` there for now.

jwendell: It is. For instance, this is the latest one: https://console.cloud.google.com/gcr/images/istio-testing/global/build-tools@sha256:0469340f2ad428a11219f7d09cd18f6578e6f1bb9a6230563d818ecab01fcc31/details

But like I said, this is not critical. We can do it in a follow up.

PR #53139, opened 2024-06-12
HOSTEDCP-1446: ci-operator: hypershift: remove unused flag

stevekuznetsov:
This flag was introduced in 4.16, but only jobs older than that were ever configured to use it. The flag has clearly never worked, so it's not really clear what it's doing here or why. We can turn it on for real in the future if we need it.


sjenning:
/pj-rehearse pull-ci-openshift-hypershift-release-4.15-e2e-aws
/approve
/lgtm

PR #53158, opened 2024-06-13
Remove release-4.16-altinfra-openstack-capi-manifests in installer

mandre:
The job has no purpose and is broken. This was already removed from master in https://github.com/openshift/release/pull/52518. Let's do it for release-4.16 as well.


barbacbd (commented):
/approve

PR #53183, opened 2024-06-13
MULTIARCH-4593: Use multiarch base image for promtail

jeffdyoung:
Use multiarch base image for promtail


deepsm007:
Required by https://github.com/openshift/loki/pull/304

PR #53187, opened 2024-06-13
add ovn to 4.16+ hypershift job names

jluhrsen:

this was missed in an earlier commit which only
did this for 4.15:
  https://github.com/openshift/release/commit/9674f7b44d4c6dacbc14db52b332cba349b92032


jluhrsen:
here's another one @muraee

PR #53193, opened 2024-06-13
stackrox: remove not-implemented sensor-profiling

davdhacs:
This was added during a hackathon, https://github.com/openshift/release/pull/43771, but the implementation was not added in master stackrox/stackrox.


gavin-stackrox (approved):
/lgtm

PR #53200, opened 2024-06-13
Update HyperShift trigger

liswang89:


ci-operator/step-registry/trigger-jobs/trigger-jobs-commands.sh
```diff
@@ -8,6 +8,17 @@ GANGWAY_API_TOKEN=$(cat $SECRETS_DIR/gangway-api-token)
 WEEKLY_JOBS="$SECRETS_DIR/$JSON_TRIGGER_LIST"
 URL="https://gangway-ci.apps.ci.l2s4.p1.openshiftapps.com"
 
+# Check if it is hypershift jobs
+if [[ $WEEKLY_JOBS == *"hypershift"* ]]; then
+  # Get the current ISO week number (1-53)
+  WEEK_NUM=$(date +%V)
+  # Check if the week number is odd
+  if [ $((WEEK_NUM % 2)) -ne 0 ]; then
+    echo "Not Running HyperShift testing due to it being an odd-numbered week"
```
oharan2, line 17: Can we also say that hypershift testing is currently running every other week? The reason for the odd number may not be clear

liswang89: Though we wanted to run the testing every other week, we did it on the 2nd and 4th Fridays of each month (even-numbered weeks). The PR makes it clear to skip the testing on odd-numbered weeks. If approved, the difference in the Test Plan may be worth mentioning.

PR #53238, opened 2024-06-14
OCPEDGE-1105: Ibu pull secret refactor

jeff-roche:
- Move pull secrets to a file upload once at the beginning of the workflow to avoid it getting printed to console
- Update all steps that printed the pull secret to now read the pull secret from the uploaded file
- Swap OWNERS files out for symlinks to the parent folder

### DIFF
--- a/ci-operator/step-registry/openshift/image-based/upgrade/metal/config/openshift-image-based-upgrade-metal-config-commands.sh
+++ b/ci-operator/step-registry/openshift/image-based/upgrade/metal/config/openshift-image-based-upgrade-metal-config-commands.sh
@@ -213,6 +213,10 @@ PULL_SECRET=$(cat ${CLUSTER_PULL_SECRET_FILE} ${LCA_PULL_SECRET_FILE} | jq -cs '
 BACKUP_SECRET_FILE="/var/run/ibu-backup-secret/.backup-secret"
 BACKUP_SECRET=$(jq -c . ${BACKUP_SECRET_FILE})
 
+# Save the pull secrets
+echo -n "${PULL_SECRET}" > ${SHARED_DIR}/.pull_secret.json
+echo -n "${BACKUP_SECRET}" > ${SHARED_DIR}/.backup_secret.json
+
 echo "Transferring pull secrets..."
 scp "${SSHOPTS[@]}" ${SHARED_DIR}/.pull_secret.json $ssh_host_ip:$remote_workdir
 scp "${SSHOPTS[@]}" ${SHARED_DIR}/.backup_secret.json $ssh_host_ip:$remote_workdir


jeff-roche:
/pj-rehearse pull-ci-openshift-kni-lifecycle-agent-main-ibi-e2e-flow pull-ci-openshift-kni-lifecycle-agent-main-ibu-e2e-flow

ci-operator/step-registry/openshift/image-based/upgrade/seed/create/openshift-image-based-upgrade-seed-create-commands.sh
```diff
@@ -12,12 +12,10 @@ SSHOPTS=(-o 'ConnectTimeout=5'
   -o LogLevel=ERROR
   -i "${CLUSTER_PROFILE_DIR}/ssh-privatekey")
 
-PULL_SECRET_FILE="/var/run/pull-secret/.dockerconfigjson"
-PULL_SECRET=$(cat ${PULL_SECRET_FILE})
-BACKUP_SECRET_FILE="/var/run/ibu-backup-secret/.backup-secret"
-BACKUP_SECRET=$(cat ${BACKUP_SECRET_FILE})
-SEED_VM_NAME="seed"
 remote_workdir=$(cat ${SHARED_DIR}/remote_workdir)
+PULL_SECRET_FILE=$(cat ${SHARED_DIR}/pull_secret_file)
```
eggfoobar, line 16: Small nit, since we cat the file here, technically `PULL_SECRET_FILE` and `BACKUP_SECRET_FILE` is not referencing a file any longer, can we name it just `PULL_SECRET` or `BACKUP_SECRET` if we cat the file?

ci-operator/step-registry/openshift/image-based/upgrade/e2e/openshift-image-based-upgrade-e2e-commands.sh
```diff
@@ -16,23 +15,20 @@ remote_workdir=$(cat ${SHARED_DIR}/remote_workdir)
 instance_ip=$(cat ${SHARED_DIR}/public_address)
 host=$(cat ${SHARED_DIR}/ssh_user)
 ssh_host_ip="$host@$instance_ip"
-PULL_SECRET_FILE="/var/run/pull-secret/.dockerconfigjson"
-PULL_SECRET=$(cat ${PULL_SECRET_FILE})
+PULL_SECRET_FILE=$(cat ${SHARED_DIR}/pull_secret_file)
```
eggfoobar, line 18: Same nit regarding catting a file

ci-operator/step-registry/openshift/image-based/upgrade/seed/create/openshift-image-based-upgrade-seed-create-commands.sh
jeff-roche: The contents of `pull_secret_file` is actually the path to the file on the ec2 instance so it is in fact a filepath and not actually the pull secret itself. With that context, what do you think in terms of variable name? I was back and forth on what it should be called

ci-operator/step-registry/openshift/image-based/upgrade/e2e/openshift-image-based-upgrade-e2e-commands.sh
```diff
@@ -16,23 +15,20 @@ remote_workdir=$(cat ${SHARED_DIR}/remote_workdir)
...
 remote_artifacts_dir=${remote_workdir}/artifacts
 
 cat <<EOF > ${SHARED_DIR}/e2e_test.sh
 #!/bin/bash
-set -xeuo pipefail
+set -euo pipefail
 
 export KUBECONFIG='${target_kubeconfig}'
-export PULL_SECRET='${PULL_SECRET}'
+export PULL_SECRET=\$(<${PULL_SECRET_FILE})
```
suleymanakbas91, line 29: I'm confused here. This will try to read the content from the file path, but `PULL_SECRET_FILE` is not a file path anymore but it includes the content of the file because of the 'cat', right?

jeff-roche: The contents of the file is now the path on the ec2 instance to the `.pull-secret.json`. The value of the pull secret is no longer being passed in these steps but rather read from the file on the ec2 instance. 

The command that would be run on the ec2 instance in this case would look something like `export PULL_SECRET=$(</usr/ec2-user/workdir/.pull-secret.json)` which forward the pull secret into the variable without printing it

PR #53243, opened 2024-06-14
Disable install retries for AWS

dgoodwin:
Install retries on AWS can collide when re-using install assets. Disabling for now.


patrickdillon:
Change looks good to me. The two non-code caveats:

1. The collisions will be resolved by https://github.com/openshift/installer/pull/8593 so this can just be temporary
2. This will potentially show a decrease in stability for 4.15 and earlier if they have regular infrastructure failure retries

dgoodwin:
Time is critical here, given the diff in the change, I'm going to ack rehersals immediately. We need to know if something is real or not for next weeks decisions and this may expose the underlying problem.

/pj-rehearse ack

PR #53261, opened 2024-06-16
Update frequency of GPU tests

empovit:
* Make 4.16 run daily
* Make 4.14 and 4.15 run weekly


empovit:
/pj-rehearse skip 
as only the job frequency is affected

PR #53294, opened 2024-06-17
openstack/SRIOV: don't install from source anymore

EmilienM:
We have the SRIOV images in the release payload.


MaysaMacedo:
/lgtm

happy to see we can remove this code

PR #53299, opened 2024-06-17
Remove gcp-4.0

psalajova:
This config map is not used in any test. 

A part of the initiative to clean up possibly deprecated derivative cluster profiles (profiles which share both secret and leases with the original aws/gcp/vsphere cluster profile).

/cc @droslean @vrutkovs 
/hold


vrutkovs:
/lgtm

Leaving for you or Nikolaos to remove the hold

PR #53332, opened 2024-06-18
Bump timeout of serial jobs to 5h

derekhiggins:
These are timing out with no obvious errors, bump up the timeout to see if they eventually succeed.


dtantsur:
/lgtm

We probably need to continue investigating, we cannot bump the timeout infinitely.

PR #53340, opened 2024-06-18
OSASINFRA-3488: hypershift: Optional presubmit test for OpenStack

mdbooth:
Initial presubmit test for OpenStack jobs. Test is optional and does not yet
pass.

The job is not required to pass for merging. We will merge the job before the code which implements it.


mdbooth:
This will obviously fail, but I want to check it fails in the expected manner.

/pj-rehearse

mdbooth:
That appears to have failed in the expected manner.

EmilienM:
/lgtm
/approve
/pj-rehearse ack

To me that looks like a solid start, thanks.

ci-operator/jobs/openshift/hypershift/openshift-hypershift-main-presubmits.yaml
```diff
@@ -757,6 +757,85 @@ presubmits:
         secret:
           secretName: result-aggregator
     trigger: (?m)^/test( | .* )e2e-kubevirt-metal-conformance,?($|\s.*)
+  - agent: kubernetes
+    always_run: true
```
MaysaMacedo, line 761: should we always run this job? maybe it's better to only run if we explicitly trigger it

ci-operator/step-registry/hypershift/openstack/e2e/execute/hypershift-openstack-e2e-execute-commands.sh
```diff
@@ -0,0 +1,34 @@
...
+        --e2e.latest-release-image=${OCP_IMAGE_LATEST} \
+        --e2e.previous-release-image=${OCP_IMAGE_PREVIOUS} \
+        --e2e.pull-secret-file=/etc/ci-pull-credentials/.dockerconfigjson \
+        --e2e.node-pool-replicas=2 \
+        --e2e.base-domain=ci.hypershift.devcluster.openshift.com \
+        --e2e.external-dns-domain=service.ci.hypershift.devcluster.openshift.com \
+        --test.run='^TestCreateCluster$' \
+        --e2e.platform="OpenStack" \
+        --e2e.openstack-node-flavor="${OPENSTACK_COMPUTE_FLAVOR}" \
+        --e2e.openstack-node-external-network-name="${OPENSTACK_EXTERNAL_NETWORK}" \
```
MaysaMacedo, line 33: this doesn't look specific to the node, so it's worth removing node from the name

ci-operator/step-registry/hypershift/openstack/e2e/execute/hypershift-openstack-e2e-execute-commands.sh
```diff
@@ -0,0 +1,34 @@
...
+        --e2e.openstack-credentials-file="${SHARED_DIR}/clouds.yaml"
```
MaysaMacedo, line 34: is the framework going to create the secret?

ci-operator/step-registry/hypershift/openstack/e2e/execute/hypershift-openstack-e2e-execute-commands.sh
```diff
@@ -0,0 +1,34 @@
...
+        --test.timeout=0 \
+        --e2e.latest-release-image=${OCP_IMAGE_LATEST} \
+        --e2e.previous-release-image=${OCP_IMAGE_PREVIOUS} \
+        --e2e.pull-secret-file=/etc/ci-pull-credentials/.dockerconfigjson \
+        --e2e.node-pool-replicas=2 \
+        --e2e.base-domain=ci.hypershift.devcluster.openshift.com \
+        --e2e.external-dns-domain=service.ci.hypershift.devcluster.openshift.com \
+        --test.run='^TestCreateCluster$' \
+        --e2e.platform="OpenStack" \
+        --e2e.openstack-node-flavor="${OPENSTACK_COMPUTE_FLAVOR}" \
```
MaysaMacedo, line 32: If we're also handling node pool, do we also need an image?

ci-operator/jobs/openshift/hypershift/openshift-hypershift-main-presubmits.yaml
mdbooth: Hmm. No, this was definitely not my intention. What did I miss? 🤔

ci-operator/step-registry/hypershift/openstack/e2e/execute/hypershift-openstack-e2e-execute-commands.sh
mdbooth: No, that's generated by a prior step. I'll add a check here that it has been generated, though.

ci-operator/step-registry/hypershift/openstack/e2e/execute/hypershift-openstack-e2e-execute-commands.sh
mdbooth: I think we're going to have to create that ourselves. That's going to be a PITA, btw.

mdbooth:
Most recent change:
- updates `always_run` to false
- verifies that `clouds.yaml` has been generated so the e2e test will fail early if misconfigured
- renames the `--e2e.openstack-external-network-name` argument

mdbooth:
Looking good.

davidvossel (commented):
/approve

The job looks great. The main things we want here are to ensure it is both optional and set to not always run (at least initially until this is stable)

I see the job is failing right now due to `flag provided but not defined: -e2e.openstack-credentials-file` I'm guessing there's a chicken and egg issue here where you haven't added some new cli arg yet to the hypershift test binary, but you want a job you can execute to ensure the test driver changes work.

i gave the PR an approve, so feel free to merge it whenever you all are ready. I don't see anything here that risks stability or merge gating.

PR #53360, opened 2024-06-18
Finish migrating some metal jobs to OVN

elfosardo:


elfosardo:
I don't expect the 4.18 jobs to pass

elfosardo:
/pj-rehearse ack
thinking about it we need this in any case, so just moving on

PR #53367, opened 2024-06-18
hypershift: Use build cache

mdbooth:
See https://docs.ci.openshift.org/docs/architecture/ci-operator/#using-the-build-cache

The following data is taken from the job rehearsals from this PR, which appear to be uncached. Time from job start until `Build hypershift-amd64 succeeded`. Note that in all cases, job start to `Building src` is only a few seconds.

* e2e-aws: 13 minutes
* e2e-aws-metrics: 10 minutes
* e2e-azure: 8 minutes
* e2e-conformance: 9 minutes
* e2e-conformance-azure: 9 minutes

I'd expect all of these to be about the same, so presumably there is significant jitter.

We can compare these to the same jobs after merging to see what the impact was on build times.

### DIFF
--- a/ci-operator/config/openshift/hypershift/openshift-hypershift-main.yaml
+++ b/ci-operator/config/openshift/hypershift/openshift-hypershift-main.yaml
@@ -22,6 +22,7 @@ base_images:
 binary_build_commands: make build
 build_root:
   from_repository: true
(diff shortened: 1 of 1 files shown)


mdbooth:
Nothing failed to build, so

/pj-rehearse ack

PR #53372, opened 2024-06-18
Move HyperShift control plane to multi-arch nodepool

bear-redhat:


deepsm007:
/lgtm 
to unblock broken cluster-bot

PR #53381, opened 2024-06-18
lightspeed-operator: raise synk threshold to high and set .synk file as policy config

raptorsun:
lightspeed-operator: raise synk threshold to high and set .synk file as policy config

### DIFF
--- a/ci-operator/config/openshift/lightspeed-operator/openshift-lightspeed-operator-main.yaml
+++ b/ci-operator/config/openshift/lightspeed-operator/openshift-lightspeed-operator-main.yaml
@@ -110,6 +110,8 @@ tests:
 - as: security
   optional: true
   steps:
+    env:
+      SNYK_CODE_ADDITIONAL_ARGS: --severity-threshold=high --policy-path=.snyk
     workflow: openshift-ci-security
 zz_generated_metadata:
   branch: main


ci-operator/config/openshift/lightspeed-operator/openshift-lightspeed-operator-main.yaml
```diff
@@ -110,6 +110,9 @@ tests:
 - as: security
   optional: true
   steps:
+    env:
+      PROJECT_NAME: lightspeed-operator
```
bparees, line 114: any reason you needed to add this?

see the discussion i had w/ TP about this here:
https://redhat-internal.slack.com/archives/CBN38N3MW/p1716315436075249?thread_ts=1716314770.838859&cid=CBN38N3MW

i'd rather let them default it than us set it(so that if at some point in the future the expected value changes, we get it for free), unless there is a reason we think the default is wrong.

ci-operator/config/openshift/lightspeed-operator/openshift-lightspeed-operator-main.yaml
```diff
@@ -110,6 +110,9 @@ tests:
 - as: security
   optional: true
```
bparees, line 111: once you have this passing, we should remove it from "optional"

ci-operator/config/openshift/lightspeed-operator/openshift-lightspeed-operator-main.yaml
raptorsun: Yes, let's use default project name. New commit is pushed.

ci-operator/config/openshift/lightspeed-operator/openshift-lightspeed-operator-main.yaml
raptorsun: A Jira ticket is created to track this https://issues.redhat.com/browse/OLS-751

PR #53386, opened 2024-06-18
ipi-install-install: Limit gathered capi artifacts

patrickdillon:
openshift/installer#8600 started saving the etcd data store into the .clusterapi_output directory. This changes our install step so that we don't save the data store into the artifacts dir.

The .clusterapi_output directory contains some artifacts which should not be gathered, because they are useless or, in the case of the etcd data store, very large. Instead we should select the artifacts we want, which are the manifests (*.yaml) and controller logs (*.log).

The etcd data dir (.clusterapi_output/etcd) does not contain any .yaml or .log files, so this commit does not explicitly exclude the directory, instead we just designate the files we want:

{,**/}: looks in the current dir and all subfolders
*.{log,yaml}: selects all .yaml & .log files

### DIFF
--- a/ci-operator/step-registry/ipi/install/install/aws/ipi-install-install-aws-commands.sh
+++ b/ci-operator/step-registry/ipi/install/install/aws/ipi-install-install-aws-commands.sh
@@ -36,7 +36,8 @@ function populate_artifact_dir() {
   # Copy CAPI-generated artifacts if they exist
   if [ -d "${dir}/.clusterapi_output" ]; then
     echo "Copying Cluster API generated manifests..."
-    cp -rpv "${dir}/.clusterapi_output" "${ARTIFACT_DIR}/" 2>/dev/null
+    mkdir -p "${ARTIFACT_DIR}/clusterapi_output/"
+    cp -rpv "${dir}/.clusterapi_output/"{,**/}*.{log,yaml} "${ARTIFACT_DIR}/clusterapi_output/" 2>/dev/null
   fi
 }
 
--- a/ci-operator/step-registry/ipi/install/install/ipi-install-install-commands.sh
+++ b/ci-operator/step-registry/ipi/install/install/ipi-install-install-commands.sh
@@ -127,7 +127,8 @@ function populate_artifact_dir() {
   # Copy CAPI-generated artifacts if they exist
   if [ -d "${dir}/.clusterapi_output" ]; then
     echo "Copying Cluster API generated manifests..."
-    cp -rpv "${dir}/.clusterapi_output" "${ARTIFACT_DIR}/" 2>/dev/null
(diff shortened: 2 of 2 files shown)


patrickdillon:
/cc @r4f4 

I'm super confident this bash is going to work seamlessly

r4f4:
/pj-rehearse pull-ci-openshift-installer-master-altinfra-e2e-aws-ovn

ci-operator/step-registry/ipi/install/install/aws/ipi-install-install-aws-commands.sh
```diff
@@ -36,7 +36,7 @@ function populate_artifact_dir() {
   # Copy CAPI-generated artifacts if they exist
   if [ -d "${dir}/.clusterapi_output" ]; then
     echo "Copying Cluster API generated manifests..."
-    cp -rpv "${dir}/.clusterapi_output" "${ARTIFACT_DIR}/" 2>/dev/null
+    cp -rpv "${dir}/.clusterapi_output/"{,**/}*.{log,yaml} "${ARTIFACT_DIR}/" 2>/dev/null
```
r4f4, line 39: Would you be against
```suggestion
    cp -rpv "${dir}/.clusterapi_output/"{,**/}*.{log,yaml} "${ARTIFACT_DIR}/clusterapi_output" 2>/dev/null
```

The shell-fu [is working](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_release/53386/rehearse-53386-pull-ci-openshift-installer-master-altinfra-e2e-aws-ovn/1803128670945021952/artifacts/e2e-aws-ovn/ipi-install-install/artifacts/) but I think I preferred when the capi output was inside its own directory.

r4f4: (I think you'd have to `mkdir -p ${ARTIFACTS_DIR}/clusterapi_output` first, but you get the idea).

patrickdillon: > Would you be against

> I preferred when the capi output was inside its own directory.

No, it was my intent as well to keep this stuff in its own dir. incorporated changes into the most recent push and running a rehearsal.

r4f4 (commented):
/lgtm

r4f4:
Looking good now: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_release/53386/rehearse-53386-pull-ci-openshift-installer-master-e2e-aws-ovn/1803790342852448256/artifacts/e2e-aws-ovn/ipi-install-install/artifacts/

PR #53393, opened 2024-06-18
NO-JIRA: increase azure arm quota

jeffdyoung:
increase azure arm quota


deepsm007:
need to run `make boskos-config`

PR #53416, opened 2024-06-19
Patch cron again

shaior:


natifridman:
Could this cause issues if the tests need to run for more than 3 hours?

PR #53431, opened 2024-06-19
ART-9391 Bump 4.15 rhel8 repos to ubi

locriandev:
Ref. https://issues.redhat.com/browse/ART-9391


joepvd:
/lgtm
Timing of merge should be controlled, setting this out of an abundance of caution:
/hold

PR #53433, opened 2024-06-19
Bump all metal-ipi serial job timouts to 5h

derekhiggins:
We previously bumped up the timeout on periodic and CBO serial jobs. 
This seems to have helped in periodic jobs. Only bumping up for releases >= 4.16.


derekhiggins:
cc @stbenjam @dgoodwin ,  setting all metal-ipi serial jobs to match that we already do for the periodic jobs

dtantsur:
/approve

for ICC

soltysh (approved):
/approve

PR #53447, opened 2024-06-20
remove sdn clsuter on 4.17

zhaozhanqi:
SDN cluster is not supported on 4.17. 
@jianlinliu @liangxia Could you help review it, thanks


jianlinliu:
/lgtm

btw, what is the policy on upgrade path for those sdn cluster installed from old versions?

PR #53458, opened 2024-06-20
[Regular Check] Optimize HyperShift AWS jobs

LiangquanLi930:
Consider balancing cloud cost saving and our business request matching, QE plan to reduce the frequency of AWS jobs and remove some of jobs which can be replaced by other high priority jobs.


LiangquanLi930:

PR #53318, opened 2024-06-18
OCPQE-22436 Adding y version upgrade jobs for OCP 4.17

JianLi-RH:
/cc @jianlinliu @shellyyang1989 @jiajliu @dis016 @evakhoni

PR #53470, opened 2024-06-20
do migration to ovn if target version is 4.16 and sdn

zhaozhanqi:
@JianLi-RH @shellyyang1989 @weliang1 Could you help review this.

### DIFF
--- a/ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
+++ b/ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
@@ -254,6 +254,35 @@ EOF
     ansible-playbook -i "${SHARED_DIR}/ansible-hosts" /tmp/repo.yaml -vvv
 }
 
+# Do sdn migration to ovn since sdn is not supported from 4.17 version
+function sdn2ovn(){
+    oc patch network.operator.openshift.io cluster --type='merge'  -p='{"spec":{"defaultNetwork":{"ovnKubernetesConfig":{"ipv4":{"internalJoinSubnet": "100.65.0.0/16"}}}}}' 
+    oc patch network.operator.openshift.io cluster --type='merge'  -p='{"spec":{"defaultNetwork":{"ovnKubernetesConfig":{"ipv4":{"internalTransitSwitchSubnet": "100.85.0.0/16"}}}}}' 
+    oc patch Network.config.openshift.io cluster --type='merge' --patch '{"metadata":{"annotations":{"network.openshift.io/network-type-migration":""}},"spec":{"networkType":"OVNKubernetes"}}'
+    timeout 300s bash <<EOT
+    until 
+       oc get network -o yaml | grep NetworkTypeMigrationInProgress > /dev/null
+    do
+       echo "Migration is not started yet"
+       sleep 10
+    done
+EOT
+    echo "Start Live Migration process now"
+    # Wait for the live migration to fully complete
+    timeout 3600s bash <<EOT
+    until 
+       oc get network -o yaml | grep NetworkTypeMigrationCompleted > /dev/null && \
+       for NODE in \$(oc get nodes -o custom-columns=NAME:.metadata.name --no-headers); do oc get node \$NODE -o yaml | grep "k8s.ovn.org/node-transit-switch-port-ifaddr:" | grep "100.85";  done > /dev/null && \
+       for NODE in \$(oc get nodes -o custom-columns=NAME:.metadata.name --no-headers); do oc get node \$NODE -o yaml | grep "k8s.ovn.org/node-gateway-router-lrp-ifaddr:" | grep "100.65";  done > /dev/null && \
+       oc get network.config/cluster -o jsonpath='{.status.networkType}' | grep OVNKubernetes > /dev/null;
+    do
+       echo "Live migration is still in progress"
+       sleep 300
+    done
+EOT
+    echo "The Migration is completed"
+}
+
 # Upgrade RHEL node
 function rhel_upgrade(){
     echo "Upgrading RHEL nodes"
@@ -998,6 +1027,13 @@ for target in "${TARGET_RELEASES[@]}"; do
     fi
     check_history
     health_check
+    currentPlugin=$(oc get network.config.openshift.io cluster -o jsonpath='{.status.networkType}')
+    if [[ ${TARGET_MINOR_VERSION} == "16" && ${currentPlugin} == "OpenShiftSDN" ]]; then
+	echo "The cluster is running version 4.16 with OpenShift SDN, and it needs to be migrated to OVN before upgrading"
+	sdn2ovn
+	health_check
+    fi
+
     if [[ -n "${E2E_RUN_TAGS}" ]]; then
         run_upgrade_e2e "${index}"
     fi


zhaozhanqi:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.16-amd64-nightly-4.16-upgrade-from-stable-4.7-gcp-ipi-f30

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
```diff
@@ -998,6 +1053,13 @@ for target in "${TARGET_RELEASES[@]}"; do
     fi
     check_history
     health_check
+    currentVersion=$(oc version -o yaml | grep openshiftVersion | grep -o '[0-9]*[.][0-9]*' | head -1)
+    currentPlugin=$(oc get network.config.openshift.io cluster -o jsonpath='{.status.networkType}')
+    if [[ ${currentVersion} == "4.16" && ${currentPlugin} == "OpenShiftSDN" ]]; then
+	echo "the cluster is 4.16 and openshift-sdn need to migration to ovn and then upgrade if need"
```
weliang1, line 1059: Suggest to use "The cluster is running version 4.16 with OpenShift SDN, and it needs to be migrated to OVN before upgrading"

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
```diff
@@ -998,6 +1053,13 @@ for target in "${TARGET_RELEASES[@]}"; do
     fi
     check_history
     health_check
+    currentVersion=$(oc version -o yaml | grep openshiftVersion | grep -o '[0-9]*[.][0-9]*' | head -1)
```
JianLi-RH, line 1056: maybe we can use global variable TARGET_VERSION or TARGET_MINOR_VERSION, because after health_check currentVersion should be same with TARGET_VERSION otherwise this step should fail.

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
```diff
@@ -998,6 +1053,13 @@ for target in "${TARGET_RELEASES[@]}"; do
...
+    currentPlugin=$(oc get network.config.openshift.io cluster -o jsonpath='{.status.networkType}')
+    if [[ ${currentVersion} == "4.16" && ${currentPlugin} == "OpenShiftSDN" ]]; then
```
JianLi-RH, line 1058: here we can check TARGET_MINOR_VERSION == "16"

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
```diff
@@ -254,6 +254,61 @@ EOF
     ansible-playbook -i "${SHARED_DIR}/ansible-hosts" /tmp/repo.yaml -vvv
 }
 
+# Do sdn migration to ovn since sdn is not supported from 4.17 version
+function sdn2ovn(){
+    TARGET=${TARGET:-OVNKubernetes}
+    timeout 600s bash <<EOT
```
JianLi-RH, line 260: if we run sdn2ovn after health_check, I think we don't need to wait co ready anymore.

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
```diff
@@ -254,6 +254,61 @@ EOF
...
+       for NODE in \$(oc get nodes -o custom-columns=NAME:.metadata.name --no-headers); do oc get node \$NODE -o yaml | grep "k8s.ovn.org/node-gateway-router-lrp-ifaddr:" | grep "100.65";  done > /dev/null && \
+       oc get network.config/cluster -o jsonpath='{.status.networkType}' | grep OVNKubernetes > /dev/null;
+    do
+       echo "Live migration is still in progress"
+       sleep 300
+    done
+EOT
+    echo "The Migration is completed"
+    # Check all ClusterOperators back to normal after live migration
+    timeout 600s bash <<EOT
```
JianLi-RH, line 298: we have similar function:
```
        if check_clusteroperators; then
            echo "Passed #${continous_successful_check}"
            (( continous_successful_check += 1 ))
        else
            echo "cluster operators are not ready yet, wait and retry..."
            continous_successful_check=0
        fi
```

JianLi-RH, line 298: we have similar function:
```
        if check_clusteroperators; then
            echo "Passed #${continous_successful_check}"
            (( continous_successful_check += 1 ))
        else
            echo "cluster operators are not ready yet, wait and retry..."
            continous_successful_check=0
        fi
```

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
zhaozhanqi: thanks @JianLi-RH 
hmm, I'm not sure the TARGET_VERSION or TARGET_MINOR_VERSION if will be changed, eg. 4.17 or 4.18.
So I get the current cluster version which should be more accurate I think.

zhaozhanqi:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.16-amd64-nightly-4.16-upgrade-from-stable-4.7-gcp-ipi-f30

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
```diff
@@ -998,6 +1053,13 @@ for target in "${TARGET_RELEASES[@]}"; do
     fi
     check_history
     health_check
+    currentVersion=$(oc version -o yaml | grep openshiftVersion | grep -o '[0-9]*[.][0-9]*' | head -1)
+    currentPlugin=$(oc get network.config.openshift.io cluster -o jsonpath='{.status.networkType}')
+    if [[ ${currentVersion} == "4.16" && ${currentPlugin} == "OpenShiftSDN" ]]; then
+	echo "the cluster is 4.16 and openshift-sdn need to migration to ovn and then upgrade if need"
+	sdn2ovn
```
jianlinliu, line 1060: Does this mean after cluster upgrade completed, need to migrate sdn to ovn ?

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
jianlinliu: agree.

jiajliu:
hi @zhaozhanqi could u share more context on this pr? If `sdn2ovn` is only needed after upgrade, sounds like a new step is more better than adding into upgrade step.
btw, i see `sdn2ovn` only added in chain upgrade, do we intend to do the migration for chain upgrade only, but not considering one hop upgrade(like 4.15-4.16).

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
zhaozhanqi: Yes,  since sdn cannot upgrade to 4.17,  so it needs to migration to ovn and then upgrade to 4.17

zhaozhanqi:
> hi @zhaozhanqi could u share more context on this pr? If `sdn2ovn` is only needed after upgrade, sounds like a new step is more better than adding into upgrade step. btw, i see `sdn2ovn` only added in chain upgrade, do we intend to do the migration for chain upgrade only, but not considering one hop upgrade(like 4.15-4.16).

For openshift-sdn cluster from 4.16, which cannot upgrade to 4.17. So for these test path from old version ( eg. 4.4/4.6 with openshift-sdn) which need to migration to OVN firstly then go on upgrading to 4.17..4.18 etc.

jianlinliu:
> For openshift-sdn cluster from 4.16, which cannot upgrade to 4.17. So for these test path from old version ( eg. 4.4/4.6 with openshift-sdn) which need to migration to OVN firstly then go on upgrading to 4.17..4.18 etc.

This is a on-purpose coverage or just update codes to make sure the job can move forwards?

If the latter one, we just need to drop the job in 4.17 ci.
If the former one, I agree with @jiajliu, we also need to take one-hope upgrade (`cucushift-upgrade-toimage`) into consideration.

zhaozhanqi:
> > For openshift-sdn cluster from 4.16, which cannot upgrade to 4.17. So for these test path from old version ( eg. 4.4/4.6 with openshift-sdn) which need to migration to OVN firstly then go on upgrading to 4.17..4.18 etc.
> 
> This is a on-purpose coverage or just update codes to make sure the job can move forwards?
> 
> If the latter one, we just need to drop the job in 4.17 ci. If the former one, I agree with @jiajliu, we also need to take one-hope upgrade (`cucushift-upgrade-toimage`) into consideration.

Yeah, actually we want to cover this 

1. Upgraded from old version cluster can do migration to ovn on 4.16
2. Cluster still can do upgrade successfully after migration to OVN. 

After talked to @jiajliu about the one hope upgrade in slack.  will update one-hope steps to coverage migration ovn

shellyyang1989:
> > > For openshift-sdn cluster from 4.16, which cannot upgrade to 4.17. So for these test path from old version ( eg. 4.4/4.6 with openshift-sdn) which need to migration to OVN firstly then go on upgrading to 4.17..4.18 etc.
> > 
> > 
> > This is a on-purpose coverage or just update codes to make sure the job can move forwards?
> > If the latter one, we just need to drop the job in 4.17 ci. If the former one, I agree with @jiajliu, we also need to take one-hope upgrade (`cucushift-upgrade-toimage`) into consideration.
> 
> Yeah, actually we want to cover this
> 
> 1. Upgraded from old version cluster can do migration to ovn on 4.16
> 2. Cluster still can do upgrade successfully after migration to OVN.
> 
> After talked to @jiajliu about the one hope upgrade in slack. will update one-hope steps to coverage migration ovn

We've dropped sdn 4.16 to 4.17 upgrade in https://github.com/openshift/release/pull/53318. So you'd like to cover the scenario that starts from 4.16 sdn, migrates to ovn, upgrades to 4.17?

zhaozhanqi:
> > > > For openshift-sdn cluster from 4.16, which cannot upgrade to 4.17. So for these test path from old version ( eg. 4.4/4.6 with openshift-sdn) which need to migration to OVN firstly then go on upgrading to 4.17..4.18 etc.
> > > 
> > > 
> > > This is a on-purpose coverage or just update codes to make sure the job can move forwards?
> > > If the latter one, we just need to drop the job in 4.17 ci. If the former one, I agree with @jiajliu, we also need to take one-hope upgrade (`cucushift-upgrade-toimage`) into consideration.
> > 
> > 
> > Yeah, actually we want to cover this
> > 
> > 1. Upgraded from old version cluster can do migration to ovn on 4.16
> > 2. Cluster still can do upgrade successfully after migration to OVN.
> > 
> > After talked to @jiajliu about the one hope upgrade in slack. will update one-hope steps to coverage migration ovn
> 
> We've dropped sdn 4.16 to 4.17 upgrade in #53318. So you'd like to cover the scenario that starts from 4.16 sdn, migrates to ovn, upgrades to 4.17?

There is one sdn profile from 4.15 upgrade to 4.16,  we can add one migration step after upgrade to 4.16 version in https://github.com/openshift/release/blob/master/ci-operator/step-registry/openshift/upgrade/qe/test/openshift-upgrade-qe-test-chain.yaml#L12. 
```
$ grep "\- as:" openshift-openshift-tests-private-release-4.16__amd64-nightly-4.16-upgrade-from-stable-4.15.yaml|grep sdn
- as: vsphere-ipi-sdn-f28
```
I know 4.15 already do not supported in fresh installation.  but sounds like there still customer use it as this SE mentioned https://issues.redhat.com/browse/SUPPORTEX-19752

shellyyang1989:
> > > > > For openshift-sdn cluster from 4.16, which cannot upgrade to 4.17. So for these test path from old version ( eg. 4.4/4.6 with openshift-sdn) which need to migration to OVN firstly then go on upgrading to 4.17..4.18 etc.
> > > > 
> > > > 
> > > > This is a on-purpose coverage or just update codes to make sure the job can move forwards?
> > > > If the latter one, we just need to drop the job in 4.17 ci. If the former one, I agree with @jiajliu, we also need to take one-hope upgrade (`cucushift-upgrade-toimage`) into consideration.
> > > 
> > > 
> > > Yeah, actually we want to cover this
> > > 
> > > 1. Upgraded from old version cluster can do migration to ovn on 4.16
> > > 2. Cluster still can do upgrade successfully after migration to OVN.
> > > 
> > > After talked to @jiajliu about the one hope upgrade in slack. will update one-hope steps to coverage migration ovn
> > 
> > 
> > We've dropped sdn 4.16 to 4.17 upgrade in #53318. So you'd like to cover the scenario that starts from 4.16 sdn, migrates to ovn, upgrades to 4.17?
> 
> There is one sdn profile from 4.15 upgrade to 4.16, we can add one migration step after upgrade to 4.16 version in https://github.com/openshift/release/blob/master/ci-operator/step-registry/openshift/upgrade/qe/test/openshift-upgrade-qe-test-chain.yaml#L12.
> 
> ```
> $ grep "\- as:" openshift-openshift-tests-private-release-4.16__amd64-nightly-4.16-upgrade-from-stable-4.15.yaml|grep sdn
> - as: vsphere-ipi-sdn-f28
> ```
> 
> I know 4.15 already do not supported in fresh installation. but sounds like there still customer use it as this SE mentioned https://issues.redhat.com/browse/SUPPORTEX-19752

If you just need the 4.15 to 4.17 coverage on sdn, then current change is able to handle it. I'm wondering do you need to cover 4.16 to 4.17 on sdn?

zhaozhanqi:
> > > > > > For openshift-sdn cluster from 4.16, which cannot upgrade to 4.17. So for these test path from old version ( eg. 4.4/4.6 with openshift-sdn) which need to migration to OVN firstly then go on upgrading to 4.17..4.18 etc.
> > > > > 
> > > > > 
> > > > > This is a on-purpose coverage or just update codes to make sure the job can move forwards?
> > > > > If the latter one, we just need to drop the job in 4.17 ci. If the former one, I agree with @jiajliu, we also need to take one-hope upgrade (`cucushift-upgrade-toimage`) into consideration.
> > > > 
> > > > 
> > > > Yeah, actually we want to cover this
> > > > 
> > > > 1. Upgraded from old version cluster can do migration to ovn on 4.16
> > > > 2. Cluster still can do upgrade successfully after migration to OVN.
> > > > 
> > > > After talked to @jiajliu about the one hope upgrade in slack. will update one-hope steps to coverage migration ovn
> > > 
> > > 
> > > We've dropped sdn 4.16 to 4.17 upgrade in #53318. So you'd like to cover the scenario that starts from 4.16 sdn, migrates to ovn, upgrades to 4.17?
> > 
> > 
> > There is one sdn profile from 4.15 upgrade to 4.16, we can add one migration step after upgrade to 4.16 version in https://github.com/openshift/release/blob/master/ci-operator/step-registry/openshift/upgrade/qe/test/openshift-upgrade-qe-test-chain.yaml#L12.
> > ```
> > $ grep "\- as:" openshift-openshift-tests-private-release-4.16__amd64-nightly-4.16-upgrade-from-stable-4.15.yaml|grep sdn
> > - as: vsphere-ipi-sdn-f28
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I know 4.15 already do not supported in fresh installation. but sounds like there still customer use it as this SE mentioned https://issues.redhat.com/browse/SUPPORTEX-19752
> 
> If you just need the 4.15 to 4.17 coverage on sdn, then current change is able to handle it. I'm wondering do you need to cover 4.16 to 4.17 on sdn?

@shellyyang1989 ok, thanks.  if this path ( fresh install with 4.15 with sdn --> 4.16 --> migration to ovn --> 4.17 ) is covered. that's should be enough.  then we don't need to update one-hope chain. @jiajliu @jianlinliu WDYT?

jiajliu:
>  then we don't need to update one-hope chain. @jiajliu @jianlinliu WDYT?

@zhaozhanqi SGTM, just like we discussed, it totally depends on the requirement. If 4.15-4.16 one-hop upgrade against `vsphere-ipi-sdn-f28` is not in your scope, then no need for `cucushift-upgrade-toimage`.

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
jianlinliu: Agree. In the toimage step, we always run health check, also including the co ready testing.

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
zhaozhanqi: updated, thanks.

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
zhaozhanqi: Ok, updated. thanks. @JianLi-RH @jianlinliu

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
zhaozhanqi: updated, thanks

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
zhaozhanqi: OK, I removed this steps. thanks.

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
zhaozhanqi: thanks, I reused `headlth_check` steps after migration.

zhaozhanqi:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.16-amd64-nightly-4.16-upgrade-from-stable-4.12-azure-ipi-sdn-f14

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
```diff
@@ -254,6 +254,36 @@ EOF
     ansible-playbook -i "${SHARED_DIR}/ansible-hosts" /tmp/repo.yaml -vvv
 }
 
+# Do sdn migration to ovn since sdn is not supported from 4.17 version
+function sdn2ovn(){
+    TARGET=${TARGET:-OVNKubernetes}
```
jianlinliu, line 259:
```suggestion
    local target_network_plugin=${target_network_plugin:-OVNKubernetes}
```
`TARGET` is already used in other code, to avoid conflicts, suggest to use a different var name.

zhaozhanqi: ah, good catch.  Sounds like `TARGET` is not used for this function,  I removed it. thanks.

PR #53474, opened 2024-06-20
stackrox: always run upgrade test

janisz:
Our data shows that upgrade test is reasonable stable. To prevent accidental breaks let's run it on every PR. Once we get more data we can mark it as required.

See:
- https://github.com/openshift/release/pull/52766


gavin-stackrox (approved):
/lgtm

PR #53499, opened 2024-06-20
Remove test job for RC trigger

calebevans:


oharan2:
Do we want to keep this job for development? I'm curious about the need for its initial use

PR #53500, opened 2024-06-20
Test Compliance Operator on ROSA HCP periodically

rhmdnd:
Now that we have CI that runs Compliance Operator testing on ROSA HCP,
let's wire it up to periodic CI so that we can monitor it over time.


Vincent056 (approved):
/lgtm

PR #53537, opened 2024-06-21
Remove openshift-ansible 3.11 gcp jobs

patrickdillon:
We have AWS jobs that were added by TRT team and can be used instead. These are broken anyway and it seems that 3.11 will be gone soon.


barbacbd (commented):
/lgtm

PR #53564, opened 2024-06-24
Adjust skipped profiles

JianLi-RH:
/cc @jianlinliu @shellyyang1989 @evakhoni @jiajliu @dis016

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/tools/get_missed_upgrade_jobs.py
+++ b/ci-operator/config/openshift/openshift-tests-private/tools/get_missed_upgrade_jobs.py
@@ -51,13 +51,13 @@ def get_missed_profiles(e2e_jobs, upgrade_jobs):
             - as: vsphere-upi-zones-f28
     """
     missed_profiles = []
-    skipped_profiles = 'disasterrecovery|longduration|destructive|hypershift|ocm|rosa|aro|ui'
+    skipped_profiles = 'agent|alibaba|aro|cloud|day2-64k-pagesize|destructive|disasterrecovery|fips-check|hive|hypershift|longduration|mco|netobserv|nokubeadmin|ocm|rhel|rosa|security-profiles-operator|ui|winc|wrs'
+    str_upgrade_jobs = ','.join(upgrade_jobs)
     for job in e2e_jobs:
         match = re.search(skipped_profiles, job) 
         if not match:
             e2e_job_name = job.replace("- as: ", '')
             e2e_job_name = re.split(r'-f\d+', e2e_job_name)[0]
-            str_upgrade_jobs = ','.join(upgrade_jobs)
     
             if e2e_job_name not in str_upgrade_jobs:
                 missed_profiles.append(job)


ci-operator/config/openshift/openshift-tests-private/tools/get_missed_upgrade_jobs.py
```diff
@@ -51,13 +51,13 @@ def get_missed_profiles(e2e_jobs, upgrade_jobs):
             - as: vsphere-upi-zones-f28
     """
     missed_profiles = []
-    skipped_profiles = 'disasterrecovery|longduration|destructive|hypershift|ocm|rosa|aro|ui'
+    skipped_profiles = 'agent|alibaba|aro|destructive|disasterrecovery|hypershift|longduration|ocm|rhel|rosa|ui|winc'
```
shellyyang1989, line 54: I'm wondering why rhel, ui and winc?

JianLi-RH: For winc: https://redhat-internal.slack.com/archives/CF8SMALS1/p1713842442500839?thread_ts=1713841950.285019&cid=CF8SMALS1

JianLi-RH: rhel: I think `rhel workers are being deprecated in 4.16` https://redhat-internal.slack.com/archives/CF8SMALS1/p1715306294842649, and this script is most likely be used in future work, so we can skip rhel in it

JianLi-RH: ui: I think the jobs which name include `ui` means they are ui team owned jobs, we don't need to add them in upgrade. need ui team to confirm. @yapei

yapei: May I know what are we going to achieve by this PR?

JianLi-RH: @yapei This script is used to find the jobs/profiles which include in e2e config file but not in upgrade config file. In other word, we want to find out the profiles which have not implemented in upgrade.

yapei: thanks for clarification @JianLi-RH 
okay to skip jobs with `ui` in it

JianLi-RH:
If there are no objections, please help review and merge this pr

shellyyang1989:
/lgtm

@liangxia PTAL. Thank you!

ci-operator/config/openshift/openshift-tests-private/tools/get_missed_upgrade_jobs.py
liangxia: Maybe also need to include cloud/mco/netobserv/hive/wrs/security-profiles-operator/fips-check/obo/day2-64k-pagesize/nokubeadmin

JianLi-RH: I have confirmed with @lihongyan1 for `obo`, it should be automated for upgrade job.

PR #53609, opened 2024-06-25
[4.16] move some disconnected ci jobs to single arch ci

jianlinliu:
multi-arch payload is much bigger than single-arch payload, it will become unstable to mirror a big size of image, so move some multi-arch disconnected ci jobs to single-arch ci.


jianlinliu:
cc @yunjiang29 @jinyunma @liangxia to review.

PR #53622, opened 2024-06-25
envoyproxy: update BC & add readme

Prucek:
Explain why we have our version of `gangway_api_descriptor.pb`.

/cc @jmguzik


jmguzik (approved):
LGTM but please open discussion with upstream how we could address this in longterm

PR #53625, opened 2024-06-25
Remove FCOS based OKD for now

ausil:
As part of the pivot to have everything CentOS stream based for now we will remove the FCOS based okd for 4.16 on. It may come back down the road as part of a community effort.


bradmwilliams:
An alternative to this would be to simply add the `"endOfLife": true,` option to the config.  This will preserve the config (for later) and tell the release-controller to ignore the inbound imagestream all together.

ausil:
I have updated to end of life them

bradmwilliams (approved):

PR #53645, opened 2024-06-25
MULTIARCH-4701 libvirt UPI workflow additional logging

tvardema:
Adding extra log lines to individual UPI steps
Also updating the install step to include the fix addressing the missing static pod controller degradation.

### DIFF
--- a/ci-operator/step-registry/ipi/conf/etcd/on-ramfs/ipi-conf-etcd-on-ramfs-commands.sh
+++ b/ci-operator/step-registry/ipi/conf/etcd/on-ramfs/ipi-conf-etcd-on-ramfs-commands.sh
@@ -6,6 +6,7 @@ set -o pipefail
 
 if [[ "${USE_RAMFS:=false}" == "true" ]]; then
 
+echo "Creating the manifest_etcd-on-ramfs-mc.yml file..."
 cat >> "${SHARED_DIR}/manifest_etcd-on-ramfs-mc.yml" << EOF
 kind: MachineConfig
 apiVersion: machineconfiguration.openshift.io/v1
--- a/ci-operator/step-registry/ipi/debug/missing-static-pod-controller-degraded/ipi-debug-missing-static-pod-controller-degraded-commands.sh
+++ b/ci-operator/step-registry/ipi/debug/missing-static-pod-controller-degraded/ipi-debug-missing-static-pod-controller-degraded-commands.sh
@@ -6,6 +6,7 @@ set -o pipefail
 
 if [[ "${STATIC_POD_CHECK_WORKAROUND:=false}" == "true" ]]; then
 
+echo "Creating the manifest_static-pod-check-workaround-master-mc.yml file..."
 #
 # This is taken from https://github.com/redhat-ztp/dev-utils/tree/static-pod-check/static-pod-check
 # Which was the answer for https://issues.redhat.com/browse/OCPBUGS-2474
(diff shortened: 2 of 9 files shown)


tvardema:
/pj-rehearse periodic-ci-openshift-multiarch-master-nightly-4.17-ocp-e2e-ovn-remote-libvirt-ppc64le

tvardema:
/pj-rehearse periodic-ci-openshift-multiarch-master-nightly-4.17-ocp-e2e-ovn-remote-libvirt-ppc64le

jaypoulz (approved):
I like the addition of extra trace logging.
I feel a little mixed on two points:
1. I don't love that we use /tmp as our install directory, since it seems to me like there's a possibility that other services contaminate /tmp. It's relatively unlikely, but are we guaranteed to have a fresh /tmp for the install pod, and is it modified between steps? What if the SHARED_DIR was actually just /tmp, could there be other files passed into this step that cause issues. I think we actually know what the shared dir is, so it's probably a non issue. I guess my preference would be that we create a directory inside tmp specific to the job, and then all commands related to the install dir use a variable like INSTALL_DIR to specify the directory. I don't know that it's necessary since our installs are *already working*, but I figured I'd explain my thinking for evaluation.
2. I'm mixed that we repeat ourselves a few times for each of the configuration manifests. For IPI, we just copy over all `.yaml` files from the SHARED_DIR. This is less explicit (and thus harder to debug), but easier to add additional steps for since you only have to create steps that add a manifest yaml to the SHARED_DIR, which I understand to be a fairly established design pattern in CI.

I don't think either of my points need to be addressed, so I'm marking as approved.

tvardema:
> 1. I don't love that we use /tmp as our install directory, since it seems to me like there's a possibility that other services contaminate /tmp. It's relatively unlikely, but are we guaranteed to have a fresh /tmp for the install pod, and is it modified between steps? What if the SHARED_DIR was actually just /tmp, could there be other files passed into this step that cause issues. I think we actually know what the shared dir is, so it's probably a non issue. I guess my preference would be that we create a directory inside tmp specific to the job, and then all commands related to the install dir use a variable like INSTALL_DIR to specify the directory. I don't know that it's necessary since our installs are _already working_, but I figured I'd explain my thinking for evaluation.
`/tmp` is "unique to the pod".  The reason I've copied content from `SHARED_DIR` to `/tmp` is to basically ensure that `SHARED_DIR` contains all the relevant historical information that happens over the entirety of the job, where `/tmp` will be used _safely_ by any individual step.  The only thing that could "contaminate" the `/tmp` directory would be done within that same `commands.sh` file.  I'm not exactly opposed to making another directory off `/tmp` to house that particular build, but its already unique to the pod, so I don't think its all that necessary.
> 2. I'm mixed that we repeat ourselves a few times for each of the configuration manifests. For IPI, we just copy over all `.yaml` files from the SHARED_DIR. This is less explicit (and thus harder to debug), but easier to add additional steps for since you only have to create steps that add a manifest yaml to the SHARED_DIR, which I understand to be a fairly established design pattern in CI.
I generally prefer to be explicit when adding configurations... I'm not sure I can guarantee that every `.yaml` or `.yml` file in the `SHARED_DIR` will be used for the cluster deployment.  What I *would* want would be to create a directory off `SHARED_DIR` where we could house those specific configs and just import from said directory.  Given the limited space and lack of `dir` support off `SHARED_DIR`, that's not currently possible.

PR #53648, opened 2024-06-25
Promote Openshift artifacts image for OKD scos

Prashanth684:
Previously, https://github.com/openshift/release/pull/52117 was introduced to create the artifacts image for OKD. Since then, it has been decided that the coreos team will work on providing an scos base image on which we will overlay the ART rpms rather than create the artifacts container ourselves. In the interim, this is a temporary solution to use Openshift's artifacts image for OKD.


soltysh (approved):
/approve

PR #53658, opened 2024-06-26
OCPQE-23189: Fix tests in aws-ipi-public-ipv4-pool-amd-fXX-destructive

liangxia:
Based on the test name, it's a destructive tests. But in the log, it shows we are executing non-destructive tests.


yunjiang29:
@liangxia thanks for the fix

PR #53663, opened 2024-06-26
decrease test frequency fo rhel worker ci jobs

jianlinliu:
Now 4.16 is close to GA, rhel workers are deprecated, so decrease the test frequency for all the rhel worker install profile. 

Also removed a UI ci job on a rhel worker install profile per the [discussion](https://redhat-internal.slack.com/archives/CPJQCSN49/p1719367274491449).


jianlinliu:
cc @gpei @yapei to review

PR #53671, opened 2024-06-26
Add keyword longduration to long duration tests

liangxia:
Those profiles are running long duration tests, but missing the keyword `longduration` in the test name to indicate that.


LiangquanLi930:
/lgtm
Thanks!

fxierh:
Thanks. 

/lgtm

PR #53683, opened 2024-06-26
[OCPQE-22132] Added not Auth tag to skip baselinecap auth test to avoid failure

dpuniaredhat:
Added a not auth tag to ensure the test does not fail due to the baselinecap profile missing image-registry capabilities, which caused failures like `dockercfg` missing secret.


xingxingxia:
Per Slack discussion, there is `skip_this_scenario` https://github.com/openshift/verification-tests/blob/16f00558205e4508fa37d0fe9211da04e154028b/features/step_definitions/storage_class.rb#L305 .

PR #53684, opened 2024-06-26
Add tools to github-ldap-user-group-creator

deepsm007:
Failing [job](https://deck-internal-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/origin-ci-private/logs/periodic-github-ldap-user-group-creator/1805947848148127744) due to missing tools

/cc @openshift/test-platform

### DIFF
--- a/ci-operator/config/openshift/ci-tools/openshift-ci-tools-master.yaml
+++ b/ci-operator/config/openshift/ci-tools/openshift-ci-tools-master.yaml
@@ -612,6 +612,7 @@ images:
   to: clusterimageset-updater
 - dockerfile_literal: |-
     FROM registry.access.redhat.com/ubi9/ubi-minimal:latest
+    RUN microdnf install -y tar gzip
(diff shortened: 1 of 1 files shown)


ci-operator/config/openshift/ci-tools/openshift-ci-tools-master.yaml
```diff
@@ -612,6 +612,7 @@ images:
   to: clusterimageset-updater
 - dockerfile_literal: |-
     FROM registry.access.redhat.com/ubi9/ubi-minimal:latest
+    RUN microdnf install -y tar gzip
```
droslean, line 615: Why this is needed? I see that the job is failing but I am falling to find where exactly the job is running tar commands.....

smg247: Ah, good to know. I also see that the only other infra-periodic that uses `tar` that way is `pr-reminder`. Perhaps we have found them all already.

PR #53698, opened 2024-06-26
aws-ipi-ovn-custom-cidr

jechen0648:


ci-operator/step-registry/ovn/conf/custom-cidr/ovn-conf-custom-cidr-ref.yaml
```diff
@@ -8,7 +8,7 @@ ref:
       memory: 100Mi
   env:
   - name: CLUSTER_NETWORK_CIDR
-    default: "10.128.0.0/20"
```
jluhrsen, line 11: not sure why you are changing this, but we will need to understand what it will affect.

Can you put more details in the commit message please?

jechen0648: I do not look to merge this PR, I only wanted to be able to pj-rehearse to get a cluster with 10.18.0.0/16 cidr.

jechen0648:
done with my need

PR #53781, opened 2024-06-28
medik8s: fix tide and cleanup

slintes:
- add missing 4.16, and all 4.17 configs
- add script for adding new OCP version to all repos for main and latest release branches
- remove configs of unsupported versions
- move tide config to org level for deduplication
- make jobs optional; they will be made required by branch protection

Reason for the latter:
Our tests use skip_if_only_changed, which makes them not required,
what we don't want. That's why we added branch protection config.
However, for tide those tests are configured twice now: as "required"
from branch protection, and "required if present" from job config. That failed
tide. To fix this, we make the jobs optional. Letting tide ignore branch protection
would be nicer, but it's not possible to override that setting.


ci-operator/config/medik8s/common/new-ocp-version.sh
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env bash
```
clobrano, line 1: wow, nice script :D

clobrano (approved):

ci-operator/config/medik8s/common/new-ocp-version.sh
slintes: thx :)

PR #53786, opened 2024-06-28
Mirror golangci-lint v1.59.1 image

danilo-gemoli:
Required for `ci/prow/lint` to pass in https://github.com/openshift/ci-tools/pull/4203.
We are upgrading to Golang `v1.22.x` in `o/ci-tools` and it seems that tools like `golangci-lint` needs to keep up with the same minor version:
```sh
$ golangci-lint --version
golangci-lint has version v1.59.1 built with go1.22.4 ...
```

The image has been mirrored like so:
```sh
$ oc image mirror \
  'docker.io/golangci/golangci-lint:v1.59.1' 'quay.io/rh_ee_dgemoli/golangci-lint:v1.59.1' \
  --keep-manifest-list=true
```

### DIFF
--- a/core-services/image-mirroring/_config.yaml
+++ b/core-services/image-mirroring/_config.yaml
@@ -637,6 +637,8 @@ supplementalCIImages:
     image: registry.k8s.io/git-sync/git-sync:v4.2.3
   ci/golangci-lint:v1.59.0:
     image: quay.io/openshift-install/golangci-lint:v1.59.0
+  ci/golangci-lint:v1.59.1:
+    image: quay.io/rh_ee_dgemoli/golangci-lint:v1.59.1
   ci/govulncheck:latest:
     namespace: ci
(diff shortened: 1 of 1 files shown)


core-services/image-mirroring/_config.yaml
```diff
@@ -637,6 +637,8 @@ supplementalCIImages:
     image: registry.k8s.io/git-sync/git-sync:v4.2.3
   ci/golangci-lint:v1.59.0:
     image: quay.io/openshift-install/golangci-lint:v1.59.0
+  ci/golang-lint:v1.59.1:
```
deepsm007, line 640:
```suggestion
  ci/golangci-lint:v1.59.1:
```

danilo-gemoli: Multitasking is killing me...

PR #53800, opened 2024-06-28
stackrox: move release-4.4 tests from ocp 4.15 to 4.16

davdhacs:
/hold
/uncc


tommartensen:
/lgtm

@davdhacs What about moving the lower bound for OCP version? 4.12 has just gone out of life last week: https://docs.openshift.com/dedicated/osd_architecture/osd_policy/osd-life-cycle.html#sd-life-cycle-dates_osd-life-cycle

tommartensen:
@davdhacs Thanks for checking!

PR #53801, opened 2024-06-28
Update golangci-lint in o/ci-tools

danilo-gemoli:
This should solve the `ci/prow/lint` failure in https://github.com/openshift/ci-tools/pull/4203.


ci-operator/config/openshift/ci-tools/openshift-ci-tools-master.yaml
```diff
@@ -10,7 +10,7 @@ base_images:
   golangci-lint:
     name: golangci-lint
     namespace: ci
-    tag: v1.54.2
+    tag: v1.59.0
```
droslean, line 13: Is this multi-arch? We can enforce the job to run on arm nodes only.

PR #53830, opened 2024-07-01
OCPQE-23300: Refine 4.14 test frequency

liangxia:
Change `f4/f7/f14/f30` to `f28` for 4.14 jobs.

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__amd64-nightly-4.14-upgrade-from-aro-4.14.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__amd64-nightly-4.14-upgrade-from-aro-4.14.yaml
@@ -34,8 +34,8 @@ resources:
       cpu: 100m
       memory: 200Mi
 tests:
-- as: aro-f30
-  cron: 28 7 4 * *
+- as: aro-f28
+  cron: 4 13 15 * *
   steps:
     cluster_profile: azure-qe
     env:
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__amd64-nightly-4.14-upgrade-from-eus-4.12.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__amd64-nightly-4.14-upgrade-from-eus-4.12.yaml
@@ -61,7 +61,7 @@ resources:
       memory: 200Mi
 tests:
 - as: azure-ipi-workers-rhel8-f28
-  cron: 35 2 6 * *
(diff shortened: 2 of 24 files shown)


jianlinliu:
I wonder why not f28, but f30 (Feb do not have 30 days).

liangxia:
> I wonder why not f28, but f30 (Feb do not have 30 days).

Both of them spread out the jobs from day1 to day30. 
For f28, the increment is 28, while for f30, the increment is 30. As we only allow one execution for f28/f30 in a month, increment 28 or 30 has the same effect.
Even we use f28 (or f30), we still have cron entry with day of 29/30.

liangxia:
Found some jobs are using f28, while several of the others are using f30. Since they have the same effect, trying to unify them into one. I'm fine to unify f28 to f30, or f30 to f28.

@jianlinliu  Let me know which one you prefer.

jianlinliu:
> Found some jobs are using f28, while several of the others are using f30. Since they have the same effect, trying to unify them into one. I'm fine to unify f28 to f30, or f30 to f28.
> 
> @jianlinliu Let me know which one you prefer.

I prefer f28.

jianlinliu:
> Both of them spread out the jobs from day1 to day30. For f28, the increment is 28, while for f30, the increment is 30. As we only allow one execution for f28/f30 in a month, increment 28 or 30 has the same effect. Even we use f28 (or f30), we still have cron entry with day of 29/30.

Take an example:
```
- as: aws-ipi-disc-priv-sts-ep-fips-f30
  cron: 33 17 30 * *
```
That means the job would never be run in Feb, right?

liangxia:
> > Both of them spread out the jobs from day1 to day30. For f28, the increment is 28, while for f30, the increment is 30. As we only allow one execution for f28/f30 in a month, increment 28 or 30 has the same effect. Even we use f28 (or f30), we still have cron entry with day of 29/30.
> 
> Take an example:
> 
> ```
> - as: aws-ipi-disc-priv-sts-ep-fips-f30
>   cron: 33 17 30 * *
> ```
> 
> That means the job would never be run in Feb, right?

As explained above, f28 and f30 make no diffences in Feb, as they have the same [issue](https://github.com/openshift/release/blob/master/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.14__amd64-nightly.yaml#L2139-L2140),
```
- as: ibmcloud-ipi-dedicatedhost-dynamic-f28-destructive
  cron: 39 4 30 * *
```

liangxia:
> > Found some jobs are using f28, while several of the others are using f30. Since they have the same effect, trying to unify them into one. I'm fine to unify f28 to f30, or f30 to f28.
> > @jianlinliu Let me know which one you prefer.
> 
> I prefer f28.

@jianlinliu   Done. Unified them to f28.

PR #53844, opened 2024-07-01
cert-rotation: run on Rocky 9

vrutkovs:
Use Rocky 9 as a default OS for cert rotation instead of specifying it every time.

This also removes 4.14/4.15 cert rotation periodics as these are permabroken


ci-operator/config/openshift-priv/cluster-kube-apiserver-operator/openshift-priv-cluster-kube-apiserver-operator-release-4.14.yaml
```diff
@@ -310,102 +310,6 @@ tests:
   steps:
     cluster_profile: azure4
     workflow: openshift-e2e-azure
-- always_run: false
```
danmanor, line 313: I believe `openshift-priv` config shouldn't be touched by a non-bot user

danmanor: https://docs.ci.openshift.org/docs/architecture/private-repositories/#openshift-priv-organization

vrutkovs: This happens due to testplatform issue where one repo blocked openshift-priv updates, so we have to roll with this update to make it pass

PR #53855, opened 2024-07-01
Add new profile: equinix-ocp-hcp

LiangquanLi930:
https://github.com/openshift/ci-tools/pull/4207
https://issues.redhat.com/browse/HOSTEDCP-1805


_boskos.yaml
```diff
@@ -0,0 +1,2951 @@
+# generated with generate-boskos.py; do not edit directly
```
smg247: This file is added in error

LiangquanLi930: Thanks, I noticed it, I just changed the code

PR #53873, opened 2024-07-01
[azure-workload-identity] rename Dockerfile

ashwindasr:
Alongside https://github.com/openshift/azure-workload-identity/pull/23


ashwindasr:
/hold

Needs to merge with https://github.com/openshift/azure-workload-identity/pull/23

PR #53882, opened 2024-07-02
[4.13] move some disconnected ci jobs to single arch ci

jianlinliu:
1. [4.13] multi-arch payload is much bigger than single-arch payload, it will become unstable to mirror a big size of image, so move some multi-arch disconnected ci jobs to single-arch ci.
2. In https://github.com/openshift/release/pull/53828, forget to drop the migrated ci jobs in 4.14, so drop them here.
3. Remove remove redundant tp jobs.


jianlinliu:
cc @yunjiang29 and @jinyunma to review.

PR #53888, opened 2024-07-02
OCPQE-23341: Add fips to indicate fips testing

liangxia:
`nutanix-ipi-sdn-live-migration-ovn` set `FIPS_ENABLED: "true"`, but does not indicate fips state in test name.


liangxia:
Rebase and resolve conflicts.

PR #53898, opened 2024-07-02
Remove some exceptions granted for cluster pools

jmguzik:


core-services/cluster-pools/_config.yaml
```diff
@@ -9,31 +9,12 @@
...
         - acs-fleet-manager
-    - org: red-hat-storage #exception granted
-      repos:
-        - mcg-ms-console
-        - mcg-osd-deployer
-        - ocs-operator
-        - odf-console
-        - odf-multicluster-orchestrator
-        - odf-operator
     - org: codeready-toolchain #exception granted
```
jmguzik, line 12: maybe this can be removed as well, i will create a separate pr to check

PR #53900, opened 2024-07-02
odf: add ibm_storage to allowed Bugzilla groups

nixpanic:
The  OpenShift bugzilla bot reports problems like this for some bugs:

```
Bugzilla bug ... is in a bug group that is not in the allowed groups for this repo.
Allowed groups for this repo are:

   - qe_staff
   - redhat
```

This happens when the bug has the `ibm_storage` group added, which is pretty much the default for bugs against the ODF product (all engineers working on the product are part of IBM).

/assign agarwal-mudit


agarwal-mudit (approved):
/lgtm

PR #53915, opened 2024-07-02
upgrade status should only log yet unseen errors during upgrade

evakhoni:
follow up [slack thread](https://redhat-internal.slack.com/archives/CF8SMALS1/p1719889316376549) the recent updates made to upgrade status command, broke the "log only errors" filter in prow upgrade testing. this PR is addressing the issue, while also implementing a better filtering to minimize the repeating of already seen errors.

/hold
while testing, also needs porting to chainupgrade and maybe elsewhere

### DIFF
--- a/ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
+++ b/ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
@@ -342,18 +342,28 @@ function upgrade() {
 
 # Monitor the upgrade status
 function check_upgrade_status() {
-    local wait_upgrade="${TIMEOUT}" out avail progress cluster_version stat oldstat='empty' filter='[0-9]+h|[0-9]+m|[0-9]+s|\s+|\n'
+    local wait_upgrade="${TIMEOUT}" interval=1 out avail progress cluster_version stat_cmd stat='empty' oldstat='empty' filter='[0-9]+h|[0-9]+m|[0-9]+s|[0-9]+%|[0-9]+.[0-9]+s|[0-9]+ of|\s+|\n'
     if [[ -n "${1:-}" ]]; then
         cluster_version="$1"
     else
         cluster_version="${TARGET_VERSION}"
     fi
     echo "Starting the upgrade checking on $(date "+%F %T")"
+    # print once to log (including full messages)
+    oc adm upgrade || true
+    # log oc adm upgrade (excluding garbage messages)
+    stat_cmd="oc adm upgrade | grep -vE 'Upstream is unset|Upstream: https|available channels|No updates available|^$'"
+    # if available (version 4.16+) log "upgrade status" instead
+    if [[ -n "${TARGET_MINOR_VERSION}" ]] && [[ "${TARGET_MINOR_VERSION}" -ge "16" ]] ; then
+        stat_cmd="env OC_ENABLE_CMD_UPGRADE_STATUS=true oc adm upgrade status 2>&1 | grep -vE 'no token is currently in use|for additional description and links'"
+    fi
     while (( wait_upgrade > 0 )); do
-        sleep 5m
-        wait_upgrade=$(( wait_upgrade - 5 ))
-        if ! ( run_command "oc adm upgrade" ); then
-            continue
+        sleep ${interval}m
+        wait_upgrade=$(( wait_upgrade - interval ))
+        # if output is different from previous (ignoring irrelevant time/percentage difference), write to log
+        if stat="$(eval "${stat_cmd}")" && [ -n "$stat" ] && ! diff -qw <(sed -zE "s/${filter}//g" <<< "${stat}") <(sed -zE "s/${filter}//g" <<< "${oldstat}") >/dev/null ; then
+            echo -e "=== Upgrade Status $(date "+%T") ===\n${stat}\n\n\n\n"
+            oldstat=${stat}
         fi
         if ! out="$(oc get clusterversion --no-headers || false)"; then
             echo "Error occurred when getting clusterversion"
@@ -369,18 +379,6 @@ function check_upgrade_status() {
             UPGRADE_RHEL_WORKER_BEFOREHAND="triggered"
             echo -e "Upgrade stuck at updating RHEL worker, need to run the RHEL worker upgrade later...\n\n"
             return 0
-        fi
-        # assuming oc to equal TARGET_MINOR_VERSION, log abnormal "upgrade status" if available (version 4.16+)
-        if [[ -n "${TARGET_MINOR_VERSION}" ]] && [[ "${TARGET_MINOR_VERSION}" -ge "16" ]] ; then
-            # capture "update health" section of status, ignoring the title and the last line
-            # whenever an error is present, "Run with --details=health" is the last line we should remove.
-            # with no error, it'll remove "Update is proceeding well" resulting in empty $stat that we simply ignore
-            stat="$(env OC_ENABLE_CMD_UPGRADE_STATUS=true oc adm upgrade status 2>&1 | sed -ne "/= Update Health =/,$ p" | sed '2d;$d' || true)"
-            # if update health messages exist, and the mesage is different from previous (ignoring time difference using $filter), dump "Update Health" to log
-            if [ -n "$stat" ] && ! diff -qw <(sed -zE "s/${filter}//g" <<< "${stat}") <(sed -zE "s/${filter}//g" <<< "${oldstat}") >/dev/null ; then
-                echo "${stat}"
-                oldstat=${stat}
-            fi
         fi
     done
     if [[ ${wait_upgrade} -le 0 ]]; then
--- a/ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
+++ b/ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
@@ -684,28 +684,25 @@ function upgrade() {
     echo "Upgrading cluster to ${TARGET} gets started..."
 }
 
-# Log abnormal cluster status
-function dump_status_if_unexpected() {
-    # expecting oc to equal TARGET_MINOR_VERSION, skip if less than .16
-        if [[ -n "${TARGET_MINOR_VERSION}" ]] && [[ "${TARGET_MINOR_VERSION}" -ge "16" ]] ; then
-            local out; out="$(env OC_ENABLE_CMD_UPGRADE_STATUS=true oc adm upgrade status 2>&1 || true)"
-            # if upgrading, and not progressing well, dump status to log
-            # "resource name may not be empty" is a known issue, remove once OCPBUGS-32682 is fixed
-            if ! grep -qE 'The cluster version is not updating|Upgrade is proceeding well|resource name may not be empty' <<< "${out}" ; then
-                echo "${out}"
-            fi
-        fi
-}
-
 # Monitor the upgrade status
 function check_upgrade_status() {
-    local wait_upgrade="${TIMEOUT}" out avail progress
+    local wait_upgrade="${TIMEOUT}" interval=1 out avail progress stat_cmd stat='empty' oldstat='empty' filter='[0-9]+h|[0-9]+m|[0-9]+s|[0-9]+%|[0-9]+.[0-9]+s|[0-9]+ of|\s+|\n'
     echo "Starting the upgrade checking on $(date "+%F %T")"
+    # print once to log (including full messages)
+    oc adm upgrade || true
+    # log oc adm upgrade (excluding garbage messages)
+    stat_cmd="oc adm upgrade | grep -vE 'Upstream is unset|Upstream: https|available channels|No updates available|^$'"
+    # if available (version 4.16+) log "upgrade status" instead
+    if [[ -n "${TARGET_MINOR_VERSION}" ]] && [[ "${TARGET_MINOR_VERSION}" -ge "16" ]] ; then
+        stat_cmd="env OC_ENABLE_CMD_UPGRADE_STATUS=true oc adm upgrade status 2>&1 | grep -vE 'no token is currently in use|for additional description and links'"
+    fi
     while (( wait_upgrade > 0 )); do
-        sleep 5m
-        wait_upgrade=$(( wait_upgrade - 5 ))
-        if ! ( run_command "oc get clusterversion" ); then
-            continue
+        sleep ${interval}m
+        wait_upgrade=$(( wait_upgrade - interval ))
+        # if output is different from previous (ignoring irrelevant time/percentage difference), write to log
+        if stat="$(eval "${stat_cmd}")" && [ -n "$stat" ] && ! diff -qw <(sed -zE "s/${filter}//g" <<< "${stat}") <(sed -zE "s/${filter}//g" <<< "${oldstat}") >/dev/null ; then
+            echo -e "=== Upgrade Status $(date "+%T") ===\n${stat}\n\n\n\n"
+            oldstat=${stat}
         fi
         if ! out="$(oc get clusterversion --no-headers || false)"; then
             echo "Error occurred when getting clusterversion"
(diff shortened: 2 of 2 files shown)


evakhoni:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.17-arm64-stable-4.17-upgrade-from-stable-4.16-aws-ipi-proxy-sts-f28

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -383,7 +370,18 @@ function check_upgrade_status() {
...
+        # assuming oc to equal TARGET_MINOR_VERSION, log abnormal "upgrade status" if available (version 4.16+)
+        if [[ -n "${TARGET_MINOR_VERSION}" ]] && [[ "${TARGET_MINOR_VERSION}" -ge "16" ]] ; then
+            # capture "update health" section of status, ignoring the title and the last line
+            # whenever an error is present, "Run with --details=health" is the last line we should remove.
+            # with no error, it'll remove "Update is proceeding well" resulting in empty $stat that we simply ignore
+            stat="$(env OC_ENABLE_CMD_UPGRADE_STATUS=true oc adm upgrade status 2>&1 | sed -ne "/= Update Health =/,$ p" | sed '1d;2d;$d' || true)"
+            # if update health messages exist, and the mesage is different from previous (ignoring time difference using $filter), dump "Update Health" to log
+            if [ -n "$stat" ] && ! diff -qw <(sed -zE "s/${filter}//g" <<< "${stat}") <(sed -zE "s/${filter}//g" <<< "${oldstat}") >/dev/null ; then
+                echo "${stat}"
+                oldstat=${stat}
```
jianlinliu, line 382: In the [log](https://gcsweb-qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/qe-private-deck/pr-logs/pull/openshift_release/53915/rehearse-53915-periodic-ci-openshift-openshift-tests-private-release-4.17-arm64-stable-4.17-upgrade-from-stable-4.16-aws-ipi-proxy-sts-f28/1808142830237913088/artifacts/aws-ipi-proxy-sts-f28/cucushift-upgrade-toimage/build-log.txt), there is a shell syntax issue.
```
environment: line 381: oldstat: unbound variable
```

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -383,7 +370,18 @@ function check_upgrade_status() {
...
-        dump_status_if_unexpected
+        # assuming oc to equal TARGET_MINOR_VERSION, log abnormal "upgrade status" if available (version 4.16+)
+        if [[ -n "${TARGET_MINOR_VERSION}" ]] && [[ "${TARGET_MINOR_VERSION}" -ge "16" ]] ; then
+            # capture "update health" section of status, ignoring the title and the last line
+            # whenever an error is present, "Run with --details=health" is the last line we should remove.
+            # with no error, it'll remove "Update is proceeding well" resulting in empty $stat that we simply ignore
+            stat="$(env OC_ENABLE_CMD_UPGRADE_STATUS=true oc adm upgrade status 2>&1 | sed -ne "/= Update Health =/,$ p" | sed '1d;2d;$d' || true)"
+            # if update health messages exist, and the mesage is different from previous (ignoring time difference using $filter), dump "Update Health" to log
+            if [ -n "$stat" ] && ! diff -qw <(sed -zE "s/${filter}//g" <<< "${stat}") <(sed -zE "s/${filter}//g" <<< "${oldstat}") >/dev/null ; then
+                echo "${stat}"
```
jianlinliu, line 381: For better understanding where the message on the [log](https://gcsweb-qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/qe-private-deck/pr-logs/pull/openshift_release/53915/rehearse-53915-periodic-ci-openshift-openshift-tests-private-release-4.17-arm64-stable-4.17-upgrade-from-stable-4.16-aws-ipi-proxy-sts-f28/1808142830237913088/artifacts/aws-ipi-proxy-sts-f28/cucushift-upgrade-toimage/build-log.txt) is from, shall we add some prompt message to indicated that before printing out them ?

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -383,7 +370,18 @@ function check_upgrade_status() {
             echo -e "Upgrade stuck at updating RHEL worker, need to run the RHEL worker upgrade later...\n\n"
             return 0
         fi
-        dump_status_if_unexpected
+        # assuming oc to equal TARGET_MINOR_VERSION, log abnormal "upgrade status" if available (version 4.16+)
+        if [[ -n "${TARGET_MINOR_VERSION}" ]] && [[ "${TARGET_MINOR_VERSION}" -ge "16" ]] ; then
+            # capture "update health" section of status, ignoring the title and the last line
+            # whenever an error is present, "Run with --details=health" is the last line we should remove.
+            # with no error, it'll remove "Update is proceeding well" resulting in empty $stat that we simply ignore
+            stat="$(env OC_ENABLE_CMD_UPGRADE_STATUS=true oc adm upgrade status 2>&1 | sed -ne "/= Update Health =/,$ p" | sed '1d;2d;$d' || true)"
```
jianlinliu, line 378: I have some concerns of only printing unhealthy message, sometimes the whole output from `oc adm upgrade status` can help us find out some useful info, for example, https://issues.redhat.com/browse/OCPQE-22351 was found when checking the complete output of `oc adm upgrade status`

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
evakhoni: good catch. it is basically unbound for the first use because it still haven't been set to $stat yet. I"m fixing this by assigning an initial value to it, shouldn't matter which as "the first time" we see any value in stat we can print it safely.

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
evakhoni: this one is a dilemma. on one side, having a title to each status message sounds like a more readable log, however on the other hand, status update health messages are easily recognizable because they are table structured as such
```
1h4m23s   Error     API Availability   Cluster Operator authentication is degraded (OAuthServerConfigObservation_Error)
6m37s     Warning   Update Stalled     Cluster Version version is failing to proceed with the update (ClusterOperatorDegraded)
-         Warning   Update Stalled     Outdated nodes in a paused pool 'mcpfoo' will not be updated
```
since it can be a number of lines, adding an inline title will look wrong, then adding a title on a separate line is an option, however if there's multiple error messages printed throughout the upgrade, that title will also appear multiple times, making the log longer to read, while providing no additional value.  
another idea however, since both "oc adm upgrade" and "upgrade status" are running one after the other at the same time, we can change instead of printing the current `Running command: oc adm upgrade` as generated by `run_command`, replace by something like : echo "Running oc adm upgrade and upgrade status (filtered)"

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
evakhoni: that one is a dilemma as well. firstly, after upgrade failure, `function debug` will print the full version, +including details.   
during upgrade, we can print either the full long message, or the Update Health. both options have pros and cons, so I have asked [in the thread](https://redhat-internal.slack.com/archives/CF8SMALS1/p1719907992107919?thread_ts=1719889316.376549&cid=CF8SMALS1) for our teammates opinion, and both @JianLi-RH as well as @jiajliu commented they prefer shorter log with only "Update Health" section

jianlinliu: Actually I am thinking if we can  increase the intervals to save some output, but keep a competed output from the command when it was executed, also adding some empty lines as separators to make the log easier to read.

evakhoni: good point about monitoring functionality. this was the original intention (alongside obviously providing insights) to implement upgrade status into the CI. however as for increasing the interval, this would have an obvious negative effect of missing certain temporary conditions, because upgrade status does not save information but only prints live

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -365,7 +352,7 @@ function check_upgrade_status() {
     while (( wait_upgrade > 0 )); do
         sleep 5m
         wait_upgrade=$(( wait_upgrade - 5 ))
-        if ! ( run_command "oc get clusterversion" ); then
+        if ! ( run_command "oc adm upgrade" ); then
```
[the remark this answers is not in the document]
jianlinliu: Actually I am thinking the original purpose of the line, per my understanding, its initial purpose is to help users can easily get the current cluster version to help debug purpose. So actually I do not think we need to change it. Beside that, if we decide always print out more details info from `oc adm upgrade status`, then we can skip the command (on old version, we need to keep to print out cluster version, but not `oc adm upgrade status`)

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
evakhoni: Imo, since we're anyway skipping the happy state status of "Update is proceeding well" as well as the duplicates, so a log being too long should not be a concern after this PR, whatever we decide to print the full message or the health only, in any case should be a good idea to actually decrease intervals!

when I tested the status command while working at the epic, I even used a zero interval, but a bit more aggressive filter to ignore all time and percentage differences, and ended up with acceptable sized log.

jianlinliu: For me, the original full output from `oc adm upgrade status` is not a pain point for me, after all, I used the output (even it is a healthy message) to find out some automation issues. And standing from checking the functionality of the command, I prefer to keep a full output. From my side, I think the only enhancement is to make the whole log output easier to read, but not limit the log's length.

JianLi-RH: > From my side, I think the only enhancement is to make the whole log output easier to read, but not limit the log's length. cc @JianLi-RH to think about it.

Agree with @jianlinliu , we can add separators between each upgrade status output.

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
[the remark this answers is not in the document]
evakhoni: ok I solved it by printing the full upstream/channel/.. message one time, but not repeating it later. this way we have all the needed info in log, but still readable and no garbage.

evakhoni:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.16-amd64-nightly-4.16-upgrade-from-stable-4.15-aws-ipi-shared-vpc-phz-f14

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -340,33 +340,31 @@ function upgrade() {
...
     else
         cluster_version="${TARGET_VERSION}"
     fi
     echo "Starting the upgrade checking on $(date "+%F %T")"
+    # print once to log (including full messages)
+    oc adm upgrade || true
+    # log oc adm upgrade (excluding garbage messages)
+    stat_cmd="oc adm upgrade | grep -vE 'Upstream is unset|Upstream: https|available channels|No updates available|^$'"
+    # if available (version 4.16+) log "upgrade status" instead
+    if [[ -n "${TARGET_MINOR_VERSION}" ]] && [[ "${TARGET_MINOR_VERSION}" -ge "16" ]] ; then
```
jianlinliu, line 357: If the token issue get fixed, we need to update the condition again, right?

evakhoni: as far as I know, we don't. I'm only skipping the error if its there.

jianlinliu: So if the issue is supposed to be fixed, but if regression happen again, we would not catch it by watching logs, right?

evakhoni: well, its not exactly an "issue" that planned to be "fixed". its just a limitation of how our CI works (have no token by default) vs how the "current" tech preview upgrade status implementation (built inside "oc" so cannot workaround the "no token" situation).  
next version of upgrade status will be built inside cvo, so no more token requirement to display "alerts"

evakhoni: on a side note, for catching regressions of that sort, we have test cases specifically for upgrade status. the CI implementation for this command is more of a "usability testing", feedback collecting, and trying to be helpful in ci watching

jianlinliu: Got it.

jianlinliu:
btw, shall we also need to update `cucushift-chainupgrade-toimage` step?

evakhoni:
> btw, shall we also need to update `cucushift-chainupgrade-toimage` step?

that's correct! just want to ensure current implementation works as expected (I'll test on both >= 4.16 as well as <4.16) then I'll go ahead and copy + unhold

btw, do you think any other step also need an update? `cucushift-upgrade-toversion`? something else?

jianlinliu:
> btw, do you think any other step also need an update? `cucushift-upgrade-toversion`? something else?

I think no others. Last time, we only enabled status command on the two steps, this time, we are following the same rule.

evakhoni:
> /pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.16-amd64-nightly-4.16-upgrade-from-stable-4.15-aws-ipi-shared-vpc-phz-f14

 ```exec /tools/entrypoint: argument list too long```
:man_facepalming: 

/pj-rehearse cancel

evakhoni:
ok, 4.16+ [looks nice](https://gcsweb-qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/qe-private-deck/pr-logs/pull/openshift_release/53915/rehearse-53915-periodic-ci-openshift-openshift-tests-private-release-4.17-arm64-stable-4.17-upgrade-from-stable-4.16-aws-ipi-proxy-sts-f28/1808874303303716864/artifacts/aws-ipi-proxy-sts-f28/cucushift-upgrade-toimage/build-log.txt) 

testing <4.16 ...
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-arm64-stable-4.15-upgrade-from-stable-4.14-aws-ipi-proxy-sts-f28

evakhoni:
ah what a shame pushing `cucushift-chainupgrade-toimage` canceled the last cucushift-upgrade rehearse I that just started :angry:   
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-arm64-stable-4.15-upgrade-from-stable-4.14-aws-ipi-proxy-sts-f28  
this one is for <4.16, for old clusters without explicit upgrade status command. if it will end up looking good as well, then pr is ready to merge   
/unhold

evakhoni:
/cc @dis016   
for awareness. you will notice this change when watching CI

evakhoni:
ok, <4.16 rehearsal also worked as expected
/pj-rehearse ack

evakhoni:
core-valid failure looks orthogonal  
/retest

PR #53924, opened 2024-07-02
Bump dev-scripts serial job timouts to 5h

derekhiggins:
I missed this in  46894c04de179e12efa1748459e3aa21a45f0918


elfosardo:
@derekhiggins looks like we passed the timeout, I would be ok adding the rehearsals-ack

PR #53933, opened 2024-07-02
[coredns] rename Dockerfile

ashwindasr:
Alongside https://github.com/openshift/coredns/pull/124


ashwindasr:
/close

Image owners are working on a rename, and will let ART know once they are ready.

PR #53941, opened 2024-07-02
e2e:hypershift: add KUBECONFIG env

Tal-or:
It's convenient to have `KUBECONFIG` env set as the hosted cluster kubeconfig path, because the test tries to fetch the rest.Config:
- https://github.com/openshift/cluster-node-tuning-operator/blob/master/test/e2e/performanceprofile/functests/utils/client/clients.go#L140
- https://github.com/openshift/cluster-node-tuning-operator/blob/master/test/e2e/performanceprofile/functests/utils/pods/pods.go#L155

for the hostedcluster, so by setting the KUBECONFIG we won't have to handle separate flow for hypershift and non-hypershift in the code.

We keep the old `HYPERSHIFT_HOSTED_CLUSTER_KUBECONFIG` env for backward compatability, until we'll update the test code.


rbaturov:
This looks like a good idea to me.
/lgtm

Tal-or:
/pj-rehearse pull-ci-openshift-cluster-node-tuning-operator-master-e2e-hypershift-pao

PR #53944, opened 2024-07-02
Populate $SHARED_DIR/operator-versions with the versions of installed operators

calebevans:
These values will eventually be used to add operator versions as labels in firewatch tickets.

### DIFF
--- a/ci-operator/step-registry/operator/install/operator-install-commands.sh
+++ b/ci-operator/step-registry/operator/install/operator-install-commands.sh
@@ -6,12 +6,16 @@ set -o pipefail
 set -o verbose
 
 export KUBECONFIG=${SHARED_DIR}/kubeconfig
-
+operator_configs=$(env | grep -E '^OPERATOR[0-9]+_CONFIG' | sort --version-sort)
 RUN_COMMAND="poetry run python ocp_addons_operators_cli/cli.py --action install --kubeconfig ${KUBECONFIG} "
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/operator/install/operator-install-commands.sh
```diff
@@ -47,3 +47,12 @@ if [ "${INSTALL_FROM_IIB}" = "true" ]; then
 fi
 
 ${RUN_COMMAND}
+
+for operator_value in $(env | grep -E '^OPERATOR[0-9]+_CONFIG' | sort --version-sort); do
```
rnetser, line 51: please save `$(env | grep -E '^OPERATOR[0-9]+_CONFIG' | sort --version-sort)` and re-use

ci-operator/step-registry/operator/install/operator-install-commands.sh
```diff
@@ -47,3 +47,12 @@ if [ "${INSTALL_FROM_IIB}" = "true" ]; then
...
+    operator_config=$(echo "$operator_value" | sed -E 's/^OPERATOR[0-9]+_CONFIG=//')
```
rnetser, line 52: duplicate code; please extract to a function

rnetser (changes requested):

rnetser (approved):

PR #53987, opened 2024-07-03
add aos-art-psi to pull from quay-proxy

ashwindasr:
To allow ART members to pull from quay-proxy


ashwindasr:
/close

Looks like `podman login -u=$(oc whoami) -p=$(oc whoami -t) [quay-proxy.ci.openshift.org](http://quay-proxy.ci.openshift.org/)` works without this PR, which is all we need for now.

PR #54008, opened 2024-07-03
move more jobs from vsphere-8-vpn to elastic

jcpowermac:


jcpowermac:
/hold 

until after we have enough networks

PR #54069, opened 2024-07-05
lightspeed-operator: export manager logs when e2e test fails

raptorsun:
We are using `hypershift-hostedcluster-workflow` for `e2e-bundle` test. This flow does not have the must-gather or must-gather-extra collected. 
This PR allow us to put the operator log into the stdout so that we can check the cause of failure.


tisnik (commented):
looks ok

PR #54075, opened 2024-07-05
[ptp-operator] Rename dockerfiles

ashwindasr:
For https://github.com/openshift/ptp-operator/pull/500

Unsure whether additional files need to change here, since two components build out from the same master repo


jzding:
need to change this for 4.17 and 4.18 files too

jzding (approved):
/lgtm

PR #54079, opened 2024-07-05
[sriov-network-operator] Rename dockerfile

ashwindasr:
For https://github.com/openshift/sriov-network-operator/pull/955

Not sure if additional files need to change here since more than one dockerfile names were changed, please verify


ashwindasr:
Thanks @zeeke, I've updated and repushed. ptal when you get the chance

PR #54092, opened 2024-07-08
gangway: revert envoy to stable

Prucek:


psalajova:
@Prucek is this PR still relevant?

PR #54112, opened 2024-07-08
[MSINTEROP-1167] Add step to populate additional labels for product to product scenarios

calebevans:


ci-operator/step-registry/p2p/additional-labels/p2p-additional-labels-commands.sh
```diff
@@ -0,0 +1,15 @@
...
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+# Login to cluster
+eval "$(cat "${SHARED_DIR}/api.login")"
+
+# Populate operator version labels
+cat "${SHARED_DIR}/operator-versions" >> "${SHARED_DIR}/firewatch-additional-labels"
```
chetna14manku, line 11: How about adding condition here to check if file `${SHARED_DIR}/operator-versions` exists? otherwise it would throw error in firewatch step I guess.

chetna14manku: Incase operator-install step didn't run and job failed before that.

calebevans: Agreed, fixed

PR #54113, opened 2024-07-08
OCM-9426 | RHCS: Increase day1 supplemental timeout

radtriste:
https://issues.redhat.com/browse/OCM-9426

### DIFF
--- a/ci-operator/step-registry/rhcs/e2e/general-tests/rhcs-e2e-general-tests-ref.yaml
+++ b/ci-operator/step-registry/rhcs/e2e/general-tests/rhcs-e2e-general-tests-ref.yaml
@@ -7,7 +7,7 @@ ref:
     requests:
       cpu: 100m
       memory: 300Mi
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/rhcs/e2e/general-tests/rhcs-e2e-general-tests-ref.yaml
```diff
@@ -7,7 +7,7 @@ ref:
     requests:
       cpu: 100m
       memory: 300Mi
-  timeout: 3h0m0s
+  timeout: 6h0m0s
```
yasun1, line 10: The [timeout](https://github.com/openshift/release/blob/master/ci-operator/step-registry/rhcs/e2e/general-tests/rhcs-e2e-general-tests-commands.sh#L77) in the `ginkgo` command is set to `2h`. Since you are extending the running time, I think you may also need to extend it.

PR #53783, opened 2024-06-28
ART-9391 Bump 4.13 rhel8 repos

locriandev:

PR #54120, opened 2024-07-08
ART-9391 Bump 4.17 RHEL8 repos to 8.10

sdodson:


core-services/release-controller/_repos/ocp-4.17-rhel86.repo
```diff
@@ -1,6 +1,6 @@
 [rhel-8.6-baseos]
 name = rhel-8.6-baseos
-baseurl = https://cdn.redhat.com/content/eus/rhel8/8.6/x86_64/baseos/os/
+baseurl = https://cdn.redhat.com/content/dist/rhel8/8/x86_64/baseos/os/
```
locriandev, line 3: IIUC, https://cdn.redhat.com/content/dist/rhel8/8 points at rhel 8.10. If this is true, having this file named after 8.6 might be confusing. That's why I've decided to rename it to 8.10 in places like [this](https://github.com/openshift/release/pull/53783/files#diff-1ba81807ea4e5ace5a20667f4a736ca11e50ad8023b1df9b6d8d345c59287e7d).

PR #54157, opened 2024-07-09
OSASINFRA-3523: Disable ShiftStack CI hwoffload periodic jobs

MiguelCarpio:
Based on the fact that "RHOSO 18 GA will ship without HWOL or VDPA support", hwoffload ShiftStack periodic jobs will be disabled in this PR. 

This is the first step to turn off hwoffload jobs, which consists only of removing the periodic jobs from the CI config. The involved jobs are the following:

e2e-openstack-nfv-hwoffload in:

- shiftstack-ci-release-4.17.yaml
- shiftstack-ci-release-4.16.yaml
- shiftstack-ci-release-4.15.yaml
- shiftstack-ci-release-4.14.yaml
- shiftstack-ci-release-4.13.yaml
- shiftstack-ci-release-4.12.yaml


EmilienM:
/lgtm
I'm fine with this change for now, we don't have known customers using that feature anyway.

PR #54170, opened 2024-07-09
Replace wait_timeout with expected_recovery_time

chaitanyaenr:
This will enable user to track regressions in terms of pod not recovering within expected recovery time.


paigerube14 (approved):
/lgtm

PR #54182, opened 2024-07-10
allow skip on success for 4.12 auto release jobs

rioliu-rh:
add control flag `allow_skip_on_success` for 4.12 auto release jobs to skip data gathering steps when job status is success


rioliu-rh:
as I know, this control flag does not take effect on rehearse job, correct me if I'm wrong

PR #54195, opened 2024-07-10
[azure] fix cert issue when using azure-cli on ash

jinyunma:
Failed job: https://qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/qe-private-deck/logs/periodic-ci-openshift-openshift-tests-private-release-4.17-amd64-nightly-azure-stack-upi-f28-destructive/1810865649568714752

Error:
```
INFO[2024-07-10T02:47:37Z] Logs for container test in pod azure-stack-upi-f28-destructive-upi-install-azurestack: 
INFO[2024-07-10T02:47:37Z] ~/installer /output
cat: '/usr/lib64/az/lib/python*/site-packages/certifi/cacert.pem': No such file or directory 
```

The method to install azure-cli in upi-installer image is changed via https://github.com/openshift/installer/pull/8643
seems copying az-cli from quay.io/ocp-splat/pwsh:v7.3.12, and version does not specify to 2.49.0.
current version:
```
$ az version
{
  "azure-cli": "2.61.0.post20240516044921",
  "azure-cli-core": "2.61.0.post20240516044921",
  "azure-cli-telemetry": "1.1.0.post20240516044921",
  "extensions": {}
}
```

### DIFF
--- a/ci-operator/step-registry/ipi/conf/azurestack/ipi-conf-azurestack-commands.sh
+++ b/ci-operator/step-registry/ipi/conf/azurestack/ipi-conf-azurestack-commands.sh
@@ -62,9 +62,16 @@ TENANT_ID=$(jq -r .tenantId "${SHARED_DIR}/osServicePrincipal.json")
 cat >> "${SHARED_DIR}/azurestack-login-script.sh" << EOF
 
 if [[ -f "${CLUSTER_PROFILE_DIR}/ca.pem" ]]; then
-  cp "${CLUSTER_PROFILE_DIR}/ca.pem" /tmp/ca.pem
-  cat /usr/lib64/az/lib/python*/site-packages/certifi/cacert.pem >> /tmp/ca.pem
-  export REQUESTS_CA_BUNDLE=/tmp/ca.pem
+    cp "${CLUSTER_PROFILE_DIR}/ca.pem" /tmp/ca.pem
+    if ls /usr/lib64/az/lib/python*/site-packages/certifi/cacert.pem > /dev/null 2>&1; then
+        cat /usr/lib64/az/lib/python*/site-packages/certifi/cacert.pem >> /tmp/ca.pem
+    elif ls /go/src/github.com/openshift/installer/azure-cli/lib64/python*/site-packages/certifi/cacert.pem > /dev/null 2>&1; then
+        cat /go/src/github.com/openshift/installer/azure-cli/lib64/python*/site-packages/certifi/cacert.pem >> /tmp/ca.pem
+    else
+        echo "ERROR: unable to find cacert.pem in pyhton library that azure-cli depends on, exit..."
+        exit 1
+    fi
+    export REQUESTS_CA_BUNDLE=/tmp/ca.pem
 fi
 az cloud register \
     -n ${cloud_name} \


jinyunma:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.17-amd64-nightly-azure-stack-upi-f28

jinyunma:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.17-amd64-nightly-azure-stack-upi-f28

jinyunma:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.17-amd64-nightly-azure-stack-upi-f28

jinyunma:
cc @jianlinliu @gpei to review

checked [job](https://qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/qe-private-deck/pr-logs/pull/openshift_release/54195/rehearse-54195-periodic-ci-openshift-openshift-tests-private-release-4.17-amd64-nightly-azure-stack-upi-f28/1810967271162515456),  `az login` was running successfully.

Regarding azure cli version, only see the change on 4.17 via https://github.com/openshift/installer/pull/8643, so on old ocp version, still use az-2.49.0
1. maybe we need to verify latest az cli and then update upi-installer image on older branch to always install latest one? 
2. verify and update all steps(except upi-install-azure and upi-install-azurestack) which have az cli to use 4.17 upi-installer image?
or any other suggestion?

jianlinliu:
If the old version of az cli in upi-instealler image still works, then I would like to keep untouch with them.

I remember there is some reason we locked the az cli version in upi-installer image.

From  https://github.com/openshift/installer/pull/8643, actually the main purpose is only for installing pwsh, I have no idea why it changed the way of installing az cli, maybe we can talk with the author the PR to see if we can rollback the old way of installing az cli. After all, if we keep to install az-cli from pwsh image, in some day, we would hit some unexpected az cli version update issue.

jinyunma:
minor update to fix error message in [log](https://gcsweb-qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/qe-private-deck/pr-logs/pull/openshift_release/54195/rehearse-54195-periodic-ci-openshift-openshift-tests-private-release-4.17-amd64-nightly-azure-stack-upi-f28/1810967271162515456/artifacts/azure-stack-upi-f28/upi-install-azurestack/build-log.txt):
```
~/installer /output
ls: cannot access '/usr/lib64/az/lib/python*/site-packages/certifi/cacert.pem': No such file or directory
WARNING: Switched active cloud to 'XXX'.
```
@jianlinliu PTAL again ,thanks.

jinyunma:
> if the latest az-cli does not bring any trouble for us. I am okay with it.

okay, UPI installation on both azure public and azure stack hub are successful with az 2.61.0 installed in pwsh image(version: v7.3.12), will keep it as now. Other related steps are still using az 2.49.0, if needed, we can update those steps to use 4.17+ upi installer image in the future.

ci-operator/step-registry/ipi/conf/azurestack/ipi-conf-azurestack-commands.sh
```diff
@@ -62,9 +62,16 @@ TENANT_ID=$(jq -r .tenantId "${SHARED_DIR}/osServicePrincipal.json")
 cat >> "${SHARED_DIR}/azurestack-login-script.sh" << EOF
 
 if [[ -f "${CLUSTER_PROFILE_DIR}/ca.pem" ]]; then
-  cp "${CLUSTER_PROFILE_DIR}/ca.pem" /tmp/ca.pem
-  cat /usr/lib64/az/lib/python*/site-packages/certifi/cacert.pem >> /tmp/ca.pem
-  export REQUESTS_CA_BUNDLE=/tmp/ca.pem
+    cp "${CLUSTER_PROFILE_DIR}/ca.pem" /tmp/ca.pem
+    if ls /usr/lib64/az/lib/python*/site-packages/certifi/cacert.pem > /dev/null 2>&1; then
```
jianlinliu, line 66: I wonder why not use `-f` here?

jinyunma: Because the file path contains wildcards "*", in this way, we don't care about python version.
Checked that ` if [[ -f /tmp/ca* ]]; then echo "ok"; fi` didn't work yesterday. But tried again just now on local host, ` if [ -f /tmp/ca* ]; then echo "ok"; fi` works, let me update to use this condition.

jinyunma:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.17-amd64-nightly-azure-stack-upi-f28

ci-operator/step-registry/ipi/conf/azurestack/ipi-conf-azurestack-commands.sh
jianlinliu: I guess you are right about the file path contains wildcards "*":
```
[jianlinliu@jialiu-laptop installer]$ ls ./D*
./Dockerfile.ci  ./Dockerfile.ci.rhel7  ./Dockerfile.upi.ci  ./Dockerfile.upi.ci.rhel8
[jianlinliu@jialiu-laptop installer]$ if [ -f Dockerfile.upi.* ]; then echo "ok"; fi
bash: [: Dockerfile.upi.ci: binary operator expected
```

gpei:
> I have no idea why it changed the way of installing az cli
Seen from https://github.com/openshift-splat-team/pwsh/commit/e5eeb60979e07c7d8ac4443f966bdf3ec855604c, the az cli is installed as a dependency of powershell, so which means the `az` installed into the upi-instealler image will work for both vSphere UPI and Azure installations.

ci-operator/step-registry/ipi/conf/azurestack/ipi-conf-azurestack-commands.sh
jinyunma: hmm, `[ ]` works on my local host, checked step `upi-install-azurestack` in [job](https://qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/qe-private-deck/pr-logs/pull/openshift_release/54195/rehearse-54195-periodic-ci-openshift-openshift-tests-private-release-4.17-amd64-nightly-azure-stack-upi-f28/1811264411780780032), also works well, no error this time.
```
~/installer /output
WARNING: Switched active cloud to 'WWT'.
WARNING: Use 'az login' to log in to this cloud.
WARNING: Use 'az account set' to set the active subscription.
```

gpei:
So now the az is installed always with the latest(including pre-releases)https://github.com/openshift-splat-team/pwsh/blob/main/install-azure-cli.sh#L5

> --upgrade-strategy <upgrade_strategy>
> Determines how dependency upgrading should be handled [default: only-if-needed]. “eager” - dependencies are upgraded regardless of whether the currently installed version satisfies the requirements of the upgraded package(s). “only-if-needed” - are upgraded only when they do not satisfy the requirements of the upgraded package(s).
> 
> 
> --pre
> Include pre-release and development versions. By default, pip only finds stable versions.

ci-operator/step-registry/ipi/conf/azurestack/ipi-conf-azurestack-commands.sh
jinyunma: if more than one files are matched, indeed get some issues. It should be more safe to use `ls <file_path> > /dev/null 2>&1`
```
[jima@mylaptop tmp]$ ls *.sh
azurestack-login-script-1.sh  azurestack-login-script.sh  cucushift-installer-check-azure-cp-identity-commands.sh  test.sh
[jima@mylaptop tmp]$ if [ -f *.sh ]; then echo "ok"; fi
bash: [: too many arguments
[jima@mylaptop tmp]$ if ls *.sh > /dev/null 2>&1; then echo "ok"; fi
ok

```

jianlinliu: Okay, if the wildcards path only locate one file, then `[]` works.

jianlinliu: okay, I am okay to restore to use `ls`

jinyunma:
> Once we found some latest az version doesn't work with our Azure scripts, I guess we have to pin it a specific version again.

current version is `2.61.0.post20240516044921`, this might be pre-release or development version? 
```
$ az version
{
  "azure-cli": "2.61.0.post20240516044921",
  "azure-cli-core": "2.61.0.post20240516044921",
  "azure-cli-telemetry": "1.1.0.post20240516044921",
  "extensions": {}
}
```
The stable version is as below:
```
$ az version
{
  "azure-cli": "2.52.0",
  "azure-cli-core": "2.52.0",
  "azure-cli-telemetry": "1.1.0",
  "extensions": {}
}

```
I think we should use stable version to test, right? maybe need to remove `--pre` option?

ci-operator/step-registry/ipi/conf/azurestack/ipi-conf-azurestack-commands.sh
jinyunma: change back to use `ls <file_path> > /dev/null 2>&1`.
PTAL

jianlinliu:
> I think we should use stable version to test, right? maybe need to remove `--pre` option?

Yeah, I agree to use stable version.

> The stable version is as below: - 2.52.0"

If that would hit https://github.com/openshift/installer/pull/7297 ?

PR #54210, opened 2024-07-10
DPTP-4088: Fixes issue by removing DNS configs

bear-redhat:


deepsm007:
/retitle DPTP-4088: Fixes issue by removing DNS configs

wking (approved):
I don't know if this will be sufficient or not, but I have no problems guessing and checking.  Last week's openshift/managed-cluster-config#2158 doesn't seem to have been sufficient.

/lgtm

wking:
I manually removed `nodePlacement` from build03 and build04 to catch up with this^ using:

```console
$ oc --as system:admin patch dns.operator.openshift.io default --type json -p '[{"op": "remove", "path": "/spec/nodePlacement"}]'
```

build05 still had some config, so the openshift/release GitOps tooling automatically removed that property on that cluster.

PR #54225, opened 2024-07-10
HIVE-2559: mce-2.5 branch 4.16 deployment compat

celebdor:
As part of the effort to allow Hive's version in MCE-2.5 to be able to deploy 4.16 with FIPS enabled, we need to backport Hive's mce-2.6 work for building the hive operator container with both rhel8 and rhel9 compatibility. This change will allow for such build to happen in CI.


celebdor:
/close
in favor of #60723

PR #54253, opened 2024-07-11
ptp ci switch time for 4.16/4.17

jzding:
This PR is switching scheduled 4.16/4.17 PTP nightly CI so that 4.17 runs later than 4.16 and finishes in the morning. The reason being 4.17/master is where we have most code churns. By leaving the CI cluster setup in 4.17 in the morning, developers can run dev CIs during the day without re-setup 4.17 cluster.

### DIFF
--- a/ci-operator/config/openshift/release/openshift-release-master__nightly-4.16.yaml
+++ b/ci-operator/config/openshift/release/openshift-release-master__nightly-4.16.yaml
@@ -795,7 +795,7 @@ tests:
     workflow: telco5g-sno-cnf
 - as: e2e-telco5g-ptp
   cluster: build05
-  cron: 0 9 */1 * *
+  cron: 0 4 */1 * *
   steps:
     env:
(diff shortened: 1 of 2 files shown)


aneeshkp (approved):
/lgtm

vrutkovs:
Is this "we're tweaking cron time to avoid hitting other jobs because jobs don't use https://docs.ci.openshift.org/docs/architecture/quota-and-leases/"? Again?

jzding:
> Is this "we're tweaking cron time to avoid hitting other jobs because jobs don't use https://docs.ci.openshift.org/docs/architecture/quota-and-leases/"? Again?

no. we simply switch these 2 jobs so that 4.17 setup ends up in the morning and developers and use it to do their dev testing on 4.17 during the day.

vrutkovs:
Oh, okay
/approve

it would be handy to have it in the description :)

PR #54265, opened 2024-07-11
Storage OLM e2e jobs should set OO_TARGET_NAMESPACES to !install

dobsonj:
When we set `OO_TARGET_NAMESPACES` to `!all` or leave it blank, the presubmit job installs the CSV to _ALL_ namespaces. See this [gcp-filestore](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_gcp-filestore-csi-driver-operator/96/pull-ci-openshift-gcp-filestore-csi-driver-operator-main-operator-e2e/1811089655865544704/artifacts/operator-e2e/gather-extra/artifacts/oc_cmds/clusterserviceversions) job for example. I think this came from an example and then got copied around to our other operators.

In contrast, if we set `OO_TARGET_NAMESPACES` to `!install`, the CSV only gets installed to the namespace where the operator is installed. See this [LSO](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_local-storage-operator/489/pull-ci-openshift-local-storage-operator-master-e2e-operator/1811089588978978816/artifacts/e2e-operator/gather-extra/artifacts/oc_cmds/clusterserviceversions) job for example.

I don't think this is actively causing problems, it's just messy and it would be better for our presubmit jobs to install the CSV to a single namespace like a customer would typically do.

/cc @openshift/storage

### DIFF
--- a/ci-operator/config/openshift/csi-operator/openshift-csi-operator-master.yaml
+++ b/ci-operator/config/openshift/csi-operator/openshift-csi-operator-master.yaml
@@ -224,6 +224,9 @@ tests:
       OO_INDEX: ci-index-aws-efs-csi-driver-operator-bundle
     env:
       OO_CHANNEL: stable
+      OO_INSTALL_NAMESPACE: openshift-cluster-csi-drivers
+      OO_PACKAGE: aws-efs-csi-driver-operator
+      OO_TARGET_NAMESPACES: '!install'
       TEST_SUITE: openshift/csi
     workflow: openshift-e2e-aws-csi-efs
 - as: aws-efs-operator-e2e-extended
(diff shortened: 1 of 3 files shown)


dobsonj:
> ClusterServiceVersion gcp-filestore-csi-driver-operator.v4.16.0 status message: OwnNamespace InstallModeType not supported, cannot configure to watch own namespace

Nevermind, this was a failed experiment. Each operator specifies supported install modes in the CSV. In gcp-filestore for example:
https://github.com/openshift/gcp-filestore-csi-driver-operator/blob/06ed529f8242c1862b681efb89c5695f8125d15e/config/manifests/stable/gcp-filestore-csi-driver-operator.clusterserviceversion.yaml#L58-L66

So for these operators, they really should set `OO_TARGET_NAMESPACES` to `!all`, I was just puzzled why LSO was different.

PR #54271, opened 2024-07-11
Limit informing (optional) jobs to two retries

sdodson:
There are no recent examples where a third retry yields success unless it's tied to an infrastructure outage that just happens to clear by the third retry. Therefore and according to recent TRT blog post, limit those jobs to two retries to speed up iteration cycles.

The blog I cite is internal so I will not link it, but it mentions the following

> For single jobs, your job must pass all tests at least once in a single run, and we can retry your job no more than twice.


sdodson:
/cc @stbenjam 
PTAL

PR #52636, opened 2024-05-30
vSphere UPI on VCM

rvanderp3:

PR #54302, opened 2024-07-12
rename vsphere-8 ci jobs

jianlinliu:
follow up https://github.com/openshift/release/pull/52636, now almost all the jobs are going to run on vsphere-8, so rename the vsphere-8 ci jobs name.


jianlinliu:
cc @liangxia to review.

PR #54323, opened 2024-07-12
OCP storage should use -mod=readonly for verify & unit tests

dobsonj:
Various upstream projects need to build test tools from the internet (like go install sigs.k8s.io/logtools/logcheck@v0.8.2 ) in  make tests  / make verify. These fail with -mod=vendor ([AWS EBS rebase](https://github.com/openshift/aws-ebs-csi-driver/pull/270), [attacher rebase](https://github.com/openshift/csi-external-attacher/pull/76))

We want to keep -mod=vendor for the binary build, but in some case we need -mod=readonly for unit tests and verify jobs. We already have a precedent for this in [secrets-store-csi-driver](https://github.com/openshift/release/blob/4a96112600e6fe59e0a705616599f07e1f553176/ci-operator/config/openshift/secrets-store-csi-driver/openshift-secrets-store-csi-driver-main.yaml#L81) for example.

/cc @openshift/storage

### DIFF
--- a/ci-operator/config/openshift/aws-ebs-csi-driver/openshift-aws-ebs-csi-driver-master.yaml
+++ b/ci-operator/config/openshift/aws-ebs-csi-driver/openshift-aws-ebs-csi-driver-master.yaml
@@ -52,7 +52,7 @@ tests:
 - as: verify
   commands: |
     # Set $HOME for golangci-lint cache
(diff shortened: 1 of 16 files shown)


dobsonj:
/pj-rehearse pull-ci-openshift-ibm-powervs-block-csi-driver-main-verify

PR #54349, opened 2024-07-15
CNF-13439: Add kube-compare image build

natifridman:
Build kube-compare image postsubmit

### DIFF
--- a/ci-operator/config/openshift/kube-compare/openshift-kube-compare-main.yaml
+++ b/ci-operator/config/openshift/kube-compare/openshift-kube-compare-main.yaml
@@ -1,14 +1,50 @@
 base_images:
+  base:
+    name: "4.16"
+    namespace: ocp
+    tag: base
   golangci-lint:
     name: golangci-lint
     namespace: ci
     tag: v1.59.0
+  ocp_builder_rhel-8-golang-1.21-openshift-4.16:
+    name: builder
+    namespace: ocp
+    tag: rhel-8-golang-1.21-openshift-4.16
+  ocp_builder_rhel-9-golang-1.21-openshift-4.16:
+    name: builder
+    namespace: ocp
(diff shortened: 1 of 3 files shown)


ci-operator/config/openshift/kube-compare/openshift-kube-compare-main.yaml
```diff
@@ -1,14 +1,50 @@
...
+      as:
+      - registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.21-openshift-4.16
+    ocp_builder_rhel-9-golang-1.21-openshift-4.16:
+      as:
+      - registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.21-openshift-4.16
+  to: kube-compare-artifacts
+promotion:
+  to:
+  - namespace: ocp
+    tag: "4.16"
```
nocturnalastro, line 37: Is that the tag its being pushed as? I think it would just be lastest ATM

natifridman: I think the convention is to use Openshift versions here

nocturnalastro: But the tool should be independent of the openshift version as the reference is external, this could be confusing for users.

natifridman: How k8s packages going to be managed? for example, I think `kubectl` is supported within one minor version

natifridman: I think it also pushes to latest but I'm not sure about it...

natifridman:
/pj-rehearse pull-ci-openshift-kube-compare-main-unit-report-creator

PR #54350, opened 2024-07-15
Access to QCI registry for openshift-serverless admins

mgencur:
This is following instructions to gain access to QCI registry:
* https://docs.ci.openshift.org/docs/how-tos/use-registries-in-build-farm/#how-do-i-gain-access-to-qci
* https://docs.ci.openshift.org/docs/how-tos/use-registries-in-build-farm/#token-for-programmatic-access-to-qci


mgencur:
@danilo-gemoli Thanks!

PR #54370, opened 2024-07-15
MULTIARCH-4696: update ocp tests private lease times

prb112:


prb112:
Hi @liangxia Could you approve this cronjob timing changes? This is per @MayXuQQ @juliemathew Thanks, Paul

PR #54403, opened 2024-07-16
Remove vendor/ before calling `make verify`

jsafrane:
AWS EBS CSI driver upstream does not have vendor/ directory in the repository. `make verify` then checks that all *.go files in the repo have the right license header.

In OCP, we do have vendor dir checked into the repository and `make verify` complains that some files there miss the license header. Remove the vendor dir before the check then.


jsafrane:
/close
this breaks `verify-kustomize`.

PR #54425, opened 2024-07-16
adding runningCount to all rhoe cluster pools

acornett21:
- add runningCount to all rhoe cluster pools as a % of the total size.
- remove ocp-4-15 east 1 cluster pool, now that ocp 4.16 has a few dot releases.


komish (approved):
/lgtm

PR #54442, opened 2024-07-17
Remove OKO dataplane-operator jobs

bshephar:
This change removes the dataplane-operator jobs and config since the dataplane-operator has been merged with openstack-operator


bshephar:
Not required any more.

PR #54533, opened 2024-07-18
Failover build02 to build09

deepsm007:
/cc @danilo-gemoli @openshift/test-platform


deepsm007:
/hold 
with the new changes it looks like the cluster should be fine

PR #54541, opened 2024-07-18
Add 4.17 periodic jobs for OpenShift Compliance testing

rhmdnd:


ci-operator/config/ComplianceAsCode/content/ComplianceAsCode-content-master__4.17.yaml
```diff
@@ -0,0 +1,893 @@
+build_root:
+  image_stream_tag:
+    name: release
+    namespace: openshift
+    tag: golang-1.18
```
Vincent056, line 5: I wonder if we need to update the tag here

rhmdnd: I don't think it would hurt to update it. Before we do that we might also want to update the go version in ocp4e2e, too.

https://github.com/ComplianceAsCode/ocp4e2e/blob/main/go.mod#L3

rhmdnd: Kicking the tires on that in https://github.com/ComplianceAsCode/ocp4e2e/pull/47

rhmdnd:
CI spins up the right cluster version, but we need to adjust the expected results to match security and compliance posture for 4.17.

```
helpers.go:879: E2E-FAILURE: The expected result for the e2e-cis-node-master-etcd-unique-ca rule didn't match. Expected 'map[master:PASS worker:NOT-APPLICABLE]', Got 'FAIL'
```

We'll need to do this in a separate PR to the ComplianceAsCode/content repository.

@Vincent056 @yuumasato do you have any feedback on this approach?

yuumasato (approved):
/lgtm

Let's see how the rehearsals go

PR #54546, opened 2024-07-18
stackrox: add auto team where gavin is

davdhacs:
After this:
```
$ find . -name OWNERS | grep stackrox | xargs -I{} grep -l gavin {} | while read -r fn; do cat "$fn" | grep -v gavin | grep 'Brad\|tom\|davd' >/dev/null || echo $fn; done
$
```

I also removed some long-gone teammates, but I did not remove all who transcended.


gavin-stackrox (approved):
/lgtm

PR #54560, opened 2024-07-19
Fix RHDH slack user mention in nightly alert

subhashkhileri:
Fix RHDH slack user mention which is sent as simple text instead of mention.
https://issues.redhat.com/browse/RHIDP-3222


josephca (approved):
/lgtm
/approve

PR #54582, opened 2024-07-19
Exclude CI from running on github workflow patches

rhmdnd:
We use GitHub workflows in ComplianceAsCode/compliance-operator for
building container images and vulnerability scanning, but we don't need
to exercise full e2e testing on these patches since updates to the
workflows don't directly affect the operator code.

Let's save on resources by not running CI on patches that update GitHub
workflows.


Vincent056 (approved):
/lgtm

rhmdnd:
/pj-rehearse ack

acking the rehearsals since this is just an exclusion and not changing the jobs themselves.

PR #54592, opened 2024-07-19
Cleanup Dashboard, align with current setup

andrewballantyne:
Does a few things to catch up
* Removes incubation configs since we don't use the branch anymore
* Allows us to maintain a specific set of requirements for branch protections
* Updates the default merge strategy to squash
* ~Allows for a merge strategy if we need it via a label on PRs~ nvm, seems unsupported by OpenShift/release


andrewballantyne:
Breaking this up into 2 PRs -- Cleaning up incubation stuff can be in its own PR to avoid blocking everything by unknown duration for reviews from the dptp team.

PR #54595, opened 2024-07-19
Removal of the configurations around incubation branch, we don't use it anymore

andrewballantyne:
Removal of the `incubation` branch settings, the branch is no longer used and images are still built from it.


lucferbux (approved):
/lgtm

PR #54640, opened 2024-07-22
Add Step2/MetalLB operator install

josecastillolema:


josecastillolema:
/pj-rehearse pull-ci-openshift-qe-ocp-qe-perfscale-ci-main-metal-4.16-latest-x86-day2-metallb

josecastillolema:
@jtaleric ptal

ci-operator/step-registry/openshift-qe/installer/bm/day2/metallb/openshift-qe-installer-bm-day2-metallb-commands.sh
```diff
@@ -0,0 +1,75 @@
...
+  source: redhat-operators
+  sourceNamespace: openshift-marketplace
+EOF
+
+# Wait for the operator to be ready
+  until [ "$(kubectl get csv -n metallb-system | grep metallb-operator > /dev/null; echo $?)" == 0 ];
+    do echo "Waiting for MetalLB operator"
+    sleep 5
+  done
+  kubectl wait --for jsonpath='{.status.phase}'=Succeeded --timeout=10m -n metallb-system "$(kubectl get csv -n metallb-system -oname)"
```
jtaleric, line 60: should we have the `10m` timeout be something we default to in the ref file, but allow the user to override the timeout?

ci-operator/step-registry/openshift-qe/installer/bm/day2/metallb/openshift-qe-installer-bm-day2-metallb-commands.sh
```diff
@@ -0,0 +1,75 @@
...
+spec:
+  channel: stable
+  name: metallb-operator
+  source: redhat-operators
+  sourceNamespace: openshift-marketplace
+EOF
+
+# Wait for the operator to be ready
+  until [ "$(kubectl get csv -n metallb-system | grep metallb-operator > /dev/null; echo $?)" == 0 ];
+    do echo "Waiting for MetalLB operator"
```
jtaleric, line 57: This loop could go on forever, should we print out more debug?

ci-operator/step-registry/openshift-qe/installer/bm/day2/metallb/openshift-qe-installer-bm-day2-metallb-commands.sh
josecastillolema: I guess most of the operators are installed way faster than 10 min. It is the same setup we have for the SRIOV one and so far has been consistent. Anyway, if you feel that the timeout should be configurable let me know and I'll quickly introduce the TIMEOUT var.

ci-operator/step-registry/openshift-qe/installer/bm/day2/metallb/openshift-qe-installer-bm-day2-metallb-commands.sh
josecastillolema: So, even if we print the output of:
```
kubectl get csv -n metallb-system
```

I guess at this moment it would not bring any relevant additional info. Do you have anything else in mind?

ci-operator/step-registry/openshift-qe/installer/bm/day2/metallb/openshift-qe-installer-bm-day2-metallb-commands.sh
jtaleric: ah ok - i am good with having or not having it. I didn't know if there was situations where we might want to bump that timeout.

ci-operator/step-registry/openshift-qe/installer/bm/day2/metallb/openshift-qe-installer-bm-day2-metallb-commands.sh
jtaleric: at the moment no.. My concern is waiting forever, so in the vent we timeout due to waiting, it might be nice for just some debug in the step that failed.

PR #54670, opened 2024-07-23
Disable Prow branch protection for caikit-nlp repository

vaibhavjainwiz:
We need to disable in [opendatahub-io/caikit-nlp repo](https://github.com/opendatahub-io/caikit-nlp) the setting in branch protection rules that `Requires status checks to pass before merging`


dtrifiro:
@vaibhavjainwiz 

Do we need to set unmanaged for this? I was looking at the docs https://docs.prow.k8s.io/docs/components/optional/branchprotector/ and it seems that we can just disable status checks.

PR #54674, opened 2024-07-23
support to create multi custom mcp

jianlinliu:
support users to define multiple custom mcp with their wanted label filter and pool size.
This also help https://issues.redhat.com/browse/OCPQE-24143

### DIFF
--- a/ci-operator/step-registry/mco/conf/day2/create-custom-mcp/mco-conf-day2-create-custom-mcp-commands.sh
+++ b/ci-operator/step-registry/mco/conf/day2/create-custom-mcp/mco-conf-day2-create-custom-mcp-commands.sh
@@ -4,12 +4,6 @@ set -e
 set -u
 set -o pipefail
 
-if  [ "$MCO_CONF_DAY2_CUSTOM_MCP_NAME" == "" ]; then
-  echo "This installation does not need to create any custom MachineConfigPool"
-  exit 0
-fi
-
-
 function set_proxy () {
     if [ -s "${SHARED_DIR}/proxy-conf.sh" ]; then
         echo "Setting the proxy ${SHARED_DIR}/proxy-conf.sh"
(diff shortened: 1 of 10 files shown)


jianlinliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.16-multi-nightly-aws-ipi-arm-mixarch-f9-day2-64k-pagesize

jianlinliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.16-multi-nightly-aws-ipi-arm-mixarch-f9-day2-64k-pagesize

ci-operator/step-registry/mco/conf/day2/create-custom-mcp/mco-conf-day2-create-custom-mcp-commands.sh
```diff
@@ -60,32 +73,46 @@ EOF
     echo "It has been requested to add 0 nodes to the pool. No node will be added to the custom MachineConfigPool."
     return
   fi
-  [ -z "$MCP_NUM_NODES" ] \
-    && echo "MCO_CONF_DAY2_CUSTOM_MCP_NUM_NODES variable is empty. All nodes matching the '$MCP_LABEL' label will be added to the custom pool" \
-    || echo "Labeling $MCP_NUM_NODES worker nodes in order to add them to the new custom MCP"
-  # shellcheck disable=SC2046
-  oc label node $(oc get nodes -l "$MCP_LABEL" -ojsonpath="{.items[:$MCP_NUM_NODES].metadata.name}") node-role.kubernetes.io/"$MCO_CONF_DAY2_CUSTOM_MCP_NAME"=
+  if (( ${#NODES_LIST[@]} == 0 )); then
```
rioliu-rh, line 76: suggest to add a checkpoint here to make sure node num in NODES_LIST >= MCP_NUM_NODES

ci-operator/step-registry/mco/conf/day2/create-custom-mcp/mco-conf-day2-create-custom-mcp-ref.yaml
```diff
@@ -10,7 +10,7 @@ ref:
     - name: MCO_CONF_DAY2_CUSTOM_MCP_NAME
       default: ""
       documentation: |
-        The name of the custom MachineConfigPool that will be created
+        The name list of the custom MachineConfigPool that will be created
```
rioliu-rh, line 13:
```suggestion
        Support multiple pools creation, the MachineConfigPool name list can be defined as space delimited string
```

rioliu-rh: Suggest to sync this change to README.md as well

ci-operator/step-registry/mco/conf/day2/create-custom-mcp/mco-conf-day2-create-custom-mcp-commands.sh
rioliu-rh: So the node num is not a sensitive param, by default we will create multiple pools with same size if possible
Is there any requirement to create different pools with different size, e.g. I have a mix-arch cluster, the cluster has 3 workers, 1 amd64 worker, 2 arm64 workers, then I want to create 2 pools for different arches separately.

sergiordlr:
This change could have an impact in arm and multiarch jobs. We can ping @aleskandro  so that he can have a look a this change.

ci-operator/step-registry/mco/conf/day2/create-custom-mcp/mco-conf-day2-create-custom-mcp-commands.sh
jianlinliu: Actually I have no specific requirement to create different pools with different size, but during my coding, I think that is also a possible case. E.g: 3 workers in total, but set `MCP_NUM_NODES` to 2, then the 2nd mcp's size will be different with the 1st one. Of course, if users want to create multiple pools with same size, they need to adjust the setting of `MCP_NUM_NODES` or scale up worker number to make the two mcp size the same. But from code perspective, that is NOT a MUST requirement.

jianlinliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.16-multi-nightly-aws-ipi-arm-mixarch-f9-day2-64k-pagesize

ci-operator/step-registry/mco/conf/day2/create-custom-mcp/mco-conf-day2-create-custom-mcp-commands.sh
rioliu-rh: `same size` here means we only support user to configure the pool size for all the pools by param `MCO_CONF_DAY2_CUSTOM_MCP_NUM_NODES`, to make this step more robust, It's better to create a pool with 2 params, `label` and `size`. e.g. according to above scenario, if I want to create 2 pools for arm64 and amd64 workers, I need to filter out the nodes by arch related `label`,then label the `size` nodes with new mcp name

jianlinliu: I did not hear it is a MUST that have to two mcp with the same size, I think that totally depends on user's desire. 

> according to above scenario

Are you talking about "I have a mix-arch cluster, the cluster has 3 workers, 1 amd64 worker, 2 arm64 workers, then I want to create 2 pools for different arches separately" ? What size you are want to set for each mcp ?

> if I want to create 2 pools for arm64 and amd64 workers, I need to filter out the nodes by arch related label,then label the size nodes with new mcp name

I think that is reasonable, what behavior are you expecting ?

rioliu-rh: > Are you talking about "I have a mix-arch cluster, the cluster has 3 workers, 1 amd64 worker, 2 arm64 workers, then I want to create 2 pools for different arches separately" ? What size you are want to set for each mcp ?

Yes

> I think that is reasonable, what behavior are you expecting ?

for example, I have a mix-arch cluster provision step, I can configure node num for different arches, so before creating custom pool I already know the pool size of different arches

then we can use 3 params to configure pool name, pool size and node label e.g.
`MCO_CONF_DAY2_CUSTOM_MCP_NAME=mcp4amd64 mcp4arm64`
`MCO_CONF_DAY2_CUSTOM_MCP_FROM_LABEL=kubernetes.io/arch=amd64 kubernetes.io/arch=arm64`
`MCO_CONF_DAY2_CUSTOM_MCP_NUM_NODES=1 2`

for every pool we have 3 params, the data struct is like below 
```
{
   "pool_name": "mcp4amd64",
   "pool_size": 1,
   "node_label": "kubernetes.io/arch=amd64"
}
```

To create every pool the logic is like, fill the `NODES_LIST` by `node_label`, create the custom pool with `pool_name`, loop the `NODES_LIST` to get nodes by `pool_size` and label these nodes with `pool_name`

rioliu-rh: @sergiordlr @aleskandro WDYT

jianlinliu: > MCO_CONF_DAY2_CUSTOM_MCP_NAME=mcp4amd64 mcp4arm64
> MCO_CONF_DAY2_CUSTOM_MCP_FROM_LABEL=kubernetes.io/arch=amd64 kubernetes.io/arch=arm64
> MCO_CONF_DAY2_CUSTOM_MCP_NUM_NODES=1 2

Do you mean you want `MCO_CONF_DAY2_CUSTOM_MCP_FROM_LABEL` and `MCO_CONF_DAY2_CUSTOM_MCP_NUM_NODES` to support multi values, even they are mapped in order?

```
{ "pool_name": "mcp4amd64", "pool_size": 1, "node_label": "kubernetes.io/arch=amd64" }
{ "pool_name": "mcp4arm64", "pool_size": 2, "node_label": "kubernetes.io/arch=arm64" }
```

If my understand is right, the parameter should be kind of hash dict data, but not use 3 ENV var to map them in order, it is not a robust way. If we want the accurate mcp definition, we need to a big change to replace the 3 ENV with a new ENV using json format value.  

In current implementation, it is supposed to support multi mcp by only setting `MCO_CONF_DAY2_CUSTOM_MCP_NAME`, both `MCO_CONF_DAY2_CUSTOM_MCP_FROM_LABEL` and `MCO_CONF_DAY2_CUSTOM_MCP_NUM_NODES` are shared for each mcp creation.

jianlinliu: Just like what I explained in the above, the current implemention do NOT support that, it only support create multi mcp with the same label filter and size setting. I do not have plan to change it too much to avoid impact on the existing ci jobs.

jianlinliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.16-multi-nightly-aws-ipi-arm-mixarch-f9-day2-64k-pagesize

jianlinliu:
> I have some ideas to provide support for more than one custom MCP using different labels and different number of nodes. The idea is to use 3 lists, and a default value if any list is not provided (or complete)

3 list ENV vars will lead a lot of codes to manipulate these lists and map them, I think that is not flexy enough, and hard to maintain the code, I prefer to ask users to pass a json format string as their desire, that will be more flexy and not have to maintain the complex codes.

E.g: 
```
MCO_CONF_DAY2_CUSTOM_MCP='[{"mcp_name": "aa", "mcp_node_num": "2", "mcp_node_label": "kubernetes.io/arch=arm64"}, {"mcp_name": "bb", "mcp_node_num": "1", "mcp_node_label": "kubernetes.io/arch=amd64"}]'
```

But this will impact all the existing ci jobs.

sergiordlr:
Yes, that's why I tried the 3 lists approach, in order not to impact the existing code.
The main user of this step is @aleskandro . Please, @aleskandro could you give us your opinion regarding this subject and the impact of this change in arch jobs?

ci-operator/step-registry/mco/conf/day2/create-custom-mcp/mco-conf-day2-create-custom-mcp-commands.sh
aleskandro: I like @rioliu-rh idea of defining a better 'data model' for this step since ever. However, I also understand the complexity of that. 

Can you clarify what are the reasons triggering the changes in this step now? Is there a new feature that is going to make use of this step?

jianlinliu: Right, this change is from a test requirement from a slack [discussion](https://redhat-internal.slack.com/archives/GH9JHQFKN/p1721311802315329?thread_ts=1721107446.852209&cid=GH9JHQFKN).

If everyone like the idea of 'data model', I can have a shot.

jianlinliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.16-multi-nightly-aws-ipi-arm-mixarch-f9-day2-64k-pagesize

jianlinliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.16-multi-nightly-aws-ipi-arm-mixarch-f9-day2-64k-pagesize

jianlinliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.16-multi-nightly-aws-ipi-arm-mixarch-f9-day2-64k-pagesize

jianlinliu:
cc @aleskandro @rioliu-rh @sergiordlr to review, from [log](https://gcsweb-qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/qe-private-deck/pr-logs/pull/openshift_release/54674/rehearse-54674-periodic-ci-openshift-openshift-tests-private-release-4.16-multi-nightly-aws-ipi-arm-mixarch-f9-day2-64k-pagesize/1816074946854522880/artifacts/aws-ipi-arm-mixarch-f9-day2-64k-pagesize/mco-conf-day2-create-custom-mcp/build-log.txt), it works well.

sergiordlr:
Do you think that having a default value for "mcp_node_label" (`node-role.kubernetes.io/worker`) and "mcp_node_num" (`1`) so that they don't have to be defined always is a good idea?

jianlinliu:
> Do you think that having a default value for "mcp_node_label" (`node-role.kubernetes.io/worker`) and "mcp_node_num" (`1`) so that they don't have to be defined always is a good idea?

My original idea is to always ask user input them, because now `MCO_CONF_DAY2_CUSTOM_MCP` is the parameter, but not any element in the json data string. But if you like that style, I have no strong objection.

Updated for your proposal.

jianlinliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-multi-nightly-azure-ipi-arm-mixarch-disc-fullypriv-f28-day2-64k-pagesize

jianlinliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-multi-nightly-aws-ipi-arm-mixarch-f28-day2-64k-pagesize

jianlinliu:
@sergiordlr @aleskandro this PR is ready to review and merge, please take a look.

aleskandro (commented):
/lgtm

Thanks @jianlinliu

aleskandro:
/hold

To allow others reviews

rioliu-rh:
LGTM from my side

sergiordlr:
LGTM 

Thank you very much!!

jianlinliu:
cc @liangxia to review

PR #54692, opened 2024-07-23
CI config for ironic-agent-image 4.18

elfosardo:


danilo-gemoli:
@elfosardo as far as I know `openshift-priv/*` gets updated automatically by our tools. Why isn't this the case? Do we have a policy or something that rules this repository out?

PR #54698, opened 2024-07-23
Reduce periodics exec frequency after GA

jmguzik:


ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__ppc64le-nightly.yaml
```diff
@@ -75,7 +75,7 @@ resources:
 tests:
 - as: powervs-ipi-f7
   cluster: build01
-  cron: 0 0 * * 4
+  cron: 19 22 8,28 * *
```
liangxia, line 78: @jhou1 @jianlinliu  Are those batch update for `openshift-tests-private` jobs expected ?

liangxia: CC @aishwaryabk for the ppc64le, too

liangxia: Just take ppc64le as an example.  Also amd64/arm64/multi are affected in this PR, too.

PR #54716, opened 2024-07-23
Updating makefile images

deepsm007:
/cc @openshift/test-platform 

/hold to verify

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,6 @@
 SHELL=/usr/bin/env bash -o errexit
 
-.PHONY: help check check-boskos check-core check-services dry-core core dry-services services all update template-allowlist release-controllers checkconfig jobs ci-operator-config registry-metadata boskos-config prow-config validate-step-registry new-repo branch-cut prow-config multi-arch-gen message
-
-message:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -3,8 +3,7 @@ SHELL=/usr/bin/env bash -o errexit
 .PHONY: help check check-boskos check-core check-services dry-core core dry-services services all update template-allowlist release-controllers checkconfig jobs ci-operator-config registry-metadata boskos-config prow-config validate-step-registry new-repo branch-cut prow-config multi-arch-gen message
 
 message:
-	@printf '\e[1mThis Makefile was recently migrated to use QCI as respository of images. Please make sure you are logged in to QCI before executing targets. Copy the token from https://oauth-openshift.apps.ci.l2s4.p1.openshiftapps.com/oauth/token/request and run `oc login`\e[0m\n'
-	@printf '\e[1mTo login, use your container engine, example for podman: podman login -u=$$(oc whoami) -p=$$(oc whoami -t) quay-proxy.ci.openshift.org\e[0m\n'
+	@printf '\e[1mThis Makefile was recently migrated to use QCI as respository of images.\e[0m\n'
```
smg247, line 6: I don't know that there is any value in having this message appear at all anymore, now that there is no action required from the user there is no point. If this doesn't work for some reason, we will just revert the whole thing and fix it.

PR #54739, opened 2024-07-24
Fix RHDH slack user mention in nightly alert

subhashkhileri:
Fix RHDH slack user mention in nightly alert


rhdh-qe (approved):
/approve
/lgtm

PR #54776, opened 2024-07-24
Bump mem for infra nodes on build02

bear-redhat:


jupierce:
For historical breadcrumbs: this is a test to see if we can accommodate spikes in internal image registry pod memory usage with large blob sizes configured for R2 integration. 
Due to an incompatibility in the R2 implementation of the S3 API for multi-part upload, we have temporarily set build02 blobsize to 5G. This change can be reverted if we do either of the following:
- We move away from R2 and reduce blob size.
- R2 fixes multi-part uploads and we reduce blob size.

PR #54778, opened 2024-07-24
Use n2-highmem-32

bear-redhat:


jupierce:
https://github.com/openshift/release/pull/54776#issuecomment-2248592762

PR #54785, opened 2024-07-24
Failover build02 to build09

deepsm007:
We are testing with different instance type on build02 to handle memory required for chunksize.

/cc @openshift/test-platform 

/hold


deepsm007:
Build02 is back to normal

PR #54787, opened 2024-07-24
Fix PROW schedular to deal with potential race conditions

vrindle:


ci-operator/step-registry/dpu-operator/e2e-tests-workflow/tests/dpu-operator-e2e-tests-workflow-tests-commands.sh
```diff
@@ -23,7 +22,11 @@ do
     result=$(echo "$job_check" | jq -r '.result')
     if [ $result != "null" ]; then
 	sleep $((RANDOM % 61 + 60))
-        break
+	job_check=$(curl -k -s --resolve "$endpoint_resolve" "$job_url/api/json")
```
wizhaoredhat, line 25: This is a copy paste of the code above. Could you think of a way to do it without copying?

wizhaoredhat: Also this is an infinite loop if the exit condition is never met.

vrindle: okay

ci-operator/step-registry/dpu-operator/e2e-tests-workflow/tests/dpu-operator-e2e-tests-workflow-tests-commands.sh
```diff
@@ -1,13 +1,24 @@
...
 token_dpu_operator_key=$(cat "/var/run/token/dpu-token/dpu-key")
 endpoint=$(cat "/var/run/token/dpu-token/url")
 
-job_url="https://${endpoint}/job/99_Lab217_IPU_Deploy"
-
+job_url="https://${endpoint}/job/99_Lab217_E2E_IPU_Deploy"
 endpoint_resolve="${endpoint}:443:10.0.180.88"
 
-max_sleep_duration=86400  # Maximum sleep duration in seconds (2 hours)
+max_sleep_duration=432000  # Maximum sleep duration in seconds (5 days)
```
wizhaoredhat, line 21: why did you bump this to 5 days?

vrindle: This is just a safeguard. In case there are a lot of PRs I don't want it to timeout prematurely.

wizhaoredhat: Multiple PRs can be in the middle of this script in parallel?

vrindle: No, what this code does is wait for other PRs to finish executing PRIOR to firing. Hence the need for the long timetout. Let's say there are like 20 PR's in the queue. It is going to take a long time to clear everything out.

wizhaoredhat: If multiple PRs **can not** be in the middle of this script in parallel then why do we need this timeout? If multiple instances of this script would be running in PROW then would there be a better way to make sure prow serializes them instead of running them in parallel.

vrindle: I don't think there is any way to stop prow from running multiple instances of this script in parallel. Did you have something in mind?

wizhaoredhat: https://github.com/openshift/release/blob/master/ci-operator/step-registry/telco5g/cnf/tests/telco5g-cnf-tests-commands.sh

Does the Telco 5g test also have such a wait in their scripts? It just seems quite fragile to wait and block on the output of the Jenkins job. It seems that we could potentially have a race.

wizhaoredhat: E.g. Reading from the Job (Result is empty) -> do stuff -> (Jenkins Job Result is now not empty due to another running PR) -> Try to run the Job.

The thing is that there is no proper mutual exclusion locking.

ci-operator/step-registry/dpu-operator/e2e-tests-workflow/tests/dpu-operator-e2e-tests-workflow-tests-commands.sh
```diff
@@ -1,72 +1,99 @@
 #!/bin/bash
 
+manage_queue() {
+	GET_QUEUE=$(curl -k -s --resolve "$endpoint_resolve" "https://${endpoint}/queue/api/json")
+
+	QUEUE_FILTERED_JOB_NAME=$(echo $GET_QUEUE | jq '.items[] | select(.task.name == "bare-test")')
```
wizhaoredhat, line 6: bare-test should not be hardcoded. We should have a variable for this that we can replace with the 217 job in one place.

ci-operator/step-registry/dpu-operator/e2e-tests-workflow/tests/dpu-operator-e2e-tests-workflow-tests-commands.sh
```diff
@@ -1,72 +1,99 @@
...
-            exit 1
-        fi
-        sleep_counter=$((sleep_counter + 60))
-        sleep 60
-    fi
-done
+
+manage_queue
+
+curl -k --resolve "${endpoint_resolve}" "https://${endpoint}/view/dpu-test/job/bare-test/buildWithParameters?token=$token_dpu_operator_key&pullnumber=$PULL_NUMBER"
```
wizhaoredhat, line 94: bare-test should not be hardcoded.

ci-operator/step-registry/dpu-operator/e2e-tests-workflow/tests/dpu-operator-e2e-tests-workflow-tests-commands.sh
```diff
@@ -1,72 +1,99 @@
...
+    		fi
+	done
+}
 token_dpu_operator_key=$(cat "/var/run/token/dpu-token/dpu-key")
 endpoint=$(cat "/var/run/token/dpu-token/url")
+header=$(cat "/var/run/token/header")
 
-job_url="https://${endpoint}/job/99_Lab217_IPU_Deploy"
-
+job_url="https://${endpoint}/view/dpu-test/job/bare-test/lastBuild"
```
wizhaoredhat, line 82: variable for "bare-test"

wizhaoredhat:
@vrindle Please change the title of the PR and the commit title.

ci-operator/step-registry/dpu-operator/e2e-tests-workflow/tests/dpu-operator-e2e-tests-workflow-tests-commands.sh
```diff
@@ -1,72 +1,100 @@
 #!/bin/bash
 
+manage_queue() {
+	GET_QUEUE=$(curl -k -s --resolve "$endpoint_resolve" "https://${endpoint}/queue/api/json")
+
+	QUEUE_FILTERED_JOB_NAME=$(echo $GET_QUEUE | jq --arg testname "$test_name" '.items[] | select(.task.name == $testname)')
```
wizhaoredhat, line 6: If this fails then we should exit with error.

vrindle: okay I added a error check

PR #54789, opened 2024-07-24
OCPBUGS-37513: Add vsphere steps for csi-provisioner

gnufied:
xref https://issues.redhat.com/browse/OCPBUGS-37513

PR #54790, opened 2024-07-24
add e2e-vsphere-csi to csi-external-provisioner presubmit jobs

dobsonj:
We recently had to revert a csi-external-provisioner rebase because it broke on vsphere. This PR adds a presubmit job for vsphere to the csi-external-provisioner repo to catch this kind of thing earlier.

/cc @openshift/storage

### DIFF
--- a/ci-operator/config/openshift/csi-external-provisioner/openshift-csi-external-provisioner-master.yaml
+++ b/ci-operator/config/openshift/csi-external-provisioner/openshift-csi-external-provisioner-master.yaml
@@ -56,6 +56,13 @@ tests:
   steps:
     cluster_profile: gcp
     workflow: openshift-e2e-gcp-csi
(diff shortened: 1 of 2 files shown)


dobsonj:
The rehearsal should actually fail until https://github.com/openshift/csi-external-provisioner/pull/102 merges

gnufied:
ha I also opened - https://github.com/openshift/release/pull/54789

So this PR in itself will not catch intree PV deletion failing bug, because if you see - https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/241 then vsphere-csi test suite actually is passing, but it is generic `e2e-vsphere` that is failing.

dobsonj:
/close
in favor of https://github.com/openshift/release/pull/54789

PR #54794, opened 2024-07-25
OCPQE-24200 support configurable releases for prow job

rioliu-rh:
currently we cannot configure supported releases for different qe-release-gate-test job, even the runtime can skip triggering job if no test job is defined for current release+arch in test job registry, but the method is not efficient, the change in this PR can support us to configure supported releases (has test job definition) for different jobs


rioliu-rh:
/pj-rehearse periodic-ci-openshift-release-tests-master-qe-release-gate-test

PR #54816, opened 2024-07-25
Fix username for pierreprinetti

mandre:
This pointed to the wrong user. This was also the only occurrence in the repo, as `ag "pierre\b"` returns no results anymore.


pierreprinetti:
oh?
/lgtm

PR #54820, opened 2024-07-25
Add workflow for e2e in openshift-pipelines console-plugin

vikram-raj:
Add workflow to run e2e on the OCP cluster.


jerolimov:
@vikram-raj Didn't we make the decision that the dyn. plugin runs only on 4.15+? Then we might can remove the 4.12-4.14 jobs, or?

For the workflow property. It lgtm, but it would be nice if Jakub could take a look as well. :)

/cc @jerolimov @jhadvig

sanketpathak:
@vikram-raj I think we also need a cluster_profile specification
like here [https://github.com/openshift/release/blob/master/ci-operator/config/openshift/open[…]ift-tests-private/openshift-openshift-tests-private-master.yaml](https://github.com/openshift/release/blob/master/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-master.yaml#L210)

vikram-raj:
/pj-rehearse pull-ci-openshift-pipelines-console-plugin-release-4.15-e2e

vikram-raj:
/pj-rehearse pull-ci-openshift-pipelines-console-plugin-main-e2e

PR #54842, opened 2024-07-25
remove ocp4-15 testing from release-4.5

davdhacs:
ocp 4.16 is already configured to always_run or run_if_changed https://github.com/openshift/release/blob/e8357d0734743c2bf547c53ee131e6c5e4a780df/ci-operator/config/stackrox/stackrox/stackrox-stackrox-release-4.5__ocp-4-16.yaml#L24
ocp 4.15 does not need to be tested on the patch releases since ocp 4.16 will be tested
/hold
/uncc


tommartensen (changes requested):
Please also prepare the `release-x.y` template for the next release with the targetted OCP version.

PR #54848, opened 2024-07-26
Add one more skipped profiles

JianLi-RH:
/cc @jianlinliu @shellyyang1989 @liangxia


JianLi-RH:
Discussed by https://redhat-internal.slack.com/archives/CF8SMALS1/p1721888905449329

shellyyang1989:
/lgtm

Just a suggestion, would it be better to split it to y and z? For y stream, we need to skip tp as well.

PR #54878, opened 2024-07-26
stackrox: release-x.y ocp 4.15 -> 4.16

davdhacs:
ocp4.16 has been GA for 1 month
Previously delayed for ocp GA and acs 4.5 release.


tommartensen (commented):
Question: While I understand that we want to have all tests on the same OCP version stream, why were some on `fast` before?

tommartensen (approved):
/lgtm

PR #52008, opened 2024-05-14
Adding 416 gcp support

paigerube14:
Need to have the same cloud configuration that we ran in 4.15 on 4.16

PR #54898, opened 2024-07-29
Run zone-outages as part of the same workflow

chaitanyaenr:
This will save costs as well as time for the runs. This commit also renames cloud-api-tests to node-tests to align with the naming in the other configs.


paigerube14:
@chaitanyaenr I was working on updating the gcp and azure runs to use the node workflow here: https://github.com/openshift/release/pull/52008

The zone outage right now is only for aws so that workflow isn't applicable to gcp or azure yet

chaitanyaenr:
> @chaitanyaenr I was working on updating the gcp and azure runs to use the node workflow here: #52008
> 
> The zone outage right now is only for aws so that workflow isn't applicable to gcp or azure yet

@paigerube14 Great! So basically we need to run the zone-outage as a separate step ( test:
    - ref: redhat-chaos-zone-outage ) on the same cluster as aws node scenarios instead of a separate workflow which includes installation leading to higher costs + time duration. Thoughts?

PR #54944, opened 2024-07-29
Update HyperShift CLI for creating nested HCP mgmt clusters

Patryk-Stefanski:
Updating Hypershift CLI to 4.17 for creating nested MGMT clusters. This is to fix our e2e which is failing due atm due to a mismatch in CRDs between 4.16 and 4.17 CLI.


sjenning:
blocked on https://github.com/openshift/hypershift/pull/4445 for now

PR #54947, opened 2024-07-29
TRT-1754: Enable resource watch observer for some multiarch jobs

xueqzhan:


p0lyn0mial:
@xueqzhan thanks for this pr. FYI `tide` says it is `Not mergeable` because `Must be by author openshift-bot OR Needs rehearsals-ack label.`

PR #54951, opened 2024-07-29
add art-members to ci namespace

ashwindasr:
To allow ART members to see build runs in `ci` namespace in app.ci


ashwindasr:
/hold

does not include builds / pods and its related permissions

PR #54958, opened 2024-07-30
Fix stream version resolution for upgrade job

bradmwilliams:
The current configuration is attempting to resolve a stream name: `4.16.0-0.okd-scos-scos`, which is not correct.  This PR updates the configuration to resolve: `4.16.0-0.okd-scos` instead.

### DIFF
--- a/core-services/release-controller/_releases/release-okd-scos-4.17.json
+++ b/core-services/release-controller/_releases/release-okd-scos-4.17.json
@@ -36,7 +36,7 @@
       "upgrade": true,
       "upgradeFromRelease": {
         "candidate": {
-          "stream": "okd-scos",
+          "stream": "okd",
           "version": "4.16"
         }
       }


core-services/release-controller/_releases/release-okd-scos-4.17.json
```diff
@@ -36,7 +36,7 @@
       "upgrade": true,
       "upgradeFromRelease": {
         "candidate": {
-          "stream": "okd-scos",
+          "stream": "okd",
```
Prashanth684, line 39: so where is the `-scos` coming from ?

bradmwilliams: The logic that assembles the stream name is located here:
https://github.com/openshift/release-controller/blob/c556d485536ef41361f681c950b2fc6fb8d3b16b/cmd/release-controller/sync_verify.go#L221
So, in this case, the `-scos` comes from the configuration here:
https://github.com/openshift/release/blob/4e3137cb95921c9fe3b00b29ad250829e99a22f5/core-services/release-controller/_releases/release-okd-scos-4.17.json#L3

PR #54973, opened 2024-07-30
Renaming jobs

josecastillolema:
- Renaming of jobs:
   - **_openshift-qe-ocp-qe-perfscale-ci-main__metal-4.17-latest-x86.yaml_** -> **_openshift-qe-ocp-qe-perfscale-ci-main__metal-dev-latest-x86.yaml_**
   - All the jobs that do not depend on deploy ⇒ move them to **_openshift-qe-ocp-qe-perfscale-ci-main__metal-x86.yaml_**
   I.e:
     - **_pull-ci-openshift-qe-ocp-qe-perfscale-ci-main-metal-4.16-latest-x86-data-path-6nodes_** -> deploys the cluster on 4.16 and runs the corresponding test
     - **_pull-ci-openshift-qe-ocp-qe-perfscale-ci-main-metal-x86-data-path-6nodes_** -> runs the corresponding test on the preexisting cluster
 - Fix issues when deploying dev releases on weekly Wednesdays
 - Adds slack messaging integration for BM jobs

### DIFF
--- a/ci-operator/jobs/openshift-qe/ocp-qe-perfscale-ci/openshift-qe-ocp-qe-perfscale-ci-main-periodics.yaml
+++ b/ci-operator/jobs/openshift-qe/ocp-qe-perfscale-ci/openshift-qe-ocp-qe-perfscale-ci-main-periodics.yaml
@@ -4468,6 +4468,17 @@ periodics:
     job-release: "4.17"
     pj-rehearse.openshift.io/can-be-rehearsed: "true"
   name: periodic-ci-openshift-qe-ocp-qe-perfscale-ci-main-metal-dev-latest-x86-weekly-6nodes
+  reporter_config:
+    slack:
+      channel: '#ocp-qe-scale-ci-results'
+      job_states_to_report:
+      - success
+      - failure
+      - error
+      report_template: '{{if eq .Status.State "success"}} :white_check_mark: Job *{{.Spec.Job}}*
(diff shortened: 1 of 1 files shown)


rsevilla87 (approved):
Idea looks good to me, let's get some rehearsals done before merging

qiliRedHat:
@josecastillolema I have some questions other than comment.
1. The job name 'weekly-6nodes' is different from our common naming convention, I think that's because the tests in those job are not a single control-plane, data-path or a workload. What are the thoughts here for what tests are chosen and what are put together into one job.
2. For 'All the jobs that do not depend on deploy', what is the consideration to have a preexisting cluster and run tests on it? Is that to save time of deploy and be more flexible when doing some manual triggered tests?

josecastillolema:
Hi @qiliRedHat :
 1. You are right, this is a special job in the sense that deploys the cluster and then runs a suite of various tests on top of it. Usually, in cloud environments, you get a fresh cluster and run the test. In baremetal we can work slightly different, looking into having the cluster living longer and finding different kinds of issues. This is part of an ongoing discussion and we don't have the final answer, but for now we would like to explore this approach.
 2. Exactly, the deploy takes approximately 1h30 minutes. The idea is to be flexible and allow the user to decide if he wants to deploy or just run the test directly in the existing cluster, i.e. when doing manual triggered tests as you said.

PR #54987, opened 2024-07-30
Change hub cluster

shaior:


natifridman:
Do we need to rehearse this change or it's using the same credentials?

natifridman (approved):
/lgtm

PR #54997, opened 2024-07-30
no-jira: Remove redundant installer tests

barbacbd:
** Remove GCP altifra tests that are the default when the tech preview tag is no longer required.


r4f4:
We don't want to remove those while the altinfra image exists. As long as no more than one GCP test triggers automatically, there is no harm in keeping them.

PR #55003, opened 2024-07-30
PERFSCALE-3290 - Moving crd-scale to the end so it doesn't interfere …

afcollins:
…with node-density

crd-scale was inserted between cdv2 and node-density.

We have enough failed runs so far, and I have a strong enough suspicion that it is causing chaos with node-density, which we actually use for regression detection.

Any new workloads bundled within control-plane should be appended so as to not interfere with existing tests we rely on.


afcollins:
/pj-rehearse periodic-ci-openshift-qe-ocp-qe-perfscale-ci-main-aws-4.15-nightly-x86-payload-control-plane-6nodes

afcollins:
/pj-rehearse periodic-ci-openshift-qe-ocp-qe-perfscale-ci-main-aws-4.15-nightly-x86-payload-control-plane-6nodes

PR #55007, opened 2024-07-30
OSASINFRA-3555: hypershift: openstack new workflow

EmilienM:
3 commits:
- hypershift: make agent/metallb step re-usable
- hypershift: move agent/create/metallb to hypershift
- hypershift/openstack: new workflow

The last one does the following:
* Use a dedicated workflow for OpenStack, based on nested AWS workflow.
* Create a HostedCluster
* Runs day 2 operations for Ingress and DNS
* Test with conformance
* Dump and destroy resources

These steps will help us until we get Ingress figured out for the OpenStack platform when running Hypershift.
This aims to test a complete workflow of what has been documented for HCP on OpenStack in dev-preview.


mdbooth:
I'm not a fan of this. This is an enormous amount of copy/paste just to enable day 2 operations. I'm convinced we can implement just the functionality we need within the existing framework. I would prefer we did not merge this. I would prefer not to have to maintain this.

Also, although I'd accept it for dev-preview, I don't understand why we need MetalLB. From a networking standpoint it is identical to keepalived/haproxy. Adding MetalLB adds complexity and hides the fact that we don't understand why keepalived is not deploying correctly, which we will eventually have to understand.

PR #55062, opened 2024-07-31
[INTEROP-7456][INTEROP-7458] Update GitOps config to run FIPS; add `slack_reporter` to config.

dfrazzette:


dfrazzette:
The rehearsal job failed with test failures not related to PR changes; these are also failing in periodic/production runs, and can be debugged in weekly job run.

PR #53677, opened 2024-06-26
OCPQE-23057: Enhance hcp ci and ocp ci reuse idp step

xingxingxia:
This PR removes folder `ci-operator/step-registry/cucushift/hypershift-extended/idp-htpasswd` , merge it into folder `ci-operator/step-registry/idp/htpasswd` to avoid duplicate maintenance of two folders that are mostly same with only differences in fields' names.
@dpuniaredhat @heliubj18 , will ask you to review once the Prow tests pass.

PR #55074, opened 2024-08-01
NO-JIRA: Restore my OWNERS permissions which were inadvertently removed

fxierh:
Restore my OWNERS permissions which were inadvertently removed #53677. 

ref: https://github.com/openshift/release/pull/53677#discussion_r1694475234


fxierh:
/cc @xingxingxia 

Could you please help to approve this

PR #55106, opened 2024-08-01
fix: konflux-e2e: update regex

psturc:
we shouldn't be triggering openshift-ci jobs based on the changes in following directories:
* .vscode
* integration-tests
* pipelines
* templates

this PR also removes skipping CI in case files under `tests/release/pipelines/` directory are changed

### DIFF
--- a/ci-operator/config/konflux-ci/e2e-tests/konflux-ci-e2e-tests-main.yaml
+++ b/ci-operator/config/konflux-ci/e2e-tests/konflux-ci-e2e-tests-main.yaml
@@ -46,7 +46,7 @@ tests:
     product: ocp
     timeout: 1h0m0s
     version: "4.14"
-  skip_if_only_changed: ^docs/|^\.github/|^\.tekton/|\.md$|^tests/release/pipelines/|^tests/load-tests/|^cmd/loadTests.go$|^(?:.*/)?(?:\.gitignore|OWNERS|OWNERS_ALIASES|PROJECT|LICENSE)$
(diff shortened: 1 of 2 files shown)


ci-operator/config/konflux-ci/e2e-tests/konflux-ci-e2e-tests-main.yaml
```diff
@@ -46,7 +46,7 @@ tests:
     product: ocp
     timeout: 1h0m0s
     version: "4.14"
-  skip_if_only_changed: ^docs/|^\.github/|^\.tekton/|\.md$|^tests/release/pipelines/|^tests/load-tests/|^cmd/loadTests.go$|^(?:.*/)?(?:\.gitignore|OWNERS|OWNERS_ALIASES|PROJECT|LICENSE)$
```
dheerajodha, line 49: why was `release` removed from this pattern?

dheerajodha: because it was pointing to a subdirectory?

psturc: hmm I assumed that we no longer need to skip triggering e2e jobs when `tests/release/pipelines/` test files are changed thanks to the rule engine Danny has implemented.

@jinqi7 can you confirm, or is it not ready yet and I should revert that change?

jinqi7: If the test files in `tests/release/pipelines/` directory  are changed, we need to trigger e2e-tests jobs.

psturc: thanks, so I'll keep it

psturc: could you please add `/lgtm` if the change looks ok?

PR #55110, opened 2024-08-01
MULTIARCH-4860 Multiarch CI missing upgrade jobs

tvardema:
The automation when branching seems to miss our nightly upgrade jobs, so we're adding them in manually.


Shilpa-Gokul (commented):
/lgtm

PR #55128, opened 2024-08-02
Remove the OKD tests from the samples operator.

metlos:
Samples Operator doesn't doesn't support OKD anymore so there's no point in having these (failing) tests.

NOTE: As far as I understand the YAML files, the okd-based image was being promoted to OpenShift Origin 4.17. Because, as far as I gathered, OKD images are now replaced by the SCOS images, I tried to replace the okd-based image in origin with the scos-based one. I have no clue what and why I'm doing here though, so I would appreciate a pair of knowledgeable eyes on this to check that this actually makes sense.


metlos:
Hmm... so the rehearsal fails for the very same reason that we're trying to solve by removing OKD from our CI. I'm way out of my knowledge here....

ci-operator/config/openshift/cluster-samples-operator/openshift-cluster-samples-operator-master__okd-scos.yaml
```diff
@@ -26,17 +26,12 @@ images:
   to: cluster-samples-operator
 promotion:
   to:
-  - name: "4.17"
+  - name: scos-4.17
```
Prashanth684, line 29: the `ocp_builder_stream9` above should we replaced with the origin one:
```
 origin_scos-4.17_base-stream9:
    name: scos-4.17
    namespace: origin
    tag: base-stream9
```

Prashanth684:
@metlos to give you some context - okd is undergoing a rework: https://issues.redhat.com/browse/OKD-210 where we are building it on top of scos. This is an ongoing effort. but changes like this are greatly appreciated if each team can handle its own jobs.

ci-operator/config/openshift/cluster-samples-operator/openshift-cluster-samples-operator-master__okd-scos.yaml
```diff
@@ -26,17 +26,12 @@ images:
...
-  initial:
-    integration:
-      name: "4.17"
-      namespace: origin
   latest:
     integration:
-      include_built_images: true
-      name: "4.17"
+      name: scos-4.17
       namespace: origin
```
Prashanth684, line 35: this will add the created image to the scos-4.17 imagestream. To mirror to quay you will have to add it in the mapping below.

core-services/image-mirroring/openshift/mapping_origin_4_17
```diff
@@ -1,5 +1,4 @@
 registry.ci.openshift.org/origin/4.17:baremetal-installer quay.io/openshift/origin-baremetal-installer:4.17 quay.io/openshift/origin-baremetal-installer:4.17.0 quay.io/openshift/origin-baremetal-installer:latest
-registry.ci.openshift.org/origin/4.17:cluster-samples-operator quay.io/openshift/origin-cluster-samples-operator:4.17 quay.io/openshift/origin-cluster-samples-operator:4.17.0 quay.io/openshift/origin-cluster-samples-operator:latest
```
Prashanth684, line 2: i would recommend that once you have verified that the jobs pass and the image is created in registry.ci, i.e `registry.ci.openshift.org/origin/scos-4.17:cluster-samples-operator`, you can add the mirroring stanza

ci-operator/config/openshift/cluster-samples-operator/openshift-cluster-samples-operator-master__okd-scos.yaml
metlos: Thanks, this seems to have worked! 🥇

core-services/image-mirroring/openshift/mapping_origin_4_17
metlos: Since this file is autogenerated, how do I reference the scos image here?

ci-operator/config/openshift/cluster-samples-operator/openshift-cluster-samples-operator-master__okd-scos.yaml
metlos: Thanks, this seems to have worked! 🥇

ci-operator/jobs/openshift/cluster-samples-operator/openshift-cluster-samples-operator-master-presubmits.yaml
```diff
@@ -609,550 +609,6 @@ presubmits:
         secret:
           secretName: result-aggregator
     trigger: (?m)^/test( | .* )images,?($|\s.*)
-  - agent: kubernetes
```
Prashanth684, line 612: you want to add equivalent okd-scos jobs here, by defining the test config in `openshift-cluster-samples-operator-master__okd-scos.yaml`

metlos:
Hmm, so the only way I found to keep the image of cluster-samples-operator in the origin image mapping is to replicate the `name` (4.17) and `namespace` (origin) in its `releases`.

The rehearsal passes if the name is `scos-4.17` but fails when the name is `4.17`. This leads me to believe that there is some kind of interplay between our outdated repository and the rehearsal.

So there's a chicken-and-egg problem I don't know the way out of:
1) If we remove the image mirroring to `origin` (which I think we don't want to), we get a passing build.
2) If we put the SCOS image into image mirrors such that it replaces our old image, we get a rehearsal failure.

I know I'm going in circles here and I'm sorry but I think I still need help :(

Prashanth684:
/pj-rehearse pull-ci-openshift-cluster-samples-operator-master-okd-scos-images

ci-operator/config/openshift/cluster-samples-operator/openshift-cluster-samples-operator-master__okd-scos.yaml
```diff
@@ -26,18 +23,28 @@ images:
   to: cluster-samples-operator
 promotion:
   to:
-  - name: scos-4.17
+  - name: "4.17"
```
Prashanth684, line 26: these should be scos-4.17

Prashanth684:
> Hmm, so the only way I found to keep the image of cluster-samples-operator in the origin image mapping is to replicate the `name` (4.17) and `namespace` (origin) in its `releases`.
> 
> The rehearsal passes if the name is `scos-4.17` but fails when the name is `4.17`. This leads me to believe that there is some kind of interplay between our outdated repository and the rehearsal.
> 
> So there's a chicken-and-egg problem I don't know the way out of:
> 
> 1. If we remove the image mirroring to `origin` (which I think we don't want to), we get a passing build.
> 2. If we put the SCOS image into image mirrors such that it replaces our old image, we get a rehearsal failure.
> 
> I know I'm going in circles here and I'm sorry but I think I still need help :(

All references need to be to the scos-4.17 imagestream. "4.17" would refer to the okd fcos imagestream which we are moving away from. Once we get the job passing, we can figure out the mirroring.

ci-operator/config/openshift/cluster-samples-operator/openshift-cluster-samples-operator-master__okd-scos.yaml
```diff
@@ -26,18 +23,28 @@ images:
...
 releases:
+  initial:
+    integration:
+      name: "4.17"
+      namespace: origin
   latest:
     integration:
-      name: scos-4.17
+      include_built_images: true
+      name: "4.17"
```
Prashanth684, line 36:
```suggestion
      name: "scos-4.17"
```

metlos:
PR https://github.com/openshift/release/pull/55408 did a very similar thing to what I've been trying to do here, so I think it is safe to close this.

PR #54888, opened 2024-07-26
OCPQE-24286: [ARO-HCP] Support workflow-launching an azure-aks-hypershift-byo-vnet cluster

fxierh:
A template job has been added to `infra-periodics.yaml` to enable the use of the `azure-qe` cluster profile within `workflows-config.yaml`. 

This allows us to operate within our own budget and use our usual `BASE_DOMAIN` and `HYPERSHIFT_BASE_DOMAIN`.

PR #55131, opened 2024-08-02
OCPQE-24437: Support workflow-launching a hosted cluster on AWS using QE cluster profile

fxierh:
A template job has been added to `infra-periodics.yaml` to enable the use of the `aws-qe` cluster profile within workflows-config.yaml. 

This allows us to operate within our own budget and use our usual `BASE_DOMAIN` and `HYPERSHIFT_BASE_DOMAIN`.

Note: Similar changes were made for Azure in PR #54888, and Azure-hosted clusters can now be successfully workflow-launched using the expected cluster profile.

### DIFF
--- a/core-services/ci-chat-bot/workflows-config.yaml
+++ b/core-services/ci-chat-bot/workflows-config.yaml
@@ -259,8 +259,8 @@ workflows:
     platform: aws
   cucushift-installer-rehearse-aws-ipi-ovn-custom-cidr:
     platform: aws
-  cucushift-installer-rehearse-aws-ipi-ovn-hypershift-cluster-bot:
-    platform: aws
+  cucushift-installer-rehearse-aws-ipi-ovn-hypershift:
+    platform: aws-qe
     base_images:
       hypershift-operator:
         name: hypershift-operator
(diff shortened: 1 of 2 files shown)


core-services/ci-chat-bot/workflows-config.yaml
```diff
@@ -259,8 +259,8 @@ workflows:
     platform: aws
   cucushift-installer-rehearse-aws-ipi-ovn-custom-cidr:
     platform: aws
-  cucushift-installer-rehearse-aws-ipi-ovn-hypershift-cluster-bot:
-    platform: aws
+  cucushift-installer-rehearse-aws-ipi-ovn-hypershift:
+    platform: aws-qe
```
bradmwilliams, line 263: Before this change can merge, you'll need to add `aws-qe` as a support platform in the clusterbot:
https://github.com/openshift/ci-chat-bot/blob/5305b6ceebe072fb2918b487b8747e109702e510/pkg/manager/prow.go#L56
or the workflow will almost immediately fail with an unsupported platform error.

bradmwilliams (commented):
/hold
Needs changes to the clusterbot before this workflow will work.

bradmwilliams (approved):

core-services/ci-chat-bot/workflows-config.yaml
fxierh: Thanks for the comment

fxierh: For future reference: 
This works because workflow-launch uses different validation logic. As [discussed](https://redhat-internal.slack.com/archives/CNHC2DK2M/p1722609379193539) in the channel, we'll keep the current SupportedPlatforms unchanged to avoid exposing the -qe platforms to the public.

PR #55139, opened 2024-08-02
Use `hyperdisk-balanced` disk on build02 infra nodes

danilo-gemoli:
Build02 infra nodes use machines of type `c4-highmem-32` that support Hyperdisks only: https://cloud.google.com/compute/docs/general-purpose-machines#c4-highmem

/cc @deepsm007 @bear-redhat


danilo-gemoli:
/hold
until we figure out how to work around this issue
```
machinesets.machine.openshift.io/build0-gstfj-w-b-infra
error: machinesets.machine.openshift.io "build0-gstfj-w-b-infra" could not be patched: admission webhook "validation.machineset.machine.openshift.io" denied the request: providerSpec.disks[0].type: Unsupported value: "hyperdisk-balanced": supported values: "pd-balanced", "pd-ssd", "pd-standard"
```

PR #55154, opened 2024-08-02
Use external network ID in Hypershift cmdline

MaysaMacedo:
We need to pass the external network ID instead of name for the hypershift command line. This commit achieves that by looking for the ID of the external network once the network is fetched from OpenStack by name.


EmilienM:
since your PR has been approved and is going to merge I think we can safely land it
/lgtm
/approve

ci-operator/step-registry/hypershift/openstack/e2e/execute/hypershift-openstack-e2e-execute-commands.sh
```diff
@@ -36,6 +36,6 @@ hack/ci-test-e2e.sh \
         --e2e.platform="OpenStack" \
 	--e2e.ssh-key-file="${CLUSTER_PROFILE_DIR}/ssh-publickey" \
         --e2e.openstack-credentials-file="${SHARED_DIR}/clouds.yaml" \
-        --e2e.openstack-external-network-name="${OPENSTACK_EXTERNAL_NETWORK}" \
+        --e2e.openstack-external-network-name="${OPENSTACK_EXTERNAL_NETWORK_ID}" \
```
EmilienM, line 39:
```suggestion
        --e2e.openstack-external-network-id="${OPENSTACK_EXTERNAL_NETWORK_ID}" \
```

PR #54851, opened 2024-07-26
[azure] enable CAPI on 4.17 jobs - part1

jinyunma:
Enable CAPI install for following 4.17 CI jobs:
openshift-openshift-tests-private-release-4.17__multi-nightly.yaml: 
```
./openshift-openshift-tests-private-release-4.17__multi-nightly.yaml-- as: azure-ipi-amd-f7-netobserv
./openshift-openshift-tests-private-release-4.17__multi-nightly.yaml-- as: azure-ipi-amd-mixarch-f28-destructive
./openshift-openshift-tests-private-release-4.17__multi-nightly.yaml-- as: azure-ipi-basecap-vset-additionalcap-amd-f28-destructive
./openshift-openshift-tests-private-release-4.17__multi-nightly.yaml-- as: azure-ipi-basecap-vset-amd-f28-destructive
./openshift-openshift-tests-private-release-4.17__multi-nightly.yaml-- as: azure-ipi-des-day2-fips-amd-f28-destructive
./openshift-openshift-tests-private-release-4.17__multi-nightly.yaml-- as: azure-ipi-des-mini-perm-fips-amd-f28-destructive
./openshift-openshift-tests-private-release-4.17__multi-nightly.yaml-- as: azure-ipi-ingress-controller-amd-mixarch-f28-destructive
./openshift-openshift-tests-private-release-4.17__multi-nightly.yaml-- as: azure-ipi-ingress-custom-controller-fips-amd-f28-destructive
./openshift-openshift-tests-private-release-4.17__multi-nightly.yaml-- as: azure-ipi-oidc-amd-f28-destructive
./openshift-openshift-tests-private-release-4.17__multi-nightly.yaml-- as: azure-ipi-ovn-etcd-encryption-compact-amd-f28-destructive
./openshift-openshift-tests-private-release-4.17__multi-nightly.yaml-- as: azure-ipi-ovn-ipsec-azurefile-csi-amd-f28-destructive
./openshift-openshift-tests-private-release-4.17__multi-nightly.yaml-- as: azure-ipi-private-spec-net-type-spec-perm-amd-f28-destructive
./openshift-openshift-tests-private-release-4.17__multi-nightly.yaml-- as: azure-ipi-sno-etcd-encryption-amd-f28-destructive
./openshift-openshift-tests-private-release-4.17__multi-nightly.yaml-- as: azure-ipi-ultrassd-disks-fips-amd-f28-destructive

```
[...]

PR #55170, opened 2024-08-05
[azure] remove feature gates on 4.17 jobs with mixarch

jinyunma:
Cont. to https://github.com/openshift/release/pull/54851

Installer storage account name is changed to use "infraID" as prefix with CAPI-based installation, which cause the the step [`ipi-install-heterogeneous`](https://github.com/openshift/release/blob/master/ci-operator/step-registry/ipi/install/heterogeneous/ipi-install-heterogeneous-commands.sh#L241) failed.

```
azure-ipi-ingress-controller-amd-mixarch-f28-destructive
azure-ipi-amd-mixarch-f28-destructive
```

We suggest to change back to use "cluster" in bug https://issues.redhat.com/browse/OCPBUGS-37953, so temporarily removed feature gates for 4.17 mixarch jobs to unblock e2e test, and wait for the fix of bug OCPBUGS-37953 to decide whether needs to update step `ipi-install-heterogeneous` or not.


jinyunma:
cc @lwan-wanglin @jianlinliu @gpei to review

jinyunma:
/pj-rehearse ack

cc @liangxia to review

PR #55178, opened 2024-08-05
NO-ISSUE: build image-service image in CI with centos stream 9

danmanor:
This PR is a leftover from https://github.com/openshift/assisted-image-service/pull/266


adriengentil:
can you remove other references to centos 8 ? We still have it here for example: https://github.com/openshift/release/blob/master/ci-operator/config/openshift/assisted-image-service/openshift-assisted-image-service-main.yaml#L10

danmanor:
@adriengentil Yes, done

adriengentil (approved):

PR #55182, opened 2024-08-05
openstack: Fetch the network ID in a separate step

pierreprinetti:
Before this patch, the step openstack-conf-resourcenames attempted to reach the OpenStack API before a clouds.yaml was available. This resulted in all jobs permafailing.

With this patch, the act of fetching the external network ID based on the external network name is delayed until after a clouds.yaml is available.

The external network ID is necessary for the hypershift workflow.

The external network ID is not hardcoded in openstack-conf-resourcenames because it is susceptible to change on clouds that are reprovisioned often (like mecha-central).


pierreprinetti:
The two end-to-end tests passed install and are running the conformance suite, which means that the issue was fixed. Let's just run this last job (of another kind) and then I think we can merge

/pj-rehearse periodic-ci-shiftstack-ci-main-quotas-vexxhost

EmilienM:
Adding a step for that seems overkill to me, but I'm fine with that.
Now we need to add that step to the HyperShift chain.

pierreprinetti:
> Adding a step for that seems overkill to me

correct >>> optimised

PR #55208, opened 2024-08-05
OSASINFRA-3556: Updating the openstack-vh-mecha-az0-quota-slice to 4

MiguelCarpio:
Update openstack-vh-mecha-az0-quota-slice to admit four concurrent general jobs on mecha-az0


mandre (commented):
/lgtm

PR #55211, opened 2024-08-05
hypershift/openstack: fix base domain

EmilienM:
The base domain is in fact `origin-ci-int-aws.dev.rhcloud.com`
in the aws-nested environment, we need to stop hardcoding it from the
execute command and use $BASE_DOMAIN.


EmilienM:
/pj-rehearse ack
No need to run rehearse on this one because it'll fail since we need https://github.com/openshift/hypershift/pull/4474

PR #55221, opened 2024-08-06
hypershift/openstack: export cluster name

EmilienM:
We have to specify a cluster name for e2e tests because at this point we already
created the DNS record for the HostedCluster Ingress endpoint.
Because of that we can't run more than one test in parallel.

This is very likely a temporary hack until we find a better solution.


ci-operator/step-registry/hypershift/openstack/e2e/execute/hypershift-openstack-e2e-execute-commands.sh
```diff
@@ -34,6 +46,7 @@ hack/ci-test-e2e.sh \
         --e2e.base-domain="${BASE_DOMAIN}" \
         --e2e.external-dns-domain="service.${BASE_DOMAIN}" \
         --test.run='^TestCreateCluster$' \
+	--test.parallel=1 \
```
EmilienM, line 49: here I assume that's the number of concurrent `HostedClusters` that can be deployed at the same time.

PR #55229, opened 2024-08-06
Do not run e2e tests always in Pull Requests

flacatus:
Due to konflux migration, lets do a step away in openshift ci and only run e2e on comments


ci-operator/config/konflux-ci/e2e-tests/konflux-ci-e2e-tests-main.yaml
```diff
@@ -36,7 +36,8 @@ resources:
       cpu: 100m
       memory: 200Mi
 tests:
-- as: redhat-appstudio-e2e
+- always_run: false
```
psturc, line 39: not sure if that changes anything... the job [already has `always_run: false` set](https://github.com/openshift/release/blob/7bb86d31f805b0cd4a8aedae92ebdb1fb5e9baa5/ci-operator/jobs/konflux-ci/e2e-tests/konflux-ci-e2e-tests-main-presubmits.yaml#L545-L550)

I think we should probably remove [this line](https://github.com/openshift/release/pull/55229/files#diff-da92c0d36b5ef11d26abf7debafeeb4aed6b766d61bd9eefb9aba76b4eeee571R50) and set `optional: true` for the job config

PR #55247, opened 2024-08-06
Bump expected recovery time to 120s

chaitanyaenr:
This is needed to provide additional buffer time as the components to disrupt are picked randomly. 120 seconds recovery time is a good target to avoid failing where expected in addition to flagging components taking long time to recover.


paigerube14 (approved):
/lgtm

shahsahil264 (approved):
LGTM

PR #55267, opened 2024-08-07
Update vsphere cluster profile

weinliu:
Update it to use DEVQEdatacenter instead of cidatacenter


weinliu:
@liangxia coud you help? Thanks!

PR #55286, opened 2024-08-07
Add configuration for Python 3.11 based notebooks

caponetto:
Depends on https://github.com/opendatahub-io/notebooks/pull/659

### DIFF
--- a/ci-operator/config/opendatahub-io/notebooks/opendatahub-io-notebooks-main.yaml
+++ b/ci-operator/config/opendatahub-io/notebooks/opendatahub-io-notebooks-main.yaml
@@ -184,6 +184,132 @@ images:
   dockerfile_path: Dockerfile
   from: rocm-ubi9-python-3.9
   to: runtime-rocm-tensorflow-ubi9-python-3.9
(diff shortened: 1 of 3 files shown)


jiridanek:
just a reminder that two PRs will be needed, one for opendatahub-io and another for red-hat-data-services, same as in 2024a

* https://github.com/openshift/release/pull/49709
* https://github.com/openshift/release/pull/49706

jiridanek:
What's `pull-ci-openshift-release-master-build11-dry`? I dont remember seeing this before.

jiridanek:
@caponetto looks good to me, tests failed because rocm image was not built, which is known issue; wdyt?

caponetto:
> @caponetto looks good to me, tests failed because rocm image was not built, which is known issue; wdyt?

I agree. I thought the `*-e2e-tests` (anaconda, codeserver, habana) were supposed to pass since they don't have to do with rocm.

jiridanek:
The images test always tries to build all images and if that fails, the dependent tests automatically fail. Harshad once mentioned it would be nice if we could improve this but tbh I don't understand openshift či enough to jump in and just do it.

PR #55346, opened 2024-08-08
DNM - nodepool e2e tests for HCP on OSP 

EmilienM:
I wanted to check whether they pass and how much time it'll take since we run them in serial.


EmilienM:
We might have to skip one Nodepool test too like it's done in https://github.com/openshift/hypershift/pull/3562 
In the meantime I'll try to increase the timeout.

EmilienM:
i think the NTO test will fail because our flavor doesn't have hugepages. I requested it to our OpenStack cloud provider.

PR #55348, opened 2024-08-08
Remove redundant unit test from siteconfig

dislbenn:
Since `siteconfig` was onboarded with `konflux` we can remove the `unit-test` job from the prow builds.


carbonin:
I'd prefer to wait for @sakhoury to take a look at this one.
Right before his time off he was working on some openshift CI enablement plus he did all of the konflux work so it would probably be best for him to ensure this is okay.

sakhoury (approved):
/lgtm

PR #55369, opened 2024-08-08
Fix issue DPTP-4131 Add HyperConverged CR

bear-redhat:


bear-redhat:
retest next week, when b03/05 upgraded to 4.16

PR #55372, opened 2024-08-08
Add a presubmit for node removal job

carbonin:
This is optional and not run by default, but would be useful to have when editing logic involved in the node removal process

https://issues.redhat.com/browse/MGMT-10006


CrystalChun (approved):
/lgtm

PR #55383, opened 2024-08-09
Revert "Update vsphere cluster profile"

weinliu:
Reverts openshift/release#55267
Revert since it's not working as expected


weinliu:
@liangxia could you help ?Thanks!

PR #55390, opened 2024-08-09
CFE-1099: Use RHEL8 for ansible-operator-plugins image

chiragkyal:
**Description of the change:**

- Update builder to use rhel-8-golang-1.22-openshift-4.17 to fix `GLIBC` version issue.
- Remove `no-qe` in favour of https://github.com/openshift/ansible-operator-plugins/pull/7#issuecomment-2276243875

**Motivation for the change:**
builder and base image should use same RHEL version.

### DIFF
--- a/core-services/prow/02_config/openshift/ansible-operator-plugins/_prowconfig.yaml
+++ b/core-services/prow/02_config/openshift/ansible-operator-plugins/_prowconfig.yaml
@@ -73,7 +73,7 @@ tide:
     - approved
     - jira/valid-reference
     - lgtm
-    - qe-approved,no-qe
+    - qe-approved
     missingLabels:
     - backports/unvalidated-commits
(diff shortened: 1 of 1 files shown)


core-services/prow/02_config/openshift/ansible-operator-plugins/_prowconfig.yaml
```diff
@@ -73,7 +73,7 @@ tide:
     - approved
     - jira/valid-reference
     - lgtm
-    - qe-approved,no-qe
+    - qe-approved
```
arkadeepsen, line 76: Thanks for removing the `no-qe` label from the prowconfig. It cannot be applied as it not available in the openshift repos. Ref: https://github.com/openshift/ansible-operator-plugins/pull/7#issuecomment-2276243875

PR #55395, opened 2024-08-09
OCPEDGE-799: feat: make aggregated sno upgrade job blocking

eggfoobar:
/hold

Hold until after branch cut.


eggfoobar:
Hey @neisw, if we're good with the branching, this should be good to merge in, wdyt?

PR #55400, opened 2024-08-09
NO-JIRA: Fix incorrect prow configurations of cli manager and cli-manager-operator

ardaguclu:
SSIA

### DIFF
--- a/core-services/prow/02_config/openshift/cli-manager/_prowconfig.yaml
+++ b/core-services/prow/02_config/openshift/cli-manager/_prowconfig.yaml
@@ -1,6 +1,71 @@
 tide:
   queries:
   - includedBranches:
+    - openshift-4.10
+    - openshift-4.11
+    - openshift-4.12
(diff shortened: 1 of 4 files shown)


core-services/prow/02_config/openshift/cli-manager/_prowconfig.yaml
```diff
@@ -1,5 +1,70 @@
...
+    missingLabels:
+    - backports/unvalidated-commits
+    - do-not-merge/hold
+    - do-not-merge/invalid-owners-file
+    - do-not-merge/work-in-progress
+    - jira/invalid-bug
+    - needs-rebase
+    repos:
+    - openshift/cli-manager
+  - includedBranches:
```
ingvagabund, line 50: This part looks to be exactly the same as for the previous `includedBranches`. Should be safe to move `openshift-4.17` and `release-4.17` there. The same for CLMO configuration.

PR #55431, opened 2024-08-11
networking-console-plugin: Add OKD-scos for 4.18 and enable mirroring to origin

orenc1:
In order for the new `networking-console-plugin` (that will be deployed by cluster-network-operator) to work on OKD, we need the networking-console-plugin image of the okd-scos variant to be mirrored into `quay.io/openshift/origin-networking-console-plugin:latest`.
We need the mirroring enabled to unblock https://github.com/openshift/cluster-network-operator/pull/2332.

### DIFF
--- a/ci-operator/config/openshift/networking-console-plugin/openshift-networking-console-plugin-release-4.17__okd-scos.yaml
+++ b/ci-operator/config/openshift/networking-console-plugin/openshift-networking-console-plugin-release-4.17__okd-scos.yaml
@@ -26,12 +26,12 @@ images:
   to: networking-console-plugin
 promotion:
   to:
-  - name: "4.18"
+  - name: "4.17"
     namespace: origin
 releases:
   latest:
     integration:
-      name: "4.18"
+      name: "4.17"
       namespace: origin
 resources:
   '*':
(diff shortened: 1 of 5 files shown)


ci-operator/config/openshift/networking-console-plugin/openshift-networking-console-plugin-main__okd-scos.yaml
```diff
@@ -26,12 +26,12 @@ images:
   to: networking-console-plugin
 promotion:
   to:
-  - name: scos-4.17
+  - name: "4.18"
```
jupierce, line 29: scos-4.x is the appropriate integration stream for OKD. "4.x" was for the old fcos builds and is deprecated.

jupierce:
Looks like some overlap with https://github.com/openshift/release/pull/55469 which moves master to build into scos-4.18.

ci-operator/config/openshift/networking-console-plugin/openshift-networking-console-plugin-main__okd-scos.yaml
orenc1: just for the record, initially i tried putting `scos-4.18`, but the `openshift-image-mirror-mappings` check deleted the entry in question. however, it turns out that actually pushing the image to `quay.io/openshift-origin-*` is not required.

orenc1:
closing as not required, as per @jupierce 
/close

PR #55439, opened 2024-08-12
Remove slack reporting for hypershift/kubevirt

orenc1:
we can now remove the individual jobs slack reports since we have a bot that aggragates all of the periodic results to a single message once a day.


davidvossel (commented):
/lgtm
/approve

PR #55445, opened 2024-08-12
Fix scos jobs in ironic-image and ironic-agent-image

elfosardo:


elfosardo:
actually this is not need anymore

PR #55454, opened 2024-08-12
USHIFT-3520: Disable workload partitioning for conformance

pacevedom:


ci-operator/step-registry/openshift/microshift/infra/conformance-setup/openshift-microshift-infra-conformance-setup-commands.sh
```diff
@@ -1,5 +1,30 @@
...
+  ServerAliveCountMax 1200
+EOF
+chmod 0600 "${HOME}/.ssh/config"
+
 cp /go/src/github.com/openshift/microshift/origin/skip.txt "${SHARED_DIR}/conformance-skip.txt"
 cp "${SHARED_DIR}/conformance-skip.txt" "${ARTIFACT_DIR}/conformance-skip.txt"
+
+# Disable workload partitioning for annotated pods to avoid throttling.
+ssh "${INSTANCE_PREFIX}" "sudo sed -i 's/resources/#&/g' /etc/crio/crio.conf.d/11-microshift-ovn.conf"
+ssh "${INSTANCE_PREFIX}" "sudo systemctl daemon-reload"
```
ggiguash, line 26: This is probably unnecessary as we're restarting crio

ci-operator/step-registry/openshift/microshift/infra/conformance-setup/openshift-microshift-infra-conformance-setup-commands.sh
```diff
@@ -1,5 +1,30 @@
...
+# Just for safety, restart everything from scratch.
+ssh "${INSTANCE_PREFIX}" "echo 1 | sudo microshift-cleanup-data --all"
```
ggiguash, line 28: Add `--keep-images` to save time?

ci-operator/step-registry/openshift/microshift/infra/conformance-setup/openshift-microshift-infra-conformance-setup-commands.sh
```diff
@@ -1,5 +1,30 @@
...
+ssh "${INSTANCE_PREFIX}" "sudo systemctl restart crio"
+ssh "${INSTANCE_PREFIX}" "sudo systemctl start microshift"
```
ggiguash, line 30: Do we want to enable microshift service so that it's restarted by systemd if it crashes?
Or, we start on purpose not to miss crashes?

If it's the latter, let's add a comment please.

pacevedom: It is not enabled on purpose. If it crashes that shouldnt happen and we need to address it, so better to force a failure.

ci-operator/step-registry/openshift/microshift/infra/conformance-setup/openshift-microshift-infra-conformance-setup-commands.sh
pacevedom: Good point. Done

ci-operator/step-registry/openshift/microshift/infra/conformance-setup/openshift-microshift-infra-conformance-setup-commands.sh
pacevedom: I got complaints when restarting it because of not doing it.

PR #55461, opened 2024-08-12
Update ironic builds jobs for OCP 4.18

elfosardo:


elfosardo:
/pj-rehearse ack
I checked all the failures and they don't seem related to this change
we have other 2 PRs open to fix the root image error https://github.com/openshift/ironic-agent-image/pull/144 and https://github.com/openshift/ironic-image/pull/542
this is good to go

PR #55465, opened 2024-08-12
POST-BRANCHING [4.19]: jira validation criteria

Prucek:
Step 1


core-services/prow/02_config/_plugins.yaml
```diff
@@ -214,6 +214,15 @@ bugzilla:
       - 4.19.0
       target_release: 4.18.0
       validate_by_default: true
+    openshift-4.19:
```
jmguzik, line 217: bugzilla not needed, I would not add it and soon we will remove it completely

jmguzik (approved):

PR #55487, opened 2024-08-12
OLS update bundle tag to latest

xrajesh:
Updated the bundle tag to latest so that we can use for prow latest test.


raptorsun:
/cc @williamcaban 
Could you please approve this PR?

PR #55489, opened 2024-08-12
stackrox: release-4.5 has ocp4-16, remove ocp4-15

davdhacs:
We do not need to keep testing on OCP 4.15 on release 4.5 because we have testing on OCP 4.16.


tommartensen (approved):
/lgtm

BradLugo (approved):
/lgtm

PR #55562, opened 2024-08-14
METAL-1122 ofcir: shellcheck acquire and release bash scripts

tdomnesc:
Fix every shellcheck warning on OFCIR acquire and release bash scripts


derekhiggins:
/lgtm 
thanks

PR #55574, opened 2024-08-14
Update firewatch-cluster to add ipi-install-rbac ref

chetna14manku:
Add `ipi-install-rbac` ref to pull images from build clusters


calebevans (approved):
/lgtm

PR #55579, opened 2024-08-14
Increase vexxhost quota slice

EmilienM:
We reduced the vh-managed quota from 18 to 9 in the past because:
* we didn't have much of quota
* swift wasn't available so we had a lot of cinder volumes being created

Now:
* our storage quota is now doubled
* we have swift in vexxhost, so we don't allocate the 100GB storage for the image registry anymore

So we can double our capacity.


mandre (commented):
/lgtm

mandre:
This effectively reverts https://github.com/openshift/release/pull/33087

PR #55586, opened 2024-08-14
OCPQE-24848: Support launching the cucushift-installer-rehearse-azure-aks-hypershift-install workflow

fxierh:


fxierh:
/cc @liangxia 

Could you please take a look ? Thank you

PR #55592, opened 2024-08-14
Added hector-vido

hector-vido:
Added hector-vido to OWNERS_ALIASES file.


deepsm007:
/override ci/build-farm/build11-dry

We are syncing the cluster

PR #55611, opened 2024-08-15
Remove "set -x" in case of ignored sensitive line

xingxingxia:
Originally the global `set -x` was added in place so that any reported failure can be debugged quickly, and individual `set +x`s were added for sensitive lines. But some sensitive lines may be not noticed. Thus remove all anyway.

### DIFF
--- a/ci-operator/step-registry/idp/htpasswd/idp-htpasswd-commands.sh
+++ b/ci-operator/step-registry/idp/htpasswd/idp-htpasswd-commands.sh
@@ -3,7 +3,6 @@
 set -e
 set -u
 set -o pipefail
-set -x
 
 function check_if_hypershift_env () {
     if [ -f "${SHARED_DIR}/nested_kubeconfig" ]; then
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/idp/htpasswd/idp-htpasswd-commands.sh
```diff
@@ -64,12 +61,10 @@ function check_idp () {
...
     current_idp_config=$(oc get $TARGET_RESOURCE -o jsonpath='{range '$IDP_FIELD'[*]}{.name}{" "}{.type}{"\n"}{end}')
-    set +x
     if [ -n "$current_idp_config" ] && [ "$current_idp_config" != "null" ] && [ -n "$USERS" ]; then
         echo -e "Skipping addition of new htpasswd IDP because already configured IDP as below:\n$current_idp_config"
         exit 0
     fi
-    set -x
 }
 
 function set_proxy () {
```
xingxingxia, line 70: This is the so-called "some sensitive lines may be not noticed". Under `set -x`, `source "${SHARED_DIR}/proxy-conf.sh"` prints proxy sensitive info, but this was not noticed.

xingxingxia:
/pj-rehearse skip
@liangxia bother you to help check / lgtm, thanks!

PR #55630, opened 2024-08-15
Kube-compare: fix tigger for report creator tool tests

nocturnalastro:
/cc @lack @AlinaSecret @natifridman

### DIFF
--- a/ci-operator/config/openshift/kube-compare/openshift-kube-compare-main.yaml
+++ b/ci-operator/config/openshift/kube-compare/openshift-kube-compare-main.yaml
@@ -73,7 +73,7 @@ tests:
   commands: GOFLAGS="-mod=readonly" make test-report-creator
   container:
     from: bin
-  run_if_changed: ^addon-tools/report-creator/*$
+  skip_if_only_changed: (^docs/)|((^|/)OWNERS(_ALIASES)?$)|((^|/)[A-Z]+\.md$)
 - as: golint
   commands: |
     make golangci-lint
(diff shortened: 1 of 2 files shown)


ci-operator/config/openshift/kube-compare/openshift-kube-compare-main.yaml
```diff
@@ -73,7 +73,7 @@ tests:
   commands: GOFLAGS="-mod=readonly" make test-report-creator
   container:
     from: bin
-  run_if_changed: ^addon-tools/report-creator/*$
+  skip_if_only_changed: (^docs/)|((^|/)OWNERS(_ALIASES)?$)|((^|/)[A-Z]+\.md$)
```
natifridman, line 76: Maybe we should also add `/vendors` and `/hack`?

lack: I think `/vendors` is safe to exclude; The only time it should change is if go modules have changed, so there would always be an associated change to `go.mod` which should always trigger a build (in case a go module change would break the build and/or test).

`/hack` is trickier; If our CI ever relies on scripts in `/hack` excluding it could be dangerous.  I'm not sure if it does today or not...

nocturnalastro: Hack has scripts for installing and running markdown and go linters.

lack:
/pj-rehearse pull-ci-openshift-kube-compare-main-unit-report-creator

PR #55662, opened 2024-08-16
Revert "mirror `pod-scaler` image for use in consumers on build farm clusters"

smg247:
Reverts openshift/release#55637


smg247:
/close
We need this mirror after all

PR #55681, opened 2024-08-16
`pod-scaler`: update probes and requests for consumers

smg247:


danilo-gemoli:
/lgtm
whenever you're ready to watch the rollout @smg247

PR #55686, opened 2024-08-16
revert multi-arch openshift-velero-plugin 1.0 1.3

kaovilai:
unblocks https://github.com/openshift/openshift-velero-plugin/pull/273

Since `.config.prowgen` in the same dir sets default multi_arch to true, we need to specify multi_arch as false in oadp-1.0 and oadp-1.3 branch to revert behavior.

### DIFF
--- a/ci-operator/config/openshift/openshift-velero-plugin/.config.prowgen
+++ b/ci-operator/config/openshift/openshift-velero-plugin/.config.prowgen
@@ -1,0 +1,1 @@
+multi_arch: yes


kaovilai:
[blocked by test-infra](https://redhat-internal.slack.com/archives/C0144ECKUJ0/p1724343483827169?thread_ts=1723733305.681339&cid=C0144ECKUJ0)

ci-operator/config/openshift/openshift-velero-plugin/.config.prowgen
mateusoliveira43, line 1: from "linter" job, this seems to remove multi-arch from the other branches (tell them that they do not need build10 cluster)

mateusoliveira43 (approved):
only amd images are being created by the jobs

PR #55697, opened 2024-08-19
[OCPQE-25518] Adding slack notification to Hive long duration cases for 4.17.

huangmingxia:
We need to have ci watcher to monitor this channel and do the analysis.

It will be easier for us to search our analysis after long time as Peter's Web UI can only store failure result for 1-2 weeks.


huangmingxia:
@bandrade Please help approval this PR when you have time, thanks.

PR #46354, opened 2023-12-04
remove autoscaling from the default setting

yasun1:
Remove the autoscaling from the default setting.
Remove the useless workflow `rosa-aws-sts-rosacli-e2e-workflow`.
Remove the useless chain `rosa-aws-sts-deprovision-byo-vpc-chain`.

[OCPQE-18428](https://issues.redhat.com//browse/OCPQE-18428)

PR #55723, opened 2024-08-19
Interop - Updated the firewatch-rosa-aws-sts workflow with removing redundant env vars

oharan2:
Main changes:

1. Following https://github.com/openshift/release/pull/46354 which disables the `ENABLE_AUTOSCALING` env var by setting it to false in all ROSA related refs. 
According to [OCPQE-18428](https://issues.redhat.com/browse/OCPQE-18428) setting it to true causes unexpected errors.
2. Also, in our case `DISABLE_WORKLOAD_MONITORING` is redundent since it's overriding the same value in [rosa-cluster-provision](https://github.com/openshift/release/blob/master/ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-ref.yaml#L90).

rh-pre-commit.version: 2.0.3
rh-pre-commit.check-secrets: ENABLED


calebevans (approved):
/lgtm

PR #55728, opened 2024-08-19
MULTIARCH-4963: Update base images and builder to 4.17 and golang 1.22

aleskandro:
Required by https://github.com/openshift/multiarch-tuning-operator/pull/225


aleskandro:
/pj-rehearse ack 

(no need for e2e in this case)

PR #55425, opened 2024-08-09
[WIP] refactor vertical-pod-autoscaler-operator tests for operator-sdk upgrade

maxcao13:
Ignore for now, just checking to see if it works against my branch.

This is so that we can fix tests that have to migrated to the operator-sdk upgrade in the VPA-operator sooner than later since a lot has changed in the repo.

The first commit is to make sure that images get pulled correctly and simple tests (go vet, fmt ,etc. get tested correctly).

Further commits will test the e2e tests that require a cluster.

Which should let us get instant feedback on test changes when we merge VPA upgrade PR instead of having to change there, then come back here and change it.

PR in question here: https://github.com/openshift/vertical-pod-autoscaler-operator/pull/169

PR #55730, opened 2024-08-19
PODAUTO-89: Update config/jobs to vpa operator-sdk migration format

maxcao13:
This PR changes the CI for the VPA-operator CI jobs and config to fit the overhauled VPA-operator operator-sdk upgrade. That PR is detailed here: https://github.com/openshift/vertical-pod-autoscaler-operator/pull/169

We will first merge this PR in first and allow tests to run on that PR to validate it.

I've already created a draft PR on my fork to test these changes. That PR is here: https://github.com/openshift/release/pull/55425

This PR should just be cloning that, but preserving the `openshift/master` instead of `maxcao13/branch-name` and keeping the `promotion` stanza.

Additionally this PR adds a `scorecard` test that runs an operator-sdk scorecard validation test suite against a hypershift cluster. More info here: https://sdk.operatorframework.io/docs/testing-operators/scorecard/#built-in-tests

### DIFF
--- a/ci-operator/config/openshift/vertical-pod-autoscaler-operator/openshift-vertical-pod-autoscaler-operator-master.yaml
+++ b/ci-operator/config/openshift/vertical-pod-autoscaler-operator/openshift-vertical-pod-autoscaler-operator-master.yaml
@@ -1,4 +1,8 @@
 base_images:
+  golangci-lint:
+    name: golangci-lint
+    namespace: ci
+    tag: v1.59.1
   ocp_4.17_base-rhel9:
     name: "4.18"
     namespace: ocp
@@ -7,18 +11,20 @@ base_images:
     name: release
     namespace: openshift
     tag: rhel-9-release-golang-1.22-openshift-4.17
+  operator-sdk:
+    name: "4.18"
+    namespace: origin
+    tag: operator-sdk
 build_root:
   from_repository: true
+  use_build_cache: true
 images:
-- inputs:
-    ocp_4.17_base-rhel9:
-      as:
-      - registry.ci.openshift.org/ocp/4.17:base-rhel9
-    openshift_release_rhel-9-release-golang-1.22-openshift-4.17:
-      as:
-      - registry.ci.openshift.org/openshift/release:rhel-9-release-golang-1.22-openshift-4.17
-  to: vertical-pod-autoscaler-operator
-- dockerfile_path: images/operator-registry/Dockerfile.registry.ci
+- build_args:
+  - name: TARGETOS
+    value: linux
+  - name: TARGETARCH
+    value: amd64
+  dockerfile_path: Dockerfile
   inputs:
     ocp_4.17_base-rhel9:
       as:
@@ -26,7 +32,17 @@ images:
     openshift_release_rhel-9-release-golang-1.22-openshift-4.17:
       as:
       - registry.ci.openshift.org/openshift/release:rhel-9-release-golang-1.22-openshift-4.17
-  to: vpa-operator-registry
+  to: vertical-pod-autoscaler-operator
+operator:
+  bundles:
+  - as: vertical-pod-autoscaler-operator-bundle
+    dockerfile_path: bundle.Dockerfile
+    skip_building_index: true
+  substitutions:
+  - pullspec: quay.io/openshift/origin-vertical-pod-autoscaler-operator.*
+    with: pipeline:vertical-pod-autoscaler-operator
+  - pullspec: quay.io/openshift/origin-vertical-pod-autoscaler.*
+    with: stable:vertical-pod-autoscaler
 promotion:
   to:
   - name: "4.18"
@@ -41,6 +57,11 @@ releases:
       include_built_images: true
       name: "4.18"
       namespace: ocp
+  multilatest:
+    release:
+      architecture: multi
+      channel: stable
+      version: "4.15"
 resources:
   '*':
     requests:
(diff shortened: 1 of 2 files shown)


maxcao13:
Acking since master won't pass on these tests.
/pj-rehearse ack

maxcao13:
@jkyros @joelsmith Should we consider adding an "upgrade" test case? From our meeting it was mentioned about the possibility of an upgrade from the previous version to this new layout, so I'm wondering if we could cover some of that risk through a test case here. I've done a little digging and noticed our friends from WINC does something similar to what I think we should do: https://github.com/openshift/release/blob/2df80956b53d93dc28b4e4043be5056cafc68be7/ci-operator/config/openshift/windows-machine-config-operator/openshift-windows-machine-config-operator-master.yaml#L141
Basically, they promote the previous version of the bundle image and the operator image they are testing (`previous-bundle` and `previous-wmco`), import it as a base image, run the `previous-bundle` image with the `previous-wmco` replacement, and do `operator-sdk upgrade-bundle` and run the tests to see if anything breaks. 

It's also documented here: https://docs.ci.openshift.org/docs/how-tos/testing-operator-sdk-operators/#operator-upgrade-testing

ci-operator/config/openshift/vertical-pod-autoscaler-operator/openshift-vertical-pod-autoscaler-operator-master.yaml
```diff
@@ -1,4 +1,8 @@
 base_images:
+  golangci-lint:
+    name: golangci-lint
+    namespace: ci
+    tag: v1.59.1
```
jkyros, line 5: Do you know, will this auto-bump when they cut new channels, or are we going to have to police this? ( I just know if we skew too far from the golang version weird stuff happens)

ci-operator/config/openshift/vertical-pod-autoscaler-operator/openshift-vertical-pod-autoscaler-operator-master.yaml
```diff
@@ -7,26 +11,38 @@ base_images:
...
+  - name: TARGETARCH
+    value: amd64
+  dockerfile_path: Dockerfile
   inputs:
     ocp_4.17_base-rhel9:
       as:
       - registry.ci.openshift.org/ocp/4.17:base-rhel9
     openshift_release_rhel-9-release-golang-1.22-openshift-4.17:
       as:
       - registry.ci.openshift.org/openshift/release:rhel-9-release-golang-1.22-openshift-4.17
```
jkyros, line 34: We're kind of mixing and matching 4.17 and 4.18 here, I assume we will true up the 4.17/4.18 business once https://github.com/openshift/vertical-pod-autoscaler-operator/pull/169 merges?

ci-operator/config/openshift/vertical-pod-autoscaler-operator/openshift-vertical-pod-autoscaler-operator-master.yaml
```diff
@@ -96,21 +118,43 @@ tests:
...
+    test:
+    - as: test
+      cli: latest
+      commands: make test-scorecard
+      from: src
+      resources:
+        requests:
+          cpu: 100m
+          memory: 200Mi
+    workflow: hypershift-hostedcluster-workflow
```
jkyros, line 157: This is clever, running the scorecard tests against a hypershift node because it's less expensive. Was this cribbed off another operator or is this bespoke?

jkyros (commented):
Looks good, just a couple of questions for understanding.  I'm fine with it going in as-is, I don't think there's anything here that will cause problems.

ci-operator/config/openshift/vertical-pod-autoscaler-operator/openshift-vertical-pod-autoscaler-operator-master.yaml
maxcao13: I think this would have to be policied by us. It should probably match whatever is in our `Makefile`'s version. So when we do operator-sdk upgrades (by release chore cycles or however we want to do it), we should change it here as well so nothing unexpected happens.

jkyros:
> Should we consider adding an "upgrade" test case? From our meeting it was mentioned about the possibility of an upgrade breaking from the previous version to this new layout

I would love one. I've always wanted one.  Well, I've always wanted one for the CMA, but I'll take one for the VPA if you're offering one. We've just kind of been uh...getting by without one and having QE test it. :smile: 

If you think it will take too long and block everything else, we can come back for it and just make a point to have QE test the heck out of upgrades this cycle.

ci-operator/config/openshift/vertical-pod-autoscaler-operator/openshift-vertical-pod-autoscaler-operator-master.yaml
maxcao13: Yeah, 4.18 got cut in the middle of this PR, so I think we can have followup PRs on both repos to update the images to 4.18 on both after the whole migration is merged.

ci-operator/config/openshift/vertical-pod-autoscaler-operator/openshift-vertical-pod-autoscaler-operator-master.yaml
maxcao13: Noticed an example in the docs :D There is an example here: https://docs.ci.openshift.org/docs/architecture/ci-operator/#testing-with-a-cluster-from-hypershift

jkyros: heck yeah, love it, clearly I should go check the docs again periodically :smile:

maxcao13:
> > Should we consider adding an "upgrade" test case? From our meeting it was mentioned about the possibility of an upgrade breaking from the previous version to this new layout
> 
> I would love one. I've always wanted one. Well, I've always wanted one for the CMA, but I'll take one for the VPA if you're offering one. We've just kind of been uh...getting by without one and having QE test it. 😄
> 
> If you think it will take too long to make/test/get right here and block everything else, we can come back for it and just make a point to have QE test the heck out of upgrades this cycle.

Hm... I think we can get this change in and I can do a followup PR after everything else is merged in? Just so this isn't blocking the vpa PR.

jkyros:
Going to merge this, then going to run it against https://github.com/openshift/vertical-pod-autoscaler-operator/pull/169
/lgtm

PR #55754, opened 2024-08-20
OSASINFRA-3552: Updating default hwoffload compute flavor

MiguelCarpio:
hwoffload cloud was rebuilt without hwoffload configs, and now it doesn't support nfv flavors.


pierreprinetti:
Is this a temporary workaround ?

/approve
/lgtm

PR #55770, opened 2024-08-20
Fix for registry url auth string

deepsm007:
/cc @openshift/test-platform 

/hold wait for https://github.com/openshift/ci-tools/pull/4276 to merge

### DIFF
--- a/core-services/ci-secret-bootstrap/_config.yaml
+++ b/core-services/ci-secret-bootstrap/_config.yaml
@@ -616,7 +616,7 @@ secret_configs:
         registry_url: quay-proxy.ci.openshift.org
       - auth_field: auth
         item: quayio-ci-read-only-robot
-        registry_url: quay.io/openshift
+        registry_url: quay.io/openshift/
       - auth_field: auth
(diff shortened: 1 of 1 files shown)


core-services/ci-secret-bootstrap/_config.yaml
```diff
@@ -616,7 +616,7 @@ secret_configs:
         registry_url: quay-proxy.ci.openshift.org
       - auth_field: auth
         item: quayio-ci-read-only-robot
-        registry_url: quay.io/openshift
+        registry_url: quay.io/openshift/
```
droslean, line 619: This doesn't look like a fix. On the first look it seems that the bug exist in ci-secret-bootstrap tool, and this change is just a workaround.

deepsm007: The above change was a workaround for [this](https://redhat-internal.slack.com/archives/C07HL9Y8MKP/p1724164441103899) but later on fixed with a different change in https://github.com/openshift/release/pull/55833/files

PR #55774, opened 2024-08-20
[hcp][kubevirt] skip disconnected periodics

tiraboschi:
Temporary skip periodic execution of disconnected
test lanes for HCP with the Kubevirt provider
untile we are able to make them more stable.


davidvossel (commented):
/lgtm
/approve

PR #55781, opened 2024-08-20
Fix issue DPTP-4137 Use new token generation

bear-redhat:


bear-redhat:
/hold
ill run a local test first.

PR #55787, opened 2024-08-20
Temporarily disable LCA-4.14 image mirroring

lack:
This is causing a failure which stops all other images from being mirrored.  We will reinstate this once we figure out the underlying reason.


rauhersu (approved):

lack:
No longer needed; we fixed the underlying source image that was causing the problem, and the latest mirror job passed.

PR #55789, opened 2024-08-20
OSSM maistra add update charts into 2.6 jobs

yxun:


dgn (approved):
why not add it to the job that updates 2.6 charts?

ci-operator/config/maistra/istio-operator/maistra-istio-operator-maistra-2.4.yaml
```diff
@@ -135,6 +135,30 @@ tests:
   secrets:
   - mount_path: /creds-github
     name: ossm-github-simple-job
+- as: update-charts-in-2-6
```
jwendell, line 138: we must remove the previous job, i.e., we only update the latest operator (2.6) - no need to update 2.5

yxun: https://github.com/openshift/release/pull/55797

PR #55795, opened 2024-08-20
[INTEROP-7475] Set weekly trigger static cron for OCP 4.17 FIPS testing.

dfrazzette:
Temporary change for OCP 4.17 FIPS weekly trigger job to:  cron: 0 6 9 9 * to correct the current value and prevent the FIPS trigger from running on Wednesday mornings.


Simranpal (approved):
/lgtm

PR #55804, opened 2024-08-21
Remove CAPI configuration in 4.15 on AWS

yunjiang29:
CAPI has been GAed in 4.16, so removing test configuration in 4.15 now


yunjiang29:
@jianlinliu PTAL

jianlinliu:
cc @liangxia to review.

PR #55817, opened 2024-08-21
Reset host PDU if ipmi unable to connect

sgoveas:
When ipmi is unable to connect to the host the `prepare_host_for_boot`
[line 38](https://github.com/openshift-qe/baremetal-qe-infra/blob/master/images/fcos-bastion-image/root/usr/bin/prepare_host_for_boot#L38) loops till `timeout` is killed effectively killing the function itself,
so the pdu reset in the later part of the `reset_host()` never executes.

Above is fixed with PR [Avoid infinite loop for node power off](https://github.com/openshift-qe/baremetal-qe-infra/pull/79)
This patch will catch the issue and trigger a pdu reset before the other
steps.

### DIFF
--- a/ci-operator/step-registry/baremetal/lab/post/wipe/baremetal-lab-post-wipe-commands.sh
+++ b/ci-operator/step-registry/baremetal/lab/post/wipe/baremetal-lab-post-wipe-commands.sh
@@ -29,7 +29,7 @@ function wait_for_power_down() {
   local host_str
   host_str="#${bmc_forwarded_port##1[0-9]}"
   sleep 90
-  local retry_max=40 # 15*40=600 (10 min)
+  local retry_max=20 # 30*20=600 (10 min)
   while [ $retry_max -gt 0 ] && ! ipmitool -I lanplus -H "${AUX_HOST}" -p "${bmc_forwarded_port}" \
     -U "$bmc_user" -P "$bmc_pass" power status | grep -q "Power is off"; do
     echo "$host_str is not powered off yet... waiting"
@@ -54,6 +54,23 @@ function wait_for_power_down() {
   return 0
 }
 
+function reset_pdu() {
+  local pdu_uri="${1}"
+  pdu_host=${pdu_uri%%/*}
+  pdu_socket=${pdu_uri##*/}
+  pdu_creds=${pdu_host%%@*}
+  pdu_host=${pdu_host##*@}
+  pdu_user=${pdu_creds%%:*}
+  pdu_pass=${pdu_creds##*:}
+  # pub-priv key auth is not supported by the PDUs
+  echo "${pdu_pass}" > /tmp/ssh-pass
+
+  timeout -s 9 1m sshpass -f /tmp/ssh-pass ssh "${SSHOPTS[@]}" "${pdu_user}@${pdu_host}" <<EOF || true
+olReboot $pdu_socket
+quit
+EOF
+}
+
 function reset_host() {
   local bmc_address="${1}"
   local bmc_forwarded_port="${2}"
@@ -64,24 +81,31 @@ function reset_host() {
   local pdu_uri="${7:-}"
   local host="${bmc_forwarded_port##1[0-9]}"
   host="${host##0}"
+  if [ -n "${pdu_uri}" ] && ipmitool -I lanplus -H "${AUX_HOST}" -p "${bmc_forwarded_port}" \
+    -U "$bmc_user" -P "$bmc_pass" power status | grep -iq "Error: Unable to establish IPMI"; then
+    echo "Resetting ${host} pdu"
+    reset_pdu "${pdu_uri}"
+    local max_try=20
+    while [ "$max_try" -gt 0 ] && ! ipmitool -I lanplus -H "${AUX_HOST}" -p "${bmc_forwarded_port}" \
+      -U "$bmc_user" -P "$bmc_pass" power status | grep -iq "Chassis Power is"; do
+      echo "Waiting for PDU to become available"
+      sleep 30
+      max_try=$(( max_try - 1 ))
+    done
+    if [ $max_try -le 0 ]; then
+      echo "${host} PDU is unreachable, contact @metal-qe-team"
+      echo "$bmc_host:$bmc_forwarded_port" >> /tmp/failed
+      return 1
+    fi
+  fi
+
   timeout -s 9 10m ssh "${SSHOPTS[@]}" "root@${AUX_HOST}" prepare_host_for_boot "${host}" "pxe"
   if ! wait_for_power_down "$bmc_address" "$bmc_forwarded_port" "$bmc_user" "$bmc_pass" "$vendor" "$ipxe_via_vmedia"; then
     echo "$bmc_host:$bmc_forwarded_port" >> /tmp/failed
   fi
   [ -z "${pdu_uri}" ] && return 0
-  pdu_host=${pdu_uri%%/*}
-  pdu_socket=${pdu_uri##*/}
-  pdu_creds=${pdu_host%%@*}
-  pdu_host=${pdu_host##*@}
-  pdu_user=${pdu_creds%%:*}
-  pdu_pass=${pdu_creds##*:}
-  # pub-priv key auth is not supported by the PDUs
-  echo "${pdu_pass}" > /tmp/ssh-pass
 
-  timeout -s 9 10m sshpass -f /tmp/ssh-pass ssh "${SSHOPTS[@]}" "${pdu_user}@${pdu_host}" <<EOF || true
-olReboot $pdu_socket
-quit
-EOF
+  reset_pdu "${pdu_uri}"
   if ! wait_for_power_down "$bmc_address" "$bmc_forwarded_port" "$bmc_user" "$bmc_pass" "$vendor" "$ipxe_via_vmedia"; then
     echo "$bmc_address:$bmc_forwarded_port" >> /tmp/failed
   fi
@@ -121,7 +145,7 @@ for bmhost in $(yq e -o=j -I=0 '.[]' "${SHARED_DIR}/hosts.yaml"); do
   # shellcheck disable=SC2154
   echo "Ejecting virtual media from ${name}"
   # shellcheck disable=SC2154
-  timeout -s 9 10m ssh "${SSHOPTS[@]}" "root@${AUX_HOST}" umount.vmedia "${host}"
+  timeout -s 9 5m ssh "${SSHOPTS[@]}" "root@${AUX_HOST}" umount.vmedia "${host}"
 done
 
 if [ -s /tmp/failed ]; then


ci-operator/step-registry/baremetal/lab/post/wipe/baremetal-lab-post-wipe-commands.sh
```diff
@@ -64,24 +81,19 @@ function reset_host() {
...
   local host="${bmc_forwarded_port##1[0-9]}"
   host="${host##0}"
+  if ssh "${SSHOPTS[@]}" "root@${AUX_HOST}" ipmicmd "${host}" power status | grep -q "Error: Unable to establish IPMI v2" \
+  && [ -n "${pdu_uri}" ]; then
+    echo "Resetting ${host} pdu"
+    reset_pdu "${pdu_uri}"
+    sleep 30
+  fi
   timeout -s 9 10m ssh "${SSHOPTS[@]}" "root@${AUX_HOST}" prepare_host_for_boot "${host}" "pxe"
   if ! wait_for_power_down "$bmc_address" "$bmc_forwarded_port" "$bmc_user" "$bmc_pass" "$vendor" "$ipxe_via_vmedia"; then
```
aleskandro, line 91: Your implementation is like an initial check and reboot if we fail to retrieve the status. I would be ok with it if:
1. we can use `timeout` before the ssh command
2. no sleep 30 -> some hardware may need more time, other less. Can we 'wait until the status is reported successfully (with timeout)'?

An alternative is to bind the timeout that fails in the if statement:

```suggestion
  if timeout -s 9 10m ssh "${SSHOPTS[@]}" "root@${AUX_HOST}" prepare_host_for_boot "${host}" "pxe" && wait_for_power_down "$bmc_address" "$bmc_forwarded_port" "$bmc_user" "$bmc_pass" "$vendor" "$ipxe_via_vmedia"; then
    echo "Cleaned up successfully"
  else 
```

sgoveas: Regarding the suggestion: When the host is not reachable by IPMI, the issue is caused in `prepare_host_for_boot` on [line 38](https://github.com/openshift-qe/baremetal-qe-infra/blob/master/images/fcos-bastion-image/root/usr/bin/prepare_host_for_boot#L38) which will keep trying infinitely, not getting the expected string `Chassis Power is off`. In such case `timeout` is killed after 10m and when it is killed it even terminates the function so nothing else will execute in `reset_host()`. The `wait_for_power_down` function and everything else will not execute.

I have added timeout before the ssh and the `wait_for_power_down` for the check

aleskandro: Oh, got it! We'd fix it too though. We should loop up to a maximum number of retries

aleskandro: Also, wait_for_power_down may be not enough. If the server is set to restore the previous state when the power is back, it will not be off

sgoveas: I think the until loop is fine as the host should not be unreachable. If it is, then it will be fixed with the pdu reboot in the previous step. We can remove the `wait_for_power_down` and add a sleep of 90-120, as the next step is for the until loop that will check for power off state for at least 10min. WDYT?

aleskandro: I think it is prone to become an infinite loop... therefor having something like

For i in {1..X} max; do
 If I == max
   Break
 CheckPower
 Sleep Y
done

sgoveas: Replaced `wait_for_power_down` with `sleep 90` and added PR to [Avoid infinite loop for node power off](https://github.com/openshift-qe/baremetal-qe-infra/pull/79)

aleskandro: Can't the retry at line 49 suffice?

sgoveas: Seems like line 49 would not do much. It does not have `$pdu_uri` passed to it. Besides this, there are other issues.
After `prepare_host_for_reboot` exits with failure with `reset_host()` execution first time, the `wait_for_power_down` would run for 21.30 minutes then will execute `reset_host()` second time without the pdu_uri (doesn't matter at this point). Timeout `prepare_host_for_reboot` fails again as pdu is not rebooted yet. `wait_for_power_down` runs for another 21.30 mins. After 43 mins the pdu is finally rebooted without the disk clean  up.

aleskandro: Should we fix all of it then?

aleskandro: The initial issue for which we tried twice wasn't related to the BMC unreachability, but rather to kernel panic events happening right while executing the cleanup. The only way for recovering was to reboot the BMC via PDU (because that BMC doesn't support any reboot endpoints via Redfish).

We need to maintain this behavior and add a resolution for the issue you're facing with the other vendor. 


Ideally, re-executing at most twice, anytime a 'fatal error' occurs (kernel panic, BMc fully unreachable, etc,...)

sgoveas: I think this patch fixes the issues mentioned above. If a host is unreachable we would have to do a pdu reset for it to move forward. The node would be at least rebooted twice if there are any other issues.

aleskandro: That's ok if we want to introduce this further reboot for the pdu, but while you're fixing this and given the analysis

> After prepare_host_for_reboot exits with failure with reset_host() execution first time, the wait_for_power_down would run for 21.30 minutes, then will execute reset_host() second time without the pdu_uri (doesn't matter at this point). Timeout prepare_host_for_reboot fails again as pdu is not rebooted yet. wait_for_power_down runs for another 21.30 mins. After 43 mins + 15 miins (of prepare_host_for_reboot running 2 times) the pdu is finally rebooted without the disk clean up.

Can we better set up the timeouts?

ci-operator/step-registry/baremetal/lab/post/wipe/baremetal-lab-post-wipe-commands.sh
```diff
@@ -54,6 +54,23 @@ function wait_for_power_down() {
...
+  pdu_host=${pdu_uri%%/*}
+  pdu_socket=${pdu_uri##*/}
+  pdu_creds=${pdu_host%%@*}
+  pdu_host=${pdu_host##*@}
+  pdu_user=${pdu_creds%%:*}
+  pdu_pass=${pdu_creds##*:}
+  # pub-priv key auth is not supported by the PDUs
+  echo "${pdu_pass}" > /tmp/ssh-pass
+
+  timeout -s 9 10m sshpass -f /tmp/ssh-pass ssh "${SSHOPTS[@]}" "${pdu_user}@${pdu_host}" <<EOF || true
```
aleskandro, line 68:
```suggestion
  timeout -s 9 1m sshpass -f /tmp/ssh-pass ssh "${SSHOPTS[@]}" "${pdu_user}@${pdu_host}" <<EOF || true
```

should we enough for this task, right?

ci-operator/step-registry/baremetal/lab/post/wipe/baremetal-lab-post-wipe-commands.sh
```diff
@@ -64,24 +81,20 @@ function reset_host() {
   local pdu_uri="${7:-}"
   local host="${bmc_forwarded_port##1[0-9]}"
   host="${host##0}"
+  if timeout -s 9 5m ssh "${SSHOPTS[@]}" "root@${AUX_HOST}" ipmicmd "${host}" power status | grep -iq "Error: Unable to establish IPMI" \
```
aleskandro, line 84:
```suggestion
  if timeout -s 9 1m ssh "${SSHOPTS[@]}" "root@${AUX_HOST}" ipmicmd "${host}" power status | grep -iq "Error: Unable to establish IPMI" \
```

ci-operator/step-registry/baremetal/lab/post/wipe/baremetal-lab-post-wipe-commands.sh
```diff
@@ -64,24 +81,20 @@ function reset_host() {
...
+  && [ -n "${pdu_uri}" ]; then
+    echo "Resetting ${host} pdu"
+    reset_pdu "${pdu_uri}"
+    sleep 90
+  fi
+
   timeout -s 9 10m ssh "${SSHOPTS[@]}" "root@${AUX_HOST}" prepare_host_for_boot "${host}" "pxe"
```
aleskandro, line 91:
```suggestion
  timeout -s 9 5m ssh "${SSHOPTS[@]}" "root@${AUX_HOST}" prepare_host_for_boot "${host}" "pxe"
```

ci-operator/step-registry/baremetal/lab/post/wipe/baremetal-lab-post-wipe-commands.sh
```diff
@@ -29,7 +29,7 @@ function wait_for_power_down() {
   local host_str
   host_str="#${bmc_forwarded_port##1[0-9]}"
   sleep 90
-  local retry_max=40 # 15*40=600 (10 min)
+  local retry_max=40 # 30*40=1200 (20 min)
```
aleskandro, line 32:
```suggestion
  local retry_max=20 # 30*20=1200 (10 min)
```
Are 10 mins enough?

ci-operator/step-registry/baremetal/lab/post/wipe/baremetal-lab-post-wipe-commands.sh
sgoveas: Left this unchanged.  `timeout` setting of 5 mins will be shorter than the 15  (7 mins 30 sec) retries to power off the host in `prepare_host_for_reboot`. Shorter timeout would effectively kill the function itself as explained in my earlier comments.

ci-operator/step-registry/baremetal/lab/post/wipe/baremetal-lab-post-wipe-commands.sh
```diff
@@ -64,24 +81,20 @@ function reset_host() {
   local pdu_uri="${7:-}"
   local host="${bmc_forwarded_port##1[0-9]}"
   host="${host##0}"
+  if timeout -s 9 5m ssh "${SSHOPTS[@]}" "root@${AUX_HOST}" ipmicmd "${host}" power status | grep -iq "Error: Unable to establish IPMI" \
+  && [ -n "${pdu_uri}" ]; then
+    echo "Resetting ${host} pdu"
+    reset_pdu "${pdu_uri}"
+    sleep 90
```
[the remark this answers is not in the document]
aleskandro: Any news on this?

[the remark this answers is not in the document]
aleskandro: ok... do you think we can remove the `sleep 90`?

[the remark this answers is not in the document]
sgoveas: done

aleskandro (approved):
Thanks @sgoveas 
/lgtm
/approve

sgoveas:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.17-multi-nightly-baremetal-ipi-ovn-ipv4-vmedia-amd-f7

sgoveas:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.16-amd64-rollback-nightly-baremetal-ipi-ovn-ipv4-f7

PR #55826, opened 2024-08-21
Remove unmaintained okd-scos jobs config

elfosardo:
The okd-scos jobs are unmaintaiened and broken since 4.16 Removing them to save some CI resources, they need to be overridden anyway.


sdodson:
I would think we'd try to fix them in 4.16 and 4.17? Though not clear to me when OKD 4.16 goes EOL, I'd assume not for a few more months.

elfosardo:
> I would think we'd try to fix them in 4.16 and 4.17? Though not clear to me when OKD 4.16 goes EOL, I'd assume not for a few more months.

@sdodson probably missing context here, I left you a message on slack, let's continue the discussion there

sdodson:
/lgtm
/approve
We'll reintroduce these as release blocking jobs on latest supported OKD build once the dust has settled.

ausil:
Please revert this immediately

sdodson:
https://github.com/openshift/release/pull/55886

PR #55860, opened 2024-08-22
OCPBUGS-38973: Fix LVM install failure in versions <4.15

LiangquanLi930:
https://issues.redhat.com/browse/OCPBUGS-38973 
The PR was submitted to fix the issue because LVM versions prior to 4.15 do not have the `owned-by.topolvm.io/name=xxx` label, causing a failure in version 4.14.


Phaow:
Since the lvm-operator does not have label `owned-by.topolvm.io/name` for 4.14 and below -> https://github.com/openshift/lvm-operator/blob/release-4.14/controllers/topolvm_storageclass.go#L53 and the vg name in this step is hard value. Current fix looks good to me.
cc to @aleskandro

PR #55871, opened 2024-08-22
Add openshift ci cluster pools for 4.16 and 4.17

fedepaol:
The pools definitions were missing and here we are adding them.


fedepaol:
cc @deepsm007

PR #55873, opened 2024-08-22
ROX: do not skip postsubmits tests

janisz:
Currently we run some jobs depending on changed paths. This PR always runs all postsubmits tests to allow us to quickly detect if a commit introduced a breaking change.   

Replace:
```yaml
  postsubmit: true
  (run_if_changed|skip_if_only_changed): .*
```
with
```yaml
  postsubmit: true
```

/uncc


porridge (approved):
/lgtm

PR #55883, opened 2024-08-22
[SecureSign] add user/fghanmi to OWNERS

fghanmi:


JasonPowr (commented):
Can you also add @miyunari @SequeI and @bkabrda, thanks :)

JasonPowr:
@jupierce pinging you as I see you are in the root code owners file for the repo https://github.com/openshift/release/blob/master/OWNERS, I am wondering how we can add the users above to the org so we can add them to our Owners file?

PR #55890, opened 2024-08-22
Deploy jobset-controller as part of odh/kueue e2e tests

ChristianZaccaria:
Context for this change:

It seems as part of the jobset integration, there has been added e2e tests to verify behaviour of the jobset controller with Kueue. This change is to verify its integration and not diverge too much from upstream/Kueue. Moreover, eventually it will be supported natively by Kueue.

Some references:
https://github.com/kubernetes-sigs/kueue/issues/1463
https://github.com/kubernetes-sigs/kueue/pull/1902


ChristianZaccaria:
cc: @sutaakar

sutaakar:
@ChristianZaccaria Can you try to rebase the PR? It may fix the PR check issue.

PR #55906, opened 2024-08-25
OCPQE-20013: Fix non-deletion of vlan tagged interface

sgoveas:
Use last octet of master node 1 as the vlan id.

### DIFF
--- a/ci-operator/step-registry/baremetal/lab/post/provisioning-network/baremetal-lab-post-provisioning-network-commands.sh
+++ b/ci-operator/step-registry/baremetal/lab/post/provisioning-network/baremetal-lab-post-provisioning-network-commands.sh
@@ -10,10 +10,8 @@ fi
 [ -z "${AUX_HOST}" ] && { echo "\$AUX_HOST is not filled. Failing."; exit 1; }
 [ -z "${architecture}" ] && { echo "\$architecture is not filled. Failing."; exit 1; }
 
-# As the API_VIP is unique in the managed network and based on how it is reserved in the reservation steps,
-# we use the last part of it to define the VLAN ID.
-# TODO: find a similar unique value for dual stack and ipv6 single stack configurations?
-VLAN_ID=$(yq ".api_vip" "${SHARED_DIR}/vips.yaml")
+# Use the last octet of the first master node as the VLAN ID to keep it unique in the managed network
+VLAN_ID=$(yq '.[] | select(.name|test("master-00")).ip' "${SHARED_DIR}/hosts.yaml")
 VLAN_ID=${VLAN_ID//*\./}
 CLUSTER_NAME="$(<"${SHARED_DIR}/cluster_name")"
 SSH_KEY_PATH="${CLUSTER_PROFILE_DIR}/ssh-key"
--- a/ci-operator/step-registry/baremetal/lab/pre/provisioning-network/baremetal-lab-pre-provisioning-network-commands.sh
+++ b/ci-operator/step-registry/baremetal/lab/pre/provisioning-network/baremetal-lab-pre-provisioning-network-commands.sh
@@ -19,10 +19,8 @@ SSHOPTS=(-o 'ConnectTimeout=5'
   -o LogLevel=ERROR
   -i "${CLUSTER_PROFILE_DIR}/ssh-key")
 
-# As the API_VIP is unique in the managed network and based on how it is reserved in the reservation steps,
-# we use the last part of it to define the VLAN ID.
-# TODO: find a similar unique value for dual stack and ipv6 single stack configurations?
-VLAN_ID=$(yq ".api_vip" "${SHARED_DIR}/vips.yaml")
+# Use the last octet of the first master node as the VLAN ID to keep it unique in the managed network
+VLAN_ID=$(yq '.[] | select(.name|test("master-00")).ip' "${SHARED_DIR}/hosts.yaml")
 VLAN_ID=${VLAN_ID//*\./}
 CLUSTER_NAME="$(<"${SHARED_DIR}/cluster_name")"
 SSH_KEY_PATH="${CLUSTER_PROFILE_DIR}/ssh-key"


ci-operator/step-registry/baremetal/lab/ipi/install/baremetal-lab-ipi-install-commands.sh
```diff
@@ -75,6 +75,13 @@ if [ "${DISCONNECTED}" == "true" ] && [ -f "${SHARED_DIR}/install-config-mirror.
 fi
 file /tmp/openshift-baremetal-install
 
+
+if [ "${EXTERNAL_LB}" == "true" ] && [ -f "${SHARED_DIR}/external_vips.yaml" ]; then
+  VIPS="${SHARED_DIR}/external_vips.yaml"
+else
+  VIPS="${SHARED_DIR}/vips.yaml"
+fi
+
```
aleskandro, line 84: This is going to be a potential issue for future implementations using other LBs (as we already have). We are trying to avoid adding new variables when they are not really needed. 

Can we rather limit the changes for this bugfix to the provisioning network pre and post steps?

As an example, we can just use the last octet of the first master's IPv4 and we should actually solve a TODO we had in the pre step's code too wrt ipv6.

sgoveas: Done

ci-operator/step-registry/baremetal/lab/ipi/conf/external-lb/baremetal-lab-ipi-conf-external-lb-commands.sh
```diff
@@ -25,4 +25,7 @@ platform:
       type: UserManaged
 EOF
 
+# Backup vips.yaml file to move it back after the install-config.yaml is
+# processed in the install step
+cp "$SHARED_DIR/vips.yaml" "$SHARED_DIR/vips_backup_for_ipi_external_lb.yaml"
```
aleskandro, line 30: Can we rather limit the changes for this bugfix to the provisioning network pre and post steps?

sgoveas: This file is the pre step in which the vips.yaml file is overwritten by the external_vips.yaml file. I had added the reverting of the file in the ipi-install step to not lose it in case the cluster is preserved, prune_nodes would use the overwritten file and fail to delete the correct vlan interface. I have updated to save the backup on bastion now so we should have it in case the cluster is preserved. Vlan is already created with the last octet of the first master node. It is just overwritted with the external_vips.yaml for external_lb install.

aleskandro: Yes, what I'm saying is that if we can get that Id from the hosts.yaml instead of the API_VIP, we may decouple from the vips.yaml file that can change in several cases or be actually uneffective (that external lb one, the disconnected ipi case, ipv6, some hypershift scenarios).

How much would be the effort of changing this reference in the pre step of the provisioning network and in the post one? Do we also need to change it in the pruner script?

sgoveas: Done

ci-operator/step-registry/baremetal/lab/post/provisioning-network/baremetal-lab-post-provisioning-network-commands.sh
```diff
@@ -13,7 +13,7 @@ fi
 # As the API_VIP is unique in the managed network and based on how it is reserved in the reservation steps,
 # we use the last part of it to define the VLAN ID.
 # TODO: find a similar unique value for dual stack and ipv6 single stack configurations?
-VLAN_ID=$(yq ".api_vip" "${SHARED_DIR}/vips.yaml")
+VLAN_ID=$(yq ".[0].ip" "${SHARED_DIR}/hosts.yaml")
```
aleskandro, line 16:
```suggestion
VLAN_ID=$(yq '.[] | select(.name|test("master-00")).ip' "${SHARED_DIR}/hosts.yaml")
```

let's ensure it's the first master.

ci-operator/step-registry/baremetal/lab/post/provisioning-network/baremetal-lab-post-provisioning-network-commands.sh
```diff
@@ -13,7 +13,7 @@ fi
 # As the API_VIP is unique in the managed network and based on how it is reserved in the reservation steps,
 # we use the last part of it to define the VLAN ID.
 # TODO: find a similar unique value for dual stack and ipv6 single stack configurations?
```
aleskandro, line 15: Can you update this comment and remove the TODO regarding IPv6?

ci-operator/step-registry/baremetal/lab/pre/provisioning-network/baremetal-lab-pre-provisioning-network-commands.sh
```diff
@@ -22,7 +22,7 @@ SSHOPTS=(-o 'ConnectTimeout=5'
 # As the API_VIP is unique in the managed network and based on how it is reserved in the reservation steps,
```
aleskandro, line 22: Same about the comment.

aleskandro (approved):
Thanks @sgoveas 
/lgtm
/approve

sgoveas:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.18-multi-nightly-baremetal-ipi-ovn-ipv4-external-lb-arm-f7 periodic-ci-openshift-openshift-tests-private-release-4.15-arm64-stable-baremetal-pxe-ha-agent-ipv4-static-connected-f14

PR #55917, opened 2024-08-26
add env for sdn plugin

zhaozhanqi:
@JianLi-RH @jiajliu @shellyyang1989 ^^


shellyyang1989:
Hi @zhaozhanqi what issue does it fix?

zhaozhanqi:
> Hi @zhaozhanqi what issue does it fix?

for sdn cluster from 4.15 which is not supported for installation, so we need this parameter to hack the cluster to force setup openshift-sdn plugin.

zhaozhanqi:
@liangxia Could you help take a look and approve this?

PR #55994, opened 2024-08-27
Add metal jobs for cgroups v1 and crun in 4.18

harche:
This PR adds jobs for `crun` and `cgroups v1` for `metal`. All jobs are set to run `daily` for about a month, and then the frequency would be reduced to `weekly`.


harche:
I verified that all the jobs, irrespective of their successful or failed run, were correctly configured to run with `crun` or `cgroups v1` respectively. 

/pj-rehearse ack

vrutkovs (approved):
/approve
/lgtm

PR #56006, opened 2024-08-27
rename scanner-v4-tests to scanner-v4-install-tests

RTann:
Rename the Scanner V4 installation tests to reflect just that: they are installation tests. This is accompanied by https://github.com/stackrox/stackrox/pull/12530.


ci-operator/jobs/stackrox/stackrox/stackrox-stackrox-master-postsubmits.yaml
```diff
@@ -186,16 +186,16 @@ postsubmits:
       skip_cloning: true
     labels:
       ci.openshift.io/generator: prowgen
-    max_concurrency: 6
```
RTann, line 189: This was done automatically. Should I revert this?

ci-operator/config/stackrox/stackrox/stackrox-stackrox-master__ocp-stable-scanner-v4.yaml
```diff
@@ -121,4 +121,4 @@ zz_generated_metadata:
   branch: master
   org: stackrox
   repo: stackrox
-  variant: ocp-stable-scanner-v4
+  variant: ocp-stable-scanner-v4-install
```
RTann, line 124: this is not ok according to CI. Can someone help me understand what this is?

dcaravel (commented):
The changes LGTM, is there a way to test the re-names (if not already)?

RTann:
> The changes LGTM, is there a way to test the re-names (if not already)?

I think only some of them (the ones related to PRs) may be tested, and it's honestly a huge pain, as it requires coordination with https://github.com/stackrox/stackrox/pull/12530 and I think I'd have to create new files here with the branch's name

dcaravel (approved):
LGTM

RTann:
@BradLugo @mclasmeier now that the 4.6 branch has been cut, can we come back to this? The team aims to add some scanner v4 e2e tests so it may be best to make it clear the tests these reference, here, are installation-related

BradLugo (approved):
/lgtm

Do we want to update the release branches as well? My instinct is "no" because we'd need to backport changes from https://github.com/stackrox/stackrox/pull/12530, which doesn't really seem worth it, but wanted to make sure that's a conscious decision

RTann:
> /lgtm
> 
> Do we want to update the release branches as well? My instinct is "no" because we'd need to backport changes from [stackrox/stackrox#12530](https://github.com/stackrox/stackrox/pull/12530), which doesn't really seem worth it, but wanted to make sure that's a conscious decision

I don't think we need to. This is really meant to ensure future Scanner V4 tests will not be mistaken for the current installation tests. I think it's ok for previous releases to keep the old name, as the non-install Scanner V4 tests do not exist in those releases

mclasmeier (approved):

PR #56024, opened 2024-08-28
OSASINFRA-3577: Update ipi-conf-inframachineset monitor conditionals

MiguelCarpio:
The current monitor conditional waits 10 minutes to move infra pods from workers to all the new infra nodes. There is a case where the pods are already moved to infra nodes but not replicated in all of them; however, they are already moved and should pass the conditional. So, instead of validating the following,

 ` if [[ -z "$(echo "${MONITORING_NODES[@]}" "${INFRA_NODE_NAMES[@]}" | tr ' ' '\n' | sort | uniq -u)" ]]; then`

I'm proposing this:

 `if [[ ! "${MONITORING_NODES[*]}" =~ "worker" ]]; then`

Where I'm assuring, there are no infra pods for workers. 

Besides, the monitor for pods to settle doesn't need to consider installer pods.


MiguelCarpio:
/pj-rehearse periodic-ci-openshift-release-master-ci-4.16-upgrade-from-stable-4.15-e2e-aws-sdn-upgrade-infra

PR #56041, opened 2024-08-28
Add openshift/standalone-kube-proxy

danwinship:
New repo for standalone kube-proxy builds, which previously were happening out of openshift/sdn. (The repo is currently basically empty, but https://github.com/openshift/standalone-kube-proxy/pull/1 shows what _will_ be there... it turns out I can't actually merge until the plugins are configured, because the repo is owned by "Openshift Team SDN Maintainers" which as it turns out is empty.)

I saw that there's a `make new-repo` but I just copied stuff over from `openshift/sdn` instead, with appropriate renaming.

There are currently no tests defined for the repo other than the optional `security` test, because:
- There is no golang code in `standalone-kube-proxy` itself to validate/test.
- The vendored k8s code does not include unit tests, so there are no unit tests to run.
- Our CI currently has no (good) way to bring up a cluster with a standalone kube-proxy, so we can't e2e test.

(Hopefully the last part will eventually be fixed, but none of this is a regression from the old `openshift/sdn`-based builds; we were never testing the standalone kube-proxy image.)


danwinship:
we're doing this a different way

PR #56050, opened 2024-08-28
OCPQE-24986: fixing defer recovery

evakhoni:
https://issues.redhat.com/browse/OCPQE-24986 there's a corner case when skopeo fails just before the clear command is issued, causing the restored graph and channel to unblock multiarch migration and progress prematurely. the fix is applying the clear command prior to restoring the graph, which in case of a lack of failure from the test case, will have no adverse effect on the progress.


evakhoni:
a minor change, and the command guarded by `or true` so could we skip the rehearsal? wdty?

/cc @jianlinliu @JianLi-RH @dis016

PR #56074, opened 2024-08-29
[fix] the run-if condition for the rocm e2e tests job

jstourac:
Having the `||` means that even empty string is matched which results to this job being triggered everytime with any change to the repository.


jstourac:
@jiridanek @harshad16 @atheo89 please check, I don't have rights to add you as reviewers, thanks!

jiridanek:
or like this

/assign @jiridanek @harshad16 @atheo89

PR #56080, opened 2024-08-29
Add a waiting time before checking route53 status

yunjiang29:
To reduce `Rate exceeded` errors,  adding a waiting time before checking records' status.


yunjiang29:
@jianlinliu PTAL

PR #56115, opened 2024-08-30
OCPQE-23928 drop signaturestore cases

jiajliu:


jiajliu:
cc @jianlinliu PTAL, thank you.

PR #56030, opened 2024-08-28
OCPQE-25160: Decommission hypershift install jobs on Azure with OCP MC

fxierh:
All Azure test cases have been transitioned to AKS jobs. Decommission the `azure-ipi-ovn-hypershift-install` job.

PR #56179, opened 2024-09-02
OCPQE-25160: Decommission hypershift install jobs on Azure with OCP MC

fxierh:
This was already done for 4.16 and 4.17 in #56030 but wasn't carried over to 4.18. This PR applies the same changes to 4.18.


fxierh:
/cc @liangxia 

Could you please take a look ? (This PR basically does the same thing as #56030)

PR #56180, opened 2024-09-02
Add docker dind image for ODH multiarch builds

mdafsanhossain:
This PR adds docker in docker image for building ODH dashboard/operator for multiple archs (ppc64le, s390x).

### DIFF
--- a/core-services/image-mirroring/_config.yaml
+++ b/core-services/image-mirroring/_config.yaml
@@ -519,8 +519,8 @@ supplementalCIImages:
     image: quay.io/powercloud/powervs-actions-container:latest
   ocp-on-powervs/powervs-container-host:powervs-ocp-jenkins-deployment:
     image: quay.io/powercloud/powervs-ocp-jenkins-deployment:latest
-  odh-multi/docker-dind:latest:
-    image: quay.io/afsanhossain/docker-dind:latest
+  odh-multi/odh-dind:latest:
+    image: quay.io/powercloud/odh-dind:latest
   openshift/centos:stream9:
     image: quay.io/centos/centos:stream9
   openshift/centos:stream8:


core-services/image-mirroring/_config.yaml
```diff
@@ -519,6 +519,8 @@ supplementalCIImages:
     image: quay.io/powercloud/powervs-actions-container:latest
   ocp-on-powervs/powervs-container-host:powervs-ocp-jenkins-deployment:
     image: quay.io/powercloud/powervs-ocp-jenkins-deployment:latest
+  odh-multi/docker-dind:latest:
```
mkumatag, line 522: whats the source for this image? I'm wondering if we can push the container image to `quay.io/powercloud` with instructions how to build.

mkumatag (changes requested):

core-services/image-mirroring/_config.yaml
mdafsanhossain: It's the official docker image just mirrored to my personal quay.io account. It would be better if it can be pushed to powercloud.

mkumatag: in that case, why not directly mirror from the original location instead of mirroring to person repo? 🤔

mdafsanhossain: OpenShift CI cannot pull from `docker.io` due to pull rate limit. 
`We cannot mirror images from docker.io due to rate limiting constraints. Please, instead, push up an image to quay and mirror that to QCI.`

mkumatag: As discussed lets move of one of the org repo and use that instead.

mdafsanhossain:
@mkumatag I have updated the image. Can I get a `lgtm`?

mkumatag:
@deepsm007 ptal

danilo-gemoli (approved):

PR #56073, opened 2024-08-29
[4.13] merge efs to aws byo iam and s3 cloudfront to byo route53

jianlinliu:
To save aws cost, merge efs to aws byo iam and s3 cloudfront to byo route53 profile

PR #56208, opened 2024-09-03
missing efs base image

jianlinliu:
1. many upgrade ci jobs are missing efs base image setting (search them by `for i in `grep -r "as: .*-efs-" *|awk -F: '{print $1}'| sort -u`; do if grep -q "aws-efs-csi-operator-create-efs:" $i; then :; else echo $i; fi; done`), that would lead all the efs install workflow failed.
2. merge efs to aws byo iam and s3 cloudfront to byo route53 for 4.13 upgrade ci jobs that missed in https://github.com/openshift/release/pull/56073
3. fix some version setting issue in some upgrade jobs, the version should match with the source version


jianlinliu:
cc @Phaow @JianLi-RH to review

/pj-rehearse skip

jianlinliu:
cc @liangxia to approve

PR #56213, opened 2024-09-03
[CIGolang]add ibmcloud priv user

kuiwang02:
/cc @Xia-Zhao-rh @sandeepknd


kuiwang02:
@sandeepknd could you please add lgtm? thanks

PR #56223, opened 2024-09-03
OCPQE-25339: CI Integration x86 Control plane + ARM date plane (MGMT)

LiangquanLi930:


LiangquanLi930:

ci-operator/step-registry/hypershift/mce/aws/create/hypershift-mce-aws-create-chain.yaml
```diff
@@ -19,6 +19,9 @@ chain:
         - name: ENABLE_ICSP
           default: "false"
           documentation: "if true, add image content sources config(path=${SHARED_DIR}/mgmt_icsp.yaml)"
+        - name: HYPERSHIFT_INSTANCE_TYPE
+          default: "m5.xlarge"
+          documentation: "Instance type for the cluster nodes"
```
liangxia, line 24: Is it possible to use a small instance to save the cost ?

ci-operator/step-registry/hypershift/mce/aws/create/hypershift-mce-aws-create-chain.yaml
```diff
@@ -19,6 +19,9 @@ chain:
...
         - name: BREW_IMAGE_REGISTRY_USERNAME
           default: '|7820aea2-0d75-11e7-9259-28d244ea5a6d.hhav.f63e13'
```
liangxia, line 26: Not related to this PR. But I think this kind of info should be stored in vault. CC @YuLi517  @jianlinliu @jhou1

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.17__amd64-nightly.yaml
```diff
@@ -615,6 +615,40 @@ tests:
...
+  cron: 35 14 5,12,19,28 * *
+  steps:
+    cluster_profile: aws-qe
+    dependencies:
+      RELEASE_IMAGE_LATEST: release:multi-latest
+    env:
+      BASE_DOMAIN: qe.devcluster.openshift.com
+      EXTRA_ARGS: --arch=arm64
+      HYPERSHIFT_INSTANCE_TYPE: m6g.xlarge
+      MCE_VERSION: "2.7"
```
liangxia, line 628: I can see some 4.17 jobs are using MCE 2.6. Could you explain a little bit that how to decide we should use 2.6 or 2.7 ?

ci-operator/step-registry/hypershift/mce/aws/create/hypershift-mce-aws-create-chain.yaml
jianlinliu: agree.

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.17__amd64-nightly.yaml
LiangquanLi930: Based on the current rules, we should all be using version 2.7. Currently, e2e is using 2.7, while upgrade is using 2.6. I will submit a new PR to update all the upgrade CI to version 2.7.

ci-operator/step-registry/hypershift/mce/aws/create/hypershift-mce-aws-create-chain.yaml
LiangquanLi930: The default value for HYPERSHIFT_INSTANCE_TYPE is m5.xlarge. If you want to change the default value, you need to override HYPERSHIFT_INSTANCE_TYPE. As far as I know, we should not be able to use smaller instances, as this would result in many failed cases.

ci-operator/step-registry/hypershift/mce/aws/create/hypershift-mce-aws-create-chain.yaml
LiangquanLi930: Since these steps are maintained by multiple teams and I am not the owner of BREW_IMAGE_REGISTRY_USERNAME, I need to communicate with the other teams and then open a PR to take it.

liangxia: I'm fine to let this PR go.  However it's better to make sure there are jira issues to track the mentioned issues, especially this one.

LiangquanLi930: create https://issues.redhat.com/browse/HOSTEDCP-1949 to check

LiangquanLi930:
@liangxia Could you help to review again? Thanks

PR #56259, opened 2024-09-04
loki-operator: Add branch config and image mirroring for release-6.0

periklis:
/cc @xperimental

### DIFF
--- a/ci-operator/config/openshift/loki/openshift-loki-release-6.0.yaml
+++ b/ci-operator/config/openshift/loki/openshift-loki-release-6.0.yaml
@@ -1,9 +1,9 @@
 base_images:
   base:
-    name: "4.17"
+    name: "4.16"
     namespace: ocp
     tag: base-rhel9
-  loki-tag:
+  loki:
     name: v3.1.0
     namespace: logging
     tag: loki
@@ -11,10 +11,6 @@ base_images:
     name: mirror-5.y
     namespace: logging
     tag: lokistack-gateway
-  ocp_builder_rhel-9-golang-1.21-builder-multi-openshift-4.16:
-    name: builder
-    namespace: ocp
-    tag: rhel-9-golang-1.21-builder-multi-openshift-4.16
   opa-openshift:
     name: mirror-5.y
     namespace: logging
@@ -25,21 +21,6 @@ build_root:
     namespace: ocp
     tag: rhel-9-golang-1.21-builder-multi-openshift-4.16
 images:
-- dockerfile_path: Dockerfile.ocp
-  from: base
-  inputs:
-    ocp_builder_rhel-9-golang-1.21-builder-multi-openshift-4.16:
-      as:
-      - registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.21-openshift-4.16
-  multi_arch: true
-  to: loki
-- dockerfile_path: Dockerfile.promtail.ocp
-  from: base
-  inputs:
-    ocp_builder_rhel-9-golang-1.21-builder-multi-openshift-4.16:
-      as:
-      - registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.21-openshift-4.16
-  to: promtail
 - context_dir: operator
   dockerfile_path: Dockerfile
   from: base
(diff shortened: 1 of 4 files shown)


ci-operator/config/openshift/loki/openshift-loki-release-6.0.yaml
```diff
@@ -0,0 +1,161 @@
...
+  lokistack-gateway:
+    name: mirror-5.y
+    namespace: logging
+    tag: lokistack-gateway
+  ocp_builder_rhel-9-golang-1.21-builder-multi-openshift-4.16:
+    name: builder
+    namespace: ocp
+    tag: rhel-9-golang-1.21-builder-multi-openshift-4.16
+  opa-openshift:
+    name: mirror-5.y
```
JoaoBraveCoding, line 19: Shouldn't we also edit `core-services/image-mirroring/_config.yaml` to add mirror-6.y?

JoaoBraveCoding (commented):
just a small question but lgtm

ci-operator/config/openshift/loki/openshift-loki-release-6.0.yaml
```diff
@@ -0,0 +1,161 @@
...
+      - registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.21-openshift-4.16
+  multi_arch: true
+  to: loki
+- dockerfile_path: Dockerfile.promtail.ocp
+  from: base
+  inputs:
+    ocp_builder_rhel-9-golang-1.21-builder-multi-openshift-4.16:
+      as:
+      - registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.21-openshift-4.16
+  to: promtail
```
xperimental, line 42: Shouldn't these only be present for `main`?

ci-operator/config/openshift/loki/openshift-loki-release-6.0.yaml
```diff
@@ -0,0 +1,161 @@
+base_images:
+  base:
+    name: "4.16"
+    namespace: ocp
+    tag: base-rhel9
+  loki-tag:
```
xperimental, line 6: I think you don't need this rename when `loki` and `promtail` are removed from `images`.

ci-operator/config/openshift/loki/openshift-loki-release-6.0.yaml
```diff
@@ -0,0 +1,161 @@
...
+operator:
+  bundles:
+  - as: loki-operator-bundle
+    context_dir: operator/bundle/openshift
+    dockerfile_path: bundle.Dockerfile
+  substitutions:
+  - pullspec: quay.io/openshift-logging/loki-operator:0.1.0
+    with: pipeline:loki-operator
+  - pullspec: quay.io/openshift-logging/loki:v3.1.0
+    with: pipeline:loki-tag
```
xperimental, line 57: Similar to the beginning: I think this can be `loki` without the images in the build.

ci-operator/config/openshift/loki/openshift-loki-release-6.0.yaml
xperimental: Maybe `mirror-main` would be "more true"? Would then make sense in both versions.

periklis: Yes this is best served in separate cleanup PR once we get the branch config in.

JoaoBraveCoding (approved):
/lgtm

ci-operator/config/openshift/loki/openshift-loki-release-6.0.yaml
```diff
@@ -0,0 +1,146 @@
...
+    namespace: logging
+    tag: loki
+  lokistack-gateway:
+    name: mirror-5.y
+    namespace: logging
+    tag: lokistack-gateway
+  ocp_builder_rhel-9-golang-1.21-builder-multi-openshift-4.16:
+    name: builder
+    namespace: ocp
+    tag: rhel-9-golang-1.21-builder-multi-openshift-4.16
```
xperimental, line 17: I think this one is not needed anymore?

xperimental (commented):
Just found one image we don't need anymore. The rehearsals should still work though. It's possible that a few of them are not needed anymore, now that some of the builds were removed but are still present in the PR checks.

PR #56280, opened 2024-09-04
[cluster-bot] omit the `readOnly` key on the `ocmcred` mount

hoxhaeris:
That volume must be writable, as the Rosa client will persist the tokes there.


hoxhaeris:
/close
The k8s secrets are read-only, this will not work as expected.

hoxhaeris:
The k8s secrets are read-only, this will not work as expected.

PR #56292, opened 2024-09-04
Remove Pelorus periodic job and create stub trigger for the PR job

mpryc:
This PR is in fact disabling Pelorus e2e jobs without removing last CI job config which may be used/reworked in the future if needed.


etsauer (approved):
/lgtm

PR #56296, opened 2024-09-04
Update azure upgrade jobs to 5h30m timeout

sdodson:
Jobs are commonly waiting up to an hour for an Azure lease which is fine, boskos is doing its job, but then we start a job that takes longer than the remaining timeout just for it to fail. By extending the timeout and reducing the number of overall jobs that get re-run I believe we'll actually bring down lease utilization during peak usage.


sdodson:
/pj-rehearse periodic-ci-openshift-release-master-ci-4.14-e2e-azure-ovn-upgrade

wking (approved):
/lgtm

PR #56323, opened 2024-09-05
OCPQE-25359: Reduce test frequency for 4.16

liangxia:
Change f9/f14 -> f28,  f5/f7 -> f14, f2 -> f7, but excluding hypershift jobs.


jhou1:
Thank you Liang
/lgtm

PR #56343, opened 2024-09-05
move FILTERS_ADDITIONAL to eus chain

jianlinliu:
1. move FILTERS_ADDITIONAL to eus chain, so that we only need to set it in one place.
2. A minor update for OCP-21588, move it to another test profile, because `openshift-upgrade-qe-test-heterogeneous` does not include OTA test step in the chain.


jianlinliu:
cc @jiajliu to review.

jianlinliu:
cc @liangxia to review.

PR #56354, opened 2024-09-05
cincinnati: Install openssl-devel and gcc for cargo-test

DavidHurta:
/cc @petr-muller


petr-muller:
```
error: could not compile `prometheus-query` (lib) due to 6 previous errors
```

David is fixing this downstream in https://github.com/openshift/cincinnati/pull/934

/pj-rehearse ack

petr-muller (approved):

PR #56359, opened 2024-09-05
Set up tests for opendatahub-io Data Science Pipelines Operator repository

anishasthana:
cc @HumairAK


anishasthana:
For the sake of documenting -- in the future we will need to use the $KUBECONFIG variable to pass through the right credentials for `make integrationtest`

PR #56404, opened 2024-09-06
cincinnati: install cargo2junit into src-rust

petr-muller:
cargo2junit: https://github.com/johnterickson/cargo2junit

Rust / Cargo seems to have surprisingly crappy tooling in this domain. The cargo2junit seems to have its
[problems](https://github.com/johnterickson/cargo2junit/issues) and is not that actively maintained, but it seems to do the job for us.

https://github.com/openshift/cincinnati/pull/936

Goal is to run tests directly with `cargo test` and create artifacts for Prow
to consume and present in its JUnit lens, so that we do need to dig through
the raw output.

### DIFF
--- a/ci-operator/config/openshift/cincinnati/openshift-cincinnati-master.yaml
+++ b/ci-operator/config/openshift/cincinnati/openshift-cincinnati-master.yaml
@@ -25,6 +25,8 @@ images:
       rustup install 1.64.0
     # install yamllint
     RUN pip3 install --user yamllint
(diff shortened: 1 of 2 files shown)


petr-muller:
An interesting alternative would be to use the nextest runner: https://nexte.st/docs/machine-readable/junit/

petr-muller:
/close

Superseded by https://github.com/openshift/release/pull/56928 , we're going with Nextest

PR #56420, opened 2024-09-06
updating the scope of failed_jobs var

vi-patel:
Updating the scope of failed_jobs var so it doesn't get cleared out before final usage.

The following logs show the error, where a job fails to trigger (shows up correctly in the logs as: `periodic-ci-sclorg-ansible-tests-master-rhscl-ocp4.17-lp-interop-rhscl-interop-aws FAILED, a manual re-run is needed`; however, at the end of the log it says: `No jobs failed to be triggered.`

log file: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/periodic-ci-rhpit-interop-tests-main-weekly_trigger-ocp-self-managed-layered-product-interop-417/1825412586229731328/artifacts/ocp-self-managed-layered-product-interop-417/trigger-jobs/build-log.txt


Simranpal (approved):
/lgtm

PR #56440, opened 2024-09-09
[WIP] Check uesr access to ci-jenkins

shaior:
Check that the service account being used has access to a different Jenkins instance.


shaior:
Closing as the SA has access to launch job on the new instance.

PR #56441, opened 2024-09-09
Fix weekly trigger script: Convert `month_day` var from octal to integer.

dfrazzette:
Convert the `month_day` variable from octal to integer in order to calculate the compare date expressions.


dfrazzette:
/pj-rehearse periodic-ci-rhpit-interop-tests-main-weekly_trigger-ocp-self-managed-layered-product-interop-417

PR #56444, opened 2024-09-09
OCPQE-24827 Switch the profile of 4.9->4.17 upgrade path from azure-ipi to aws-ipi

JianLi-RH:
It seems reach TTL of azure, so switch to aws-ipi.

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.17__amd64-nightly-4.17-upgrade-from-stable-4.9.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.17__amd64-nightly-4.17-upgrade-from-stable-4.9.yaml
@@ -89,12 +89,12 @@ resources:
       cpu: 100m
       memory: 200Mi
 tests:
-- as: azure-ipi-f28
-  cron: 30 7 7 * *
+- as: aws-ipi-f28
+  cron: 34 5 5 * *
   steps:
-    cluster_profile: azure-qe
+    cluster_profile: aws-qe
     env:
-      BASE_DOMAIN: qe.azure.devcluster.openshift.com
+      BASE_DOMAIN: qe.devcluster.openshift.com
       LOKI_ENABLED: "false"
     test:
     - as: set-upgrade-releases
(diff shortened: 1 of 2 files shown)


JianLi-RH:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.17-amd64-nightly-4.17-upgrade-from-stable-4.9-aws-ipi-f28

JianLi-RH:
hit `exec /tools/entrypoint: argument list too long` issue
/cc @jianlinliu @jiajliu @shellyyang1989 can we merge this without a success run?

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.17__amd64-nightly-4.17-upgrade-from-stable-4.9.yaml
```diff
@@ -122,8 +127,10 @@ tests:
         requests:
           cpu: 100m
           memory: 200Mi
-    - chain: openshift-upgrade-qe-test-chainupgrade
```
jianlinliu, line 125: Why we need this change? I saw we are reference `openshift-upgrade-qe-test-chainupgrade` even in 4.8 -> 4.17 chain upgrade.

JianLi-RH: I refereed https://github.com/openshift/release/blob/master/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.17__amd64-nightly-4.17-upgrade-from-stable-4.5.yaml#L112 which is the latest aws-ipi job.

jianlinliu: The source version is 4.5, that is too old, I guess that is because olm does not support yet on 4.5, so we have to split the install workflow.

For the source version on 4.9, I think we should use install workflow and upgrade chain as usual.

JianLi-RH: ok, let us try the new workflow

JianLi-RH: done

PR #56465, opened 2024-09-09
mirror-images: make it re-usable outside of QE env

EmilienM:
These steps were needed so we could re-use this script outside of the QE
environment in our CI.

Context: jira/OSASINFRA-3583

* Load "${SHARED_DIR}/proxy-conf.sh" if it exists, it'll allow to reach
  mirrors that are themselves behind a proxy.
* Move `oc registry login` earlier so it runs before getting the
  metadata version.

### DIFF
--- a/ci-operator/step-registry/mirror-images/by-oc-adm/mirror-images-by-oc-adm-commands.sh
+++ b/ci-operator/step-registry/mirror-images/by-oc-adm/mirror-images-by-oc-adm-commands.sh
@@ -39,6 +39,11 @@ fi
 echo "MIRROR_REGISTRY_HOST: $MIRROR_REGISTRY_HOST"
 echo "OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE: ${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE}"
 
+# since ci-operator gives steps KUBECONFIG pointing to cluster under test under some circumstances,
+# unset KUBECONFIG to ensure this step always interact with the build farm.
+unset KUBECONFIG
+oc registry login
+
 readable_version=$(oc adm release info "${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE}" -o jsonpath='{.metadata.version}')
 echo "readable_version: $readable_version"
 
@@ -51,11 +56,6 @@ target_release_image="${target_release_image_repo}:${readable_version}"
 
 echo "target_release_image: $target_release_image"
 echo "target_release_image_repo: $target_release_image_repo"
-
-# since ci-operator gives steps KUBECONFIG pointing to cluster under test under some circumstances,
-# unset KUBECONFIG to ensure this step always interact with the build farm.
-unset KUBECONFIG
-oc registry login
 
 # combine custom registry credential and default pull secret
 registry_cred=$(head -n 1 "/var/run/vault/mirror-registry/registry_creds" | base64 -w 0)
@@ -94,6 +94,17 @@ if oc adm release mirror -h | grep -q -- --print-mirror-instructions; then
     args+=(--print-mirror-instructions="${mirror_crd_type}")
 else
     echo "This version of oc does not support --print-mirror-instructions=, skip it."
+fi
+
+# For disconnected or otherwise unreachable mirrors, we want to
+# have steps use an HTTP(S) proxy to reach the registry. This proxy
+# configuration file should export HTTP_PROXY, HTTPS_PROXY, and NO_PROXY
+# environment variables, as well as their lowercase equivalents (note
+# that libcurl doesn't recognize the uppercase variables).
+if test -f "${SHARED_DIR}/proxy-conf.sh"
+then
+        # shellcheck disable=SC1090
+        source "${SHARED_DIR}/proxy-conf.sh"
 fi
 
 # execute the mirror command


EmilienM:
I don't know which job to rehearse, please help.

EmilienM:
ping @yunjiang29 @jianlinliu @jinyunma for review, thanks

ci-operator/step-registry/mirror-images/by-oc-adm/mirror-images-by-oc-adm-commands.sh
```diff
@@ -39,6 +50,12 @@ fi
 echo "MIRROR_REGISTRY_HOST: $MIRROR_REGISTRY_HOST"
 echo "OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE: ${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE}"
 
+# since ci-operator gives steps KUBECONFIG pointing to cluster under test under some circumstances,
+# unset KUBECONFIG to ensure this step always interact with the build farm.
+unset KUBECONFIG
+registry_url=$(echo "${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE}" | cut -d'/' -f1)
```
jianlinliu, line 56: I wonder why we need this change.

Compare the log in your [job](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_release/56373/rehearse-56373-periodic-ci-shiftstack-ci-release-4.18-e2e-openstack-singlestackv6/1833300143412088832/artifacts/e2e-openstack-singlestackv6/mirror-images-by-oc-adm/build-log.txt) with this change and another regular ci [job](https://gcsweb-qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/qe-private-deck/logs/periodic-ci-openshift-openshift-tests-private-release-4.17-arm64-nightly-aws-ipi-disc-priv-tp-f14/1833212701204025344/artifacts/aws-ipi-disc-priv-tp-f14/mirror-images-by-oc-adm/build-log.txt) without this change.

Your job log with this change:
```
MIRROR_REGISTRY_HOST: mirror-registry.lizmfnfl-e1579.shiftstack.devcluster.openshift.com:5000
OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE: registry.build03.ci.openshift.org/ci-op-lizmfnfl/release@sha256:56f25591b1473f6c2ac7a92bc72e6d6e69b684848db531a8aadf73b588ed7302
Saved credentials for registry.build03.ci.openshift.org into /alabama/run/containers/auth.json
```

OCP QE's job with this change:
```
MIRROR_REGISTRY_HOST: ec2-54-236-255-73.compute-1.amazonaws.com:5000
OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE: registry.build03.ci.openshift.org/ci-op-gy6hzm72/release@sha256:ce715d16b17099e218e786ce2873b2cb0f74dd6bf91925e2f6ac2a7c5fd7c933
readable_version: 4.17.0-0.nightly-arm64-2024-09-09-150502
target_release_image: ec2-54-236-255-73.compute-1.amazonaws.com:5000/ci-op-gy6hzm72/release:4.17.0-0.nightly-arm64-2024-09-09-150502
target_release_image_repo: ec2-54-236-255-73.compute-1.amazonaws.com:5000/ci-op-gy6hzm72/release
info: Using registry public hostname registry.build03.ci.openshift.org
Saved credentials for registry.build03.ci.openshift.org into /alabama/run/containers/auth.json
```

Even without this change, `oc registry login` can locate the correct registry url, so I guess we do not need this change.

ci-operator/step-registry/mirror-images/by-oc-adm/mirror-images-by-oc-adm-commands.sh
```diff
@@ -4,6 +4,17 @@ set -o nounset
 set -o errexit
 set -o pipefail
 
+# For disconnected or otherwise unreachable environments, we want to
+# have steps use an HTTP(S) proxy to reach the API server. This proxy
+# configuration file should export HTTP_PROXY, HTTPS_PROXY, and NO_PROXY
+# environment variables, as well as their lowercase equivalents (note
+# that libcurl doesn't recognize the uppercase variables).
+if test -f "${SHARED_DIR}/proxy-conf.sh"
```
jianlinliu, line 12: if the mirror registry is behind a proxy, I think we should move the export proxy block to a place as possible as close to the oc mirror command, e.g: https://github.com/openshift/release/pull/56465/files#diff-bdea30af18dd50cc47a8f900a78e8b74d7a1f09860173647789b129acf41dc8dR118

Once export proxy here, all the traffic will go through it, even interact with build farm cluster.

ci-operator/step-registry/mirror-images/by-oc-adm/mirror-images-by-oc-adm-commands.sh
EmilienM: @jianlinliu

1. I have a `403 Permission Denied` on the second link.
2. Here is a [log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_release/56373/rehearse-56373-periodic-ci-shiftstack-ci-release-4.18-e2e-openstack-singlestackv6/1833242885663756288/artifacts/e2e-openstack-singlestackv6/mirror-images-by-oc-adm/build-log.txt) when we don't give the registry hostname. Maybe it doesn't happen in your environment because of some magic? But as you can see in my logs it's needed. And also it doesn't hurt to have it, in fact it's suggested that we provide it to avoid confusion. Please accept that change if this doesn't break you.

ci-operator/step-registry/mirror-images/by-oc-adm/mirror-images-by-oc-adm-commands.sh
EmilienM: I'm fine with this request, I'll do it right now.

EmilienM:
/pj-rehearse ack
I believe there is no rehearsal since the jobs using this step are private.

ci-operator/step-registry/mirror-images/by-oc-adm/mirror-images-by-oc-adm-commands.sh
jianlinliu: > Maybe it doesn't happen in your environment because of some magic

Hmm, actually I can not remember what special magic we did. I did some search, found a similar issue we fixed before, you can reference it https://github.com/openshift/release/pull/42848

> And also it doesn't hurt to have it, in fact it's suggested that we provide it to avoid confusion. 

The step is being referenced widely by many install workflows, and per my experience, there are several special ways to trigger the install workflow, not only regular periodic ci , e.g: call prow api gateway by overriding a specific payload; call cluster-robot to trigger the install workflows. I have some concern this change might bring some side effect on those special trigger way, though I can not confirm if this change will break them or not, but currently it works well, and meet all different requirements, so I prefer not to change it to avoid some unexpected issues happened. And from https://github.com/openshift/release/pull/42848, I think we can fix your original issue  by moving proxy export block somewhere.

EmilienM: ok good, I'll remove it now so you can approve if the rest is good for you. Thanks!
PS: I'll let you know the results on my testing in the other PR.

EmilienM: it worked

PR #56496, opened 2024-09-10
fix: Remove ARM builds for 1.3 and 1.0 OADP Branches

mateusoliveira43:
Remove ARM builds for 1.3 and 1.0 OADP Branches, as their Dockerfile do not support multi-arch, yet.


rayfordj (approved):
/lgtm
/approve

PR #56519, opened 2024-09-10
ROX-25438: Upgrade stackrox-ui-test to prepare for Go 1.22.5

BradLugo:
## Description

I missed updating the Go version in the `stackrox-ui-test` CI image. These changes include the new image that has the corrected Go version.

/cc @davdhacs @tommartensen

### DIFF
--- a/ci-operator/config/stackrox/scanner/stackrox-scanner-release-2.34.yaml
+++ b/ci-operator/config/stackrox/scanner/stackrox-scanner-release-2.34.yaml
@@ -7,7 +7,7 @@ build_root:
   image_stream_tag:
     name: apollo-ci
     namespace: stackrox
-    tag: scanner-test-0.3.69
+    tag: scanner-test-0.4.4
 resources:
   '*':
     requests:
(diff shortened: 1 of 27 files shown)


ci-operator/config/stackrox/scanner/stackrox-scanner-release-2.34.yaml
```diff
@@ -7,7 +7,7 @@ build_root:
   image_stream_tag:
     name: apollo-ci
     namespace: stackrox
-    tag: scanner-test-0.3.69
+    tag: scanner-test-0.4.4
```
RTann, line 10: Do we need to update the Scanner in 4.5 to use Go 1.22?

ci-operator/config/stackrox/stackrox/stackrox-stackrox-release-3.74.yaml
```diff
@@ -23,7 +23,7 @@ build_root:
   image_stream_tag:
     name: apollo-ci
     namespace: stackrox
-    tag: stackrox-ui-test-0.4.3
+    tag: stackrox-ui-test-0.4.4
```
RTann, line 26: did we even need to update the Go version for 3.74 in the first place? I believe we do not need to fix any vulnerabilities for 3.74 at this point, so I wonder if it's best to keep this on whatever version of Go it was originally on?

PR #56520, opened 2024-09-10
add periodic logging job

jcantrill:
Add periodic job to ensure claims remain viable

cc @cahartma @xperimental


jcantrill:
/pj-rehearse periodic-ci-openshift-cluster-logging-operator-master-functional-periodic

xperimental:
I think the issue with the stale clusters should be a non-issue once development restarts again. It might also make sense to reduce the number of "ready clusters" in the pool to the number needed by a single PR (maybe +1), so that they get cycled faster.

PR #56524, opened 2024-09-10
OKO - Increase kuttl test step's timeout

viroel:
mariadb-operator is getting more kuttl tests which is taking longer than 2h to finish. This PR increases kuttl step's timeout to 4h and also increase mariadb-operator kuttl job to 6h.


viroel:
/pj-rehearse ack

no need to rerun jobs, we are just increasing timeouts to add more tests

PR #56491, opened 2024-09-10
[4.17] correct release image config

jianlinliu:
1. openshift-openshift-tests-private-release-4.17__multi-stable.yaml is using nightly build to run testing, e.g: this [job](https://gcsweb-qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/qe-private-deck/logs/periodic-ci-openshift-openshift-tests-private-release-4.16-multi-stable-azure-ipi-disc-fullypriv-arm-mixarch-f28/1831974069222772736/artifacts/azure-ipi-disc-fullypriv-arm-mixarch-f28/cucushift-installer-check-info/build-log.txt), fix it.
2. Correct release:target image to a stable build in openshift-openshift-tests-private-release-4.17__amd64-stable.yaml and openshift-openshift-tests-private-release-4.17__arm64-stable.yaml
3. In openshift-openshift-tests-private-release-4.17__multi-nightly.yaml, set release:latest to multi arch image by default to save multiple overriding lines.

PR #56537, opened 2024-09-11
[4.18] correct release image config

jianlinliu:
Similar to https://github.com/openshift/release/pull/56491, in openshift-openshift-tests-private-release-4.18__multi-nightly.yaml, set release:latest to multi arch image by default to save multiple overriding lines.


jianlinliu:
@liangxia can you help review and approve?

PR #56540, opened 2024-09-11
opeshift-loki: Fix multi-arch builds in main

periklis:


JoaoBraveCoding (commented):
So IIUC this change we are removing the builds from main since we are building loki and promtail already in the upstream branches and we are removing multi_arch build from main because we are only interested in them in release branches. Is this correct?

periklis:
> So IIUC this change we are removing the builds from main since we are building loki and promtail already in the upstream branches and we are removing multi_arch build from main because we are only interested in them in release branches. Is this correct?

I am removing this from main and soon also from the `release-5.*` and `release-6.0` branches because it creates unnecessary noice and failures for. The CI team (nowadays called TRT previously DPTP) is and always was interested in multi-arch builds of Promtail. We satisfy this need with enabling `multi_arch: true` in the `upstream-v*` branches only.

JoaoBraveCoding (approved):
/lgtm

PR #56572, opened 2024-09-11
opeshift-loki: Disable multi-arch builds in release branches

periklis:
Follow-up of #56540

/cc @JoaoBraveCoding


JoaoBraveCoding (approved):
/lgtm

PR #56585, opened 2024-09-11
removing rhoe 4.12 cluster pool

acornett21:
- OCP 4.12 has no been marked EOL in pyxis, so we can remove our cluster pool.


skattoju (approved):

komish (approved):
/lgtm

PR #56617, opened 2024-09-12
opendatahub/vllm: use build10 cluster for smoke test

dtrifiro:
build05 cluster has an issue with GPU drivers which has caused the smoke test pod to fail to schedule for the past month.

See original discussion: https://redhat-internal.slack.com/archives/CBN38N3MW/p1726075499300699?thread_ts=1724079927.308519&cid=CBN38N3MW


dtrifiro:
Not needed anymore, `build05` is back up and running

PR #56654, opened 2024-09-12
Add GPU operator and nodes to build10

bear-redhat:


smg247:
@bear-redhat, is this just copy/pasted from `build05`? If so, can we define a common folder? If not, why not?

PR #56728, opened 2024-09-14
[CIGolang]fix KUBEADMIN_REMOVED

kuiwang02:
/cc @Xia-Zhao-rh @rioliu-rh 

when I support kubeadmin, I made PR https://github.com/openshift/release/pull/56674 which checks KUBEADMIN_REMOVED in all go steps.

actually KUBEADMIN_REMOVED is only in opneshift-extended-test step.

so, remove it from other golang steps.


kuiwang02:
move KUBEADMIN_REMOVED logic to remove_kubeadmin_user function.

PR #56775, opened 2024-09-16
NO-JIRA: fix: remove sno informing micro upgrade

eggfoobar:
removing upgrade micro sno lane, this throws off the aggregated run and is not really needed anymore


sosiouxme (approved):
/lgtm

PR #56777, opened 2024-09-16
Rename builds

vaibhavjainwiz:


vaibhavjainwiz:
/pj-rehearse pull-ci-opendatahub-io-vllm-main-cuda-pr-image-mirror

ci-operator/config/opendatahub-io/vllm/opendatahub-io-vllm-main.yaml
```diff
@@ -84,15 +84,15 @@ tests:
       SOURCE_IMAGE_REF: vllm-build-main-cuda
     env:
       IMAGE_REPO: vllm
-      RELEASE_VERSION: cuda-pr
+      RELEASE_VERSION: cuda
```
dtrifiro, line 87: This should be `fast`

vaibhavjainwiz: `cuda` gives more clarity to consumers. If we just tag it `fast` then we need to explictly mention it in doc which people usually don't read. With `cuda` tag it would be very clear by its name only that this vLLM image version is to support NVIDIA CCDA.

vaibhavjainwiz:
/pj-rehearse pull-ci-opendatahub-io-vllm-main-cuda-pr-image-mirror

vaibhavjainwiz:
/pj-rehearse pull-ci-opendatahub-io-vllm-main-cuda-pr-image-mirror

ci-operator/config/opendatahub-io/vllm/opendatahub-io-vllm-main.yaml
Xaenalt: Hmmm, is that true about the fast tag? If so, then it should be fine, but given the landscape these days it makes sense to have the CUDA build be the default, since that's what most customers will consume. Can we add both fast and cuda tags to the image?

PR #56816, opened 2024-09-17
Reduce the frequency of NVIDIA GPU operator jobs on 4.17

empovit:
Reduce the frequency to weekly until the version is stable and has the NFD operator in the catalog


wabouhamad (commented):
/lgtm

PR #56848, opened 2024-09-18
only run tests when spec file changed

yapei:
trigger cluster installation and tests only when test file changed, refer doc https://docs.prow.k8s.io/docs/jobs/#triggering-jobs


yapei:
Hi @kuiwang02 Could you help take a look whether the changes are correct? Thanks very much!

memodi (approved):
/lgtm
thanks @yapei

PR #56871, opened 2024-09-18
CNF-14557: Add pao updating profile on hypershift lane

rbaturov:
This lane is needed to run suites testing the performance profile update functionality. Currently, we will enable this lane to run only manually by user request (always_run=false).

### DIFF
--- a/ci-operator/config/openshift/cluster-node-tuning-operator/openshift-cluster-node-tuning-operator-master.yaml
+++ b/ci-operator/config/openshift/cluster-node-tuning-operator/openshift-cluster-node-tuning-operator-master.yaml
@@ -86,6 +86,25 @@ tests:
           cpu: 100m
           memory: 200Mi
     workflow: hypershift-aws-e2e-cluster
+- always_run: false
+  as: e2e-pao-updating-profile-hypershift
+  steps:
+    cluster_profile: hypershift
+    test:
+    - as: e2e-pao-updating-profile-hypershift
+      cli: latest
+      commands: |-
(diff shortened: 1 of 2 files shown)


ci-operator/config/openshift/cluster-node-tuning-operator/openshift-cluster-node-tuning-operator-master.yaml
```diff
@@ -86,6 +86,25 @@ tests:
           cpu: 100m
           memory: 200Mi
     workflow: hypershift-aws-e2e-cluster
+- as: e2e-pao-updating-profile-hypershift
+  always_run: false
```
Tal-or, line 90: Hey @jmencak I recall you prefer to have it this way and not always running.

Just wanted to make sure you're pleased with that

jmencak: Hey Talor, cannot remember the conversation, but yes, unless we need the tests blocking merges, this makes sense.  Do we also want `optional: true` here?

Tal-or: @jmencak not necessarily. 
If this lane intended to run only if mentioned explicitly, then the least I would expect is for it to be required and not optional.

Tal-or:
/hold 
let's wait for https://github.com/openshift/cluster-node-tuning-operator/pull/1168 to get merged first.
Then we should rerun this lane and see that it passes.

Tal-or:
/lgtm
/hold cancel
Awesome! Thanks!

PR #56874, opened 2024-09-18
Update aro-e2e test to use new profile and workflow

slintes:
- add test, which uses new workflow and cluster profile
- add `releases`
  - will be used to inject correct `oc` binary into the test container
  - can be used for getting release image and ocp version in workflow (not implemented yet) 
- allow access to RH network (https://docs.ci.openshift.org/docs/architecture/ci-operator/#allowing-tests-to-access-red-hat-intranet)


hlipsig (approved):
lgtm

PR #56890, opened 2024-09-18
openstack: add ssv6 jobs to MCO

EmilienM:
Add OpenStack single stack v6 jobs to the MCO.


yuqi-zhang (approved):
Workflow seems fine, all optional and not always_run, lgtm

PR #56901, opened 2024-09-19
update target to use candidate channel

jiajliu:
Although the ota dedicated job is calling `cucushift-upgrade-toversion` and target version in ci config is not the actual upgrade target, to be consistent with other upgrade jobs, update target in the config to use candidate channel instead of nightly build in stable-to-stable upgrade jobs.


jiajliu:
/pj-rehearse periodic-ci-openshift-verification-tests-master-ota-multi-stable-4.17-upgrade-from-stable-4.16-azure-ipi-amd-mixarch-f28

PR #56933, opened 2024-09-19
machine-os-images: convert SDN jobs to OVN

elfosardo:


elfosardo:
/pj-rehearse ack
don't really need to run the tests here

PR #56935, opened 2024-09-19
openshift-e2e-test: adjustments for disconnected install

EmilienM:
* Load proxy env before any `oc` command, or it'll fail in disconnected
  environment.
* If `${SHARED_DIR}/mirror-tests-image` exists, add it to the TEST_ARGS
  as `--from-repository`. It's necessary when the tests image has been
  mirrored.
* Run image import with `--insecure=true`. The mirror might use a
  self-signed certificate so we need this argument to avoid TLS errors.


EmilienM:
/pj-rehearse periodic-ci-openshift-release-master-ci-4.18-e2e-aws-ovn

PR #56937, opened 2024-09-19
Make mirror-images-qe-test-images re-usable outside QE env

EmilienM:
* Skip the step if ${SHARED_DIR}/mirror_registry_url doesn't exist.
  It allows to run that step in shared workflows and the step is skipped
  when a mirror is not used by the job.
* Allow to provide a pre-created list of images to mirror. This can be
  triggered by `QE_IMAGES` set to false (default is true for backward
  compatibility). This allows us to generate the list in a previous step
  and give the file here so the mirror can be populated.
  We are not interested by QE images but by the script to push the
  images into the mirror.
* Load proxy environment for disconnected clusters and mirrors.

### DIFF
--- a/ci-operator/step-registry/mirror-images/qe-test-images/mirror-images-qe-test-images-commands.sh
+++ b/ci-operator/step-registry/mirror-images/qe-test-images/mirror-images-qe-test-images-commands.sh
@@ -8,21 +8,22 @@ trap 'CHILDREN=$(jobs -p); if test -n "${CHILDREN}"; then kill ${CHILDREN} && wa
 
 new_pull_secret="${SHARED_DIR}/new_pull_secret"
 
+if [ ! -f "${SHARED_DIR}/mirror_registry_url" ]; then
+    echo "File ${SHARED_DIR}/mirror_registry_url does not exist, skipping..."
+    exit 0
+fi
 # private mirror registry host
 # <public_dns>:<port>
 MIRROR_REGISTRY_HOST=`head -n 1 "${SHARED_DIR}/mirror_registry_url"`
-if [ ! -f "${SHARED_DIR}/mirror_registry_url" ]; then
-    echo "File ${SHARED_DIR}/mirror_registry_url does not exist."
-    exit 1
-fi
 echo "MIRROR_REGISTRY_HOST: $MIRROR_REGISTRY_HOST"
 
 # since ci-operator gives steps KUBECONFIG pointing to cluster under test under some circumstances,
 # unset KUBECONFIG to ensure this step always interact with the build farm.
 unset KUBECONFIG
 
-# Create qe test image list
-cat <<EOF > "/tmp/mirror-images-list.yaml"
+if [ "${QE_IMAGES}" == "true" ]; then
+    echo "Create qe test image list"
+    cat <<EOF > "/tmp/mirror-images-list.yaml"
 quay.io/openshifttest/iperf3:latest=MIRROR_REGISTRY_PLACEHOLDER/openshifttest/iperf3:latest
 quay.io/openshifttest/hello-sdn:1.2.0=MIRROR_REGISTRY_PLACEHOLDER/openshifttest/hello-sdn:1.2.0
 quay.io/openshifttest/hello-openshift-fedora:latest=MIRROR_REGISTRY_PLACEHOLDER/openshifttest/hello-openshift-fedora:latest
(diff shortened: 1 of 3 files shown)


ci-operator/step-registry/mirror-images/qe-test-images/mirror-images-qe-test-images-commands.sh
```diff
@@ -107,16 +120,23 @@ openshifttest_auth_password=$(cat "/var/run/vault/mirror-registry/registry_quay_
...
-# MIRROR IMAGES
-# Creating ICSP for quay.io/openshifttest is in enable-qe-catalogsource-disconnected step
-# Set Node CA for Mirror Registry is in enable-qe-catalogsource-disconnected step
-sed -i "s/MIRROR_REGISTRY_PLACEHOLDER/${MIRROR_REGISTRY_HOST}/g" "/tmp/mirror-images-list.yaml" 
+# For disconnected or otherwise unreachable environments, we want to
+# have steps use an HTTP(S) proxy to reach the API server. This proxy
+# configuration file should export HTTP_PROXY, HTTPS_PROXY, and NO_PROXY
+# environment variables, as well as their lowercase equivalents (note
+# that libcurl doesn't recognize the uppercase variables).
+if test -f "${SHARED_DIR}/proxy-conf.sh"
```
jianlinliu, line 128: Once enabled this block, all the mirror traffic would go though proxy to mirror registry, even mirror registry can be directly connected (especially in some QE's scenario on public clouds). Per my original understanding, "${SHARED_DIR}/proxy-conf.sh" is mainly for reaching cluster api endpoints which is not exposed to public.

`mirror-images-qe-test-images` step is bit different with `mirror-images-by-oc-adm`, `mirror-images-qe-test-images` step generally is executed in post-installation, while `mirror-images-by-oc-adm` is executed in pre-installation. 

In  pre-installation phase, when `mirror-images-by-oc-adm` is being executed, `${SHARED_DIR}/proxy-conf.sh` is not created yet in most of cases, so it seldom brings side effect for QE's ci jobs, while in post-installation phase, `${SHARED_DIR}/proxy-conf.sh` is created, so all the traffic will go through proxy to mirror registry. 

To avoid side effect as much as possible, I am thinking the following solutions for your reference:
1. Introduce a separate env to decide source `"${SHARED_DIR}/proxy-conf.sh"` or not, we can set it on job config level as our demand.
2. Create a new proxy config file with a different name, e.g: "${SHARED_DIR}/mirror-proxy-conf.sh" to indicated the file is only used to reach mirror registry behind a proxy.

EmilienM: I like idea number 2. If that's ok for you. I'll push it and let you decide. I'm happy with 1 or 2 and would avoid 3 if possible. New steps are expensive to maintain sometimes and this is probably the last step we need from your team, we got a successful CI run with it.

jianlinliu: If you like option 2, I am okay with it, too. BTW, do you mind to update `mirror-images-by-oc-adm` to use `"${SHARED_DIR}/mirror-proxy-conf.sh"` instead of `"${SHARED_DIR}/proxy-conf.sh"`

EmilienM: done.

ci-operator/step-registry/mirror-images/qe-test-images/mirror-images-qe-test-images-commands.sh
```diff
@@ -107,16 +120,23 @@ openshifttest_auth_password=$(cat "/var/run/vault/mirror-registry/registry_quay_
...
-do
-    oc image mirror $image  --insecure=true -a "${new_pull_secret}" \
- --skip-missing=true --skip-verification=true --keep-manifest-list=true --filter-by-os='.*'
-done
+# MIRROR IMAGES
+ # To avoid 409 too many request error, mirroring image one by one
+ for image in `cat /tmp/mirror-images-list.yaml`
+ do
+     oc image mirror $image  --insecure=true -a "${new_pull_secret}" \
+         --skip-missing=true --skip-verification=true --keep-manifest-list=true --filter-by-os='.*' --insecure=true
```
jianlinliu, line 139: There are two duplicated option - `--insecure=true` ?

EmilienM: oops

jianlinliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.15-amd64-stable-aws-c2s-ipi-disc-priv-fips-f28

PR #56984, opened 2024-09-20
disable appstudio-hac-e2e job

jrichter1:


ci-operator/config/redhat-appstudio/infra-deployments/redhat-appstudio-infra-deployments-main.yaml
```diff
@@ -17,7 +17,8 @@ resources:
       cpu: 100m
       memory: 200Mi
 tests:
-- as: appstudio-hac-e2e-tests
+- always_run: false
+  as: appstudio-hac-e2e-tests
```
psturc, line 21: I think this won't work - see https://github.com/openshift/release/pull/55234/files as a reference for making the job optional (i.e. it will run only if requested via `/test <job-name>` command)
it needs to have the field `optional: true` and `skip_if_only_changed` field removed

psturc (approved):
/lgtm
/pj-rehearse ack

PR #56998, opened 2024-09-20
OPRUN-3542: Remove catalogd upstream e2e

tmshort:
The upstream e2e fails downstream; there is not enough setup.

This harmonizes the e2e between catalogd and operator-controller.


tmshort:
Closing in favor of #57229

PR #57004, opened 2024-09-21
Set cluster to build05

jadhaj:
Cluster was mistakenly set to build03


jadhaj:
@hoxhaeris any idea why ordered-prow-config job is failing ?

PR #57023, opened 2024-09-23
OpenStack: Fix sriov step to add config only on 4.16+

mandre:
This is a followup to 425d6e5dc1d134181e412fb6ca53b0cacc9e45c9 that fixed the 4.16 job but unfortunately broke the jobs for other versions.

On versions lower than 4.16, openshift complains that the config already exists. Instead of switching the command to apply, we're now conditionally creating this for 4.16+ versions only, so that we don't overwrite config with different settings by mistake.


mandre:
/pj-rehearse pull-ci-openshift-machine-api-provider-openstack-release-4.15-e2e-openstack-nfv-intel pull-ci-openshift-machine-api-provider-openstack-release-4.16-e2e-openstack-nfv-intel

pierreprinetti:
Thanks!
/lgtm
Now let’s see what the CI thinks of it.

PR #57029, opened 2024-09-23
chore: upgrade ocp to 4.18 in common templates tests

ksimon1:
chore: upgrade ocp to 4.18 in common templates tests


ksimon1:
replaced by https://github.com/openshift/release/pull/57802

PR #57051, opened 2024-09-23
Fix some steps that are not multi-arch-enabled and use multi-arch images before scaling up arm64 nodes

aleskandro:
Some images used by the steps are already multi-arch, either because they are promoted as multi-arch images or because they come from a multi-arch image built elsewhere (e.g., the ART multi-arch base images). Since pods using images built also for arm64 may land on arm64 nodes whenever the buildfarm cluster runs arm64 nodes, these jobs may fail.

I found these references of jobs using the multi-arch ci-operator:latest imagestreamtag while supporting an investigation about a failure of the ci-chat-bot  pods when they run on arm64.
The new versions seem not downloading the CLI or other tools in place.

I didn't go through the whole logic of the test cases, but this PR ensures the CLI tools are  downloaded according to the architecture of the nodes where they run.

/cc @deepsm007

### DIFF
--- a/ci-operator/jobs/openshift/release/openshift-release-release-4.10-periodics.yaml
+++ b/ci-operator/jobs/openshift/release/openshift-release-release-4.10-periodics.yaml
@@ -18,7 +18,7 @@ periodics:
         # prow doesn't allow init containers or a second container
         export PATH=$PATH:/tmp/bin
         mkdir /tmp/bin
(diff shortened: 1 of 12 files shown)


aleskandro:
Failures are unrelated. Other jobs can't be rehearsed.

aleskandro:
The failures are unrelated. Some of the rehearsals are based on releases that no longer exist, and others are not rehears-able.

I will check the results in the deck.

PR #57024, opened 2024-09-23
[CIGolang]use oc4pre for pre-upgrade oc inject

kuiwang02:
/cc @shellyyang1989 @Xia-Zhao-rh 

for upgrade, there are two kinds of upgrade: 
A: single step. for example, N->N+1
B: multiple steps. for example, N-> N+1 -> N+2 ...>N+final
for these jobs, latest is used for N version installation, and target is used for final version (N+1 or N+final) installation.

originally, pre-upgrade step is executed on N version cluster whether it is single-step or multiple-steps upgrade, and we re-use latest to inject oc.
                 post upgrade is executed final version cluster, and we use target to inject oc.

Now, there is change for pre-upgrade step: it is executed on final-1 version cluster (per https://github.com/openshift/release/pull/57024#discussion_r1772451718, some still execute on N version cluster.). 
so, we need to inject oc according this version, or else some operators are not supported by old oc.

so, we introduce dedicated tag oc4pre for pre-upgrade to inject oc.
so,
for single-step upgrade, oc4pre tag should same with latest.
for multi-steps upgrade, oc4pre tag should be same with final-1 tag definition (per https://github.com/openshift/release/pull/57024#discussion_r1772451718, it is still N for openshift-upgrade-qe-test-cpou). note: if it is not amd64 or multi, rememer to change it to amd64 or multi for oc4pre tag.
as summary, you need to determine which load's oc is used for your pre step and the above is general suggestion.

you could follow the example in this PR to define oc4pre tag.

PR #57071, opened 2024-09-24
[CIGolang]new pre step to take non latest tag for oc inject

kuiwang02:
/cc @Xia-Zhao-rh @jianlinliu @jiajliu @shellyyang1989 

per https://github.com/openshift/release/pull/57024#issuecomment-2370123260, make this PR.

### DIFF
--- a/ci-operator/step-registry/openshift-extended/upgrade/pre-custom-cli/OWNERS
+++ b/ci-operator/step-registry/openshift-extended/upgrade/pre-custom-cli/OWNERS
@@ -1,0 +1,7 @@
+approvers:
+  - jianzhangbjz
+  - kuiwang02
+reviewers:
+  - jianzhangbjz
+  - kuiwang02
+  - xia-zhao-rh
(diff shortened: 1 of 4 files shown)


ci-operator/step-registry/openshift-extended/upgrade/prenonNoc/openshift-extended-upgrade-prenonNoc-ref.yaml
```diff
@@ -0,0 +1,65 @@
+ref:
+  as: openshift-extended-upgrade-prenonNoc
```
jianlinliu, line 2: The content of this step is the same as `openshift-extended-upgrade-pre`, the only difference is `cli: oc4pre`, right?

BTW, I wonder what means `nonNoc` ?

kuiwang02: yes.
nonNoc means non N oc
I can not make good step name, could you please provide one? thanks

jianlinliu: How about naming it `openshift-extended-upgrade-pre-custom-cli`, once that, maybe we also need to change `cli: oc4pre` to `cli: custom`

kuiwang02: ok, done

PR #57098, opened 2024-09-24
MGMT-19003: Add CI job to test 4, 5 control plane node cluster installation

danmanor:
Adding jobs to test OCP installation with 4/5 control planes

### DIFF
--- a/ci-operator/config/openshift/assisted-service/openshift-assisted-service-master__edge.yaml
+++ b/ci-operator/config/openshift/assisted-service/openshift-assisted-service-master__edge.yaml
@@ -1390,6 +1390,50 @@ tests:
       ASSISTED_CONFIG: |
         PLATFORM=external
     workflow: assisted-baremetal
+- always_run: false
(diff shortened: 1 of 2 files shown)


ci-operator/config/openshift/assisted-service/openshift-assisted-service-master__edge.yaml
```diff
@@ -1390,6 +1390,50 @@ tests:
       ASSISTED_CONFIG: |
         PLATFORM=external
     workflow: assisted-baremetal
+- as: e2e-metal-assisted-5-control-planes-4-17
```
adriengentil, line 1393:
```suggestion
- always_run: false
  as: e2e-metal-assisted-5-control-planes-4-17
```

I don't think we want to run these jobs each time a PR is opened?

danmanor: Yes thank you !

adriengentil (approved):

PR #57112, opened 2024-09-24
remove config for SBO

sadlerap:
SBO has been [deprecated and archived](https://github.com/redhat-developer/service-binding-operator?tab=readme-ov-file#deprecation-notice), so this configuration is no longer needed.


dperaza4dustbit (approved):
/lgtm

PR #57141, opened 2024-09-25
use openshift 4.18 in rook CI

Nikhil-Ladha:
update rook master branch ci config to use openshift 4.18


Nikhil-Ladha:
@iamniting can you please review?

iamniting (approved):

Nikhil-Ladha:
@iamniting seems like we are missing rehearsals-ack label here, can you please add that?

PR #57155, opened 2024-09-25
openstack: new chain for disconnected e2e

EmilienM:
We need a dedicated chain so images can be mirrored before running e2e.


mandre (commented):
/lgtm

PR #57173, opened 2024-09-25
[openshift-logging] Fix CI jobs pointing to old eventrouter image and remove obsolete jobs

xperimental:
The image mirroring is currently failing because it's still looking for the `5.y` image which is not built anymore, this PR aims to fix that by removing the mirroring for that image and also changing the configuration of CI jobs still depending on it.

I also tried to identify CI jobs that we do not need anymore and deleted them. There are probably more jobs to remove, please have a look at the directories I touched, if I can remove more files.

cc @jcantrill @periklis


ci-operator/config/ViaQ/logging-fluentd/ViaQ-logging-fluentd-release-5.6.yaml
```diff
@@ -32,7 +32,7 @@ base_images:
     namespace: logging
     tag: logging-elasticsearch6
   logging-eventrouter:
-    name: 5.y
+    name: 6.y
```
jcantrill, line 35: Normally I would suggest making this a specific 5.x image but given the ER doesn't really change and 5.6 is going away in a few months this is fine.

ci-operator/config/ViaQ/logging-fluentd/ViaQ-logging-fluentd-ruby27-v1.14.x.yaml
```diff
@@ -52,7 +52,7 @@ base_images:
     namespace: logging
     tag: logging-elasticsearch6
   logging-eventrouter:
-    name: 5.y
+    name: 6.y
```
jcantrill, line 55: same comment here.  I guess 5.y was pruned because there are no jobs referencing it any longer :thinking:

ci-operator/config/ViaQ/logging-fluentd/ViaQ-logging-fluentd-v1.7.4.yaml
```diff
@@ -52,7 +52,7 @@ base_images:
     namespace: logging
     tag: logging-elasticsearch6
   logging-eventrouter:
-    name: 5.y
+    name: 6.y
```
jcantrill, line 55: We should consider reviewing the fluent jobs to see what versions we actually need.  I think there are several of these which can be removed

ci-operator/config/ViaQ/logging-fluentd/ViaQ-logging-fluentd-ruby27-v1.14.x.yaml
jcantrill: ref: https://github.com/ViaQ/logging-fluentd

periklis (commented):
/lgtm

xperimental:
Rebased and also removed additional CI branches in the vector and fluentd repositories.

ci-operator/config/ViaQ/logging-fluentd/ViaQ-logging-fluentd-v1.7.4.yaml
xperimental: I've only kept the two branches listed in the version matrix for our current versions.

ci-operator/config/ViaQ/logging-fluentd/ViaQ-logging-fluentd-ruby27-v1.14.x.yaml
xperimental: For me, it looks as if images get pruned if nothing is _producing_ them anymore. I guess this is an effort to keep the number of images small and also to avoid CI running with stale images that are not being updated anymore.

periklis (commented):
/lgtm

ci-operator/config/ViaQ/vector/ViaQ-vector-release-6.0.yaml
```diff
@@ -36,7 +36,7 @@ base_images:
     namespace: logging
     tag: logging-elasticsearch6
   logging-eventrouter:
-    name: 5.y
+    name: 6.y
```
jcantrill, line 39: Can't think off-hand why these test fails now, maybe they haven't been run:

`make test-functional`
`make test-e2e`

PR #57182, opened 2024-09-25
Updated deprecated git-sync options

hector-vido:
Updated the other files after the test done with PRs https://github.com/openshift/release/pull/57176/ and https://github.com/openshift/release/pull/57181.

Removed `GIT_SYNC_DEST` variable and replaced it with the parameter `--link`.

Resolves: [DPTP-4177](https://issues.redhat.com/browse/DPTP-4177)

The following script was used:

```bash
while read FILE; do

 grep -- --branch $FILE > /dev/null 2>&1
  if [ "$?" -eq 0 ]; then
    echo $FILE --branch
    sed -i 's,--branch=,--ref=,g' $FILE
  fi

  grep -- --wait $FILE > /dev/null 2>&1
  if [ "$?" -eq 0 ]; then
    echo $FILE --wait
    sed -Ei 's,--wait=([0-9]+),--period=\1s,g' $FILE
  fi

  grep -- --max-sync-failures $FILE > /dev/null 2>&1
  if [ "$?" -eq 0 ]; then
    echo $FILE --max-sync-failures
    sed -i 's,--max-sync-failures=,--max-failures=,g' $FILE
  fi

  grep GIT_SYNC_DEST $FILE > /dev/null 2>&1
  if [ "$?" -eq 0 ]; then
    echo $FILE GIT_SYNC_DEST
    sed -Ezi 's,env:\s*- name: GIT_SYNC_DEST\s*value: release,- --link=release,g' $FILE
  fi

done <<EOF
`grep -r git-sync 2>/dev/null | cut -d: -f1 | uniq | grep .yaml`
EOF
```

> **Important:** Some manifests are generated by a python script and a single one had a different indentation.

### DIFF
--- a/clusters/app.ci/assets/admin_ci-images-mirror.yaml
+++ b/clusters/app.ci/assets/admin_ci-images-mirror.yaml
@@ -70,13 +70,11 @@ spec:
         - /git-sync
         args:
         - --repo=https://github.com/openshift/release.git
-        - --branch=master
+        - --ref=master
         - --root=/tmp/git-sync
         - --one-time=true
         - --depth=1
-        env:
-        - name: GIT_SYNC_DEST
-          value: release
+        - --link=release
         image: quay-proxy.ci.openshift.org/openshift/ci:ci_git-sync_v4.2.3
         volumeMounts:
         - name: release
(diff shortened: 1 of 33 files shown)


Prucek:
I still see this: 
```{"logger":"","ts":"2024-09-25 20:25:08.282266","caller":{"file":"main.go","line":393},"level":0,"msg":"setting --link from deprecated --dest"}```
https://console-openshift-console.apps.ci.l2s4.p1.openshiftapps.com/k8s/ns/ci/pods/payload-testing-prow-plugin-6bbdfcc496-2t8ld/logs

hector-vido:
@Prucek I found no files using `--link` or `--dest`, this looks like a default message from the binary to warn people about these options.

PR #57190, opened 2024-09-25
stackrox: retry stackrox cr creates

davdhacs:
/cc @mtodor @BradLugo @tommartensen 
I've seen the custom-resource creation fail about 1/20 times in my testing of perf&scale runs (like with https://github.com/openshift/release/pull/57120). It passes on retry.


tommartensen (approved):
/lgtm

PR #57193, opened 2024-09-25
stackrox: switch acs tests to ocp-4.17 after GA oct1st

davdhacs:
After GA of OCP 4.17, we will move the ACS tests on stackrox master to OCP 4.17 from 4.16.


BradLugo (approved):
/lgtm

PR #57211, opened 2024-09-26
Add trigger plugin for eco-ci-cd repo

shaior:


core-services/prow/02_config/openshift-kni/eco-ci-cd/_pluginconfig.yaml
```diff
@@ -0,0 +1,44 @@
+approve:
+- commandHelpLink: ""
+  repos:
+  - openshift-kni/eco-ci-cd
+  require_self_approval: false
+external_plugins:
```
natifridman, line 6: I think most of this list is inherited from the defaults. Maybe we can start without them and add it later if needed?

natifridman (approved):
/lgtm

PR #57221, opened 2024-09-26
OpenStack: Restore creation of default config in provision-sriov-networknodepolicy

mandre:
This partially reverts 425d6e5dc1d134181e412fb6ca53b0cacc9e45c9 and fixes the openstack pre-submit jobs that don't run the `openstack-provision-sriov-operator` step.

### DIFF
--- a/ci-operator/step-registry/openstack/provision/sriov-networknodepolicy/openstack-provision-sriov-networknodepolicy-commands.sh
+++ b/ci-operator/step-registry/openstack/provision/sriov-networknodepolicy/openstack-provision-sriov-networknodepolicy-commands.sh
@@ -118,6 +118,20 @@ wait_for_injector_webhook() {
   fi
 }
 
+create_default_sriov_operator_config() {
+    oc apply -f - <<EOF
+apiVersion: sriovnetwork.openshift.io/v1
+kind: SriovOperatorConfig
+metadata:
+  name: default
+  namespace: openshift-sriov-network-operator
+spec:
+  enableInjector: true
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/openstack/provision/sriov-networknodepolicy/openstack-provision-sriov-networknodepolicy-commands.sh
```diff
@@ -172,6 +186,11 @@ fi
 
 wait_for_sriov_pods
 
+# This is only needed on ocp 4.16+
+# introduced https://github.com/openshift/sriov-network-operator/pull/887
+# u/s https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/617
+create_default_sriov_operator_config
```
EmilienM, line 192: should not we check for the OCP version then?

mandre: we could, but that's not necessary since this is `oc apply` which will overwrite the config if it already exists. This would be different if we were using `oc create`, where it would fail if the object already exists (below 4.16, or when we already ran the `openstack-provision-sriov-operator` step).

EmilienM:
/lgtm

I'll let you handle the rehearse if you want any

mandre:
/pj-rehearse pull-ci-openshift-sriov-network-operator-master-e2e-openstack-nfv

PR #57255, opened 2024-09-27
Fix reporting bug in prow ci

vrindle:
Since we have concurrent builds enabled in Jenkins, when multiple builds are queued, all the blocked values are set to false (as they can run but are waiting for an executor). Instead of checking the blocked field to determine if a job is available, we should examine whether the job is set to "waiting for executor" to assess its availability.

### DIFF
--- a/ci-operator/step-registry/dpu-operator/e2e-tests-workflow/tests/dpu-operator-e2e-tests-workflow-tests-commands.sh
+++ b/ci-operator/step-registry/dpu-operator/e2e-tests-workflow/tests/dpu-operator-e2e-tests-workflow-tests-commands.sh
@@ -16,10 +16,15 @@ wait_for_job_to_run() {
 	do
 		JSON_FILE=$(mktemp /tmp/pullnumber.XXX)
 		curl -k --resolve "$endpoint_resolve" -X GET $1/api/json > $JSON_FILE
-		blocked=$(cat $JSON_FILE | jq .blocked)
-		if [[ "$blocked" == "true" ]]; then
+		blocked=$(cat $JSON_FILE | jq .why)
+
+                if [[ "$blocked" != "null" && "$blocked" =~ ^\"(.*)\"$ ]]; then
+    			blocked="${BASH_REMATCH[1]}"
+                fi
+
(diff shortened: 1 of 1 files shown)


vrindle:
@wizhaoredhat ptal

ci-operator/step-registry/dpu-operator/e2e-tests-workflow/tests/dpu-operator-e2e-tests-workflow-tests-commands.sh
```diff
@@ -16,10 +16,10 @@ wait_for_job_to_run() {
 	do
 		JSON_FILE=$(mktemp /tmp/pullnumber.XXX)
 		curl -k --resolve "$endpoint_resolve" -X GET $1/api/json > $JSON_FILE
-		blocked=$(cat $JSON_FILE | jq .blocked)
-		if [[ "$blocked" == "true" ]]; then
+		blocked=$(cat $JSON_FILE | jq .why)
+		if [[ "$blocked" == "\"Waiting for next available executor on ‘ipu’\"" ]]; then
```
wizhaoredhat, line 20: Is it possible to do a generic regex instead of hardcoding "ipu"? I feel like if we change this, we will forget to change it here.

ci-operator/step-registry/dpu-operator/e2e-tests-workflow/tests/dpu-operator-e2e-tests-workflow-tests-commands.sh
```diff
@@ -16,10 +16,15 @@ wait_for_job_to_run() {
 	do
 		JSON_FILE=$(mktemp /tmp/pullnumber.XXX)
 		curl -k --resolve "$endpoint_resolve" -X GET $1/api/json > $JSON_FILE
-		blocked=$(cat $JSON_FILE | jq .blocked)
-		if [[ "$blocked" == "true" ]]; then
+		blocked=$(cat $JSON_FILE | jq .why)
+
+                if [[ "$blocked" != "null" ]]; then
+                        blocked="${blocked:1:-1}"
```
wizhaoredhat, line 22: Why do we have to offset this string? Looks a bit fragile...

PR #57289, opened 2024-09-30
NO-ISSUE: Add release and presubmit jobs for CRI-O versions 1.30 and 1.31

kwilczynski:
The CRI-O versions 1.30 and 1.31 have been released some time ago, but the jobs were never added.

Thus, add the missing release and presubmit jobs for both CRI-O releases to fill the test coverage gap.

The release job is also particularly useful, as it would generate an artefact that can be used to bring up an OpenShift cluster with it, which is extremely useful when testing and troubleshooting issues.

### DIFF
--- a/ci-operator/config/cri-o/cri-o/cri-o-cri-o-release-1.30.yaml
+++ b/ci-operator/config/cri-o/cri-o/cri-o-cri-o-release-1.30.yaml
@@ -1,0 +1,68 @@
+base_images:
+  base-machine:
+    name: centos
+    namespace: openshift
+    tag: stream9
+  rhel-coreos-9-base:
+    name: "4.17"
+    namespace: ocp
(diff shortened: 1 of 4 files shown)


kwilczynski:
@openshift/cri-o-maintainers, please have a look. Thank you!

kwilczynski:
Related:

- https://github.com/cri-o/cri-o/pull/8619

kwilczynski:
**To the reviewer:**

I am not sure if there is an RHEL 10 or CentOS Stream 10 image available yet that we could use for 4.18 (CRI-O v1.31) tests.

haircommander:
4.18 should also be on 9.4 AFAIU so the same images as 4.17 should work
/approve

let's make sure the jobs mostly work then we can merge

sohankunkerkar (approved):
/lgtm

PR #57301, opened 2024-09-30
drop deprecated devconsole-api and devconsole-git from redhat-developer

MatousJobanek:
One of our team members (sbryzak) left RedHat. While going through all files he was mentioned in, I figured out that there was devconsole-api and devconsole-git configs (with his name as well as my name). 
These two repos (I was partially responsible for) were deprecated many years ago, so the best way is to remove them from this repo completely.


MatousJobanek:
Never mind

PR #57307, opened 2024-09-30
openshift/operator-framework: require catalogd and operator-controller branches to be up-to-date to merge

joelanford:
The operator-controller and catalogd repositories both include upstream sync automation that requires a consistent view of the downstream repository. If downstream PRs are created after and merge before a previously created sync PR, then subsequent syncs will lose track of the commits that came in with the downstream PR.

In the future it may be possible to teach the syncer to find these "lost" downstream commits, but for now it is simpler to require that PRs are up-to-date with the latest state of the repo. That way sync PRs will not be mergeable if other changes are merged to the `main` branch in the mean time.

See branchprotector docs here: https://docs.prow.k8s.io/docs/components/optional/branchprotector/

### DIFF
--- a/core-services/prow/02_config/openshift/operator-framework-catalogd/_prowconfig.yaml
+++ b/core-services/prow/02_config/openshift/operator-framework-catalogd/_prowconfig.yaml
@@ -1,3 +1,13 @@
+branch-protection:
+  orgs:
+    openshift:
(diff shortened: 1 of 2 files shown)


tmshort:
You might need to do a `make config`?

PR #57329, opened 2024-10-01
[INTEROP-7527] OCP 4.18 expansion for DevSpaces

dfrazzette:
Include OCP 4.18 FIPS-enabled testing. Also add `.config.prowgen`/`slack_reporter` config and default Jira assignee.


calebevans (approved):
/lgtm

PR #57374, opened 2024-10-01
core-services: hypershift: remove per release mirroring

sjenning:


core-services/image-mirroring/hypershift/mapping_hypershift_latest
```diff
@@ -0,0 +1 @@
+registry.ci.openshift.org/hypershift/hypershift-operator:latest quay.io/hypershift/hypershift:latest
```
csrwng, line 1: should be `quay.io/hypershift/hypershift-operator:latest` no?

sjenning: yes, nice catch

PR #57383, opened 2024-10-02
NO-JIRA: remove agent-based installer okd fcos jobs

andfasano:
After the okd 4.15 reshaping the support for agent-based installer was broken, and currently not available, so removing the permafailing job to avoid wasting CI resources


r4f4 (commented):
/lgtm

PR #57392, opened 2024-10-02
Some of the build images were missing the YearIndex in the generated tag

jstourac:
This should fix this for the both notebook repositories we have.

---

I ran into this during my work on the opendatahub-io/notebooks#716. I checked each of our config file we care about with the following command:

```
yq '.tests[].steps.env.IMAGE_TAG' <filename>
```

Any `null` tells us that it is missing expected env. Note that for the `main` branch, we expect that because there are some special testing jobs that don't build images (7 occurrences of null at the end of the file).


jstourac:
@atheo89 @harshad16 please, take a look, thank you!

PR #57394, opened 2024-10-02
OSASINFRA-3624: [CI] Stop e2e-openstack-sdn periodic jobs

MiguelCarpio:
ShiftStack supports OVN Kubernetes network technology by default. Let's be focused only on OVN shiftstack periodic jobs. 

Affected releases:
- 4.12
- 4.13
- 4.14


mandre (commented):
/hold

I don't think we should remove those jobs. These are still supported configuration. Perhaps we may want to run them less often instead.

PR #57395, opened 2024-10-02
Create prow configuration for nbde-tang-server

sarroutbi:


core-services/prow/02_config/medik8s/_prowconfig.yaml
```diff
@@ -12,3 +12,27 @@ tide:
...
+    - lgtm
+    missingLabels:
+    - backports/unvalidated-commits
+    - do-not-merge/hold
+    - do-not-merge/invalid-owners-file
+    - do-not-merge/work-in-progress
+    - jira/invalid-bug
+    - needs-rebase
+    orgs:
+    - medik8s
```
psalajova, line 38: This is already present in the config

sarroutbi: Removed. Thanks for your catch

PR #57422, opened 2024-10-03
Align CI to match OSC upstream repo

gkurz:
Bump the OCP and golang versions to match the current state of OSC upstream.
Switch back to the default bundle Dockerfile which is now track in the OSC repo.

Fixes: https://issues.redhat.com/browse/KATA-3410


ci-operator/config/openshift/sandboxed-containers-operator/openshift-sandboxed-containers-operator-devel.yaml
```diff
@@ -1,33 +1,33 @@
 base_images:
   base:
-    name: "4.13"
+    name: "4.16"
```
bpradipt, line 3: /lgtm

nit: can you also change `OO_CHANNEL: stable-1.3` to `OO_CHANNEL: stable` 

We are only using `stable` channel name.

bpradipt (approved):
/lgtm

PR #57430, opened 2024-10-03
rhds notebooks repository protected branch to false release-2024b

jstourac:
This is required for the GHA to sync to downstream `release-2024b` branch, see #53736.

### DIFF
--- a/core-services/prow/02_config/red-hat-data-services/notebooks/_prowconfig.yaml
+++ b/core-services/prow/02_config/red-hat-data-services/notebooks/_prowconfig.yaml
@@ -10,6 +10,8 @@ branch-protection:
               protect: false
             release-2024a:
               protect: false
+            release-2024b:
+              protect: false
 tide:
   merge_method:
     red-hat-data-services/notebooks: merge


jstourac:
In upstream repo we don't have this protection branch configured for some reason and we seem to be happy with that: https://github.com/openshift/release/blob/master/core-services/prow/02_config/opendatahub-io/notebooks/_prowconfig.yaml

jstourac:
@harshad16 JFYI

harshad16:
> In upstream repo we don't have this protection branch configured for some reason and we seem to be happy with that: https://github.com/openshift/release/blob/master/core-services/prow/02_config/opendatahub-io/notebooks/_prowconfig.yaml

The reason to only have it here , is due to this: https://github.com/openshift/release/pull/53736

jstourac:
> The reason to only have it here , is due to this: #53736

Thanks for the context, Harshad. I updated the description of this. Also, if I'm not missing anything, we want this change, correct? Leaving it open for some lgtm.

harshad16:
> So, I'm a bit confused here. We were able to create and release the release-2024b images without this being merged. Shall I close this PR as unnecessary anymore? 🤔

Yes , we were able to release this ,
lets include this for sanity purpose
and re check on requirement of this bits and document on our end.

harshad16 (approved):
/lgtm

thanks :+1:

PR #57442, opened 2024-10-03
Change MON to 1 for orion cron expression

afcollins:
Changing cron expression for gsheet script compatibility


paigerube14 (approved):
/lgtm

PR #57444, opened 2024-10-03
[DRAFT] OU-405: Adjust the monitoring CI to publish images in quay

zhuje:


ci-operator/config/openshift/troubleshooting-panel-console-plugin/openshift-troubleshooting-panel-console-plugin-release-0.2.yaml
```diff
@@ -20,7 +20,6 @@ images:
 promotion:
   to:
   - excluded_images:
-    - troubleshooting-panel-console-plugin
```
PeterYurkovich, line 23: This looks like it should have been the test image I think

zhuje: Yeah, I think the same.

PR #57485, opened 2024-10-04
rm aws-ovn-ipsec-serial from origin presubmits

jluhrsen:
the job is permafailing


bertinatto:
Job is permafailing and as per Slack discussion we agreed to remove it:

/lgtm

CC @tssurya @pperiyasamy

PR #57410, opened 2024-10-02
HCP/OSP: add job for testing NFV features

EmilienM:
Add a job that execute more tests than the regular job. It'll allow us to iterate on making the Nodepool tests working on OpenStack and also implement our own Nodepool tests for NFV use cases (e.g. additional ports, etc).

PR #57493, opened 2024-10-04
HCP/OSP/e2e: runs NodePool tests

EmilienM:
The NodePool tests are valuable and we're also adding our own nodepool
tests in HCP. We need to exercise them.


EmilienM:
doing it via https://github.com/openshift/release/pull/57410

PR #57504, opened 2024-10-06
[INTEROP-7527] OCP 4.18 expansion for Quay 3.12 lp-interop.

dfrazzette:
Include OCP 4.18 FIPS-enabled testing. Also add `.config.prowgen`/`slack_reporter` config and default Jira assignee.


dfrazzette:
Quay 4.18 FIPS-enabled rehearsal job failed with test failures. Will investigate as part of next monthly OCP 4.18 FIPS-enabled run.

PR #57533, opened 2024-10-07
move ci-tools jobs to multi-arch cluster due to failover build10

smg247:
it also shuffled some other jobs to build11, now that that is the hardcoded multi-arch cluster, when running `make jobs`.


aleskandro:
Are the multiarch-tuning-operator jobs going to build multiarch images in their new cluster?

deepsm007:
> Are the multiarch-tuning-operator jobs going to build multiarch images in their new cluster?

Yes, the new cluster is identical to the current one

PR #57549, opened 2024-10-07
Set cluster-density-v2 warm up threshold

afcollins:
overrides default pod ready threshold for warmup runs that should not care, and even if they did care they don't index metrics so we can't root-cause the failure anyway.

Fixes: cluster-density-v2 warmup run failed recently due to default pod ready threshold.

### DIFF
--- a/ci-operator/step-registry/openshift-qe/node-density-cni/openshift-qe-node-density-cni-commands.sh
+++ b/ci-operator/step-registry/openshift-qe/node-density-cni/openshift-qe-node-density-cni-commands.sh
@@ -22,7 +22,7 @@ pushd e2e-benchmarking/workloads/kube-burner-ocp-wrapper
 export WORKLOAD=node-density-cni
 
 # A non-indexed warmup run
-ES_SERVER="" EXTRA_FLAGS="--pods-per-node=50" ./run.sh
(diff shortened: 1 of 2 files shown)


afcollins:
/pj-rehearse periodic-ci-openshift-qe-ocp-qe-perfscale-ci-main-aws-4.17-nightly-x86-payload-control-plane-6nodes

afcollins:
Results are correct. threshold is set to 2m on the warmup run and still keeps the 13s threshold on the competition run.

```
/tmp/kube-burner-ocp cluster-density-v2 --log-level=info --qps=20 --burst=20 --gc=true --uuid f8f45250-fdf0-4d7d-8d97-ed70093c2612 --pod-ready-threshold=2m --iterations=6 --churn=false
...
/tmp/kube-burner-ocp cluster-density-v2 --log-level=info --qps=20 --burst=20 --gc=true --uuid 53f24042-dfd3-4c2e-b0b2-7ad6773a5d82 --churn-duration=20m --pod-ready-threshold=13s --local-indexing --gc-metrics=true --profile-type=both --iterations=90 --churn=true --es-server=https://XXXXXX:XXXXXXXXXXXXXXXXXXXX@search-XXXXXX-perf-scale-test-elk-hcm7wtsqpxy7xogbu72bor4uve.us-east-1.es.amazonaws.com --es-index=ripsaw-kube-burner
```

Gonna push similar changes for the other node-density* jobs in case they see something similar.

PR #57550, opened 2024-10-07
change order of tests

memodi:


Amoghrd:
Reason for this change?

ci-operator/step-registry/openshift/e2e/test/netobserv-qe/openshift-e2e-test-netobserv-qe-chain.yaml
```diff
@@ -22,8 +22,8 @@ chain:
   steps:
   - ref: idp-htpasswd
   - ref: cucushift-pre
-  - ref: openshift-extended-web-tests
   - ref: openshift-extended-test
+  - ref: openshift-extended-web-tests
```
Amoghrd, line 26: Reason for this change?

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__amd64-nightly.yaml
```diff
@@ -3505,13 +3505,6 @@ tests:
     test:
     - chain: openshift-e2e-test-qe-destructive
     workflow: cucushift-installer-rehearse-vsphere-upi-zones
-- as: vsphere-ipi-amd-f14-netobserv
```
Amoghrd, line 3508: Why vsphere removed from 4.16?

memodi: https://redhat.enterprise.slack.com/docs/T027F3GAJ/F06Q9BMS3HV refer this matrix, 4.17 is new qualifying release and will be replaced by 4.18 soon.

ci-operator/step-registry/openshift/e2e/test/netobserv-qe/openshift-e2e-test-netobserv-qe-chain.yaml
memodi: because in last few runs when there openshift-extended-web-tests step fails, it doesn't get to openshift-extended-test. In past https://redhat-internal.slack.com/archives/C03SHCD5C9M/p1720723976144419  I had brought up similar behavior, however probably it also occurs when the openshift-extended-web-tests never exits in some cases (for us because of test were broken due to this PR https://github.com/openshift/openshift-tests-private/pull/20067) unless timeout is reached . So this change would have openshift-extended-test run first and then run openshift-extended-web-tests in the end

PR #56886, opened 2024-09-18
HOSTEDCP-1569: hypershift: move all repos to use latest HO and e2e tests

sjenning:
Now that hypershift periodic e2es confirm that HO/e2e from hypershift `main` can successfully test all previous releases, cut over all repos to use HO/e2e from hypershift `main`.

PR #57562, opened 2024-10-08
[CIGolang]update tag fro hypershift-operator

kuiwang02:
/cc @Xia-Zhao-rh 

cc @heliubj18 @fxierh

there is no branch tag anymore for hypershift-operator, and only latest now.

now all PR are blocked here, so change to latest tag for a short time. if it needs other tag, please @heliubj18 @fxierh tell us.


fxierh:
@kuiwang02 Thanks for the update. This might be related to [a recent change](https://github.com/openshift/release/pull/56886) from the Hypershift dev team that updates all repos to use the latest Hypershift Operator in CI.

PR #57565, opened 2024-10-08
OCPQE-25553: Disable Azure jobs with OCP MC

fxierh:
Given that:

1. For Azure HCs 4.17 and later, it's required by design to have:
- External DNS enabled (ref. https://github.com/openshift/hypershift/blob/f1f04d30a801a49eb5fef81b0a54464bf7c69699/api/hypershift/v1beta1/hostedcluster_types.go#L351-L355)
- Shared ingress enabled (ref. https://github.com/openshift/hypershift/blob/f1f04d30a801a49eb5fef81b0a54464bf7c69699/cmd/install/install.go#L176)

2. It's currently impossible to host an Azure HC with shared ingress from an OCP MC (ref. https://redhat-internal.slack.com/archives/G01QS0P2F6W/p1721639805710709). 

3. We recently switched to using the latest (4.18) HO for all branches (ref. https://github.com/openshift/release/pull/56886)

It's no longer possible to host Azure HCs from an OCP MC. 

This PR disables Azure job with OCP MC, retaining the job definitions for future reference.

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__amd64-nightly.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__amd64-nightly.yaml
@@ -1611,8 +1611,8 @@ tests:
     - ref: openshift-extended-test-longduration
     - ref: openshift-e2e-test-qe-report
     workflow: cucushift-installer-rehearse-azure-aks-hypershift-install
-- as: azure-ipi-ovn-hypershift-guest-f14
-  cron: 23 18 6,22 * *
+- as: azure-ipi-ovn-hypershift-guest-f360
+  cron: 0 0 29 2 *
   steps:
     cluster_profile: azure-qe
     env:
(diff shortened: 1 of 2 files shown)


fxierh:
/cc @liangxia 

Could you please take a look ?

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.16__amd64-nightly.yaml
```diff
@@ -1611,8 +1611,8 @@ tests:
     - ref: openshift-extended-test-longduration
     - ref: openshift-e2e-test-qe-report
     workflow: cucushift-installer-rehearse-azure-aks-hypershift-install
-- as: azure-ipi-ovn-hypershift-guest-f14
-  cron: 23 18 6,22 * *
+- as: azure-ipi-ovn-hypershift-guest-f365
```
liangxia, line 1614: For historical reasons, the other disabled jobs are using f360. Let's use f360 here, too.

fxierh: Updated

PR #57571, opened 2024-10-08
Adding clusterpool yaml for version 4.17

TomerNewman:
Due to the release of OpenShift 4.17, we need to create a new clusterpool of 4.17 as well.
This commit adds the matching 4.17 clusterpool yaml that will lead to test KMM with 4.17 as well.

---

/assign @ybettan 
/cc @yevgeny-shnaidman

### DIFF
--- a/clusters/hosted-mgmt/hive/pools/rh-ecosystem-edge/ocp-4-17-0_cluster-pool.yaml
+++ b/clusters/hosted-mgmt/hive/pools/rh-ecosystem-edge/ocp-4-17-0_cluster-pool.yaml
@@ -1,20 +1,20 @@
 apiVersion: hive.openshift.io/v1
 kind: ClusterPool
 metadata:
-  name: rh-ecosystem-edge-4-12-cluster-pool
+  name: rh-ecosystem-edge-4-17-cluster-pool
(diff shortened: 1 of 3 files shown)


clusters/hosted-mgmt/hive/pools/rh-ecosystem-edge/ocp-4-17-0_cluster-pool.yaml
```diff
@@ -0,0 +1,29 @@
...
+  skipMachinePools: true
+  platform:
+    aws:
+      credentialsSecretRef:
+        name: aws-credentials # the name of the secret with the credentials of the cloud account
+      region: us-east-1
+  pullSecretRef:
+    name: pull-secret
+  size: 2 # the number of clusters that Hive should keep provisioned and waiting for use.
+  maxSize: 4 # the maximum number of clusters that can exist at the same time.
```
ybettan, line 29:
```suggestion
  size: 6 # the number of clusters that Hive should keep provisioned and waiting for use.
  maxSize: 9 # the maximum number of clusters that can exist at the same time.
```

This pull is the bigger one, therefore, it should have the same size of what we had in 4.15 (and 4.15 needs to be scaled down to the same size of 4.14).

PR #57621, opened 2024-10-08
NO-JIRA: Update/Reduce multi-arch release test cadence

jeffdyoung:
Currently arm64,s390x, ppc64le, and multi release controllers seem to run every 2-3 hours. We don't need to run that often. Proposing this schedule (which is will be less frequent than x86)

4.14: every 24 hrs
4.15: every 12 hrs
4.16: every 8 hrs
4.17: every 8 hrs
4.18: every 6 hrs

thoughts @Prashanth684 @aleskandro @tvardema ?

### DIFF
--- a/core-services/release-controller/_releases/priv/release-ocp-4.14-arm64.json
+++ b/core-services/release-controller/_releases/priv/release-ocp-4.14-arm64.json
@@ -2,6 +2,7 @@
     "expires": "168h",
     "maxUnreadyReleases": 1,
     "message": "<!-- GENERATED FROM PUBLIC ANNOTATION CONFIG - DO NOT EDIT. -->This release contains OSBS official image builds of all arm64 code in release-4.14 (master) branches, and is updated after those builds are synced to quay.io.",
+    "minCreationIntervalSeconds": 86400,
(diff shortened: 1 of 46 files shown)


bradmwilliams:
The release-controller's will only spin new releases if/when the content of the inbound imagestream has changed from the previous release.  Updating this setting will force the release-controller's to wait for the specified amount of time before performing the inbound imagestream comparison. 
As long as this is the behavior you want, then I approve.
/approve

jeffdyoung:
> The release-controller's will only spin new releases if/when the content of the inbound imagestream has changed from the previous release. Updating this setting will force the release-controller's to wait for the specified amount of time before performing the inbound imagestream comparison. As long as this is the behavior you want, then I approve. /approve

@bradmwilliams this is exactly what I'm hoping to do. IMO we (multi-arch) spend too much $ on testing GA'd nightlys' every few hrs and we've had issues with hitting aws/azure/gcp account limits from multiple concurrent runs.  
I assume you did [this](https://github.com/openshift/release/pull/41535/files) for similar reasons?

bradmwilliams:
> I assume you did [this](https://github.com/openshift/release/pull/41535/files) for similar reasons?

Correct

PR #57632, opened 2024-10-09
OSSM maistra update quay cred names

yxun:


FilipB:
@yxun Maybe we could also remove 2.3 jobs in this PR?

FilipB:
@yxun it seems the credential problem is fixed but some jobs are failing because of different issues. Do you want to fix those issues in this PR or in separated PR?

PR #57639, opened 2024-10-09
Removing old CI job file

TomerNewman:
Due to the release of 4.17, we should not test kmm 1.0 anymore with ocp 4.12.
Therefore deleting `rh-ecosystem-edge-kernel-module-management-release-1.0.yaml`

---

/cc @ybettan @yevgeny-shnaidman


ybettan:
I would just remove the `e2e-in-cluster-build` job which is claiming for a 4.12 cluster from our pool which doesn't exist anymore.
It doesn't harm us to still `lint`, `unit-test` etc on rare cherry-picks to 4.12 if we need such in the future.

TomerNewman:
Creating a new PR for this in order for it to be cleaner, closing.

PR #57650, opened 2024-10-09
USHIFT-4570: Reduce conformance periodics frequency to daily

pacevedom:


ci-operator/config/openshift/microshift/openshift-microshift-release-4.18__periodics.yaml
```diff
@@ -80,15 +80,15 @@ tests:
       MICROSHIFT_OS: rhel-9.4
     workflow: openshift-microshift-e2e-metal-tests
 - as: e2e-aws-ovn-ocp-conformance
-  cron: 0 4-19/3 * * *
+  cron: 0 4,16 * * *
```
ggiguash, line 83: Do we need this twice a day? Maybe, only once?

pacevedom: Fair enough. Was thinking about running twice to pick up different time zone changes to MicroShift while either of them are within working hours.
It is true though that now we are blocking and therefore part of the payload jobs. Let me reduce it.

PR #57664, opened 2024-10-09
Add new e2e-monitoring test to library-go

dgrisonnet:
New test added in https://github.com/openshift/library-go/pull/1823

### DIFF
--- a/ci-operator/jobs/openshift/library-go/openshift-library-go-master-presubmits.yaml
+++ b/ci-operator/jobs/openshift/library-go/openshift-library-go-master-presubmits.yaml
@@ -80,6 +80,85 @@ presubmits:
           secretName: result-aggregator
     trigger: (?m)^/test( | .* )e2e-aws-encryption,?($|\s.*)
   - agent: kubernetes
+    always_run: false
+    branches:
+    - ^master$
+    - ^master-
+    cluster: build09
+    context: ci/prow/e2e-monitoring
+    decorate: true
+    labels:
+      ci-operator.openshift.io/cloud: aws
+      ci-operator.openshift.io/cloud-cluster-profile: aws
+      ci.openshift.io/generator: prowgen
+      pj-rehearse.openshift.io/can-be-rehearsed: "true"
+    name: pull-ci-openshift-library-go-master-e2e-monitoring
+    rerun_command: /test e2e-monitoring
+    run_if_changed: ^(pkg/operator/resource/resourceapply/monitoring.go)|^(test/e2e-monitoring)
+    spec:
+      containers:
+      - args:
+        - --gcs-upload-secret=/secrets/gcs/service-account.json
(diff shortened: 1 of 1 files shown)


deads2k:
make sure this runs prior to merge.  Feel free to release after.

/approve
/hold

dgrisonnet:
/pj-rehearse pull-ci-openshift-library-go-master-e2e-monitoring

p0lyn0mial:
`ci/rehearse/openshift/library-go/master/e2e-monitoring` is green.

/lgtm

rexagod:
/hold

Did the test actually run: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_release/57664/rehearse-57664-pull-ci-openshift-library-go-master-e2e-monitoring/1844711757919031296/artifacts/e2e-monitoring/test/build-log.txt?

dgrisonnet:
>Did the test actually run?
> > make: Nothing to be done for 'test-e2e-monitoring'.

This is the expected result, it ran but we just added a placeholder Makefile rule (https://github.com/openshift/library-go/pull/1829)

rexagod:
ACK, apologies for blocking the original PR, and I'm just thinking out loud here, but shouldn't the integration tests actually run on PRs and actively block them if they fail? This is in line with the fact that vetting `master` currently fails for [`encryption_test.go`](https://github.com/openshift/library-go/blob/11a9d18b8fdd34062c3b2e1fae008a83af256450/test/e2e-encryption/encryption_test.go#L85) as that signature was changed, but uncaught. Despite this being a compilation error rather than a logical one, I'd root for an integration suite that runs these tests over a `go vet` step (over a placeholder step).

cc @p0lyn0mial

dgrisonnet:
> ACK, apologies for blocking the original PR, and PLMK if I'm missing something here, but shouldn't the integration tests actually run on PRs and actively block them if they fail? 

Yes but we don't really need to run them if the code path hasn't been touched.

> Despite this being a compilation error rather than a logical one, I'd root for an integration suite that runs these tests over a go vet step

If the reason was logical it would be flagged by the integration test since the code path would be touched. But here it is just a compilation error that can easily be caught once we try to update the dependency in one of the consumer, we don't really need to run the integration suite for that?

dgrisonnet:
The placeholder step is just here because pj-rehearse doesn't support targeting a specific PR. We have to first merge the PR here with a test that is never failing and then update the rule in the library-go PR to test the code.

rexagod:
Ah, I see, the integration tests have been split up into separate targets so they can be run lazily based on the related [codepath changes](https://github.com/openshift/release/pull/57664/files#diff-73874d1076feb7b067c8ca47766353e467799fccbb7438eb09e700670cc5ecbaR60). It seems the `e2e-encryption` suite had its rule here added later on as it's still just a placeholder, but I'll send a patch for it.

PR #57675, opened 2024-10-09
hypershift/openstack: e2e adjustments

EmilienM:
* Don't run parallel tests on NFV, we don't have enough resources for
  the flavor used by nodepools.

* Create a temporary, optional, on-demand CI job that will be used to
  make the nodepool tests working in Hypershift.
  Once we make these tests working, we'll run the tests in
  `e2e-openstack` and remove that job.


EmilienM:
I'm fixing the tests here: https://github.com/openshift/hypershift/pull/4868

PR #57686, opened 2024-10-10
fix and adjust test frequency on cpou jobs

jiajliu:
1. Fix rename issue imported in https://github.com/openshift/release/pull/56612
2. Adjust cpou test frequency from f14 to f28 because the GA target was moved from 4.17 to 4.18 and so there is enough time for the test result collection. And to help reduce the upgrade ci workload at the same time.

cc @jianlinliu @JianLi-RH to review, thank you.


jiajliu:
cc @liangxia to review, thanks.

PR #57689, opened 2024-10-10
chore(tests): use info as log level for dummy model server

vprashar2929:
This commit uses info as a log level for the dummy model server instead of warn


IshwarKanse (approved):
/lgtm

PR #57696, opened 2024-10-10
OSSM-8242: make sure we test using upstream version.yaml

FilipB:
This is required after https://github.com/openshift-service-mesh/sail-operator/pull/146 to be still testing more istio versions from ./versions.yaml

### DIFF
--- a/ci-operator/config/openshift-service-mesh/sail-operator/openshift-service-mesh-sail-operator-main.yaml
+++ b/ci-operator/config/openshift-service-mesh/sail-operator/openshift-service-mesh-sail-operator-main.yaml
@@ -36,6 +36,7 @@ tests:
   commands: |
     make test.unit \
     BUILD_WITH_CONTAINER="0" \
+    VERSIONS_YAML_FILE="versions.yaml" \
     GOBIN="/gobin" \
     GOCACHE="/tmp/cache" \
     GOMODCACHE="/tmp/cache" \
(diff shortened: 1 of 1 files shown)


dgn:
Looks like the one failing test is due to us not force-pushing and the job has already run today. Maybe try deleting the branch in the fork. But overall, I don't think it matters, as the other jobs show that the change works

ci-operator/config/openshift-service-mesh/sail-operator/openshift-service-mesh-sail-operator-main.yaml
```diff
@@ -402,7 +419,10 @@ tests:
            git config --global user.name 'Jonh Wendell'; \
            cd /work && \
            export BUILD_WITH_CONTAINER="0" && \
+           export VERSIONS_YAML_FILE=versions.yaml && \
```
dgn, line 422: this change is not required, we already remove the entire vendor directory and Makefile for the nightlies (as they're supposed to be community nightlies)

FilipB: Right, it's there just for clarity and in case somebody will decide to remove `rm Makefile.vendor.mk` in the future and use just env variables. Should I still remove it or it's ok to keep it?

dgn: hmm. I guess it can stay, does no harm

dgn (approved):

FilipB:
/override ci/rehearse/periodic-ci-openshift-service-mesh-sail-operator-main-nightly-images

PR #57730, opened 2024-10-10
fix image pull in ptp e2e

jzding:


jzding:
/pj-rehearse pull-ci-openshift-ptp-operator-master-operator-e2e

nishant-parekh (commented):
Why use the regex this change should be part of the change to use new release images correct?

jzding:
> Why use the regex this change should be part of the change to use new release images correct?

regex would match `4.18` and replace the matched image with pipeline image.

PR #57731, opened 2024-10-10
update inframachineset ipi commands script

elmiko:
this change is being proposed to update the metrics configuration. The configuration key `k8sPrometheusAdapter` is being deprecated in favor of `metricsServer`. The openshift documentation[0] shows the current recommended values for monitoring.

This change also leaves the previous value in place so that older versions can be tested. Unrecognized keys will be ignored by the monitoring configuration.

[0]: https://docs.openshift.com/container-platform/4.17/observability/monitoring/configuring-the-monitoring-stack.html#configurable-monitoring-components_configuring-the-monitoring-stack


elmiko:
related conversation in red hat slack, https://redhat-internal.slack.com/archives/CBN38N3MW/p1728591246058389

PR #57759, opened 2024-10-11
CORS-3707: ipi/conf/aws: add minimal permissions user config option

r4f4:
If `AWS_INSTALL_USE_MINIMAL_PERMISSIONS` is set to "yes" in an AWS job, the Openshift Installer will be used to generate a permissions policy with the required permissions for installation. Then a user will be provisioned with those permissions and used to run the cluster install.

### DIFF
--- a/ci-operator/config/openshift/installer/openshift-installer-master.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-master.yaml
@@ -446,6 +446,19 @@ tests:
       AWS_METADATA_SERVICE_AUTH: Required
     workflow: openshift-e2e-aws
   timeout: 6h0m0s
+- always_run: false
+  as: e2e-aws-ovn-min-perms
+  optional: true
+  steps:
+    cluster_profile: aws
+    env:
+      AWS_INSTALL_USE_MINIMAL_PERMISSIONS: "yes"
+      USER_TAGS: |
+        keyA valueA
+        keyB valueB
+        keyC valueC
+    workflow: openshift-e2e-aws
+  timeout: 6h0m0s
 - as: e2e-gcp-ovn
   run_if_changed: (gcp|google)
   steps:
--- a/ci-operator/config/openshift/installer/openshift-installer-release-4.18.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-release-4.18.yaml
@@ -447,6 +447,19 @@ tests:
       AWS_METADATA_SERVICE_AUTH: Required
     workflow: openshift-e2e-aws
   timeout: 6h0m0s
+- always_run: false
+  as: e2e-aws-ovn-min-perms
+  optional: true
+  steps:
+    cluster_profile: aws
+    env:
+      AWS_INSTALL_USE_MINIMAL_PERMISSIONS: "yes"
+      USER_TAGS: |
+        keyA valueA
+        keyB valueB
+        keyC valueC
+    workflow: openshift-e2e-aws
+  timeout: 6h0m0s
 - as: e2e-gcp-ovn
   run_if_changed: (gcp|google)
   steps:
--- a/ci-operator/jobs/openshift/installer/openshift-installer-master-presubmits.yaml
+++ b/ci-operator/jobs/openshift/installer/openshift-installer-master-presubmits.yaml
@@ -4159,6 +4159,87 @@ presubmits:
     - ^master$
     - ^master-
     cluster: build05
+    context: ci/prow/e2e-aws-ovn-min-perms
+    decorate: true
+    decoration_config:
+      timeout: 6h0m0s
+    labels:
+      ci-operator.openshift.io/cloud: aws
+      ci-operator.openshift.io/cloud-cluster-profile: aws
+      ci.openshift.io/generator: prowgen
+      pj-rehearse.openshift.io/can-be-rehearsed: "true"
+    name: pull-ci-openshift-installer-master-e2e-aws-ovn-min-perms
+    optional: true
+    rerun_command: /test e2e-aws-ovn-min-perms
+    spec:
+      containers:
+      - args:
+        - --gcs-upload-secret=/secrets/gcs/service-account.json
+        - --image-import-pull-secret=/etc/pull-secret/.dockerconfigjson
+        - --lease-server-credentials-file=/etc/boskos/credentials
+        - --report-credentials-file=/etc/report/credentials
+        - --secret-dir=/secrets/ci-pull-credentials
+        - --secret-dir=/usr/local/e2e-aws-ovn-min-perms-cluster-profile
+        - --target=e2e-aws-ovn-min-perms
+        command:
+        - ci-operator
+        image: ci-operator:latest
+        imagePullPolicy: Always
+        name: ""
+        resources:
+          requests:
+            cpu: 10m
+        volumeMounts:
+        - mountPath: /etc/boskos
+          name: boskos
+          readOnly: true
+        - mountPath: /secrets/ci-pull-credentials
+          name: ci-pull-credentials
+          readOnly: true
+        - mountPath: /usr/local/e2e-aws-ovn-min-perms-cluster-profile
+          name: cluster-profile
+        - mountPath: /secrets/gcs
+          name: gcs-credentials
+          readOnly: true
+        - mountPath: /secrets/manifest-tool
+          name: manifest-tool-local-pusher
+          readOnly: true
+        - mountPath: /etc/pull-secret
+          name: pull-secret
+          readOnly: true
+        - mountPath: /etc/report
+          name: result-aggregator
+          readOnly: true
+      serviceAccountName: ci-operator
+      volumes:
+      - name: boskos
+        secret:
+          items:
+          - key: credentials
+            path: credentials
+          secretName: boskos-credentials
+      - name: ci-pull-credentials
+        secret:
+          secretName: ci-pull-credentials
+      - name: cluster-profile
+        secret:
+          secretName: cluster-secrets-aws
+      - name: manifest-tool-local-pusher
+        secret:
+          secretName: manifest-tool-local-pusher
+      - name: pull-secret
+        secret:
+          secretName: registry-pull-credentials
+      - name: result-aggregator
+        secret:
+          secretName: result-aggregator
+    trigger: (?m)^/test( | .* )e2e-aws-ovn-min-perms,?($|\s.*)
+  - agent: kubernetes
+    always_run: false
+    branches:
+    - ^master$
+    - ^master-
+    cluster: build05
     context: ci/prow/e2e-aws-ovn-proxy
     decorate: true
     decoration_config:
--- a/ci-operator/jobs/openshift/installer/openshift-installer-release-4.18-presubmits.yaml
+++ b/ci-operator/jobs/openshift/installer/openshift-installer-release-4.18-presubmits.yaml
@@ -4157,6 +4157,87 @@ presubmits:
     - ^release-4\.18$
     - ^release-4\.18-
     cluster: build09
+    context: ci/prow/e2e-aws-ovn-min-perms
+    decorate: true
+    decoration_config:
+      timeout: 6h0m0s
+    labels:
+      ci-operator.openshift.io/cloud: aws
+      ci-operator.openshift.io/cloud-cluster-profile: aws
+      ci.openshift.io/generator: prowgen
+      pj-rehearse.openshift.io/can-be-rehearsed: "true"
+    name: pull-ci-openshift-installer-release-4.18-e2e-aws-ovn-min-perms
+    optional: true
+    rerun_command: /test e2e-aws-ovn-min-perms
+    spec:
+      containers:
+      - args:
+        - --gcs-upload-secret=/secrets/gcs/service-account.json
+        - --image-import-pull-secret=/etc/pull-secret/.dockerconfigjson
+        - --lease-server-credentials-file=/etc/boskos/credentials
+        - --report-credentials-file=/etc/report/credentials
+        - --secret-dir=/secrets/ci-pull-credentials
+        - --secret-dir=/usr/local/e2e-aws-ovn-min-perms-cluster-profile
+        - --target=e2e-aws-ovn-min-perms
+        command:
+        - ci-operator
+        image: ci-operator:latest
+        imagePullPolicy: Always
+        name: ""
+        resources:
+          requests:
+            cpu: 10m
+        volumeMounts:
+        - mountPath: /etc/boskos
+          name: boskos
+          readOnly: true
+        - mountPath: /secrets/ci-pull-credentials
+          name: ci-pull-credentials
+          readOnly: true
+        - mountPath: /usr/local/e2e-aws-ovn-min-perms-cluster-profile
+          name: cluster-profile
+        - mountPath: /secrets/gcs
+          name: gcs-credentials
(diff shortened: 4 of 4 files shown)


r4f4:
The installer QE team has a [similar step](https://github.com/openshift/release/tree/master/ci-operator/step-registry/aws/provision/iam-user/minimal-permission) to create an AWS user with minimal permissions that is used only for the QE workflows. Future work can try to combine both approaches into one.

r4f4:
/pj-rehearse pull-ci-openshift-installer-master-e2e-aws-ovn

Existing jobs should be not affected by this change

/pj-rehearse pull-ci-openshift-installer-master-e2e-aws-ovn-min-perms

This job should fail until https://github.com/openshift/installer/pull/8704 merges.

r4f4:
/hold

Until I squash the fixup commits.

r4f4:
/pj-rehearse pull-ci-openshift-installer-master-e2e-aws-ovn pull-ci-openshift-installer-master-e2e-aws-ovn-min-perms

r4f4:
Update: changed installer cmd invocation after updates to https://github.com/openshift/installer/pull/8704/ and added installer QE team members as owners.

/cc @yunjiang29 @jianlinliu

r4f4:
/pj-rehearse pull-ci-openshift-installer-master-e2e-aws-ovn pull-ci-openshift-installer-master-e2e-aws-ovn-min-perms

r4f4:
* `e2e-aws-ovn-min-perms`: failed as expected.
* `e2e-aws-ovn`: succeeded as expected and we see:
```
Custom AWS user with minimal permissions is disabled. Using AWS user from cluster profile.
```
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_release/57759/rehearse-57759-pull-ci-openshift-installer-master-e2e-aws-ovn/1845084211640274944/artifacts/e2e-aws-ovn/ipi-conf-aws-min-perms-user/build-log.txt
```
/tmp/secret/aws_user_names is missing.
/tmp/secret/aws_policy_arns is missing.
```
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_release/57759/rehearse-57759-pull-ci-openshift-installer-master-e2e-aws-ovn/1845084211640274944/artifacts/e2e-aws-ovn/aws-deprovision-users-and-policies/build-log.txt

ci-operator/step-registry/ipi/conf/aws/min-perms-user/ipi-conf-aws-min-perms-user-commands.sh
```diff
@@ -0,0 +1,131 @@
...
+  echo "Custom AWS user with minimal permissions is not supported in target latest version ${ocp_version}; bypassing configuration"
+  exit 0
+fi
+
+export AWS_SHARED_CREDENTIALS_FILE="${CLUSTER_PROFILE_DIR}/.awscred"
+
+REGION="${LEASED_RESOURCE}"
+CLUSTER_NAME="${NAMESPACE}-${UNIQUE_HASH}"
+
+installer_generate_policies
```
jianlinliu, line 96: QE developed a step to provision a minimal permission user to run install, like `aws-provision-iam-user-minimal-permission`. But we were using a static permission policy list, now sounds 4.18 installer would have a new command to create the permission policy list by `openshift-install create permissions-policy`. 

Shall we merge the new command to `aws-provision-iam-user-minimal-permission`, so that we can keep the same step to cover the minimal permission user installation on all versions? cc @yunjiang29

r4f4: @jianlinliu I was thinking about doing that as a follow-up, once we have the `create permissions-policy` working. My idea was to merge these changes first so I can test the installer PR introducing the new command at https://github.com/openshift/installer/pull/8704

jianlinliu: Sounds a plan for me.

ci-operator/step-registry/ipi/conf/aws/min-perms-user/ipi-conf-aws-min-perms-user-commands.sh
```diff
@@ -0,0 +1,131 @@
...
+  RELEASE_IMAGE_INSTALL="${RELEASE_IMAGE_LATEST:-}"
+fi
+cp ${CLUSTER_PROFILE_DIR}/pull-secret /tmp/pull-secret
+oc registry login --to /tmp/pull-secret
+ocp_version=$(oc adm release info --registry-config /tmp/pull-secret ${RELEASE_IMAGE_INSTALL} --output=json | jq -r '.metadata.version' | cut -d. -f 1,2)
+ocp_major_version=$( echo "${ocp_version}" | awk --field-separator=. '{print $1}' )
+ocp_minor_version=$( echo "${ocp_version}" | awk --field-separator=. '{print $2}' )
+rm /tmp/pull-secret
+
+if (( ocp_major_version < 4 || ( ocp_major_version == 4 && ocp_minor_version < 18 ))); then
```
jianlinliu, line 86: If we are using `AWS_INSTALL_USE_MINIMAL_PERMISSIONS` to control the behavior, do we still need the version judgement here ? If some jobs enabled `AWS_INSTALL_USE_MINIMAL_PERMISSIONS` on an unsupported version, that sounds like a job config issue.

r4f4: Right. I added this because, once it's all working, we could invert the logic for the `AWS_INSTALL_USE_MINIMAL_PERMISSIONS` to make it opt-out by default, instead of opt-in. I can drop this until that's the case.

ci-operator/step-registry/ipi/conf/aws/min-perms-user/ipi-conf-aws-min-perms-user-commands.sh
```diff
@@ -0,0 +1,131 @@
...
+fi
+
+export AWS_SHARED_CREDENTIALS_FILE="${CLUSTER_PROFILE_DIR}/.awscred"
+
+REGION="${LEASED_RESOURCE}"
+CLUSTER_NAME="${NAMESPACE}-${UNIQUE_HASH}"
+
+installer_generate_policies
+
+USER_POLICY_FILE="${SHARED_DIR}/aws-permissions-policy-mint-creds.json"
```
mtulio, line 98: are you planning to keep it fixed for mint now, then update to support other modes?

r4f4: Ops, nice catch! The installer will generate only one file and its contents will depend on the `CredentialsMode` set in the install-config.yaml. I'll fix this.

mtulio:
> * `e2e-aws-ovn-min-perms`: failed as expected.
> * `e2e-aws-ovn`: succeeded as expected and we see:
> 

/lgtm
/approve

ci-operator/step-registry/ipi/conf/aws/min-perms-user/ipi-conf-aws-min-perms-user-ref.yaml
```diff
@@ -0,0 +1,21 @@
+ref:
+  as: ipi-conf-aws-min-perms-user
+  from_image:
+    namespace: ocp
+    name: "4.18"
```
jianlinliu, line 5: Because the step has dependency of installer, sounds like we can not use specific version tag of upi-installer image, but use `from: upi-installer`, so that the step can used is different versions of OCP ci jobs.

ci-operator/step-registry/ipi/conf/aws/min-perms-user/ipi-conf-aws-min-perms-user-commands.sh
```diff
@@ -0,0 +1,131 @@
...
+    eval "${cmd}"
+}
+
+function installer_generate_policies()
+{
+	local dir=/tmp/min_perms/
+
+	mkdir -p ${dir}
+
+	# Make a copy of the install-config.yaml since the installer will consume it.
```
yunjiang29, line 22: > installer will consume it

`create manifests` and `create ignition-configs` consume config file as well, what's the correct order after `permissions-policy` is added? is `permissions-policy` -> `manifests` -> `ignition-configs` ?

ci-operator/step-registry/ipi/conf/aws/min-perms-user/ipi-conf-aws-min-perms-user-commands.sh
yunjiang29: sounds good to me

ci-operator/step-registry/ipi/conf/aws/min-perms-user/ipi-conf-aws-min-perms-user-commands.sh
r4f4: As long as the workdir is preserved, you can run `create permissions-policy` at any point after `create install-config` since the installer will use either the `install-config.yaml` file from disk or from the state file `<workdir>/.openshift_install_state.json`. The `permissions-policy` commands generates the same set of permissions that is used during install-config validation.

ci-operator/step-registry/ipi/conf/aws/min-perms-user/ipi-conf-aws-min-perms-user-ref.yaml
r4f4: The QE step uses `name: "4.12"`. So you think I can remove the `name: "4.18"` part?

ci-operator/step-registry/ipi/conf/aws/min-perms-user/ipi-conf-aws-min-perms-user-commands.sh
r4f4: The ignition configs depend on the manifests (you could edit the manifests generated by the installer), so `manifests -> ignition-configs` is the correct order.

ci-operator/step-registry/ipi/conf/aws/min-perms-user/ipi-conf-aws-min-perms-user-ref.yaml
jianlinliu: I meant for this step, we should use:
```
from: upi-installer
```
instead of:
```
from_image:
    namespace: ocp
    name: "4.18"
    tag: upi-installer
```
So that the step would rely on upi-installer base image setting in ci yaml file.

E.g:  upi-installer base image is supposed to be 4.19 in 4.19 ci yaml file, , then for this step, it would use 4.19 installer to run `create permissions-policy`, but not use a fixed version of upi-installer to run this step.

r4f4: Fair enough. I'll change it.

ci-operator/step-registry/ipi/conf/aws/min-perms-user/ipi-conf-aws-min-perms-user-commands.sh
r4f4: Fixed.

ci-operator/step-registry/ipi/conf/aws/min-perms-user/ipi-conf-aws-min-perms-user-commands.sh
r4f4: I left the check so the step fails with an error message if a job is misconfigured. It won't matter once we merge this with QE's step to generate a policy for all release versions.

r4f4:
Update: addressed review comments.

r4f4:
/pj-rehearse pull-ci-openshift-installer-master-e2e-aws-ovn-min-perms pull-ci-openshift-installer-master-e2e-aws-ovn pull-ci-openshift-installer-release-4.16-e2e-aws-ovn

ci-operator/step-registry/ipi/conf/aws/min-perms-user/ipi-conf-aws-min-perms-user-commands.sh
```diff
@@ -0,0 +1,129 @@
...
+export AWS_SHARED_CREDENTIALS_FILE="${CLUSTER_PROFILE_DIR}/.awscred"
+
+REGION="${LEASED_RESOURCE}"
+CLUSTER_NAME="${NAMESPACE}-${UNIQUE_HASH}"
+
+if ((ocp_major_version < 4 || (ocp_major_version == 4 && ocp_minor_version < 18))); then
+	# FIXME: generate an ad-hoc policy here
+	echo "Openshift installer cannot generate permissions policy prior to release 4.18"
+	exit 1
+else
```
barbacbd, line 92: nit: this else is not needed right?

r4f4: I left it as a placeholder for when we address the `# FIXME` comment.

barbacbd (commented):
/lgtm

ci-operator/step-registry/ipi/conf/aws/min-perms-user/ipi-conf-aws-min-perms-user-commands.sh
```diff
@@ -0,0 +1,129 @@
...
+	installer_generate_policies
+fi
+
+USER_POLICY_FILE="${SHARED_DIR}/aws-permissions-policy-creds.json"
```
jianlinliu, line 96: The file is supposed to be generated from `installer_generate_policies`, right? How about use the same name in `installer_generate_policies` instead of `*-creds.json` to keep consistent, help users easy locate where the file is generated. Or make `installer_generate_policies` to return the file name, assign the value to `USER_POLICY_FILE`.

paigerube14:
@r4f4 this change effects a common workflow that many people/teams use and is now causing them to fail because its not able to find the image

https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-qe-ocp-qe-perfscale-ci-main-aws-4.18-nightly-x86-payload-control-plane-6nodes/1846559713655787520
`Failed to pull image "image-registry.openshift-image-registry.svc:5000/ci-op-04k80f8p/stable:upi-installer": reading manifest upi-installer in image-registry.openshift-image-registry.svc:5000/ci-op-04k80f8p/stable: manifest unknown`

The way the **from: <image>** is defined would require every user of the workflow to add the upi-installer as a base image to each of the config files. Are we able to update the ref to be similar to 
[ipi-conf-aws](https://github.com/openshift/release/blob/b6efdd4b4dbefbcf4b87975dc5efdfdaae3432bf/ci-operator/step-registry/ipi/conf/aws/ipi-conf-aws-ref.yaml#L3-L6), that every team using this workflow won't have to edit their configs. 

I opened PR for an update on this: https://github.com/openshift/release/pull/57893

r4f4:
@paigerube14 thanks for catching that. That's what I had before but I changed it after a review feedback https://github.com/openshift/release/pull/57759#discussion_r1800335922. Let's continue discussions in your PR.

PR #57768, opened 2024-10-12
Update ROSA e2e for ComplianceAsCode/content CI

rhmdnd:
A billing account is now required. Update the env so that CI supplies it
when provisioning ROSA environments. Without this CI fails.

### DIFF
--- a/ci-operator/config/ComplianceAsCode/content/ComplianceAsCode-content-master__4.15.yaml
+++ b/ci-operator/config/ComplianceAsCode/content/ComplianceAsCode-content-master__4.15.yaml
@@ -558,6 +558,7 @@ tests:
     env:
       BYO_OIDC: "true"
       CHANNEL_GROUP: stable
+      ENABLE_BILLING_ACCOUNT: "yes"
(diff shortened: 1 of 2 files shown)


rhmdnd:
@Vincent056 @yuumasato this should resolve the ROSA e2e issues for cac/content CI.

ci-operator/config/ComplianceAsCode/content/ComplianceAsCode-content-master__4.15.yaml
```diff
@@ -558,6 +558,7 @@ tests:
     env:
       BYO_OIDC: "true"
       CHANNEL_GROUP: stable
+      ENABLE_BILLING_ACCOUNT: "yes"
```
yuumasato, line 561: I see that we are "locked" onto 4.15 for the e2e-rosa tests.
Is it because it is only supported on 4.15? Or is it to reduce number of tests (reduce costs)?

yuumasato (approved):
/approve

PR #57783, opened 2024-10-14
Rename longduration to longrun

liangxia:
Rename several `longduration` test chains to `longrun`, so that we can rename the test name accordingly in another PR.
Currently several test name already reached the max char limits, which block us from adjust test frequency in them ( from `fN` to `fMN` which increase the test name to one or two more chars).


kuiwang02:
I am ok.

sergiordlr:
/lgtm
thank you!

PR #57786, opened 2024-10-14
OCPQE-26275: Drop Hypershift AKS jobs on 4.16

fxierh:
Remove Hypershift AKS jobs for 4.16 as ARO-HCP will move to 4.17. Additionally, these jobs have been unstable due to compatibility issues.

This will significantly reduce our costs. 

ref. https://redhat-internal.slack.com/archives/G01QS0P2F6W/p1728647323291399


fxierh:
/cc @heliubj18 

Could you please take a look ?

fxierh:
/cc @liangxia 

Could you please take a look ?

PR #57787, opened 2024-10-14
MULTIARCH-4972: remove multiarch team non-working cluster pools

lwan-wanglin:
there is an epic [MULTIARCH-4850](https://issues.redhat.com/browse/MULTIARCH-4850) for **updating ClusterClaims target environments in PROW to have available multi-arch targets**, so seems we don't need to add ourselves clusterpool, becasue currently except for amd64 we need to update clusterimageset manually for other arches  and also hit other difficultlies like our accout enabled auto-prune,  so **removing the multiarch team non-working cluster pools and wait for epics done**.

cc @aleskandro to check


aleskandro:
/lgtm
/assign @deepsm007 
Can you merge?

PR #57790, opened 2024-10-14
MGMT-18937: Update OCP 4.18/4.19/master external jobs which rely on assisted images to ocm-2.13 (latest)

danmanor:
Part of ocm-2.12 feature freeze


adriengentil (approved):

adriengentil:
> @adriengentil what do you think about these failures ?

They do not look related to any assisted jobs ?

PR #57750, opened 2024-10-11
NO-JIRA: cleaning up aks guest annotations logic for hypershift

Patryk-Stefanski:

PR #57823, opened 2024-10-14
Revert "NO-JIRA: cleaning up aks guest annotations logic for hypershift"

bryan-cox:
Reverts openshift/release#57750


sjenning:
@fxierh can we get approval to fix the `e2e-aks` regression?

PR #40212, opened 2023-06-09
FIO: Bump the pullspec

jhrozek:
Needed for a clean upgrade run.

PR #38215, opened 2023-04-11
Bump the FIO pull spec to match the latest version

rhmdnd:
We do this to workaround issues in upgrade testing. This needs to be updated to the 1.2.1 version so that the upgrade job pulls the correct version.

PR #57827, opened 2024-10-14
Update bundle version for FIO upgrade tests

rhmdnd:
This commit updates the bundle test to use the latest version of FIO.

### DIFF
--- a/ci-operator/config/openshift/file-integrity-operator/openshift-file-integrity-operator-master.yaml
+++ b/ci-operator/config/openshift/file-integrity-operator/openshift-file-integrity-operator-master.yaml
@@ -29,7 +29,9 @@ operator:
     context_dir: .
     dockerfile_path: bundle.Dockerfile
   substitutions:
-  - pullspec: quay.io/file-integrity-operator/file-integrity-operator:1.3.0
+  - pullspec: quay.io/file-integrity-operator/file-integrity-operator:1.3.3
+    with: pipeline:file-integrity-operator
+  - pullspec: quay.io/file-integrity-operator/file-integrity-operator:1.3.4
     with: pipeline:file-integrity-operator
 promotion:
   to:
@@ -134,7 +136,7 @@ tests:
     env:
       OO_CHANNEL: alpha
       OO_INSTALL_NAMESPACE: file-integrity
-      OO_LATEST_CSV: file-integrity-operator.v1.3.0
+      OO_LATEST_CSV: file-integrity-operator.v1.3.3
       OO_PACKAGE: file-integrity-operator
       OO_TARGET_NAMESPACES: '!install'
     test:


yuumasato (approved):
/approve

yuumasato:
@rhmdnd based on https://github.com/openshift/release/pull/40212,

Maybe you also need to replace the `substitutions:` key?

Here the image is updated to use the tag: 1.3.4: https://github.com/openshift/file-integrity-operator/pull/529/files#diff-3f66712777bc246da12ba3733e948beb641703ff242abf09f8c095234933687eR102
But it doesn't exist yet. It is pushed once the release PR is merged, right?

rhmdnd:
> @rhmdnd based on #40212
> 
> Maybe you also need to replace the `substitutions:` key?
> 
> Here the image is updated to use the tag: 1.3.4: https://github.com/openshift/file-integrity-operator/pull/529/files#diff-3f66712777bc246da12ba3733e948beb641703ff242abf09f8c095234933687eR102
> 
> But it doesn't exist yet: https://quay.io/repository/file-integrity-operator/file-integrity-operator?tab=tags
> 
> It is pushed once the release PR is merged, right?

Yeah - you're right. We've updated both in the past.

https://github.com/openshift/release/pull/38215

rhmdnd:
/pj-rehearse pull-ci-openshift-file-integrity-operator-master-e2e-bundle-aws-upgrade

ci-operator/config/openshift/file-integrity-operator/openshift-file-integrity-operator-master.yaml
```diff
@@ -29,7 +29,7 @@ operator:
     context_dir: .
     dockerfile_path: bundle.Dockerfile
   substitutions:
-  - pullspec: quay.io/file-integrity-operator/file-integrity-operator:1.3.0
+  - pullspec: quay.io/file-integrity-operator/file-integrity-operator:1.3.3
```
yuumasato, line 32: @rhmdnd What if we had substitutions for the current tag `1.3.3` and the future one `1.3.4`?
This way it will be a bit more future proof.

There can be more than on substitutions:
https://docs.ci.openshift.org/docs/how-tos/testing-operator-sdk-operators/#building-operator-bundles

yuumasato: It is not clear to me in which order the bundle update in CI happens compared to the release PR. Which one is merged first?

It seems to me that this one comes first, as it has substitutions to make the release PR work.

rhmdnd:
/pj-rehearse pull-ci-openshift-file-integrity-operator-master-e2e-bundle-aws-upgrade

ci-operator/config/openshift/file-integrity-operator/openshift-file-integrity-operator-master.yaml
rhmdnd: Correct - we have to merge this one first, so the bundle upgrade will work in the release PR. Then when we land the release PR, I expect we'll need to come back an bump this so that it's using 1.3.4 (the latest release).

ci-operator/config/openshift/file-integrity-operator/openshift-file-integrity-operator-master.yaml
```diff
@@ -134,7 +136,7 @@ tests:
     env:
       OO_CHANNEL: alpha
       OO_INSTALL_NAMESPACE: file-integrity
-      OO_LATEST_CSV: file-integrity-operator.v1.3.0
+      OO_LATEST_CSV: file-integrity-operator.v1.3.3
```
rhmdnd, line 139: This should probably be `file-integrity-operator.v1.3.4` based on the documentation:

_OO_LATEST_CSV 	The name of the CSV being upgraded to._

yuumasato: And I wonder why is `INITIAL_CSV` is not set.

````
INITIAL_CSV 	Parameter[[?](https://docs.ci.openshift.org/docs/architecture/step-registry/#parameters)] 
(For upgrade tests) The name of the initial CSV to install.
```

rhmdnd:
This may not be necessary depending on the outcome of https://github.com/openshift/release/pull/57930

rhmdnd:
Ended up getting bundle upgrades working in https://github.com/openshift/release/pull/57930 with an approach that doesn't require us to constantly bump the bundle version.

PR #57842, opened 2024-10-15
    HOSTEDCP-1472: add hypershift n1,n2 release image params to aks e2e for NodePoolPrevReleaseCreateTest

Patryk-Stefanski:


ci-operator/config/openshift/hypershift/openshift-hypershift-main.yaml
```diff
@@ -79,6 +79,18 @@ releases:
       product: ocp
       stream: ci
       version: "4.17"
+  n-1minor:
+    candidate:
+      product: ocp
+      relative: 1
```
sjenning, line 85: don't need `relative: 1`

ci-operator/config/openshift/hypershift/openshift-hypershift-main.yaml
```diff
@@ -79,6 +79,18 @@ releases:
...
+      stream: ci
+      version: "4.17"
+  n-2minor:
+    candidate:
+      product: ocp
+      relative: 1
```
sjenning, line 91: don't need `relative: 1`

sjenning (changes requested):

sjenning:
Also, this only changes the `aks` workflow.  We should do this for the `aws` workflow as well.

sjenning:
/hold

until https://github.com/openshift/hypershift/pull/4911 merges

PR #57855, opened 2024-10-15
[sg-core] Onboard sg-core to openstack-k8s-operators

vkmc:


paramite:
I don't see any issues with this PR. I'm not sure though if anything can be missing, given the fact that I have experience with adding jobs for openstack-operators only and sg-core is not an operator.

danilo-gemoli (approved):

PR #57876, opened 2024-10-16
Update RHDH nightlies Slack reporter URLs

zdrapela:
Update the RHDH nightlies to use emojis and add ReportPortal launch URL.

Fixes: https://issues.redhat.com/browse/RHIDP-4371


zdrapela:
We will add another reporting Slack workflow instead.

PR #52625, opened 2024-05-30
ZTP Telco Left shifting

ccardenosa:
The intention of this PR is to use the Openshift-CI infrastructure to deploy and test in an automated way the nightly builds of the OCP to be used in Telco deployments using ZTP to deploy Spoke clusters, so that, in case of any problem, the corresponding OCP release can be requested to be blocked.

For this purpose, it has been decided to divide this task into 2 main blocks, this one for the hub cluster part, which will be deployed once every 1 or 2 weeks, keeping them active (i.e. Openshift-CI will not decommission the hub cluster) for that period (which should be configurable).

The main activities performed in this PR are:

- Add PR debug mode to help during PR development for a better control of the work during the development of the functions
- Deploy SNO Hub cluster and share its profile for persistency so that the cluster is not decommissioned once the job is finished and its access data is stored in a known location so that jobs that make use of this Hub cluster can interact with it.
- Set the expiration time for the OCP clusters so that, on a weekly basis, it is reinstalled using a more updated version. In this case, the new jobs will deploy an SNO Hub cluster with OCP version 4.17 with Creation time: On Saturday, 07.00 UTC and Expiration time: On Friday 20.00 UTC.
- Add the option to generate PreGA catalogs and improve catalog selection. Due to the availability of the catalogs in Pre GA and GA versions varies in the OCP versions we want to test, the intention of this change is to provide more flexibility in selecting the catalog to use.

The other main block will be the Spoke cluster deployment [see the related PR [here](https://github.com/openshift/release/pull/53610)], using the ACM infra deployed in the Hub cluster for each OCP nightly builds.

PR #57888, opened 2024-10-16
Let the `wait` step's changes be backward compatible

aleskandro:
#52625 let the wait step to execute in PR only when used in a workflow. However, some jobs were mistakenly using it in non-presubmit jobs.

This commit restore the previous behavior as default, still allowing the in-PR only logic to work when requested.

### DIFF
--- a/ci-operator/step-registry/telcov10n/sno-hub-ztp/telcov10n-sno-hub-ztp-workflow.yaml
+++ b/ci-operator/step-registry/telcov10n/sno-hub-ztp/telcov10n-sno-hub-ztp-workflow.yaml
@@ -10,5 +10,8 @@ workflow:
     post:
     - ref: wait
     - chain: baremetal-lab-post
+    env:
+      TIMEOUT: "+10 hour"
+      PR_ONLY: "true"
   documentation: |-
     Install a hub cluster using Agent Based Installer (ABI) and then deprovision the cluster.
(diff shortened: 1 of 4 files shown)


ci-operator/step-registry/wait/wait-ref.yaml
```diff
@@ -1,21 +1,23 @@
 ref:
   as: wait
-  from: src
+  from: cli
```
ccardenosa, line 3: Is it `cli: latest` no longer needed?

aleskandro: never needed i think.

ci-operator/step-registry/telcov10n/sno-hub-ztp/telcov10n-sno-hub-ztp-workflow.yaml
```diff
@@ -10,5 +10,8 @@ workflow:
     post:
     - ref: wait
     - chain: baremetal-lab-post
+    env:
+      TIMEOUT: "+30 min"
```
ccardenosa, line 14: Remove [this line](https://github.com/ccardenosa/release/blob/master/ci-operator/config/openshift-kni/eco-ci-cd/openshift-kni-eco-ci-cd-main__sno-hub-ztp-4.17-latest.yaml#L54) then

PR #57893, opened 2024-10-16
Adding upi installer base image

paigerube14:
Adding a base version image

### DIFF
--- a/ci-operator/step-registry/ipi/conf/aws/min-perms-user/ipi-conf-aws-min-perms-user-ref.yaml
+++ b/ci-operator/step-registry/ipi/conf/aws/min-perms-user/ipi-conf-aws-min-perms-user-ref.yaml
@@ -1,6 +1,9 @@
 ref:
   as: ipi-conf-aws-min-perms-user
-  from: upi-installer
+  from_image:
+    namespace: ocp
+    name: "4.14"
+    tag: upi-installer
   grace_period: 10m
   commands: ipi-conf-aws-min-perms-user-commands.sh
   resources:


ci-operator/step-registry/ipi/conf/aws/min-perms-user/ipi-conf-aws-min-perms-user-ref.yaml
```diff
@@ -1,6 +1,9 @@
 ref:
   as: ipi-conf-aws-min-perms-user
-  from: upi-installer
+  from_image:
+    namespace: ocp
+    name: "4.14"
```
r4f4, line 5: I think we cannot hardcode this to the 4.14 version since the functionality we need from the installer will be introduced in 4.18.

r4f4: But I can also fix that in a follow-up so we get CI jobs working again.

paigerube14: Understand, is the 4.18 version ready for upi-installer that we could specify that version instead? Or does the functionality not work for previous versions

Does that ref need to be in a common workflow or can you add it in your specific use case, as this change touches 10025 jobs

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pj-rehearse/openshift/release/57893/ba9d4c6d99af618be6534c734dcf905ca07d199e

jianlinliu: Maybe we can use `from: installer` to run the step, then use phthon json module to replace `jq` to run those commands in the step, so that satisfy everyone.

r4f4: @jianlinliu we can't because we need the aws sdk. Let's discuss on https://github.com/openshift/release/pull/57897

PR #57903, opened 2024-10-16
Add slack notifications for fail/error oadp-operator periodics

kaovilai:
Intention long term is to add images build failures also to slack but will follow later when https://issues.redhat.com/browse/DPTP-4213 completes.

### DIFF
--- a/ci-operator/config/openshift/oadp-operator/.config.prowgen
+++ b/ci-operator/config/openshift/oadp-operator/.config.prowgen
@@ -1,1 +1,15 @@
 multi_arch: yes
+slack_reporter:
+- channel: "#oadp-cicd"
+  job_states_to_report:
+  - failure
+  - error
+  report_template: '{{if eq .Status.State "success"}} :rainbow: Job *{{.Spec.Job}}*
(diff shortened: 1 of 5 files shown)


ci-operator/config/openshift/oadp-operator/.config.prowgen
```diff
@@ -1 +1,15 @@
 multi_arch: yes
+slack_reporter:
+- channel: "#oadp-cicd"
+  job_states_to_report:
+  - failure
+  - error
+  report_template: '{{if eq .Status.State "success"}} :rainbow: Job *{{.Spec.Job}}*
```
mateusoliveira43, line 7: there is a preview of the message we can check to see how it will look?

kaovilai: another repo with same exact template: https://github.com/kaovilai/release/blob/e6751376bf8e796fd7ac1e776dd51c2dd70be55a/ci-operator/config/openshift/image-based-install-operator/.config.prowgen#L7-L10

https://redhat.enterprise.slack.com/archives/C078MLC0SP4

mateusoliveira43 (approved):

PR #57907, opened 2024-10-16
rhacs scanner 2.35

RTann:
/cc @BradLugo @dcaravel @daynewlee @jvdm

Starting Scanner 2.35 release process. The release branch doesn't exist yet, but that's ok. This is copied from the release-2.34 file, and there are not (expected) changes. It should be ok to merge this so it's available when we cut the release branch


BradLugo (approved):
/lgtm

PR #57951, opened 2024-10-17
`tide`: set global `optional-regex-contexts` for `multi-pr` presubmit runs

smg247:
These `check_runs` should always be optional in `tide`.

For: https://issues.redhat.com/browse/DPTP-4148


smg247:
The `ci/prow/prow-config` check won't succeed until we bump prow in `ci-tools`. I will work on that

smg247:
This won't function properly until https://github.com/kubernetes-sigs/prow/pull/304 is merged and we update tide to that version, but it is safe to merge this now anyways

PR #57963, opened 2024-10-18
[PERFSCALE-3525] Add OVN PerfScale AWS Cluster Profile

krishvoor:
Add OVN PerfScale AWS Cluster Profile
Depends on: https://github.com/openshift/ci-tools/pull/4379

### DIFF
--- a/core-services/ci-secret-bootstrap/_config.yaml
+++ b/core-services/ci-secret-bootstrap/_config.yaml
@@ -2259,7 +2259,7 @@ secret_configs:
     name: cluster-secrets-azure-qe
     namespace: ci
   - cluster_groups:
-    - non_app_ci:
+    - non_app_ci
     name: cluster-secrets-aws-ovn-perfscale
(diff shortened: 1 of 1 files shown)


krishvoor:
/assign @smg247 

@smg247 Can you help review this PR, thank you!

core-services/ci-secret-bootstrap/_config.yaml
```diff
@@ -2258,6 +2258,10 @@ secret_configs:
     - non_app_ci
     name: cluster-secrets-azure-qe
     namespace: ci
+  - cluster_groups:
+    - non_app_ci:
```
jtaleric, line 2262: i don't think there should be a `:` at the end of this line?

krishvoor: Good catch, updated the logic!

krishvoor (commented):
/lgtm

PR #57980, opened 2024-10-18
Only run bundle e2e when bundle files change for FIO

rhmdnd:
Let's be more conservative with our testing resources and time for
bundle testing since we're already running e2e on each PR. Let's make
sure we're only running bundle e2e on changes that actually modify the
bundle.


rhmdnd:
/pj-rehearse ack

Applying the ack now since we know these jobs pass, and this change is only modifying when they run.

PR #58019, opened 2024-10-21
Use latest openshift virt for ovn-k master jobs.

qinqon:


qinqon:
/hold

Wait for @jluhrsen

ci-operator/step-registry/kubevirt/install/kubevirt-install-commands.sh
```diff
@@ -29,7 +29,11 @@ elif [ -n "${CNV_PRERELEASE_CATALOG_IMAGE}" ]
   then
   CNV_RELEASE_CHANNEL=stable
 else
-  CNV_RELEASE_CHANNEL=nightly-$(ocp_version)
+  if [ "${CNV_PRERELEASE_LATEST_CHANNEL}" == "true" ]; then
```
jluhrsen, line 32: why not make this the actual channel name or version string so you can tweak it to other values
in the future if needed?

default would be ""

so, code would be something like:

if [ "${CNV_PRERELEASE_LATEST_CHANNEL_VERSION}" != "" ]; then
 CNV_RELEASE_CHANNEL=nightly-${CNV_PRERELEASE_LATEST_CHANNEL_VERSION}
  else
    CNV_RELEASE_CHANNEL=nightly-$(ocp_version)
  fi

qinqon: We don't want to test wrong channel versions at master branch, also this is a very temporal knob that will be removed after stuff settle up at kubevirt.

jluhrsen: ok, sounds like you think it's overkill then.

/lgtm

qinqon:
/hold

Kubevirt installtion not working.

qinqon:
All good, master jobs are failing at not finding the test suite (as expected) and CNV is installed with primary UDN binding 
```
 featureGates:                                                               
      alignCPUs: false                                                          
      autoResourceLimits: false                                                 
      deployKubeSecondaryDNS: false                                             
      deployKubevirtIpamController: true                                        
      deployTektonTaskResources: false                                          
      deployVmConsoleProxy: false                                               
      disableMDevConfiguration: false                                           
      downwardMetrics: false                                                    
      enableApplicationAwareQuota: false                                        
      enableCommonBootImageImport: false                                        
      enableManagedTenantQuota: false                                           
      nonRoot: true                                                             
      persistentReservation: false                                              
      primaryUserDefinedNetworkBinding: true                                    
      withHostPassthroughCPU: false           
```

PR #58077, opened 2024-10-22
CORS-3704: GCP Tech Preview Job Custom DNS

barbacbd:
** Create new tech preview job for user provisioned DNS installs (GCP)

### DIFF
--- a/ci-operator/config/openshift/installer/openshift-installer-release-4.18.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-release-4.18.yaml
@@ -475,6 +475,16 @@ tests:
       FEATURE_SET: TechPreviewNoUpgrade
     workflow: openshift-e2e-gcp
   timeout: 6h0m0s
+- always_run: false
+  as: e2e-gcp-user-provisioned-dns
+  optional: true
+  steps:
+    cluster_profile: gcp
+    env:
+      FEATURE_SET: TechPreviewNoUpgrade
+      USER_PROVISIONED_DNS: "yes"
+    workflow: openshift-e2e-gcp
+  timeout: 6h0m0s
 - as: e2e-gcp-secureboot
   optional: true
   run_if_changed: (gcp|google)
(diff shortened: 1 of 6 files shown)


sadasu (approved):
/approve

ci-operator/config/openshift/installer/openshift-installer-release-4.18.yaml
```diff
@@ -475,6 +475,16 @@ tests:
...
   timeout: 6h0m0s
+- always_run: false
+  as: e2e-gcp-user-provisioned-dns
+  optional: true
+  steps:
+    cluster_profile: gcp
+    env:
+      FEATURE_SET: TechPreviewNoUpgrade
+      USER_PROVISIONED_DNS: "yes"
+    workflow: openshift-e2e-gcp
```
patrickdillon, line 486: This is using the existing e2e workflow, but I assumed we would need a new workflow in which we (users) create the public DNS record. This is what we expect users to do and I suspect will be necessary for e2e tests to be able to run.

patrickdillon (commented):
This looks good to me in terms of testing installs, but further work will need to be done to run e2e tests. Not sure if we want to handle that now, or if we want to put tests in place immediately to enable install testing, which is fine with me.

ci-operator/config/openshift/installer/openshift-installer-release-4.18.yaml
sadasu: I am not sure what the issues the e2e tests would encounter. Once the install has completed, there would be a CoreDNS pod running on the control plane that can provided DNS resolution for API and API-Int. I am not sure if the e2e tests need anything more than that. 

At this time using the test to make sure install is successful would be the 1st step. We can evaluate additional setup necessary for e2e tests to pass after we reach that milestone.

patrickdillon:
The [generated config job is failing,](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/58077/pull-ci-openshift-release-master-generated-config/1856671892744704000) looks like you just need to regenerate the jobs

ci-operator/config/openshift/installer/openshift-installer-release-4.18.yaml
patrickdillon: The problem is that when the e2e test pod tries to connect to the cluster (via kubeconfig), it will not be able to resolve the dns for the api. The coreDNS pods can only provide dns for the cluster itself.

PR #58082, opened 2024-10-22
Do not retry Azure installs on infrastructure provisioning failures.

patrickdillon:
Stop retrying installs on Azure. Retrying installs with the same directory worked in Terraform, but was not a supported use case. With the move to CAPI, doing this always fails with:

```
level=error msg=--------------------------------------------------------------------------------
level=error msg=RESPONSE 200: 200 OK
level=error msg=ERROR CODE: BadRequest
level=error msg=--------------------------------------------------------------------------------
level=error msg={
level=error msg=  "startTime": "2024-10-21T02:16:36.2713825+00:00",
level=error msg=  "endTime": "2024-10-21T02:16:41.3495579+00:00",
level=error msg=  "status": "Failed",
level=error msg=  "error": {
level=error msg=    "code": "BadRequest",
level=error msg=    "message": "The source blob https://ciopcwcnsbzs3ea1ajvn47sa.blob.core.windows.net/vhd/rhcos1qcrd.vhd is not accessible."
level=error msg=  },
level=error msg=  "name": "fd3f7997-e2ac-4c24-bf1b-733b6af5c149"
level=error msg=}
level=error msg=--------------------------------------------------------------------------------
level=error
```

We would like to make this work in the future, but let's stop it for now. In fact this particular issue will definitely go away with the move to marketplace images.

### DIFF
--- a/ci-operator/step-registry/ipi/install/install/ipi-install-install-commands.sh
+++ b/ci-operator/step-registry/ipi/install/install/ipi-install-install-commands.sh
@@ -753,6 +753,10 @@ case $JOB_NAME in
     # Do not retry because aws resources can collide when re-using installer assets
     max=1
     ;;
+  *azure)
+    # Do not retry because azure resources always collide when re-using installer assets
+    max=1
+    ;;
   *)
     max=3
     ;;


patrickdillon:
/pj-rehearse pull-ci-openshift-installer-master-e2e-azure-ovn

PR #58097, opened 2024-10-23
remove pact jobs from hac

Katka92:
There is no backed running Pact contract tests.
Removing Pact related jobs from a hac-dev repository.

https://issues.redhat.com/browse/KFLUXUI-115


sahil143 (approved):
/lgtm
/approve

PR #58139, opened 2024-10-24
USHIFT-4334: Merge AWS EC2 CI steps and EC2 logs

vanhalenar:


ci-operator/step-registry/openshift/microshift/infra/aws/ec2/openshift-microshift-infra-aws-ec2-ref.yaml
```diff
@@ -1,9 +1,9 @@
 ref:
   as: openshift-microshift-infra-aws-ec2
   from_image:
-    namespace: ocp
-    name: "4.16"
-    tag: upi-installer
```
pacevedom, line 6: Why change the upi-installer?

vanhalenar: @pmtk and I thought it might not be working because the image used in steps and logs is different, so we're trying this

pacevedom: Just realized, you can not scp if using the `upi-installer`, right?

pmtk: Yeah, the ssh and scp don't work because of the UID tweaks to the image

pmtk:
```
unknown user 1001750000
```

PR #58151, opened 2024-10-24
reconcile memory requests with reality

smg247:
It appears that quite a few of these are very far off, and haven't been updated for multiple years. I didn't do intense digging into any of these, just looked at what the running pod (the largest if there were multiple) was actively using and updated it to match that.

For: https://issues.redhat.com/browse/DPTP-4220


smg247:
I also checked the `git-sync` container's memory usage and it is under `1Gi`, so the issue is the workload itself rather than `git-sync`.

PR #58176, opened 2024-10-25
KONFLUX-191: Remove status checks on konflux triggered jobs

bharath-b-rh:
Removing status checks on konflux triggered jobs because tests vary depending on the changes and all tests on single PR wouldn't happen.

Also when changes are for updating operator images reference in bundle, check causes to execute operator build and again for updating the bundle another PR would be created and it continues in a loop, which is not desired.

### DIFF
--- a/core-services/prow/02_config/openshift/cert-manager-operator-release/_prowconfig.yaml
+++ b/core-services/prow/02_config/openshift/cert-manager-operator-release/_prowconfig.yaml
@@ -12,14 +12,7 @@ branch-protection:
               protect: true
               required_status_checks:
                 contexts:
-                - Red Hat Konflux/cert-manager-operator-1-15-on-pull-request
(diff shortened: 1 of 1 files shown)


core-services/prow/02_config/openshift/cert-manager-operator-release/_prowconfig.yaml
```diff
@@ -10,16 +10,6 @@ branch-protection:
               protect: true
             release-1.15:
               protect: true
-              required_status_checks:
-                contexts:
```
swghosh, line 14: Does using just `- Red Hat Konflux` help here?
I understand the loopy behaviour but wanted to keep some checks so it's easier in the future to automate merge(s), say for the reference update PRs by the konflux bot; without required checks it can be harder and can break unknown things in between multiple PR merges.

PR #58201, opened 2024-10-25
Mark failing e2e job in aro-e2e optional

lranjbar:
Marking the aro-e2e repository job as optional for now. Once we get it working we can make it required again.


hlipsig (approved):
LGTM

PR #58213, opened 2024-10-28
Introduce test frequency f999 for disabled jobs

liangxia:
Introduce test frequency f999 for disabled jobs, and set the cron entry to `8 8 29 2 *` (Feb 29), so that we do not need to change the cron entry for those disabled jobs in the next 3 years ( until year 2028).

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/README.md
+++ b/ci-operator/config/openshift/openshift-tests-private/README.md
@@ -50,6 +50,7 @@ Job frequency is defined by cron according to the test requirements
 - f4: every 4 days
 - f5: every 5 days
 - ...
+- f999: disable the job temporarely
 
 ~~~
 NOTE: We can use below script to generate cron settings against the frequency:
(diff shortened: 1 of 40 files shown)


rioliu-rh:
we'd better to record this spec somewhere, it's useful to make some jobs disabled since they will be triggered by Gangway API proactively e.g. auto release jobs

liangxia:
> we'd better to record this spec somewhere, it's useful to make some jobs disabled since they will be triggered by Gangway API proactively e.g. auto release jobs

The readme has that info, and it's update together in this PR.

ci-operator/config/openshift/openshift-tests-private/README.md
```diff
@@ -50,6 +50,7 @@ Job frequency is defined by cron according to the test requirements
 - f4: every 4 days
 - f5: every 5 days
 - ...
+- f999: disable the job temporarely
```
rioliu-rh, line 53:
```suggestion
- f999: disable the job temporarily
```

liangxia: Good catch. Fix in https://github.com/openshift/release/pull/58217

PR #58231, opened 2024-10-28
[cnv-ci] enable v4.18

Axuba:


Axuba:
cc/ @tiraboschi if you can check if data is correct

Axuba:
Can we merge this? @tiraboschi

Axuba:
@tiraboschi are those test failing normal? 4.10 I understand doesn't matter as it's an old version, but some tests for 4.18 are failing.

cc/ @tal-hason

PR #58254, opened 2024-10-29
core-services/release-controller/_releases/release-ocp-4.12: Use AWS for 4.11-to-4.12

wking:
Recent GCP-side changes break mint-mode GCP installs in accounts that do not enable backupdr APIs (which includes our CI accounts).  We've [patched 4.12 and later to avoid that issue][1], but [4.11 is end-of-life][2] and will not be patched.  So the install-mint-mode-GCP-4.11 needed by the old `gcp-ovn-upgrade-4.12-minor` would never pass.

This commit pivots to use AWS as the blocking 4.11-to-4.12 job.  And instead we only test patch-level 4.12-to-4.12 GCP updates, which will start passing once one nightly with [OCPBUGS-43872][1] has been manually accepted.

[1]: https://issues.redhat.com/browse/OCPBUGS-43872
[2]: https://access.redhat.com/support/policy/updates/openshift#dates

### DIFF
--- a/core-services/release-controller/_releases/priv/release-ocp-4.12.json
+++ b/core-services/release-controller/_releases/priv/release-ocp-4.12.json
@@ -102,13 +102,14 @@
                 "name": "periodic-ci-openshift-release-master-ci-4.12-e2e-aws-sdn-techpreview-serial-priv"
             }
         },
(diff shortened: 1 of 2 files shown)


mrunalp (approved):
/lgtm

PR #58301, opened 2024-10-30
Optimize tag filter for disconnected upgrade

liangxia:
Set the tag `UPGRADE_CHECK/PRE_RUN_TAGS` in the chain `openshift-upgrade-qe-test-disconnected`, so that we do not need to set it in every disconnected upgrade jobs.


liangxia:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.17-arm64-nightly-4.17-upgrade-from-stable-4.17-aws-ipi-disc-priv-sts-basecap-none-f60

PR #58312, opened 2024-10-30
Update of the OCP versions used in ci jobs for notebooks and kubeflow repos

jstourac:
This updates OCP versions used in both upstream and downstream repositories for kubeflow and notebooks codebase.

Bumps all jobs we care to OCP 4.14 images at least - where newer image was used, I kept it that way. Also, I added an explicit definition to downstream notebooks to have this unified with the upstream.

I chosed the OCP 4.14 as it seems to be a decently old release that will be supported for couple of months more [1]. OCP 4.13 will get out of support in 3 weeks.

* [1] https://access.redhat.com/support/policy/updates/openshift#dates

### DIFF
--- a/ci-operator/config/opendatahub-io/kubeflow/opendatahub-io-kubeflow-main.yaml
+++ b/ci-operator/config/opendatahub-io/kubeflow/opendatahub-io-kubeflow-main.yaml
@@ -26,7 +26,7 @@ releases:
     release:
       architecture: amd64
       channel: fast
-      version: "4.14"
+      version: "4.15"
 resources:
   '*':
     requests:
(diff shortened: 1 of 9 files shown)


atheo89:
Thanks Jan for checking on this. 
/lgtm
Let's see if that maybe fix also the e2e tests

jiridanek:
/hold
let's try one more thing first

jstourac:
Thank you @atheo89 for the review. Per the discussion on Slack regarding the issues [1,2], I added one commit so that we use OCP4.15 at least, which shouldn't be affected. Let's see how it will go with this now. If it is good, I can then rearrange commits and commit message so it make sense.

* [1] https://issues.redhat.com/browse/OHSS-38255
* [2] https://issues.redhat.com/browse/OCPBUGS-43821

jstourac:
Okay, looks like this didn't really helped.

jstourac:
/pj-rehearse ci/rehearse/opendatahub-io/kubeflow/main/odh-notebook-controller-e2e

jstourac:
> I'm not sure we want this; we don't really care about ocp version, so 4.15 is just as good as 4.14

Yeah, I agree. Maybe we can ditch the 4.15 upgrade now and just sync all the resource to ocp 4.14 only (except those that are on 4.16 already).

PR #58313, opened 2024-10-30
NO-ISSUE: Use Dockerfile.build for prow job

pastequo:
Prow jobs should be running using the same golang version as the one define in Dockerfile.build

/cc @danmanor


danmanor (approved):

danmanor (changes requested):
add `clone: true`

danmanor (approved):

PR #58322, opened 2024-10-30
Stop relying on Cincinnati channels for upgrade paths

sdodson:
A few times a year we run into scenarios where we latch on either needing a critical fix for an external change like recent GCP changes or due to a regression and it'd be nice to unlatch those by simply accepting or rejecting a release without having to wait on something to be promoted to candidate channels.

This does call into question should we rename all of the "from-stable" to from "from-nightly" or simply "from". Really "from" here means that it's a minor version upgrade rather than specifically from a channel. For now leave it as is, will start a discussion w/ TRT.


sdodson:
/pj-rehearse ack
All jobs have at least completed the initial install which is what this would affect

wking (approved):
/lgtm

PR #58349, opened 2024-10-31
Add 3hops step for cpou

liangxia:
Replace the literal step `set-upgrade-releases` in the configs to a reusable step `cucushift-upgrade-cpou-set-edge-3hops` stored in the step registry.


liangxia:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.18-arm64-nightly-4.18-cpou-upgrade-from-4.15-azure-ipi-fullyprivate-proxy-f28

liangxia:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.17-multi-nightly-4.17-cpou-upgrade-from-4.14-gcp-ipi-ovn-ipsec-arm-mixarch-f28

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.18__multi-nightly-4.18-cpou-upgrade-from-4.15.yaml
```diff
@@ -32,13 +32,13 @@ base_images:
     namespace: ci
     tag: latest
 releases:
-  intermediate416:
```
jianlinliu, line 35: If I remember right, we keep the intermediate release name with version number on purpose, to help us understanding the upgrade path, also align with other similar upgrade ci jobs.  cc @jiajliu

jiajliu: yes, this style is inherited from chain upgrade.

liangxia: Those names are just pointers to the release payloads. Any name should work as long the reference use the same name.
Actually I have the plan the change the chain upgrade after this one. ( The original plan is change chain upgrade together with this one, change the mind later as I want to make changes small, simple and specific ).

ci-operator/step-registry/cucushift/upgrade/cpou/set-edge/4jumps/cucushift-upgrade-cpou-set-edge-4jumps-ref.yaml
```diff
@@ -0,0 +1,16 @@
+ref:
+  as: cucushift-upgrade-cpou-set-edge-4jumps
```
jianlinliu, line 2: it is 3 hops, but not 4 hops, right?

jiajliu, line 2: Is this new step only used for 3 hops cpou jobs(4.14-4.17, 4.15-4.18)? how about 2 hops(for example, 4.16-4.18)?

liangxia: That will be done in another PR, with another step.

liangxia: Updated, `s/4jumps/3hops/g`

jiajliu: hmm, I got u. My concern is that if we will add several steps to support 2 hops/3 hops and N hops in chain upgrade?

liangxia: Currently, I only have plan for 2 hops / 3 hops (because there are multiple tests/profiles on it).   For N hops, e.g, for 4.1->4.18, we don't change it very often, also we only have one or two profile on it. I don't think it can save us too much effort based on what I know currently.  We can talk/decide on the N hops later.

jiajliu: > For N hops, e.g, for 4.1->4.18, we don't change it very often, also we only have one or two profile on it.

Seems not. It has more workloads in chain upgrade because both hops and profiles are more than cpou. take 4.17 for example, there are N paths(4.1/4.2/4.3/.../4.15 to 4.17, which means 15 paths) * M profiles(EOL has only one profile, but others has a maximum of 9 profiles) 

```
$ grep -rh "^- as:" *4.17-upgrade-from-stable-4.14*
- as: ibmcloud-ipi-ovn-ipsec-f28
- as: nutanix-ipi-proxy-fips-f28
- as: vsphere-ipi-ovn-dualstack-privmaryv6-f28
- as: baremetal-upi-ovn-ipv4-f28
- as: aws-ipi-proxy-sts-arm-f28
- as: azure-ipi-basecap-vset-additionalcap-arm-f28
- as: gcp-ipi-sdn-arm-f28
- as: aws-ipi-proxy-sts-arm-f28

$ grep -rh "^- as:" *4.17-upgrade-from-stable-4.13*
- as: baremetalds-ipi-ovn-ipv6-fips-f28
- as: gcp-upi-private-xpn-sdn-f28
- as: vsphere-ipi-ovn-dualstack-f28
- as: nutanix-ipi-proxy-fips-f28
- as: gcp-upi-private-xpn-sdn-f28
- as: baremetal-upi-ovn-ipv4-f28
- as: aws-ipi-proxy-sts-arm-f28
- as: aws-upi-basecap-none-arm-f28
- as: azure-ipi-ovn-ipsec-azurefile-csi-arm-f28
```
> We can talk/decide on the N hops later.

Fine from my side to take it into consideration later.

jiajliu:
lgtm

ci-operator/step-registry/cucushift/upgrade/cpou/set-edge/4jumps/cucushift-upgrade-cpou-set-edge-4jumps-ref.yaml
liangxia: > take 4.17 for example, there are N paths(4.1/4.2/4.3/.../4.15 to 4.17, which means 15 paths) * M profiles(EOL has only one profile, but others has a maximum of 9 profiles)

That's a different thing.  4.1->4.17 and 4.2->4.17 are different, because the hops are different. When the hops are different, a new step is needed (because different hops requires different upgrade edges, thus the dependencies numbers and shell commands to write the upgrade edge are different). That means we will need a new step (for each of 4.1->4.17, 4.2->4.17, 4.3->4.17, ...), but it only applies to one or two profiles, which is not that valuable. For 4.14->4.17 and 4.15->4.17 that have multiple profiles in it, we should be able to apply it to the 3 hops / 2 hops steps.

jiajliu: Here we are talking the whole workload on implementing a dedicated step in chain upgrade, right? If so, both hops/steps number and profiles number will increase the workload. 
Or do u plan to split the chain upgrade into two styles, one is using the step as the implement in this pr, and another is keeping current setting in the yaml file. I'm afraid it's not easily understood for the maintainer.

ci-operator/step-registry/cucushift/upgrade/cpou/set-edge/3hops/cucushift-upgrade-cpou-set-edge-3hops-ref.yaml
```diff
@@ -0,0 +1,16 @@
+ref:
+  as: cucushift-upgrade-cpou-set-edge-3hops
```
jianlinliu, line 2: I think this step also can be reused in chain upgrade, so maybe we can rename it to `cucushift-upgrade-set-edge-3hops`

ci-operator/step-registry/cucushift/upgrade/cpou/set-edge/4jumps/cucushift-upgrade-cpou-set-edge-4jumps-ref.yaml
liangxia: That looks like much more complicated than I have thought. 
Let's focus on 2 hops and 3 hops for cpou for now. We can discuss the chained upgrade in slack for a long term plan.

ci-operator/step-registry/cucushift/upgrade/cpou/set-edge/3hops/cucushift-upgrade-cpou-set-edge-3hops-ref.yaml
liangxia: Relocated/renamed the step.

ci-operator/step-registry/cucushift/upgrade/setedge/3hops/cucushift-upgrade-setedge-3hops-ref.yaml
```diff
@@ -0,0 +1,16 @@
+ref:
+  as: cucushift-upgrade-setedge-3hops
```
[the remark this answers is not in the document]
liangxia: Yes, just following the same pattern in that folder, `setchannel`, `toimage`, `toversion`

PR #58512, opened 2024-11-05
Adding e2e-aws-ovn-ipsec-serial lane back for origin

pperiyasamy:


ci-operator/config/openshift/origin/openshift-origin-master.yaml
```diff
@@ -326,7 +326,6 @@ tests:
 - always_run: false
   as: e2e-aws-ovn-ipsec-serial
   optional: true
-  skip_if_only_changed: ^(?:docs|\.github)/|\.md$|^(?:\.gitignore|OWNERS|OWNERS_ALIASES|PROJECT|LICENSE)$
```
jluhrsen: curious as to what/why this is getting removed? is this what was making it run every time even though it was
'always_run: false' ?

if that's right, can you add that info to the commit message?

bertinatto (commented):
Rehearsal are failing, but a fix is on the works and the idea is to run these jobs manually only.

@pperiyasamy please make sure to follow-up once the job is fixed.

/lgtm

PR #58514, opened 2024-11-05
[SDCICD-1422] Remove osde2e-based conformance test jobs

mmazur:
We're switching to an osde2e-less approach with #58078.

Fixes [SDCICD-1422](https://issues.redhat.com//browse/SDCICD-1422)

### DIFF
--- a/core-services/release-controller/_releases/priv/release-ocp-4.15.json
+++ b/core-services/release-controller/_releases/priv/release-ocp-4.15.json
@@ -464,25 +464,11 @@
                 "name": "periodic-ci-openshift-osde2e-main-nightly-4.15-osd-aws-priv"
             }
         },
-        "openshift-dedicated-aws-conformance": {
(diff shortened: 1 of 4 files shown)


jbpratt:
Need to update another file @mmazur 

```
 [2024-11-05 16:40:30,203] If you are receiving this message, then there is a discrepancy in the release-controller's configuration:
    - If your PR has changes that have removed any of the aforementioned jobs, then you must also remove these jobs from the release-controller's configuration files under:
        core-services/release-controller/_releases, and then run a "make release-controllers" from the root of the repo.
    - If your PR has changes that have added any of the aforementioned jobs, then you must also ensure that these jobs have been defined in their respective location under:
        ci-operator/jobs
    - If you have no idea why you've received this error, then it's most likely do to another commit that introduced the problem.  Please reach out to #forum-ocp-crt in Slack. 
```

PR #58548, opened 2024-11-06
Run e2e-openstack jobs on a cloud that has manila enabled

MaysaMacedo:
Right now we're not running openstack specific tests for manila given the cloud configured on e2e-openstack doesn't have manila endpoint available and the e2e-openstack-manila-csi only runs the csi test suite from openshift. This commit fixes the issue by changing the cloud used on the e2e-openstack job to one that has manila enabled.


stephenfin:
So this is why we're using mecha in multiple places.

/lgtm
/cc @EmilienM

PR #58621, opened 2024-11-07
AGENT-992: installer presubmit job to test control plane with 4 and 5 replicas

rwsu:
The job's name is e2e-agent-5control-ipv4.

Requires: https://github.com/openshift/assisted-service/pull/6917

### DIFF
--- a/ci-operator/config/openshift/installer/openshift-installer-master.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-master.yaml
@@ -1011,6 +1011,30 @@ tests:
     cluster_profile: equinix-ocp-metal
     workflow: agent-e2e-ha-dualstack
   timeout: 6h0m0s
+- as: e2e-agent-4control-ipv4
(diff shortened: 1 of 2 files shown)


ci-operator/config/openshift/installer/openshift-installer-master.yaml
```diff
@@ -1011,6 +1011,17 @@ tests:
     cluster_profile: equinix-ocp-metal
     workflow: agent-e2e-ha-dualstack
   timeout: 6h0m0s
+- as: e2e-agent-5control-ipv4
+  cluster: build05
```
andfasano, line 1015: shouldn't be optional this job?

rwsu: Yes, it should be optional. Thanks for catching it.

rwsu:
I updated the PR to add a job for 4 replicas.

PR #58632, opened 2024-11-08
kni: compute: cleanup old images

shajmakh:
4.11 and below are out of maintenance, remove related old images.


ffromani (commented):
can we clean up only our compute stuff first please? let's defer the ztp cleanup to a followup PR

shajmakh:
This PR touches only the compute images (cnf-tests, NROP images, performanceprofile-operator and dpdk). since the lines were reorganized, this may reflect as if a change was done to ztp related line, but that isn't the case.

ffromani (commented):
/approve
/lgtm

thanks!

PR #58635, opened 2024-11-08
ACM Obs: use integration pipeline and OCP4.17

jacobbaungard:
Currently (and maybe permenently) the `dev` pipeline of stolostron/pipeline doesn't have any snapshots. This causes the e2e-test to fail as it's not able to deploy ACM. This commit sets the pipeline stage to integration, which at least has some snapshots. Long term, we likely need to migrate this test to run on Konflux.

Also, change the OCP version to 4.17.


jacobbaungard:
/pj-rehearse pull-ci-stolostron-multicluster-observability-operator-release-2.13-test-e2e

saswatamcode (approved):

PR #58636, opened 2024-11-08
Serverles: Set size=0 for 4.15 and 4.16 cluster pools

mgencur:
The size specifies the number of hibernated clusters that wait for use. The older clusters will be rarely used from now that we have 4.17 in place.
We need at least 4 clusters for each PR so it makes sense to have the size either `0` or `>=4` for old clusters. Otherwise we don't get any speed up but still need to pay for hibernated clusters.
We keep the original `maxSize` so that custers can be started on demand up to `maxSize` clusters.


ReToCode (approved):
/lgtm
/approve

PR #58643, opened 2024-11-08
IBMCloud: Wait for Resource Group creation.

cjschaef:
Creation of Resource Groups may take a few minutes, before it is ready and can be referenced. Wait up to 5 minutes for those Resource Groups to show up before moving on.


MayXuQQ:
@cjschaef I added the resource group checking in the latest step, such as ci-operator/step-registry/ibmcloud/provision/vpc/ibmcloud-provision-vpc-commands.sh, still has the  "No resource group found" issue.

PR #58658, opened 2024-11-09
PODAUTO-228: CMA: explicitly tag test suite into pipeline

jkyros:
It seems to take awhile to get the promoted test image in the operator test when we update the operand, I'm suspecting it's because we aren't explicitly tagging it into the pipeline here. 

If we don't need this, I'll close it, but I figured it was worth a try while we were waiting.


jkyros:
/pj-rehearse pull-ci-openshift-custom-metrics-autoscaler-operator-main-cma-e2e-aws-ovn

PR #58662, opened 2024-11-10
CNF-14676:numaresources-operator: fixing image mirroring

Tal-or:
The `name` in the `promotion` stanza is the name of the repository under the QCI registry.

The `images[].to` field represents the `tag`.

So if we want to mirror the  image of the `numaresources-operator-bundle` we should mirror the `registry.ci.openshift.org/ocp-kni/numaresources-operator:numaresources-operator-bundle` image.


ffromani (commented):
/lgtm

thanks for pinpointing the problem and for the explanation

PR #58703, opened 2024-11-12
NO-JIRA: Mark e2e-aws-ovn-builds as optional in oc

ardaguclu:
`e2e-aws-ovn-builds` job is nearly always permafailing due to various reasons and usually blocks the oc. For example, this PR https://github.com/openshift/oc/pull/1910 fixes the unit test that is blocking all the other PRs but we can't merge it, since e2e-aws-ovn-builds is permafailing. If there is any change aside from unit tests, owner of the change should additionally check the failures in this job as of now it will be optional.


wking (approved):
I'm not an approver for this repo, but it already has approved, so:

/lgtm

to help unblock openshift/oc#1910 and other pull requests.  Do we know of any trackers for firming up the test's reliability, so we can eventually make this blocking again?

PR #58705, opened 2024-11-12
Merge request for configuring the cases that are marked as not_pr not to run in e2e tests(Insights Operator)

BaiyangZhou:
Modify the e2e tests for Insights Operator not to run the cases that are marked as not_pr

### DIFF
--- a/ci-operator/config/openshift/insights-operator/openshift-insights-operator-release-4.14.yaml
+++ b/ci-operator/config/openshift/insights-operator/openshift-insights-operator-release-4.14.yaml
@@ -89,7 +89,7 @@ tests:
     - as: test
       cli: latest
       commands: |
-        pytest.sh -k "not tp_only and not UI and not slow and not not_pr" --junitxml=$(pwd)/test-report.xml --rp_enabled --rp_name=Insights_operator_e2e_tests_pr_checks_4.14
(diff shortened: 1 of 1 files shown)


ci-operator/config/openshift/insights-operator/openshift-insights-operator-master.yaml
```diff
@@ -93,7 +93,7 @@ tests:
     - as: test
       cli: latest
       commands: |
-        pytest.sh -k "not tp_only and not UI and not slow" --junitxml=$(pwd)/test-report.xml --rp_enabled --rp_name=Insights_operator_e2e_tests_pr_checks
+        pytest.sh -k "not tp_only and not UI and not slow and not not_pr" --junitxml=$(pwd)/test-report.xml --rp_enabled --rp_name=Insights_operator_e2e_tests_pr_checks
```
tremes, line 96: nit: Personally, I always find the double negation a bit confusing. We should have probably called the decorator as `pr` or `pr_only`. :thinking:

BaiyangZhou: These 2 labels are also a little bit confusing because we only mark the cases that shouldn't run PR, would it be better as `skip_on_pr`? ^^

BaiyangZhou: I'm afraid these 2 labels are also a little bit confusing because we just marked the cases that shouldn't run in the PR instead the cases that should run. Would it be better as `skip_on_pr`?^^

ncaak (approved):
/lgtm

PR #58715, opened 2024-11-12
Bumping kmm jobs to ocp 4.17

TomerNewman:
Bumping KMM jobs in CI to use ocp 4.17.

/cc @ybettan @yevgeny-shnaidman

### DIFF
--- a/ci-operator/config/rh-ecosystem-edge/kernel-module-management/rh-ecosystem-edge-kernel-module-management-main.yaml
+++ b/ci-operator/config/rh-ecosystem-edge/kernel-module-management/rh-ecosystem-edge-kernel-module-management-main.yaml
@@ -94,7 +94,7 @@ releases:
   latest:
     release:
       channel: fast
-      version: "4.15"
+      version: "4.17"
 resources:
   '*':
     limits:
@@ -174,7 +174,7 @@ tests:
     owner: rh-ecosystem-edge
     product: ocp
     timeout: 1h0m0s
-    version: "4.15"
+    version: "4.17"
   steps:
     test:
     - as: install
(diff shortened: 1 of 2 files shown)


ci-operator/config/rh-ecosystem-edge/kernel-module-management/rh-ecosystem-edge-kernel-module-management-main.yaml
```diff
@@ -174,7 +174,7 @@ tests:
     owner: rh-ecosystem-edge
     product: ocp
     timeout: 1h0m0s
-    version: "4.15"
+    version: "4.16"
```
ybettan, line 177: Why are we testing this one with 4.16 instead of 4.17? I don't think we even have a pool for 4.16.

TomerNewman: you are right, Im trying to isolate the issue because it looks like the problem is not with the busybox image

ybettan:
/hold

I have a comment.

ci-operator/config/rh-ecosystem-edge/kernel-module-management/rh-ecosystem-edge-kernel-module-management-release-2.2.yaml
```diff
@@ -94,7 +94,7 @@ releases:
   latest:
     release:
       channel: fast
-      version: "4.15"
+      version: "4.17"
```
ybettan, line 97: If we bump the release version in 2.2 we should do the same in the `main` branch.

PR #58747, opened 2024-11-12
 PRE-BRANCHING: Populate generated 4.19 release-gating jobs

hector-vido:


afcollins:
Do we need to include all of the tests under `openshift-qe/ocp-qe-perfscale-ci` in this PR?

Or do we only want those payload-informing jobs? We only have one: `payload-control-plane-6nodes`

One reason to not copy all of them now: We need to shift the cron schedules so they don't overlap. Simply copy-pasting the files will leave jobs running at the same time.

PR #58769, opened 2024-11-13
Add SAST scanning to art-config

joepvd:


joepvd:
rehearsal failed, because it could not clone from the private repo. Gonna tag for merge, maybe it works better when it is really in? Might need a follow-up PR. 

/pj-rehearse ack

PR #58799, opened 2024-11-13
increase 4h->5h for aws-ovn-upgrade in ovnk

jluhrsen:
it's taking aprox 50m to get from the start of the job through the build process before starting the actual test work. The periodic job that does not need to build takes ~3h to complete so using the default 4h timeout here is too close and most jobs are timing out. this increases it to 5h


trozet (approved):
/lgtm

PR #58862, opened 2024-11-15
kni: nrop: operator-sdk 1.31-> 1.36

shajmakh:
Consume operator-sdk version that's similar to what's used in the project repo.


shajmakh:
bad idea, tag not found on quay

PR #6991, opened 2020-02-03
 Revert "label-sync: allow 47h runtime" 

petr-muller:
Reverts: https://github.com/openshift/release/pull/6990

Should merge once `periodic-label-sync` works through its backlog and succeeds, expected to happen after 2 more runs, so we should inspect the job on Friday 2020-02-07 and if all went well, merge this revert.

/hold
/cc @openshift/openshift-team-developer-productivity-test-platform

PR #58894, opened 2024-11-17
NO-ISSUE: capi disconnected fails due to timeout

eifrach:


eifrach:
related to fix  [#6991](https://github.com/openshift/assisted-service/pull/6991)

PR #58921, opened 2024-11-18
Mirror centos images to ocp namespace

joepvd:
Currently, ART is mirroring the centos base images to the CI cluster. This works, but our config is based on OCP version, and this is global for all versions. Think the sync better lives in the release repository.

I confirmed that e.g. registry.ci.openshift.org/ci/yq:4.30.5 is just as multi arch as quay.io/multi-arch/yq:4.30.5.


Prucek:
@joepvd we already mirror those images: 
https://github.com/openshift/release/blob/master/core-services/image-mirroring/_config.yaml#L760-L761

PR #58948, opened 2024-11-18
Add timeout workaround to CI

vrindle:
This PR introduces a timeout workaround. When a pull request (PR) with a particular SHA has already been run but the previous job timed out, the queue manager now returns a UUID immediately, allowing the request to be resolved without re-queuing. If the SHA has not been processed before, the request is placed in the queue as usual.


ci-operator/step-registry/dpu-operator/e2e-tests-workflow/tests/dpu-operator-e2e-tests-workflow-tests-commands.sh
```diff
@@ -31,20 +40,31 @@ check_pull_number() {
...
-if [ $? -ne 0 ]; then
-    echoerr "Error: Failed to trigger build."
-    exit 1
-fi
+#Check timeout wa
+timed_out_response=$(check_timeout_status)
+
+check_timed_out_return_code=$(echo "$timed_out_response" | jq -r '.return_code')
+
+if [ "$check_timed_out_return_code" -eq 400 ]; then
```
SalDaniele, line 49: Can we remove the nesting? It makes it harder to tell exactly what we are doing here? What does a timeout return code of 400 mean? A comment would be helpful

PR #58962, opened 2024-11-19
[OCPERT-18]Change platform value for vsphere-ipi-remote-worker

ming1013:
When run workflow-launch cucushift-installer-rehearse-vsphere-ipi-remote-worker via Cluster Bot, it always got error:
/bin/bash: line 15: vsphere_extra_portgroup_1: unbound variable
Try to change the platform to fix this.


ming1013:
cc @liangxia @WenXinWei @sgaoshang to review

PR #58992, opened 2024-11-19
OpenStack: drop 4.13 periodics

mandre:
As the release entered EOL recently, we no longer need to test it.


mandre:
> /lgtm But don't we set them to be once-per-year instead usually?

I don't think we're doing it anymore, the configs for the previous versions were all removed (at least for openstack)

PR #59080, opened 2024-11-21
allow ci-operator to manages images accross all namespaces in the build clusters

droslean:
/cc @deepsm007 @openshift/test-platform


deepsm007:
/test all

The test seems to be stuck for few hours now

PR #59126, opened 2024-11-22
WIP: Improve results and debuggability for quay-e2e job

dgoodwin:
- **Fail quay e2e test step when e2e tests fail**
- **Enable resource watch observer on quay-quay-tests-master-ocp-418-quay-quay-e2e-tests-quay313-ocp418-lp-interop**

### DIFF
--- a/ci-operator/config/quay/quay-tests/quay-quay-tests-master__ocp-418-quay.yaml
+++ b/ci-operator/config/quay/quay-tests/quay-quay-tests-master__ocp-418-quay.yaml
@@ -31,9 +31,6 @@ tests:
   cron: 0 0 11 6 *
   steps:
     cluster_profile: aws-cspi-qe
-    observers:
-      enable:
-      - observers-resource-watch
     env:
       BASE_DOMAIN: cspilp.interop.ccitredhat.com
       COMPUTE_NODE_TYPE: m5.4xlarge
@@ -55,6 +52,9 @@ tests:
       QUAY_OPERATOR_CHANNEL: stable-3.13
       USER_TAGS: |
         scenario quay
+    observers:
+      enable:
+      - observers-resource-watch
     test:
     - ref: quay-tests-deploy-quay-aws-s3
     - ref: quay-tests-test-quay-e2e


dgoodwin:
/pj-rehearse periodic-ci-quay-quay-tests-master-ocp-418-quay-quay-e2e-tests-quay313-ocp418-lp-interop

dgoodwin:
This PR demonstrates a couple concepts I wanted to share for the interop QE jobs. 

Many seem to have test steps that cannot fail (return non-zero), this is a critical problem that needs to be addressed for all tooling right from prow on down to sippy. None of the tooling can properly understand the job until this is resolved.

The second is something I think could be a dramatic improvement for investigating and categorizing failures as either infrastructure or a product or test bug. I started digging into some job failures and found that some of the tools I would normally use are not present because these tests are not run through origin. (big future concept around that mentioned below) 

But we do have an interesting tool called the resource watch observer that runs origin's "monitortests" as well. This enables a long running process throughout the life of the job, including product installation. It watches for interesting things happening in the cluster, monitors for network disruption to a number of different network stacks in the cluster (tested by requests coming from the CI build farm), and creates a git repo with commits that represent many of the most important kube api objects in the cluster as they change, and when.

Enabling this observer is demonstrated and is quite simple. Then you end up with an artifact directory like this: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_release/59126/rehearse-59126-periodic-ci-quay-quay-tests-master-ocp-418-quay-quay-e2e-tests-quay313-ocp418-lp-interop/1860029073342861312/artifacts/quay-e2e-tests-quay313-ocp418-lp-interop/observers-resource-watch/artifacts/



The tar is the interesting part, extracting this you will find:
  - a number of html charts showing a timeline of those interesting things happening, this is incredibly powerful for determine what was going on in the cluster at the time a failure occurred. covers many things, alerts, disruption, kube events, operator state, node reboots, etc.
  - the git repo/ directory with the history of key resources in the cluster as they change over time

Both of these I think would be very powerful for interop jobs. 

There is a big project underway for components and layered products to be able to define their tests in their own repos, but be run via origin. (enhancement https://github.com/openshift/enhancements/pull/1676) This would require go tests, but will eventually allow for dramatically better tooling to be used consistently. This is a ways out and I gather many of these jobs are not using go tests at this time, but if any one product might be interested in being an early adopter we could get those discussions going sooner.

PR #59180, opened 2024-11-25
Add dptp members to `core-services/image-mirroring/openshift/OWNERS`

danilo-gemoli:
As per title.

### DIFF
--- a/core-services/image-mirroring/openshift/OWNERS
+++ b/core-services/image-mirroring/openshift/OWNERS
@@ -2,7 +2,3 @@ approvers:
 - mrunalp
 - jupierce
 - sdodson
-- jmguzik
-
-options:
-  no_parent_owners: true


Prucek (approved):

smg247:
I think the entire point of the no-parent-owners config is to keep us from approving these PRs. This may no longer be desired, but if we want this, we may as well just remove that line.

danilo-gemoli:
If we allow parents and add someone other than `dptp` in one of these:
- `core-services/image-mirroring/OWNERS`
- `core-services/OWNERS`  

wouldn't he be allowed to approve PRs in `core-services/image-mirroring/openshift/*` too?

smg247:
> If we allow parents and add someone other than `dptp` in one of these:
> 
> * `core-services/image-mirroring/OWNERS`
> * `core-services/OWNERS`
> 
> wouldn't he be allowed to approve PRs in `core-services/image-mirroring/openshift/*` too?

Yes, but I don't see any reason why we would do that. If we did do that, it would make sense that they would also have rights in `core-services/image-mirroring/openshift`.

danilo-gemoli:
updated

smg247:
Now there is no reason to add `dptp` at all because we are already root level approvers.

PR #59282, opened 2024-11-27
Revert "[DTK] Removing old nightly jobs."

ybettan:
Reverts openshift/release#59219

---

4.12 is still supported and although 4.13 is EOL it still gets backport fixes in their way to 4.12.


ybettan:
Need rebase.
Closing and reverting again.

PR #59286, opened 2024-11-27
disable the retester for openshift/dpu-operator repo

droslean:
/cc @bn222


bn222:
/lgtm 

weird unexpected behavior observed in our repo

PR #59298, opened 2024-11-27
Add job for arm64 on crio

xw19:
Trying to fix https://github.com/cri-o/cri-o/issues/8714


xw19:
@saschagrunert  Can you please help? Added a new job as per instruction

saschagrunert:
I guess we need an vm image first.

PR #59299, opened 2024-11-27
PodAutoscaling: Add skip_if_only_changed to 4.20 also, fix regex across all

joelsmith:
Also, don't skip the verify history job

### DIFF
--- a/ci-operator/config/openshift/kedacore-keda/openshift-kedacore-keda-release-4.20.yaml
+++ b/ci-operator/config/openshift/kedacore-keda/openshift-kedacore-keda-release-4.20.yaml
@@ -47,10 +47,12 @@ tests:
   commands: env && GOFLAGS="" make fmt
   container:
     from: src
+  skip_if_only_changed: ^docs/|^\.github|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|OWNERS_ALIASES|PROJECT|LICENSE)$
 - as: govet
   commands: env && GOFLAGS="" make vet
   container:
     from: src
+  skip_if_only_changed: ^docs/|^\.github|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|OWNERS_ALIASES|PROJECT|LICENSE)$
 - as: cma-verify-history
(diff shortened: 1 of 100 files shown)


ci-operator/config/openshift/kedacore-keda/openshift-kedacore-keda-release-4.20.yaml
```diff
@@ -60,6 +62,7 @@ tests:
     clientset-generate
   container:
     from: src
+  skip_if_only_changed: ^docs/|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|PROJECT|LICENSE)$
 - as: security
   optional: true
   skip_if_only_changed: \.md$|^\.github|^(?:.*/)?(?:\.gitignore|OWNERS|OWNERS_ALIASES|PROJECT|LICENSE)$
```
maxcao13, line 68: Seems like there already is an existing `skip_if_only_changed` for other tests that include a bit more. Should we just use that instead?
```yaml
- as: keda-e2e-aws-ovn
  skip_if_only_changed: \.md$|^\.github|^(?:.*/)?(?:\.gitignore|OWNERS|OWNERS_ALIASES|PROJECT|LICENSE)$
  steps:
```

joelsmith: I don't think it would hurt to add the .github dir since the KEDA repos use that, but I don't know that it will make a difference since the only time .github will be changing is on an upstream rebase PR which will usually (always?) have other changes as well. I think we should probalby keep docs. I don't think that any of our projects are using OWNER_ALIASES, so I think we're fine to leave that off.

joelsmith: Okay, I added .github and OWNER_ALIASES. It doesn't hurt anything to ignore stuff that doesn't exist. Someday if we start using them, we'll already have them ignored.

maxcao13 (approved):
/lgtm

Thanks for doing that!

PR #59331, opened 2024-11-28
HCP/openstack: remove RHCOS_IMAGE_NAME from main, 4.20 and 4.19 jobs

EmilienM:
We don't need to set this variable anymore after 4.18 because the
it has a safe default in the step registry.

Also, moving forward this variable won't be needed anymore because we
are working on [OSASINFRA-3492](https://issues.redhat.com//browse/OSASINFRA-3492) which will take care of the RHCOS image.


mandre (commented):
/lgtm
/approve

EmilienM:
/pj-rehearse ack
No need to trigger jobs here.

PR #59338, opened 2024-11-28
Update ODH operator to Go 1.22

grdryn:
Related: https://github.com/opendatahub-io/opendatahub-operator/pull/1424

### DIFF
--- a/ci-operator/config/opendatahub-io/opendatahub-operator/opendatahub-io-opendatahub-operator-main.yaml
+++ b/ci-operator/config/opendatahub-io/opendatahub-operator/opendatahub-io-opendatahub-operator-main.yaml
@@ -7,7 +7,7 @@ build_root:
   image_stream_tag:
     name: release
     namespace: openshift
-    tag: golang-1.21
+    tag: golang-1.22
 images:
 - context_dir: .
   dockerfile_path: Dockerfiles/Dockerfile


ci-operator/config/opendatahub-io/opendatahub-operator/opendatahub-io-opendatahub-operator-feature-operator-refactor.yaml
```diff
@@ -7,7 +7,7 @@ build_root:
   image_stream_tag:
     name: release
     namespace: openshift
-    tag: golang-1.21
+    tag: golang-1.22
```
zdtsw, line 10: if the original PR is only targeting incubation, we can drop this one. soon feature-oprator-refactor will be killed

ci-operator/config/opendatahub-io/opendatahub-operator/opendatahub-io-opendatahub-operator-main.yaml
```diff
@@ -7,7 +7,7 @@ build_root:
   image_stream_tag:
     name: release
     namespace: openshift
-    tag: golang-1.21
+    tag: golang-1.22
```
zdtsw, line 10: and we are not using main for long time. so better keep it sync with the code on that branch

ci-operator/config/opendatahub-io/opendatahub-operator/opendatahub-io-opendatahub-operator-rhoai.yaml
```diff
@@ -7,7 +7,7 @@ build_root:
   image_stream_tag:
     name: release
     namespace: openshift
-    tag: golang-1.21
+    tag: golang-1.22
```
zdtsw, line 10: this will be needed, but need wait till ODH PR land to rhoai branch

grdryn: @zdtsw I'm confused by the chicken-egg-ness of this: I thought that this would be needed on any branch _before_ those branches would be able to start using the newer go, or else the tests wouldn't be able to pass?

zdtsw: what i meant is:
we could have new commits to rhoai branch before we get the 1.22 PR to rhoai.
(if we are going step by step: merge 1.22 to incubation, test it,  days after.............., sync the PR to rhoai)
during these "days between", e2e from this PR uses 1.22 to build our code, in theory, they should work

grdryn: Ah right, understood, thanks :+1: I've removed it from here for now then

grdryn:
@zdtsw @VaishnaviHire I've changed this to target the main branch now, since I understand that that's the new development branch. I'll replace https://github.com/opendatahub-io/opendatahub-operator/pull/1400 with an equivalent targeting main also, but I believe this would be the first step.

zdtsw:
grrrrrrrrrrrr. so i suddenly have the "permission to approve"?

PR #59280, opened 2024-11-27
correct capacity check in upgrade cases

jianlinliu:
Follow up https://github.com/openshift/release/pull/58914#issuecomment-2503252220

For those capacities which are not existing in source version and introduced for the 1st time in target version, after upgrade, they are supposed to be disabled by default.

PR #59355, opened 2024-11-29
enhance the way of getting new capacities list in upgrade cases

jianlinliu:
Since https://github.com/openshift/release/pull/59280

```
# extract_potential_new_caps "${4.15-manifest-dir}" "${4.18-manifest-dir}"
Build Console Insights MachineAPI marketplace OperatorLifecycleManagerV1 Storage
```
The detected new cap list is not accurate, actually since 4.15, only `OperatorLifecycleManagerV1` is a newly introduced fresh operator capacity, which is not existing on the source version, only existing on the target version. 

Fortunately the new caps calculated for the 1st time between 4.15 and 4.18 is `CloudControllerManager Ingress OperatorLifecycleManagerV1`, so when strip the list using the output of `extract_potential_new_caps`, only strip `OperatorLifecycleManagerV1`, that happened to help us get a correct implicit and enabled cap list. 

Think about that, if the upgrade happened between other versions, e.g: 4.13 -> 4.18, the new caps list calculated for the 1st time is `MachineAPI Build DeploymentConfig ImageRegistry OperatorLifecycleManager CloudCredential CloudControllerManager Ingress OperatorLifecycleManagerV1`, when strip the list using the output of `extract_potential_new_caps`, `Build` would be removed together with `OperatorLifecycleManagerV1`, that means, `Build` would not in the implicit and enabled cap list. But actually the source version always install "Build" operator. That would lead the testing exit as failure, but that is wrong. 

The root cause is `extract_potential_new_caps` did not get the exact caps that newly introduced operator later than source version. 

To improve it, introduced `get_transfered_new_caps` function, in the function, it is going to do 2 checks:
1. check if there is the same mainfest files for the specific new cap in the source payload, if yes, then it will not be identified as a known capacity for source version
2. if 1 does not work, continue to try to grep related resource for the specific new cap in source manifest, if yes, then it will not be identified as a known capacity for source version
3. otherwise, the capacity will be identified as a total new capacity that source version does not know.

The above logic is a bit complicated, and not reliable enough, especially some new special operator get introduced in the future.

So as final, we decide to introduce a dedicated array to record these newly introduced capacity.
E.g:
new_caps_version["OperatorLifecycleManagerV1"]="4.18"
new_caps_version["xxx"]="4.19"
new_caps_version["yyy"]="4.20"
Then compare the version number with the source version to decide which capacity should be removed from transfered_new_caps list.

### DIFF
--- a/ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
+++ b/ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
@@ -205,32 +205,81 @@ function extract_migrated_caps() {
     done
 }
 
-# Get the capabilities which are introdcued for the 1st time in the target payload
-# By comparing two payloads at manifest level
-function extract_potential_new_caps() {
-    local source_dir="$1"
-    local target_dir="$2"
-    local target_cap_files target_file filename source_file target_cap
-    local candidate_new_caps=""
+# not used yet, keep it here for future usage
+function lowerVersion() {
+    local version="$1" target_version="$2"
 
-    # Check for missing environment variables
-    if [ -z "$source_dir" ] || [ -z "$target_dir" ]; then
-        echo "ERROR: Missing required arguments"
-        return 1
+    local ret=0
+    if [[ "$(printf '%s\n' "${target_version}" "${version}" | sort --version-sort | head -n1)" == "${target_version}" ]]; then
+        ret=1
     fi
+    return $ret
+}
 
-    target_cap_files=$(grep -rl "capability.openshift.io/name" ${target_dir})
-    for target_file in ${target_cap_files}; do
-        filename=$(basename "${target_file}")
-        source_file="$source_dir/$filename"
+# Get the new caps transfered from an existing operator on source verion
+# by removing those capacities that are GAed as optional ones
+function get_transfered_new_caps() {
+    local candidate_new_caps="$1" source_ver="$2"
+    local cap tmp_candidate_new_caps
 
-        if [ ! -f "$source_file" ]; then
-            target_cap=$(grep 'capability.openshift.io/name:' "$target_file" | cut -d ':' -f2 | tr -d '"' | sort | uniq )
-            candidate_new_caps="${candidate_new_caps} ${target_cap}"
+    tmp_candidate_new_caps="${candidate_new_caps}"
+    source_ver=$(echo "$source_ver" | cut -d. -f1,2)
+    for cap in ${candidate_new_caps}; do
+        #shellcheck disable=SC2076
+        if [[ " ${!new_caps_version[*]} " =~ " ${cap} " ]]; then
+            echo "${cap} is GAed as optional capacity later than source version - ${source_ver}"
+            tmp_candidate_new_caps="${tmp_candidate_new_caps/$cap}"
         fi
     done
-    echo "${candidate_new_caps}" | xargs -n1 | sort -u | xargs
+    transfered_new_caps=$(echo "${tmp_candidate_new_caps}" | xargs -n1 | sort -u | xargs)
 }
+# The function is complicated, and might be unreliable if some new change in 
+# manifest level, so disable it for now.
+#function get_transfered_new_caps() {
+#    local source_dir="$1" target_dir="$2" candidate_new_caps="$3" found="no"
+#    local cap target_cap_files target_file filename source_file target_cap crd_files cr_kind tmp_candidate_new_caps
+#
+#    tmp_candidate_new_caps="${candidate_new_caps}"
+#    for cap in ${candidate_new_caps}; do
+#        target_cap_files=$(grep -rl "capability.openshift.io/name: ${cap}$" ${target_dir} || true)
+#        if [[ -z "$target_cap_files" ]]; then
+#            echo "Did not find out $cap capacity manifest files in target payload"
+#            continue
+#        fi
+#        # most of capacities can be identified if it is new or not on source version by comparing file name if the capacity manifests are not changed a lot
+#        found="no"
+#        for target_file in ${target_cap_files}; do
+#            filename=$(basename "${target_file}")
+#            source_file="$source_dir/$filename"
+#            if [[ -f "$source_file" ]] && [[ -z "$(grep 'release.openshift.io/feature-set:' $source_file | grep -E 'CustomNoUpgrade|TechPreviewNoUpgrade|DevPreviewNoUpgrade')" ]]; then
+#                echo "Found related manifest - $filename in source payload manifests for $cap capacity, it is known (not TP) for source version"
+#                found="yes"
+#                break
+#            fi
+#        done
+#        # but if capacity manifest is changed between source and target version, try to search CR type for the capacity
+#        # e.g: "Build", it isn't operator type, related manifests defining "Build" are totally changed from 4.13 to 4.18
+#        # Another special capacity "DeploymentConfig", there is no any relalted manifests defining "DeploymentConfig" in both 4.13 and 4.18
+#        crd_files=""
+#        [[ "$found" == "no" ]] && crd_files=$(grep -l "^kind: CustomResourceDefinition$" ${target_cap_files} || true)
+#        for target_file in ${crd_files}; do
+#            cr_kind=$(yq-go r "${target_file}" "spec.names.kind")
+#            [[ -n "$cr_kind" ]] && cr_kind_files=$(grep -rl "kind: ${cr_kind}$" "${source_dir}" || true)
+#            if [[ -n "$cr_kind_files" ]] && [[ -z "$(grep -E 'CustomNoUpgrade|TechPreviewNoUpgrade|DevPreviewNoUpgrade' $cr_kind_files)" ]]; then
+#                echo "Found related ${cr_kind} resource in source payload manifests for $cap capacity, it is known (not TP) for source version"
+#                found="yes"
+#                break
+#            fi
+#        done
+#        # remove total unknown capacity, the source cluster never install it, even upgraded, there would be no such resource.
+#        # some exception, some fresh capacities introduced later than source version maybe installed automatically as some enabled capacity's dependency during upgrade 
+#        if [[ "${found}" == "no" ]]; then
+#            echo "${cap} is totally new for source version..."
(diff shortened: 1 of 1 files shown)


jianlinliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.18-multi-nightly-4.18-upgrade-from-stable-4.17-aws-ipi-sno-etcd-encryption-basecap-none-arm-f28 periodic-ci-openshift-openshift-tests-private-release-4.18-multi-nightly-4.18-upgrade-from-stable-4.15-gcp-ipi-basecap-none-additionalcaps-arm-f28

jianlinliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.18-multi-nightly-4.18-upgrade-from-stable-4.15-gcp-ipi-basecap-none-additionalcaps-arm-f28

jianlinliu:
cc @jiajliu to review.

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
```diff
@@ -205,31 +205,56 @@ function extract_migrated_caps() {
...
+        if [[ -z "$target_cap_files" ]]; then
+            echo "Did not find out $cap capacity manifest files in target payload"
+            continue
+        fi
+        # most of capacities can be identified if it is new or not on source version by comparing file name if the capacity manifests are not changed a lot
+        found="no"
+        for target_file in ${target_cap_files}; do
+            filename=$(basename "${target_file}")
+            source_file="$source_dir/$filename"
+            if [[ -f "$source_file" ]] && [[ -z "$(grep 'release.openshift.io/feature-set:' $source_file | grep -E 'CustomNoUpgrade|TechPreviewNoUpgrade')" ]]; then
```
jiajliu, line 230: shall `DevPreviewNoUpgrade` be excluded too?

```
# grep -rh 'release.openshift.io/feature-set:' manifest/|sort -u
    release.openshift.io/feature-set: CustomNoUpgrade
        release.openshift.io/feature-set: CustomNoUpgrade,TechPreviewNoUpgrade
    release.openshift.io/feature-set: CustomNoUpgrade,TechPreviewNoUpgrade
    release.openshift.io/feature-set: Default
    release.openshift.io/feature-set: DevPreviewNoUpgrade
            release.openshift.io/feature-set: TechPreviewNoUpgrade
        release.openshift.io/feature-set: TechPreviewNoUpgrade
    release.openshift.io/feature-set: TechPreviewNoUpgrade
```

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
```diff
@@ -205,31 +205,56 @@ function extract_migrated_caps() {
...
-        filename=$(basename "${target_file}")
-        source_file="$source_dir/$filename"
-
-        if [ ! -f "$source_file" ]; then
-            target_cap=$(grep 'capability.openshift.io/name:' "$target_file" | cut -d ':' -f2 | tr -d '"' | sort | uniq )
-            candidate_new_caps="${candidate_new_caps} ${target_cap}"
+    for cap in ${candidate_new_caps}; do
+        target_cap_files=$(grep -rl "capability.openshift.io/name: ${cap}$" ${target_dir} || true)
+        if [[ -z "$target_cap_files" ]]; then
+            echo "Did not find out $cap capacity manifest files in target payload"
```
jiajliu, line 222: If the new_cap is not found in the target payload, it should not be expected, right? Should the test exit?

jianlinliu: Take `DeploymentConfig` as example, there is no any capacity manifest files can be found, so introduce this if block to continue.

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
jianlinliu: Done.

jianlinliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.18-multi-nightly-4.18-upgrade-from-stable-4.17-aws-ipi-sno-etcd-encryption-basecap-none-arm-f28

ci-operator/step-registry/cucushift/upgrade/check/capability/cucushift-upgrade-check-capability-commands.sh
```diff
@@ -205,32 +205,80 @@ function extract_migrated_caps() {
...
+    local cap tmp_candidate_new_caps
 
-        if [ ! -f "$source_file" ]; then
-            target_cap=$(grep 'capability.openshift.io/name:' "$target_file" | cut -d ':' -f2 | tr -d '"' | sort | uniq )
-            candidate_new_caps="${candidate_new_caps} ${target_cap}"
+    tmp_candidate_new_caps="${candidate_new_caps}"
+    source_ver=$(echo "$source_ver" | cut -d. -f1,2)
+    for cap in ${candidate_new_caps}; do
+        #shellcheck disable=SC2076
+        if [[ " ${!new_caps_version[*]} " =~ " ${cap} " ]] && lowerVersion "${source_ver}" "${new_caps_version[$cap]}"; then
```
jiajliu, line 228: when will `${source_ver} > ${new_caps_version[$cap]}`?

jianlinliu: Good question. In the current real world, there is no such case.

But I can foresee a case in the future. 
1. Assumed on 4.N version, a new operator is introduced and GAed. Let us call it "A". On the 4.N+1 version, the operator become an optional operator. `new_caps_version["A"]="4.N"`
2. Assumed there is no the condition `lowerVersion "${source_ver}" "${new_caps_version[$cap]}"` in this check.
3. In 4.N -> 4.N+1 upgrade case, `candidate_new_caps` is "A", after the function is executed, `transfered_new_caps` is set to null. It will lead to a failed result (A is installed in the upgraded cluster, but the testing is expecting it is disabled).
4. In 4.N-1 -> 4.N+1 upgrade case, `candidate_new_caps` is "A", after the function is executed, `transfered_new_caps` is set to null. It will lead to a passed result.

In the above point 2, if we have the `lowerVersion` condition.
3.  In 4.N -> 4.N+1 upgrade case, `candidate_new_caps` is "A", after the function is executed, `transfered_new_caps` is set to "A". It will lead to a passed result.
4.  In 4.N-1 -> 4.N+1 upgrade case, `candidate_new_caps` is "A", after the function is executed, `transfered_new_caps` is set to null. It will lead to a passed result.

jiajliu: > Assumed on 4.N version, a new operator is introduced and GAed. Let us call it "A". On the 4.N+1 version, the operator become an optional operator.

In this scenario, "A" does not meet first introduced operational cap, right? So "A" should not be defined in new_caps_version[], but only added into caps_string of get_caps_for_version_capset(). 
Then, in 4.N -> 4.N+1 upgrade case, candidate_new_caps is "A", after the function is executed, transfered_new_caps is still set to "A". It will lead to a passed result.

In 4.N-1 -> 4.N+1 upgrade case, candidate_new_caps is "A", after the function is executed, transfered_new_caps is still set to "A". It will lead to a passed result. "A" will be enabled.

jianlinliu: > In 4.N-1 -> 4.N+1 upgrade case, candidate_new_caps is "A", after the function is executed, transfered_new_caps is still set to "A". It will lead to a passed result. "A" will be enabled.

For 4.N-1, the resources belonging to "A" would not be installed, after the upgrade, there would be no any resources belonging to "A" installed, even the cluster is on 4.N+1 version, then the testing would be failed, because the script is expecting "A" was enabled.

jiajliu: 4.N-1 -> 4.N+1 will go to 4.N first, right? "A" is a new operator but not an optional cap in 4.N, so during 4.N-1-> 4.N, "A" will be installed after upgrade to 4.N by default. And then during 4.N->4.N+1, "A" is still installed.

jianlinliu: Even go to 4.N first, "A" would not be installed automatically, though it is not an optional cap in 4.N.

jiajliu: Hmm, if "A" is a new operator(non optional) in 4.N but not 4.N-1, will "A" be installed or not if setting `BaselineCapabilitySet=None` during installation? I just checked the api definition in [doc](https://github.com/openshift/api/blob/6f735e7109c87826edee3ca02a753071ecc933b9/config/v1/types_cluster_version.go#L297), `BaselineCapabilitySet=None` seems only control optional caps. So I guess "A" is not in the optional caps, then it should be installed, right?

jianlinliu: Maybe you are right, in https://gcsweb-qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/qe-private-deck/logs/periodic-ci-openshift-openshift-tests-private-release-4.12-amd64-stable-4.12-upgrade-from-stable-4.11-aws-c2s-ipi-disc-priv-fips-f60/1863851873665028096/artifacts/aws-c2s-ipi-disc-priv-fips-f60/, `control-plane-machine-set` is introduced in 4.12 for the 1st time, after upgrade to 4.12 from 4.11, the co is installed in the upgraded cluster.

Let me update the script.

jianlinliu: Done, I keep `lowerVersion` function in case of possible demand in the future.

jianlinliu:
Just removed one condition in the last commit to loosen the check, low risk to introduce regression issue, would skip pj-rehearse testing.

/pj-rehearse skip

PR #59383, opened 2024-12-02
[PERFSCALE-3525] Update OVN PerfScale AWS Cluster Profile us-east-1

krishvoor:
Use us-east-1 as choice for region, given it has more AZ.


dceara:
From our testing this seems to work.  FWIW, the change looks good to me.

PR #59386, opened 2024-12-02
OLS-1252 - Separating periodic jobs 

JoaoFula:
from pre-commits so that we can call a separate script.
This is done to stop test_model_response from running in pre-commit testing


tisnik (commented):
nice one!

PR #59416, opened 2024-12-02
Pin sippy e2e presubmit to build10 due to build06 secret problem

dgoodwin:


dgoodwin:
forgot make jobs, pushed again.

PR #59441, opened 2024-12-03
Use Azure for full e2e periodics of hypershift-kubevirt

orenc1:
the AWS-BM approach is failing frequently, because we're reaching the maxPods limit on the single BM worker on that environment. trying to switch to azure with 3 worker nodes capable of running VMs and check if it can be more stable. starting with 4.18 as a PoC.


orenc1:
/pj-rehearse periodic-ci-openshift-hypershift-release-4.18-periodics-e2e-kubevirt-azure-ovn

PR #59444, opened 2024-12-03
Remove last cdc from FIPs POC ClusterPool

lewisdenny:
It was decided to keep FIPs testing D/S, this POC is no longer needed so the cdc can be added to the 4.18 cluster pool.


lewisdenny:
/cc @frenzyfriday @pablintino 

Just need a lgtm if you don't have any concerns.

PR #59099, opened 2024-11-22
CORS-3698: Introduce default configuration step on AWS

yunjiang29:
All of install-config on AWS includes many of optional configurations (see step [ipi-conf-aws](https://github.com/openshift/release/blob/master/ci-operator/step-registry/ipi/conf/aws/ipi-conf-aws-ref.yaml)), like zones, instance types, in some cases, the test requires a _default_ configuration, to cover cases like [OCPBUGS-43453](https://issues.redhat.com/browse/OCPBUGS-43453), hence, `ipi-conf-aws-default` step is introduced, which will generate install-config like:
```
apiVersion: v1
metadata:
  name: ci-op-cxhzw500-2fe85
baseDomain: qe.devcluster.openshift.com
platform:
  aws:
    region: us-east-1
    userTags:
      ...
controlPlane:
  architecture: arm64
  name: master
  replicas: 3
  platform: {}
compute:
- architecture: arm64
  name: worker
  replicas: 3
  platform: {}
```

PR #59488, opened 2024-12-04
CORS-3698 - Add test for AWS default install config

gpei:
As a followup PR for https://github.com/openshift/release/pull/59099 to implement [CORS-3698](https://issues.redhat.com//browse/CORS-3698)

Based on the description in [CORS-3698](https://issues.redhat.com//browse/CORS-3698), `openshift-e2e-test` is not added in the test.
>We should add these for each platform. The tests do not need to run full e2e tests. An install should be sufficient.

### DIFF
--- a/ci-operator/config/openshift/installer/openshift-installer-master.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-master.yaml
@@ -428,6 +428,22 @@ tests:
     - ref: openshift-e2e-test
   timeout: 6h0m0s
 - always_run: false
+  as: e2e-aws-default-config
+  optional: true
+  run_if_changed: aws
+  steps:
+    cluster_profile: aws
+    env:
+      AWS_INSTALL_USE_MINIMAL_PERMISSIONS: "yes"
+    post:
+    - chain: ipi-aws-post
+    pre:
+    - ref: ipi-conf
+    - ref: ipi-conf-aws-default
+    - chain: aws-provision-iam-user-minimal-permission
+    - chain: ipi-install
+  timeout: 6h0m0s
+- always_run: false
   as: e2e-aws-ovn-edge-zones
   optional: true
   run_if_changed: aws
--- a/ci-operator/config/openshift/installer/openshift-installer-release-4.18.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-release-4.18.yaml
@@ -428,6 +428,22 @@ tests:
     - ref: openshift-e2e-test
   timeout: 6h0m0s
 - always_run: false
+  as: e2e-aws-default-config
+  optional: true
+  run_if_changed: aws
+  steps:
+    cluster_profile: aws
+    env:
+      AWS_INSTALL_USE_MINIMAL_PERMISSIONS: "yes"
+    post:
+    - chain: ipi-aws-post
+    pre:
+    - ref: ipi-conf
+    - ref: ipi-conf-aws-default
+    - chain: aws-provision-iam-user-minimal-permission
+    - chain: ipi-install
+  timeout: 6h0m0s
+- always_run: false
   as: e2e-aws-ovn-edge-zones
   optional: true
   run_if_changed: aws
--- a/ci-operator/config/openshift/installer/openshift-installer-release-4.19.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-release-4.19.yaml
@@ -429,6 +429,22 @@ tests:
     - ref: openshift-e2e-test
   timeout: 6h0m0s
 - always_run: false
+  as: e2e-aws-default-config
+  optional: true
+  run_if_changed: aws
+  steps:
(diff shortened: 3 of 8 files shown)


gpei:
/pj-rehearse pull-ci-openshift-installer-master-e2e-aws-default-config

gpei:
cc @yunjiang29 @jianlinliu for the review

ci-operator/config/openshift/installer/openshift-installer-master.yaml
```diff
@@ -427,6 +427,21 @@ tests:
     test:
     - ref: openshift-e2e-test
   timeout: 6h0m0s
+- always_run: false
+  as: e2e-aws-default-config
```
jianlinliu, line 431: Shall we also need to add the job in openshift-installer-release-4.{18,19,20}.yaml ?

jianlinliu: I am not sure if `backport` PR would also trigger the pre-submit jobs, if yes, the jobs are defined in `openshift-installer-release-4.18.yaml. For earlier version, so far, it only supports 4.18, so we do not need to consider that yet.

@r4f4 what is your option here ?

r4f4: I think we can add the 4.18 job in this PR. In theory we wouldn't need the 4.19 one but I always add just in case.

r4f4:
@gpei Thank you for working on this! Seeing the generated `install-config.yaml`, it's pretty minimal:
```
apiVersion: v1
metadata:
  name: ci-op-gjcfnmtc-2f6d9
sshKey: <redacted>
baseDomain: origin-ci-int-aws.dev.rhcloud.com
platform:
  aws:
    region: us-west-1
    userTags:
      expirationDate: 2024-12-04T09:20+00:00
      clusterName: ci-op-gjcfnmtc-2f6d9
controlPlane:
  architecture: amd64
  name: master
  replicas: 3
  platform: {}
compute:
- architecture: amd64
  name: worker
  replicas: 3
  platform: {}
```
But we could also consider removing either the `controlPlane` or `compute` stanza to catch cases where the installer doesn't check for `nil`

ci-operator/config/openshift/installer/openshift-installer-master.yaml
```diff
@@ -427,6 +427,21 @@ tests:
     test:
     - ref: openshift-e2e-test
   timeout: 6h0m0s
+- always_run: false
```
r4f4, line 430: I think we can make this a `run_if_changed: aws` so it always runs for aws jobs.

ci-operator/config/openshift/installer/openshift-installer-master.yaml
gpei: sure, I think I can add the job to openshift-installer-release-4.{18,19,20}.yaml to make sure there's nothing missed.

gpei:
> But we could also consider removing either the `controlPlane` or `compute` stanza to catch cases where the installer doesn't check for `nil`

@r4f4 do you mean something like 
```
controlPlane: {}
```
or just removing the `controlPlane` field from the install-config completely? 

but one concern here is the test is for a **default** configuration just like what `openshift-install create install-config` generates , @jianlinliu @yunjiang29 wdyt?

r4f4:
> > But we could also consider removing either the `controlPlane` or `compute` stanza to catch cases where the installer doesn't check for `nil`
> 
> @r4f4 do you mean something like
> 
> ```
> controlPlane: {}
> ```
> 
> or just removing the `controlPlane` field from the install-config completely?

I was thinking about removing it entirely.

> but one concern here is the test is for a **default** configuration just like what `openshift-install create install-config` generates , @jianlinliu @yunjiang29 wdyt?

Oh, if that's the goal then what you have right now is fine.

jianlinliu:
> one concern here is the test is for a default configuration just like what openshift-install create install-config generates

Yeah, the yaml file created from `openshift-install create install-config` has both controlPlane or compute stanza.

yunjiang29:
> > > But we could also consider removing either the `controlPlane` or `compute` stanza to catch cases where the installer doesn't check for `nil`
> > 
> > 
> > @r4f4 do you mean something like
> > ```
> > controlPlane: {}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > or just removing the `controlPlane` field from the install-config completely?
> 
> I was thinking about removing it entirely.
> 
> > but one concern here is the test is for a **default** configuration just like what `openshift-install create install-config` generates , @jianlinliu @yunjiang29 wdyt?
> 
> Oh, if that's the goal then what you have right now is fine.

The PR looks good to me, it's almost the same as what `create install-config` generated.

For the above concerns, I think we can add a control flag, to allow `ipi-conf-aws-default` to create configs like `controlPlane: {}` if necessary

PR #59517, opened 2024-12-04
Remove HCP on OSP jobs for 4.17

EmilienM:
We don't need them now we have 4.18 jobs and it's still dev preview.
It's not worth keeping both as we document that HyperShift Operator
should be pulled from latest anyway.


yuqi-zhang (approved):
/lgtm


for the MCO jobs

PR #59528, opened 2024-12-04
Update Vsphere02 ownership and common manifests are moved

deepsm007:
This will make sure any build farm wide changes are not applied to vsphere02 cluster. Any required changes needs to go through SPLAT team owners

We have applyconfig sync which will sync any configs stored under this directory, if someone modifies the file on cluster it will be synced back from the configs stored here.

/cc @rvanderp3 @jcpowermac 
/cc @openshift/test-platform


smg247:
Do we still want to keep the `common` folder name in there? I think it might be confusing if we maintain that structure.

deepsm007:
We can remove the folder structure completely and let the team decide what they want for vsphere02

smg247 (approved):

deepsm007:
updated 
`options:
  no_parent_owners: true`

PR #59542, opened 2024-12-05
OCPQE-27389 - Add default configuration step on Azure

gpei:
To generate a standard default install-config.yaml in Prow test.
```
install-config.yaml
-------------------
apiVersion: v1
metadata:
  name: ci-op-wyht7bfn-d342f
sshKey: XXX
baseDomain: qe.azure.devcluster.openshift.com
platform:
  azure:
    baseDomainResourceGroupName: os4-common
    region: centralus
controlPlane:
  architecture: arm64
  name: master
  platform: {}
compute:
- architecture: arm64
  name: worker
  replicas: 3
  platform: {}
```


gpei:
/pj-rehearse periodic-ci-openshift-verification-tests-master-installation-nightly-4.18-azure-ipi-default-mini-perm-f7

ci-operator/step-registry/ipi/conf/azure/default/ipi-conf-azure-default-commands.sh
```diff
@@ -0,0 +1,116 @@
...
+REQUIRED_OCP_VERSION="4.12"
+isOldVersion=true
+if [ -n "${version}" ] && [ "$(printf '%s\n' "${REQUIRED_OCP_VERSION}" "${version}" | sort --version-sort | head -n1)" = "${REQUIRED_OCP_VERSION}" ]; then
+  isOldVersion=false
+fi
+
+PUBLISH=$(yq-go r "${CONFIG}" "publish")
+echo "publish: ${PUBLISH}"
+echo "is Old Version: ${isOldVersion}"
+if [ ${isOldVersion} = true ] || [ -z "${PUBLISH}" ] || [ X"${PUBLISH}" == X"External" ]; then
```
jianlinliu, line 105: I think we drop the if block here, because we want a pretty minimal install-config.yaml in this testing.

ci-operator/step-registry/ipi/conf/azure/default/ipi-conf-azure-default-commands.sh
```diff
@@ -0,0 +1,116 @@
...
+  else
+    echo "${OUTBOUND_TYPE} is not supported yet" || exit 1
+  fi
+fi
+
+# User tags went GA in 4.14. In 4.14+, tag resources with an expiration date to facilitate cleanup.
+if (( minor_version > 13 && major_version == 4 )); then
+  expiration_date=$(date -d '8 hours' --iso=minutes --utc)
+  printf 'Setting user tag expirationDate: %s\n' "${expiration_date}"
+  yq-go write -i "${CONFIG}" "platform.azure.userTags.expiration_date" "${expiration_date}"
```
jianlinliu, line 93: Shall we drop `userTags.expiration_date` setting here to get a pretty minimal install-config.yaml ? or we can skip the version check, directly set `platform.azure.userTags.expiration_date` in install-config.yaml, because the testing is only run in 4.16+

ci-operator/config/openshift/verification-tests/openshift-verification-tests-master__installation-nightly-4.16.yaml
```diff
@@ -311,6 +311,15 @@ tests:
     test:
     - chain: cucushift-installer-check-cluster-health
     workflow: cucushift-installer-rehearse-azure-ipi-generation
+- as: azure-ipi-default-mini-perm-f7
```
jianlinliu, line 314: suggest to use `arm` to run testing to save some cost.

BTW, do we need to run it once per 7 days ?  How about f14 ?

ci-operator/step-registry/ipi/conf/azure/default/ipi-conf-azure-default-commands.sh
gpei: Looks like the `baseDomainResourceGroupName` parameter is generated by default in the install-config, here's one example install-config created manually from the `openshift-install create install-config` command.
```
platform:
  azure:
    baseDomainResourceGroupName: os4-common
    cloudName: AzurePublicCloud
    outboundType: Loadbalancer
    region: eastus2
```

@jinyunma do you know if we can run a normal IPI installation without specifying the `baseDomainResourceGroupName`?  and maybe also `outboundType`, looks like `outboundType` would be "Loadbalancer" if we don't set it?

jianlinliu: Ah, I remember it now.
When < 4.12, whatever private or public cluster install, `baseDomainResourceGroupName` is always required, there is a known issue on private cluster install.
When >=4.12, when it is a private cluster, it can be skipped, when it is a external cluster, we have to set it.

So in this case, I think we can directly set `baseDomainResourceGroupName` in install-config.yaml with skipping the version check.

ci-operator/step-registry/ipi/conf/azure/default/ipi-conf-azure-default-commands.sh
```diff
@@ -0,0 +1,116 @@
...
+  name: master
+  platform: {}
+compute:
+- architecture: ${OCP_ARCH}
+  name: worker
+  replicas: ${workers}
+  platform: {}
+EOF
+
+if [ -z "${OUTBOUND_TYPE}" ]; then
```
jianlinliu, line 72: I think we can skip the `OUTBOUND_TYPE` check, not set `outboundType` (by default, it is "outboundType: Loadbalancer") to get a minimal install-config.yaml

ci-operator/step-registry/ipi/conf/azure/default/ipi-conf-azure-default-commands.sh
jinyunma: `baseDomainResourceGroupName` is required for common IPI installation, and not required when installing private cluster. For default configuration test, as we only test common IPI installation, I think we can always specify `baseDomainResourceGroupName` here.

For `outboundType`, I think it can be skipped in install-config, because the default value is `Loadbalancer`, I just checked that there is no error reported when creating manifests, and outboundType is set to `Loadbalancer` in "cluster-config.yaml". Let me do more test when creating cluster without specifying `outboundType`

jinyunma: > Let me do more test when creating cluster without specifying outboundType

Installation is running into stage of "waiting for bootstrapping to complete", external load balancer is created successful, so `outboundType` can also be skipped here.

ci-operator/step-registry/ipi/conf/azure/default/OWNERS
```diff
@@ -0,0 +1,4 @@
+approvers:
+- patrickdillon
+- jianlinliu
+- yunjiang29
```
jianlinliu, line 4:
```suggestion
- yunjiang29
- jinyunma
```

ci-operator/step-registry/ipi/conf/azure/default/ipi-conf-azure-default-commands.sh
gpei: OK, dropped both of `baseDomainResourceGroupName` and `outboundType` condition checks, set `baseDomainResourceGroupName` in the install-config directly and omit the `outboundType`

ci-operator/step-registry/ipi/conf/azure/default/ipi-conf-azure-default-commands.sh
gpei: dropped

ci-operator/config/openshift/verification-tests/openshift-verification-tests-master__installation-nightly-4.16.yaml
gpei: updated to `arm`. But I still prefer f7 to run once at least one week to catch blocker bugs like https://issues.redhat.com/browse/OCPBUGS-42296

jianlinliu: okay.

gpei:
/pj-rehearse periodic-ci-openshift-verification-tests-master-installation-nightly-4.18-azure-ipi-default-mini-perm-arm-f7

PR #59544, opened 2024-12-05
Mirror c9s/rhel-9.6 CoreOS base images

jlebon:
These images are built as part of the CoreOS pipeline. They will be used as bases for building the node images containing OCP-versioned content for CI.

Part of https://github.com/openshift/enhancements/pull/1637.


Prucek:
@jlebon the image needs to be public

jlebon:
> @jlebon the image needs to be public

Hmm gotcha, that explains it.

Unfortunately, I don't think we can make it public. Is there any way to give the necessary privileges to have the mirror service be able to access it? Alternatively, we can have the CoreOS pipeline push to QCI but that would require giving us more privileges. Alternatively, are there any registry repos today that is not public but that the mirror service has access to that we could push to?

Prucek:
the image is accessible with our tooling, I think it could have been caused by manipulation of the image? anyway, it is green now

PR #59556, opened 2024-12-05
Update CRI-O to use go 1.23

saschagrunert:


saschagrunert:
@openshift/cri-o-maintainers @sohankunkerkar @kwilczynski PTAL

saschagrunert:
@haircommander @sohankunkerkar it still looks like that we have no go 1.23 in CI. would it work for us to "hack around" by pulling the latest golang version into the image?

kwilczynski:
@saschagrunert I wonder if we could leverage https://github.com/kevincobain2000/gobrew. Just curious, though.

saschagrunert:
> @saschagrunert I wonder if we could leverage https://github.com/kevincobain2000/gobrew. Just curious, though.

Honestly I'm not even sure if the current approach is something we should do. If we upgrade CRI-O to require go 1.23, will we be able to publish it downstream?

kwilczynski:
> > @saschagrunert I wonder if we could leverage [kevincobain2000/gobrew](https://github.com/kevincobain2000/gobrew). Just curious, though.
> 
> Honestly I'm not even sure if the current approach is something we should do. If we upgrade CRI-O to require go 1.23, will we be able to publish it downstream?

@saschagrunert, I thought we are waiting for Kubernetes, which seems to have moved to 1.23—albeit, we in CRI-O can be held up by runc interoperability, I suppose. Plus there is downstream, too... yeah. Bummer.

saschagrunert:
/pj-rehearse pull-ci-cri-o-cri-o-main-ci-cgroupv2-integration

saschagrunert:
I think we already do something like this in CRI-O:

https://github.com/cri-o/cri-o/blob/413d2f6fe62f431858907037616d38af140ce2ec/contrib/test/ci/build/golang.yml#L7-L20

kwilczynski:
> I think we already do something like this in CRI-O:
> 
> [cri-o/cri-o@`413d2f6`/contrib/test/ci/build/golang.yml#L7-L20](https://github.com/cri-o/cri-o/blob/413d2f6fe62f431858907037616d38af140ce2ec/contrib/test/ci/build/golang.yml#L7-L20)

@saschagrunert I wonder if we should add a test that uses `tip`, which can also fail without consequences, for us to be able to catch issues early with the upcoming releases of Go. Thoughts?

saschagrunert:
> > I think we already do something like this in CRI-O:
> > [cri-o/cri-o@`413d2f6`/contrib/test/ci/build/golang.yml#L7-L20](https://github.com/cri-o/cri-o/blob/413d2f6fe62f431858907037616d38af140ce2ec/contrib/test/ci/build/golang.yml#L7-L20)
> 
> @saschagrunert I wonder if we should add a test that uses `tip`, which can also fail without consequences, for us to be able to catch issues early with the upcoming releases of Go. Thoughts?

Yes, as GitHub action for simpler maintainability?

kwilczynski:
> > > I think we already do something like this in CRI-O:
> > > [cri-o/cri-o@`413d2f6`/contrib/test/ci/build/golang.yml#L7-L20](https://github.com/cri-o/cri-o/blob/413d2f6fe62f431858907037616d38af140ce2ec/contrib/test/ci/build/golang.yml#L7-L20)
> > 
> > 
> > @saschagrunert I wonder if we should add a test that uses `tip`, which can also fail without consequences, for us to be able to catch issues early with the upcoming releases of Go. Thoughts?
> 
> Yes, as GitHub action for simpler maintainability?

@saschagrunert, yeah.

PR #59563, opened 2024-12-05
CRI-O: Add arm64 images

saschagrunert:
Refers to https://github.com/openshift/release/pull/59298, https://github.com/cri-o/cri-o/issues/8714


saschagrunert:
This does not seem to be the right way.

PR #59585, opened 2024-12-05
Reduce the number of jobs executions

ccardenosa:
From this commit, the scheduled number of executions will be 1 per day, both for the job that deploys the Hub cluster and the Spoke cluster.

However, it is intended to maximize the availability of the Hub cluster for as long as possible, in order to meet possible unscheduled Spoke cluster deployment requests.

To this end, each day a new Hub cluster will be deployed, which, in case of success, will be preserved for the rest of the week, although at the same time the previous deployment will be invalidated, so that its resources will be decommissioned and from that moment on, new requests for Spoke deployments will be served by the newest available Hub.

If for some reason, the last Hub deployment attempt failed, the last one deployed would still be available, reducing the chances of failure in the job that deploys the Spoke cluster (and that is the one to become Informing) due to the unavailability of the Hub cluster.


ccardenosa:
@natifridman, @neisw,

This PR is ready to be reviewed.

natifridman (approved):
/lgtm

PR #59590, opened 2024-12-05
Fix pullspec of OSC operator

gkurz:
The `substitutions:` stanza will change the pullspec of the OSC controller
in the CSV so that CI pulls the image that was built and pushed to the CI
registry instead of the default one.

This is done with `sed` and it requires the regular expression to fully
match the original pullspec from the CSV, i.e. both the registry part
and the image tag.

Align with the current state of the OSC repo.


bpradipt (approved):
/lgtm

snir911 (approved):
LGTM thanks

PR #59598, opened 2024-12-05
hcp/openstack: rename OSP jobs when running on AWS mgt cluster

EmilienM:
This is an initial effort to separate the hcp/openstack jobs when running on an AWS mgt cluster or an OpenStack mgt cluster.
For now all the jobs run on an AWS mgt cluster so let's make it explicit, and re-use the Kubevirt jobs name.


EmilienM:
I checked https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pj-rehearse/openshift/release/59598/ebc76f50b6072230325a6c53f46e03fbfe200312 and they all contain `aws` which is the initial goal of this PR, so I think it's good to go.
/pj-rehearse ack

PR #59213, opened 2024-11-26
OCPQE-27280 - Add the test for public cluster migrating to private for AWS

gpei:
There's a dedicated section in the official doc https://docs.openshift.com/container-platform/4.17/post_installation_configuration/configuring-private-cluster.html guiding users to migrate the public cluster to a private one, so adding the test job to cover the steps.

PR #59604, opened 2024-12-06
OCPQE-27348 - Move the AWS public-to-private test to E2E job list

gpei:
To keep align with the Azure test in https://github.com/openshift/release/pull/59213/, move the AWS public-to-private job to E2E job list to see if any cluster operator would be affected on this profile.


gpei:
cc @yunjiang29 @jianlinliu @liangxia for the review, thanks.

PR #59614, opened 2024-12-06
AGENT-1035: extend agent informing compact jobs with add-nodes workflow

andfasano:
This patch extends the agent-based installer informing compact jobs by also including the add nodes workflow. Only 4.17+ jobs are affected, as the feature has been introduced just since 4.17 version


andfasano:
Jobs are failing due e2e tests failures, not directly related to the content of the patch

PR #59670, opened 2024-12-09
hcp/osp: increase wait timeout

EmilienM:
A hostedCluster can take more than 15 min to deploy in our CI if our
infrastructure is busy. So let's increase the timeout to 30m which
should help us to have greener jobs.


mandre (commented):
/lgtm

PR #59783, opened 2024-12-11
4.18 vsphere controle plane regression test

skordas:


skordas:
[OCPQE-26748](https://issues.redhat.com/browse/OCPQE-26748)

PR #59785, opened 2024-12-11
Enable 'make ci-validate' for telco-reference release branches

lack:
Signed-off-by: Jim Ramsay <jramsay@redhat.com>

### DIFF
--- a/ci-operator/config/openshift-kni/telco-reference/openshift-kni-telco-reference-release-4.16.yaml
+++ b/ci-operator/config/openshift-kni/telco-reference/openshift-kni-telco-reference-release-4.16.yaml
@@ -1,0 +1,25 @@
+build_root:
+  project_image:
+    dockerfile_literal: |
+      FROM registry.ci.openshift.org/ocp/ubi-python-311:9
+      USER 0
+      RUN dnf -y install git python3 python3-pip diffutils && python3 -m pip install yamllint
+      RUN curl -L https://github.com/openshift/kube-compare/releases/download/v0.6.0/kube-compare_linux_amd64.tar.gz | tar -xz -C /usr/bin kubectl-cluster_compare
+  use_build_cache: true
+resources:
+  '*':
+    limits:
+      memory: 4Gi
+    requests:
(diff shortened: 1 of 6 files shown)


ci-operator/config/openshift-kni/telco-reference/openshift-kni-telco-reference-release-4.16.yaml
```diff
@@ -0,0 +1,25 @@
+build_root:
+  project_image:
+    dockerfile_literal: |
+      FROM registry.ci.openshift.org/ocp/ubi-python-311:9
```
yuvalk, line 4: dont you prefer to have the dockerfile in the original telco-reference repo?

lack: I think I would prefer this, but @imiller0 didn't set it up that way.

imiller0: I put the dockerfile here/inline and enabled use_build_cache below in an attempt to speed up the CI. It would be interesting to compare the average CI time using this model vs dockerfile in the repo.

lack: I when I said I would prefer the in-repo dockerfile, I wasn't thinking about it from a "speed of CI run" but more of a "maintainability" perspective.  If there is a speed-up I'd love to know...

imiller0 (commented):
/lgtm

ci-operator/config/openshift-kni/telco-reference/openshift-kni-telco-reference-release-4.16.yaml
yuvalk: the maintainability aspect, is that it supposed to be easier for the project developers to maintain that file
I suggest also using it in the Makefile build 
so it is being used and exercised regularly

PR #59796, opened 2024-12-11
Enable build06 with limited capacity and vpn capability

deepsm007:
/cc @openshift/test-platform 
/cc @danilo-gemoli we need to monitor build06 tomorrow


danilo-gemoli:
/lgtm
but waiting for my fellow @deepsm007

PR #59799, opened 2024-12-11
csi-operator: openstack jobs are now optional

EmilienM:
They should not be required as their rate of stability is not great enough.


EmilienM:
/close
sounds like we don't need it after all.

PR #59802, opened 2024-12-12
Add changes for amq streams expansion and remove FIPS config

smpsgill:


im-konge:
@Simranpal I just wanted to ask - why is the FIPS config removed? It's because of OCP 4.18 that should be tested with FIPS in the future?

Simranpal:
> @Simranpal I just wanted to ask - why is the FIPS config removed? It's because of OCP 4.18 that should be tested with FIPS in the future?

@im-konge yes that is correct. We are not testing on previous ocp versions. This PR is to update the ocp z-stream configs and removing old FIPS config

PR #59818, opened 2024-12-12
[WIP] Reduce node number for long chain upgrade job

JianLi-RH:
Because some long chain upgrade jobs exceed the TTL of Prow, so we have to reduce the execution time to make sure chain upgrade works.

By default we create 3 master node and 3 worker node, and to upgrade each node will take around 20 minutes, so we can try to reduce the node numbers to short chain upgrade time.


JianLi-RH:
close

PR #59826, opened 2024-12-12
Use the jump_ci for the 'metal' test

kpouget:


kpouget:
@dagrayvid , could we merge this PR? this is for a POC to get TOPSAIL running with the bare-metal clusters, in a similar way to what we currently have with Jenkins. (https://github.com/openshift-psap/topsail/pull/607)

PR #59837, opened 2024-12-12
Switch base image of OSC to rhel9

gkurz:
For the sake of being consistent with the builder which is rhel9 already.


littlejawa (approved):
/lgtm
Thanks @gkurz !

PR #59851, opened 2024-12-13
CORS-3790: Remove existing CI for AWS Outposts

yunjiang29:
Shutdown any resource created by engineering in the AWS Outpost account.

The cluster profile will be removed in a separate PR.


yunjiang29:
cc @gpei @jianlinliu @jiajliu @qiliRedHat for review and awareness

jiajliu:
upgrade part lgtm

jianlinliu:
/lgtm

Does that mean we do not need to cover outpost any more ?

yunjiang29:
> @yunjiang29 Does that mean we do not need to cover outpost any more ? 

Yes.

> How about keeping the steps and install workflow yaml files, just remove those ci jobs

OK

yunjiang29:
@liangxia PTAL, thanks

PR #59853, opened 2024-12-13
[CIGolang]enhance step to handle CLUSTER_PROFILE_DIR

kuiwang02:
CLUSTER_PROFILE_DIR is set by dptp only when cluster_profile is set in your test per
https://docs.ci.openshift.org/docs/architecture/step-registry/#available-environment-variables.
as we known, now some tests do not set it because of using cluster claim, so there is no CLUSTER_PROFILE_DIR,
and then step will fail with CLUSTER_PROFILE_DIR: unbound variable
here we provide default value although it is not used for env indeed, but it could keep step running and then the test should get the related resource in other way if there is no cluster profile for your test.

/cc @Xia-Zhao-rh

### DIFF
--- a/ci-operator/step-registry/openshift-extended/test/disruptive/openshift-extended-test-disruptive-commands.sh
+++ b/ci-operator/step-registry/openshift-extended/test/disruptive/openshift-extended-test-disruptive-commands.sh
@@ -35,12 +35,19 @@ function exit_trap {
     fi
 }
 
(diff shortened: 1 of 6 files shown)


kuiwang02:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.18-amd64-nightly-aws-ipi-shared-vpc-phz-freeipa-f14

PR #59865, opened 2024-12-13
Remove configs for non-existing repo & team

m1kola:
Removing `openshift/azure-sop` from the config. This is used to be a private repo which ARO SRE team used to store standard operating procedure (SOP) docs for ARO based on OpenShift 3.

This repo is no longer relevant and the service has been retired some time ago.


m1kola:
Had to rebase because it conflicted with https://github.com/openshift/release/pull/59815

PR #59910, opened 2024-12-16
Use only jira plugin for Konflux build team repositories

mmorhun:


psturc:
@mmorhun I think if you want to keep only jira plugin, the job configs for these repos need to be removed

PR #59923, opened 2024-12-16
CNV-52131: Switch to a cluster-profile that only includes AWS regions providing …

aleskandro:
…arm64 BM Machines


aleskandro:
(Test passed)

PR #59924, opened 2024-12-16
MGMT-19465: Add CI jobs to test user managed load balancer in assisted-service / assisted-test-infra

danmanor:
relies on - https://github.com/openshift/assisted-test-infra/pull/2597


CrystalChun (approved):
/lgtm

PR #59972, opened 2024-12-17
OCPBUGS-46523: Remove e2e-agent-5control-ipv6-pxe test due to failures

bfournie:
Remove this test fow now. It will be reinstated once it is passing.


sosiouxme:
huh... i thought i was an approver
/approve
?

PR #60008, opened 2024-12-18
HyperShift: switch nested mgmt clusters to use public IP

csrwng:
Switches nested mgmt clusters to use public IP


enxebre:
/lgtm
pending rehearsals

PR #60039, opened 2024-12-19
OCPBUGS-57475: albo: skip operator CredentialsRequest creation

alebedev87:
The operator can bootstrap itself with credentials, eliminating the need for a pre-created CredentialsRequest. This change avoids intermittent issues with CCO secret provisioning caused by the operator namespace being created after the pre-install step.


alebedev87:
You are not stale my friend, still good, keep up!

PR #60041, opened 2024-12-19
opendatahub/vllm: increase timeout for all builds

dtrifiro:


dtrifiro:
Current timeouts seem ok, for now.

PR #60046, opened 2024-12-19
NO-ISSUE: Extend ZTP day2 jobs timeouts

danmanor:
We are getting a lot of timeout failures in day2 ZTP jobs where the job has reached its final 4 hours (default) timeout, for example - https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_assisted-service/7114/pull-ci-openshift-assisted-service-release-ocm-2.12-e2e-ai-operator-ztp-sno-day2-workers-late-binding/1869653475260895232

This can be caused by many reasons like time to get a lease, time to build images, etc.

This jobs are performing hosts installations after the regular cluster installations in day1 which takes extra time. We want to give these jobs enough time to complete. We already have timeouts on assisted-service deployment scripts / specific steps, we don't need a short timeout for the entire job as well.


CrystalChun (approved):
/lgtm

PR #60063, opened 2024-12-20
[INTEROP-7903] OpenShift Pipelines: Change cron expressions to prevent periodic jobs from bypassing weekly trigger job and running on scheduled date.

dfrazzette:
Original periodic job design included setting cron value to a recently passed date. However, they are all configured to run at least once a year; many of these on Dec. 25.

Changing the cron expression for these jobs to syntactically correct but non-existent date:
`cron: 0 23 31 2 * (At 11:00 PM, on day 31 of the month, only in February)`

These changes are to periodic jobs which may be reviewed and merged by OWNERS in `openshift-pipelines` file.


srivickynesh (approved):
/lgtm

PR #60070, opened 2024-12-20
OCPQE-27748: add test-supplementary in destructive job

huali9:
25436 will not run in openshift-extended-test-disruptive after https://github.com/openshift/openshift-tests-private/pull/21972, so add openshift-extended-test-supplementary to ensure 25436 run in destructive jobs. 
@sunzhaohua2 @miyadav @shellyyang1989 PTAL, thanks!


huali9:
Just found the destructive and non-destructive jobs are the same, so only run this case in non-destructive is ok, this pr is not needed.
/close

PR #60176, opened 2024-12-26
OCPQE-26586: Remove cucushit cases in vSphere multi-vcenter profile

duanwei33:
The CSIMigration is not supported in the vSphere multi-vcenter env, so in-tree cases will fail in CI. 
Also it is part of fix for: https://issues.redhat.com/browse/OCPQE-26586.

cc @Phaow @chao007


duanwei33:
cc @WenXinWei, I removed the feature gate in 4.19, just FYI.

duanwei33:
Hi @liangxia, could you help take a look?

PR #52547, opened 2024-05-29
update featuregates for gatewayapi

lihongan:
follow up https://github.com/openshift/release/pull/48873
fix the name in feature_gates

Adding `ClusterAPIInstallAWS=true` to workaround install issue with `CustomNoUpgrade` set, but if they fix it from installer side, we could remove this.

PR #60217, opened 2024-12-30
adding Gateway API conformance tests

lihongan:
Adding optional job for upstream gateway api conformance test

### DIFF
--- a/ci-operator/config/openshift/cluster-ingress-operator/openshift-cluster-ingress-operator-master.yaml
+++ b/ci-operator/config/openshift/cluster-ingress-operator/openshift-cluster-ingress-operator-master.yaml
@@ -143,6 +143,22 @@ tests:
         requests:
           cpu: 100m
     workflow: ipi-aws
+- always_run: false
+  as: e2e-aws-gatewayapi-conformance
+  optional: true
+  steps:
+    cluster_profile: aws
+    env:
+      FEATURE_SET: TechPreviewNoUpgrade
+    test:
+    - as: test
+      cli: latest
+      commands: make gatewayapi-conformance
+      from: src
+      resources:
+        requests:
+          cpu: 100m
+    workflow: ipi-aws
 - as: e2e-aws-ovn-techpreview
   optional: true
   steps:
--- a/ci-operator/jobs/openshift/cluster-ingress-operator/openshift-cluster-ingress-operator-master-presubmits.yaml
+++ b/ci-operator/jobs/openshift/cluster-ingress-operator/openshift-cluster-ingress-operator-master-presubmits.yaml
@@ -74,6 +74,79 @@ presubmits:
           secretName: result-aggregator
     trigger: (?m)^/test( | .* )e2e-aws-gatewayapi,?($|\s.*)
   - agent: kubernetes
+    always_run: false
+    branches:
+    - ^master$
+    - ^master-
+    cluster: build03
+    context: ci/prow/e2e-aws-gatewayapi-conformance
+    decorate: true
+    labels:
+      ci-operator.openshift.io/cloud: aws
+      ci-operator.openshift.io/cloud-cluster-profile: aws
+      ci.openshift.io/generator: prowgen
+      pj-rehearse.openshift.io/can-be-rehearsed: "true"
+    name: pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-gatewayapi-conformance
+    optional: true
+    rerun_command: /test e2e-aws-gatewayapi-conformance
+    spec:
+      containers:
+      - args:
+        - --gcs-upload-secret=/secrets/gcs/service-account.json
+        - --image-import-pull-secret=/etc/pull-secret/.dockerconfigjson
+        - --lease-server-credentials-file=/etc/boskos/credentials
+        - --report-credentials-file=/etc/report/credentials
+        - --secret-dir=/secrets/ci-pull-credentials
+        - --target=e2e-aws-gatewayapi-conformance
+        command:
+        - ci-operator
+        image: ci-operator:latest
+        imagePullPolicy: Always
+        name: ""
+        resources:
+          requests:
+            cpu: 10m
+        volumeMounts:
+        - mountPath: /etc/boskos
+          name: boskos
+          readOnly: true
+        - mountPath: /secrets/ci-pull-credentials
+          name: ci-pull-credentials
+          readOnly: true
+        - mountPath: /secrets/gcs
+          name: gcs-credentials
+          readOnly: true
+        - mountPath: /secrets/manifest-tool
+          name: manifest-tool-local-pusher
+          readOnly: true
+        - mountPath: /etc/pull-secret
+          name: pull-secret
+          readOnly: true
+        - mountPath: /etc/report
+          name: result-aggregator
+          readOnly: true
+      serviceAccountName: ci-operator
+      volumes:
+      - name: boskos
+        secret:
+          items:
+          - key: credentials
+            path: credentials
+          secretName: boskos-credentials
+      - name: ci-pull-credentials
+        secret:
+          secretName: ci-pull-credentials
+      - name: manifest-tool-local-pusher
+        secret:
+          secretName: manifest-tool-local-pusher
+      - name: pull-secret
+        secret:
+          secretName: registry-pull-credentials
+      - name: result-aggregator
+        secret:
+          secretName: result-aggregator
+    trigger: (?m)^/test( | .* )e2e-aws-gatewayapi-conformance,?($|\s.*)
+  - agent: kubernetes
     always_run: true
     branches:
     - ^master$


ci-operator/step-registry/ingress/gatewayapi/conformance-tests/ingress-gatewayapi-conformance-tests-commands.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+echo "Go version: $(go version)"
+cd /go/src/github.com
+mkdir kubernetes-sigs && cd kubernetes-sigs
```
Miciah, line 9: Using `&&` here means that the `cd` command won't run if the directory already exists or if the `mkdir` command fails for any other reason:
```text
$ mkdir foo && cd foo
$ pwd
/home/mmasters/foo
$ cd ..
$ mkdir foo && cd foo
mkdir: cannot create directory ‘foo’: File exists
$ pwd
/home/mmasters
$ 
```
`mkdir -p` doesn't return an error if the directory already exists:
```text
$ ls -d foo
foo
$ mkdir -p foo && cd foo
$ pwd
/home/mmasters/foo
$ 
```
So it is better to use `mkdir -p` to avoid an error if the directory already exists (I assume we want to use the existing directory in this case), and use a separate `cd` command so that a failing `mkdir` command causes the script to exit on error:
```suggestion
mkdir -p kubernetes-sigs
cd kubernetes-sigs
```

ci-operator/config/openshift/cluster-ingress-operator/openshift-cluster-ingress-operator-master.yaml
```diff
@@ -143,6 +143,18 @@ tests:
...
           cpu: 100m
     workflow: ipi-aws
+- always_run: false
+  as: e2e-aws-gatewayapi-conformance
+  optional: true
+  steps:
+    cluster_profile: aws
+    env:
+      FEATURE_GATES: '[GatewayAPI=true]'
+      FEATURE_SET: CustomNoUpgrade
```
Miciah, line 153: We will need to update this once https://github.com/openshift/api/pull/2081 merges (which we want to happen as soon as possible).

ci-operator/step-registry/ingress/gatewayapi/conformance-tests/ingress-gatewayapi-conformance-tests-commands.sh
```diff
@@ -0,0 +1,21 @@
...
+# currently CRD v1.0.0 is supported
+git clone --branch release-1.0 https://github.com/kubernetes-sigs/gateway-api
```
Miciah, line 11: It is good that the CRD version is explicit.  However, I wonder whether it would be better and practical to specify the version in the openshift/cluster-ingress-operator repository.  We might eventually want to use this same job for multiple release branches with various CRD versions, and anyway, it would make version bumps easier if the CRD version for conformance tests and the CRD manifests were both in the same repository (namely openshift/cluster-ingress-operator).  

If it isn't easy to read the CRD version from another repository, then we can leave this as a follow-up.

Miciah: I suppose the simplest thing would be to add a `Makefile` target in openshift/cluster-ingress-operator for running the conformance tests.

ci-operator/config/openshift/cluster-ingress-operator/openshift-cluster-ingress-operator-master.yaml
```diff
@@ -143,6 +143,18 @@ tests:
         requests:
           cpu: 100m
     workflow: ipi-aws
+- always_run: false
```
candita, line 146: Shouldn't we remove the entry for `e2e-aws-gatewayapi` above?  Doesn't this replace it?

lihongan: I don't think the conformance test can replace existing `e2e-aws-gatewayapi`, the conformance is to test the implementation matches the API specification, but implementation specific features will not be covered by it. For example, `e2e-aws-gateway` covers the OSSM and DNSRecord tests but conformance doesn't.
We could run conformance tests as part of our full e2e test suite. see https://gateway-api.sigs.k8s.io/concepts/conformance/#conformance

ci-operator/config/openshift/cluster-ingress-operator/openshift-cluster-ingress-operator-master.yaml
lihongan: OK, once https://github.com/openshift/api/pull/2081 merges we could just set `FEATURE_SET: TechPreviewNoUpgrade`.
And finally we will promote to GA in 4.19 right?

ci-operator/step-registry/ingress/gatewayapi/conformance-tests/ingress-gatewayapi-conformance-tests-commands.sh
lihongan: Good suggestion, let me try if I can make it with `Makefile`.

ci-operator/config/openshift/cluster-ingress-operator/openshift-cluster-ingress-operator-master.yaml
candita: Thanks.  I misunderstood. Now I see the tests are completely different:
  - ref: ingress-gatewayapi-conf-gatewayclass refers to ci-operator/step-registry/ingress/gatewayapi/conf-gatewayclass/ingress-gatewayapi-conf-gatewayclass-ref.yaml
  - ref: ingress-gatewayapi-conformance-tests refers to ci-operator/step-registry/ingress/gatewayapi/conformance-tests/ingress-gatewayapi-conformance-tests-ref.yaml

This is a new pattern for us.  Normally we make very few code additions here because we keep test code in the repo it is used in.  I wonder what the benefit is to keep it in the step-registry?

lihongan: Yes, Miciah also recommend to add a Makefile target in openshift/cluster-ingress-operator for running the conformance tests, I'll try it later. If that works then we don't need keep it in the step-registry.

Currently to run conformance test with AWS we need:
1. ensure one gatewayclass is ready
2. update the timeout const "MaxTimeToConsistency" from 30s to 90s (otherwise some tests will fail)
I'm not sure if the 2nd one is acceptable since we need to update upstream code.

lihongan:
/hold
rehearsable test needs https://github.com/openshift/cluster-ingress-operator/pull/1176 merge

candita:
/pj-rehearse pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-gatewayapi-conformance

candita:
The conformance test log looks good:  https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_release/60217/rehearse-60217-pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-gatewayapi-conformance/1883943002972360704/artifacts/e2e-aws-gatewayapi-conformance/test/build-log.txt

Miciah:
/hold cancel
per https://github.com/openshift/release/pull/60217#issuecomment-2579966335 and https://github.com/openshift/release/pull/60217#issuecomment-2616856807.

ci-operator/config/openshift/cluster-ingress-operator/openshift-cluster-ingress-operator-master.yaml
candita: @lihongan if we need to update upstream code to set the MaxTimeToConsistency we should start that soon.  Hopefully we could do that easily with a runtime flag, e.g. some examples in https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/utils/flags/flags.go.

ci-operator/step-registry/ingress/OWNERS
candita, line 1: It would be better to have the OWNERS change in a separate PR, just in case we need to revert this one.

ci-operator/config/openshift/cluster-ingress-operator/openshift-cluster-ingress-operator-master.yaml
```diff
@@ -143,6 +143,23 @@ tests:
...
+  as: e2e-aws-gatewayapi-conformance
+  optional: true
+  steps:
+    cluster_profile: aws
+    env:
+      FEATURE_GATES: '[GatewayAPI=true]'
```
candita, line 152: You didn't need to add the workaround feature gate `ClusterAPIInstallAWS` here, as you did in https://github.com/openshift/release/pull/52547 for other tests.  Should it be removed from the other tests too?

ci-operator/config/openshift/cluster-ingress-operator/openshift-cluster-ingress-operator-master.yaml
lihongan: Currently we workaround it in the [hack/gatewayapi-conformance](https://github.com/openshift/cluster-ingress-operator/blob/edf5e71e8b08ef23a4d8f0b3fee5630c66625967/hack/gatewayapi-conformance.sh#L46-L48) of C-I-O repo, the original `ref` already removed from this PR.
Will improve it with more flexible way like using CustomTimeoutConfig.

ci-operator/config/openshift/cluster-ingress-operator/openshift-cluster-ingress-operator-master.yaml
lihongan: Yes, the feature gate `ClusterAPIInstallAWS` could be removed since it is temp workaround for 4.16, but it is harmless and we could remove it in another PR. It was added for the job `e2e-aws-gatewayapi` and the job could be removed once GWAPI is promoted to TechPreview/GA, I think.

ci-operator/config/openshift/cluster-ingress-operator/openshift-cluster-ingress-operator-master.yaml
lihongan: Should we hold this PR until https://github.com/openshift/api/pull/2081 get merged ?

candita: Regarding a hold until the feature is merged as TechPreview:

Pro: we won't need another change within a month or so.  We will still need another change to GA, so one change instead of two would be more efficient.

Con: we lose a month or so of conformance tests against OSSM 3.0 GA.

I vote no, it would be better to have the tests against OSSM 3.0.  @Miciah do you agree?

lihongan:
/pj-rehearse ack
the rehearse job passed

ci-operator/config/openshift/cluster-ingress-operator/openshift-cluster-ingress-operator-master.yaml
```diff
@@ -143,6 +143,22 @@ tests:
         requests:
           cpu: 100m
     workflow: ipi-aws
+- always_run: false
+  as: e2e-aws-gatewayapi-conformance
+  optional: true
+  steps:
+    cluster_profile: aws
+    env:
+      FEATURE_SET: TechPreviewNoUpgrade
```
lihongan, line 152: The PR https://github.com/openshift/api/pull/2081 already got merged, so update env `FEATURE_SET` to `TechPreviewNoUpgrade` accordingly.

PR #60273, opened 2025-01-02
[KMM] Chaning the container image for the `lint` job.

ybettan:
The image contain a version of `golangci-lint` which is too old and there is no release of the image with an up-to-date version.

We will let KMM install `golangci-lint` on its own the way it does when there is not other binary present - this will also ensure we are all testing with the same linter version.

---

/assign @yevgeny-shnaidman @TomerNewman


ybettan:
/pj-rehearse ack

We only care about the `lint` job and it wasn't triggered anyway so there is no reason to wait for CI.

PR #60297, opened 2025-01-03
Sync Serverless CI config/

serverless-qe:


mgencur:
There's no need to merge this PR now when it includes the same changes as https://github.com/openshift/release/pull/60069 (which is being merged).

PR #60369, opened 2025-01-07
Adding slack notification when job is aborted

TomerNewman:
We saw that when jobs were being aborted, we did not get notification on slack, which led to unused aws resources that were been kept.
This commit will also send us on the slack channel notification when a job is being aborted.

---

/cc @empovit


empovit:
Nice! I have two comments though
1. It looks like you need to re-generate the jobs (`make jobs`)
2. Is it possible to have different messages for different statuses? I'd like to remind that `aborted` can mean that the job has left orphaned cloud resources that must be reviewed and released. It looks like you can add `{{if eq .Status.State "aborted"}}` and then a detailed explanation, e.g. "If the job was aborted, check for orphaned cloud resources".

PR #60381, opened 2025-01-07
Sync release controller updated deployments for vsphere

jcpowermac:


bradmwilliams:
Ultimately, I'd like to see this be integrated into the release-controller's generation scripts (which can be run via a `make release-controllers` from the root of the repository). 
I can add the necessary support for this and that would take care of the burden of keeping these different locations in sync

PR #60400, opened 2025-01-08
update fast foward task for all ALC repos

xiangjingli:


xiangjingli:
@fxiang1  @dtthuynh  I noticed the fast forward merge from main to release-2.13 failed in all ALC repos. The PR is to apply the fast forward task implemented in the other stolostorn repos. Hope it will make the fast forward work.

Also I reviewed the fast forward task in CLC repos and they have applied this way.

fxiang1:
@xiangjingli Yes, your change is the same as what CLC does for fast forwarding.

PR #60291, opened 2025-01-03
with resource group id to avoid 'No resource group found' error

MayXuQQ:
based on https://cloud.ibm.com/unifiedsupport/cases?number=CS4162762
use resource group id in "ibmcloud target",  try to resolve "No resource group found" issue

PR #60413, opened 2025-01-08
fix bug in #60291

MayXuQQ:
rg should be resource_group


MayXuQQ:
@liangxia  could you give me a lgtm label? thank so much

PR #60427, opened 2025-01-08
Stolostron hypershift golang upgrade

rokej:
Upgraded golang build image from 1.19 to 1.20 for stolostron hypershift addon operator


mikeshng (approved):
/approve

/lgtm

PR #59392, opened 2024-12-02
[IBMCloud] CORS-3364	e2e testing automation: Replace Terraform with CAPI Provider for IBM Cloud

MayXuQQ:
[CORS-3364](https://issues.redhat.com/browse/CORS-3364)	e2e testing automation: Replace Terraform with CAPI Provider for IBM Cloud
verify [CORS-3278](https://issues.redhat.com/browse/CORS-3278) Replace Terraform with CAPI Provider for IBM Cloud
Verify [OCPBUGS-46355](https://issues.redhat.com/browse/OCPBUGS-46355)
Verify [OCPBUGS-45984](https://issues.redhat.com/browse/OCPBUGS-45984)

https://github.com/openshift/installer/blob/d2bb75079b1d73ed533c3b998550090893237bd5/pkg/types/ibmcloud/metadata.go#L51-L56 
changed the serviceEndpoints order in install-config

PR #60462, opened 2025-01-09
fix bug in #59392

MayXuQQ:
_FEATURE_SET: TechPreviewNoUpgrade_ should changed to _FEATURE_SET: CustomNoUpgrade_


MayXuQQ:
@gpei  can I get a lgtm label? thanks a lot

PR #60466, opened 2025-01-09
Update openshift-kni mirroring OWNERS

lack:
Signed-off-by: Jim Ramsay <jramsay@redhat.com>

### DIFF
--- a/core-services/image-mirroring/openshift-kni/OWNERS
+++ b/core-services/image-mirroring/openshift-kni/OWNERS
@@ -6,7 +6,9 @@ approvers:
 - yanirq
 - yuvalk
 - lack
-- nishant-parekh
+- jc-rh
+- browsell
+- donpenney
 options: {}
 reviewers:
 - cynepco3hahue
@@ -17,4 +19,3 @@ reviewers:
 - slintes
 - vladikr
 - yanirq
-- nishant-parekh


core-services/image-mirroring/openshift-kni/OWNERS
```diff
@@ -6,7 +6,9 @@ approvers:
 - yanirq
 - yuvalk
 - lack
-- nishant-parekh
+- jc-rh
+- browsell
+- donpenney
```
donpenney, line 11: These would be needed under reviewers list as well, I think?

lack: I thought approver implied reviewer, but I think you're right that we should all be in both places.

lack: Or maybe I'm right, according to this:

https://www.kubernetes.dev/docs/guide/owners/#quirks-of-the-process

donpenney: The wording there seems a little ambiguous to me. An approver giving a `lgtm` implies an `approve`, but does that imply an approver is able to give a `lgtm`? I know in repos where I'm both reviewer and approver, I can give an `lgtm` without it also adding an `approve`, so it's not clear to me. Could be something to experiment with

PR #60538, opened 2025-01-10
add kueue operator prow jobs

kannon92:


ci-operator/config/openshift/kueue-operator/openshift-kueue-operator-main.yaml
```diff
@@ -0,0 +1,21 @@
+build_root:
+  image_stream_tag:
+    name: release
+    namespace: openshift
+    tag: golang-1.22
```
mrunalp, line 5: Should it be 1.23?

mrunalp (approved):
/lgtm

PR #58388, opened 2024-11-01
add USER_TAGS to help with cloud cost

yasun1:
[OCPQE-26391](https://issues.redhat.com//browse/OCPQE-26391)

PR #60574, opened 2025-01-13
Sanitize PR authors so bot accounts can provision ROSA envs

rhmdnd:
https://github.com/openshift/release/pull/58388 added github users as a tag in AWS environments to help track resources. But certain characters are not allowed as tags, like [ and ], which is how github represents CI accounts (`renovate[bot]`).

Keeping [ and ] in the tag means any CI bot can't provision ROSA environments.

This commit removes [ and ] from the github author string before setting the tag.


rhmdnd:
@yasun1 @pruan-rht 👋🏻 curious if you could take a look at the approach here.

yasun1:
Looks good to me. @pruan-rht  is the change OK to you?

PR #60587, opened 2025-01-13
metal: add missing etcd image coming in k8s 1.29.11

bertinatto:
This image is only used in tests.

We have to do this until https://issues.redhat.com/browse/TRT-1286 is fixed.

/assign @jerpeter1


jerpeter1:
@elfosardo could I get an /approve please?

PR #60612, opened 2025-01-14
Snyk will "fail" only on high and critical vulnerabilities

hector-vido:


smg247:
@hector-vido, this is probably not something that we want. This step is used by other repos as well, and it is unlikely that they want the threshold changed. Maybe we can introduce an env var to set the threshold (and document its existence) though.

PR #60621, opened 2025-01-14
Temporarily disable permafailing 4.19 `e2e-kubevirt-azure` job

orenc1:
this job is permafailing for quite some time.
we're disabling it and in the meanwhile investigate how can we stabilize the job so conformance tests will pass on azure on 4.19.


davidvossel (commented):
/lgtm
/approve

PR #60622, opened 2025-01-14
Update 2.5 and 2.6 IBIO branches to go 1.22

carbonin:
cc @giladravid16 @gamli75 

This is causing the 2.5 and 2.6 CVE PRs to fail CI


carbonin:
Hopefully these will pass, but I'm not sure. I think this needs to be merged even if they fail to get thee tests on the PRs to fix the issues to pass.

carbonin:
Yeah, that's about what I expected.

/pj-rehearse ack

PR #60636, opened 2025-01-14
Add skip_if_only_changed to ODH operator main

grdryn:
The pattern is taken from what was there on the incubation and rhoai
branches.


AjayJagan:
lgtm

PR #60639, opened 2025-01-14
stackrox: add stable image tag for stackrox flavors

davdhacs:
This change is to add mirroring of a "*-stable" set of tags for the stackrox cluster-flavor images.

("latest" will be the staging images for testing changes to the flavor provisioning. "stable" will be the images that tests run on, and it will be updated after tests pass on "latest")


BradLugo (approved):
Verified with https://github.com/stackrox/automation-flavors/actions/runs/12793196931
/lgtm

tommartensen (approved):
/lgtm

PR #60642, opened 2025-01-14
Use docker mirrors where it's supported

jwendell:
For Istio 1.24 onwards, Istio builder image supports the `DOCKER_REGISTRY_MIRRORS` variable, that is used to configure the docker daemon with docker mirrors.

This change makes use of it to setup 2 mirrors for now, Google and Amazon..

### DIFF
--- a/ci-operator/step-registry/servicemesh/envoy/e2e/servicemesh-envoy-e2e-commands.sh
+++ b/ci-operator/step-registry/servicemesh/envoy/e2e/servicemesh-envoy-e2e-commands.sh
@@ -80,6 +80,8 @@ spec:
       privileged: true
       runAsUser: 0
     env:
+    - name: DOCKER_REGISTRY_MIRRORS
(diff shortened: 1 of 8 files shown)


yxun:
You should either 
adding the new env variable in each .sh file below 
or 
adding them in workflow yaml files (you have done that in your PR now) and parsing them to .sh file.

envoy:
https://github.com/openshift/release/blob/master/ci-operator/step-registry/servicemesh/envoy/e2e/servicemesh-envoy-e2e-commands.sh#L82
proxy:
https://github.com/openshift/release/blob/master/ci-operator/step-registry/servicemesh/proxy/e2e/servicemesh-proxy-e2e-commands.sh#L77
istio:
https://github.com/openshift/release/blob/master/ci-operator/step-registry/servicemesh/istio/e2e/servicemesh-istio-e2e-commands.sh#L78

yxun (approved):

PR #60695, opened 2025-01-16
SDN-5555: Restore monitor tests for ipsec serial lane

pperiyasamy:


jluhrsen:
/retitle SDN-5555: Restore monitor tests for ipsec serial lane
/lgtm

PR #60715, opened 2025-01-16
Let hypershift presubmits to install with TECH_PREVIEW_NO_UPGRADE

enxebre:
Driving use case atm is testing autoNode via karpenter. Periodics will continue to run without the flag.


enxebre:
/hold
for the reharsal

enxebre:
```
    create_cluster_test.go:869: Running validation "When OpenStack value is set as platform type it should fail"
    create_cluster_test.go:876: 
        Expected an error to have occurred.  Got:
            <nil>: nil
    --- FAIL: TestOnCreateAPIUX/HostedCluster_creation (0.42s)
```

Needs https://github.com/openshift/hypershift/pull/5410

PR #60526, opened 2025-01-10
NO-JIRA: feat: use different go version

eggfoobar:

PR #60742, opened 2025-01-17
NO-JIRA: Use different go version for LVMS 4.13.z

suleymanakbas91:
Applies [this fix](https://github.com/openshift/release/pull/60526) into 4.13.


suleymanakbas91:
cc @eggfoobar

PR #60756, opened 2025-01-17
Make FIO rosa testing opt-in

rhmdnd:
FIO doesn't support ROSA, but we wired up the testing for it so we can
develop support for it. The CI is already optional, but it runs on every
PR and will always result in a failure until we have time to develop
support for ROSA.

This commit makes it so ROSA CI doesn't run automatically on every
PR, saving compute resources. We still have the ability to run testing,
we just need to opt into it with a PR comment.

We can make this CI run on each PR again once we have support for ROSA
in FIO.


rhmdnd:
@Vincent056 @yuumasato this should help cleanup false negative results in FIO ci.

rhmdnd:
/pj-rehearse ack

No need to run rehearsals here since we're just make the ci-config optional for a single job.

yuumasato (approved):
/approve

PR #60765, opened 2025-01-17
Remove command to install loki operator from e2e-target test in master & 6.1

Clee2691:
This PR removes the command to deploy the loki operator at the start of the `e2e-target` test.

The lokistack tests already install the loki operator through a subscription.

/cc @cahartma @vparfonov @xperimental


jcantrill (commented):
/approve
/lgtm

PR #60776, opened 2025-01-19
add aws cluster profile for ansible-operator-plugin

chiragkyal:
We should run our e2e test into multiple cluster profiles for better verification.

### DIFF
--- a/ci-operator/config/openshift/ansible-operator-plugins/openshift-ansible-operator-plugins-main.yaml
+++ b/ci-operator/config/openshift/ansible-operator-plugins/openshift-ansible-operator-plugins-main.yaml
@@ -69,6 +69,21 @@ tests:
         requests:
           cpu: 100m
     workflow: ipi-gcp
+- as: e2e-ansible-aws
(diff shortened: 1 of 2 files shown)


arkadeepsen:
Are there anything specific to AWS we are checking in this new e2e? If not, then I am not sure we'd need to add it as this would increase the cost of the CI.

chiragkyal:
> Are there anything specific to AWS we are checking in this new e2e? If not, then I am not sure we'd need to add it as this would increase the cost of the CI.

I faced hiccups with GCP while working on https://github.com/openshift/ansible-operator-plugins/pull/28. The cluster was not getting created within the timeout time, but the e2e was passing fine with AWS.

There might be some temporary issue at that time which would have caused the problem. So, this is an optional PR to have the e2e tests running on both the cloud providers. 

Feel free to close if you think we don't need it :)

arkadeepsen:
Closing this as this test will not provide any additional benefit. We can revisit this later if we feel there are any benefits to this.
/close

PR #60809, opened 2025-01-20
Increase cluster resources for obs 4.18 cluster pools

Clee2691:
## Description
This PR increases the cluster pool node sizes for obs 4.18 cluster pools from `default` to `m4.2xlarge`


Clee2691:
@alanconway can I get an approval for this?

PR #60811, opened 2025-01-20
set red-hat-data-services/kubeflow to branch protection false

ckhordiasma:
unmanaged: true did not work as intended, so using protect: false instead


harshad16 (approved):
/lgtm
:+1:

PR #60838, opened 2025-01-21
Add optional job to run Envoy tests from within the proxy repo

jwendell:
It will not run automatically on every PR. It has to be explicitly triggered by commenting `/test envoy`. It won't also block the merge if the job fails.


jwendell:
Related to OSSM-8460

yxun (approved):

PR #60861, opened 2025-01-22
Add telcov10n-ci ns to registry-access

shaior:
cc: @kononovn @natifridman

### DIFF
--- a/clusters/app.ci/registry-access/telcov10n-ci/telcov10n-ci.yaml
+++ b/clusters/app.ci/registry-access/telcov10n-ci/telcov10n-ci.yaml
@@ -1,0 +1,22 @@
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  annotations:
+    openshift.io/description: Published Images for telcov10n-ci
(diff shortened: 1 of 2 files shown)


clusters/app.ci/registry-access/telcov10n-ci/telcov10n-ci.yaml
```diff
@@ -0,0 +1,22 @@
...
+  name: telcov10n-ci-image-puller-binding
+  namespace: telcov10n-ci
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: system:image-puller
+subjects:
+- apiGroup: rbac.authorization.k8s.io
+  kind: Group
+  name: system:authenticated
```
natifridman, line 22: Don't we need the LDAP group name here?

shaior: Not specifically, it means every authenticated user can use the images in this namespace.

natifridman (approved):

PR #60882, opened 2025-01-22
hypershift/openstack: set CleanupOrcImageResourcesAnnotation to false

EmilienM:
In our e2e tests, we want to recycle a Glance image that already exists
so it's being adopted by ORC (openstack-resource-controller). That will
save time and networking transfers.

The images used by e2e will never be deleted by the Image controller.
Instead, our CI has cleanup jobs which take care of removing images
older than 5 hours.

### DIFF
--- a/ci-operator/step-registry/hypershift/openstack/e2e/execute/hypershift-openstack-e2e-execute-commands.sh
+++ b/ci-operator/step-registry/hypershift/openstack/e2e/execute/hypershift-openstack-e2e-execute-commands.sh
@@ -49,4 +49,5 @@ hack/ci-test-e2e.sh ${E2E_EXTRA_ARGS} \
         --e2e.openstack-credentials-file="${SHARED_DIR}/clouds.yaml" \
         --e2e.openstack-external-network-id="${OPENSTACK_EXTERNAL_NETWORK_ID}" \
         --e2e.openstack-node-flavor="${OPENSTACK_COMPUTE_FLAVOR}" \
-	--e2e.openstack-node-image-name="${RHCOS_IMAGE_NAME}"
+	--e2e.openstack-node-image-name="${RHCOS_IMAGE_NAME}" \
+	--e2e.annotations="hypershift.openshift.io/cleanup-orc-image-resource=false"


ci-operator/step-registry/hypershift/openstack/e2e/execute/hypershift-openstack-e2e-execute-commands.sh
```diff
@@ -49,4 +49,5 @@ hack/ci-test-e2e.sh ${E2E_EXTRA_ARGS} \
         --e2e.openstack-credentials-file="${SHARED_DIR}/clouds.yaml" \
         --e2e.openstack-external-network-id="${OPENSTACK_EXTERNAL_NETWORK_ID}" \
         --e2e.openstack-node-flavor="${OPENSTACK_COMPUTE_FLAVOR}" \
-	--e2e.openstack-node-image-name="${RHCOS_IMAGE_NAME}"
+	--e2e.openstack-node-image-name="${RHCOS_IMAGE_NAME}" \
+	--e2e.annotations="hypershift.openshift.io/cleanup-orc-image-resource=false"
```
MaysaMacedo, line 53: If we have both this settings provided, how will Hypershift know which image it should use?

I see this test is specifically using the node-image https://github.com/openshift/hypershift/pull/5139/files#diff-591071886a0db131efd18d3adde08b8d5d7932635e8c5f2895da043fcb4ff6c2R114

Which other test will use the annotation?

EmilienM: In our e2e tests, we want to test the two following scenarios:
* A customer deploys a HostedCluster with a pre-created image in Glance. That's what `--e2e.openstack-node-image-name` will trigger since we have this [test](https://github.com/openshift/hypershift/pull/5139/files#diff-591071886a0db131efd18d3adde08b8d5d7932635e8c5f2895da043fcb4ff6c2R49) that will cover it.
* A customer doesn't provide a pre-created image (what our PR in HCP is doing), but we want the image to not be cleaned up. That's what `--e2e.annotations="hypershift.openshift.io/cleanup-orc-image-resource=false"` will do.

Does it make sense?

EmilienM: We stop populate the node image name via https://github.com/openshift/hypershift/pull/5139/files#diff-b95a8b2f1c46eff71a4a7efa9d1c5b64513c6ea3eb87deffebd979b6dca2f8e6L228
and therefore I think we reach the else. this should be confirmed by the e2e tests

PR #60906, opened 2025-01-23
CONSOLE-4437: Automated Content Security Policy testing in CI

jhadvig:
This PR is adding additional test to openshift/console CI for running CSP violation test.
The test is added by https://github.com/openshift/console/pull/14675

/assign @vojtechszocs 

/hold

### DIFF
--- a/ci-operator/config/openshift/console/openshift-console-main.yaml
+++ b/ci-operator/config/openshift/console/openshift-console-main.yaml
@@ -84,6 +84,11 @@ tests:
   container:
     from: test-bin
   skip_if_only_changed: \.md$|(?:^|/)(?:\.gitignore|OWNERS|LICENSE)$
+- as: csp
+  commands: cd ./frontend && NO_SANDBOX=true yarn test-puppeteer-csp
+  container:
+    from: test-bin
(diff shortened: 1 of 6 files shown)


ci-operator/config/openshift/console/openshift-console-main.yaml
```diff
@@ -84,6 +84,11 @@ tests:
   container:
     from: test-bin
   skip_if_only_changed: \.md$|(?:^|/)(?:\.gitignore|OWNERS|LICENSE)$
+- as: csp
+  commands: cd ./frontend && NO_SANDBOX=true yarn test-puppeteer-csp
+  container:
+    from: test-bin
+  skip_if_only_changed: \.md$|(?:^|/)(?:\.gitignore|OWNERS|LICENSE)$
```
spadgett, line 91: Hey, @jhadvig. If this runs on every PR (or almost every PR), it would be good to make it part of the existing integration test so that we don't create an additional cluster just for CSP tests. It could save us money on CI. openshift/console is a pretty high traffic repo.

@vojtechszocs @jhadvig Are there any limitations with these tests that require a separate job?

vojtechszocs: @spadgett Agreed, we can make it part of an existing integration test to avoid unnecessary overhead.

PR #60916, opened 2025-01-23
Removing tide as we are switching to github actions

frenzyfriday:


frenzyfriday:
Screenshot from 2025-01-25 22-58-27
Screenshot from 2025-01-25 22-58-10
Tested on a dummy repo

PR #60921, opened 2025-01-23
Block openshift/console master branch PRs from merging due to PF6 update

jhadvig:
Adding new `pf6-approval-required` for a short period of time we need to block the console's merge queue from PRs to get merged, due to PF6 update. 
/assign @spadgett @rhamilto

### DIFF
--- a/core-services/prow/02_config/_labels.yaml
+++ b/core-services/prow/02_config/_labels.yaml
@@ -618,6 +618,11 @@ repos:
         addedBy: prow
   openshift/console:
     labels:
+      - color: e11d21
+        description: Indicates if the change is addressing only urgent fixes or CVEs.
+        name: approved-fixes-only
+        target: prs
+        addedBy: humans
       - color: 00D8FF
         description: Related to local-storage-operator-plugin
         name: component/lso
--- a/core-services/prow/02_config/openshift/console/_pluginconfig.yaml
+++ b/core-services/prow/02_config/openshift/console/_pluginconfig.yaml
@@ -33,6 +33,16 @@ label:
       assign_on:
       - label: backport-risk-assessed
       label: cherry-pick-approved
+    - allowed_users:
+      - jhadvig
+      - rhamilto
+      - spadgett
+      - TheRealJon
+      - kmcfaul
+      - nicolethoen
+      - vikram-raj
+      - rohitkrai03
+      label: approved-fixes-only
 plugins:
   openshift/console:
     plugins:
--- a/core-services/prow/02_config/openshift/console/_prowconfig.yaml
+++ b/core-services/prow/02_config/openshift/console/_prowconfig.yaml
@@ -6,6 +6,7 @@ tide:
     labels:
     - acknowledge-critical-fixes-only
     - approved
+    - approved-fixes-only
     - docs-approved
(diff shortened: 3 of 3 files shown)


core-services/prow/02_config/openshift/console/_pluginconfig.yaml
```diff
@@ -33,6 +33,12 @@ label:
       assign_on:
       - label: backport-risk-assessed
       label: cherry-pick-approved
+    - allowed_users:
+      - jhadvig
+      - rhamilto
+      - spadgett
+      - TheRealJon
+      label: pf6-approval-required
```
spadgett, line 41: I think there's another file where the label needs to be defined so it's added to the repo. Do we know if the label gets removed later if we remove it from the config file?

We could go with something more generic that we can use again in the future. `additional-approval-required`? `approved-fixes-only`? I'm open to suggestions :) There's the existing critical fixes only label, but I don't think we can lock that down to just this set of approvers.

spadgett: > I think there's another file where the label needs to be defined so it's added to the repo.

Looks like release/core-services/prow/02_config/_labels.yaml

jhadvig: Yeah, the wording is always hard. I like the `approved-fixes-only` 👍 

Will update.

Thanks @spadgett

core-services/prow/02_config/openshift/console/_pluginconfig.yaml
```diff
@@ -33,6 +33,14 @@ label:
       assign_on:
       - label: backport-risk-assessed
       label: cherry-pick-approved
+    - allowed_users:
+      - jhadvig
+      - rhamilto
+      - spadgett
+      - TheRealJon
+      - kmcfaul
+      - nicolethoen
```
spadgett, line 42: I think we'd also want to add @vikram-raj and @rohitkrai03 from ODC.

@vikram-raj @rohitkrai03 We plan to hold other non-critical changes until the PF6 PR merges, so we're adding a label.

core-services/prow/02_config/_labels.yaml
```diff
@@ -618,6 +618,11 @@ repos:
         addedBy: prow
   openshift/console:
     labels:
+      - color: e11d21
+        description: Indicates if the change is addressing only urgent fixes or CVEs.
+        name: approved-fixes-only
+        target: both
```
spadgett, line 624: Should this be...?
```suggestion
        target: prs
```

jhadvig: It looks like it should 👍

core-services/prow/02_config/openshift/console/_pluginconfig.yaml
jhadvig: added both

vikram-raj:
@jhadvig @spadgett Is it still needed as PF6 PR has already been merged?

spadgett:
@vikram-raj It's not. Holding for now. We might want to introduce these labels in the future for other large features, but it's not urgent.

/hold

PR #60941, opened 2025-01-24
off-board red-hat-data-services/kubeflow from tide

jiridanek:
This is in response to instructions in a slack thread https://redhat-internal.slack.com/archives/CHY2E1BL4/p1737626272677609

> If you want to use tide please install the app to these repos,
> if not, please remove your tide-config from openshift/release.


jstourac:
Just to be sure - what is the process of merging changes there once we remove the tide?

jstourac:
Based on that slack discussion, I presume we want this now.

/lgtm

PR #60943, opened 2025-01-24
CORS-3860 - Enhance the testing on Azure fully private cluster with proxy enabled

gpei:
In previous Azure fully private + proxy testing, the cluster was using a vNet that's accessible to the Azure Cloud services, which is not particularly ideal, with the bug fix of https://issues.redhat.com/browse/OCPBUGS-24412, bootstrap also could use proxy to fetch the ignition from the Azure Blob Storage for 4.17+ versions, then we don't need the vNet to access to the Internet or Azure Cloud service anymore, because all the outgoing traffic could go through the proxy server to finish everything

### DIFF
--- a/ci-operator/step-registry/azure/provision/vnet/azure-provision-vnet-commands.sh
+++ b/ci-operator/step-registry/azure/provision/vnet/azure-provision-vnet-commands.sh
@@ -41,11 +41,31 @@ function create_disconnected_network() {
     for nsg in $subnet_nsgs; do
         run_command "az network nsg rule create -g ${rg} --nsg-name '${nsg}' -n 'DenyInternet' --priority 1010 --access Deny --source-port-ranges '*' --source-address-prefixes 'VirtualNetwork' --destination-address-prefixes 'Internet' --destination-port-ranges '*' --direction Outbound"
         if [[ "${CLUSTER_TYPE}" != "azurestack" ]]; then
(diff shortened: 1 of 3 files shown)


gpei:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.18-multi-nightly-azure-ipi-fullyprivate-proxy-arm-f7

gpei:
also cc @jinyunma @jianlinliu for review

jianlinliu:
> In previous Azure fully private + proxy testing, the cluster was using a vNet that's accessible to the Internet

In the fullly private proxy install workflow, we did not enable `azure-provision-nat`, and there is no any public lb/IP attached for the vms, that means all the nodes in the cluster would have no internet access natively. That is why we did not enable `RESTRICTED_NETWORK="yes"` explicitly.

gpei:
> that means all the nodes in the cluster would have no internet access natively. That is why we did not enable `RESTRICTED_NETWORK="yes"` explicitly.

Got it, but in previous cluster, all the nodes(including bootstrap) still have the ability to access to Azure Cloud services right?
In https://issues.redhat.com/browse/OCPBUGS-24412, customer was expecting the bootstrap using proxy to fetch the ignition from Azure blob storage, but it didn't, while our testing in the same scenario was working well because the Azure Cloud services are still avaialble.

jianlinliu:
> but in previous cluster, all the nodes(including bootstrap) still have the ability to access to Azure Cloud services right? 

This is also what I am thinking of now, per my understanding, all nodes would have no the ability to access to Azure Cloud services, unless these Azure Cloud service has some internal IP endpoints, the nodes are using internal IP to access them. But I did not hear of that yet.


> In https://issues.redhat.com/browse/OCPBUGS-24412, customer was expecting the bootstrap using proxy to fetch the ignition from Azure blob storage, but it didn't, while our testing in the same scenario was working well because the Azure Cloud services are still avaialble.

I am thinking how we can reproduce it using our install workflow.

PR #60937, opened 2025-01-24
OCPQE-28195: Refine test frequency for 4.18/4.19

liangxia:
Context in https://redhat-internal.slack.com/archives/CH76YSYSC/p1737687653199659

Update test jobs with `f2/f4/f5` to `f7`

PR #60944, opened 2025-01-24
OCPQE-28195: Refine test frequency for 4.18 ci jobs

jianlinliu:
Some jobs are missing in https://github.com/openshift/release/pull/60937


jianlinliu:
cc @liangxia to review.

PR #60948, opened 2025-01-24
MGMT-19657: Adding nmstate operator to assisted installer

zszabo-rh:
Depends on
https://github.com/openshift/assisted-service/pull/7215
https://github.com/openshift/assisted-test-infra/pull/2619


eliorerz (approved):
LGTM

PR #60956, opened 2025-01-24
shiftstack/prune: relax images cleanup on VEXXHOST

EmilienM:
With Hypershift e2e jobs, the RHCOS image will be pushed into Glance for
HostedCluster. The link of the qcow2 is taken from the release payload.
In order to save network transfers, we will keep the RHCOS image longer
than 7h so it can be re-used by other jobs which need it until there is
a new image that is published.

For now, we'll cleanup the images every 2 weeks (only where the job
runs, so on VEXXHOST) to see how that works but this could be adjusted in the future.

### DIFF
--- a/ci-operator/config/shiftstack/ci/shiftstack-ci-main.yaml
+++ b/ci-operator/config/shiftstack/ci/shiftstack-ci-main.yaml
@@ -28,6 +28,7 @@ tests:
   steps:
     env:
       CLUSTER_TYPE_OVERRIDE: openstack-vexxhost
+      IMAGES_TTL: 336h
       OPENSTACK_AUTHENTICATION_METHOD: password
     workflow: shiftstack-cleanup
 - as: cleanup-openstack-nfv
--- a/ci-operator/step-registry/shiftstack/cleanup/shiftstack-cleanup-commands.sh
+++ b/ci-operator/step-registry/shiftstack/cleanup/shiftstack-cleanup-commands.sh
@@ -7,4 +7,16 @@ export OS_CLIENT_CONFIG_FILE="${SHARED_DIR}/clouds.yaml"
 export CLUSTER_TYPE="${CLUSTER_TYPE_OVERRIDE:-${CLUSTER_TYPE:-}}"
 SLACK_HOOK="$(</var/run/slack-hooks/shiftstack-bot)"
 
-prune --resource-ttl="$RESOURCE_TTL" --slack-hook="$SLACK_HOOK" --no-dry-run > "${ARTIFACT_DIR}/result.json"
+EXCLUDE_ARG=""
+if [ -n "${IMAGES_TTL:-}" ]; then
+    EXCLUDE_ARG="--exclude=images"
+    prune --resource-ttl="$IMAGES_TTL" --slack-hook="$SLACK_HOOK" --no-dry-run --include="images" > "/tmp/images-result.json"
+fi
+
+prune --resource-ttl="$RESOURCE_TTL" --slack-hook="$SLACK_HOOK" --no-dry-run "$EXCLUDE_ARG" > "/tmp/all-result.json"
+
+if [ -n "${IMAGES_TTL:-}" ]; then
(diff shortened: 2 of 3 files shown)


ci-operator/step-registry/shiftstack/cleanup/shiftstack-cleanup-commands.sh
```diff
@@ -7,4 +7,16 @@ export OS_CLIENT_CONFIG_FILE="${SHARED_DIR}/clouds.yaml"
 export CLUSTER_TYPE="${CLUSTER_TYPE_OVERRIDE:-${CLUSTER_TYPE:-}}"
 SLACK_HOOK="$(</var/run/slack-hooks/shiftstack-bot)"
 
-prune --resource-ttl="$RESOURCE_TTL" --slack-hook="$SLACK_HOOK" --no-dry-run > "${ARTIFACT_DIR}/result.json"
+EXCLUDE_ARG=""
+if [ -n "${IMAGES_TTL:-}" ]; then
+    EXCLUDE_ARG="--exclude=images"
+    prune --resource-ttl="$IMAGES_TTL" --slack-hook="$SLACK_HOOK" --no-dry-run "$EXCLUDE_ARG" --include="images" > "/tmp/images-result.json"
```
MaysaMacedo, line 13: why should we include and exclude images from the prune?

EmilienM: because we would separate the cleanup between images (every 2 weeks) and the rest of the resources (every 7h)

MaysaMacedo: In the description of [this PR](https://github.com/shiftstack/prune/pull/26) you mention resource types can not be included and excluded simultaneously. Is that the same that is happening here?

MaysaMacedo: > because we would separate the cleanup between images (every 2 weeks) and the rest of the resources (every 7h)

shouldn't it be just --include?

EmilienM: in the same command we can't exclude and include at the same time.
so the proposal here is to run 2 prune commands:
* one that only include images
* one that exclude images

that way we can control image TTL and other resources TTL.

EmilienM: sorry I fixed my error

ci-operator/config/shiftstack/ci/shiftstack-ci-main.yaml
```diff
@@ -28,13 +28,15 @@ tests:
...
       CLUSTER_TYPE_OVERRIDE: openstack-vexxhost
+      IMAGES_TTL: 336h
       OPENSTACK_AUTHENTICATION_METHOD: password
     workflow: shiftstack-cleanup
 - as: cleanup-openstack-nfv
   interval: 6h
   steps:
     env:
       CLUSTER_TYPE_OVERRIDE: openstack-nfv
+      IMAGES_TTL: 336h
```
MaysaMacedo, line 39: why all these clouds need to have that specific clean up?
Can you share the exact jobs that are expected to have ORC images running?

EmilienM: Right now the HCP e2e job only runs on VEXXHOST AFIK but I didn't want to create a snowflake so I added this everywhere for consistency. Happy to discuss this and change it.

MaysaMacedo: I thought we could set it up only on the cloud that is really needed, which is vexxhost. In case there any issue related to that, it would be isolated there.
There are other hypershift related jobs that run on mecha for example (hypershift-e2e-openstack-aws-csi-cinder), but it does not run the hypershift e2e test.
I don't have strong opinion on this topic, if we decide to do it in all of them, let's also add to az0.

EmilienM: sure let me update it then

PR #60961, opened 2025-01-24
bump golangci-lint to latest

patrickdillon:
Switch main branch to use latest version mirrored in CI.

The latest version I see mirrored is v1.62.2, so if we see that version in the rehearsal, it works!


barbacbd (commented):
/lgtm

PR #61013, opened 2025-01-27
telcov10n: hcp: add tier3 tests and fix for timeout

rshemtov13:
After PR #60874 has been merged, the timeout can now be increased up to 24hours in order to run our entire hcp test suite. Changed the workflow timeout to 24 hours and added tier3 ginkgo label.

### DIFF
--- a/ci-operator/step-registry/telcov10n/functional/hcp-cnf/tests/telcov10n-functional-hcp-cnf-tests-ref.yaml
+++ b/ci-operator/step-registry/telcov10n/functional/hcp-cnf/tests/telcov10n-functional-hcp-cnf-tests-ref.yaml
@@ -5,7 +5,7 @@ ref:
     name: telco-runner
     tag: latest
   commands: telcov10n-functional-hcp-cnf-tests-commands.sh
-  timeout: 8h
+  timeout: 24h
   resources:
     requests:
       cpu: 100m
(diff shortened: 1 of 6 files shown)


ci-operator/step-registry/telcov10n/functional/hcp-cnf/tests/telcov10n-functional-hcp-cnf-tests-ref.yaml
```diff
@@ -5,7 +5,7 @@ ref:
     name: telco-runner
     tag: latest
   commands: telcov10n-functional-hcp-cnf-tests-commands.sh
-  timeout: 8h
+  timeout: 24h
```
natifridman, line 8: The timeout of 24 hours is for the entire job, including some CI Operator build steps so we might not be able to use the entire 24 hours

rshemtov13: Yes as of now the entire suite should be about ~21 hours +- 1 hour, with the deployment of hypershift together.
So I think it's fine to leave it like this

sshnaidm: But you set 24hr only for one step. 24 hours might be timeout for the whole job and 21 hour for this step.

rshemtov13: Ah okay. So putting the timeout [here](https://github.com/openshift/release/blob/master/ci-operator/step-registry/telcov10n/functional/hcp-cnf/telcov10n-functional-hcp-cnf-workflow.yaml) would make better sense?

rshemtov13: Ah that was the actual issue. We cannot set in the config a higher timeout then 8 hours we're getting an error for that. This bypasses it.

PR #61015, opened 2025-01-27
pin golangci-lint to v1.63.4

tthvo:
Pin CI golangci-lint to `v1.63.4` on main. Using `latest` might bump version and introduce breaking changes.

The latest at the moment is `v1.63.4` (sorry, there was a typo when I said v1.64.3).

### DIFF
--- a/ci-operator/config/openshift/installer/openshift-installer-main.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-main.yaml
@@ -30,7 +30,7 @@ base_images:
   golangci-lint:
     name: golangci-lint
     namespace: ci
-    tag: latest
+    tag: v1.63.4
   nested-environment-builder:
     name: nested-environment-builder
     namespace: ci


tthvo:
Hm, it looks like `v1.63.4` tag is not available but `latest` does point to it. I do need to change the mirror configuration:

https://github.com/openshift/release/blob/03ce60190b689d2006783fe6a5bb41d7c38fe116/core-services/image-mirroring/_config.yaml#L746

@patrickdillon Is it okay to do so?

patrickdillon:
> Hm, it looks like `v1.63.4` tag is not available but `latest` does point to it. I do need to change the mirror configuration:
> 
> https://github.com/openshift/release/blob/03ce60190b689d2006783fe6a5bb41d7c38fe116/core-services/image-mirroring/_config.yaml#L746
> 
> @patrickdillon Is it okay to do so?

Oh, I see, the `latest` tag is a literal tag on the mirrored image... it doesn't select the latest version :facepalm:  We can't change the `latest` tag, but you could potentially add a tag for `v1.63.4` pointing to the same image

tthvo:
> Oh, I see, the `latest` tag is a literal tag on the mirrored image... it doesn't select the latest version 🤦 We can't change the `latest` tag, but you could potentially add a tag for `v1.63.4` pointing to the same image

Sounds good! I separate the patch to another PR: https://github.com/openshift/release/pull/61018

patrickdillon:
/approve
/lgtm

waiting on rehearsal

PR #61018, opened 2025-01-27
mirror: create v1.63.4 tag for golangci-lint

tthvo:
Create a tag `v1.63.4` to point to golangci-lint `v1.63.4` (`latest`). This allows other to pin version of the tool.


patrickdillon:
/lgtm

We will need approval from the test platform @danilo-gemoli PTAL

PR #61020, opened 2025-01-27
Add dualstack vsphere presubmits to installer

MaysaMacedo:


MaysaMacedo:
/cc @mkowalski 

I noticed the installer didn't have vsphere dualstack jobs, so I'm proposing to add some.
Let me know if you would additional versions to be covered.
I took the same job definition from [here.](https://github.com/openshift/release/blob/master/ci-operator/config/openshift/cluster-network-operator/openshift-cluster-network-operator-master.yaml#L160C1-L171C50)

MaysaMacedo:
/hold

for Mat's input.

ci-operator/config/openshift/installer/openshift-installer-release-4.19.yaml
```diff
@@ -1140,6 +1140,19 @@ tests:
...
+  optional: true
+  steps:
+    cluster_profile: vsphere-elastic
+    env:
+      EXTRA_MG_ARGS: --host-network
+      TEST_SKIPS: Services Secondary IP Family \[LinuxOnly\] should function for pod-Service\|
+        Services Secondary IP Family \[LinuxOnly\] should function for node-Service\|
+        Services Secondary IP Family \[LinuxOnly\] should function for endpoint-Service\|
+        Services Secondary IP Family \[LinuxOnly\] should function for service endpoints
+        using hostNetwork\| should have ipv4 and ipv6 internal node ip
```
mkowalski, line 1154: We shouldn't skip those tests any more. I know they are still skipped in CNO because it's impossible to get a quick approval there, but starting from around 4.15 or 4.16 those tests should actually pass

mkowalski:
It is great to have this test added ! We never cared about o/installer repo because we do not do much development here, but it's actually very helpful

Just one nit regarding skipped tests and we are good to go

mkowalski:
/pj-rehearse pull-ci-openshift-installer-release-4.18-e2e-vsphere-ovn-dualstack

mkowalski:
>Just making sure there isn't any expectation of the ipv6 address being routable or used to connect services (registry,vcenter,etc)

That's correct, no any requirement like this

mkowalski:
/pj-rehearse ack

4.18 job succeeded. 4.19 fails but it's a problem with 4.19, we observe similar issues in other repo

MaysaMacedo:
@mkowalski can you take another look in this PR and apply lgtm if it makes sense to you?

mkowalski:
I can see you are doing
```
- always_run: false
  optional: true
```
what means the job will never run automatically, you will always need to go and explicitly write `/test [...]` in the PR in o/installer.

Is it expected? I mean, I guess it helps when you yourself develop something there and want the test to run, but it will not prevent someone else from introducing a regression because no one will deliberately execute this particular test.

I know it's a tradeoff between quick&cheap CI versus reliable CI, but I would personally try having it optional + always_run. That way it won't block merge, but will give a clear signal that stuff is wrong. Also for you and me will be easy to see which commit introduced regression

MaysaMacedo:
> I can see you are doing
> 
> ```
> - always_run: false
>   optional: true
> ```
> 
> what means the job will never run automatically, you will always need to go and explicitly write `/test [...]` in the PR in o/installer.
> 
> Is it expected? I mean, I guess it helps when you yourself develop something there and want the test to run, but it will not prevent someone else from introducing a regression because no one will deliberately execute this particular test.
> 
> I know it's a tradeoff between quick&cheap CI versus reliable CI, but I would personally try having it optional + always_run. That way it won't block merge, but will give a clear signal that stuff is wrong. Also for you and me will be easy to see which commit introduced regression

@mkowalski we often try to do it this way with OpenStack because we're limited on cloud capacity and we don't want every job that runs consuming some of our capacity, just the needed ones. I'm not sure if the same happens with VSphere, so I decide to propose the economic approach. If you would prefer other way, let me know

mkowalski:
> @mkowalski we often try to do it this way with OpenStack because we're limited on cloud capacity and we don't want every job that runs consuming some of our capacity, just the needed ones. I'm not sure if the same happens with VSphere, so I decide to propose the economic approach. If you would prefer other way, let me know

Good for me, no problem. I just know from my experience that if you create optional not-always-run job, you will be the only person ever using it ;-)

PR #61024, opened 2025-01-28
[INTEROP-7949] Add Firewatch key to fail with test failures. (Exit 1)

dfrazzette:
Currently, Prow reports Quay OCP CI job as Passed, even with test failures. Introducing Firewatch flag to fail job with test failures.


dfrazzette:
Test failures expected; test passed to confirm Firewatch flag works.

Simranpal (approved):
/lgtm

PR #61032, opened 2025-01-28
CM-480: update pipeline image tags to use v1.15.5

siddhibhor-56:
tag updates in o/cert-manager-operator configs with version [v1.15.5]


swghosh:
/retitle CM-480: update pipeline image tags to use v1.15.5

PR #61049, opened 2025-01-28
[INTEROP-7949] Quay: Add Firewatch flag to FIPS config to fail with test failures.

dfrazzette:
Adding flag to fail the Quay on OCP 4.18 periodic job if there are test failures .


dfrazzette:
Test failures expected; test passed to confirm Firewatch flag works and Prow marks test as failed for FIPS config.

PR #61060, opened 2025-01-28
PERFSCALE-3664 - Increase 252node ipsec timeouts

afcollins:
It looks counterintuitive, but the setting in _config.yaml appears to be a default timeout, and then we remove the timeout from the individual job configs, which uses the defaults.  Removing from the job config is required becuase a value over 8h fails validation.


chentex (approved):
/lgtm

PR #61062, opened 2025-01-28
installer: make azure-upi test optional

patrickdillon:
This test fails frequently and sometimes blocks PR. Make it optional!


barbacbd (commented):
/lgtm

/pj-rehease ack

PR #61071, opened 2025-01-29
Use correct upstream branch in sync job

FilipB:


FilipB:
`./ossm/merge_upstream.sh: No such file or directory` will be fixed when https://github.com/openshift-service-mesh/sail-operator/pull/207 is merged

dgn (approved):

PR #61088, opened 2025-01-29
Moving the opp app test to be the last test

gparvin:
We are moving this to try allowing other tests to run in scenarios where some of OPP fails to install but we choose to continue testing. The specific scenario is a quay install failure on new OCP builds that we want to bypass.  The risk is whether the app test will work after running the component specific tests, but I think that's a low risk. Note that it is expected this test will always fail if all of OPP is not functioning correctly.


gparvin:
/pj-rehearse periodic-ci-stolostron-policy-collection-main-opp-ocp4.18-interop-vsphere-openshift-plus-interop-vsphere

PR #61090, opened 2025-01-29
MOC support OCL using v1 API

sergiordlr:
Support using v1 API to enable OCL.  The mco-conf-day2-enable-ocl step will be able to configure OCL using v1alpha API (<=4.18 clusters) and v1 API (4.19+ clusters).

### DIFF
--- a/ci-operator/step-registry/mco/conf/day2/enable-ocl/mco-conf-day2-enable-ocl-commands.sh
+++ b/ci-operator/step-registry/mco/conf/day2/enable-ocl/mco-conf-day2-enable-ocl-commands.sh
@@ -4,6 +4,10 @@ set -e
 set -u
 set -o pipefail
 
+# v1 is used in clusters 4.19+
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/mco/conf/day2/enable-ocl/mco-conf-day2-enable-ocl-commands.sh
```diff
@@ -4,6 +4,8 @@ set -e
 set -u
 set -o pipefail
 
+MOSC_V1_API_VERSION="v1"
+MOSC_V1ALPHA_API_VERSION="v1alpha1"
```
ptalgulk01, line 8: Can we please add in comment the OCP version here, so it will help to get more info about this parameters .

sergiordlr: Done

PR #61110, opened 2025-01-29
Adds perfscale 4.19 slack reporter configs

afcollins:
Perfscale 4.19 jobs were missing slack reporter configs, so this adds them in.


paigerube14 (approved):
/lgtm

PR #61140, opened 2025-01-30
[release-1.35] Initialize Serverless Operator CI config

serverless-qe:
Serverless Operator CI config for release-1.35.


pierDipi:
I don't think there is a need to test this

PR #61148, opened 2025-01-30
add e2e tests to instaslice

cpmeadors:
This adds e2e test to the the main configuration.  It includes building the images, installing the deps, installing the operator, and running the e2e tests


ci-operator/config/openshift/instaslice-operator/openshift-instaslice-operator-main.yaml
```diff
@@ -3,11 +3,47 @@ base_images:
     name: golangci-lint
     namespace: ci
     tag: v1.54.2
+  ocp_builder_rhel-9-golang-1.22-openshift-4.19:
```
kannon92, line 6: 4.19 is using golang 1.23.

PR #61160, opened 2025-01-31
Adding step to stop ephemeral rehearsal cluster 

JoaoFula:
at that point so that we can run the must gather and hopefully replicate the issue.

Recently in OLS, the must-gather has failed several times but in a flaky manner. We were asked to add this step and perform investigation on the cluster.


JoaoFula:
/pj-rehearse periodic-ci-openshift-lightspeed-service-main-4.17-e2e-ols-cluster-periodics

PR #61181, opened 2025-01-31
prow/orc: update config

EmilienM:
using CAPO as reference, add more config, especially the approved plugin and the backports.


EmilienM:
We first need https://github.com/openshift/openstack-resource-controller/pull/1
/hold
/pj-rehearse ack

PR #61187, opened 2025-02-01
Disable health check from SNO jetlag deploys

josecastillolema:
Disabling the health-check step in jetlag SNO deployments because it expects a public vlan setup.
A proper alternative would look like:

   - Implementing the health check within jetlag playbooks
   - Finishing the public vlan setup for SNOs that @rsevilla87 was working on


josecastillolema:
cc @akrzos

josecastillolema:
@akrzos @jtaleric can I have some eyes here?
Thanks!

akrzos (approved):

PR #61209, opened 2025-02-03
orc: sync from release-1.0 instead of main

EmilienM:
We don't want to sync from main, it was a mistake.


mandre (commented):
/lgtm
/approve

PR #61215, opened 2025-02-03
WIP test

bear-redhat:


dtrifiro:
@bear-redhat  Just fyi: the smoke test has some issues that cause it to hang, but we can't debug it because it produces no output and just hangs

PR #61260, opened 2025-02-04
hypershift: create separate techpreview job

sjenning:
We need the required presub test to run exactly the way the periodic/payload jobs do.

However, running an additional job with HO techpreview enabled on every PR would strain CI resources.

The PR
* Turns off TechPreview for the required presub
* Add a new optional presub that runs with TechPreview enabled


enxebre:
/pj-rehearse pull-ci-openshift-hypershift-main-e2e-aws-techpreview

PR #61283, opened 2025-02-05
add security scanning workflow for all the ztp releases

rauhersu:


rauhersu:
@Missxiaoguo , can you '/ok-to-test' :) ?

PR #61284, opened 2025-02-05
kni: cnf-tests: support arm64 arch

shajmakh:
For all supported versions, support buidling cnf-tests over arm systems. ref: https://docs.ci.openshift.org/docs/how-tos/multi-architecture/


shajmakh:
/hold
for pre-merge testing

PR #61352, opened 2025-02-06
Update netobserv-console-plugin build recipe

jotak:


jotak:
@memodi the "image" rehearse job passed, I guess it's the important one here, right?

memodi:
> @memodi the "image" rehearse job passed, I guess it's the important one here, right?

yes, that's right. thanks!

/lgtm

PR #61357, opened 2025-02-06
Hive: Bump periodics for OCP 4.19, go 1.23

2uasimojo:
[HIVE-2740](https://issues.redhat.com//browse/HIVE-2740)


suhanime (commented):
@2uasimojo I noticed it in the other PR, but for the e2e where we claim the cluster from AWS pool, we should be updating the OCP version to a relevant stable one. Maybe 4.17 in this case if not 4.18? What do you think?

2uasimojo:
> @2uasimojo I noticed it in the other PR, but for the e2e where we claim the cluster from AWS pool, we should be updating the OCP version to a relevant stable one. Maybe 4.17 in this case if not 4.18? What do you think?

That's just the hub, so I haven't been worrying about it.

PR #61361, opened 2025-02-06
release-controller: add rhel-9.6-early-kernel

jlebon:
This is actually just the same as the 4.19 plashet, but scoped to the kernel only (with `includepkgs`) so that it's usable in contexts where we only want RHEL packages. See also https://issues.redhat.com/browse/ART-11248.


aaradhak (approved):
LGTM

PR #61364, opened 2025-02-06
disable ovnk techpreview presubmits (excl BGP)

jluhrsen:
now that UDN is GA for 4.19 and 4.18 we can save cloud spending by making these techpreview jobs `always_run: false`

they can still be run with the /test comment in PRs that may want them


jluhrsen:
@tssurya @jcaamano I still think we want this in right? it's just sitting idle

jluhrsen:
> @tssurya @jcaamano I still think we want this in right? it's just sitting idle

@tssurya @jcaamano just noticed this still open in my list. NBD to me. just pinging you here just in case you still want it.

PR #61398, opened 2025-02-07
turning off events when nodes are restarted

paigerube14:
When a node is restarted/stopped-> started the list of cluster events are super long and I don't think we necessarily care about them other than the alerts that are fired at the end of the run. Want to turn off cluster event collection on power outage and node scenarios

Example of a lot of events here: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/periodic-ci-redhat-chaos-prow-scripts-main-4.15-nightly-krkn-hub-node-tests-415/1887713312439799808/artifacts/krkn-hub-node-tests-415/redhat-chaos-node-disruptions-master-outage/build-log.txt


chaitanyaenr (approved):
/lgtm

PR #61406, opened 2025-02-07
Hive: Use newer clusterpool

2uasimojo:
Two e2e jobs in each config were using old (4.12 and 4.14) clusterpools to grab hub clusters. Update to 4.17.

[HIVE-2770](https://issues.redhat.com//browse/HIVE-2770)

### DIFF
--- a/ci-operator/config/openshift/hive/openshift-hive-master.yaml
+++ b/ci-operator/config/openshift/hive/openshift-hive-master.yaml
@@ -89,7 +89,7 @@ tests:
     owner: openshift-ci
     product: ocp
     timeout: 1h0m0s
-    version: "4.14"
+    version: "4.17"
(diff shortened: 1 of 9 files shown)


2uasimojo:
/test ci/prow/ordered-prow-config

Hopefully doesn't require a rebase.

2uasimojo:
mce-2.3 e2e is busted on the OVNKubernetes transition. Not concerned about it since 2.3 is EOL -- we'll fix it if we ever need to resurrect this job (probably by simply using an older OCP release).

suhanime:
@2uasimojo Want to drop 2.3 (and below) files from the commit? I'll approve then

We can even delete them - in general - since they're EOL

2uasimojo:
/pj-rehearse ack

@suhanime I considered it, but there's some small chance we're going to have to use those branches again at some point, and resurrecting the CI config would be painful. In any case I'd like to defer the decision to drop those configs -- along with dropping their branches from the hive repo -- via a separate card.

(Note also that they would be failing regardless of this change.)

PR #61433, opened 2025-02-10
RHDH authentication providers

albarbaro:
Adds a new periodic job for rhdh auth providers testing.

### DIFF
--- a/ci-operator/step-registry/redhat-developer/rhdh/auth-providers-nightly/redhat-developer-rhdh-auth-providers-nightly-ref.metadata.json
+++ b/ci-operator/step-registry/redhat-developer/rhdh/auth-providers-nightly/redhat-developer-rhdh-auth-providers-nightly-ref.metadata.json
@@ -3,14 +3,12 @@
 	"owners": {
 		"approvers": [
 			"josephca",
-			"rnapoles-rh",
 			"subhashkhileri",
 			"zdrapela"
 		],
 		"reviewers": [
 			"albarbaro",
 			"josephca",
-			"rnapoles-rh",
 			"subhashkhileri",
 			"zdrapela"
 		]


ci-operator/step-registry/redhat-developer/rhdh/auth-providers-nightly/OWNERS
```diff
@@ -0,0 +1,18 @@
+# DO NOT EDIT; this file is auto-generated using https://github.com/openshift/ci-tools.
+# Fetched from https://github.com/redhat-developer/rhdh root OWNERS
+# If the repo had OWNERS_ALIASES then the aliases were expanded
+# Logins who are not members of 'openshift' organization were filtered out
+# See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md
+
+approvers:
+- josephca
+- rnapoles-rh
```
zdrapela, line 9: We are now removing rnapoles from all the jobs.

ci-operator/config/redhat-developer/rhdh/redhat-developer-rhdh-main.yaml
```diff
@@ -155,6 +155,11 @@ tests:
     - ref: redhat-developer-rhdh-osd-gcp-nightly
     workflow: redhat-developer-rhdh-osd-gcp-claim-cluster
   timeout: 3h40m0s
+- as: e2e-tests-nightly-auth-providers
+  cron: 0 7 * * *
```
zdrapela, line 159: If we want to be able to run the auth providers job from a PR check, you can add this and generate the presubmit job.
```suggestion
  cron: 0 7 * * *
  optional: true
  presubmit: true
```
Note that after the job is generated by the `make` commands, make sure to change the `always_run: true` to `always_run: false` for your job so it doesn't trigger automatically on new PRs

albarbaro: We are not adding the PR trigger at the moment, as this job will run a static cluster due to the Github redirect urls limitations.

albarbaro: I'll have a Jira anyway to track the PR checks trigger work. We will definitely need that but will require more work.

PR #61476, opened 2025-02-10
 OSD-26766 - Adding test to check for a leading zero

anispate:
Adding a Prow post-merge check for a leading zero and checking if those test fails as an imperfect workaround for now as part of [OSD-26766](https://issues.redhat.com//browse/OSD-26766) ticket

### DIFF
--- a/ci-operator/config/openshift/certman-operator/openshift-certman-operator-master.yaml
+++ b/ci-operator/config/openshift/certman-operator/openshift-certman-operator-master.yaml
@@ -48,6 +48,24 @@ tests:
   commands: make validate
   container:
     from: src
+- as: leading-zero-check
+  postsubmit: true
+  steps:
+    test:
+    - as: check-commit-message
+      commands: |
+        COMMIT_HASH=$(git rev-parse --short HEAD)
+        if [[ $COMMIT_HASH =~ ^0 ]]; then
+            echo "Git commit hash starts with a leading zero. Please generate a dummy commit to generate a new hash."
+            echo "we're doing this due to an OLM bug (https://issues.redhat.com/browse/OSD-13681) and the fix is to add a dummy commit"
+            exit 1
+        fi
+        exit 0
+      from: base
+      resources:
+        requests:
+          cpu: 10m
+          memory: 100Mi
 zz_generated_metadata:
   branch: master
   org: openshift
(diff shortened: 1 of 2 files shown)


ci-operator/config/openshift/certman-operator/openshift-certman-operator-master.yaml
```diff
@@ -48,6 +48,23 @@ tests:
...
     from: src
+- as: leading-zero-check
+  postsubmit: true
+  steps:
+    test:
+    - as: check-commit-message
+      commands: |
+        COMMIT_MESSAGE=$(git log -1 --pretty=%B)
+        if [[ $COMMIT_MESSAGE =~ ^0 ]]; then
+            echo "Error: Commit message starts with a leading zero."
```
clcollins, line 59:
```suggestion
            echo "Error: Git commit hash starts with a leading zero.  Please generate a new commit hash."
```

clcollins: You know...we _might_ be able to do this ourselves automatically.  There are automations in prow that generate new commits (eg: managed-cluster-config generating the final YAML and adding to the users' commits automatically).  We could theoretically create a "dummy" commit automatically and re-test for the leading zero.  That may help reduce overall toil.

ci-operator/jobs/openshift/certman-operator/openshift-certman-operator-master-postsubmits.yaml
```diff
@@ -1,5 +1,54 @@
...
+        - --target=leading-zero-check
+        command:
+        - ci-operator
+        image: ci-operator:latest
+        imagePullPolicy: Always
+        name: ""
+        resources:
+          requests:
+            cpu: 10m
+        volumeMounts:
```
clcollins, line 28: Not having any other knowledge of how this works: are these volumes needed, or were they copied from another example?

ci-operator/config/openshift/certman-operator/openshift-certman-operator-master.yaml
```diff
@@ -48,6 +48,23 @@ tests:
...
     from: src
+- as: leading-zero-check
+  postsubmit: true
+  steps:
+    test:
+    - as: check-commit-message
+      commands: |
+        COMMIT_HASH=$(git rev-parse --short HEAD)
+        if [[ $COMMIT_HASH =~ ^0 ]]; then
+            echo "Git commit hash starts with a leading zero. Please generate a new commit hash."
```
abyrne55, line 59: I would suggest expanding this error message a bit to explain that we're doing this due to an OLM bug (bonus points for linking the bug) and that the fix is to add a dummy commit

PR #61516, opened 2025-02-11
Update user list to get label permission of Jenkins component and plugins

prdhamdh:
- [x]  Update user list to get label permission of Jenkins component and plugins


adambkaplan (approved):
/approve

PR #61521, opened 2025-02-11
Update openstack-rhoso cluster profile

rlobillo:
Adding titan03 and titan08 on the openstack-rhoso pool.

Depends-on: https://github.com/openshift/ci-tools/pull/4455


MiguelCarpio:
@rlobillo could you add the fqdn instead of the IP address? You may need to set the DNS server.

rlobillo:
> @rlobillo could you add the fqdn instead of the IP address? You may need to set the DNS server.

We already tested that and it wasn't working. There is a task for that but I'd like to implement a valid job first. I will work on using FQDNs later on.

PR #61527, opened 2025-02-11
openshift/os: drop rhcos-96* tests; create rhcos-9next tests

dustymabe:
In the future we'll use the rhcos-9next tests when proofing out a new RHEL minor version. This will prevent us from having to update it here as much.


jlebon:
/lgtm

Rehearsals for the new tests will fail since those entrypoints don't exist yet. We'll iterate on it in https://github.com/openshift/os/pull/1742.

/pj-rehearse ack

PR #61528, opened 2025-02-11
prow/orc: require acknowledge-critical-fixes-only label

EmilienM:
Like other components in the release payload, require this label to be
set until things stabilize.


mdbooth:
This looks right, but I don't fully understand this config, tbh.

/lgtm

PR #61529, opened 2025-02-11
remove operator-framework/operator-registry master & nonexistent release branches from prow

grokspawn:
operator-framework/operator-registry is used to release OCP versions 4.1...4.9. No other v4.X branches exist or will be created. Since 4.9 this repository only releases upstream components.

master uses GHA / merge queues for releasing, and we require the same jobs to pass there as what prow required, so we don't need any prow protection.


grokspawn:
RH is donating prow/tide resources for the o-f repositories, so we're actually going to go the opposite direction here.  So I'll close this. :)

PR #61568, opened 2025-02-12
Allow the use of custom RHCOS ISOs for PreGA version

ccardenosa:
Since some features are not available in PreGA versions, a mechanism is needed to allow us to use customized versions of some of the artifacts that are necessary for the installation process via ZTP.

The reason for this particular change is to avoid the problem reported in the following bug already reported: https://issues.redhat.com/browse/OCPBUGS-50585


ccardenosa:
/pj-rehearse periodic-ci-openshift-kni-eco-ci-cd-main-nightly-4.19-telcov10n-metal-single-node-spoke

ccardenosa:
/pj-rehearse periodic-ci-openshift-kni-eco-ci-cd-main-nightly-4.19-telcov10n-metal-single-node-spoke

ccardenosa:
Honestly, I'm not so sure it's a _workaround_, but I see what you mean. But since we are testing PreGA versions and that necessarily means that all components may not always be available (as in this case), we expect this sort of patches.

What you would have to think about is how to detect this kind of situation, because at some point you would have to revert the change (like associate this kind of PreGA stuff to some Jira and have it stop applying when it closes or something like that. However, I am not sure how hard that it would be to develop).

ci-operator/step-registry/telcov10n/jiras/telcov10n-jiras-commands.sh
```diff
@@ -0,0 +1,65 @@
...
+  echo
+  echo "####################################################################################################"
+  echo
+
+}
+
+function closed_jiras {
+
+  # TODO: Send a Slack notification or email to signal that a PreGA workaround
+  # should be reviewed
```
ccardenosa, line 57: @natifridman ,

I have tried to implement something very naive but good enough to track PreGA missing components. If any of you have an idea to extend it so that a notification can be sent in the even of a tracked Jira is closed, that would be of great help. WDYT?

ci-operator/step-registry/telcov10n/jiras/telcov10n-jiras-commands.sh
```diff
@@ -0,0 +1,65 @@
...
+  echo
```
ccardenosa, line 58: @natifridman ,

I have tried to implement something very naive but good enough to track PreGA missing components. If any of you have an idea to extend it so that a notification can be sent in the even of a tracked Jira is closed, that would be of great help. WDYT?

ccardenosa:
/pj-rehearse periodic-ci-openshift-kni-eco-ci-cd-main-nightly-4.19-telcov10n-metal-single-node-spoke

ccardenosa:
/pj-rehearse periodic-ci-openshift-kni-eco-ci-cd-main-nightly-4.19-telcov10n-metal-single-node-spoke

ccardenosa:
/pj-rehearse periodic-ci-openshift-kni-eco-ci-cd-main-nightly-4.19-telcov10n-metal-single-node-spoke

ci-operator/step-registry/telcov10n/jiras/telcov10n-jiras-commands.sh
natifridman: I'm not sure I fully understand what you are looking for, but if you need to send Slack notifications you can use [webhooks](https://api.slack.com/messaging/webhooks)

ccardenosa: Actually two things are needed. The slack webhooks and Jira API Rest access via Token to check out whether the related Jira has been closed and it is time to revert the workaround so that the job will continue testing as much final GA artefacts as possible.

For now, this is just one way to follow up on what you commented. At least this way we will know what we have been changing so that the Job can work with the PreGA builds.

natifridman: Maybe it's something we can do manually by following the Jira issue to get a notification? not sure if it's something that needs automation here...

ccardenosa: Ok. Let's agree on the following procedure then:
- Find out the related Jira (or create a new one to track if no one exist) and add yourself to the watchers group.
-  Create an new entry in the Job config like the one...

```
JIRAS: |-
  - description: OCP 4.19 failing to deploy due to bootkube.sh error, missing oc command
    links:
      - https://issues.redhat.com/browse/OCPBUGS-50585
    PRs:
      - https://github.com/openshift/release/pull/61568
```

  ... where you put the tracking info needed to revert the WA along with the WA PR itself.
  - Finally, at info will be dumped in the Job's logs.
  
  Make sense?
  
  I am going to rework this commit to remove the automation proposal.

natifridman: Sounds good

ccardenosa:
/pj-rehearse periodic-ci-openshift-kni-eco-ci-cd-main-nightly-4.19-telcov10n-metal-single-node-spoke

ccardenosa:
/pj-rehearse periodic-ci-openshift-kni-eco-ci-cd-main-nightly-4.19-telcov10n-metal-single-node-spoke

PR #61623, opened 2025-02-13
OSD-26766 - Fixing the test by adding presubmit flag

anispate:
[OSD-26766](https://issues.redhat.com//browse/OSD-26766) - Adding the presubmit flag as well as updating the image

### DIFF
--- a/ci-operator/config/openshift/certman-operator/openshift-certman-operator-master.yaml
+++ b/ci-operator/config/openshift/certman-operator/openshift-certman-operator-master.yaml
@@ -48,7 +48,24 @@ tests:
   commands: make validate
   container:
     from: src
-- as: leading-zero-check
+- as: pre-leading-zero-check
+  steps:
+    test:
+    - as: check-commit-message
+      commands: |
+        COMMIT_HASH=$(git rev-parse --short HEAD)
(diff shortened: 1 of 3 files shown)


ci-operator/config/openshift/certman-operator/openshift-certman-operator-master.yaml
```diff
@@ -50,6 +50,7 @@ tests:
     from: src
 - as: leading-zero-check
   postsubmit: true
+  presubmit: true
```
deepsm007, line 53: You cannot add postsubmit and presubmit in the same config

anispate: I added both as part of the same config due to the response here: https://redhat-internal.slack.com/archives/CBN38N3MW/p1739457362371669?thread_ts=1739395878.023609&cid=CBN38N3MW

deepsm007: Add two separate config, one which contains `postsubmit: true` and other without postsubmit and run `make jobs` it will create a postsubmit and presubmit job for you

PR #61628, opened 2025-02-13
Move logging-vector to component version build

jcantrill:
This PR:

* Modifies the CI for logging-vector to be based upon vector component version
* Collects vector image mirroring into a single file

cc @xperimental @vparfonov

### DIFF
--- a/ci-operator/config/ViaQ/vector/ViaQ-vector-v0.37.1-rh.yaml
+++ b/ci-operator/config/ViaQ/vector/ViaQ-vector-v0.37.1-rh.yaml
@@ -61,8 +61,8 @@ images:
   to: logging-vector-test-unit
 promotion:
   to:
-  - name: "6.1"
-    namespace: logging
+  - namespace: logging
+    tag: v0.37.1
 releases:
   latest:
     release:
@@ -174,6 +174,6 @@ tests:
           cpu: 100m
     workflow: generic-claim
 zz_generated_metadata:
-  branch: release-6.1
+  branch: v0.37.1-rh
   org: ViaQ
   repo: vector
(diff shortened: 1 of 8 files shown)


ci-operator/config/ViaQ/vector/ViaQ-vector-v0.37.1-rh.yaml
```diff
@@ -61,8 +61,8 @@ images:
   to: logging-vector-test-unit
 promotion:
   to:
-  - name: "6.1"
-    namespace: logging
+  - namespace: logging
+    tag: v0.37.1
```
xperimental, line 65: Is `tag` correct here? I have only seen `name` so far. Loki is also using `name: v3.3.2`.

xperimental: One more: Should we keep the promotion to `6.1`, so that the builds that currently depend on this image still get it? Then we can do follow-up PRs to remove that dependency and, as a last one, remove the promotion?

jcantrill: It is.  You can only use name or tag.  If you use tag, then the push to the registry uses buildname:tag instead of the openshift pattern which is the name spec'd here and the "tag" is the buildname.

jcantrill: there is a version of 6.1 which already exists in the catalog so all existing builds will still pass

xperimental: > It is. You can only use name or tag. If you use tag, then the push to the registry uses buildname:tag instead of the openshift pattern which is the name spec'd here and the "tag" is the buildname.

Just had a look into the CI cluster and I think I need to migrate the Loki builds to that variant as well, it makes much more sense than using the version as the name like in the other pattern :+1:

PR #61677, opened 2025-02-14
remove periodic bgp-techpreview job

jluhrsen:
this job is permafailing and running it as a periodic will end up causing component readiness reports to flag it as a regression. We need this job to be stable and passing before we add it as a periodic.


jluhrsen:
/close
Not needed now that this job is fixed using [this PR](https://issues.redhat.com/browse/OCPBUGS-50636)

PR #61678, opened 2025-02-14
Revert - Increase timeouts for payload jobs to land k8s 1.32

deepsm007:
/cc @bertinatto 

This is an improvement over older method for increasing timeout. For future you can simply update these 2 parameters and we are good to go.


bertinatto (commented):
/lgtm

PR #61712, opened 2025-02-17
StackRox: Add `red-hat-konflux` to trusted apps of `stackrox` and `scanner`

msugakov:
This should make the [Konflux bot](https://github.com/apps/red-hat-konflux) trusted in [StackRox](https://github.com/stackrox/stackrox/) and [Scanner (v2)](https://github.com/stackrox/scanner/) repos so that PRs that are open by Mintmaker/Renovate (similar thing as dependabot) from this bot's account do not need explicit `/ok-to-test` comment to start testing.

From <https://docs.prow.k8s.io/docs/announcements/>:

> January 24, 2022 It is possible now to define GitHub Apps bots as trusted users to allow automatic tests trigger without relying on /ok-to-test from organization member. Trigger and DCO plugins configuration now support additional field trusted_apps, which contains list of GitHub Apps bot usernames without [bot] suffix.

Ref https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1739559859027839

### DIFF
--- a/core-services/prow/02_config/stackrox/scanner/_pluginconfig.yaml
+++ b/core-services/prow/02_config/stackrox/scanner/_pluginconfig.yaml
@@ -37,3 +37,4 @@ triggers:
   - stackrox/scanner
   trusted_apps:
   - dependabot
+  - red-hat-konflux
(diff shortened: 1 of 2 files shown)


msugakov:
It worked. https://github.com/stackrox/stackrox/pull/14321 needs no `ok-to-test` comment.

PR #61718, opened 2025-02-17
[WIP] Ztp left shifting telco lvm issue

ccardenosa:
We have been experiencing [a recurring problem](https://redhat-internal.slack.com/archives/C073NR2AUH3/p1737354636121869) with the time on our jobs, due to this bug.

A way to avoid it might be by selecting always the same disk device, as it is propose here), because, as you can read [here](https://github.com/ccardenosa/release/blob/master/ci-operator/step-registry/baremetal/lab/post/wipe/baremetal-lab-post-wipe-commands.sh#L119) the baremetal-lab-post-wipe-commands.sh step just wipes the disk used to install coreos in the hosts and that means that if, from time to time, the selected server to be used to install the Spoke cluster was previously used for the other job and since they often have two disks, in a previous installation of a OCP cluster the /dev/sda and /dev/sdb were selected the opposite each other and /dev/sdb (that now is /dev/sda) was used to as LVM storage, leading into the issue we see (See further details [here](https://redhat-internal.slack.com/archives/C049U2HRWJU/p1737621219192339)).

This workaround will try to clean up also the partition used by LVM during post step, avoiding the afore mentioned issue.

### DIFF
--- a/ci-operator/step-registry/telcov10n/metal-single-node-spoke/telcov10n-metal-single-node-spoke-workflow.yaml
+++ b/ci-operator/step-registry/telcov10n/metal-single-node-spoke/telcov10n-metal-single-node-spoke-workflow.yaml
@@ -10,8 +10,8 @@ workflow:
     - ref: telcov10n-metal-single-node-spoke-get-and-share-hub-cluster-profile
     - chain: telcov10n-metal-single-node-spoke-deploy
     post:
-    - ref: wait
     - chain: telcov10n-metal-single-node-spoke-clean-up-hub-services
+    - ref: wait
     - chain: baremetal-lab-post
     env:
       TIMEOUT: "+10 hour"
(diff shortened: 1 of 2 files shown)


ccardenosa:
The simplest solution is not valid.

During the execution of the job that deploys the Spoke cluster, an associated Hub cluster is expected to be active, which creates and uses the CR `AgentServiceConfig`, which in turn makes use of the PVs that are created from the CR `LVMCluster`, whose associated operator has created the LVMs that will support those PVs.

As the Hub cluster is deployed in a job that does not delete it once finished (it must prevail to be used by the job that deploys the spokes clusters), controlling the timing of the deletion and cleaning of the disks is a non-trivial operation, always leaving one of the disks with its partition table under the use of LVM.

If that server is selected from the pool of servers to try to deploy a Spoke cluster and the order of how the devices are defined is just the opposite of the previous time (`/dev/sda` is now `/dev/sdb` and vice versa), we end up with the problem that, via ZTP and using assisted installer, the installation will fail due to [this bug](https://issues.redhat.com/browse/MGMT-19737). Once fixed, this problem would be over. Another way to overcome this would be to select always the same disk for the installation. That's something also expected to be available at some point (check [this](https://issues.redhat.com/browse/OCPQE-28186) out).

PR #61730, opened 2025-02-17
Add IBIO backplane-2.9 branch CI file

carbonin:
cc @danmanor @linoyaslan


danmanor:
the error should be fixed now

danmanor (approved):

PR #61738, opened 2025-02-17
remove: kuadrant-console-plugin

jasonmadigan:
Removing kuadrant-console-plugin


jasonmadigan:
@david-martin when you get a moment

PR #61767, opened 2025-02-18
[KMM] Changing the merge-metod from `squash` to `rebase`.

ybettan:
In CI, we have a test to make sure each PR consist of a single commit, in some case, we want to have more than 1 commit in a PR (a good example will be cherry-picking some commits for a release) and in this case, the CI job will be overwritten.

When, it is intended to have multiple commits in a PR, then those commits shouldn't be squashed.

---

/assign @yevgeny-shnaidman @TomerNewman


ybettan:
/pj-rehearse ack

(not need for running the CI jobs as we only change the merge method of PRs).

PR #61768, opened 2025-02-18
Add cherry-pick approvers

kasturinarra:
Currently users in the PR does not have permissions to add cherry-pick-approved labels, updating the users  so that they could add the label.


r4f4 (commented):
/lgtm

PR #61798, opened 2025-02-19
hypershift: remove skip_if_only_changed from optional techpreview job

sjenning:
`skip_if_only_changed` causes the test to run if it is not matched, even with `optional: true` and `always_run: false`


devguyio:
shouldn't we also add `always_run: false` ? According to [the doc](https://docs.ci.openshift.org/docs/architecture/ci-operator/#pre-submit-tests) :

>`always_run`: Set to `false` to disable automatic triggering on every PR. This defaults to `true` (run on every PR) unless `run_if_changed` or `skip_if_only_changed` is set.

devguyio:
I'll remove the hold so that we test it anyway, worst case is that nothing changes and it keeps always running as it does now then we can set `away_run` to false in a separate PR

PR #61816, opened 2025-02-19
Remove leftover stackrox/collector jobs

Molter73:
We've moved away from using OSCI for the collector repo for some time, for some reason these files were not removed and I believe it is causing the openshift-ci bot to continue labeling our PRs.


JoukoVirtanen (approved):
LGTM

PR #61820, opened 2025-02-19
Add role & binding for `ci-operator` to be able to create secretproviderclasses

psalajova:
- A part of the GCP Secret Manager epic, jira link: https://issues.redhat.com/browse/DPTP-4259
- This PR adds the ClusterRole and ClusterRoleBinding so that ci-operator is able to create the `secretproviderclasses.secrets-store.csi.x-k8s.io` resource

### DIFF
--- a/clusters/build-clusters/common/ci-operator/admin_ci-operator_rbac.yaml
+++ b/clusters/build-clusters/common/ci-operator/admin_ci-operator_rbac.yaml
@@ -388,3 +388,28 @@ items:
   - kind: ServiceAccount
     name: ci-operator
     namespace: ci
+
+- kind: ClusterRole
+  apiVersion: rbac.authorization.k8s.io/v1
+  metadata:
+    name: secret-provider-class-creator
+  rules:
+  - apiGroups:
+    - secrets-store.csi.x-k8s.io
+    resources:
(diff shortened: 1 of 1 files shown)


clusters/build-clusters/common/ci-operator/admin_ci-operator_rbac.yaml
```diff
@@ -388,3 +388,28 @@ items:
   - kind: ServiceAccount
     name: ci-operator
     namespace: ci
+
+- kind: ClusterRole
```
Prucek, line 392: Does it have to be a `ClusterRole`? Isn't `Role` enough?

psalajova: hm, ci-operator needs to be able to create those resources (`secretproviderclasses.secrets-store.csi.x-k8s.io`) in all namespaces which will be used for running tests, so idk if Role would be enough (correct me if I'm wrong though)

Prucek: That's what I wanted to hear :D

Prucek (approved):

PR #61826, opened 2025-02-19
CAPO: security job is now required

EmilienM:
The job is passing and it should not be optional.


mandre (commented):
I'm generally in favor of making CI jobs required, however I have concern making the `security` job required due to the little control we have over it. It could break at any time without us able to do anything about it.

mandre (commented):
OpenShift recommends to make this job required...

https://docs.ci.openshift.org/docs/how-tos/add-security-scanning/#how-to-add-the-security-scanning-workflow-as-a-test

/lgtm
/approve

PR #61829, opened 2025-02-19
Add 'skip-if-only-changed' configuration to cluster-group-upgrades-operator CI

fontivan:
- Originally I had borrowed the value from assisted installer but it seems to not work as expected
- Tweaked it to fix the .md detection and correctly escape the . for .tekton

### DIFF
--- a/ci-operator/config/openshift-kni/cluster-group-upgrades-operator/openshift-kni-cluster-group-upgrades-operator-main.yaml
+++ b/ci-operator/config/openshift-kni/cluster-group-upgrades-operator/openshift-kni-cluster-group-upgrades-operator-main.yaml
@@ -51,7 +51,9 @@ tests:
     make ci-job
   container:
     from: src
+  skip_if_only_changed: ^(?:.*/)?(?:\.gitignore|.tekton/.*|.*\.md|OWNERS|OWNERS_ALIASES|LICENSE)$
 - as: integration
+  skip_if_only_changed: ^(?:.*/)?(?:\.gitignore|.tekton/.*|.*\.md|OWNERS|OWNERS_ALIASES|LICENSE)$
   steps:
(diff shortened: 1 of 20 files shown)


ci-operator/config/openshift-kni/cluster-group-upgrades-operator/openshift-kni-cluster-group-upgrades-operator-main.yaml
```diff
@@ -51,7 +51,9 @@ tests:
     make ci-job
   container:
     from: src
+  skip_if_only_changed: ^\.md$|^(?:.*/)?(?:\.gitignore|.tekton/.*|OWNERS|OWNERS_ALIASES|LICENSE)$
```
rauhersu, line 54: Do we need to escape: \.tekton ? Ditto for all the ocurrences.

fontivan: I had lifted this configuration from assisted installer, and assumed it would be correct

But it sure appears that the `.` for `.tekton` needs to be escaped, and it also appears that the forwards slashes are also not escaped correctly

rauhersu: 'skip_if_only_changed' is based on golang regexes, you can check yours here: https://regex101.com/ (select the golang flavour in the menu to the left).

fontivan: image
I think I am relatively happy with this one, coming up in the next commit

PR #61845, opened 2025-02-19
trt-1914: add telcov10n spoke as informing

neisw:


bradmwilliams:
@neisw The release-controller does not search for jobs in all of `openshift/release`.  There is a short list of locations where it looks: https://github.com/openshift/release/blob/21d406af31616fc58dba474220de438bc77f32e2/hack/validate-release-jobs.py#L15-L24
These jobs are not in one of the aforementioned locations.
If this is where those jobs will be, permanently, then add the location to `hack/validate-release-jobs.py` with your PR.

PR #61960, opened 2025-02-24
openshift-logging: Remove builds for 5.6 release

xperimental:
There's still a configuration in `origin-aggregated-logging` that references 5.6 among other EOL versions. Don't know if we can just remove the configuration in that directory completely or if we should update it to use 5.8.

/cc @jcantrill @JoaoBraveCoding


jcantrill:
> There's still a configuration in origin-aggregated-logging that references 5.6 among other EOL versions. Don't know if we can just remove the configuration in that directory completely or if we should update it to use 5.8.

@xperimental my recollection is the only thing we are still building in origin-aggregated-logging was .. kibana?  maybe.  I think we can remove CI for it all-to-gether.  we could maybe even ask for archival of that repo

PR #61961, opened 2025-02-24
openshift-logging: Remove unused Loki branches

xperimental:
Every non-EOL version is either using 3.3.2 or 3.4.2, so no need to keep the previous versions.

/cc @JoaoBraveCoding


JoaoBraveCoding (approved):
/lgtm

PR #61980, opened 2025-02-24
Fix latest aks version command

Patryk-Stefanski:


ci-operator/step-registry/aks/provision/aks-provision-commands.sh
```diff
@@ -77,7 +77,7 @@ fi
 if [[ -n "$AKS_K8S_VERSION" ]]; then
     AKS_CREATE_COMMAND+=(--kubernetes-version "$AKS_K8S_VERSION")
 elif [[ "$USE_LATEST_K8S_VERSION" == "true" ]]; then
-    K8S_LATEST_VERSION=$(az aks get-versions --location "${AZURE_LOCATION}" --output json --query 'max(orchestrators[*].orchestratorVersion)')
+    K8S_LATEST_VERSION=$(az aks get-versions --location "${AZURE_LOCATION}"  --output json --query 'values[?isPreview==`null`].version | sort(@) | [-1]')
```
heliubj18, line 80: the value of K8S_LATEST_VERSION is "1.31" instead of 1.31

Patryk-Stefanski: That shouldnt be an issue, but it the query doesnt seem to be working 
```
az aks get-versions --location eastus2 --output json --query 'values[?isPreview==`null`].version | sort(@) | [-1]'
ERROR: Invalid jmespath query supplied for `--query`: In function sort(), invalid type for value: None, expected one of: ['array-string', 'array-number'], received: "null"
```

PR #61983, opened 2025-02-24
Add XDG directory configuration for TALM

fontivan:
- We need to be using a data directory that we have read/write access to
- In particular, the newer versions of golanglint-ci is using a cache directory that it otherwise lacks permissions to create in the current working directory
- See also: https://github.com/openshift/release/pull/39007/files


fontivan:
Looking for a lgtm @OWNERS.reviewers:
cc. @browsell @donpenney @fedepaol @imiller0 @irinamihai @jc-rh @Missxiaoguo @pixelsoccupied @sakhoury @serngawy @sudomakeinstall2 @vitus133

PR #62040, opened 2025-02-25
PERFSCALE-3720 - Always collect gather from payload runs

afcollins:
Removing variable so we always gather from payload runs to help us debug when we need it.


afcollins:
/pj-rehearse periodic-ci-openshift-qe-ocp-qe-perfscale-ci-main-aws-4.19-nightly-x86-payload-control-plane-6nodes

PR #62046, opened 2025-02-25
Building label-sync locally - simplified

hector-vido:
This will create a standard to other `gcr.io` images.

I created a namespace `gcr-io` that will be used to build other images temporary, until we figure out what we gonna do with the upstream team.
They are triggered with `oc start-build` and are pushing directly to `QCI`.


deepsm007:
/lgtm
Verified and looking good

PR #62075, opened 2025-02-26
hypershift: Adhere to OCP backport requirements

joepvd:
To avoid the scenario of regressing in a particular bug when doing a ystream upgrade, OCP enforces that a fix has landed in `4.(Y+1)` before it can land in `4.Y`.

This PR was constructed by taking the MCO config as basis, and added the `area` requirement.

### DIFF
--- a/core-services/prow/02_config/openshift-priv/hypershift/_prowconfig.yaml
+++ b/core-services/prow/02_config/openshift-priv/hypershift/_prowconfig.yaml
@@ -54,7 +54,7 @@ tide:
     - jira/invalid-bug
     - needs-rebase
     repos:
-    - openshift/hypershift
+    - openshift-priv/hypershift
   - includedBranches:
     - openshift-4.19
     - release-4.19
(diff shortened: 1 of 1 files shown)


core-services/prow/02_config/openshift/hypershift/_prowconfig.yaml
```diff
@@ -1,6 +1,161 @@
...
+    - openshift/hypershift
+  - includedBranches:
+    - openshift-4.19
+    - release-4.19
+    labels:
+    - approved
+    - jira/valid-bug
+    - jira/valid-reference
+    - lgtm
+    - staff-eng-approved
```
sdodson, line 66: I don't think we need `staff-eng-approved` on 4.19, is that in place for other repos right now?

joepvd: The other _prowconfig.yamls have the same. I think the logic is that PRs to 4.19 need to be very special, as changes generally land as fast-forwards from master or main.

sdodson:
/approve
Honestly it's a bit hard to review this structure but if this aligns with other projects I'm fine with it. Will let Seth confirm there shouldn't be any special branch handling for the branches which don't directly affect OCP builds.

PR #62094, opened 2025-02-26
skip prow tests for .tekton exclusive changes

rauhersu:
So far, this only applies to master, but we will need to do the same for z-streams


rauhersu:
/pj-rehearse pull-ci-openshift-kni-cnf-features-deploy-master-e2e-aws-ci-tests

PR #62100, opened 2025-02-26
add perfscale workload to openshift/etcd

jtaleric:
Adding the perfscale control-plane workload


jtaleric:
@sdodson @dusk125

sdodson:
/lgtm
/approve
We can back this out if we don't like it once we're done tesing.

PR #62114, opened 2025-02-26
Removed gcr.io/envoy-ci/envoy-build since nobody is using

hector-vido:
There is no reference of this image inside `openshift/release` to `ci/envoy:1.32` or `gcr.io/envoy-ci/envoy-build` despite the mirror.


smg247:
It was added 3 weeks ago: https://github.com/openshift/release/pull/61313. Let's just leave this alone for now. If it stops working, the user/team who added it will be the ones to fix it.

PR #62133, opened 2025-02-27
Remove build04

liangxia:
Get prepared for re-provisioning build04.


deepsm007:
/close 
we will be keeping build04 as it is and instead migrate the network from sdn -> ovn

PR #62145, opened 2025-02-27
Update kmm-{approvers,reviews}

mresvanis:
This change updates the `kmm-{approvers,reviewers}` by removing users that have moved on and adding `mresvanis`.


pruan-rht:
@mresvanis, I totally missed the tag from you.  I've re-opened the PR and gave the appropriate labels.  It should be merging soon.

PR #62153, opened 2025-02-27
Enable cherrypick action for containers/container-selinux

lsm5:


lsm5:
@haircommander @rhatdan @wrabcak PTAL. Need this change for `slash-cherrypick` commands to work as in https://github.com/containers/container-selinux/pull/359#issuecomment-2687446375

lsm5:
> @liangxia Hi, looks like label requirements are met, but tide says retesting 9 jobs that I don't have visibility into.

Never mind, looks sorted now. Thanks anyway!

lsm5:
could've sworn tide said `In merge pool`. Wonder how that reverted. Anyway, I'll leave this alone until tomorrow.

PR #62155, opened 2025-02-27
`periodic-prow-auto-sippy-config-generator`: point to new 'main' default branch

smg247:


smg247:
it's probably stuck in a batch. I'll check it if it doesn't merge in some time. It says it is retesting now.

PR #62158, opened 2025-02-27
Rebasebot: use latest release for autoscaler rebases

RadekManak:
This PR changes rebase bot config to run autoscaler rebases from latest release instead of master.


RadekManak:
/close
@elmiko is going to do the rebase for this release manually. We will revisit this in the next release, when we get the latest tag detection into rebasebot itself.

elmiko (commented):
this should get us closer to having the rebase for autoscaler be automatic.

/lgtm

PR #62172, opened 2025-02-27
feat: add operator images to artifacts directory

christophermancini:
To improve ease of debugging test failures, this PR adds operator image info to the artifacts directory for persistence in S3. 

https://issues.redhat.com/browse/OSD-26489

### DIFF
--- a/ci-operator/step-registry/osd/gather-extra/osd-gather-extra-commands.sh
+++ b/ci-operator/step-registry/osd/gather-extra/osd-gather-extra-commands.sh
@@ -2,6 +2,7 @@
 
 set -o nounset
 set -o errexit
+set -o pipefail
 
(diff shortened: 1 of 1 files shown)


jbpratt:
/pj-rehearse periodic-ci-openshift-release-master-nightly-4.19-e2e-rosa-sts-ovn

btoll:
I realize I'm late to the party and this has already been approved, but it would be worth setting the `pipefail` option since this is adding a pipeline.

bradmwilliams:
@btoll My Approval was only to acknowledge that I, as an Owner, approve of this change.  It won't merge until someone else, with knowledge about the change, reviews it and tags it with an LGTM.  So, you're not late at all! ;)

bradmwilliams:
> I realize I'm late to the party and this has already been approved, but it would be worth setting the `pipefail` option since this is adding a pipeline.

I'd second this as a recommend change

PR #62293, opened 2025-03-03
ocm new grc release 2.14

JustinKuli:


dhaiducek (approved):
LGTM! Looks like the `_prowconfig.yaml` for `policy-cli` is missed (and actually it's not there for 2.13, either). But that can be a followup PR.

PR #62296, opened 2025-03-03
openshift/os: start building scos node images

Prashanth684:
- build the stream-coreos and stream-coreos-extensions images which is used in OKD releases
- promote these images to the scos imagestream as they will be used as the node images going forward in lieu of the m-o-c pipeline

### DIFF
--- a/ci-operator/config/openshift/os/openshift-os-master__okd-scos.yaml
+++ b/ci-operator/config/openshift/os/openshift-os-master__okd-scos.yaml
@@ -1,0 +1,51 @@
+base_images:
+  ocp_builder_stream9:
+    name: builder
+    namespace: ocp
+    tag: stream9
+images:
+- build_args:
+  - name: OPENSHIFT_CI
+    value: "1"
+  dockerfile_path: Containerfile
+  to: stream-coreos
+- build_args:
+  - name: VARIANT
+    value: okd-c9s
+  dockerfile_path: extensions/Dockerfile
+  inputs:
+    ocp_builder_stream9:
+      as:
+      - registry.access.redhat.com/ubi9/ubi:latest
+    stream-coreos:
+      as:
+      - registry.ci.openshift.org/rhcos-devel/rhel-coreos:latest
+  to: stream-coreos-extensions
+promotion:
+  to:
+  - name: scos-4.19
+    namespace: origin
+releases:
+  latest:
+    integration:
+      include_built_images: true
+      name: scos-4.19
+      namespace: origin
+resources:
+  '*':
+    requests:
+      cpu: 100m
+      memory: 200Mi
+tests:
+- always_run: false
+  as: e2e-aws-ovn
+  optional: true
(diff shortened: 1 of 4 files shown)


ci-operator/config/openshift/os/openshift-os-master__okd-scos.yaml
```diff
@@ -0,0 +1,56 @@
...
+  - name: VARIANT
+    value: okd-c9s
+  dockerfile_path: extensions/Dockerfile
+  inputs:
+    ocp_builder_stream9:
+      as:
+      - registry.access.redhat.com/ubi9/ubi:latest
+    stream-coreos:
+      as:
+      - registry.ci.openshift.org/rhcos-devel/rhel-coreos:latest
```
jbtrystram, line 27: this will build the okd node image on top of rhel-coreos, which is unwanted. 
We should use the c9s base image but we don't publish it to registry.ci as of now. 
We can change that though

jbtrystram (changes requested):

Prashanth684:
/pj-rehearse pull-ci-openshift-os-master-okd-scos-e2e-aws-ovn

ci-operator/config/openshift/os/openshift-os-master__okd-scos.yaml
Prashanth684: @jbtrystram this is now ready to go. Do you have any other concerns before approval?

Prashanth684: > this will build the okd node image on top of rhel-coreos, which is unwanted.

The default in then `Containerfile` of the os repo is a c9s base image (quay.io/openshift-release-dev/ocp-v4.0-art-dev:c9s-coreos) - and that is used to build the stream-coreos image which is then used to build the extensions. hope this makes it clear.

ci-operator/config/openshift/os/openshift-os-master__okd-scos.yaml
```diff
@@ -0,0 +1,56 @@
+base_images:
+  ocp_builder_stream9:
+    name: builder
+    namespace: ocp
+    tag: stream9
+build_root:
+  image_stream_tag:
+    name: coreos-assembler
+    namespace: coreos
+    tag: latest
```
jlebon, line 10: I don't think this is needed?

ci-operator/config/openshift/os/openshift-os-master__okd-scos.yaml
jlebon: We should probably use the same default in both TBH, but no need to block on that.

jlebon (approved):
Looks sane to me!
/lgtm

ci-operator/config/openshift/os/openshift-os-master__okd-scos.yaml
Prashanth684: yes. removed

PR #62354, opened 2025-03-04
Mark ocp destroy post step as optional

mtodor:
We have a situation where [ocp-4](https://issues.redhat.com//browse/ocp-4) destroy step timeouts. Generally speaking, cluster deletion is not something that should impact test execution results.

This PR changes `post` testing step that destroys the OCP cluster to be optional with a defined timeout.

- timeout is defined by randomly picking several test executions and checking duration. In most cases it was ~5mins. Additional 25mis is more than generous.

Relevant docs:
- Post step best effort: https://docs.ci.openshift.org/docs/architecture/step-registry/#marking-post-steps-best-effort
- Timeout on step level: https://docs.ci.openshift.org/docs/architecture/timeouts/#step-registry-test-process-timeouts

Relevant ticket: https://issues.redhat.com/browse/ROX-28310

### DIFF
--- a/ci-operator/step-registry/stackrox/automation-flavors/ocp-4-crun-e2e/stackrox-automation-flavors-ocp-4-crun-e2e-workflow.yaml
+++ b/ci-operator/step-registry/stackrox/automation-flavors/ocp-4-crun-e2e/stackrox-automation-flavors-ocp-4-crun-e2e-workflow.yaml
@@ -1,6 +1,7 @@
 workflow:
   as: stackrox-automation-flavors-ocp-4-crun-e2e
   steps:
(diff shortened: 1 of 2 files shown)


janisz:
What happens when we hit the timeout? If nothing than it can be lower. Also looks like we have a lot of resources for cluster deletion but this is something for another pr

janisz (approved):

tommartensen (commented):
Q: Will we still get Jira CI_Failure tickets if the cluster deletion fails?

tommartensen (approved):
/lgtm

janisz (approved):

PR #62364, opened 2025-03-04
Prune installer okd images

patrickdillon:
The okd variant is building multiple images, some of which have long-running builds, that are not used. The okd images job is frequently timing out, so we should be able to just simply remove these image builds.

### DIFF
--- a/ci-operator/config/openshift/installer/openshift-installer-main__okd-scos.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-main__okd-scos.yaml
@@ -82,29 +82,6 @@ images:
 - build_args:
   - name: TAGS
     value: scos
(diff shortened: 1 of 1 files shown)


patrickdillon:
I suspect we can also remove some or all of these okd base images, but let's start small with this pr:

https://github.com/openshift/release/blob/a7872cb9e7c06169c0c0e9a7678c62ad7134e572/ci-operator/config/openshift/installer/openshift-installer-main__okd-scos.yaml#L34-L49

patrickdillon:
/pj-rehearse pull-ci-openshift-installer-main-okd-scos-e2e-aws-ovn pull-ci-openshift-installer-main-okd-scos-images

PR #62376, opened 2025-03-05
openshift-os-master: change `from` to `inputs`

jlebon:
Using `from` will only replace the last `FROM` in the image. But we want the flexibility to switch to a multi-stage build in the future if we want. Let's switch to using `inputs` instead.


ravanelli (approved):
LGTM

PR #62404, opened 2025-03-05
Reverting change of removing the prow-jobfor the post-leading-zero-check and pre-leading-zero-check

anispate:
Reverting :
https://github.com/openshift/release/pull/61828
https://github.com/openshift/release/pull/62120


clcollins (approved):
/lgtm
/approve

PR #62471, opened 2025-03-06
[INTEROP-7913] OCP 4.19 expansion for JBoss Fuse scenario.

dfrazzette:


dfrazzette:
Rehearsal job for weekly run passed. Rehearsal on FIPS cluster showed test failures (link); will investigate and debug post-merge.

PR #62504, opened 2025-03-06
Define sno_use_lab_dhcp for latest jetlag-ci

akrzos:


akrzos:
/pj-rehearse pull-ci-redhat-performance-jetlag-main-deploy-5nodes

josecastillolema:
> 
> TASK [create-inventory : Download ocpinventory.json] ***************************
> Thursday 06 March 2025  16:31:45 -0500 (0:00:00.020)       0:00:00.408 ******** 
> fatal: [localhost]: FAILED! => {"changed": false, "content": "", "elapsed": 30, "msg": "Status code was -1 and not [200]: Request failed: ", "redirected": false, "status": -1, "url": "http://quads.rdu3.labs.perfscale.redhat.com/cloud/cloud19_ocpinventory.json"}

ci-operator/step-registry/openshift-qe/installer/bm/deploy/openshift-qe-installer-bm-deploy-commands.sh
```diff
@@ -26,6 +26,7 @@ ocp_version: $OCP_VERSION
 ocp_build: $OCP_BUILD
 networktype: OVNKubernetes
 public_vlan: $PUBLIC_VLAN
+sno_use_lab_dhcp: false
```
josecastillolema, line 29: It's is going to be always false for all scenarios (mno, sno, public, etc.) ?

akrzos: Basically I have added a var `sno_use_lab_dhcp` which is to trigger jetlag to set SNOs to have zero static networking setup, which ends up deploying the SNO on the lab's DHCP public interface. We don't have any scenarios to use this networking setup in CPT/CI (yet) but since this var is defined in the all.sample.yml, it must be defined here. Alternatively I can have it defined a 2nd time within defaults/main.yml vars file for roles in which the var is referenced however then I duplicate more instances of the var existing. (That is a possibility too)

josecastillolema: understood!

josecastillolema:
@akrzos if I understand correctly to get a suceessfull rehearsal here we need to poing at this jetlag PR right? https://github.com/redhat-performance/jetlag/pull/621

You can test a specific PR from Prow overwriting the JETLAG_PR variable [here](https://github.com/openshift/release/blob/master/ci-operator/step-registry/openshift-qe/installer/bm/deploy/openshift-qe-installer-bm-deploy-ref.yaml#L45-L48).

So what I suggest:

- point to that PR for the test
- get a successful rehearsal
- delete the PR reference for the final merge

akrzos:
> @akrzos if I understand correctly to get a suceessfull rehearsal here we need to poing at this jetlag PR right? [redhat-performance/jetlag#621](https://github.com/redhat-performance/jetlag/pull/621)
> 
> You can test a specific PR from Prow overwriting the JETLAG_PR variable [here](https://github.com/openshift/release/blob/master/ci-operator/step-registry/openshift-qe/installer/bm/deploy/openshift-qe-installer-bm-deploy-ref.yaml#L45-L48).
> 
> So what I suggest:
> 
> * point to that PR for the test
> * get a successful rehearsal
> * delete the PR reference for the final merge

This was to allow the jetlag-ci to continue to function after merging that PR for SNOs. Since the sample vars file is not copied, but fully replaced, whenever a new var is added their will be ci/cpt breakage and no inheritance of the default value for the var.

akrzos:
/pj-rehearse pull-ci-redhat-performance-jetlag-main-deploy-sno

PR #62507, opened 2025-03-06
[o/cert-manager-operator] Update e2e-operator-tech-preview to use `workflow: generic-claim`

swghosh:
Using the generic-claim workflow in the e2e reduces the turnaround time for the e2e,
cluster(s) are assigned from an existing CI pool of already created clusters which saves cluster creation time,

This is being added only the tech-preview e2e for now.

xref: https://docs.ci.openshift.org/docs/how-tos/testing-operator-sdk-operators/ - a recommenced practice for layered operator e2e(s).


lunarwhite:
/close
In favor of #67268
Thanks for your work!

PR #62509, opened 2025-03-06
[th/dpu-operator-e2e-marvell] dpu-operator: add CI test for Marvell DPU

thom311:
This test is similar to "e2e-tests-workflow", which runs on a cluster with an Intel IPU. However, these run on another cluster having a Marvell DPU.

CC: @vrindle @SalDaniele


vrindle:
pj-rehearse

PR #62531, opened 2025-03-07
mirror new nginx-alpine for QE test

lihongan:
The nginx-alpine image is updated to 1.2.4 with renew certificates, we need to add it to the mirror list.

The old one 1.2.1 are still used by some team, we could remove it from the list later once all updated to 1.2.4.

cc @melvinjoseph86 @ShudiLi @rhamini3


lihongan:
/pj-rehearse ack

Only one network_edge case failed but not related to the mirror image
```
    exit status 1
    {
        Cmd: "oc --kubeconfig=/tmp/kubeconfig-3857269278 exec -n openshift-ingress router-ocp22636-6c69d6cf75-49vz4 -- bash -c cat haproxy.config | grep \"service-unsecure\"",
        StdErr: "command terminated with exit code 1",
```

PR #62545, opened 2025-03-07
skip recert jobs for productisation exclusive logic

rauhersu:


omertuc:
Do you _know_ whether always_run false necessary for this? Or just a guess because it makes sense intuitively?

omertuc:
Will lgtm when discussion above is resolved

rauhersu:
> Do you _know_ whether always_run false is necessary for this? Or just a guess because it makes sense intuitively?

I believe I can back up the intuitive meaning:

"Jobs that set always_run: false may be configured to run conditionally based on the contents of the pull request. run_if_changed and skip_if_only_changed accept a (Golang-style) [regular expression](https://golang.org/pkg/regexp/syntax/) which is run against the path of each changed file"

rauhersu:
> Alright, please let's make sure it doesn't disable the always_run functionality after we merge

You will be seeing your usual ci/prow + konflux jobs altogether from on (see snapshot). Note that 'skip_if_only_changed + run_always false' is a typical config when for example, you do not want to run your pipelines for README.md changes exclusively . So far we have not seen not reached out due to any weird behaviour in the rest of components we have already migrated with this same config.

Let us know of any issue in any case.

ci prow

ci-operator/config/rh-ecosystem-edge/recert/rh-ecosystem-edge-recert-main.yaml
```diff
@@ -84,6 +89,7 @@ tests:
     workflow: openshift-e2e-aws-single-node-recert
 - always_run: false
   as: e2e-aws-ovn-single-node-recert-openshift-e2e-test-qe
+  skip_if_only_changed: .*\.konflux/.*|\.tekton/.*
```
omertuc, line 92: This job was already `always_run: false`

```suggestion
```

omertuc (changes requested):

ci-operator/config/rh-ecosystem-edge/recert/rh-ecosystem-edge-recert-main.yaml
rauhersu: Typically, 'always_run: false' is set when we want the job to be run under a special criteria (hence, 'skip_if_only_changed', 'run_if_changed' appear as part of the config next).

What has been the criteria to set it to 'always_run: false' in the current configuration? At first glance, I am missing an extra condition for the execution. Is this because it is expected to run only when explicitly triggered manually / another config manages this ?

omertuc: > Is this because it is expected to run only when explicitly triggered manually

yes

PR #62569, opened 2025-03-10
classify upgrade ci failures

jiajliu:
The pr refines junit report in upgrade steps to classify upgrade ci failures with var UPGRADE_FAILURE_TYPE:

- overall: general upgrade failure(default)
- cvo: upgrade not start or post-upgrade check failure of cv.history
- admin_ack: failure on admin_ack patch
- rollback: rollback checkpoints failure
- node: failure on node healthcheck during upgrade
- mco: failure on mcp healthcheck during  upgrade
- cco: pre-upgrade steps failure for cloud-credentials update
- rhel: rhel node upgrade failure
- SLO(second-level operators): Upgrade stuck at operators and reported by cv conditions. It could be one operator(such as: ingress) or several operators(such as: authentication console)

Test [job](https://qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/qe-private-deck/pr-logs/pull/openshift_release/62569/rehearse-62569-periodic-ci-openshift-openshift-tests-private-release-4.18-amd64-rollback-stable-aws-ipi-byo-route53-f28/1899629762079363072): 

- for successful upgrade step `cucushift-upgrade-toimage`, the report will be like `<testcase classname="cluster upgrade" name="upgrade should succeed: overall"/>`
- for failed rollback step `cucushift-upgrade-rollback`, the report will be like `<testcase classname="cluster upgrade" name="upgrade should succeed: authentication console">` `<failure message="openshift cluster upgrade failed at authentication console"/>`

cc @jhou1 @jianlinliu to review, thanks

### DIFF
--- a/ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
+++ b/ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
@@ -132,10 +132,14 @@ function update_cloud_credentials_oidc(){
     mkdir "${preCredsDir}" "${tobeCredsDir}"
     # Extract all CRs from live cluster with --included
     if ! oc adm release extract --to "${preCredsDir}" --included --credentials-requests; then
-        echo "Failed to extract CRs from live cluster!" && exit 1
+        echo "Failed to extract CRs from live cluster!"
+        export UPGRADE_FAILURE_TYPE="oc_update"
+        return 1
     fi
     if ! oc adm release extract --to "${tobeCredsDir}" --included --credentials-requests "${TARGET}"; then
-        echo "Failed to extract CRs from tobe upgrade release payload!" && exit 1
+        echo "Failed to extract CRs from tobe upgrade release payload!"
+        export UPGRADE_FAILURE_TYPE="oc_update"
+        return 1
     fi
 
     # TODO: add gcp and azure
@@ -148,14 +152,17 @@ function update_cloud_credentials_oidc(){
         case "${platform}" in
         "AWS")
             if [[ ! -e ${SHARED_DIR}/aws_oidc_provider_arn ]]; then
-                echo "No aws_oidc_provider_arn file in SHARED_DIR" && exit 1
+		echo "No aws_oidc_provider_arn file in SHARED_DIR"
+		return 1
             else
                 export AWS_SHARED_CREDENTIALS_FILE="${CLUSTER_PROFILE_DIR}/.awscred"
                 infra_name=${NAMESPACE}-${UNIQUE_HASH}
                 oidc_provider=$(head -n1 ${SHARED_DIR}/aws_oidc_provider_arn)
-                extract_ccoctl
+                extract_ccoctl || { export UPGRADE_FAILURE_TYPE="cloud-credential"; return 1; }
                 if ! ccoctl aws create-iam-roles --name="${infra_name}" --region="${LEASED_RESOURCE}" --credentials-requests-dir="${tobeCredsDir}" --identity-provider-arn="${oidc_provider}" --output-dir="${toManifests}"; then
-                    echo "Failed to update iam role!" && exit 1
+		    echo "Failed to update iam role!"
+		    export UPGRADE_FAILURE_TYPE="cloud-credential"
+		    return 1
                 fi
                 if [[ "$(ls -A ${toManifests}/manifests)" ]]; then
                     echo "Apply the new credential secrets."
@@ -164,8 +171,8 @@ function update_cloud_credentials_oidc(){
             fi
             ;;
         *)
-           echo "to be supported platform: ${platform}" 
-           ;;
+            echo "to be supported platform: ${platform}"
+            ;;
         esac
     fi
 }
@@ -323,7 +330,7 @@ function rhel_upgrade(){
     ansible-inventory -i "${SHARED_DIR}/ansible-hosts" --list --yaml
     echo -e "\nRunning RHEL worker upgrade"
     sed -i 's|^remote_tmp.*|remote_tmp = /tmp/.ansible|g' /usr/share/ansible/openshift-ansible/ansible.cfg
-    ansible-playbook -i "${SHARED_DIR}/ansible-hosts" /usr/share/ansible/openshift-ansible/playbooks/upgrade.yml -vvv
+    ansible-playbook -i "${SHARED_DIR}/ansible-hosts" /usr/share/ansible/openshift-ansible/playbooks/upgrade.yml -vvv || { export UPGRADE_FAILURE_TYPE="rhel"; return 1; }
 
     if [[ "${UPGRADE_RHEL_WORKER_BEFOREHAND}" == "triggered" ]]; then
         echo -e "RHEL worker upgrade completed, but the cluster upgrade hasn't been finished, check the cluster status again...\    n"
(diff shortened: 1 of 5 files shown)


jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.18-amd64-rollback-stable-aws-ipi-byo-route53-f28 periodic-ci-openshift-openshift-tests-private-release-4.18-amd64-nightly-4.18-upgrade-from-stable-4.17-azure-ipi-proxy-workers-rhcos-rhel8-f28 periodic-ci-openshift-openshift-tests-private-release-4.19-multi-stable-4.19-upgrade-from-stable-4.16-gcp-ipi-ovn-ipsec-arm-mixarch-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.18-amd64-rollback-stable-aws-ipi-byo-route53-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.19-multi-stable-4.19-upgrade-from-stable-4.16-gcp-ipi-ovn-ipsec-arm-mixarch-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.18-amd64-nightly-4.18-upgrade-from-stable-4.17-azure-ipi-proxy-workers-rhcos-rhel8-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.18-amd64-rollback-stable-aws-ipi-byo-route53-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.18-amd64-rollback-stable-aws-ipi-byo-route53-f28

jiajliu:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.18-amd64-rollback-stable-aws-ipi-byo-route53-f28

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
```diff
@@ -1033,7 +1083,9 @@ for target in "${TARGET_RELEASES[@]}"; do
         cco_annotation
     fi
     if [[ "${UPGRADE_CCO_MANUAL_MODE}" == "oidc" ]]; then
-        update_cloud_credentials_oidc
+        if ! update_cloud_credentials_oidc; then
```
jianlinliu, line 1086: Before this change, any non-zero return will lead the whole script exit as failure, but now seem like, the error was caught, and move forwards.

jianlinliu: BTW, there is some explicit exit in `update_cloud_credentials_oidc`, shall we set `UPGRADE_FAILURE_TYPE` before the exit inside the function ?

jiajliu: Good catch. Because there are more than one exit in `update_cloud_credentials_oidc`, so I add the check outside of the function to avoid duplicated codes inside the func. I just updated to add `exit` code after setting UPGRADE_FAILURE_TYPE, the same with `rhel_post_upgrade`. PTAL again.

jianlinliu: Shall we also set `UPGRADE_FAILURE_TYPE` before `exit` inside `update_cloud_credentials_oidc` ?

jiajliu: do u mean adding duplicated setting in each exit switch inside the function? it sounds like not necessary or do u have any other concern?

ci-operator/step-registry/cucushift/upgrade/rhel-worker/cucushift-upgrade-rhel-worker-commands.sh
```diff
@@ -330,13 +368,17 @@ export SOURCE_VERSION
...
-
+# Set genenral upgrade ci failure to overall as default
+export UPGRADE_FAILURE_TYPE="overall"
 if [[ $(oc get nodes -l node.openshift.io/os_id=rhel) != "" ]]; then
     run_command "oc get node -owide"
     rhel_repo
     rhel_pre_upgrade
     rhel_upgrade
-    rhel_post_upgrade
+    if ! rhel_post_upgrade; then
```
jianlinliu, line 378: Can you help update the line 262 to append `|| return 1` to ensure the failure can be caught.

jianlinliu: Similarly, for `update_cloud_credentials_oidc`, once the function is used in a `if` condition, any important lines in the function failed, need a explicit `return`, so that it can be caught by the `if` condition. So I guess we also need to append some `|| return 1` inside `update_cloud_credentials_oidc`

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
jianlinliu: Here is some local testing:
```
[jianlinliu@jialiu-laptop release]$ cat test.sh 
#!/bin/bash

set -o nounset
set -o errexit
set -o pipefail

function test1() {
    echo "testing"
    #ls /xxxx || return 1
    exit 1
    echo "done"
}

#test1
if ! test1; then
    echo "fail"
fi
[jianlinliu@jialiu-laptop release]$ sh -x test.sh 
+ set -o nounset
+ set -o errexit
+ set -o pipefail
+ test1
+ echo testing
testing
+ exit 1
```

So that means once the function execute `exit` line, there is no chance to set `UPGRADE_FAILURE_TYPE` in this `if` block. The `UPGRADE_FAILURE_TYPE` will use the default value - `overall`. Of course, if the default value is the expected type, that is not a problem.

jianlinliu: That is why generally I suggest to only use `return` inside function rather than `exit`

jianlinliu: Right, so the key point here is if we want to explicitly set  `UPGRADE_FAILURE_TYPE` for the exit lines inside `update_cloud_credentials_oidc` function. If we want those exit lines to use `overall` as `UPGRADE_FAILURE_TYPE`, ignore the comments.

jiajliu: yeah, to identify the exit line in the code is the key point. currently we only add explicit failure type for those identified exit(such as `return 1`), for unclear exit, whether inside or outside function, it makes no difference. but there are code duplicates to add `export UPGRADE_FAILURE_TYPE=xxx` everywhere.  btw, for better understanding, we could have a huddle when you have time @jianlinliu wdyt?

jiajliu: updated `exit` to `return` in `update_cloud_credentials_oidc`

ci-operator/step-registry/cucushift/upgrade/rhel-worker/cucushift-upgrade-rhel-worker-commands.sh
jiajliu: add || return 1 in both update_cloud_credentials_oidc and rhel_post_upgrade to help catch important failure.

jiajliu:
@jianlinliu after more tests locally, I found return 1 + exit outside still has potential problem, your original suggestion should be more reliable, so I move the failure setting inside to just behind each expected exit, please review again, thanks.

PR #62578, opened 2025-03-10
Use custom variable instead of the Prow one to choose the Kustomization branch to use for deploying the MTO

aleskandro:


aleskandro:
/pj-rehearse ack

The failure is unrelated

PR #62609, opened 2025-03-10
restore original 'always_run' values

DennisPeriquet:
In this [PR](https://github.com/openshift/release/pull/62431), several values were set to `true` which means more tests were running when they were not before.

This PR attempts to mitigate that.

### DIFF
--- a/ci-operator/jobs/openshift-priv/machine-config-operator/openshift-priv-machine-config-operator-main-presubmits.yaml
+++ b/ci-operator/jobs/openshift-priv/machine-config-operator/openshift-priv-machine-config-operator-main-presubmits.yaml
@@ -234,7 +234,7 @@ presubmits:
           secretName: result-aggregator
     trigger: (?m)^/test( | .* )e2e-agent-compact-ipv4,?($|\s.*)
   - agent: kubernetes
-    always_run: true
+    always_run: false
     branches:
     - ^main$
     - ^main-
@@ -400,7 +400,7 @@ presubmits:
           secretName: result-aggregator
     trigger: (?m)^/test( | .* )e2e-aws-ovn,?($|\s.*)
   - agent: kubernetes
-    always_run: false
+    always_run: true
     branches:
     - ^main$
     - ^main-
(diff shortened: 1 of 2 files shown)


yuqi-zhang (approved):
Thanks! This should be helpful for a lot of MCO PRs that are blocked.

As a follow up question, should all always_run false jobs be optional? For example,
https://github.com/openshift/machine-config-operator/pull/4907#issuecomment-2707911323
It lists `4.12-upgrade-from-stable-4.11-images` as "required". Now, this probably should have been removed from the main presubmits since it should just be a 4.12 thing, but this (and some other jobs) are now listed as required, which I presume is because of a lack of optional tag?

DennisPeriquet:
> Thanks! This should be helpful for a lot of MCO PRs that are blocked.
> 
> As a follow up question, should all always_run false jobs be optional? For example, [openshift/machine-config-operator#4907 (comment)](https://github.com/openshift/machine-config-operator/pull/4907#issuecomment-2707911323) It lists `4.12-upgrade-from-stable-4.11-images` as "required". Now, this probably should have been removed from the main presubmits since it should just be a 4.12 thing, but this (and some other jobs) are now listed as required, which I presume is because of a lack of optional tag?

Hi @yuqi-zhang , the `always_run` values  must've been set for a particular reason some time ago. When the openshift/release scripts run (e.g., 'make jobs') the `always_run` values in the `jobs` subdir get changed. I'm guessing someone changed those manually and the 'make jobs' stuff just overwrites the values. But to answer your question, I would think that if a job is not required, it should be set as optional.

yuqi-zhang:
Thanks, we will re-evaluate those.

As a follow up question, it looks like the MCO repo's existing tests still think of those as required. I guess we need to explicitly set that (or override) so PRs that did run the test can pass

DennisPeriquet:
> Thanks, we will re-evaluate those.
> 
> As a follow up question, it looks like the MCO repo's existing tests still think of those as required. I guess we need to explicitly set that (or override) so PRs that did run the test can pass

This [PR](https://github.com/openshift/release/pull/62754) sets any `always_run: false` to optional.

PR #62638, opened 2025-03-11
Remove workaround for metal zone

sgoveas:
The issue was with network acls and it is now resolved through an internal ticket. We can remove the workaround made with PRs #58727 and #58781


sgoveas:
@liangxia please review

PR #62653, opened 2025-03-11
Update cypress image

jhadvig:
Updating the `console-plugin-test-cypress` image mirroring to expose both node16 and node 18 images.

/assign @spadgett 

@vikram-raj FYI


vikram-raj (commented):
/lgtm

PR #62714, opened 2025-03-12
[DTK] Removing the `RHEL_VERSION` build-arg.

ybettan:
DTK's Dockerfile, will deduce this value from the `/etc/os-release` file in the container if it was not set as a build-arg during build time.

There is no reason to set it, especially not to a wrong value since the underlying image is using RHEL 9.4 (and not 9.2 anyway).

---

/pj-rehearse auto-ack
/cc @joepvd


ybettan:
/pj-rehearse pull-ci-openshift-driver-toolkit-release-4.20-e2e-aws

PR #62725, opened 2025-03-12
openshift-logging: Use regex for substituting Loki image

xperimental:
This should remove one step in updating Loki in the future.

/cc @JoaoBraveCoding


JoaoBraveCoding (approved):
/lgtm

PR #62726, opened 2025-03-12
Update URL for kpi reporting server

fontivan:
The url of the kpi reporting server has been updated, so update the link here accordingly


natifridman:
Hi @fontivan, this job is using the old Jenkins job we have. I think we should remove it eventually. I'm not sure if it's going to pass the rehearsal

fontivan:
> Hi @fontivan, this job is using the old Jenkins job we have. I think we should remove it eventually. I'm not sure if it's going to pass the rehearsal

is that to say I should ignore the rehearse failure or just close the PR @natifridman ?

PR #62743, opened 2025-03-12
turn off cloudgov notifs

afcollins:
4 jobs at twice a day is a lot of extra noise for us.

Disabling notifs for now.


afcollins:
Oops. Already handled by #62685.
Closed.

PR #26639, opened 2022-03-01
add cincinnati tests to graph-data

PratikMahajan:

PR #62745, opened 2025-03-12
Only run tests on internal-channels/candidate.yaml changes

sdodson:
Most of these tests verify things that once verified should not require reverification when promoting to other channels.

### DIFF
--- a/ci-operator/jobs/openshift/cincinnati-graph-data/openshift-cincinnati-graph-data-master-presubmits.yaml
+++ b/ci-operator/jobs/openshift/cincinnati-graph-data/openshift-cincinnati-graph-data-master-presubmits.yaml
@@ -15,7 +15,7 @@ presubmits:
       pj-rehearse.openshift.io/can-be-rehearsed: "true"
     name: pull-ci-openshift-cincinnati-graph-data-master-e2e
     rerun_command: /test e2e
-    run_if_changed: internal-channels/candidate.yaml
+    run_if_changed: internal-channels/candidate.yaml|graph-data.rs/.*
     spec:
       containers:
       - args:
@@ -253,7 +253,7 @@ presubmits:
       pj-rehearse.openshift.io/can-be-rehearsed: "true"
     name: pull-ci-openshift-cincinnati-graph-data-master-validate-blocked-edges
     rerun_command: /test validate-blocked-edges
-    run_if_changed: blocked-edges/*
+    run_if_changed: hack/validate-blocked-edges.py|blocked-edges/.*
     spec:
       containers:
       - args:


ci-operator/config/openshift/cincinnati-graph-data/openshift-cincinnati-graph-data-master.yaml
```diff
@@ -53,6 +54,7 @@ tests:
     timeout: 1h0m0s
     version: "4.14"
   optional: true
+  run_if_changed: internal-channels/candidate.yaml
```
sdodson, line 57: @wking This is the one I was least certain about running only on candidate changes, does this scan the entire registry? I'd really like to avoid scanning all quay.io tags in these tests whenever possible provided we're not creating a giant blind spot

wking: `e2e-latest-cincinnati` does scan the registry, while it checks to see if modern Cincinnati [can still serve a valid graph from the proposed graph-data][1].  It's `optional: true`, so maybe we should move it to a periodic or something?  I'm not clear on what sorts of graph-data changes we'd make that might break this.  But because it's optional and I don't really know what it's looking for, I'm fine with `run_if_changed: ${ANYTHING}` to save Quay load by running less often.  #26639 added the presubmit, but doesn't describe why.  CC @PratikMahajan @DavidHurta, who have a better grasp than me on what `e2e-latest-cincinnati` is trying to protect us from.

[1]: https://github.com/openshift/cincinnati/blob/074fbe1271a741d61ea15064dd0532a7354284c4/hack/e2e.sh#L3-L6

ci-operator/config/openshift/cincinnati-graph-data/openshift-cincinnati-graph-data-master.yaml
```diff
@@ -87,6 +89,7 @@ tests:
     exec hack/validate-blocked-edges.py
   container:
     from: content-checks
+  run_if_changed: blocked-edges/*
```
wking, line 92: also if the script changes, so `run_if_changed: hack/validate-blocked-edges.py|blocked-edges/*` or some such.

ci-operator/config/openshift/cincinnati-graph-data/openshift-cincinnati-graph-data-master.yaml
```diff
@@ -44,6 +44,7 @@ tests:
   commands: /usr/local/bin/cincinnati-graph-data
   container:
     from: graphdata-unittests
+  run_if_changed: internal-channels/candidate.yaml
```
wking, line 47: Also if the Rust unit-test code changes, so `run_if_changed: internal-channels/candidate.yaml|graph-data.rs/.*` or some such.

wking (approved):
I'm fine with this merging as it stands, but also holding in case you want to wait for [the scope of `e2e-latest-cincinnati`][1] to get sorted.

/lgtm
/hold

[1]: https://github.com/openshift/release/pull/62745#discussion_r1992054892

ci-operator/config/openshift/cincinnati-graph-data/openshift-cincinnati-graph-data-master.yaml
DavidHurta: > I'm not clear on what sorts of graph-data changes we'd make that might break this. But because it's optional and I don't really know what it's looking for, I'm fine with run_if_changed: ${ANYTHING} to save Quay load by running less often.

I agree. However, I am quite new to the Cincinnati project, so take my opinion with a grain of salt.

PR #62754, opened 2025-03-12
set optional: true for all always_run: false for machine-config-operator

DennisPeriquet:
where I saw `always_run: false`, I explicitly added `optional: true` to avoid having extra tests required to pass CI.

### DIFF
--- a/ci-operator/jobs/openshift/machine-config-operator/openshift-machine-config-operator-main-presubmits.yaml
+++ b/ci-operator/jobs/openshift/machine-config-operator/openshift-machine-config-operator-main-presubmits.yaml
@@ -88,6 +88,7 @@ presubmits:
       ci.openshift.io/generator: prowgen
       pj-rehearse.openshift.io/can-be-rehearsed: "true"
     name: pull-ci-openshift-machine-config-operator-main-4.12-upgrade-from-stable-4.11-images
+    optional: true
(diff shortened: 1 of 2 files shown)


ci-operator/jobs/openshift/machine-config-operator/openshift-machine-config-operator-main-presubmits.yaml
```diff
@@ -1670,6 +1674,7 @@ presubmits:
       ci.openshift.io/generator: prowgen
       pj-rehearse.openshift.io/can-be-rehearsed: "true"
     name: pull-ci-openshift-machine-config-operator-main-e2e-gcp-op
+    optional: true
```
yuqi-zhang, line 1677: Hmm, something seems weird, this test (plus some other ones in this list) should be required.

I'm auditing the tests now and will open a follow up.

DennisPeriquet:
hm... when I run `make jobs`, several `optional: true` lines get removed.  That's why the `ci/prow/generated-config` CI job failed. I don't think this PR will go anywhere because of that.

yuqi-zhang:
Noticed the same via https://github.com/openshift/release/pull/62757 which I think is closer to the actual state that MCO wants these jobs to be in.

When running make jobs it resets a lot of `always_run` flags, plus the 4.11->4.12 jobs. Could you point me to where those are being reset?

PR #62757, opened 2025-03-12
Clean up mco presubmit jobs

yuqi-zhang:
Remove cluster-bootimages. This is old and probably hasn’t been working for awhile, plus we are working to bring in bootimage updates, and as such this test will be obsolete soon.

Remove 4.12-upgrade-from-stable-4.11-e2e-aws-ovn-upgrade/4.12-upgrade-from-stable-4.11-images. This is only for the 4.12 branch.

Keep e2e-aws-ovn, e2e-aws-ovn-upgrade, e2e-gcp-op, e2e-gcp-op-single-node, e2e-hypershift, ci/prow/images to always_run: false, but these are required tests that will almost always run (minus doc changes).

Set e2e-vsphere-ovn-upi,e2e-vsphere-ovn-upi-zones, e2e-vsphere-ovn-zones to always_run: false. These are optional, non-critical tests that aren’t monitored.

Set okd-images, okd-scos-images to optional: true. I don’t believe these are being monitored.

~~Set security tests to non-optional.~~ Kept optional in par with other repos

Questions at this time:
1. running `make jobs` resets some of the above fields. What is causing that?
2. For most changes I only touched `main`. Does 4.19/4.20 also need to be updated in the same manner?


djoshy:
/lgtm
/approve

thanks for trawling through this @yuqi-zhang

PR #62765, opened 2025-03-13
telcov10: clone main branch if exists else use master for NTO git repo

mrniranjan:
checkout main branch if exist else use master branch when cloning NTO git repository


rshemtov13:
/lgtm

Thanks!

PR #62771, opened 2025-03-13
NE-1954: ingress-operator: Enable GatewayAPIController in e2e-aws-gatewayapi job

alebedev87:
This PR adds the `GatewayAPIController` feature gate to the optional `e2e-aws-gatewayapi` job in the cluster-ingress-operator. This ensures that all test cases of the `TestGatewayAPI` e2e test are executed.

### DIFF
--- a/ci-operator/config/openshift/cluster-ingress-operator/openshift-cluster-ingress-operator-master.yaml
+++ b/ci-operator/config/openshift/cluster-ingress-operator/openshift-cluster-ingress-operator-master.yaml
@@ -132,7 +132,7 @@ tests:
   steps:
     cluster_profile: aws
     env:
-      FEATURE_GATES: '[GatewayAPI=true,ClusterAPIInstallAWS=true]'
+      FEATURE_GATES: '[GatewayAPI=true,GatewayAPIController=true,ClusterAPIInstallAWS=true]'
       FEATURE_SET: CustomNoUpgrade
     test:
     - as: test


ci-operator/config/openshift/cluster-ingress-operator/openshift-cluster-ingress-operator-master.yaml
```diff
@@ -132,7 +132,7 @@ tests:
   steps:
     cluster_profile: aws
     env:
-      FEATURE_GATES: '[GatewayAPI=true,ClusterAPIInstallAWS=true]'
+      FEATURE_GATES: '[GatewayAPI=true,GatewayAPIController=true,ClusterAPIInstallAWS=true]'
```
lihongan, line 135: `ClusterAPIInstallAWS=true` could be removed, it was added in 4.16 to temporarily workaround Cluster API install issue on AWS.

alebedev87: I can do it in another PR, to avoid any impact on the ongoing development.

Miciah: I think that Hongan is saying that enabling `ClusterAPIInstallAWS` is a no-op, so keeping or removing it wouldn't impact ongoing development one way or the other.  Removing it would clean up some old cruft.  Do you still prefer to leave it for a separate PR?

Miciah:
We want to update `openshift-cluster-ingress-operator-release-4.19.yaml` too, but not `openshift-cluster-ingress-operator-release-4.18.yaml`.  Will the 4.19 job definition be updated automatically?  Can we prevent the 4.18 job definition from being updated automatically?

candita:
> We want to update `openshift-cluster-ingress-operator-release-4.19.yaml` too, but not `openshift-cluster-ingress-operator-release-4.18.yaml`. Will the 4.19 job definition be updated automatically? Can we prevent the 4.18 job definition from being updated automatically?

We expect the 4.19 job to happen automatically and the 4.18 job to be ignored.

PR #62772, opened 2025-03-13
MGMT-20125: Move all test-infra baremetal peridoc jobs to use OFCIR

danmanor:
Except `arm64` job which is not ready to move yet. Explicitly set the machine size for clarity even when using the default (`medium`)


danmanor:
/pj-rehearse periodic-ci-openshift-assisted-test-infra-master-e2e-metal-assisted-4-15-periodic

PR #62778, opened 2025-03-13
refreshed all jobs to reflect latest configs

miyadav:
@JoelSpeed @theobarberbany @sunzhaohua2 @huali9 PTAL .
The yaml-lint config was not present yet its test were running on presubmits and failing , hence refreshed the jobs again to reflect latest config .


ci-operator/jobs/openshift/machine-api-operator/OWNERS
```diff
@@ -1,23 +0,0 @@
-# DO NOT EDIT; this file is auto-generated using https://github.com/openshift/ci-tools.
```
theobarberbany, line 1: I don't think we want to delete this :)

miyadav: Thanks @theobarberbany added back

PR #62780, opened 2025-03-13
kni: telco: mark Konflux jobs as optional

shajmakh:
Until konflux migration is stabilized, mark Konflux CI jobs as optional to unblock the PRs on the Telco KNI repos.

PR #62792, opened 2025-03-13
kni: dpdk: mark Konflux jobs as optional

shajmakh:
Until konflux migration is stabilized, mark Konflux CI jobs as optional to unblock the PRs.

### DIFF
--- a/core-services/prow/02_config/_config.yaml
+++ b/core-services/prow/02_config/_config.yaml
@@ -618,6 +618,11 @@ tide:
             skip-unknown-contexts: true
           release:
             skip-unknown-contexts: true
+      openshift-kni:
+        repos:
+          dpdk-base:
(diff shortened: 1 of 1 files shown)


cgoncalves:
Is the dpdk-base Konflux CI blocking any other project?
In its own repository (https://github.com/openshift-kni/dpdk-base), Konflux is the only CI we have today.

shajmakh:
> Is the dpdk-base Konflux CI blocking any other project?

nope, but this is filed to align it with the rest of the telco repos https://github.com/openshift/release/pull/62780. If you don't see the need for this as it is not blocking I can drop it.

cgoncalves:
I don't think it is necessary. Thank you very much nonetheless for proposing this PR!

PR #62796, opened 2025-03-13
Following up #62183 to disable payload jobs on etcd repo

afcollins:
#62183 updated `always_run: false`, but this wasn't rendered into the job configs.

Payload runs from this repo don't need to be run on every commit. Only when requested.


afcollins:
@tjungblu @dusk125 If you would please lgtm and approve

PR #62826, opened 2025-03-14
reduce 3-hops cpou test frequency

jiajliu:
3-hops is not committed to support even in 4.19 based on latest confirm, so let's reduce the test frequency for an internal test only with `f60`

cc @JianLi-RH to help review.


jiajliu:
/pj-rehearse ack

cc @jhou1 to help approve since liangxia take vacation.

PR #62894, opened 2025-03-17
config/openshift/os: set OPENSHIFT_CI=1 for extensions image builds

jlebon:
We explicitly require this build arg now to trigger the logic that fetches the mirrored RPM repos. This mirrors (pun intended) how the node image is built.


jbtrystram (approved):
/lgtm
/approve

PR #62923, opened 2025-03-17
Update `boilerplate` config to enable a Go 1.23 image

joshbranham:
My attempt here is to setup a new Go 1.23 image that can be referenced in `openshift/boilerplate` as the CI image as I work on upgrading that repo to Go 1.23.

I believe this worked, although the rehearsals failed. I _think_ those are expected?

### DIFF
--- a/ci-operator/config/openshift/boilerplate/openshift-boilerplate-master.yaml
+++ b/ci-operator/config/openshift/boilerplate/openshift-boilerplate-master.yaml
@@ -3,6 +3,10 @@ base_images:
     name: builder
     namespace: ocp
     tag: rhel-8-golang-1.22-openshift-4.17
+  ocp_builder_rhel-8-golang-1.23-openshift-4.19:
+    name: builder
+    namespace: ocp
+    tag: rhel-8-golang-1.23-openshift-4.19
 build_root:
   from_repository: true
 images:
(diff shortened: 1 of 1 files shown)


ci-operator/config/openshift/boilerplate/openshift-boilerplate-master.yaml
```diff
@@ -12,6 +16,9 @@ images:
     ocp_builder_rhel-8-golang-1.22-openshift-4.17:
       as:
       - brew.registry.redhat.io/rh-osbs/openshift-golang-builder:rhel_8_1.22
+    ocp_builder_rhel-8-golang-1.23-openshift-4.19:
+      as:
+      - brew.registry.redhat.io/rh-osbs/openshift-golang-builder:rhel_8_1.23
```
danilo-gemoli, line 21: `brew.registry.redhat.io/rh-osbs/openshift-golang-builder:rhel_8_1.22` doesn't appear in the Dockerfile anymore. You can remove it and keep `rhel_8_1.23` only.

joshbranham: @danilo-gemoli so maybe I can share my goal and you can tell me if this is possible.

- In https://github.com/openshift/boilerplate/pull/503 I am trying to test if Go 1.23 works as expected. I want to see CI pass on the branch.
- In this repo, I am trying to add the new image I need to run the test suite with an image that is Go 1.23.
- I want to keep Go 1.22 configured here so that I can easily revert/not merge should something not work, while allowing others to keep working on branches with Go 1.22 and tests on master still pass.

Based on what you are telling me this seems not possible?

danilo-gemoli: Ok, if that's the case you can have both and eventually keep tuning the Containerfile in your target repository.

PR #62953, opened 2025-03-18
Improvement for copying integration test artifacts

ctartici:
This improvement will let us collect midstream istio integration test logs even if the tests fail.


FilipB (changes requested):
I think it makes sense to add that for all tests, e.g. integ-helm-test-run is missing that

ctartici:
> I think it makes sense to add that for all tests, e.g. integ-helm-test-run is missing that

Yes you are right, I have missed it. just added it. Thanks

FilipB (approved):

PR #62963, opened 2025-03-18
Chaos slack with config prowgen

paigerube14:
Making the names of jobs similar to use the prow gen config to not have to manually add the slack reporting to each job


shahsahil264 (approved):
LGTM

PR #62971, opened 2025-03-19
Apply latest hypershift operator to hosted-mgmt cluster daily

liangxia:


ci-operator/jobs/infra-periodics.yaml
```diff
@@ -119,6 +119,52 @@ periodics:
...
+  spec:
+    containers:
+    - args:
+      - -c
+      - |
+        set -o errexit
+        set -o nounset
+        set -o pipefail
+        export MGMT_AWS_CONFIG_PATH='/tmp/fake.aws'
+        echo -e "[default]\naws_access_key_id=AKKKKKKKKKKKKKKKKKKK\naws_secret_access_key=Hasdasdsadasdadsadsadsadk" > "$MGMT_AWS_CONFIG_PATH"
```
droslean, line 147: You don't need to do that on the fly. You can just mount a secret in that path directly.

bear-redhat: this is a fake credential

PR #62985, opened 2025-03-19
Build image make which contains command make

liangxia:
Build a image `make` from 4.18 ocp base, which contains `make`, `podman` and `yq`, and it will be used in https://github.com/openshift/release/pull/62971


liangxia:
Will try to get rid of requirement of podman and the new image stuff.

/close

PR #62996, opened 2025-03-19
oran: Add mirror directives for oran poc branches (#62920)

donpenney:


donpenney:
Verified that the new image tags have been promoted to the CI registry, so the new mirror directives should be safe.

PR #63021, opened 2025-03-20
cluster-ingress-operator: GatewayAPI support is now GA

alebedev87:
`GatewayAPI` and `GatewayAPIController` feature gates have been promoted to GA, making Gateway API support part of the default feature set. This PR removes `e2e-aws-gatewayapi` job because the gateway API tests are now part of the operator test jobs. Also it removes TechPreviewNoUpgrade feature set from `e2e-aws-gatewayapi-conformance` job.
Also this PR makes `e2e-aws-gatewayapi-conformance` job "always running" to enable continuous testing of the GatewayAPI implementation.


candita:
> util_gatewayapi_test.go:461: failed to get deployment openshift-operators/istio-operator, retrying...
>     gateway_api_test.go:89: failed to find expected Istio operator: error finding deployment openshift-operators/istio-operator: context deadline exceeded

/pj-rehearse

candita:
Let's figure out why the conformance tests don't work before we make them required.
/hold

alebedev87:
> Let's figure out why the conformance tests don't work before we make them required.

The conformance test will remain optional. I only changed `always_run` for it, so that it'll be run for every PR from cluster-ingress-operator repository. Even if it fails (which needs to be investigated, I agree), it won't block PRs.

alebedev87:
Holding the PR until [the problem with install plan approval](https://redhat-internal.slack.com/archives/CBWMXQJKD/p1742421080814679) is fixed.

lihongan:
Can we unhold this now ?

alebedev87:
`e2e-aws-gatewayapi` won't be needed soon (after we promote both GWAPI featuregates). Let's keep it on hold and re-iterate after all the dust will be ssettled (after branch cut on 25APR).

ci-operator/config/openshift/cluster-ingress-operator/openshift-cluster-ingress-operator-master.yaml
```diff
@@ -143,8 +142,7 @@ tests:
         requests:
           cpu: 100m
     workflow: ipi-aws
-- always_run: false
-  as: e2e-aws-gatewayapi-conformance
+- as: e2e-aws-gatewayapi-conformance
```
candita, line 145: @alebedev87 would you mind updating L135 to TechPreviewNoUpgrade and/or removing the FEATURE_SET/FEATURE_GATES env variables from here when the time is right?

alebedev87: I removed `e2e-aws-gatewayapi` job because `TestGatewayAPI` is now part of the standard test suite (`e2e-*-operator` jobs). `e2e-aws-gatewayapi-conformance` is updated too: `TechPreviewNoUpgrade` featureste is removed.

alebedev87:
/hold

As a matter of fact, the branch cut is still not done. So, let's hold this one for some more time.

alebedev87:
/unhold

The branch cut is done this morning (Europe time).

alebedev87:
/pj-rehearse ack

The PR was rebased from `master`, no changes to the contents since [the last rehearsal](https://github.com/openshift/release/pull/63021#issuecomment-2835352819).

ci-operator/jobs/openshift/cluster-ingress-operator/openshift-cluster-ingress-operator-master-presubmits.yaml
```diff
@@ -1,78 +1,5 @@
...
-          secretName: manifest-tool-local-pusher
-      - name: pull-secret
-        secret:
-          secretName: registry-pull-credentials
-      - name: result-aggregator
-        secret:
-          secretName: result-aggregator
-    trigger: (?m)^/test( | .* )e2e-aws-gatewayapi,?($|\s.*)
   - agent: kubernetes
     always_run: false
```
lihongan, line 4: Just noted that the`always_run` is not updated by `make jobs`, after updating the `always_run:` to `true` manually then run `make jobs` again and it also doesn't revert the change. Looks something wrong with `make jobs` and we might need to update the field manually for `e2e-aws-gatewayapi-conformance` job.

alebedev87: Right, seems like [TP noticed this problem too](https://github.com/openshift/ci-tools/pull/4529). Once the fix for the ci-tools is merged they'll follow up with a [PR which updates all impacted presubmits](https://github.com/openshift/release/pull/64342).

Unfortunately I cannot do much here because the presubmits for this PR will fail if I'll manually change the value to `always_run: true` in the job definition. I'll let TP fix this in https://github.com/openshift/release/pull/64342.

PR #63034, opened 2025-03-20
Bump OCP and go for OSC 1.9.0

gkurz:
OSC 1.9.0 is bumping its OCP and go versions. Prepare CI for that.

### DIFF
--- a/ci-operator/config/openshift/sandboxed-containers-operator/openshift-sandboxed-containers-operator-devel.yaml
+++ b/ci-operator/config/openshift/sandboxed-containers-operator/openshift-sandboxed-containers-operator-devel.yaml
@@ -1,6 +1,6 @@
 base_images:
   base:
-    name: "4.17"
+    name: "4.18"
     namespace: ocp
     tag: base-rhel9
 binary_build_commands: go mod download && go mod vendor && GOFLAGS="" make build
@@ -8,7 +8,7 @@ build_root:
   image_stream_tag:
     name: builder
     namespace: ocp
-    tag: rhel-9-golang-1.22-openshift-4.17
+    tag: rhel-9-golang-1.23-openshift-4.19
 images:
 - from: base
   to: openshift-sandboxed-containers-operator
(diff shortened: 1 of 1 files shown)


ci-operator/config/openshift/sandboxed-containers-operator/openshift-sandboxed-containers-operator-release-1.9.yaml
```diff
@@ -54,6 +54,6 @@ tests:
       OO_TARGET_NAMESPACES: openshift-sandboxed-containers-operator
     workflow: optional-operators-ci-azure
 zz_generated_metadata:
-  branch: release-1.8
+  branch: release-1.9
```
littlejawa, line 57: We don't have a release-1.9 branch yet... Should we create it first?
Or... this is all about building from the devel branch. Should it be "devel" here?

I'm actually wondering what this block mean :-/

gkurz: We have two files, one for devel and one for the release branch. I'm bumping everything here so that we don't forget. I'm assuming the one for the release branch will be not be used until we create the release branch.

littlejawa: Oh ! Scratch that. You're editing the 1.8 build file, that makes our CI run on the release-1.8 branch.
So you're renaming/modifying it to run on an future release-1.9 branch.

As we're not pushing anything new on release-1.8, I'm fine moving this file away and reusing it for the next release.
Maybe prow will complain that the branch is not there though - let's see.

gkurz: My guess is that we should prepare prow first or it won't be able to run when we want to push our first PR to release-1.9...

littlejawa (approved):
/lgtm
Thanks @gkurz !

gkurz:
/pj-rehearse pull-ci-openshift-sandboxed-containers-operator-devel-ci-index-openshift-sandboxed-containers-operator-bundle

PR #63036, opened 2025-03-20
No krkn hub

paigerube14:
No longer need prow tests in krkn-hub, already had moved to prow-scripts instead


paigerube14:
@pruan-rht @Liangxia @smg247 PTAL

PR #63044, opened 2025-03-20
ci-tools: sippy-config-generator Makefile needs 'which'

sosiouxme:
it's [been failing](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-prow-auto-sippy-config-generator/1902661574049402880)


sosiouxme:
this change only touches TRT's image and job
/pj-rehearse skip

PR #63068, opened 2025-03-21
Run version compatibility tests on changes

janisz:
We found that version compatibility test is broken since this PR:
- https://github.com/stackrox/stackrox/pull/14637

To prevent this in the future let's run tests if `tests/upgrade/postgres_run.sh` is changed


janisz:
Finally this change was not relevant to failure while the tag was
https://github.com/stackrox/stackrox/blob/7657f34e71713fb1fe9ab55ed9ff431ecff4dbba/.openshift-ci/get_compatibility_test_tuples.py#L47

PR #63069, opened 2025-03-21
Fix branches stanza

psalajova:
This kind of regex isn't supported by github's [branch protection](https://docs.github.com/en/rest/branches/branch-protection?apiVersion=2022-11-28#update-branch-protection), and is causing `periodic-openshift-release-fast-forward` to fail ([latest run](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-openshift-release-fast-forward/1903060153570168832) for reference). This change should disable our branch protection.

### DIFF
--- a/core-services/prow/02_config/openshift-assisted/assisted-installer-ui/_prowconfig.yaml
+++ b/core-services/prow/02_config/openshift-assisted/assisted-installer-ui/_prowconfig.yaml
@@ -4,9 +4,11 @@ branch-protection:
       repos:
         assisted-installer-ui:
           branches:
-            ^release-.*:
-              protect: false
             ^releases/v.*:
+              protect: false
+            release-4.19:
+              protect: false
+            release-4.20:
               protect: false
           enforce_admins: true
           protect: true


droslean (approved):

core-services/prow/02_config/openshift-assisted/assisted-installer-ui/_prowconfig.yaml
```diff
@@ -4,10 +4,12 @@ branch-protection:
       repos:
         assisted-installer-ui:
           branches:
-            ^release-.*:
-              protect: false
             ^releases/v.*:
```
smg247, line 7: this won't work either, right?

psalajova: I think it won't, yeah. I left it there though, as currently nothing is being forwarded to those `^releases/v.*` branches anyway, and I'm currently talking with the owners and I'm not sure if they need branch protection at all, so maybe it will just get removed.

psalajova: slack thread: https://redhat-internal.slack.com/archives/CBN38N3MW/p1742286859084419

PR #63073, opened 2025-03-21
update konflux placeholder templates

Prucek:


ci-operator/config/openshift/ci-tools/openshift-ci-tools-master.yaml
```diff
@@ -782,6 +782,7 @@ images:
       - destination_dir: .
         source_path: /go/bin/helpdesk-faq
   to: helpdesk-faq
+operator: {}
```
droslean, line 785: Is this necessary?

Prucek: I would need to add it when I want to trigger the test, or else validation will fail, can you lgtm now?

droslean (approved):

PR #63094, opened 2025-03-23
MGMT-20190: Add job for testing Node Maintenance Operator installing with assisted-service

danmanor:
For testing - https://github.com/openshift/assisted-service/pull/7444


eliorerz (approved):
LGTM

PR #63107, opened 2025-03-24
ACM-18560: Fix the e2e tests for CAPK [main, 4.17, 4.18, 4.19]

nunnatsa:
The current e2e test is trying to install ODF v4.15, while this version is not found in newer versions of OCP.

This PR fixes it by overriting the default ODF channel, to the latest version for each version.


orenc1:
odf installation succeeded.
/lgtm

PR #63129, opened 2025-03-24
[kubevirt-common-templates] add e2e test config for common templates v0.33

fossedihelm:
add e2e test config for common templates v0.33


fossedihelm:
@0xFelix Rebased and executed `make jobs`

PR #63137, opened 2025-03-24
Vboulos Propagate acm213 images to Prow

vboulos:
Promote ACM 2.13 test images for the following components

CLC (UI)
CLC (Non-ui)
ALC
GRC
Observability


smithbw88 (approved):
LGTM

PR #63156, opened 2025-03-25
Pipelines-console-plugin: Inject oc binary from latest release payload to test container

vikram-raj:
`oc` command not found error is there in the Pipelines-console-plugin e2e. This PR inject the oc binary from latest release payload to pipelines-console-plugin test container.


lokanandaprabhu (approved):
/lgtm

PR #63159, opened 2025-03-25
MULTIARCH-5347: have a periodic job per OCP release for mto

lwan-wanglin:
we already have  periodic job for 4.18 and 4.19, add periodic job for 4.16 and 4.17 too.
For 4.18 and 4.19, it runs 4 times per month. For others, it runs once per month.


lwan-wanglin:
the jobs failure seems related to https://redhat-internal.slack.com/archives/C04N203SNUW/p1743031691046639?thread_ts=1743028042.959819&cid=C04N203SNUW

aleskandro (approved):

PR #63160, opened 2025-03-25
MGMT-20210: Create skeleton files for our automation

giladravid16:
These skeleton files will be used in our automation when creating new branches for our release versions.
It will copy the content of each file to a new file, but with the following replaces:
1. assisted-version-placeholder with the branch's name.
2. ocm-version-placeholder with the most recent ocm version.
3. ocp-version-placeholder with the most recent openshift version.
4. change all release versions with the most recent openshift version.

### DIFF
--- a/ci-operator/config/openshift/assisted-installer/openshift-assisted-installer-assisted-version-placeholder.yaml
+++ b/ci-operator/config/openshift/assisted-installer/openshift-assisted-installer-assisted-version-placeholder.yaml
@@ -1,0 +1,214 @@
+base_images:
+  assisted-image-service:
+    name: assisted-version-placeholder
+    namespace: edge-infrastructure
+    tag: assisted-image-service
+  assisted-installer-agent:
+    name: assisted-version-placeholder
+    namespace: edge-infrastructure
+    tag: assisted-installer-agent
+  assisted-service:
+    name: assisted-version-placeholder
+    namespace: edge-infrastructure
+    tag: assisted-service
+  assisted-service-index:
+    name: assisted-service-index
+    namespace: edge-infrastructure
+    tag: latest
+  assisted-service-scripts:
+    name: ocm-version-placeholder
+    namespace: edge-infrastructure
+    tag: assisted-service-scripts
+  assisted-test-infra:
+    name: ocm-version-placeholder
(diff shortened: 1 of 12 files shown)


adriengentil:
I think it would make the trick, at least the all the CI config is there in one place. @danmanor can you review as well and give the approval? I think we should review the jobs running in these branches, but we can do it later.
/cc @danmanor

giladravid16:
@danmanor this is part of a [MGMT-20085](https://issues.redhat.com/browse/MGMT-20085) where the goal is to have assisted-installer-cicd create a new branch when creating a new minor release of our SaaS (for example v2.39) for each backend component.
The new branches will need to be configured with both konflux and prow configurations.
This PR in particular is for creating skeleton files for the prow configuration, so that the automation will copy those files, replace "version-placeholder" with the branch's name, and then open a PR to add those files.

danmanor:
@giladravid16 So we will have a file for each version ?

giladravid16:
@danmanor yes a file per version per component

danmanor:
@giladravid16 Are we sure we want to go this way ? it means that there going to be enormous amount of CI files which we would need to maintain.

giladravid16:
@danmanor I don't think the number of files will get that big since we don't support multiple versions at the same time in SaaS. There will be times where we have 2 versions running (for example v2.39 in prod and v2.40 in stage), but there's no need to keep older versions that are not used in either stage or prod.

danmanor:
@giladravid16 So the automation will remove the old files ?

ci-operator/config/openshift/assisted-installer/openshift-assisted-installer-assisted-version-placeholder.yaml
```diff
@@ -0,0 +1,254 @@
...
+      product: ocp
+      stream: nightly
+      version: "4.18"
+resources:
+  '*':
+    requests:
+      cpu: 100m
+      memory: 200Mi
+tests:
+- as: publish-multi-arch-images
```
danmanor, line 67: We don't need any post submit jobs in these files, lets remove them

danmanor (changes requested):

PR #63172, opened 2025-03-25
OSSM-8612: syncing build-tools images automatically

FilipB:
quay-proxy.ci.openshift.org registry requires auth. To avoid creating additional pull secret and keeping the token up to date, we are using directly gcr.io. Update of this image must be tied with update of the root_image.
This is not a problem on hyperhift as it contains pull secrets for quay-proxy.ci.openshift.org by default.


ci-operator/config/maistra/test-infra/maistra-test-infra-main.yaml
```diff
@@ -99,7 +99,7 @@ tests:
     --token-path=/creds-github/token \
     '--title=Update maistra/istio builder image' \
     --modifier=update_istio_builder_image_master \
-    '--cmd=sed -i -e "s|\(gcr.io/istio-testing/build-tools\):master.*|\1:$NEW_IMAGE_MASTER|" -e "s|\(gcr.io/istio-testing/build-tools\):release-1\.24.*|\1:$NEW_IMAGE_1_24|" core-services/image-mirroring/_config.yaml' \
+    '--cmd=sed -i -e "s|\(gcr.io/istio-testing/build-tools\):master.*|\1:$NEW_IMAGE_MASTER|" -e "s|\(gcr.io/istio-testing/build-tools\):release-1\.24.*|\1:$NEW_IMAGE_1_24|" core-services/image-mirroring/_config.yaml ci-operator/config/openshift-service-mesh/sail-operator/*' \
```
jwendell, line 102: I think this change is useless as it will not match anything.

If I understand correctly you switched to a fixed url, so there's no need for this change?

FilipB: It will change other locations which are using `gcr.io/istio-testing`. The change of MAISTRA_BUILDER_IMAGE in this PR is not related. That job runs on hypershift where we don't need to use `gcr.io/istio-testing`.

FilipB: It has already created this PR https://github.com/openshift/release/pull/63175 which is correct

jwendell (approved):

PR #63192, opened 2025-03-26
Refine test frequency for several 4.16 tests

liangxia:
4.16.z releases are scheduled to release every 3 weeks after the first 4.18.z, so update the 4.16 tests accordingly ( change from `f7` to `f14` ).


jianlinliu:
I was planing to update the 4.16 ci jobs frequency in a separate PR when I was working on another PR, glad to see you already worked it out here. Thanks.

/lgtm

PR #63213, opened 2025-03-26
Fix race condition for volume creation in upi

Shilpa-Gokul:
As we have informing jobs being triggered for every release, both the libvirt and heavy build jobs get started at the same time. So there is a possibility of race condition where the volume creation might happen concurrently leading to "volume already exists" error


tvardema:
pj-rehearse ack

PR #63307, opened 2025-03-28
Remove 1.3 CI tests per EOL

josephca:
Remove 1.3 CI tests per EOL


josephca:
Closing it as it was handled by the same duplicate: https://github.com/openshift/release/pull/63320

PR #63318, opened 2025-03-31
Add jobs for Logging 6.1 full function test

kabirbhartiRH:
This PR:
1. Enables Logging 6.1 full function test on prow CI jobs with GCP cluster scheduled to be run with once a week on Mondays.

cc. @anpingli @QiaolingTang


kabirbhartiRH:
@anpingli Please review and let me know in case more jobs need to be added. IMO, one job per week should be ok for full function test since we most likely would run acceptance tests during zstream releases.

kabirbhartiRH:
This has been marked as Obsolete and will be taken up in a different task

PR #63335, opened 2025-03-31
`secret-manager` CLI tool

psalajova:
Initial implementation of the Secret Manager CLI tool. More PRs will follow to enhance/edit the functionality.

/cc @danilo-gemoli

### DIFF
--- a/hack/gcp-secret-manager/commands/create.py
+++ b/hack/gcp-secret-manager/commands/create.py
@@ -1,3 +1,6 @@
+# Ignore dynamic imports
+# pylint: disable=E0401, C0413
+
 import re
 import typing
 from typing import Dict
--- a/hack/gcp-secret-manager/commands/list.py
+++ b/hack/gcp-secret-manager/commands/list.py
@@ -1,3 +1,6 @@
+# Ignore dynamic imports
+# pylint: disable=E0401, C0413
+
 import json
 from typing import Dict, List
 
--- a/hack/gcp-secret-manager/util.py
+++ b/hack/gcp-secret-manager/util.py
@@ -1,3 +1,6 @@
+# Ignore dynamic imports
+# pylint: disable=E0401, C0413
+
 import re
 from typing import Dict, List
 
--- a/hack/gcp-secret-manager/commands/delete.py
+++ b/hack/gcp-secret-manager/commands/delete.py
@@ -1,3 +1,6 @@
+# Ignore dynamic imports
+# pylint: disable=E0401, C0413
+
 import click
 from google.api_core.exceptions import NotFound, PermissionDenied
 from google.cloud import secretmanager
--- a/hack/gcp-secret-manager/commands/get_service_account.py
+++ b/hack/gcp-secret-manager/commands/get_service_account.py
@@ -1,3 +1,6 @@
+# Ignore dynamic imports
+# pylint: disable=E0401, C0413
+
 import json
 
 import click
--- a/hack/gcp-secret-manager/commands/login.py
+++ b/hack/gcp-secret-manager/commands/login.py
@@ -1,3 +1,6 @@
+# Ignore dynamic imports
+# pylint: disable=E0401, C0413
+
 import subprocess
 
 import click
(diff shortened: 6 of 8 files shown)


hack/gcp-secret-manager/commands/list.py
```diff
@@ -0,0 +1,83 @@
...
+
+import click
+import requests
+import yaml
+from google.api_core.exceptions import PermissionDenied
+from google.cloud import secretmanager
+from util import CONFIG_PATH, PROJECT_ID, ensure_authentication, validate_collection
+
+
+@click.command("list")
```
danilo-gemoli, line 11: I would change the way secrets collection are presented when no collection has been specified by the user. Instead of a plain list:
```
collection_1
collection_2
...
collection_n
```
I would add the group information as well:
```
aos-qe:
- collection_1
- collection_2
art-admins:
- foo_collection
...
```

hack/gcp-secret-manager/commands/list.py
```diff
@@ -0,0 +1,83 @@
...
+@click.option(
+    "-o",
+    "--output",
+    type=click.Choice(["json", "text"], case_sensitive=False),
+    default="text",
+    help="Output format, defaults to plain text but can be set to 'json'. Only applicable when a collection is specified.",
+)
+@click.option("-c", "--collection", default="", help="Name of the secret collection")
```
danilo-gemoli, line 19: As per design document, it seems we have to consider also the `--group=foobar` or `-g foobar` flag. As an example
```sh
$ hack/secret-manager.sh list --group test-platform-ci-sudoers
test-platform-ci-sudoers:
- foo
- bar
...
```

hack/gcp-secret-manager/util.py
```diff
@@ -0,0 +1,67 @@
...
+
+import click
+from google.auth import default
+from google.auth.exceptions import DefaultCredentialsError
+
+# Test Platform's project in GCP Secret Manager
+PROJECT_ID = "openshift-ci-secrets"
+
+# The YAML config which defines which groups have access to which secret collections.
+CONFIG_PATH = "https://raw.githubusercontent.com/openshift/release/master/core-services/sync-rover-groups/_config.yaml"
```
danilo-gemoli, line 11: If a user is running this script, it means he has already checked out the `openshift/release` repository and, in turn, this implies that the `core-services/sync-rover-groups/_config.yaml` file is already available on the local filesystem. What about reading it from the filesystem instead of issuing a request to GitHub?

hack/gcp-secret-manager/commands/create.py
```diff
@@ -0,0 +1,143 @@
...
+        "\nProvide a short description of how this secret was originally requested "
+        "(e.g. links to service now tickets, Jira tickets, documentation).\n"
+        "This can help future team members know who to contact in case of problems.\n"
+        "Do not include sensitive information.\n"
+    )
+    annotations[REQUEST_INFO] = prompt_for_annotation(REQUEST_INFO)
+    return annotations
+
+
+def prompt_for_annotation(msg: str) -> str:
```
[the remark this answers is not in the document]
psalajova: According to https://cloud.google.com/secret-manager/docs/creating-and-managing-annotations the limitation is: "The total size of annotation keys and values must be less than 16KiB." I'll add the check

hack/gcp-secret-manager/util.py
psalajova: Hm, good point...

psalajova: The counter argument is that the user needs to have an up-to-date release repo locally

hack/gcp-secret-manager/util.py
```diff
@@ -0,0 +1,67 @@
...
+def get_secret_name(collection, name: str) -> str:
+    return f"{collection}__{name}"
+
+
+def validate(collection: str, name: str, from_file: str, from_literal: str):
+    ensure_authentication()
+    validate_collection(collection)
+    validate_secret_name(name)
+
+    if from_literal != "" and from_file != "":
```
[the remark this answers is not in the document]
psalajova: Can you describe how reading from `stdin` would work for the user?

hack/gcp-secret-manager/util.py
danilo-gemoli: We can keep this as it is, just something to keep in mind.

PR #63376, opened 2025-04-01
remove logging 5.6,5.7 jobs

anpingli:
We had kept 5.6 and 5.7 for a while,  now we can remove these jobs.


anpingli:
@Phaow  can you merge it?

PR #63422, opened 2025-04-02
OCPEDGE-1702: feat: add presubmit optional lanes for installer and origin

eggfoobar:
Added presubmit optional lanes for two node arbiter on the origin and installer repo.


eggfoobar:
The failures correct due to issues in the installer that are being fixed and these jobs will help catch sooner

/pj-rehearse ack

PR #63435, opened 2025-04-02
Reschedule deployments for OCP 4.20 version

ccardenosa:
Until a decision is made in regards to how many Hubs should be running at the same time and how to address such condition, this job is to be run less frequency since it is expected to fail. We are doing this instead of removing it to have it as a reference in the near future.


ci-operator/config/openshift-kni/eco-ci-cd/openshift-kni-eco-ci-cd-main__ci-4.20.yaml
```diff
@@ -28,7 +28,7 @@ resources:
 tests:
 - as: telcov10n-metal-single-node-hub-ztp
   cluster: build05
-  cron: '* 1 * * *'
+  cron: 0 22 1-7 * 5
```
natifridman, line 31: Should we do something like this if we don't want it to run?
```suggestion
  cron: 0 22 31 2 5
```

natifridman (approved):
/lgtm

PR #63444, opened 2025-04-02
add oadp-non-admin oadp-1.5 jobs

weshayutin:
messed up and missed this, thanks @kaovilai


kaovilai (changes requested):
Can you also add `quay.io/konveyor/oadp-non-admin:oadp-1.5` to this line?
https://github.com/openshift/release/blob/228c0ecbc8d656d18c1e0e3b10f55286ac214913/core-services/image-mirroring/konveyor/mapping_konveyor_latest#L39

sseago:
@weshayutin I think we're still missing branch protection for oadp-non-admin for oadp-1.5
https://github.com/openshift/release/blob/master/core-services/prow/02_config/migtools/oadp-non-admin/_prowconfig.yaml#L6-L9

kaovilai:
/hold
wes do you want to add branch protection. unhold if not.
https://github.com/openshift/release/blob/master/core-services/prow/02_config/migtools/oadp-non-admin/_prowconfig.yaml#L6-L9

PR #63448, opened 2025-04-02
add multi-pr presubmit testing

weshayutin:
https://docs.ci.openshift.org/docs/how-tos/multi-pr-presubmit-testing/ 
Not 100% sure this will be useful to OADP unless we can build the containers from the other repos and create the override
* @mpryc FYI.. some to chat about when ur back


mpryc:
/remove-lifecycle rotten

To be checked this week.

PR #63454, opened 2025-04-02
nac oadp-1.5 branch protection

kaovilai:


core-services/prow/02_config/migtools/oadp-non-admin/_prowconfig.yaml
```diff
@@ -7,6 +7,9 @@ branch-protection:
             master:
               allow_force_pushes: true
               protect: true
+            oadp-1.5:
```
mateusoliveira43, line 10: do we actually need this? this is for fork repos, right? I think we can remove master instead o0f adding new branching, like oadp config https://github.com/openshift/release/blob/master/core-services/prow/02_config/openshift/oadp-operator/_prowconfig.yaml

kaovilai: need? no. nice to have? sure.

prevent accidental pushes by maintainers

kaovilai: without this if you have hit perm you could just push to it without a pr, an undesirable behavior.

sseago: @mateusoliveira43 Hmm. I would think we would want branch protection, but we probably don't need to allow force pushes. Then again, the same should be true for master.

mateusoliveira43: alright, I like @kaovilai points, LGTM to me

we could do as @sseago said and remove force push

mateusoliveira43 (approved):

core-services/prow/02_config/migtools/oadp-non-admin/_prowconfig.yaml
```diff
@@ -7,6 +7,9 @@ branch-protection:
...
+              allow_force_pushes: true
+              protect: true
           required_pull_request_reviews:
```
kaovilai, line 13:
```suggestion
              protect: true
            oadp-1.5:
              protect: true
          required_pull_request_reviews:
```

core-services/prow/02_config/migtools/oadp-non-admin/_prowconfig.yaml
kaovilai: done.

sseago (approved):

mateusoliveira43 (approved):

PR #63521, opened 2025-04-04
Remove intranet capability from build05

deepsm007:
/cc @openshift/test-platform


ggiguash:
In attempt to fix USHIFT-5475 & DPTP-4382

PR #63525, opened 2025-04-04
OCPQE-29097: Migrate baremetalds QE job to IBM cloud

sgoveas:
Adding 2 new chains to include steps `ofcir-acquire` for provisioning
and `ofcir-release` for deprovisioning in IBM cloud. New chains were
added as the existing chains are used by other teams with equinix. At
the moment for ofcir to work the jobs have to run on build05.


sgoveas:
@jianlinliu please review.

jianlinliu:
/lgtm

cc @liangxia to review.

liangxia:
> At
> the moment for ofcir to work the jobs have to run on build05.

What's the specific config/requirement here that requires build05 ?

tdomnesc:
> What's the specific config/requirement here that requires build05 ?

We are currently dependent on build05 because that is where our operator OFCIR is running. See https://github.com/openshift/release/blob/master/clusters/build-clusters/build05/ofcir/ofcir.yaml

Historically also build05 was the only cluster with intranet access.

We know this is not ideal and we already had discussions with Test Platform to migrate OFCIR and to change the jobs configs to use capabilities: - intranet. But this won't be done in time to migrate away from Equinix - the deadline is the end of this month. 

So for the time being we ask that run with this compromise.

PR #63563, opened 2025-04-07
rhdh job image update

subhashkhileri:
https://issues.redhat.com/browse/RHIDP-6853

### DIFF
--- a/ci-operator/step-registry/redhat-developer/rhdh/osd-gcp/claim-cluster/create/redhat-developer-rhdh-osd-gcp-claim-cluster-create-commands.sh
+++ b/ci-operator/step-registry/redhat-developer/rhdh/osd-gcp/claim-cluster/create/redhat-developer-rhdh-osd-gcp-claim-cluster-create-commands.sh
@@ -20,7 +20,6 @@ GITHUB_ORG_NAME="redhat-developer"
 GITHUB_REPOSITORY_NAME="rhdh"
 
 export RELEASE_BRANCH_NAME
-QUAY_REPO="rhdh/rhdh-hub-rhel9"
 # Get the base branch name based on job.
 RELEASE_BRANCH_NAME=$(echo ${JOB_SPEC} | jq -r '.extra_refs[].base_ref' 2>/dev/null || echo ${JOB_SPEC} | jq -r '.refs.base_ref')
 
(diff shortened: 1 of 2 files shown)


rm3l (commented):
LGTM, but it looks like there is a [ShellCheck error](https://storage.googleapis.com/test-platform-results/pr-logs/pull/openshift_release/63563/pull-ci-openshift-release-master-step-registry-shellcheck/1909237007771504640/build-log.txt) to fix: 
```
In .//ci-operator/step-registry/redhat-developer/rhdh/osd-gcp/claim-cluster/create/redhat-developer-rhdh-osd-gcp-claim-cluster-create-commands.sh line 23:
QUAY_REPO="rhdh/rhdh-hub-rhel9"
^-------^ SC2034: QUAY_REPO appears unused. Verify use (or export if used externally).
```

nickboldt:
The 1.7 builds are passing again. Details in https://issues.redhat.com/browse/RHIDP-4014 and https://issues.redhat.com/browse/RHIDP-6846

+1 from me to merge the above when you'd like. 

Followup improvements in https://issues.redhat.com/browse/RHIDP-6869

ci-operator/step-registry/redhat-developer/rhdh/nightly/redhat-developer-rhdh-nightly-commands.sh
```diff
@@ -52,7 +52,7 @@ GITHUB_ORG_NAME="redhat-developer"
 GITHUB_REPOSITORY_NAME="rhdh"
 
 export QUAY_REPO RELEASE_BRANCH_NAME
-QUAY_REPO="rhdh-community/rhdh"
```
zdrapela, line 55: If we switch all the nightly jobs to `rhdh/rhdh-hub-rhel9` image, the _CI test triage_ can miss when the PR checks are broken, since they use the `rhdh-community/rhdh` image. One can be broken while the other can be running just fine, as we've seen in the past.
Sure, he should be checking https://prow.ci.openshift.org/?job=pull-ci-redhat-developer-rhdh-main-e2e-tests if PR checks are passing, but it might be worth having another job, that would be running with the `rhdh-community/rhdh` image. 
I think we should consider that.

rm3l (approved):
LGTM based on the team discussion we had.

/lgtm

PR #63596, opened 2025-04-08
ossm: ztunnel: automatic upstream->downstream sync

zmiklank:


ci-operator/config/maistra/test-infra/maistra-test-infra-main.yaml
```diff
@@ -27,6 +27,26 @@ tests:
     make lint
   container:
     from: src
+- as: sync-upstream-ztunnel-1.24
```
FilipB, line 30: I think it will be better to have dedicated directory with job definition in https://github.com/openshift/release/tree/master/ci-operator/config/openshift-service-mesh

jwendell: @zmiklank I think it would be better if you adjust the logic in automator to allow both a merge *and* running a command/script. Currently it's mutually exclusive: https://github.com/maistra/test-infra/blob/main/tools/automator-main.sh#L374-L377

zmiklank: From what I understand if we keep current logic, and only allow both merge and cmd in one automator run if will result in 2 different PRs created, while I would like to have these two changes in one common PR.

jwendell: Yeah, maybe, that's part of the "adjust the logic" I mentioned.

Or, if it's too complicated, you can just create a script in the ztunnel repo, e.g. "sync_with_upstream.sh" with contents similar to [this function merge](https://github.com/openshift-service-mesh/sail-operator/blob/main/ossm/merge_upstream.sh#L44-L55), followed by a `cargo vendor`.

And then, this job would call that script instead of using `--merge` options.

FilipB: Sail operator is also using it's own [merge](https://github.com/openshift-service-mesh/sail-operator/blob/release-3.0/ossm/merge_upstream.sh) script which is used by automator like [this](https://github.com/openshift/release/blob/master/ci-operator/config/openshift-service-mesh/sail-operator/openshift-service-mesh-sail-operator-release-3.0__ocp-4.18.yaml#L92). That should work fine.

zmiklank:
The bot got 403 when wanted to create PR against the openshift-service-mesh/ztunnel repo. That is weird, because I already gave him (the openshift-merge-robot ) the access. Moreover th PR itself was created: https://github.com/openshift-service-mesh/ztunnel/pull/5. Maybe it happened when assigning labels. Will try to remove the labeling from the script.

yxun (approved):

PR #63631, opened 2025-04-09
Remove Tracing DAST jobs as the jobs are migrated to Konflux

IshwarKanse:
* Remove Israel (iblancasa) from owners files.
* Remove the DAST jobs and steps as the jobs are now migrated to run from the Konflux integration test jobs.


vprashar2929 (approved):
/lgtm

PR #63560, opened 2025-04-07
Temporary disable 4.20 to reduce overload and rate limit issues

deepsm007:
TODO: Revert this PR on Monday April 14th

[Thread](https://redhat-internal.slack.com/archives/CBN38N3MW/p1743787799320879)

/cc @openshift/test-platform @dgoodwin

PR #63652, opened 2025-04-09
Revert and Rebalance jobs

deepsm007:
/cc @openshift/test-platform @dgoodwin

PR #63657, opened 2025-04-09
Revert "Temporary disable 4.20 to reduce overload and rate limit issues"

neisw:
Reverts openshift/release#63560


neisw:
/close
in favor of https://github.com/openshift/release/pull/63652

PR #63669, opened 2025-04-09
fix security group values left over from the previous installation.

pruan-rht:
the securityGroups were carried over from the previous installation.  PTAL, @bear-redhat @deepsm007


deepsm007 (commented):
`build10-znq8n` this should be updated with  `build10-rlx6g` at every reference in the file

deepsm007 (commented):
/lgtm

PR #63700, opened 2025-04-10
fix iptables rule allowing traffic in disconnected env

bmanzari:
Temporary disabling observer pod and restoring iptables rule to block all traffic on all ports in disconnected environments

### DIFF
--- a/ci-operator/step-registry/baremetal/lab/post/firewall/baremetal-lab-post-firewall-commands.sh
+++ b/ci-operator/step-registry/baremetal/lab/post/firewall/baremetal-lab-post-firewall-commands.sh
@@ -41,6 +41,6 @@ timeout -s 9 10m ssh "${SSHOPTS[@]}" "root@${AUX_HOST}" bash -s -- \
   INTERNAL_NET_CIDR="${1}"
   IP_ARRAY="${@:2}"
   for ip in $IP_ARRAY; do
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/baremetal/lab/pre/firewall/baremetal-lab-pre-firewall-commands.sh
```diff
@@ -56,8 +56,7 @@ timeout -s 9 10m ssh "${SSHOPTS[@]}" "root@${AUX_HOST}" bash -s -- \
   IP_ARRAY="${@:2}"
   for ip in $IP_ARRAY; do
     # TODO: change to firewalld or nftables
-    # Allow connections on port 22 used by observer pod
-    iptables -A FORWARD -s ${ip} ! -d "${INTERNAL_NET_CIDR}" ! -p tcp --dport 22 -j DROP
+    iptables -A FORWARD -s ${ip} ! -d "${INTERNAL_NET_CIDR}" -j DROP
```
sgoveas, line 59: Post step needs to be updated as well

bmanzari: right! thanks steeve

PR #63780, opened 2025-04-14
Retire unused Prow job config files

JianLi-RH:
Some Prow jobs will not be used anymore, so move them into a retired folder.

1. OCP 4.12 do not support arm and multi, so retired them
2. we will not support 4.11 and earlier version for chain upgrade, so removed them.

/cc @jhou1 @jiajliu @dis016


dis016:
LGTM for me as well.

PR #62289, opened 2025-03-03
Add retries for image mirroring

dhaiducek:

PR #63804, opened 2025-04-14
(ocm) Allow `oc image mirror` failures to retry

dhaiducek:
Since adding retries for the filtering, the retries should also include retrying when `oc image mirror` returns an error rather than exiting immediately.

Followup to:
- https://github.com/openshift/release/pull/62289


mark-nc (approved):
/lgtm /approved

PR #63811, opened 2025-04-14
use rhel8 oc for configmap perfscale tests

tjungblu:
This should resolve the glibc failures we receive in our CI

### DIFF
--- a/ci-operator/step-registry/openshift-perfscale/configmap-scale/openshift-perfscale-configmap-scale-commands.sh
+++ b/ci-operator/step-registry/openshift-perfscale/configmap-scale/openshift-perfscale-configmap-scale-commands.sh
@@ -5,7 +5,7 @@ set -o pipefail
 set -x
 
 pushd /tmp
-curl -sSL https://mirror.openshift.com/pub/openshift-v4/clients/ocp/stable/openshift-client-linux.tar.gz | tar xz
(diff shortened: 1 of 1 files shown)


rsevilla87:
@tjungblu do you use this test?, I wonder if we can remove it as it's been replaced by other workloads.i.e: https://github.com/openshift/release/blob/master/ci-operator/config/openshift/etcd/openshift-etcd-openshift-4.19.yaml#L96-L109

tjungblu:
it is in use, but I didn't know about the other test that was introduced. @dusk125 do you know whether the other test is the same or similar to the configmap scale test?

dusk125:
I wasn't aware either, and I'm not sure if they do the same thing or not

rsevilla87:
> I wasn't aware either, and I'm not sure if they do the same thing or not

they don't , the other test consists of loading the cluster through creating a significant amount of objects. I wonder what's your current use case for the configmap-scale test

dusk125:
/reopen

Can we get the fix merged and bring the discussion of whether or not to remove one of the tests to another location? Slack thread maybe?

tjungblu:
@dusk125 let's remove it, it can't even reach ElasticSearch anymore. I don't think that the value we get out of this is higher than fixing it entirely

PR #63818, opened 2025-04-14
no-jira: use tech preview rather than custom no upgrade for vsphere multi-network

rvanderp3:
fixing test feature set to test multi-network in support of https://github.com/openshift/api/pull/2204


tthvo:
Let's see if I can approve :sweat_smile: 

/approve

jhixson74:
I guess installer folks can't approve this :-(

PR #63828, opened 2025-04-14
Update path for splunk-forwarder-images dockerfile

wshearn:
Moving the SFI dockerfile into build directory instead of a bespoke path.

PR in SFI with the change; https://github.com/openshift/splunk-forwarder-images/pull/67

### DIFF
--- a/ci-operator/config/openshift/splunk-forwarder-images/openshift-splunk-forwarder-images-master.yaml
+++ b/ci-operator/config/openshift/splunk-forwarder-images/openshift-splunk-forwarder-images-master.yaml
@@ -9,7 +9,7 @@ build_root:
     namespace: openshift
     tag: golang-1.15
 images:
-- dockerfile_path: build/Dockerfile.forwarder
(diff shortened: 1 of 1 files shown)


ci-operator/config/openshift/splunk-forwarder-images/openshift-splunk-forwarder-images-master.yaml
```diff
@@ -9,7 +9,7 @@ build_root:
     namespace: openshift
     tag: golang-1.15
 images:
-- dockerfile_path: containers/forwarder/Dockerfile
+- dockerfile_path: build/Dockerfile.forwarder
```
dem4gus, line 12: We should just use `build/Dockerfile` like other single-image projects, [example](https://github.com/openshift/release/blob/efc48473bf53a8db53313430a23d7a025fdd6f56/ci-operator/config/openshift/configure-alertmanager-operator/openshift-configure-alertmanager-operator-master.yaml#L4)

PR #63865, opened 2025-04-15
Optimize WORKDIR

liangxia:
If the `WORKDIR` doesn't exist, it will be created. So the `mkdir` is not necessary.

Reference doc https://docs.docker.com/reference/dockerfile/#workdir


liangxia:
/pj-rehearse pull-ci-openshift-openshift-tests-private-master-images

PR #63871, opened 2025-04-15
GitOps | Increase timeout for `operator-sdk run bundle` command

varshab1210:


jgwest (approved):
LGTM, thanks @varshab1210. This timeout parameter should be a better fit for the potential install time we are seeing!

PR #63885, opened 2025-04-15
Set default topic map for standalone Openshift docs distros

gabriel-rh:
Add default topic map so that we don't have to update every time there is a new standalone product


aireilly (approved):
/lgtm

PR #63902, opened 2025-04-16
Link gangway usage to docs.ci.openshift.org

liangxia:
We have the docs about gangway usage in docs.ci.openshift.org now. Link it here instead of maintain another copy.

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/README.md
+++ b/ci-operator/config/openshift/openshift-tests-private/README.md
@@ -126,30 +126,5 @@ releases:
 ```
 
 
-# Trigger a new test for an existing periodic job in Prow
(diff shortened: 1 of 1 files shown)


liangxia:
CC @pruan-rht

liangxia:
CC @pruan-rht

pruan-rht:
@liangxia we can merge it, but I think the current README is more specific to the openshift-test-private, which is more detailed than the specific.  How about keeping the current state and put the link as well at the end?

liangxia:
> @liangxia we can merge it, but I think the current README is more specific to the openshift-test-private, which is more detailed than the generic information from your link. How about keeping the current state and putting the link as well at the end?

I just don't want to maintain a 'duplicate' copy of the docs. I'll file a PR to update the examples in `docs.ci.openshift.org` with the override.

PR #63903, opened 2025-04-16
stackrox: retest 419interop

davdhacs:
/hold
/uncc


davdhacs:
also testing in stackrox/stackrox#14774

PR #63944, opened 2025-04-16
OLS-1662 - Increasing ols 4.18 cluster pool size

JoaoFula:
Increased the amount of jobs for most PRs to two so increasing the cluster pool to reflect the changes


JoaoFula:
This issue was tackled when migrating ols cluster claims to obs cluster pools

PR #63959, opened 2025-04-16
update INFW owner list

msherif1234:


martinkennelly (approved):

ci-operator/config/openshift/ingress-node-firewall/OWNERS
```diff
@@ -14,6 +14,9 @@ approvers:
 - knobunc
 - msherif1234
 - trozet
+- tssurya
+- martinkennelly
+-
```
liangxia, line 19: Looks like there is an extra `-` on this line

PR #63965, opened 2025-04-16
OSD-29370: Utilize .ci-operator.yaml config for osd-network-verifier

joshbranham:
This switches the config to read the build image from our repo, instead of having to come here and bump this when we make changes.

In support of https://github.com/openshift/osd-network-verifier/pull/307


abyrne55:
/lgtm

Thanks for chugging through this @joshbranham

PR #64006, opened 2025-04-17
Reduce number of AWS jobs triggered automatically

patrickdillon:
We are running A LOT of AWS jobs anytime 'aws' is matched in the path. Reduce many of these jobs so they are manually trigerred.

### DIFF
--- a/ci-operator/config/openshift/installer/openshift-installer-main.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-main.yaml
@@ -392,9 +392,9 @@ tests:
     cluster_profile: aws
     workflow: openshift-e2e-aws-upi-proxy
   timeout: 6h0m0s
-- as: e2e-aws-ovn-shared-vpc-custom-security-groups
+- always_run: false
+  as: e2e-aws-ovn-shared-vpc-custom-security-groups
   optional: true
-  run_if_changed: aws
   steps:
     cluster_profile: aws
     env:
@@ -438,7 +438,6 @@ tests:
   timeout: 6h0m0s
 - always_run: false
   as: e2e-aws-default-config
-  optional: true
   run_if_changed: aws
   steps:
     cluster_profile: aws
@@ -455,7 +454,6 @@ tests:
 - always_run: false
   as: e2e-aws-ovn-edge-zones
   optional: true
-  run_if_changed: aws
   steps:
     cluster_profile: aws
     env:
@@ -465,23 +463,22 @@ tests:
 - always_run: false
   as: e2e-aws-ovn-shared-vpc-edge-zones
   optional: true
-  run_if_changed: aws
   steps:
     cluster_profile: aws
     env:
       AWS_INSTALL_USE_MINIMAL_PERMISSIONS: "yes"
     workflow: openshift-e2e-aws-edge-zones-byo-vpc
   timeout: 6h0m0s
-- always_run: true
+- always_run: false
   as: e2e-aws-ovn-edge-zones-manifest-validation
   run_if_changed: aws
   steps:
     cluster_profile: aws
     workflow: openshift-installer-manifests-aws-edge-zones
   timeout: 3h0m0s
-- as: e2e-aws-ovn-imdsv2
+- always_run: false
+  as: e2e-aws-ovn-imdsv2
   optional: true
-  run_if_changed: aws
   steps:
     cluster_profile: aws
     env:
@@ -937,9 +934,9 @@ tests:
     cluster_profile: packet-assisted
     workflow: assisted-ofcir-baremetal
   timeout: 6h0m0s
-- as: e2e-aws-ovn-single-node
+- always_run: false
+  as: e2e-aws-ovn-single-node
   optional: true
-  run_if_changed: aws
   steps:
     cluster_profile: aws
     workflow: openshift-e2e-aws-single-node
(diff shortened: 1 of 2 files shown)


patrickdillon:
It would be good to check whether some of these have regular periodics or release jobs.

patrickdillon:
> > It would be good to check whether some of these have regular periodics or release jobs.
> 
> @patrickdillon e2e-aws-ovn-edge-zones-manifest-validation was created to prevent breaking edge zones across many enabled regions looking for invalid manifests generated by installer, it is not provision infrastructure, instead it runs many checks validating those manifests checking edge nuances, generating 600+ validations. (this test does not cover BYO VPC edge subnets scenarios)
> 
> If I can recommend: if we are going to set as optional all edge variants, the manifest validation would be something safe to keep.
> 
> There are periodic for edge jobs[1], but AFAIK there are no for release.
> 
> [1] https://prow.ci.openshift.org/?job=periodic*edge-zones*

Sounds good. I didn't touch manifests for that reason. 

I wonder if we could combine edge zones into the current BYO vnet job? So those two jobs are collapsed into one?

mtulio:
> > > It would be good to check whether some of these have regular periodics or release jobs.
> > 
> > 
> > @patrickdillon e2e-aws-ovn-edge-zones-manifest-validation was created to prevent breaking edge zones across many enabled regions looking for invalid manifests generated by installer, it is not provision infrastructure, instead it runs many checks validating those manifests checking edge nuances, generating 600+ validations. (this test does not cover BYO VPC edge subnets scenarios)
> > If I can recommend: if we are going to set as optional all edge variants, the manifest validation would be something safe to keep.
> > There are periodic for edge jobs[1], but AFAIK there are no for release.
> > [1] https://prow.ci.openshift.org/?job=periodic*edge-zones*
> 
> Sounds good. I didn't touch manifests for that reason.

IIUC [this is changing](https://github.com/openshift/release/pull/64006/files#diff-bf44a45e283d26f473de4f677c717226c9108e8617320b2d47c61def62863a54R472-R478) to optional :)


> 
> I wonder if we could combine edge zones into the current BYO vnet job? So those two jobs are collapsed into one?

That's a good idea. The `e2e-aws-ovn-shared-vpc-edge-zones` already covers BYO VPC, needing only to add the SG provisioning/config. I have only a few notes if we'll set that new job as required:
- edge zones will add two instance type to the cluster in a remote location that is: more expensive, sometime spending more time to provision the ec2
- we use 50% of the regions allowed to run CI (dedicated lease pool), because edge zones isn't supported in regions like us-east-2, us-west-1, etc, preventing to fail the job when selecting unsupported region from lease pool
- the edge node design requires node taints (applied to machine/capi manifests), making miss-configured e2e tests  to fail (unscheduled) when they wrongly selects a edge-worker node when setting up. (this is one example which I am trying to find reviewers of [permanent failed periodic-4.15][415-byovpc-edge]: https://github.com/openshift/origin/pull/29655)

[415-byovpc-edge]: https://prow.ci.openshift.org/job-history/gs/test-platform-results/logs/periodic-ci-openshift-release-master-nightly-4.15-e2e-aws-ovn-shared-vpc-edge-zones

IMO none of those items are blocker, just side notes to consider depending how we want to run it - I believe for an optional one would be ok as the developer will be under the control and intentionally aware of the tested feature.

mtulio:
@patrickdillon @tthvo FWIW I just collected a data point of the comment above, running a spike of unified BYO VPC job (https://github.com/openshift/release/pull/64100): 
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/64100/rehearse-64100-periodic-ci-openshift-release-master-nightly-4.19-e2e-aws-ovn-shared-vpc-edge-zones/1914693716002476032


looks like it ran correctly, except a, apparently, flake test* making the job to fail.

*`[sig-instrumentation] Prometheus [apigroup:image.openshift.io] when installed on the cluster shouldn't report any alerts in firing state apart from Watchdog and AlertmanagerReceiversNotConfigured`

ci-operator/config/openshift/installer/openshift-installer-main.yaml
```diff
@@ -465,23 +463,22 @@ tests:
...
-  run_if_changed: aws
   steps:
     cluster_profile: aws
     env:
       AWS_INSTALL_USE_MINIMAL_PERMISSIONS: "yes"
     workflow: openshift-e2e-aws-edge-zones-byo-vpc
   timeout: 6h0m0s
-- always_run: true
+- always_run: false
   as: e2e-aws-ovn-edge-zones-manifest-validation
```
mtulio, line 473: Do we need to make this optional? This job is [pretty stable](https://prow.ci.openshift.org/job-history/gs/test-platform-results/logs/periodic-ci-openshift-release-master-nightly-4.21-e2e-aws-ovn-edge-zones-manifest-validation), does not provision infrastructure, but makes a lot of API calls, although prevents us to be blind (e.g. unsupported instance type by new zone) in edge zones offered by OCP validating zones across many regions, even not covered by full installs by CI.

mtulio (commented):
overall lgtm, except the question of reasoning making `e2e-aws-ovn-edge-zones-manifest-validation` optional

tthvo:
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/64006/pull-ci-openshift-release-master-services-valid/1946132330695364608

There are some merge conflict I think:

> Automatic merge failed; fix conflicts and then commit the result.\n","severity":"info","time":"2025-07-18T09:04:54Z"

PR #64007, opened 2025-04-17
OKD-240: rebase OKD node images on c10s

joelcapitao:
See the commit messages

Needs https://github.com/openshift/os/pull/1806 and https://github.com/openshift-eng/ocp-build-data/pull/6628 to be merged before

### DIFF
--- a/ci-operator/config/openshift/os/openshift-os-master__okd-scos.yaml
+++ b/ci-operator/config/openshift/os/openshift-os-master__okd-scos.yaml
@@ -1,8 +1,8 @@
 base_images:
-  ocp_builder_stream9:
+  ocp_builder_stream:
     name: builder
     namespace: ocp
-    tag: stream9
+    tag: stream10
 build_root:
   image_stream_tag:
     name: coreos-assembler
@@ -19,12 +19,12 @@ images:
     value: "1"
   dockerfile_path: extensions/Dockerfile
   inputs:
-    ocp_builder_stream9:
+    ocp_builder_stream:
       as:
       - registry.access.redhat.com/ubi9/ubi:latest
     stream-coreos:
       as:
-      - registry.ci.openshift.org/rhcos-devel/rhel-coreos:latest
+      - overridden
   to: stream-coreos-extensions
 promotion:
   to:


ci-operator/config/openshift/os/openshift-os-master__okd-scos.yaml
```diff
@@ -13,6 +13,10 @@ images:
   - name: OPENSHIFT_CI
     value: "1"
   dockerfile_path: Containerfile
+  inputs:
+    stream-coreos-base:
+      as:
+      - quay.io/openshift-release-dev/ocp-v4.0-art-dev:c10s-coreos
```
Prashanth684, line 19: this does the reverse of what you think - in this case prow will look for a match for `quay.io/openshift-release-dev/ocp-v4.0-art-dev:c10s-coreos` and replace that wtih `stream-coreos-base`

Prashanth684: so we need to do the other way around which i'm not sure if we can substitute with a quay image

jcapiitao: I found out `cluster` keyword was used in the past to pull from quay.io as per [1].
Let my try out the reverse with this keyword.

[1]  https://github.com/relyt0925/release/blob/a2689a4ac024db00a7b69926d9714043f7712c8d/ci-operator/config/cri-o/cri-o/cri-o-cri-o-release-1.18.yaml#L3

jcapiitao:
/hold
we want it after branch cut

ci-operator/config/openshift/os/openshift-os-master__okd-scos.yaml
```diff
@@ -17,16 +17,14 @@ images:
...
     value: "1"
-  - name: VARIANT
-    value: okd-c9s
   dockerfile_path: extensions/Dockerfile
   inputs:
-    ocp_builder_stream9:
+    ocp_builder_stream:
       as:
-      - registry.access.redhat.com/ubi9/ubi:latest
+      - registry.access.redhat.com/ubi10-beta/ubi:latest
```
Prashanth684, line 24: revert back?

jcapiitao: Yes, let's keep ubi9 here

Prashanth684:
i guess we merge this first and then https://github.com/openshift/os/pull/1806 ? @jcapiitao

PR #64048, opened 2025-04-18
remove deprecated hibernate env var

ritmun:
HIBERNATE_AFTER_USE is deprecated since  https://github.com/openshift/osde2e/pull/2931/files


jbpratt (approved):
/lgtm

PR #64072, opened 2025-04-21
OSD-28968: add multiarch-devel to cluster app-ci for oci image pull

lwan-wanglin:
add multiarch-devel to cluster app-ci for oci image pull following the [doc ](https://docs.ci.openshift.org/docs/how-tos/use-registries-in-build-farm/#how-do-i-gain-access-to-qci)

### DIFF
--- a/clusters/app.ci/registry-access/multiarch/admin_manifest.yaml
+++ b/clusters/app.ci/registry-access/multiarch/admin_manifest.yaml
@@ -1,0 +1,65 @@
+# this is the Namespace in which your images live
+apiVersion: v1
+kind: Namespace
+metadata:
+  annotations:
+    openshift.io/description: Automation ServiceAccounts for Multiarch team
+    openshift.io/display-name: MULTIARCH team CI
+  name: multiarch
+---
+# this is the ServiceAccount whose credentials you will use to pull images
+kind: ServiceAccount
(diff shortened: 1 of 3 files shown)


core-services/sync-rover-groups/_config.yaml
```diff
@@ -59,6 +59,7 @@ groups:
     - hosted-mgmt
   multiarch-devel:
     clusters:
+    - app.ci
```
liangxia, line 62: This is for the human user. I guess you also need https://docs.ci.openshift.org/docs/how-tos/use-registries-in-build-farm/#token-for-programmatic-access-to-qci

clusters/app.ci/registry-access/multiarch/admin_manifest.yaml
```diff
@@ -0,0 +1,65 @@
+# this is the Namespace in which your images live
+apiVersion: v1
+kind: Namespace
+metadata:
+  annotations:
+    openshift.io/description: Automation ServiceAccounts for Multiarch team
+    openshift.io/display-name: MULTIARCH team CI
+  name: multiarch
```
lwan-wanglin, line 8: @liangxia must the namespace be set the same as the operator project name?

liangxia (approved):

PR #64095, opened 2025-04-21
Mount vsphere credentials for hive e2e

dlom:
xref: [HIVE-2270](https://issues.redhat.com//browse/HIVE-2270)
see also: https://github.com/openshift/hive/pull/2541

slack thread: https://redhat-internal.slack.com/archives/CH06KMDRV/p1745262442030989


2uasimojo:
I think Joe/Rich's lgtm might be enough to merge this. If not, I'll stamp it once one or both of them have acked. But it makes sense to me, assuming this is the correct one of several possible sets of vcenter creds I see floating around the release repo.

dlom:
This is the path that the `GOVC_TLS_CA_CERTS` env var is pointing to, so theoretically it should be all that's required

2uasimojo:
Alright, let's get this moving. If it's wrong, it ain't gonna break the job worse.

/lgtm

PR #64125, opened 2025-04-22
DPTP-4399: Add github-ldap-user-group-creator token on the master cluster

danilo-gemoli:
As per title.

/cc @deepsm007


deepsm007:
/retitle DPTP-4399: Add github-ldap-user-group-creator token on the master cluster

danilo-gemoli:
I've added some other secrets as well to keep this cluster consistent with the rest.

PR #64143, opened 2025-04-22
NO-JIRA: add metal techpreview serial jobs

neisw:
Needed for featuregate promotion


neisw:
/hold
@RishabhSaini for review of ipv6 PinnedImages failures

ci-operator/config/openshift/release/openshift-release-master__nightly-4.19.yaml
```diff
@@ -343,6 +343,20 @@ tests:
       enable:
       - observers-resource-watch
     workflow: baremetalds-e2e-ovn-ipv6
+- as: e2e-metal-ipi-serial-ovn-ipv6-techpreview
+  cluster: build05
+  cron: '@daily'
```
stbenjam, line 348: Given metal's limited resources, I'd suggest using a cronline directly instead to run these at a non- busy time 
image

stbenjam: The rehearsals are also close to the timeout limit, do you want to add `shard_count: 2` to split it? Should be OK as long as they're being run at non-busy times

neisw:
/pj-rehearse ack

[marking 4.19 jobs as candidate](https://github.com/openshift/sippy/pull/2571) should keep test failures in the new job from showing as regressed.

PR #64147, opened 2025-04-22
Revert "DPTP-4400: Use quay-proxy instead of registry.ci for step-registry"

pruan-rht:
Reverts openshift/release#64030


deepsm007:
/lgtm

@liangxia we need to fix the references and make sure each step can pull from QCI

PR #64162, opened 2025-04-23
ovnk: rename e2e-aws-ovn-kubevirt

qinqon:
There are some kubevirt jobs at ovnk with the name e2e-aws-ovn-virt so they are confused with the hypershift ones e2e-aws-ovn-kubevirt, this change add the hypershift word to the name so they don't get confused e2e-aws-ovn-hypershift-kubevirt

### DIFF
--- a/ci-operator/config/openshift/ovn-kubernetes/openshift-ovn-kubernetes-master.yaml
+++ b/ci-operator/config/openshift/ovn-kubernetes/openshift-ovn-kubernetes-master.yaml
@@ -373,7 +373,7 @@ tests:
       EXTRA_MG_ARGS: --host-network
     workflow: openshift-e2e-aws-ovn-serial
   timeout: 6h0m0s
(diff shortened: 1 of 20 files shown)


qinqon:
> Thanks @qinqon! While we are at it do you think we should rename the `e2e-aws-ovn-virt` job to `e2e-aws-ovn-kubevirt`? not sure why we went with the first option, maybe there is a reason.

Do you know if that affect zippy ?

kyrtapz:
> > Thanks @qinqon! While we are at it do you think we should rename the `e2e-aws-ovn-virt` job to `e2e-aws-ovn-kubevirt`? not sure why we went with the first option, maybe there is a reason.
> 
> Do you know if that affect zippy ?

I don't think so, sippy cares about tests(and their names) and not jobs iiuc.

tssurya (approved):
/approve

PR #64171, opened 2025-04-23
ACS: Add ui-component-tests to release 4.7 and release template

msugakov:
These were present in `master` and `release-4.6` but absent in `release-4.7` and in the release template (`release-x.y`), see the screenshot

image

Not sure if they are needed in the newer releases but will find out through this PR.

Another question I have is that this job only exists as presubmit, i.e. only for PRs but not for merges. If this job is meant to be permanent, we should add it as postsubmit, i.e. for merges. Is it a good time to do this already? For which releases?

https://redhat-internal.slack.com/archives/CC5UHD0KA/p1745397407589049

As part of https://issues.redhat.com/browse/ROX-29154

/uncc

### DIFF
--- a/ci-operator/config/stackrox/stackrox/stackrox-stackrox-release-4.7.yaml
+++ b/ci-operator/config/stackrox/stackrox/stackrox-stackrox-release-4.7.yaml
@@ -48,6 +48,12 @@ tests:
       COLLECTION_METHOD: core_bpf
     workflow: stackrox-stackrox-e2e-job
   timeout: 5h0m0s
+- as: ui-component-tests
+  optional: true
+  run_if_changed: ^ui/
+  steps:
+    test:
+    - ref: stackrox-initial
 - as: gke-ui-e2e-tests
   optional: true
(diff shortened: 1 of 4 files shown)


dvail:
> Not sure if they are needed in the newer releases but will find out through this PR.
> 
> Another question I have is that this job only exists as presubmit, i.e. only for PRs but not for merges. If this job is meant to be permanent, we should add it as postsubmit, i.e. for merges. Is it a good time to do this already? For which releases?

Yes this is intended to be permanent, and I think having it as postsubmit is the correct thing to do. This _could_ be added to all current releases, but there is relatively low test coverage with this job so I don't think it is critical to add it to older patch streams. Like @davdhacs mentioned above, adding it as a postsubmit moving forward is good enough and gives us one less potential problem to worry about amongst all the moving parts with our release process at the moment.

msugakov:
I looked further and I suggest to proceed with https://github.com/stackrox/stackrox/pull/15147 and https://github.com/openshift/release/pull/64461 instead of this #64171.

See https://redhat-internal.slack.com/archives/CC5UHD0KA/p1745397407589049

PR #63465, opened 2025-04-03
CM-553: update pipeline image tags to use v1.16.4

siddhibhor-56:
tag updates in o/cert-manager-operator configs with version [v1.16.4]

PR #64190, opened 2025-04-23
[o/cert-manager-operator] Use latest tag in bundle substitution

swghosh:
As a result of https://github.com/openshift/release/pull/63465, we removed the operand version tags recently. This broke the e2e(s) causing failures eg. https://github.com/openshift/cert-manager-operator/pull/258 > https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cert-manager-operator/258/pull-ci-openshift-cert-manager-operator-master-e2e-operator/1915003203531837440 with an InvalidImageName on the cert-manager pods.

Seems that CI substitution tool does not touch the tag unless the tag is already specified in source. Adding back a `latest` tag to fix the e2e config.

### DIFF
--- a/ci-operator/config/openshift/cert-manager-operator/openshift-cert-manager-operator-master.yaml
+++ b/ci-operator/config/openshift/cert-manager-operator/openshift-cert-manager-operator-master.yaml
@@ -37,11 +37,11 @@ operator:
   substitutions:
   - pullspec: openshift.io/cert-manager-operator:latest
     with: pipeline:cert-manager-operator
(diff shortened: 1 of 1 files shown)


swghosh:
Extracted the CSV out from the bundle image that was built by the rehearsal job, looking good: the patch seems to have work as expected.

```
cd $(mktemp -d) && oc image extract registry.build10.ci.openshift.org/ci-op-d5mdiksc/pipeline:cert-manager-operator-bundle
```

/pj-rehearse ack

PR #64196, opened 2025-04-23
add ocm-api-metamodel to be managed.

deads2k:
starting point to ensure workflows function.


deads2k:
already approved in concept with Victor, we're starting here.

PR #64199, opened 2025-04-23
Replace URL for route CRD install for Azure HO install

bryan-cox:
This commit replaces the URL for the route CRD which needs to be applied when installing the HO for ARO HCP clusters.


bryan-cox:
/pj-rehearse pull-ci-openshift-cluster-api-provider-azure-master-hypershift-e2e-aks

bryan-cox:
Ah I grabbed the wrong test 🤦‍♂️

PR #64248, opened 2025-04-24
CMP-3363: Make rosa e2e opt-in for FIO

rhmdnd:
We don't need to run this PR on most changes, since we haven't tackled
the ROSA HCP work for FIO, yet. This commit removes the skip condition
which makes the job optional and opt-in for FIO PRs, which is what we
want at this point.


rhmdnd:
/pj-rehearse ack

No need to rehearse this since we not doing anything with ROSA HCP at the moment.

PR #64257, opened 2025-04-24
MULTIARCH-5349: MTO adjusted to track v1.x

AnnaZivkovic:
Changed to rolling stream releases so we will be using 1.x instead of 1.0


aleskandro:
there's enough quorum from the shortest jobs.

/pj-rehearse ack

PR #64306, opened 2025-04-27
Add OperatorLifecycleManager capability for e2e-aws-ovn-no-capabilities

lihongan:
In 4.19 ingress needs `subscription` resource that provided by OperatorLifecycleManager capability.
To unblock the installation, adding the cap as workaround.


lihongan:
/close
the issue is fixed and workaround is not needed anymore

PR #64309, opened 2025-04-28
CNTRLPLANE-745: Remove hypershift-e2e-aks job from release-4.18 presubmits

bryan-cox:
We don't need these jobs since 4.19 will be the first release for ARO HCP now.


bryan-cox:
cc @Miciah

PR #64339, opened 2025-04-28
OCPBUGS-55432: baremetal: fix regressed retries when mirroring

dtantsur:
Currently, all `oc image mirror` calls are grouped in a function.
The retry logic validates the result of this function to determine if
a retry is needed. However, it will only return the outcome of the last
call.

This PR makes sure the function is aborted on the first failure.
Images are grouped in an array for readability.


dtantsur:
/cc @elfosardo 

For your information

sdodson:
/label approved
It'll still need rehearsal ack, someone else can add that

sdodson:
/pj-rehearse ack
The ipv6 job which was mentioned as being relevant passed and clearly demonstrates the new code being used.

PR #64342, opened 2025-04-28
Presubmits considering almost fields from ci-operator config

hector-vido:
This is necessary to merge https://github.com/openshift/ci-tools/pull/4529


ci-operator/jobs/stolostron/policy-collection/stolostron-policy-collection-main-presubmits.yaml
```diff
@@ -1,7 +1,7 @@
 presubmits:
   stolostron/policy-collection:
   - agent: kubernetes
-    always_run: false
+    always_run: true
```
dhaiducek, line 4: I'm not sure this is correct. We have `optional: true` in our configuration for the images stanza...

https://github.com/openshift/release/blob/dcc7d36d580546ab7f73c8d676c733c0825833f3/ci-operator/config/stolostron/policy-collection/stolostron-policy-collection-main__opp-ocp4.18-interop.yaml#L40

hector-vido: As far as I understood, `optional` means that this job status should not be required for merge. If you doesn't have `run_if_changed` or `skip_only_if_changed`, `always_run` will be assumed as `true`.

dhaiducek: That's not how I read the [operator reference](https://steps.ci.openshift.org/ci-operator-reference#line224):

```
images:
  . . .
      # Optional means the build step is not built, published, or
      # promoted unless explicitly targeted. Use for builds which
      # are invoked only when testing certain parts of the repo.
      optional: true
```

Furthermore, the `images` isn't a job that is user-configured, so as far as I know this is the only mechanism to control whether it runs in PRs since `run_if_changed` isn't available there.

dhaiducek: Thanks @jmguzik and @hector-vido!

Looks like that's this block in `ci-tools`?
```golang
	// TODO(muller): Special case images jobs for now. Some repos are marking
	// images jobs as optional for which we do not have syntax in ci-operator (should we?).
	// Tolerate manual changes for these jobs for now
	if strings.HasSuffix(merged.Name, "-images") {
```
ref: https://github.com/openshift/ci-tools/blob/302d528be3edd8eb11c31bb747a5ba5903b8d32f/pkg/jobconfig/files.go#L494-L497

I suppose we can live with extra image builds in our PRs until something else is implemented, but I thought I'd raise it since it looks like a large number of the updates here concern `-images` presubmit jobs and would catch quite a few repos by surprise it seems.

jmguzik: @dhaiducek I guess the fix for now could be in this section to mark if optional, always_run = false? We need to test it how it affects other jobs.

dhaiducek: Yes, that sounds right to me!

It _seems_ to align with the `TODO` there also since I'm not sure whether there's a use-case there when someone would mark `images` as `optional: true` but then turn around and then have `always_run: true`...

PR #64347, opened 2025-04-28
HIVE-2846: Temporary e2e-azure/4.19 presubmit

2uasimojo:
We've got this covered via periodics, but need it temporarily in a presubmit so we can run it on demand while developing the fix described in the linked issue. Will revert when that lands.


2uasimojo:
> We expect this to fail with the following

✓

/pj-rehearse ack

/assign @suhanime

PR #64353, opened 2025-04-28
CMP-3380: Use OpenShift bundle for FIO bundle upgrade testing

rhmdnd:
We have two bundle Dockerfiles in FIO, and one was recently refactored
to work with Konflux. Let's use that one for bundle upgrade testing
since it contains the updates necessary for shipped bundles.


rhmdnd:
/pj-rehearse pull-ci-openshift-file-integrity-operator-master-e2e-bundle-aws-upgrade

rhmdnd:
@Vincent056 should be ready for a review.

rhmdnd:
/hold

We still have a problem with this approach in that the bundle image built from (`bundle.openshift.Dockerfile`) does not reference the operator image from the PR under test.

We'll need to figure out how we're going to address that.

PR #64390, opened 2025-04-29
Add test arg for coverage

JustinKuli:
By default, `go test` produces coverage information for each individual package it is testing. So for a test in package `foo`, only the files in package `foo` are considered for coverage, even if you run the tests as `./...`. The search-collector repository has tests in `pkg/reconciler` which exercise code in `pkg/transforms`, so that coverage information was not previously being collected.


JustinKuli:
For reference of why this kind of coverage information is necessary, see https://github.com/stolostron/search-collector/pull/595

PR #64416, opened 2025-04-30
create priv forks for oadp repos

ashwindasr:
Productizing OADP repos in openshift-priv so that if in the future:
- there are embargoes, it can be handled in priv
- its onboarded to ART, one less step to worry about.

This change will automatically create new repos in openshift-priv, under the same name, and CI will periodically sync all branches and all commits.

### DIFF
--- a/core-services/openshift-priv/_whitelist.yaml
+++ b/core-services/openshift-priv/_whitelist.yaml
@@ -42,4 +42,4 @@ whitelist:
     - kubernetes-cli-runtime
     - kubernetes-sigs-structured-merge-diff
     - pj-rehearse # Temporary added. We will use this repo to test the publicize plugin.
-    - restic
+    - openshift-velero-plugin


core-services/openshift-priv/_whitelist.yaml
```diff
@@ -40,3 +42,4 @@ whitelist:
     - kubernetes-cli-runtime
     - kubernetes-sigs-structured-merge-diff
     - pj-rehearse # Temporary added. We will use this repo to test the publicize plugin.
+    - restic
```
rayfordj, line 45:
```suggestion
    - openshift-velero-plugin
```
I completely forgot that we don't actually build a component from this repo; its binary is included as part of velero images.  Let's use [openshift-velero-plugin](https://github.com/openshift/openshift-velero-plugin) instead, please.

PR #64448, opened 2025-05-01
Use Standard_D8s_v5

sdodson:
https://issues.redhat.com/browse/CORS-3997 proposes moving to v5 instance types. Lets get a bit of soak time on that prior to landing that change in the installer.


vrutkovs (approved):
/lgtm

PR #64511, opened 2025-05-05
OPRUN-3878: release-4.18: Add periodics for openshift-operator-framework-olm-release

camilamacedo86:


tmshort:
Compared to the `main` periodics:
- Several tests are not present:
  - e2e-gcp-ovn
  - e2e-gcp-cponsole-olm
  - e2e-upgrade
- a new docker file was added:
  - operator-framework-tools.Dockerfile
  
  Why?

camilamacedo86:
Hi @tmshort 

https://github.com/openshift/release/pull/64511#issuecomment-2851031945

We are doing the same thing that was done for 4.17.

We have introduced changes regarding the monorepo initiative. So, I suggest that we add the periodic for 4.18 as it is for 4.17, and then, if we need to add any other tests for 4.18 that do not exist for 4.17, we can go in a follow-up. WDYT?

tmshort:
Is this even the right location? Shouldn't these be in operator-framework-operator-controller if we're looking for default catalog test support?

PR #64314, opened 2025-04-28
Refine 4.15 images

liangxia:

PR #64548, opened 2025-05-06
[AUTO]make opm support rhel8

kuiwang02:
/cc @Xia-Zhao-rh @liangxia 

recently we met all opm cases fail for 4.15, for example, https://issues.redhat.com/browse/OCPQE-29518.

the reason is that PR https://github.com/openshift/release/pull/64314 take opm from https://mirror2.openshift.com/pub/openshift-v4/amd64/clients for 4.15, and it only support rhel9, while 4.15:tools is rhel8, so the error happen
```console
opm: /lib64/libc.so.6: version GLIBC_2.32' not found (required by opm)\nopm: /lib64/libc.so.6: versionGLIBC_2.33' not found (required by opm)\nopm: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by opm)
```

why it does not happen before last week:
it take https://github.com/openshift/openshift-tests-private/blob/release-4.15/images/Dockerfile.golang to make opm which support rhel8.
why there is no such issue for 4.14 and 4.13:
the opm from https://mirror2.openshift.com/pub/openshift-v4/amd64/clients for 4.14 and 4.13 only support rhel8, not rhel9.
why there is no such issue for 4.16 and higher:
the opm from https://mirror2.openshift.com/pub/openshift-v4/amd64/clients for 4.16 and higher support both rhel8 and rhel9. we could get correct opm per :tools image.

so, currently we have to compile it with correct rhel and use it.

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15.yaml
@@ -32,9 +32,17 @@ build_root:
 images:
 - dockerfile_literal: |
     FROM src AS builder
+    RUN mkdir -p /tmp/opmbuild && cd /tmp/opmbuild && pwd && \
+      MIRRORURL=https://mirror2.openshift.com/pub/openshift-v4 && CLIENTURL=${MIRRORURL}/x86_64/clients/ocp/latest-4.15 && \
+      curl -s -k -L ${MIRRORURL}/x86_64/clients/ocp/ -o ocp.html && \
+      rcgaver=$(grep -E "<a href=\"4\.15" ocp.html |cut -d"\"" -f2|cut -d"/" -f1|sort -V|tail -1) && echo "V${rcgaver}V" && \
+      if [ "V${rcgaver}V" != "VV"  ]; then CLIENTURL=${MIRRORURL}/x86_64/clients/ocp/${rcgaver}; fi && \
+      curl -s -k -L ${CLIENTURL}/opm-src.tar.gz -o opm-src.tar.gz  &&  tar -vzxf opm-src.tar.gz && opmdir=$(find ./ -type d -name "operator-framework-olm*") && \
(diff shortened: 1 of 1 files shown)


kuiwang02:
/pj-rehearse pull-ci-openshift-openshift-tests-private-release-4.15-images

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15.yaml
```diff
@@ -32,9 +32,19 @@ build_root:
 images:
 - dockerfile_literal: |
     FROM src AS builder
+    RUN mkdir -p /tmp/opmbuild && cd /tmp/opmbuild && pwd && \
+      MIRRORURL=https://mirror2.openshift.com/pub/openshift-v4 && CLIENTURL=${MIRRORURL}/x86_64/clients/ocp/candidate && \
+      curl -s -k -L ${MIRRORURL}/x86_64/clients/ocp/ -o ocp.html && curl -s -k -L ${MIRRORURL}/x86_64/clients/ocp-dev-preview/ -o pre.html && \
+      ecver=$(grep -E "<a href=\"candidate-4\.15" pre.html |cut -d"\"" -f2|cut -d"/" -f1|sort -V|tail -1) && echo "V${ecver}V" && \
```
liangxia, line 38: 4.15 has been GAed for a long time, perhaps we can remove those EC stuff.

kuiwang02:
/pj-rehearse pull-ci-openshift-openshift-tests-private-release-4.15-images

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.15.yaml
kuiwang02: @liangxia thanks. updated.

kuiwang02:
@Xia-Zhao-rh could you please add lgtm if it is OK? thanks

@liangxia could you please approve it if it is OK? thanks

PR #64552, opened 2025-05-06
snc-gather: Add logic to create GCP bucket for bundle upload

praveenkumar:


ci-operator/step-registry/gather/snc/gather-snc-commands.sh
```diff
@@ -39,12 +39,24 @@ LD_PRELOAD=/usr/lib64/libnss_wrapper.so gcloud compute scp \
...
+  echo "Bucket 'gs://crc-bundle-${PULL_NUMBER}' already exists. No action needed."
+else
+  echo "create crc-bundle-${PULL_NUMBER} bucket"
+  LD_PRELOAD=/usr/lib64/libnss_wrapper.so gcloud storage buckets create "gs://crc-bundle-${PULL_NUMBER}" \
+      --project "${GOOGLE_PROJECT_ID}"
+fi
+
 echo "Upload the bundle to gcp crc-bundle bucket"
-LD_PRELOAD=/usr/lib64/libnss_wrapper.so gsutil cp /tmp/*.crcbundle gs://crc-bundle/
+LD_PRELOAD=/usr/lib64/libnss_wrapper.so gsutil cp /tmp/*.crcbundle gs://crc-bundle-"${PULL_NUMBER}"/
```
cfergeau, line 55: Any particular reason for switching from a "global" bucket to per-PR buckets?

cfergeau (commented):
/lgtm

ci-operator/step-registry/gather/snc/gather-snc-commands.sh
praveenkumar: yes because global bucket get removed within a day so it might be possible just after bundle update the bucket get deleted :( . with associated with  per pr bucket atleast it make sure we have a day to test it out.

redbeam:
@praveenkumar this way the script exits early because of `set -e` if the bucket doesn't exist:

```bash
LD_PRELOAD=/usr/lib64/libnss_wrapper.so gcloud storage buckets describe "gs://crc-bundle-${PULL_NUMBER}" \
    --format="value(name)" > /dev/null 2>&1

if [ $? -eq 0 ]; then
```

Suggestion - something like this:

```bash
if LD_PRELOAD=/usr/lib64/libnss_wrapper.so gcloud storage buckets describe "gs://crc-bundle-${PULL_NUMBER}" \
    --format="value(name)" > /dev/null 2>&1; then
  echo "Bucket 'gs://crc-bundle-${PULL_NUMBER}' already exists. No action needed."
else
  echo "create crc-bundle-${PULL_NUMBER} bucket"
  LD_PRELOAD=/usr/lib64/libnss_wrapper.so gcloud storage buckets create "gs://crc-bundle-${PULL_NUMBER}" \
      --project "${GOOGLE_PROJECT_ID}"
fi
```

PR #64553, opened 2025-05-06
add presubmit e2e-hypershift-conformance-techpreview for Origin

lihongan:
The purpose of the pre-submit to enable techpreview feature gates e2e test on hypershift before promoting to GA and we could catch the issues earlier. 

Slack context: https://redhat-internal.slack.com/archives/C01CQA76KMX/p1744632408828129


lihongan:
Checked the job and looks `timeout: 6h0m0s` is not required, the default timeout should be fine.

PR #64559, opened 2025-05-06
NO-ISSUE: Update OCP versions/master external jobs which rely on assisted images to ocm-2.15

linoyaslan:
/cc @danmanor


linoyaslan:
IT'S JUST A TEST

PR #64577, opened 2025-05-06
OCPBUGS-55629: remove `go-verify-deps` from `openshift/console`

logonoff:
https://github.com/openshift/console/pull/15024 should be merged first


jupierce:
Since this test is broken for console, I'm fine to lgtm now. Let's just be sure it is replaced with something that enforces vendored deps correctly like: https://github.com/openshift/release/pull/64596 .

logonoff:
> Since this test is broken for console, I'm fine to lgtm now. Let's just be sure it is replaced with something that enforces vendored deps correctly like: #64596 .

I added the check in https://github.com/openshift/console/pull/15024

jhadvig (approved):
/lgtm
/approve

PR #64590, opened 2025-05-06
config: add coreos/rhel-coreos-config

jlebon:
This is a new repo that was split out of openshift/os. Add configuration for it.

Move all the cosa-based tests from openshift/os to this new repo.


prestist (approved):
Lgtm

PR #64592, opened 2025-05-06
NO-JIRA: Downgrade operator-sdk image version for CMA and VPA operators

maxcao13:
4.20 image does not exist yet and tests are failing because of it.

See https://github.com/openshift/custom-metrics-autoscaler-operator/pull/74#issuecomment-2855584978


jkyros:
Can't test with what isn't there! :stuck_out_tongue: 
/lgtm

PR #64599, opened 2025-05-07
Security Profile Operator Roll back to golang 1.23

Vincent056:
We need to roll back the base image to continue the downstream release on golang 1.23


rhmdnd (approved):
/lgtm

rhmdnd:
> We need to roll back the base image to continue the downstream release on golang 1.23

For additional context - this is due to 0.9.0 being on 1.23, and a 0.9.1 update of some kind pulling in 1.24.

PR #36150, opened 2023-02-07
OSSM-3267: Add maistra envoy presubmit jobs

yxun:

PR #64605, opened 2025-05-07
Remove unused AMI config

yunjiang29:
* `cn-north-1` and `cn-northwest-1` are not being tested in CI.
* The AMIs for `us-gov-west-1` and `us-gov-east-1` are automatically included in `coreos print-stream-json` now, byo-AMI is not required for GovCloud regions.

### DIFF
--- a/ci-operator/step-registry/ipi/conf/aws/ipi-conf-aws-commands.sh
+++ b/ci-operator/step-registry/ipi/conf/aws/ipi-conf-aws-commands.sh
@@ -296,13 +296,6 @@ rm /tmp/pull-secret
 
 # custom rhcos ami for non-public regions
 RHCOS_AMI=
-if [ "$REGION" == "us-gov-west-1" ] || [ "$REGION" == "us-gov-east-1" ] || [ "$REGION" == "cn-north-1" ] || [ "$REGION" == "cn-northwest-1" ]; then
-  # TODO: move repo to a more appropriate location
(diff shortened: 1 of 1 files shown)


yunjiang29:
@jianlinliu PTAL

ci-operator/step-registry/ipi/conf/aws/ipi-conf-aws-commands.sh
```diff
@@ -296,13 +296,6 @@ rm /tmp/pull-secret
 
 # custom rhcos ami for non-public regions
 RHCOS_AMI=
-if [ "$REGION" == "us-gov-west-1" ] || [ "$REGION" == "us-gov-east-1" ] || [ "$REGION" == "cn-north-1" ] || [ "$REGION" == "cn-northwest-1" ]; then
-  # TODO: move repo to a more appropriate location
-  curl -sL https://raw.githubusercontent.com/yunjiang29/ocp-test-data/main/coreos-for-non-public-regions/images.json -o /tmp/ami.json
```
yunjiang29, line 301: @yxun the _embedded_ AMIs are being used by GovCloud, so please update https://github.com/openshift/release/pull/36150/files#diff-fa12c2c52dcf6920a7e6d88d84100fb3e67c3ba36df80070e042047ed51bf925R273 as well. `ocp-test-data` is no longer supported.

yxun: Thanks for letting us know. I have updated this change in the PR below.

PR #64618, opened 2025-05-07
snc: Add bucket test as part of if statement to avoid failure

praveenkumar:
Since `set -e` is used, script exit as soon as there is no bucket, with following logic it will make sure not exit in case no bucket.


cfergeau (approved):
/lgtm
/approve

PR #64650, opened 2025-05-07
Add a PodDisruptionBudget for the MTO

aleskandro:
Add a PodDisruptionBudget for the MTO to make it more resilient to cases where the webhook pods couldn't spread among different nodes at creation time.

### DIFF
--- a/clusters/build-clusters/multiarch_tuning_operator/admin_40_pdb.yaml
+++ b/clusters/build-clusters/multiarch_tuning_operator/admin_40_pdb.yaml
@@ -1,0 +1,10 @@
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: pod-placement-webhook
+  namespace: openshift-multiarch-tuning-operator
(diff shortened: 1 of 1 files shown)


bear-redhat:
/hold
I think MTO shall set up antiaffinity or something to make the HA proper...?

aleskandro:
@bear we already have topologySpreadConstraint, but as of now we haven't wanted to force them (avoid the scheduling of the pods when they cannot be satisfied, e.g. when only 1 or 2 nodes can host the a replica out of 3) because in some scenarios they can cause other types of failures. We have been evaluating also PodDisruptionBudgets, but we haven't signed off about them being created automatically yet. I'd say having them here is ok for now and once we figure out whether all scenarios can be covered by them, move them upstream and remove from here. 

/unhold

aleskandro:
Also, this is not a fix to a problem that occurred. This is just another mitigation for edge cases, including manual deletion of the nodes by mistake.

PR #64654, opened 2025-05-07
Adding Joa to the OLS load-generator OWNERS group

vishnuchalla:
Adding Joa to openshift-lightspeed load tests OWNERS group.


JoaoFula:
Thank you!

PR #64672, opened 2025-05-08
OCPEDGE-1490: feat: add two node fencing presubmits

eggfoobar:
add presubmits to installer and origin for two node fencing

### DIFF
--- a/ci-operator/config/openshift/installer/openshift-installer-main.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-main.yaml
@@ -674,6 +674,15 @@ tests:
     cluster_profile: equinix-edge-enablement
     workflow: baremetalds-two-node-arbiter
   timeout: 5h0m0s
+- always_run: false
+  as: e2e-metal-ovn-two-node-fencing
+  cluster: build05
+  optional: true
+  run_if_changed: (baremetal|metal3|ironic)
+  steps:
+    cluster_profile: equinix-edge-enablement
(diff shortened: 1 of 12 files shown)


eggfoobar:
/pj-rehearse ack

Those failures are known for two-node-fencing are expected for the time being. They will be resolved as the feature is being worked on, the primary need now is to identify and resolve install failures.

ci-operator/config/openshift-priv/installer/openshift-priv-installer-main.yaml
```diff
@@ -675,6 +675,15 @@ tests:
     cluster_profile: equinix-edge-enablement
     workflow: baremetalds-two-node-arbiter
   timeout: 5h0m0s
+- always_run: false
```
patrickdillon, line 678: Do you have a reason to add the jobs to `openshift-priv/installer`? I'm not sure exactly what that is, but I think it's used by QE. If you have a reason to add to priv, no problem for me, but if you just want presubmits on PRs for devs, `openshift/installer` should be sufficient.

patrickdillon (commented):
job config looks good, but can you clarify whether openshift-priv is needed and remove if not?

ci-operator/config/openshift-priv/installer/openshift-priv-installer-main.yaml
eggfoobar: That's a good point, I'll remove them for now because I'm not sure they are needed yet.

PR #64273, opened 2025-04-25
Sync Serverless CI config/

serverless-qe:

PR #64680, opened 2025-05-08
Sync Serverless CI config/

serverless-qe:


cardil:
/close

It looks exacly the same as https://github.com/openshift/release/pull/64273

Maybe the job run in exact same time when the other was being merged?!? And it recreated it?

PR #64697, opened 2025-05-08
:broom: Removing test images for kn-event

cardil:
This should fix the https://github.com/openshift-knative/kn-plugin-event/pull/825


cardil:
/pj-rehearse ack

The rehearsals are expected to fail. However, this should unblock https://github.com/openshift-knative/kn-plugin-event/pull/825 and subsequent cherrypicks.

PR #64716, opened 2025-05-08
Update openshift-ovn-kubernetes-release-4.20.yaml

mohit-sheth:
Run orion with the correct version, 4.20


jtaleric (approved):
/lgtm

jtaleric:
> /hold automation should take care of these things? we don't do them manually every release..

The automation when we branch doesn't update fields from what I can tell -- we need the VERSION field to be updated to 4.20

jtaleric:
@tssurya can we get this in as it is holding up result collection in PR runs.

PR #64436, opened 2025-05-01
MULTIARCH-5326 use multi payload as initial/latest for MA 4.19 jobs

jeffdyoung:
We want to use the multi payload as the latest/initial release payloads instead of x86.
Why?
This ensures that any job in the multi-payload will use the matching installer and openshift-tests. 

The single arch arm, power, and z jobs will now use the latest multi-payload...historically speaking, multi has been closer to the single arch arm/p/z than the x86 release payload.

PR #64739, opened 2025-05-09
MULTIARCH-5326 use multi payload as initial/latest for MA 4.20

jeffdyoung:
same as https://github.com/openshift/release/pull/64436, but for 4.20


jeffdyoung:
@tvardema @deepsm007 we good with this?

PR #64745, opened 2025-05-09
NFD fallback image var updated

AdamKaabyia:
https://github.com/rh-ecosystem-edge/nvidia-ci/pull/146


wabouhamad (approved):
/lgtm

empovit:
Two jobs has failed, I think because this lined needs to be changed too
https://github.com/openshift/release/pull/64745/files#diff-ddec10796ee675e40c1a66466cb29d8678447dfa5cc5330ac1f9c2ef055e2699R39

empovit:
Also it looks like the job config generator like env variables sorted alphabetically, so renaming the variable will move it up a line (easier for humans to find when there are many vars).

PR #64746, opened 2025-05-09
WIP: test cri-o images

haircommander:


haircommander:
for posterity: we had a bunk CI image sneak in. the fix was to manually delete it. I'll monitor the new one created next week to see if it continues to be bunk .it's possible we need to update to RHEL 9.6 on them

PR #64758, opened 2025-05-10
Drop promtail metrics ingestion by random namespace

dgoodwin:
We've ingested about 12000 metrics today, 11500 were promtail ones which
are evaluating by namespace, which is a random string. Quite sure these
are unused, added by Vadim in 2021.


sosiouxme:
/lgtm
guess this is one way to find out if anything in there was being used...

PR #64795, opened 2025-05-12
CONSOLE-4393: update demo plugin to node 22

logonoff:


ci-operator/config/openshift/console/openshift-console-release-4.20.yaml
```diff
@@ -31,7 +27,7 @@ images:
       - registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.23-openshift-4.19
   to: console
 - dockerfile_path: Dockerfile.plugins.demo
-  from: nodejs-10
+  from: ocp_builder_rhel-9-base-nodejs-openshift-4.19
```
spadgett, line 30: Should this be 4.20?

logonoff: I think that's out of scope of this PR since I'm just using an existing image (which has the 4.19 tag). 

whenever the ART team runs their script to update all of the 4.20 configuration files to use the 4.20 images, that's when this becomes 4.20

cc @jhadvig

spadgett (approved):
/lgtm

PR #64800, opened 2025-05-12
AGENT-865: Run openshift conformance parallel tests for agent SNO IPV6 job

pawanpinjarkar:


pawanpinjarkar:
some failing tests from 4.19 and 4.20 rehearsals are knwon and bug is opened

example: https://issues.redhat.com/browse/OCPBUGS-55721
`[sig-network][OCPFeatureGate:RouteExternalCertificate][Feature:Router][apigroup:route.openshift.io] with valid setup the router should support external certificate and routes are reachable `

PR #64803, opened 2025-05-12
Fix rebasebot config on autoscaler and GCP CAPI

RadekManak:
Changes
- cluster-api-provider-gcp to use correct --source-repo option
- autoscaler branch to main

### DIFF
--- a/ci-operator/config/openshift-eng/rebasebot/openshift-eng-rebasebot-main.yaml
+++ b/ci-operator/config/openshift-eng/rebasebot/openshift-eng-rebasebot-main.yaml
@@ -125,7 +125,7 @@ tests:
     test:
     - as: cluster-api-provider-gcp
       commands: |
-        rebasebot --source kubernetes-sigs/cluster-api-provider-gcp \
+        rebasebot --source-repo kubernetes-sigs/cluster-api-provider-gcp \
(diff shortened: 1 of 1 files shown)


ci-operator/config/openshift-eng/rebasebot/openshift-eng-rebasebot-main.yaml
```diff
@@ -595,8 +595,8 @@ tests:
       commands: |
         UPSTREAM_VERSION=$(curl --silent --header "X-GitHub-Api-Version:2022-11-28" "https://api.github.com/repos/kubernetes/autoscaler/releases" | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/' | grep -E 'cluster-autoscaler-[0-9.]+' | sort -V | tail -n1)
         rebasebot --source https://github.com/kubernetes/autoscaler:"${UPSTREAM_VERSION}" \
```
nrb, line 597: Should this be `--source-repo`?

RadekManak: Autoscaler is not compatible with `_BUILTIN_/source-ref-hooks/latest-stable-release.sh`.  Because that repo has multiple projects. It will require its own custom --source-ref-hook. I won't be changing that in this PR.

nrb (approved):
/lgtm

PR #64853, opened 2025-05-13
ci-operator: use shard_count for e2e-aws-serial jobs on some cluster infra repos

damdo:
Adding shard_count was suggested in today's OCP Arch Call as a measure to fix the recurrent serial job timeouts.


damdo:
Discussed with @stbenjam this is a meta-change, as such fine to ack already.

/pj-rehearse ack

PR #64856, opened 2025-05-13
AGENT-1199: Set Up Periodic CI Job for OVE ISOBuilder

pawanpinjarkar:
Set Up Periodic CI Job for OVE ISOBuilder


andfasano:
I'm not sure if the current guidelines for [extending the release gates](https://docs.ci.openshift.org/docs/architecture/release-gating/#process-for-extending-the-release-gating-suite) require to create directly the periodic job within the `release` repo? Also, it should run the conformance tests. 
In general anyhow it does not seem to me that the job is in good shape yet to become a release candidate job, so I think that the `installer` repo may sound a more suitable location for the plain periodic

pawanpinjarkar:
> `installer`

conformance tests can't be run yet as we are not installing a cluster. I will try to move it to other location. Thank you

pawanpinjarkar:
> so I think that the installer repo may sound a more suitable location for the plain periodic

Instead why not agent-installer-utils 🤔

PR #64857, opened 2025-05-13
AGENT-1201: Improvements to OVE ISO CI job

pawanpinjarkar:


ci-operator/config/openshift/agent-installer-utils/openshift-agent-installer-utils-main.yaml
```diff
@@ -94,6 +94,8 @@ tests:
       DEVSCRIPTS_CONFIG: |
         AGENT_E2E_TEST_SCENARIO=COMPACT_IPV4
         AGENT_E2E_TEST_BOOT_MODE=ISO_NO_REGISTRY
+        AGENT_CLEANUP_ISO_BUILDER_CACHE_LOCAL_DEV=true
+        FIPS_MODE=true
```
andfasano, line 98: I'm not clear about the impact of the config var `FIPS_MODE` for this scenario, if it has any effect at all. I was expecting first to be digested somehow first in dev-scripts (and very likely an additional post-installation verification to ensure it was applied correctly). Did you maybe check it? If not, let's drop it to avoid an unnecessary config var (at least in this PR)

pawanpinjarkar: sure, right now it has not much impact because we are stopping once ISO is created.

PR #64729, opened 2025-05-09
Adding workflow cert rotation to other cloud platforms

gangwgr:
@vrutkovs @wangke19 Please review this pr

PR #64889, opened 2025-05-14
NO-JIRA: cert-rotation: wait for operators to stop progressing when observing the cluster

vrutkovs:
After we for a specified observe duration a new revision might have been scheduled at the very end. Instead of exiting immediately we should wait for it to complete so that gather- jobs won't fail on progressing operators


vrutkovs:
Included in https://github.com/openshift/release/pull/64729

PR #64894, opened 2025-05-14
OVNK BGP: add periodics for master branch

jcaamano:


tssurya:
/lgtm
I assume we don't want to do 4.20 here?

PR #64899, opened 2025-05-14
telco: cnf-tests: 4.19 is branched out, master is 4.20

shajmakh:
* point matser configuration to 4.20
* create 4.19 configuration
* update image mirroring urls
* run `make jobs` and  `make ci-operator-config`


shajmakh:
failure of master/e2e-aws-ci-tests is legit:
```
+ git fetch --all
+ git checkout origin/master
error: pathspec 'origin/master' did not match any file(s) known to git
```

master no longer exist anyway on any of the submodules repos. and either way the release-4.20 is already there so we should use that.

PR #64718, opened 2025-05-08
Bump e2e-benchmarking to v2.3.5 and kube-burner-ocp v1.6.4

afcollins:
Also, delete unused control-plane-smoke.

Other notable changes from [v2.3.1 to v2.3.5](https://github.com/cloud-bulldozer/e2e-benchmarking/compare/v2.3.1...v2.3.5):
* kube-burner: v1.6.3 to v1.6.4
* ingress perf: from 0.4.3 to 0.5.1

PR #64925, opened 2025-05-14
Remove openshift-qe-network-policy not used by anything else

afcollins:
Removing some dead code.


afcollins:
> no rehearsable tests are affected by this change

Closing this to lump into #64718

PR #65024, opened 2025-05-16
[cri-o/cri-o] Wait until the host is ready

bitoku:
This should fix https://github.com/cri-o/cri-o/issues/8390


bitoku:
@haircommander @saschagrunert Can you PTAL?

PR #65042, opened 2025-05-16
new config for kueue-fbc

cpmeadors:


kannon92:
The lint job failure is suspect.

PR #65031, opened 2025-05-16
shiftstack: Provide an image for ghira

pierreprinetti:

PR #65049, opened 2025-05-17
shiftstack: Ghira to use its own Jira token

pierreprinetti:
...instead of using the Bugwatcher's.


pierreprinetti:
Depends on https://github.com/openshift/release/pull/65031

PR #65053, opened 2025-05-18
ODH: Use Go 1.23 for build_root & FROM ubi9

grdryn:
We've already got a .ci-operator.yaml file in the repo, and it looks
like the intention when that was introduced was that it would be used
by default. That wouldn't be the case unless we change the
`build_root` stanza here to only include `from_repository: true`,
according to the docs. For some reason changing to that appears to
lead to an error where it can't find the .ci-operator.yaml file. Let's
deal with that in a different PR.

Additionally, the replacement of the final FROM instruction of the
Dockerfile in the image build should use a UBI image rather than a
CenOS stream one. This change updates that to a UBI9 image from an EL8
CentOS Stream image, corresponding to the changes in
https://github.com/opendatahub-io/opendatahub-operator/pull/1942.

### DIFF
--- a/ci-operator/config/opendatahub-io/opendatahub-operator/opendatahub-io-opendatahub-operator-main.yaml
+++ b/ci-operator/config/opendatahub-io/opendatahub-operator/opendatahub-io-opendatahub-operator-main.yaml
@@ -1,17 +1,17 @@
 base_images:
-  os:
-    name: centos
-    namespace: origin
-    tag: stream8
+  ubi:
+    name: ubi-minimal
+    namespace: ocp
+    tag: "9"
 build_root:
   image_stream_tag:
     name: release
     namespace: openshift
-    tag: golang-1.22
+    tag: golang-1.23
 images:
 - context_dir: .
   dockerfile_path: Dockerfiles/Dockerfile
-  from: os
+  from: ubi
   to: opendatahub-operator
 operator:
   bundles:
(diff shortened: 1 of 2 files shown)


grdryn:
/pj-rehearse pull-ci-opendatahub-io-opendatahub-operator-rhoai-images

grdryn:
/cc @zdtsw 

I believe this should solve the issue seen in https://github.com/opendatahub-io/opendatahub-operator/pull/1942

grdryn:
/pj-rehearse pull-ci-opendatahub-io-opendatahub-operator-rhoai-opendatahub-operator-pr-image-mirror

grdryn:
/pj-rehearse pull-ci-opendatahub-io-opendatahub-operator-rhoai-opendatahub-operator-e2e

ci-operator/config/opendatahub-io/opendatahub-operator/opendatahub-io-opendatahub-operator-incubation.yaml
```diff
@@ -1,17 +1,17 @@
 base_images:
-  os:
-    name: centos
-    namespace: origin
-    tag: stream8
+  ubi:
```
zdtsw, line 2: lets keep incubation as-is.
we are not using this branch any more, but if we need any backport fix for old EUS release, we might still need use this branch

zdtsw (approved):

grdryn:
/pj-rehearse ack

Rehearsals were run on previous iteration, and all that has changed is that the changes were dropped for the incubation branch, so the set of rehearsals previously run should cover this iteration sufficiently also.

PR #65057, opened 2025-05-19
Add configs for build09

liangxia:


liangxia:
The failure in `ci/prow/generated-config` is tracked in https://github.com/openshift/release/pull/65058

PR #65075, opened 2025-05-19
snc: Increase timeout from 4h to 6h

praveenkumar:


gbraad (approved):
/approve

praveenkumar:
/close

looks like error is not with timeout but because of https://github.com/crc-org/snc/issues/1068 one.

PR #65138, opened 2025-05-20
Add build10 pull secret for equinix-metal-qe and hcp

sgoveas:


sgoveas:
@liangxia please review

PR #65151, opened 2025-05-20
Fix: export CLUSTER_ID in osde2e-test to resolve CAD test failure.

Amarthya-v:
This PR fixes a failure in the Configuration Anomaly Detection (CAD) tests by ensuring that the CLUSTER_ID environment variable is properly set before executing the osde2e test.

the test would fail with:
'''CLUSTER_ID must be set
Expected <string>: not to be empty'''

To resolve this, the osde2e-test-commands.sh script has been updated to extract CLUSTER_ID.

Fixes: [SREP-358](https://issues.redhat.com//browse/SREP-358)

### DIFF
--- a/ci-operator/step-registry/osde2e/test/osde2e-test-commands.sh
+++ b/ci-operator/step-registry/osde2e/test/osde2e-test-commands.sh
@@ -21,6 +21,15 @@ then
    export TEST_KUBECONFIG="${SHARED_DIR}/kubeconfig"
 fi
 
+if [[ -f "${SHARED_DIR}/cluster-id" ]]; then
+    export CLUSTER_ID="$(< "${SHARED_DIR}/cluster-id")"
+fi
+
+if [[ -z "${CLUSTER_ID:-}" ]]; then
+    echo "CLUSTER_ID is not set. Aborting test execution."
+    exit 1
+fi
+
 export REPORT_DIR="${ARTIFACT_DIR}"
 
 /osde2e test --configs "${CONFIGS}" \


ci-operator/step-registry/osde2e/test/osde2e-test-commands.sh
```diff
@@ -21,7 +21,15 @@ then
    export TEST_KUBECONFIG="${SHARED_DIR}/kubeconfig"
 fi
 
+echo "Extracting CLUSTER_ID..."
```
btoll, line 24: Why `echo` this?  None of the other `exports` are logging anything.  Let's maintain this consistency.

ci-operator/step-registry/osde2e/test/osde2e-test-commands.sh
```diff
@@ -21,7 +21,15 @@ then
...
+export CLUSTER_ID="$(oc get clusterversion --kubeconfig="${TEST_KUBECONFIG:-$KUBECONFIG}" -o jsonpath='{.spec.clusterID}')"
```
btoll, line 25: Will the environment within which this is running have access to the `oc` CLI tool?  If yes, does the installed version support all of the parameters you're using?

Amarthya-v: Yes, the environment running this step has access to the oc CLI  it’s part of the base image used for this job. I also verified that the installed version supports the --kubeconfig and -o jsonpath parameters. This command has been working reliably in my local.

PR #65198, opened 2025-05-21
Move cli-ocm into oc

danilo-gemoli:
We have been building `cli-ocm` via a `BuildConfig` and push it into `registry.ci.openshift.org`.

Folks use this image into their `ci-operator` configurations like so:
```yaml
base_images:
  cli-ocm:
    name: cli-ocm
    namespace: ci
    tag: latest
```
and the image gets resolved from `quay.io/openshift/ci:ci_cli-ocm_latest`.

This means that we have to trigger a `BuildConfig` and then manually push the outcome into `quay.io` manually, every time someone needs a new version of this image.

By moving it into the `ci-operator`'s `openshift/oc` configuration, the whole mechanism of building and pushing will be handled entirely by `ci-operator`.

~~As a final yet important note: **this PR isn't enough by itself**.~~

~~`ci-operator` is going to promote the image to:~~
```
quay.io/openshift/ci:ocp_4.20_cli-ocm
```
~~whereas the `base_images` (see the snipped right above) pulls it from:~~
```
quay.io/openshift/ci:ci_cli-ocm_latest
```

~~If we want to go down this route we have to follow this up with another PR and update a bunch of `ci-operator` configurations, such as the `base_images` would look like so:~~
```yaml
base_images:
  cli-ocm:
    name: "4.20"
    namespace: ocp
    tag: cli-ocm
```

### DIFF
--- a/ci-operator/config/openshift/oc/openshift-oc-master__supplemental-images.yaml
+++ b/ci-operator/config/openshift/oc/openshift-oc-master__supplemental-images.yaml
@@ -19,21 +19,21 @@ images:
       as:
       - registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.23-openshift-4.19
   to: cli
-- from: cli
-  dockerfile_literal: |
+- dockerfile_literal: |
     FROM cli
     RUN set -eo pipefail
     RUN yum -y install yum-plugin-copr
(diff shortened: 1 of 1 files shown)


danilo-gemoli:
/pj-rehearse pull-ci-openshift-oc-master-supplemental-images-images

PR #64522, opened 2025-05-05
OSASINFRA-3795: Set shiftstack periodic jobs to use openstack-vh-bm-rhos static pool cluster profile

MiguelCarpio:
**What this PR does?**

Let's PROW automatically redistribute the shiftstack periodic jobs across the following clouds using the `openstack-vh-bm-rhos` resource pool cluster-profile: 

https://docs.ci.openshift.org/docs/architecture/quota-and-leases/#configuration-for-heterogeneous-resources

```
'openstack-vh-bm-rhos-quota-slice': {
        'openstack-hwoffload': 3,
        'openstack-vh-mecha-central': 3,
        'openstack-vh-mecha-az0': 3,
    },
```

 **Why we need it?**

Currently, we are manually redistributing the shiftstack periodic jobs on hwoffload, mecha-central and mecha-az0 RHOSP clouds. After redeploying mecha-z0, mecha-central and hwoffload clouds with the same ci-config (general.yaml) to reach the last OSP 17.1.4 release, it is possible to redistribute the jobs automatically with PROW using the `openstack-vh-bm-rhos` resource pool cluster-profile. So, this PR is actually needed to avoid the manual redistribution of shifstack periodic jobs across the RHOSP clouds.

PR #65199, opened 2025-05-21
OpenStack: Fix LEASED_RESOURCE unbound variable

MiguelCarpio:
What this PR does?

Fix LEASED_RESOURCE unbound variable error when a job is configured to use a cluster profile instead of the `openstack-vh-bm-rhos` static resource pool cluster-profile. 

Why we need it?
To support periodic and presubmit jobs that are not configured to use the `openstack-vh-bm-rhos` resource pool cluster-profile.

It seems that when a job is configured to use a cluster profile that is not a resource pool, the LEASED_RESOURCE variable doesn't have a value, in contrast to the tests in https://github.com/openshift/release/pull/64522 where if the cluster profile doesn't have a LEASED_RESOURCE, the LEASED_RESOURCE will be a number ID.


MiguelCarpio:
/pj-rehearse periodic-ci-shiftstack-ci-main-quotas-openstack-hwoffload

PR #65211, opened 2025-05-21
cri-o: Add release and presubmit jobs for CRI-O versions 1.33

bitoku:
Adding jobs for 1.33.

ref:
- https://github.com/openshift/release/pull/60098

### DIFF
--- a/ci-operator/config/cri-o/cri-o/cri-o-cri-o-release-1.33.yaml
+++ b/ci-operator/config/cri-o/cri-o/cri-o-cri-o-release-1.33.yaml
@@ -1,0 +1,68 @@
+base_images:
+  base-machine:
+    name: centos
+    namespace: openshift
+    tag: stream9
+  rhel-coreos-9-base:
+    name: "4.20"
+    namespace: ocp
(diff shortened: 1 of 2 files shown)


ci-operator/config/cri-o/cri-o/cri-o-cri-o-release-1.33.yaml
```diff
@@ -0,0 +1,68 @@
...
+    tag: stream9
+  rhel-coreos-9-base:
+    name: "4.20"
+    namespace: ocp
+    tag: rhel-coreos-9
+build_root:
+  image_stream_tag:
+    name: release
+    namespace: openshift
+    tag: rhel-9-release-golang-1.23-openshift-4.19
```
bitoku, line 14: This should be golang-1.24-openshift-4.20, but it's not available yet.
Since the main branch is still using this tag, it should be good for now.

sohankunkerkar (approved):
Just thought ..I think we can automate this presubmit job creations as part of the release process.

/lgtm

PR #65218, opened 2025-05-21
CMP-3363: Manually cleanup skip_if_only_changed from FIO ROSA jobs

rhmdnd:
https://github.com/openshift/release/pull/64248 removed the
skip_if_only_changed config, so that the e2e-rosa job was only invoked
on PRs with explicit `/test` commands. However, the actual job
configuration wasn't updated, meaning the job is still running on all
PRs proposed that don't match the regular expression.

This commit explicitly removes the skip_if_only_changed from the job
definition so we don't spin up those tests unless we run the `/test
e2e-rosa` command, which is what we want at this point since rosa
support for FIO is still under development.


Vincent056 (approved):
/lgtm

PR #65220, opened 2025-05-21
ensure ci is triggered on changes to .sh files

weshayutin:
in oadp we have some settings files under tests/e2e/scripts that should trigger ci when changed.


kaovilai (approved):
This would also match `.sh` without name.. but prob not a concern.
/lgtm

PR #65221, opened 2025-05-21
Skip FIO e2e for specific changes

rhmdnd:
Let's not run the e2e testing if the change is specific to .github CI,
tekton pipelines, documentation changes, or repository management
changes.

This makes it easier and more efficient to iterate on Konflux changes
since we don't use the Konflux build output in CI runs, yet.


Vincent056 (approved):
/lgtm

rhmdnd:
Added a last minute change for the bundle-hack directory, which Konflux bumps regularly for new operator digests.

PR #65261, opened 2025-05-22
OADP project rename of the default branch naming

mpryc:
Rename of the default branch to be oadp-dev across OADP plugins and relevant projects.

This PR is for the [openshift/oadp-operator repository](https://github.com/openshift/oadp-operator).

GitHub has the "rename" option within the settings for each repository.

Using rename instead of creating a branch off the current default allows to:

Update pull requests targeting this branch across other repositories.
Update branch protection rule that explicitly targets current default branch.
Settings->General (and the rename icon).


kaovilai:
My personal pref would be moving towards `main`..  even if it's the same name with upstream default branch names.

It's not that hard to
```
git checkout -b main_upstream upstream/main

git checkout -b main_origin origin/main
# or if you prefer to keep 'main' for your fork:
# git checkout main
# git branch --set-upstream-to=origin/main
```

PR #65374, opened 2025-05-27
use multi payload for logging arm64 jobs

anpingli:
idp-passwd  step in pending status when the cluster type is arm64. this PR fix this issue.


anpingli:
@QiaolingTang  @Phaow  PTAL

anpingli:
@liangxia  can you approve it?

PR #65398, opened 2025-05-27
NO-ISSUE: Presubmit job to test bad_dns test case in dev-scripts

rwsu:


rwsu:
The rehearsal failure is expected as a similar job is already intermittently failing. 
/pj-rehearse skip

PR #65416, opened 2025-05-28
Refine regexp in skip_if_only_changed

liangxia:
Some of the `()` are unnecessary, so removed. Also add `^` for exact match of the folders in the repo root.

Reference doc, https://docs.prow.k8s.io/docs/jobs/#triggering-jobs-based-on-changes


liangxia:
This is the regexp that determines whether we should run `e2e-aws` tests for PRs in openshift-tests-private repo.  I don't think it actually affects the tests in `e2e-aws`. So skipping the tests.

/pj-rehearse skip

PR #65424, opened 2025-05-28
OCPQE-29800: update cluster infra tests-private test image tags

sunzhaohua2:
Since the master branch has already switch to sync with 4.20, we need to change the master/release-4.20 configs' the test image tag to 4.20.
Also corrected other tags.
@huali9 @miyadav PTAL

### DIFF
--- a/ci-operator/config/openshift-priv/cloud-provider-gcp/openshift-priv-cloud-provider-gcp-main.yaml
+++ b/ci-operator/config/openshift-priv/cloud-provider-gcp/openshift-priv-cloud-provider-gcp-main.yaml
@@ -22,7 +22,7 @@ base_images:
   tests-private:
     name: tests-private
     namespace: ci
-    tag: latest
+    tag: "4.20"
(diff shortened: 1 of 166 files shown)


sunzhaohua2:
> For 4.21 we use 4.20?

Yes, confirmed with @liangxia  codes in branch 4.20, 4.21, and master we use use 4.20 currently

damdo (commented):
/approve

@sunzhaohua2  Do we need to revert this later or will this stay like this?

How will it work with the `-main.yaml` files, how will they follow the latest images when 4.21 is available?

sunzhaohua2:
> /approve
> 
> @sunzhaohua2 Do we need to revert this later or will this stay like this?
> 
> How will it work with the `-main.yaml` files, how will they follow the latest images when 4.21 is available?

@damdo I created a [card](https://issues.redhat.com/browse/OCPQE-29800) for 4.20 to update the image tag,  in the future whenever there is a new branch, I will update the image version in `-main.yaml` files

PR #65516, opened 2025-05-30
install diffutils after removing golang

rccrdpccl:
diffutils is a dependency of golang, and would be removed


carbonin:
Didn't we fix this already in https://github.com/openshift/release/pull/64445? How did it regress?

PR #65526, opened 2025-05-30
add checks and required env vars

deads2k:


deads2k:
down to six failed. We're taking it.

PR #65530, opened 2025-05-30
NO-ISSUE: (cleanup): Remove openshift-operator-framework-operator-controller-main__periodics since the latest release version has the same test

camilamacedo86:
We do not need the main configuration since it is the same of the latest release version that we have ( in this case, now 4.20). This PR remove its.


neisw:
@oceanc80 could use your approval.  We don't need the periodics in main, just the release specific configs.

PR #65536, opened 2025-05-30
Add techpreview/fix skips for vSphere MAO jobs

jcpowermac:
This PR
- adds serial and parallel techpreview jobs
- fixes vsphere skips


jcpowermac:
/hold

going to add serial as suggested by @vr4manta

PR #65544, opened 2025-06-01
feat: Add new CI job which runs daily on nightly build for a downstream version

siddardh-ra:
I've created a downstream job on the nightly build which runs daily. I've also removed the upstream jobs, because they're broken and don't work currently. Once we have a enough data for CPT , we can adjust the frequency of the job accordingly


jhutar (approved):
/lgtm

PR #65552, opened 2025-06-02
[perf-scale-openshift-psap/cluster-pool] Set the pool size to 0

kpouget:
We don't need anymore to have 2 clusters ready at anytime.


kpouget:
/cc @albertoperdomo2
can you please LGTM this PR?

PR #65562, opened 2025-06-02
`secret-manager`: Implement index secret

psalajova:
Implements `list` functionality using the `secret-index` secret which will be created for each collection and which will contain a list of all secrets in that collection. This is done so that we don't have to grant users project-wide access to list secrets (see [https://issuetracker.google.com/issues/259677917?utm_source=chatgpt.com&pli=1](https://www.google.com/url?q=https://issuetracker.google.com/issues/259677917?utm_source%3Dchatgpt.com%26pli%3D1&sa=D&source=docs&ust=1748940242741941&usg=AOvVaw1exSskA5-RlW1Rp1ZgnbXV)). Necessary changes were made also to `create`, `update` and `delete` functions, so the index is always up to date.

### DIFF
--- a/hack/gcp-secret-manager/commands/create.py
+++ b/hack/gcp-secret-manager/commands/create.py
@@ -77,7 +77,7 @@ def create(collection: str, secret: str, from_file: str, from_literal: str):
     index_secrets = get_secrets_from_index(client, collection)
     if secret in index_secrets:
         raise click.ClickException(
-            f"Secret '{secret}' already exists in collection '{collection}'."
+            f"Secret '{secret}' already exists within the collection index."
         )
     try:
         client.get_secret(name=client.secret_path(PROJECT_ID, secret_name))


hack/gcp-secret-manager/util.py
```diff
@@ -165,3 +181,66 @@ def check_if_collection_exists(collection: str) -> bool:
...
+
+    Args:
+        client (secretmanager.SecretManagerServiceClient): Secret Manager client.
+        collection (str): Name of the collection.
+
+    Returns:
+        List[str]: A list of secrets.
+    """
+    index_secret = client.secret_version_path(
+        PROJECT_ID, get_secret_name(collection, INDEX_SECRET_NAME), "latest"
```
danilo-gemoli, line 222: `get_secret_name` returns `$COLLECTION__index` rather than `$COLLECTION_____index`, as per design doc.

hack/gcp-secret-manager/util.py
```diff
@@ -7,15 +7,23 @@
...
+from google.cloud import secretmanager
 
 # Test Platform's project in GCP Secret Manager
 PROJECT_ID = "openshift-ci-secrets"
 
 # The YAML config which defines which groups have access to which secret collections.
 CONFIG_PATH = "https://raw.githubusercontent.com/openshift/release/master/core-services/sync-rover-groups/_config.yaml"
 
+# The string reserved for the index secret associated with each collection.
+INDEX_SECRET_NAME = "secret-index"
```
danilo-gemoli: Maybe set this to `__secret-index`? (or just `__index` in accordance with the design doc) See the comment below.

hack/gcp-secret-manager/commands/update.py
```diff
@@ -47,23 +49,60 @@ def update(collection: str, secret: str, from_file: str, from_literal: str):
...
+    secret_path: str,
+):
+    """
+    Ensures the secret exists both in GSM and the index, or handle inconsistencies.
+    Raises ClickException if update is not possible.
+    """
+    index_secrets = get_secrets_from_index(client, collection)
+    secret_in_index = secret in index_secrets
+
+    try:
```
danilo-gemoli, line 59: Consider introducing a new function `def secret_exists(client, secret) -> bool`

danilo-gemoli:
If my understanding is correct, we might end up having dangling secrets when a user creates a new one:
1. The secret `collection1__secret1` is created in GSM.
2. `secrets-manager` fails to update the index `collection1____index`.
    2a.  The error gets reported to the user but the secret is still there.
3. `collection1__secret1` is now a dangling secret.

Specular scenario in case of an "index-update miss" on deletion: the secret gets deleted but it's still on the index.
This is interesting:
1. The user tries to delete `collection1__secret1`.
2. The secret gets deleted in GSM but we have an index-update miss.
3. `secrets-manager` reports the failure.
4. The user tries once again to delete it but he hits another error because the secret isn't there anymore.

psalajova:
> If my understanding is correct, we might end up having dangling secrets when a user creates a new one:
> 
> 1. The secret `collection1__secret1` is created in GSM.
> 2. `secrets-manager` fails to update the index `collection1____index`.
>    2a.  The error gets reported to the user but the secret is still there.
> 3. `collection1__secret1` is now a dangling secret.
> 
> Specular scenario in case of an "index-update miss" on deletion: the secret gets deleted but it's still on the index. This is interesting:
> 
> 1. The user tries to delete `collection1__secret1`.
> 2. The secret gets deleted in GSM but we have an index-update miss.
> 3. `secrets-manager` reports the failure.
> 4. The user tries once again to delete it but he hits another error because the secret isn't there anymore.

I updated the code, hopefully with better handling of these cases. It's tricky however, making sure we always have a consistent state between the index and GCP secrets... if you have any recommendations on how to better handle this, pls let me know.

hack/gcp-secret-manager/commands/create.py
```diff
@@ -60,15 +62,31 @@ def create(collection: str, secret: str, from_file: str, from_literal: str):
...
         )
     client = secretmanager.SecretManagerServiceClient()
-    check_if_secret_already_exists(collection, secret, client)
+    secret_name = get_secret_name(collection, secret)
 
+    # Check if secret exists in either index or GCP
+    index_secrets = get_secrets_from_index(client, collection)
+    if secret in index_secrets:
+        raise click.ClickException(
+            f"Secret '{secret}' already exists in collection '{collection}'."
```
danilo-gemoli: I'd change the error message to `... exists within the collection index ...` to distinguish from the one that comes right after this.

hack/gcp-secret-manager/commands/create.py
```diff
@@ -90,30 +108,62 @@ def create(collection: str, secret: str, from_file: str, from_literal: str):
...
+                f"Failed to create secret '{secret}': {e}. "
+                f"Also failed to clean up: {cleanup_error}. "
+                "Please contact Test platform to clean up manually."
+            ) from e
+
         raise click.ClickException(f"Failed to create secret '{secret}': {e}") from e
 
 
-def check_if_secret_already_exists(
+def check_and_handle_existing_secret(
```
danilo-gemoli: Where is this function used?

psalajova: Removed it

PR #65568, opened 2025-06-02
CORS-4069: Add test for nat gateway support

rna-afk:
Adding CI test to set the install config outbound type to nat gateway to test nat gateway support.

### DIFF
--- a/ci-operator/config/openshift/installer/openshift-installer-main.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-main.yaml
@@ -1423,6 +1423,14 @@ tests:
       CHECK_MOD_LIST: "false"
     test:
     - ref: go-verify-deps
+- always_run: false
+  as: e2e-azure-nat-gateway-single-zone
+  steps:
+    cluster_profile: azure4
+    env:
+      OUTBOUND_TYPE: NATGatewaySingleZone
+    workflow: openshift-e2e-azure
+  timeout: 6h0m0s
 zz_generated_metadata:
   branch: main
   org: openshift
--- a/ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh
+++ b/ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh
@@ -103,7 +103,7 @@ EOF
 if [ -z "${OUTBOUND_TYPE}" ]; then
   echo "Outbound Type is not defined"
 else
-  if [ X"${OUTBOUND_TYPE}" == X"UserDefinedRouting" ] || [ X"${OUTBOUND_TYPE}" == X"NatGateway" ]; then
+  if [ X"${OUTBOUND_TYPE}" == X"UserDefinedRouting" ] || [ X"${OUTBOUND_TYPE}" == X"NATGatewaySingleZone" ] || [ X"${OUTBOUND_TYPE}" == X"NatGateway" ]; then
     echo "Writing 'outboundType: ${OUTBOUND_TYPE}' to install-config"
     PATCH="${SHARED_DIR}/install-config-outboundType.yaml.patch"
     cat > "${PATCH}" << EOF
--- a/ci-operator/jobs/openshift/installer/openshift-installer-main-presubmits.yaml
+++ b/ci-operator/jobs/openshift/installer/openshift-installer-main-presubmits.yaml
@@ -2773,6 +2773,80 @@ presubmits:
     - ^main$
     - ^main-
     cluster: build01
+    context: ci/prow/e2e-azure-nat-gateway-single-zone
+    decorate: true
+    decoration_config:
+      timeout: 6h0m0s
+    labels:
+      ci-operator.openshift.io/cloud: azure4
+      ci-operator.openshift.io/cloud-cluster-profile: azure4
+      ci.openshift.io/generator: prowgen
+      pj-rehearse.openshift.io/can-be-rehearsed: "true"
+    name: pull-ci-openshift-installer-main-e2e-azure-nat-gateway-single-zone
+    rerun_command: /test e2e-azure-nat-gateway-single-zone
+    spec:
+      containers:
+      - args:
+        - --gcs-upload-secret=/secrets/gcs/service-account.json
+        - --image-import-pull-secret=/etc/pull-secret/.dockerconfigjson
+        - --lease-server-credentials-file=/etc/boskos/credentials
+        - --report-credentials-file=/etc/report/credentials
+        - --secret-dir=/secrets/ci-pull-credentials
+        - --target=e2e-azure-nat-gateway-single-zone
+        command:
+        - ci-operator
+        image: ci-operator:latest
+        imagePullPolicy: Always
+        name: ""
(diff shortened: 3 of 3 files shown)


ci-operator/step-registry/ipi/conf/azure/nat-gateway/ipi-conf-azure-nat-gateway-ref.yaml
```diff
@@ -0,0 +1,14 @@
...
+    namespace: ocp
+    name: "4.20"
+    tag: installer
+  commands: ipi-conf-azure-nat-gateway-commands.sh
+  resources:
+    requests:
+      cpu: 10m
+      memory: 100Mi
+  documentation: |-
+    The IPI Azure configure step sets the nat gateway field in th
```
barbacbd:
```suggestion
    The IPI Azure configure step sets the nat gateway field in the install config.
```

ci-operator/step-registry/openshift/e2e/azure/nat-gateway/openshift-e2e-azure-nat-gateway-workflow.yaml
```diff
@@ -0,0 +1,14 @@
...
+    pre:
+    - ref: ipi-conf-azure-nat-gateway
+    - chain: ipi-azure-pre
+    test:
+    - ref: openshift-e2e-test
+    post:
+    - chain: gather-core-dump
+    - chain: ipi-azure-post
+  documentation: |-
+    The Openshift E2E Azure workflow executes the common end-to-end test suite on Azure with a default cluster configuration.
```
barbacbd: This description should be more specific to this step.

ci-operator/step-registry/ipi/conf/azure/nat-gateway/ipi-conf-azure-nat-gateway-ref.yaml
```diff
@@ -0,0 +1,14 @@
...
+    install config.
```
barbacbd:
```suggestion
```

barbacbd:
Looks like `make ci-operator-config` may solve the test issue above @rna-afk.

ci-operator/step-registry/openshift/e2e/azure/nat-gateway/openshift-e2e-azure-nat-gateway-workflow.yaml
```diff
@@ -0,0 +1,14 @@
+workflow:
+  as: openshift-e2e-azure-nat-gateway
+  steps:
+    allow_best_effort_post_steps: true
+    pre:
+    - ref: ipi-conf-azure-nat-gateway
```
jinyunma: Field `outboundType` can be specified via ENV `OUTBOUND_TYPE` in step [`ipi-conf-azure-commands.sh`](https://github.com/openshift/release/blob/master/ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh#L103-L118), so just set this ENV in the workflow instead of introducing a new step here to configure the `NatGateway` outboundType.

ci-operator/config/openshift/installer/openshift-installer-main.yaml
```diff
@@ -1350,6 +1350,12 @@ tests:
       CHECK_MOD_LIST: "false"
     test:
     - ref: go-verify-deps
+- always_run: false
+  as: e2e-azure-nat-gateway-single-zone
+  steps:
```
patrickdillon: I just discovered what jinyun already said in his comment: https://github.com/openshift/release/pull/65568/files#r2122767446

So you can switch the workflow to use a standard azure workflow and add an env var here in the config for `OUTBOUND_TYPE` .

patrickdillon:
Need to run `make jobs` and recommit the generated job.

ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh
```diff
@@ -103,7 +103,7 @@ EOF
 if [ -z "${OUTBOUND_TYPE}" ]; then
   echo "Outbound Type is not defined"
 else
-  if [ X"${OUTBOUND_TYPE}" == X"UserDefinedRouting" ] || [ X"${OUTBOUND_TYPE}" == X"NatGateway" ]; then
+  if [ X"${OUTBOUND_TYPE}" == X"UserDefinedRouting" ] || [ X"${OUTBOUND_TYPE}" == X"NATGatewaySingleZone" ]; then
```
jinyunma: Not sure if outbound type `NatGateway` is still the valid value in the future? QE currently has job `azure-ipi-nat-gateway-tp` on version 4.14 ~ 4.16, where outbound type is set to `NatGateway`. This change will break those jobs.

How about to keep `NatGateway` here and append new type `NATGatewaySingleZone`? Or we don't need to run that job any more as it does not work on 4.17+ so far?
```
if [ X"${OUTBOUND_TYPE}" == X"UserDefinedRouting" ] || [ X"${OUTBOUND_TYPE}" == X"NATGateway" ] || [ X"${OUTBOUND_TYPE}" == X"NATGatewaySingleZone" ]; then
```

rna-afk: I'm not sure if we are using NATGateway option in CAPZ. All I see is a feature gate here and we might need to update those tests anyway. @patrickdillon thoughts?

patrickdillon: I did not realize we landed an actual NAT gateway implementation in 4.14-4.16, but I see that we did in https://github.com/openshift/installer/pull/6933. 

Let's do your suggestion to append the new type and keep the old type as well. @rna-afk can you also add a comment to indicate that `NATGateway` is the valid value for 4.14-4.16 tech preview?

> Not sure if outbound type NatGateway is still the valid value in the future?

No, we will no longer accept `NATGateway` in the install config. Instead it will broken up into two choices:

- NATGatewaySingleZone
- NATGatewayHighAvailability (or something like that, I don't think we've settled on this one). 

The 4.14-4.16 `NATGateway` option would correspond to `NATGatewaySingleZone` but we don't want users to pick that simply by default, so we are switching to more expressive names.

jinyunma: @patrickdillon got it, thanks for detailed explanation.

patrickdillon:
/pj-rehearse pull-ci-openshift-installer-main-e2e-azure-nat-gateway-single-zone

ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh
patrickdillon: bumping this. @rna-afk can you keep NATGateway as an option?

rna-afk: Ugh changes got pushed to another branch. Fixed it

PR #65619, opened 2025-06-03
NETOBSERV-2212: catSrc env name update

Amoghrd:


ci-operator/step-registry/openshift-extended/web-tests/openshift-extended-web-tests-ref.yaml
```diff
@@ -27,7 +27,7 @@ ref:
   - name: CYPRESS_NOO_CATALOG_SOURCE
     default: ""
     documentation: Specific to Network Observability, to specify CatalogSource for NOO
-  - name: MULTI_STAGE_OVERRIDE_CYPRESS_NOO_CS_IMAGE
+  - name: CYPRESS_MULTISTAGE_PARAM_OVERRIDE_NOO_CS_IMAGE
```
memodi: this can't be prefixed with `CYPRESS`

Amoghrd: okay so `MULTISTAGE_PARAM_OVERRIDE_CYPRESS_NOO_CS_IMAGE` is right?

PR #65630, opened 2025-06-03
stackrox: perf&scale scanner-v4 @weekly

davdhacs:
Set as periodic `cron: 0 6 * * Mon` (6am utc each Monday) instead of every nightly commit. 

This is to reduce the data uploaded to elasticsearch until we use the data more by enabling anomaly detection (with orion) and adding dashboards or summary reports.

/cc @mtodor @tommartensen


tommartensen (approved):
/lgtm

PR #65635, opened 2025-06-03
METAL-1348: Support arm64 image build for ofcir

deepsm007:
/cc @tdomnesc


tdomnesc:
@deepsm007 - I think you forgot to run `make jobs`

PR #65672, opened 2025-06-04
Bump up devfile registry checks to use go 1.23

thepetk:
Bumps up the go version of registry checks to use go 1.23

https://github.com/devfile/registry/pull/611 needs to be merged first.


thepetk:
@michael-valdron PTAL

michael-valdron (approved):
/lgtm

PR #65685, opened 2025-06-04
jira-lifecycle-plugin: add release notes config for main

AlexNPavel:
Add the release notes requirement configs for `main` and `master` branches. The requirement existed as a default in the config, but `main` and `master` explicitly exclude defaults, so those did not apply.


bradmwilliams:
We definitely don't want this to merge until we get confirmation from management.

PR #65694, opened 2025-06-04
konflux: increase machine size in us-west-2

gbenhaim:
When running the e2e tests on a cluster that was using m6i.2xlarge, the test for running a build pipeline failed because the was no available node (because of lack of resources) for running the build.

Thus, usage large instances (m6i.4xlarge).


gbenhaim:
@psturc @flacatus can you please review?

PR #65718, opened 2025-06-05
ci-operator: coreos-assembler: update skip job list

joelcapitao:
We don't want to run those jobs when change occured in .tekton/ directory.


jbtrystram (approved):
LGTM

PR #65734, opened 2025-06-05
[CVP-4420] Update the region of 4.20 clusterpools in CVP

samvarankashyap:
$Subject


asergienk:
you also need to update these file:
https://github.com/openshift/release/blob/master/clusters/hosted-mgmt/hive/pools/cvp/install-config-aws-eu-west-2_secret.yaml#L36
https://github.com/openshift/release/blob/master/clusters/hosted-mgmt/hive/pools/cvp/fips-install-config-aws-eu-west-2_secret.yaml#L37

OpenShiftSDN -> OVNKubernetes

PR #65763, opened 2025-06-06
Use custom default region rather than LEASED_RESOURCE

yuwang-RH:
some cluster profile which runs rosa step is using dynamic lease quota, the LEASED_RESOURCE is not a valid readable region, like '[d0ce4c9f-199c-4a6f-8bb8-13c696f78ebe] '.
Because rosacli jobs don't use the region defined by job, step using LEASED_RESOURCE to config aws in steps.

### DIFF
--- a/ci-operator/step-registry/rosa/cluster/upgrade/hcp/rosa-cluster-upgrade-hcp-commands.sh
+++ b/ci-operator/step-registry/rosa/cluster/upgrade/hcp/rosa-cluster-upgrade-hcp-commands.sh
@@ -15,7 +15,7 @@ source ./tests/prow_ci.sh
 
 # functions are defined in https://github.com/openshift/rosa/blob/master/tests/prow_ci.sh
 #configure aws
-aws_region=${REGION:-$LEASED_RESOURCE}
+aws_region=${REGION:-us-east-2}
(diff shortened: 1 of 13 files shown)


yuwang-RH:
/pj-rehearse periodic-ci-openshift-rosa-master-e2e-rosa-hcp-advanced-critical-high-f3

ci-operator/step-registry/rosa/cluster/upgrade/hcp/rosa-cluster-upgrade-hcp-commands.sh
```diff
@@ -15,7 +15,8 @@ source ./tests/prow_ci.sh
 
 # functions are defined in https://github.com/openshift/rosa/blob/master/tests/prow_ci.sh
 #configure aws
-aws_region=${REGION:-$LEASED_RESOURCE}
+DEFAULT_AWS_REGION="us-east-2"
```
xueli181114: I would suggest not to use `DEFAULT_AWS_REGION` in case it would cause chaos of env variable overriding to aws/rosa client we are using in the code.

yuwang-RH:
/pj-rehearse periodic-ci-openshift-rosa-master-e2e-rosa-hcp-advanced-critical-high-f3

PR #65885, opened 2025-06-10
Publish the Stolostron Go 1.24 image builder

vthapar:


skitt:
Is this really needed? See #64555

PR #65915, opened 2025-06-10
setup image-mirroring for assisted-installer version v2.42

danmanor:


danmanor:
@giladravid16 @linoyaslan  and I recently found out the the image mirroring is not actually required (and even better not to do it). So we should probably remove it from the automation. I am closing this pull request for now.

PR #65973, opened 2025-06-11
OCPBUGS-57330: external-dns-operator: Switch cluster profile back to aws for shared VPC test

alebedev87:
`aws-2` is missing the secret for the shared vpc (`awscred_shared_account`). This PR switches the cluster profile for the shared VPC job back to `aws`.


alebedev87:
Taken over by https://github.com/openshift/release/pull/66025.

PR #65979, opened 2025-06-11
albo: Update owners for aws-load-balancer step

alebedev87:
Remove users who are no longer in the NI&D team, and add new members of the NI&D team.


alebedev87:
/pj-rehearse ack

Just an update of OWNERS, no rehearsal is needed.

PR #65988, opened 2025-06-11
WINC-1395:  Remove unsupported 4.12 version config and jobs

mansikulkarni96:
This commit removes all files including perioidics, pre-submits and post-submits for the unsupported version 4.12.


ci-operator/config/openshift/windows-machine-config-operator/openshift-windows-machine-config-operator-release-4.12.yaml
```diff
@@ -1,261 +0,0 @@
-base_images:
```
sebsoto: We still want to keep this one, only the community version should be removed.

sebsoto (changes requested):

PR #65949, opened 2025-06-11
phase-1: restrict the usage of aws qe cluster profiles

jianlinliu:
As phase-1, we are going to restrict the usage of aws qe cluster profiles for the existing ci jobs. 

Ideally the aws qe cluster is only used for qe's ci jobs. Later we are trying to migrate those non-qe ci jobs to other cluster profiles as phase-2.

PR #66006, opened 2025-06-12
phase-1: restrict the usage of azure qe cluster profiles

jianlinliu:
Similar to https://github.com/openshift/release/pull/65949

### DIFF
--- a/ci-operator/step-registry/cluster-profiles/cluster-profiles-config.yaml
+++ b/ci-operator/step-registry/cluster-profiles/cluster-profiles-config.yaml
@@ -397,6 +397,57 @@
     - org: openshift
       repos:
         - openshift-tests-private
+- profile: azure-qe
+  owners:
+    - org: openshift
+      repos:
+        - openshift-tests-private
+        - verification-tests
+        - csi-operator
+        - azure-file-csi-driver
+        - azure-file-csi-driver-operator
+        - azure-disk-csi-driver
+        - azure-disk-csi-driver-operator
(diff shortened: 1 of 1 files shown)


jianlinliu:
cc @jinyunma @gpei to review

gpei:
@jianlinliu the list is generated by filtering the `profile: azure-qe/azuremag-qe/azurestack-qe/azure-autorelease-qe` from the whole "release" repo?

jinyunma:
After this change, does it mean that only jobs defined under release/ci-operator/config/openshift/<restricted_repo_name>/ are allowed to be scheduled by using corresponding cluster profile?

jianlinliu:
> After this change, does it mean that only jobs defined under release/ci-operator/config/openshift/<restricted_repo_name>/ are allowed to be scheduled by using corresponding cluster profile?

Should only `release/ci-operator/config/<restricted_org_name>/<restricted_repo_name>/` allow to use the corresponding cluster profiles, or else, some similar error like https://redhat-internal.slack.com/archives/CH76YSYSC/p1749653036145579?thread_ts=1747761002.988469&cid=CH76YSYSC

jianlinliu:
cc @liangxia to review

jianlinliu:
> @jianlinliu the list is generated by filtering the `profile: azure-qe/azuremag-qe/azurestack-qe/azure-autorelease-qe` from the whole "release" repo?

Right, that is boring ~.

PR #66016, opened 2025-06-12
Stop syncing ironic-lib repo to openshift

elfosardo:
ironic-lib is no more upstream, we can stop and remove the automated repo sync job


elfosardo:
@damdo @JoelSpeed hi! can you please have a look at this when you get the chance? thanks!

PR #66025, opened 2025-06-12
OCPBUGS-57330: external-dns-operator: Switch cluster profile to aws-qe for shared VPC test

alebedev87:
`aws-2` cluster profile is missing the secret for the shared vpc (`awscred_shared_account`). This PR switches the cluster profile for the shared VPC job to `aws-qe`.


alebedev87:
The rehearsal failed, an instance of ExternalDNS was created, AWS client was initialized with the private zone IAM role but the expected DNS record didn't appear in the hosted zone. We need logs from the operand to be able to understand what happened.

PR #66061, opened 2025-06-13
feat: add optional presubmit checks for two-node-fencing

eggfoobar:


eggfoobar:
/pj-rehearse ack

No need to run the whole tests, as some have known failures, we just needed to validate installs

PR #66082, opened 2025-06-13
Increase time for lint and gencheck jobs ossm 4.18

fjglira:


fjglira:
Added `/pj-rehearse ack` there is no need to rerun the jobs here, we are only increasing the timeout

PR #66100, opened 2025-06-16
Keep consistent for node skew function name

xiuwang:
1. Keep consistent for node skew function name
2. Use separate env var for Capacity Reservation zone so we can point to different zone with HC


xiuwang:
@juzhao Could you approve this change? Thanks

PR #66123, opened 2025-06-16
Add openstack 17 rhel9 repo for 4.15

gryf:


elfosardo:
/hold this may break stuff as all openshift images will use packages from that repo
what's needed from that repo?

gryf:
ok, I found needed packages in rhel-8 base image instead. Closing.

PR #66124, opened 2025-06-16
OCM-13977 | Replace/Disable the 4.18/4.19 old Openshift E2E testing jobs

yingzhanredhat:
-Replace  the old rosa preparation workflow with new rosa day1

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.17__amd64-nightly.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.17__amd64-nightly.yaml
@@ -1149,14 +1149,14 @@ tests:
     cluster_profile: aws-sd-qe
     env:
       CHANNEL_GROUP: nightly
-      E2E_RUN_TAGS: '@rosa'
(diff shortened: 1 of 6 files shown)


ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.17__amd64-nightly.yaml
```diff
@@ -1149,14 +1149,14 @@ tests:
     cluster_profile: aws-sd-qe
     env:
       CHANNEL_GROUP: nightly
-      E2E_RUN_TAGS: '@rosa'
```
xueli181114, line 1152: Is it expected to remove this tag?

yingzhanredhat: Yes.It is hcp external auth profile.It is not allowed to create multi users for e2e test. It is same with OCP job too.We can only use `openshift-extended-test` test chain

yingzhanredhat:
@liangxia Would you help review it?Thanks

liangxia (approved):

PR #66148, opened 2025-06-17
rebalance gcp-qe usage

jianlinliu:

PR #66151, opened 2025-06-17
phase-2:  opt-out openshift-power-monitoring org from gcp-qe cluster profile

jianlinliu:
after https://github.com/openshift/release/pull/66148, opt-out openshift-power-monitoring org from gcp-qe cluster profile


jianlinliu:
cc @liangxia to review

PR #66162, opened 2025-06-17
STOR-2478: Enable ironic repository in efs-utils-base

jsafrane:
ironic yum repo contains python packages needed by the efs-utils: python3-coverage python3-funcsigs python3-pytest-cov python3-pytest-mock

Copied from ironic images: https://github.com/openshift/release/blob/c6736dd43d3ffa65cf8765801ad75c21705fe54a/ci-operator/config/openshift/ironic-image/openshift-ironic-image-main.yaml#L24-L30


jsafrane:
This should not be necessary, I'm trying to cut the dependencies in https://github.com/openshift/aws-efs-utils/pull/32/files.

PR #66220, opened 2025-06-18
USHIFT-5639: Adjust 4.19 rebase frequency

pacevedom:


ci-operator/config/openshift/microshift/openshift-microshift-release-4.19.yaml
```diff
@@ -85,31 +85,6 @@ tests:
...
-  capabilities:
-  - intranet
-  cron: 30 0 * * *
-  restrict_network_access: false
-  steps:
-    cluster_profile: aws-2
-    env:
-      EC2_INSTANCE_TYPE: c7g.2xlarge
-      MICROSHIFT_OS: rhel-9.6
-    workflow: openshift-microshift-e2e-metal-cache
```
ggiguash, line 112: I'm not sure we should cancel cache builds. We still do this for 4.18 and previous releases.
The problem is that bootc builds need ostree images for some tests, and those can only come from cache.

pacevedom: ah, they are not cancelled, they were duplicated. Now they are just [here](https://github.com/openshift/release/pull/66220/files#diff-d5f4d7a283ef9e9e59acfa92e15e2b8a728d25a978e21cf285b0d62d07aeb00cR44-R64)

PR #66226, opened 2025-06-18
Enable usernamespace support on the kubelet

deads2k:
Trying to resolve the error message:
  Failed to create pod sandbox: the feature gate "UserNamespacesSupport" is disabled: can't set spec.HostUsers

When creating a usernamespaced pod.

`oc --as=system:admin create --dry-run=server  -f clusters/build-clusters/build02/enable-user-namespaces.yaml` reports ok

/assign @deepsm007

### DIFF
--- a/clusters/build-clusters/build02/enable-user-namespaces.yaml
+++ b/clusters/build-clusters/build02/enable-user-namespaces.yaml
@@ -1,0 +1,14 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: KubeletConfig
+metadata:
+  annotations:
+    machineconfiguration.openshift.io/mc-name-suffix: ""
+  name: set-user-namespaces-support
+spec:
+  kubeletConfig:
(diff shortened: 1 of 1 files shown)


haircommander:
wouldn't you also need to set the feature on apiserver so it doesn't scrub the feature from client requests? or is the featuregate controller not writing the featuregate to the kubelet?

deads2k:
> or is the featuregate controller not writing the featuregate to the kubelet?

Near as I can tell, this.  Which suggests to me aren't running an e2e test on it.

deads2k:
heh.  It creates, but it doesn't work:

```
status:
  conditions:
  - lastTransitionTime: "2025-06-18T14:06:49Z"
    message: 'Error: KubeletConfiguration: featureGates is not allowed to be set,
      but contains: map[UserNamespacesSupport:true]'
    status: "False"
    type: Failure
```

That would be a great use for VAP.

PR #66235, opened 2025-06-18
Bump devfile registry operator testing to use OpenShift v4.18

michael-valdron:
## Use OpenShift v4.18 for testing

Use OpenShift v4.18 for Devfile Registry Operator integration testing.

fixes devfile/api#1669


thepetk (approved):
lgtm

PR #66243, opened 2025-06-18
CNTRLPLANE-777: Add ARO HCP AKS e2e tests as informing on CI release controller for 4.19

bryan-cox:
This PR adds the AKS e2e as an informing test on the CI payload release controllers for 4.19


sjenning:
/lgtm

TRT note: might check if renaming a test is going to mess anything up. I imagine it disconnects the history for the job.

PR #66258, opened 2025-06-19
Add DISABLE_BOOT_IMAGE_UPDATE to disable boot image update in 4.18 to 4.19 upgrade jobs

gpei:
As a follow up of https://issues.redhat.com/browse/OCPBUGS-57348, in AWS/GCP 4.18 to 4.19 upgrade jobs that's configured with custom boot image(using Marketplace Image in this case), we need to either patching the "ack-4.18-boot-image-opt-out-in-4.19" admin-ack or disabling the boot image update feature[1].
In CI, we decided to adopt the latter solution, which is to disable boot image update from MCO side, because it seems to be a more reasonable solution that is closer to user scenarios.

[1] https://docs.redhat.com/en/documentation/openshift_container_platform/4.18/html/machine_configuration/mco-update-boot-images#mco-update-boot-images-disable_machine-configs-configure

### DIFF
--- a/ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
+++ b/ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
@@ -247,6 +247,7 @@ function disable_boot_image_update() {
 
     if [ $get_status -ne 0 ]; then
         echo "Error: Failed to get current MachineConfiguration. Check cluster access."
+	export UPGRADE_FAILURE_TYPE="machine-config"
         return 1
     fi
 
@@ -261,6 +262,7 @@ function disable_boot_image_update() {
     # Edit the MachineConfiguration to disable boot image updates
     if ! oc patch MachineConfiguration cluster --type=merge --patch '{"spec":{"managedBootImages":{"machineManagers":[]}}}' -n openshift-machine-config-operator; then
         echo "Error: Failed to patch MachineConfiguration."
+	export UPGRADE_FAILURE_TYPE="machine-config"
         return 1
     fi
 
@@ -274,6 +276,7 @@ function disable_boot_image_update() {
         return 0
     else
         echo "Error: Failed to update machineManagers. Current value: $new_value"
+	export UPGRADE_FAILURE_TYPE="machine-config"
         return 1
     fi
 }
--- a/ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
+++ b/ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
@@ -246,6 +246,7 @@ function disable_boot_image_update() {
 
     if [ $get_status -ne 0 ]; then
         echo "Error: Failed to get current MachineConfiguration. Check cluster access."
+	export UPGRADE_FAILURE_TYPE="machine-config"
         return 1
     fi
 
@@ -261,6 +262,7 @@ function disable_boot_image_update() {
     # Edit the MachineConfiguration to disable boot image updates
     if ! oc patch MachineConfiguration cluster --type=merge --patch '{"spec":{"managedBootImages":{"machineManagers":[]}}}' -n openshift-machine-config-operator; then
         echo "Error: Failed to patch MachineConfiguration."
+	export UPGRADE_FAILURE_TYPE="machine-config"
         return 1
     fi
 
@@ -274,6 +276,7 @@ function disable_boot_image_update() {
         return 0
     else
         echo "Error: Failed to update machineManagers. Current value: $new_value"
+	export UPGRADE_FAILURE_TYPE="machine-config"
         return 1
     fi
 }
@@ -634,8 +637,19 @@ if [[ "${UPGRADE_CCO_MANUAL_MODE}" == "oidc" ]]; then
 fi
 if [[ "${DISABLE_BOOT_IMAGE_UPDATE}" == "true" ]]; then
     #Disable updated boot images feature for jobs with custom boot image specified in certain upgrade paths
-    disable_boot_image_update
+    echo "Checking conditions for disabling boot image updates..."
+
+    # Get platform
+    PLATFORM=$(oc get infrastructure cluster -o jsonpath='{.status.platformStatus.type}')
+
+    # Check all conditions
+    if [[ "${TARGET_MINOR_VERSION}" == "19" ]] && [[ "${PLATFORM}" =~ ^(AWS|GCP)$ ]]; then
+        disable_boot_image_update
+    else
+        echo "Skipping boot image update disablement."
+    fi
 fi
+
 upgrade
 check_upgrade_status
 


gpei:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.19-amd64-nightly-4.19-upgrade-from-stable-4.18-aws-ipi-workers-marketplace-mini-perm-f28

gpei:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.19-amd64-nightly-4.19-upgrade-from-stable-4.18-aws-ipi-workers-marketplace-mini-perm-f28

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -592,6 +600,10 @@ fi
 if [[ "${UPGRADE_CCO_MANUAL_MODE}" == "oidc" ]]; then
     update_cloud_credentials_oidc
 fi
+if [[ "${DISABLE_BOOT_IMAGE_UPDATE}" == "true" ]]; then
```
jianlinliu, line 638: Maybe we also need the change in `cucushift-chainupgrade-toimage`, but in the chain upgrade case, we also need to version check.

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -238,6 +238,14 @@ function cco_annotation(){
     fi
 }
 
+function disable_boot_image_update() {
+    echo "Disabling updated boot images by editing MachineConfiguration..."
```
jianlinliu, line 260: Suggest to add a check if MachineConfiguration/cluster is already patched or not, because cucushift-upgrade-toimage will be run as the last hop upgrade step.

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
gpei: as we don't have `marketplace-image` related upgrade testing in long chain upgrade jobs, do we still need to update the `cucushift-chainupgrade-toimage`? the DISABLE_BOOT_IMAGE_UPDATE step will only be required for jobs like `marketplace-image` where custom boot image was set.

jianlinliu: Actually I am not sure if there is future change to cover marketplace ci jobs picked by OTA team, but to be robust, maybe we can have it in `cucushift-chainupgrade-toimage`

gpei: ok, updated

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
gpei: updated

gpei:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.19-amd64-nightly-4.19-upgrade-from-stable-4.18-aws-ipi-workers-marketplace-mini-perm-f28

gpei:
From the [rehearsal job](https://qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/qe-private-deck/pr-logs/pull/openshift_release/66258/rehearse-66258-periodic-ci-openshift-openshift-tests-private-release-4.19-amd64-nightly-4.19-upgrade-from-stable-4.18-aws-ipi-workers-marketplace-mini-perm-f28/1935667535328317440) 
```
Current machineManagers value: [{"apiGroup":"machine.openshift.io","resource":"machinesets","selection":{"mode":"Partial","partial":{"machineResourceSelector":{"matchLabels":{"mcotest":"update"}}}}}]
Disabling updated boot images by editing MachineConfiguration...
machineconfiguration.operator.openshift.io/cluster patched
Verifying the change...
Successfully disabled boot image update.
```
The AMI of worker remains to be the same after [cluster fresh installation](https://gcsweb-qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/qe-private-deck/pr-logs/pull/openshift_release/66258/rehearse-66258-periodic-ci-openshift-openshift-tests-private-release-4.19-amd64-nightly-4.19-upgrade-from-stable-4.18-aws-ipi-workers-marketplace-mini-perm-f28/1935667535328317440/artifacts/aws-ipi-workers-marketplace-mini-perm-f28/cucushift-installer-check-aws-custom-ami/build-log.txt) and [post upgrade](https://gcsweb-qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/qe-private-deck/pr-logs/pull/openshift_release/66258/rehearse-66258-periodic-ci-openshift-openshift-tests-private-release-4.19-amd64-nightly-4.19-upgrade-from-stable-4.18-aws-ipi-workers-marketplace-mini-perm-f28/1935667535328317440/artifacts/aws-ipi-workers-marketplace-mini-perm-f28/gather-extra/artifacts/machinesets.json).

And the final MachineConfiguration status after upgrade
```
apiVersion: operator.openshift.io/v1
kind: MachineConfiguration
...
spec:
  logLevel: Normal
  managedBootImages:
    machineManagers: []
  managementState: Managed
  operatorLogLevel: Normal
status:
  conditions:
  - lastTransitionTime: "2025-06-19T19:26:45Z"
    message: Reconciled 0 of 0 MAPI MachineSets | Reconciled 0 of 0 CAPI MachineSets
      | Reconciled 0 of 0 CAPI MachineDeployments
    reason: MAPIMachinesetDeleted
    status: "False"
    type: BootImageUpdateProgressing
  - lastTransitionTime: "2025-06-19T19:26:45Z"
    message: 0 Degraded MAPI MachineSets | 0 Degraded CAPI MachineSets | 0 CAPI MachineDeployments
    reason: MAPIMachinesetDeleted
    status: "False"
    type: BootImageUpdateDegraded
  managedBootImagesStatus:
    machineManagers: []
...
```

gpei:
@jianlinliu @jiajliu please help to take a look, thanks.

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
```diff
@@ -1169,7 +1208,20 @@ for target in "${TARGET_RELEASES[@]}"; do
...
     fi
+    if [[ "${DISABLE_BOOT_IMAGE_UPDATE}" == "true" ]]; then
+	#Disable updated boot images feature for jobs with custom boot image specified in certain upgrade paths
+        echo "Checking conditions for disabling boot image updates..."
+
+        # Get platform
+        PLATFORM=$(oc get infrastructure cluster -o jsonpath='{.status.platformStatus.type}')
 
+        # Check all conditions
+        if [[ "${TARGET_MINOR_VERSION}" == "19" ]] && [[ "${PLATFORM}" =~ ^(AWS|GCP)$ ]]; then
```
jiajliu, line 1222: Shall we also check if it's configured with custom boot image?

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
```diff
@@ -592,6 +632,10 @@ fi
 if [[ "${UPGRADE_CCO_MANUAL_MODE}" == "oidc" ]]; then
     update_cloud_credentials_oidc
 fi
+if [[ "${DISABLE_BOOT_IMAGE_UPDATE}" == "true" ]]; then
```
jiajliu, line 638: should it be consistent with chainupgrade step?

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
jianlinliu: Generally only marketplace ci jobs need to set `DISABLE_BOOT_IMAGE_UPDATE`, of course, users can set `DISABLE_BOOT_IMAGE_UPDATE` on any user wanted custom boot image ci jobs.

jianlinliu: @gpei If we already introduced `DISABLE_BOOT_IMAGE_UPDATE` env to control the behavior, shall we still need `PLATFORM` judgement ?

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
```diff
@@ -239,6 +239,45 @@ function cco_annotation(){
...
 
+function disable_boot_image_update() {
+    # Get current machineManagers value
+    local current_value
+    current_value=$(oc get MachineConfiguration cluster -n openshift-machine-config-operator -o jsonpath='{.spec.managedBootImages.machineManagers}' 2>/dev/null)
+    local get_status=$?
+
+    if [ $get_status -ne 0 ]; then
+        echo "Error: Failed to get current MachineConfiguration. Check cluster access."
+        return 1
```
jiajliu, line 251: It's better to set UPGRADE_FAILURE_TYPE before each error exit to help identifying upgrade failure.

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
jianlinliu: boot image management is introduced in 4.19, that is why chainupgrade step has an extra condition to check OCP version.

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
gpei: @jiajliu generally we'll only set env `DISABLE_BOOT_IMAGE_UPDATE` for custom boot image upgrade jobs explicitly, so I guess we can skip the custom boot image checking here. And the logic is also a bit complicated to implement a judgment for all platforms to check whether custom boot image is used.

gpei: @jianlinliu 

> shall we still need PLATFORM judgement

this is a good question, here's my thought, obviously the boot image update support will be landed into other cloud platform like Azure in the future, and the step to disable boot image update will be required in different version upgrades than AWS/GCP, so I still kept a `platform` parameter here.

jiajliu: ah, got it. then it sounds like `platform` is also not necessary.

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
jiajliu: my concern is 4.18-4.19 upgrade jobs will be blindly copied to 4.19-4.20 jobs with DISABLE_BOOT_IMAGE_UPDATE var, so if there is not version check, does it mean disabling boot image update is also expected in 4.19-4.20 upgrade?

gpei: For chainupgrade job, if we don't add the `TARGET_MINOR_VERSION` check, the disabling bot image step won't be executed explicitly during 4.18 to 4.19, but we do want to only execute once. 
For the common 4.y to 4.y+1 upgrade jobs, we can use env `DISABLE_BOOT_IMAGE_UPDATE` to control the step conveniently.

gpei: @jiajliu that's a good point, I remember we had encountered this problem before, so that we had to remove the environment variables we had added in subsequent versions.

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
jianlinliu: > the step to disable boot image update will be required in different version upgrades than AWS/GCP, so I still kept a platform parameter here.

Good point, and agree.

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
gpei: so the most challenging thing is which team owns this failure :)

gpei: anyway I guess it should be MCO, fair enough.

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
jianlinliu: If we have the explicit `DISABLE_BOOT_IMAGE_UPDATE` setting env, I prefer to remove it when we blindly copied them on an unwanted upgrade path.

jiajliu: yeah, the ideal situation is that job maintainer is just the feature owner, then it should not be a problem. But afaik it's generally a copy-past when enabling new jobs, but not including adjusting each vars due to lack of content of each var's expected behavior. If you are both had not concern, it's fine to not add version check.

gpei: I think we can do both. To make the script stronger and the job definitions cleaner.

ci-operator/step-registry/cucushift/chainupgrade/toimage/cucushift-chainupgrade-toimage-commands.sh
jiajliu: > the step to disable boot image update will be required in different version upgrades than AWS/GCP

make sense, ack

ci-operator/step-registry/cucushift/upgrade/toimage/cucushift-upgrade-toimage-commands.sh
gpei: Created https://issues.redhat.com/browse/OCPQE-29995 to track the issue, I'll regularly review all open issues assigned to me, so it shouldn't be forgotten, hopefully.

gpei: Adding a version check here makes more sense if we treat it as a one-shot action, it won't be required in any other place or other upgrade path.

gpei:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.19-amd64-nightly-4.19-upgrade-from-stable-4.18-aws-ipi-workers-marketplace-mini-perm-f28

gpei:
@liangxia for the approval, thanks.

liangxia (approved):

PR #66261, opened 2025-06-19
[PERFSCALE-3976] Secrets for bare metal cluster profiles

rsevilla87:
We miss pull-secret  in these cluster profiles

/cc @josecastillolema


rsevilla87:
//cc @liangxia

liangxia (approved):

PR #66263, opened 2025-06-19
ossm: ztunnel: automatic upstream->downstream sync for 1.26 branch

zmiklank:


zmiklank:
That is a good point. I will rework it, also for the 1.24 branch.

PR #66293, opened 2025-06-20
Use  ubi9-gotoolset:1.23.9 Image in ODHMC CI

brettmthompson:
After golang version bump in odh-model-controller a golang image of version 1.23.8 or greater is required for CI.


brettmthompson:
Golang version bump no longer required in odh-model-controller. CVEs were resolved in https://github.com/opendatahub-io/odh-model-controller/pull/464

PR #66296, opened 2025-06-20
Use golang-1.24 Image in ODHMC

brettmthompson:
A later version of golang is required in the CI image used by odh-model-controller after the golang version bump in the repo


brettmthompson:
Closing as `golang-1.24` image is not supported

PR #66316, opened 2025-06-23
Add datarouter and reportportal

liangxia:
To support https://github.com/openshift/release/pull/66323

### DIFF
--- a/clusters/build-clusters/common_managed/dns.yaml
+++ b/clusters/build-clusters/common_managed/dns.yaml
@@ -15,11 +15,20 @@ spec:
   servers:
     - name: rh-intranet
       zones:
+        - app.intlab.redhat.com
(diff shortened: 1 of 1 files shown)


clusters/build-clusters/common_managed/dns.yaml
```diff
@@ -15,27 +15,29 @@ spec:
   servers:
     - name: rh-intranet
       zones:
+        - app.intlab.redhat.com
         - arm.eng.rdu2.redhat.com
+        - brewhub.engineering.redhat.com
         - comet.engineering.redhat.com
+        - datarouter.dno.corp.redhat.com/
```
jianlinliu: extra `/` ?

liangxia: Nice catch.

PR #66336, opened 2025-06-23
feat: add presubmits to mco for tna/tnf

eggfoobar:


eggfoobar:
/pj-rehearse ack

The rehearsals for arbiter are failing with only the featureset flag test, this is expected since TNA has not graduated in the API yet to GA (this is due to happen soon), fencing is failing with known failures that are being tracked.

PR #66342, opened 2025-06-23
OCPNODE-3385: add 4.19 testing for kueue

kannon92:
Add 4.19 testing for kueue operator and operand.

I may need to work with test platform to make sure there is a cluster bool for 4.19.


kannon92:
/hold

We need a 4.19 cluster pool.

kannon92:
/pj-rehease ack

We know why these tests are failing. Our script doesn't work on picking the right image on openshift/release.

We are okay with merging this with these failures. They will work on our repo.

PR #66378, opened 2025-06-24
Align MCO e2e-gcp-op timeout with the internal used one

pablintino:
e2e-gcp-op has started to be killed by timeout and it looks like the `go test` command used inside the Makefile has a bit more of time than the one configured here.


pablintino:
pj-rehearse

PR #66398, opened 2025-06-24
make basic job to install ARO HCP in integration

deads2k:
TODO

1. set env vars
2. ensure the chain can be called
3. make a workflow that does cleanup?
4. include secret in vault


deads2k:
replaced in https://github.com/openshift/release/pull/66398

PR #66407, opened 2025-06-24
Automation for RHDH Certification Pipeline

dcurran90:
Deploy and test RHDH plugins for RHDH Certification process.


ci-operator/step-registry/redhat-developer/rhdh/certification/redhat-developer-rhdh-certification-ref.yaml
```diff
@@ -0,0 +1,16 @@
+ref:
+  as: redhat-developer-rhdh-certification
+  cli: latest
+  commands: "redhat-developer-rhdh-certification-commands.sh"
+  credentials:
+  - mount_path: /tmp/secrets
+    name: rhdh
```
subhashkhileri: @dcurran90  Is this the same RHDH secrets collection you're using for certification? If so, it needs to be a new, separate one.

PR #66521, opened 2025-06-27
prevent the Azure/ARO-HCP presubmit from running automatically for now

deads2k:
/assign @geoberle


deads2k:
just an ordering change, reapplying lgtm

PR #66527, opened 2025-06-27
WIP: Add steps to run e2e tests for cso operator

gnufied:
xref

### DIFF
--- a/ci-operator/config/openshift/cluster-storage-operator/openshift-cluster-storage-operator-release-4.20.yaml
+++ b/ci-operator/config/openshift/cluster-storage-operator/openshift-cluster-storage-operator-release-4.20.yaml
@@ -201,6 +201,19 @@ tests:
   steps:
     cluster_profile: hypershift
     workflow: hypershift-aws-e2e-external
+- as: e2e-cluster-storage-operator-test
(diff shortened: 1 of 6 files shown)


ci-operator/config/openshift/cluster-storage-operator/openshift-cluster-storage-operator-release-4.20.yaml
```diff
@@ -201,6 +201,19 @@ tests:
   steps:
     cluster_profile: hypershift
     workflow: hypershift-aws-e2e-external
+- as: e2e-cluster-storage-operator-test
+  steps:
+    cluster_profile: aws-2
```
jsafrane, line 206: We will need to run the e2e tests on all clouds and form factors. IMHO we don't want to add a specific e2e job for each cloud. I'd prefer adding the test to openshift/origin with `[OCPFeatureGate:StoragePerformantSecurityPolicy]`. We already have e2e jobs that run origin tests + they can filter by feature gate.

jsafrane:
I think this is not needed, we have tests for `StoragePerformantSecurityPolicy` in regular jobs.
/close

PR #66632, opened 2025-07-02
switch to using kueue ga repos

kannon92:
Since we switched to using GA repos in our bundle, we should change our mirror sets.


kannon92:
/pj-rehease pull-ci-openshift-kueue-operator-release-1.0-test-e2e-upstream-4-18

PR #66700, opened 2025-07-03
ROX-29505: run sensor integration tests only on changes

janisz:
We migrated this tests to GHA on kind cluster. They are working much faster as we do not need to wait for tests. This means we can skip running this job on all PRs and run only if sensor code is changed. 

- https://github.com/stackrox/stackrox/pull/15407


janisz:
cc: @lvalerom @vikin91

PR #66704, opened 2025-07-03
Add stephenfin to approver for openshift on openstack

mandre:
Stephen is going to gradually assume team lead responsibilities for OpenShift on OpenStack. Leaving myself in these files as well for at least the transition period.


mandre:
cc @stephenfin

PR #66720, opened 2025-07-04
Add new release aws-observability-quota-slice

anpingli:
Remove https://github.com/openshift/release/pull/66669/files as  observability-aws-quota-slice is defined for dynamic resources.
Add a new lease one  aws-observability-quota-slice for cluster-profile.


core-services/prow/02_config/generate-boskos.py
```diff
@@ -523,6 +523,8 @@
     },
     'observability-aws-quota-slice': {
         'default': 50,
```
psalajova: If this is no longer needed, please remove it

anpingli: I think that  is used by  cluster pools. That is why I keep it asis and create the new one aws-observability-quota-slice'

anpingli: Is there any way to confirm if observability-aws-quota-slice' is used?

psalajova: `observability-aws-quota-slice` was used for the cluster profile ClusterProfileAWSObservabiltity. I don't think these leases are used for cluster pools

anpingli: Got it. Drop it.

anpingli: @psalajova  PTAL, thanks

PR #66790, opened 2025-07-07
add older oadp must-gather versions to image mirror

weshayutin:
OADP should have the quay image available as needed.


weshayutin:
@kaovilai fyi

PR #66409, opened 2025-06-24
[debug] [don't merge]: testing karpenter kwok e2e

maxcao13:
/hold

This PR is a testing grounds for figuring out how to configure our [kubernetes-sigs/karpenter](https://github.com/openshift/kubernetes-sigs-karpenter) fork to be able to run the kwok e2e tests.

do not merge

PR #66797, opened 2025-07-07
AUTOSCALE-256: add e2e-kwok karpenter core perioidic/presubmit test

maxcao13:
Fixes: https://issues.redhat.com/browse/AUTOSCALE-256

Depends on: https://github.com/openshift/kubernetes-sigs-karpenter/pull/8

Based on: https://github.com/openshift/release/pull/66409

~The setup registry test step looks a little grody as John may call it, but what's happening is we are getting the raw cert data attached to the internal registry's route, and then allowing the cluster's pod image puller to trust that certificate by patching the `image.config.openshift.io/cluster` object, because for whatever reason, it isn't already trusted. That process is somehwat documented here: https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/images/image-configuration#images-configuration-cas_image-configuration. Then we just wait for some components in the cluster to be ready before moving on.~

Above is obsolete. We don't need to trust the cert anymore, if we just use the correct internal registry pull spec for the pods to pull from.

This will allow the cluster that is provisioned using the `ipi-aws` workflow to pull the ko built images from it's own internal registry. And then we run the actual e2e test script that we created in https://github.com/openshift/kubernetes-sigs-karpenter/pull/8.

The test runs as a presubmit and as a periodic that runs every monday at 5am.

### DIFF
--- a/ci-operator/config/openshift/kubernetes-sigs-karpenter/openshift-kubernetes-sigs-karpenter-main.yaml
+++ b/ci-operator/config/openshift/kubernetes-sigs-karpenter/openshift-kubernetes-sigs-karpenter-main.yaml
@@ -1,4 +1,8 @@
 base_images:
+  cli:
+    name: "4.20"
+    namespace: ocp
+    tag: cli
   hypershift-operator:
     name: hypershift-operator
     namespace: hypershift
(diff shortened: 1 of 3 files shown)


maxcao13:
/hold

hold on, I think there might be a way so that we don't have to do the setup-registry step as part of the CI job steps?

elmiko:
/lgtm

i talked with @maxcao13 in slack, no rehearse here due to the pr needing to merge. but he did run it from the pr, see https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/66409/rehearse-66409-pull-ci-maxcao13-karpenter-openshift-main-e2e-kwok/1943144356294692864

maxcao13:
/pj-rehearse pull-ci-openshift-kubernetes-sigs-karpenter-main-e2e-kwok

maxcao13:
Cool!

/pj-rehearse ack

PR #66800, opened 2025-07-07
LOG-7330: Add logging cluster pool for ocp 4.19

jcantrill:
https://issues.redhat.com/browse/LOG-7330


cahartma (approved):
fyi, the status elements can be excluded as that part is overwritten

PR #66836, opened 2025-07-08
Get pull number from JOB_SPEC

liangxia:
Instead of use `PULL_NUMBER` which does not exist in periodic jobs, get pull number from `JOB_SPEC` which is more robust as it always exist.


liangxia:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.20-amd64-nightly-aws-ipi-f1

droslean (approved):

PR #38362, opened 2023-04-14
Add build job for openshift/windows_exporter

mansikulkarni96:
Add a build job for openshift/windows_exporter 

Ran commands:
```
make update
```

PR #66850, opened 2025-07-08
WINC-1367: add windows_exporter test jobs for release 4.14 to 4.19 branches

jrvaldes:


jrvaldes:
@openshift/openshift-team-windows-containers  PTAL, test passed

ci-operator/config/openshift/prometheus-community-windows_exporter/openshift-prometheus-community-windows_exporter-release-4.14.yaml
```diff
@@ -0,0 +1,42 @@
...
+    test:
+    - as: test
+      commands: |
+        git clone --branch release-4.14 --single-branch https://github.com/openshift/windows-machine-config-operator.git
+        cd windows-machine-config-operator/
+        git submodule update --init
+        cd promu/
+        GOBIN=/go/src/github.com/openshift/prometheus-community-windows_exporter go install .
+        cd /go/src/github.com/openshift/prometheus-community-windows_exporter
+        GOOS=windows ./promu build -v
```
sebsoto: I don't understand this, why is the WMCO repo being cloned into here? For the promu submodule?
If promu is needed, only promu should be gotten.

What if an update to promu is needed to build windows-exporter? You are creating a circular dependency here.

sebsoto (changes requested):

ci-operator/config/openshift/prometheus-community-windows_exporter/openshift-prometheus-community-windows_exporter-release-4.14.yaml
jrvaldes: >If promu is needed, only promu should be gotten.

True, however there is another element in to consideration; how to sync the `promu` and `prometheus-community-windows_exporter` commit hash WMCO is using at a given time. The later, may explain the need to just pull WMCO code for the given branch

jrvaldes: maybe @mansikulkarni96  can add more color, as it was introduced in 
- https://github.com/openshift/release/pull/38362/commits/2804310f731d797c3841ba11f9cb2dccfb7deaac

jrvaldes: >What if an update to promu is needed to build windows-exporter?

That will be resolved within the WMCO repository itself through submodule updates. I don't see any risk for circular dependency

mansikulkarni96: The idea was to use the same version and build it the same way we do in the wmco builds.

jrvaldes: > The idea was to use the same version and build it the same way we do in the wmco builds.

+1

PR #66855, opened 2025-07-08
skip validation on kueue repos for jira

kannon92:
I want to ignore jira lifecycle checks on kueue operators.

We don't follow OCP release cycle and we don't really have the same process for cherry picks.

### DIFF
--- a/core-services/jira-lifecycle-plugin/config.yaml
+++ b/core-services/jira-lifecycle-plugin/config.yaml
@@ -524,18 +524,6 @@ orgs:
         branches:
           '*':
             skip_target_version_check: true
-      lightspeed-console:
-        branches:
-          '*':
-            skip_target_version_check: true
-      lightspeed-service:
-        branches:
-          '*':
-            skip_target_version_check: true
-      lightspeed-operator:
-        branches:
-          '*':
-            skip_target_version_check: true
       kubernetes-sigs-kueue:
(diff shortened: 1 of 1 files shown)


core-services/jira-lifecycle-plugin/config.yaml
```diff
@@ -536,6 +536,18 @@ orgs:
...
+          '*':
+            validate_by_default: false
+      kueue-operator:
+        branches:
+          '*':
+            validate_by_default: false
+      kueue-fbc:
+        branches:
+          '*':
+            validate_by_default: false
```
bradmwilliams, line 538: I would recommend that these get bumped upto line 527 (to fall in line alphabetically).
You'll also need to copy the same block down under the `openshift-priv` organisation as well.

bradmwilliams (changes requested):

core-services/jira-lifecycle-plugin/config.yaml
kannon92: > I would recommend that these get bumped upto line 527 (to fall in line alphabetically).


done.

> ou'll also need to copy the same block down under the openshift-priv organisation as well.

I don't think I need to do that. I don't have these repos in openshift-priv. They are public repo only.

bradmwilliams (approved):

PR #66863, opened 2025-07-09
Run more tests in job which sends test results to reportportal

liangxia:
Only run a smaller test sets ( critical tests cases ) when doing the integration with ReportPortal for faster iteration.
We are able to send the test results into ReportPortal now, we'd like to run more tests to cover wider situations.


liangxia:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.20-amd64-nightly-aws-ipi-f1

PR #66899, opened 2025-07-09
telcov10n: orion tests for cpu utilization

abraham2512:
This PR adds a CI test for orion anomaly detection for cpu utilization test results

### DIFF
--- a/ci-operator/step-registry/telcov10n/orion/tests/telcov10n-orion-tests-commands.sh
+++ b/ci-operator/step-registry/telcov10n/orion/tests/telcov10n-orion-tests-commands.sh
@@ -25,8 +25,6 @@ OPENSEARCH_USERNAME=$(<"/secret/telco5g/username")
 OPENSEARCH_HOST=$(<"/secret/telco5g/hostname")
 ES_SERVER="https://${OPENSEARCH_USERNAME}:${OPENSEARCH_PASSWORD}@${OPENSEARCH_HOST}"
 
-export ES_SERVER
-
 pip install .
 export EXTRA_FLAGS=" --anomaly-detection"
 


ci-operator/step-registry/telcov10n/orion/tests/OWNERS
```diff
@@ -0,0 +1,7 @@
+approvers:
```
natifridman, line 1: Maybe we should add you as an owner here?

abraham2512: Okay I can do that

abraham2512: Done

ci-operator/step-registry/telcov10n/orion/tests/telcov10n-orion-tests-commands.sh
```diff
@@ -0,0 +1,90 @@
...
+pushd orion
+
+pip install -r requirements.txt
+
+OPENSEARCH_PASSWORD=$(<"/secret/telco5g/password")
+OPENSEARCH_USERNAME=$(<"/secret/telco5g/username")
+OPENSEARCH_HOST=$(<"/secret/telco5g/hostname")
+ES_SERVER="https://${OPENSEARCH_USERNAME}:${OPENSEARCH_PASSWORD}@${OPENSEARCH_HOST}"
+
+export ES_SERVER
```
shaior: I'm not sure if this is secure, or follow best practice, as you export sensitive information here

abraham2512: I think I can remove this export. The es_server url is also passed directly to orion in line 82, so this is redundant

abraham2512: Done

PR #66927, opened 2025-07-10
fix sandbox ci daily

rsoaresd:
# Description
 I forgot to disable the ci pairing, sorry :disappointed: For periodic job, we need to have the ci pairing disabled

Context: To achieve continuous monitoring, low-maintenance days, early detection, and more data for future metrics, we should add a Periodic CI job. This type of job will help us catch flaky tests earlier, since we expect that the pass rate is 100%, since we are testing against the code that is already in production. Also, on the Logilica side, we will be able to have metrics like success rate, failure rate, ...

## Issue ticket number and link
[SANDBOX-801](https://issues.redhat.com/browse/SANDBOX-801)


mfrancisc (approved):
/lgtm

PR #66945, opened 2025-07-10
start mechanism to launch ARO-HCP testing

deads2k:
requires https://github.com/Azure/ARO-HCP/pull/2125


deads2k:
To avoid waiting, we can merge in the reverse order: best guess at a CI job and then test the PR actually adding the e2e testing.

PR #66948, opened 2025-07-11
Append the other gsutil logs

liangxia:
The log from later gsutil commands overwrite the earlier logs, this is unexpected.  Fix that to append the logs.


liangxia:
Minor and simple change.

/pj-rehearse skip

PR #66961, opened 2025-07-11
add qe cases in cvo pre-submit jobs

jiajliu:
This pr is to integrate QE's e2e cases into pre-submit jobs. The test will execute all cvo e2e tests in openshift-tests-private repo. It can be triggered manually in cvo prs. After running for a period, we can further divide the cases into multiple jobs and then consider whether to remove `always_run: false` and control the job's execution in each PR via `skip_if_only_changed`/`run_if_changed`. Generally, there are three scenarios that we can use the test during pre-merge test:

1. For prs without specific cases, but a general regression test is needed, then choose pre-submit jobs to run overall test. 
2. For prs with existing case’s coverage, then choose pre-submit jobs including the case to test in prs. 
3. For prs related with new feature, before OTE is ready, we might continue preparing automation case in openshift-tests-private repo with dev’s pr in parallel. Then refer to 2) to run a pre-submit test in the pr directly.

### DIFF
--- a/ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-main.yaml
+++ b/ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-main.yaml
@@ -19,6 +19,10 @@ base_images:
     name: builder
     namespace: ocp
     tag: rhel-9-golang-1.24-openshift-4.20
+  tests-private:
+    name: tests-private
+    namespace: ci
+    tag: "4.20"
   yamllint:
     name: rust
     namespace: cincinnati-ci-public
@@ -92,6 +96,18 @@ tests:
           cpu: 100m
           memory: 200Mi
     workflow: openshift-e2e-azure
+- always_run: false
+  as: e2e-extended-tests
+  optional: true
+  steps:
+    cluster_profile: gcp-qe
+    env:
+      TEST_FILTERS: ~DisconnectedOnly&
+      TEST_SCENARIOS: OTA cvo
+    test:
+    - chain: openshift-e2e-test-ota-qe
+    workflow: cucushift-installer-rehearse-gcp-ipi
+  timeout: 8h0m0s
 - as: e2e-agnostic-ovn-upgrade-into-change
   skip_if_only_changed: ^docs/|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|PROJECT|LICENSE)$
   steps:
--- a/ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-release-4.16.yaml
+++ b/ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-release-4.16.yaml
@@ -19,6 +19,10 @@ base_images:
     name: builder
     namespace: ocp
     tag: rhel-9-golang-1.21-openshift-4.16
+  tests-private:
+    name: tests-private
+    namespace: ci
+    tag: "4.16"
 build_root:
   from_repository: true
 images:
@@ -88,6 +92,18 @@ tests:
           cpu: 100m
           memory: 200Mi
     workflow: openshift-e2e-azure
+- always_run: false
+  as: e2e-extended-tests
+  optional: true
+  steps:
+    cluster_profile: gcp-qe
+    env:
+      TEST_FILTERS: ~DisconnectedOnly&
+      TEST_SCENARIOS: OTA cvo
+    test:
+    - chain: openshift-e2e-test-ota-qe
+    workflow: cucushift-installer-rehearse-gcp-ipi
+  timeout: 8h0m0s
 - as: e2e-agnostic-ovn-upgrade-into-change
   skip_if_only_changed: ^docs/|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|PROJECT|LICENSE)$
   steps:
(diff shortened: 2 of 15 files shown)


jiajliu:
/pj-rehearse pull-ci-openshift-cluster-version-operator-main-e2e-extended-tests

petr-muller:
:+1: :+1: :+1: :100: 

Starting with manually triggered optional sounds like a good baseline, but I wonder if we can be more ambitious (manually triggered required is also still an option - that way humans using the job are incentivized to address flakes and be explicit in overriding failures).

Some questions (for my understanding):

-What are the typical PRs where we'd absolutely want to trigger this job? 
- And the opposite, what are the PRs that would not  (besides the obvious "docs" and the like)
- If we have this job, how much effort it saves downstream?
- If this job fails and discovers a bug, assuming other CI jobs would not discover the bug, when would we typically discover that bug today?
- Do we have some idea about how much the job overlaps with existing CI jobs?

jiajliu:
/pj-rehearse pull-ci-openshift-cluster-version-operator-release-4.19-e2e-extended-tests

jiajliu:
> but I wonder if we can be more ambitious

Generally the test will execute all cvo cases in openshift-tests-private repo and so it will take a long time to run. Setting `always_run: false` is to not bring noise for all cvo prs. QE can trigger it according to test requirements in pre-merge test. After running for a period, we can further divide the cases into multiple jobs to shorten the test time based on feedback, and then consider whether to remove `always_run: false` and control the job's execution in each PR via `skip_if_only_changed`/`run_if_changed`.

> -What are the typical PRs where we'd absolutely want to trigger this job?
> 
> * And the opposite, what are the PRs that would not  (besides the obvious "docs" and the like)

Adding QE's e2e test cases in pre-submit job is to help us do pre-merge testing and discover issues earlier, especially for prs which need regression test and backport prs. Of course, before OTE is ready, we might continue preparing automation case in openshift-tests-private repo. Then it would be easy to run a pre-submit test if the automation pr is merged before dev's pr is ready for pre-merge test.


> * If we have this job, how much effort it saves downstream?
> * If this job fails and discovers a bug, assuming other CI jobs would not discover the bug, when would we typically discover that bug today?
> * Do we have some idea about how much the job overlaps with existing CI jobs?

Regarding existing e2e tests in openshift-tests-private repo, they are currently fully covered in CI with a testing frequency of f28. If a case is reported as a failure in CI, the case owner needs to check the failure timely through CI tools(such as component readiness) . Therefore, relying on CI results in a relatively long cycle and need QE to proactively watch the ci failures.

jiajliu:
/pj-rehearse pull-ci-openshift-cluster-version-operator-release-4.19-e2e-extended-tests

jiajliu:
/assign @petr-muller 
The latest [test](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/66961/rehearse-66961-pull-ci-openshift-cluster-version-operator-release-4.19-e2e-extended-tests/1944947992863707136) has finished. @petr-muller please help review it, thanks

BTW, there are 3 cases failed(I assume it's not blocking the pr's merge): OCP-22641 and OCP-43172 sounds like flake as it passed in last [test](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/66961/rehearse-66961-pull-ci-openshift-cluster-version-operator-release-4.19-e2e-extended-tests/1944657317546954752), OCP-45879 failed in both test, cc @JianLi-RH for awareness.

petr-muller:
> Generally the test will execute all cvo cases in openshift-tests-private repo and so it will take a long time to run.

I am not concerned about a long time to run for a job that is not required for merges. When we are interested in the results of these tests then the latency is actually higher because we wait for someone to trigger the job and then we wait for the results. If it runs automatically then the person just reviews the results at the time when they would only trigger it in the first scenario. And in the cases when we are not interested in the results the run time does not matter. 

Running automatically will give us better data too; if we see failing job then we can use the history to see whether it was solid green before.

But we can make that call separately (maybe discuss on next quality call, maybe talk on Slack) and this PR is actually pretty clean (and I like the results).

Hopefully we can fix the flakes; it would be great if the job was green most of the time. But we can also make a team agreement that a `qe-approved` means that someone reviewed the failures (if present) confirmed they are flakes and then they can use `/skip` to make the job green.

petr-muller:
/pj-rehearse pull-ci-openshift-cluster-version-operator-main-e2e-extended-tests

JianLi-RH:
OCP-45879 has passed builds on all OCP versions except OCP 4.20, on OCP 4.20 there are only skipped builds.

jiajliu:
/pj-rehearse pull-ci-openshift-cluster-version-operator-main-e2e-extended-tests

jiajliu:
> Running automatically will give us better data too; if we see failing job then we can use the history to see whether it was solid green before.
> 
> But we can make that call separately (maybe discuss on next quality call, maybe talk on Slack) and this PR is actually pretty clean (and I like the results).
> 
> Hopefully we can fix the flakes; it would be great if the job was green most of the time. But we can also make a team agreement that a `qe-approved` means that someone reviewed the failures (if present) confirmed they are flakes and then they can use `/skip` to make the job green.

I agree that the test should eventually be triggered automatically. I plan to start a slack thread to get input from the team after the pr. Regarding the approach for watching test failures, your suggestion is a good one, and I think using `qe-approved` label is something we should explore.

gpei:
LGTM for the owner changes on `gcp-qe` cluster profile in `ci-operator/step-registry/cluster-profiles/cluster-profiles-config.yaml`, cc @liangxia for the approval, thanks.

liangxia (approved):

PR #66971, opened 2025-07-11
deploy latest toolchain resources

rsoaresd:
# Description
For the periodic jobs, we need o also set `DEPLOY_LATEST=true` in order to deploy the latest toolchain resources on the periodic jobs. If we do not set this, the scripts for installing the operators will be called and it will fail, ex: https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-codeready-toolchain-toolchain-e2e-master-ci-daily/1943589128746569728#1:build-log.txt%3A931.

Context: To achieve continuous monitoring, low-maintenance days, early detection, and more data for future metrics, we should add a Periodic CI job. This type of job will help us catch flaky tests earlier, since we expect that the pass rate is 100%, since we are testing against the code that is already in production. Also, on the Logilica side, we will be able to have metrics like success rate, failure rate, ...

## Issue ticket number and link
[SANDBOX-801](https://issues.redhat.com/browse/SANDBOX-801)

### DIFF
--- a/ci-operator/config/codeready-toolchain/toolchain-e2e/codeready-toolchain-toolchain-e2e-master.yaml
+++ b/ci-operator/config/codeready-toolchain/toolchain-e2e/codeready-toolchain-toolchain-e2e-master.yaml
@@ -41,7 +41,7 @@ tests:
     test:
     - as: test
       cli: latest
-      commands: make test-e2e CI_DISABLE_PAIRING=true
+      commands: make test-e2e CI_DISABLE_PAIRING=true DEPLOY_LATEST=true
       from: src
       resources:
         requests:


MatousJobanek (approved):

ci-operator/config/codeready-toolchain/toolchain-e2e/codeready-toolchain-toolchain-e2e-master.yaml
```diff
@@ -41,7 +41,7 @@ tests:
     test:
     - as: test
       cli: latest
-      commands: make test-e2e CI_DISABLE_PAIRING=true
+      commands: make test-e2e CI_DISABLE_PAIRING=true DEPLOY_LATEST=true
```
MatousJobanek, line 44: based on the other PR, shouldn't this be just:
```suggestion
      commands: make test-e2e DEPLOY_LATEST=true
```

rsoaresd: We still need `CI_DISABLE_PAIRING=true` because of the ksctl https://github.com/codeready-toolchain/toolchain-e2e/blob/master/make/ksctl.mk#L17

MatousJobanek: Gotcha, thanks. we could technically replace it with `DEPLOY_LATEST` too, but I'm not sure if Konflux doesn't rely on that in their CI. So it's fine to keep it as it is :+1:

rsoaresd: Thanks!!

PR #66973, opened 2025-07-11
Add installer/release to the owner of qe cloud cluster profiles

gpei:
To allow using qe cluster profiles in those repos.


gpei:
@jianlinliu @yunjiang29 @jinyunma @jianli-wei for the review, thanks

PR #66996, opened 2025-07-12
HyperShift: Fix hypershift-mce-kubevirt-gpu-conformance

LiangquanLi930:


ci-operator/step-registry/hypershift/mce/kubevirt/gpu/nfd-operator/hypershift-mce-kubevirt-gpu-nfd-operator-commands.sh
```diff
@@ -79,6 +79,6 @@ if [ $_apiReady -eq 0 ]; then
 fi
 echo "nfd-operator installed successfully"
 
-oc get csv -n openshift-nfd $CSVName -o jsonpath='{.metadata.annotations.alm-examples}' | jq '.[0]'  > /tmp/nodefeaturediscovery.json
+oc get csv -n openshift-nfd $CSVName -o jsonpath='{.metadata.annotations.alm-examples}' | jq '.[1]'  > /tmp/nodefeaturediscovery.json
```
orenc1, line 82: i think it's not robust to rely on the position of the item in the json array.
if we want to get the `NodeFeatureDiscovery`, i think it would be better to select it explicitly:
```
oc get csv -n openshift-nfd $CSVName -o jsonpath='{.metadata.annotations.alm-examples}' | jq '.[] | select(.kind=="NodeFeatureDiscovery")' > /tmp/nodefeaturediscovery.json
```

PR #67023, opened 2025-07-14
CNTRLPLANE-1054: Bump go version for OSSO, KDO and RODOO

ingvagabund:
SSIA


ardaguclu:
That is definitely nice to have. However, I'm under impression that if we update Go versions within the repositories, one of the bots defined for openshift/release will perform automatic update for us to reflect the reality in here.
/lgtm

ingvagabund:
You are right Arda. I need to make a note to update .ci-operator.yaml files as part of merging the dockerfile changes.

PR #67029, opened 2025-07-14
fix: allow users to approve without write access

Paul123111:
There was a bug ignoring users approving
with /approve if they didn't have write access
to community-catalog.

This commit fixes this by re-enabling
the lgtm plugin, and allowing /lgtm to act as
an approval if a user is an approver. This
prevents self-approval, while fixing
this bug.


Paul123111:
I also need `/ok-to-test`, I don't have permission to start tests myself

Paul123111:
I think this also needs `/lgtm`

PR #67066, opened 2025-07-15
CNTRLPLANE-1113: Skip E2E job execution once renovate config is modified

jparrill:


ci-operator/config/openshift/hypershift/openshift-hypershift-main.yaml
```diff
@@ -134,14 +134,14 @@ tests:
   commands: make verify
   container:
     from: custom-verify-src
-  skip_if_only_changed: (^(\.tekton|\.github|docs|examples|enhancements|contrib)/)|(^[A-Z]+\.md$)|((^|/)OWNERS$)|(/overrides\.yaml$)|(/\.testcoverage\.yml$)
+  skip_if_only_changed: (^(\.tekton|\.github|docs|examples|enhancements|contrib)/)|(^[A-Z]+\.md$)|((^|/)OWNERS$)|(/overrides\.yaml$)|(/renovate\.json$)|(/\.testcoverage\.yml$)
```
csrwng: I think it should be 
```
(^renovate.json$)
```
given that the file is at the root of the repo?

jparrill: yes, that makes sense, let me change the entries

PR #67086, opened 2025-07-15
cleanup the ARO-HCP basic install scripts to better reflect the minimums

deads2k:


deads2k:
azure portal looked ok. output looked good.

PR #67092, opened 2025-07-16
Integrate step junit2reportportal into exist chains

liangxia:


liangxia:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.20-arm64-nightly-aws-ipi-f1 periodic-ci-openshift-openshift-tests-private-release-4.20-amd64-nightly-aws-ipi-f1

ci-operator/step-registry/cucushift/installer/rehearse/aws/ipi/deprovision/cucushift-installer-rehearse-aws-ipi-deprovision-chain.yaml
```diff
@@ -1,6 +1,7 @@
 chain:
   as: cucushift-installer-rehearse-aws-ipi-deprovision
   steps:
+  - ref: junit2reportportal
```
yunjiang29: Will the depredation steps be reported if putting `junit2reportportal` to the end? I prefer to report all deprovision steps' results, @jianlinliu WDYT?

jianlinliu: Agree, let us place it in the end as much as possible.

liangxia: Yes, it can report all steps that executed before this step. I'll update

ci-operator/step-registry/junit2reportportal/junit2reportportal-commands.sh
```diff
@@ -80,6 +91,22 @@ function generate_attribute_cloud_provider() {
...
+function generate_attribute_install() {
+  install="fail"
+  for keyword in 'cucushift-e2e' \
+                 'cucushift-pre' \
+                 'openshift-extended-test' \
+                 'openshift-extended-test-supplementary' \
+                 'openshift-extended-web-tests'
+  do
+    if [[ -d "$LOCAL_DIR_ORI/$keyword" ]] ; then
+      install="succeeded"
```
jianlinliu: The attribute in jenkins pipeline, it is `succeed` rather than `succeeded`

liangxia: Updated.

ci-operator/step-registry/junit2reportportal/junit2reportportal-commands.sh
```diff
@@ -80,6 +91,22 @@ function generate_attribute_cloud_provider() {
   write_attribute cloud_provider "$cloud_provider"
 }
 
+function generate_attribute_install() {
+  install="fail"
+  for keyword in 'cucushift-e2e' \
```
jianlinliu: beside these common test steps, there are also some other test steps, e.g:

[openshift-e2e-test-mco-qe-longrun](https://github.com/openshift/release/blob/38833e0662977cca7a19c4bfdef4d6a10e07170b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.20__arm64-nightly.yaml#L129C15-L129C47)

[idp-htpasswd](https://github.com/openshift/release/blob/38833e0662977cca7a19c4bfdef4d6a10e07170b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.20__arm64-nightly.yaml#L251C12-L251C24)

[openshift-e2e-test-qe-disasterrecovery-sanity](https://github.com/openshift/release/blob/38833e0662977cca7a19c4bfdef4d6a10e07170b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.20__arm64-nightly.yaml#L304C14-L304C59)

If they are failed, the install would be marked as `fail` ?

jianlinliu: Maybe in the future, we probably have move steps are injected in the `test`, maybe we need to figure out some other ways to get pre stage results.

liangxia: Any test step (the real steps, those chains listed under `test` will be break into real steps) executed can indicate installation is success. We just need to list the common ones, no need to list all of them.

liangxia: If we have more test chains in the future, most of them will need to execute one of `cucushift-e2e (for cucushift )`, `openshift-extend-test (for ginkgo tests)`,  `openshift-extended-web-tests (for cypress tests)`, and/or `openshift-e2e-test-qe-report (for test reports in prow)`.

In case any special cases that can not be covered in above case, we can deal it later when it occurs.

liangxia: I know jenkins pipeline use `if 'Step phase pre succeeded after' not in build_log_text_job:` which search the specific string from `build-log.txt`, but this log file does not exist yet when we are in the step `junit2reportportal` as the job is still running.

yunjiang29: > maybe we need to figure out some other ways to get pre stage results.

My thought is, create a special step `generate-install-result` and add it to the end of each provision chain, then `generate_attribute_install` read the result directly. The benefit is that we don't need to maintain the _test names_ in here.
But for the current solution, it's acceptable to me, we can run it for a while and monitor to see if it is accurate.

ci-operator/step-registry/cucushift/installer/rehearse/aws/ipi/deprovision/cucushift-installer-rehearse-aws-ipi-deprovision-chain.yaml
```diff
@@ -6,5 +6,6 @@ chain:
   - ref: gather-aws-console
   - chain: ipi-deprovision
   - ref: aws-deprovision-users-and-policies
+  - ref: junit2reportportal
```
jianlinliu, line 9: How about appending this step to all the aws install workflow files, similar to https://github.com/openshift/release/pull/67158, I guess we can do that by `find . -name "*-workflow.yaml" -type f | xargs -I {} yq-v4 eval '.workflow.steps.post += {"ref": "junit2reportportal"}' {} -i`

Or I can append it directly in https://github.com/openshift/release/pull/67158, you can remove it here.

liangxia: I think we can do the append and removal in a follow up PR, because we need this line here so that we can run those rehearsal jobs.

jianlinliu: #67158 is ready to review, maybe I can append `junit2reportportal` together in that PR, once it get merged, then you can rebase your branch, and update the installation marker file list  in this PR together to run your testing.

liangxia: Because I don't want to run the rehearsal again and again, it's time consuming and boring.

jianlinliu: When we are going to update https://github.com/openshift/release/pull/67092/files#diff-f88af2897556d4dfefda4431b3391863ea3cb7422a416d99ca4a12b6cd394897R96, still need some rehearse testing, right ?

liangxia: I don't think add one more line in that place need another rehearsal.

jianlinliu: If you think it is safe, you can skip the rehearse.

liangxia:
/override ci/prow/generated-config
( this is related to https://redhat-internal.slack.com/archives/CFUGK0K9R/p1752787713565449 )

PR #67100, opened 2025-07-16
NO-JIRA: ci-operator: Specify shard_count for openshift-apiserver and oc

tchap:
Set `shard_count: 2` for selected CI jobs that take a long time.

### DIFF
--- a/ci-operator/config/openshift/openshift-apiserver/openshift-openshift-apiserver-main.yaml
+++ b/ci-operator/config/openshift/openshift-apiserver/openshift-openshift-apiserver-main.yaml
@@ -81,6 +81,7 @@ tests:
     cluster_profile: aws-2
     workflow: openshift-upgrade-aws
 - as: e2e-aws-ovn-serial
+  shard_count: 2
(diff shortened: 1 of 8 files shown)


ci-operator/config/openshift/oc/openshift-oc-main.yaml
```diff
@@ -150,26 +150,31 @@ tests:
...
+  shard_count: 2
   steps:
     cluster_profile: aws-3
     workflow: openshift-e2e-aws
 - as: e2e-aws-ovn-serial
+  shard_count: 2
   steps:
     cluster_profile: aws-2
     workflow: openshift-e2e-aws-serial
   timeout: 5h0m0s
```
ardaguclu: I increased this to 5h to mitigate the issue. But maybe from now on, we have a better solution and we can reduce this to 4h again. But this is not the main goal of this PR anyway. Just not to forget.

ardaguclu:
Thank you for taking care of this. Since we only shard aws-serial without any other changes, I don't think we need to run CI jobs for testing.
/pj-rehearse ack
/lgtm
I think, we need approval from openshift-apiserver.

tchap:
Adding a label back.

/pj-rehearse ack

dgrisonnet (approved):
/lgtm
/approve

PR #67139, opened 2025-07-16
add kueue main to prow

kannon92:
Add e2e test that uses konflux kueue main images.

This is mostly for signal on e2e in case we want to fix anything to help with rebases.


sohankunkerkar (commented):
/lgtm

PR #67167, opened 2025-07-17
periodic-org-sync: filter out `openshift-priv` 

hoxhaeris:
### Context

GitHub access management for `openshift-priv` is transitioning from the current periodic-org-sync process to **Cyborg**. Cyborg will use the Org repository as source-of-truth to dynamically generate and apply Peribolos configurations for Openshift-Priv.

### Changes

Modified `periodic-org-sync` to filter out `openshift-priv` at runtime:

- Added shell preprocessing to remove `openshift-priv` section from `org/config.yaml`
- Multiple safety checks ensure clean removal and valid output
- Uses existing distroless peribolos image (no new dependencies)
- Preserves functionality for all other organizations

**Safety features:**
- Validates `openshift-priv` exists exactly once before processing
- Confirms clean removal and well-formed YAML after processing  
- Fails fast with clear errors if any validation fails

/cc @jupierce 
/cc @bradmwilliams

### DIFF
--- a/ci-operator/jobs/infra-periodics.yaml
+++ b/ci-operator/jobs/infra-periodics.yaml
@@ -1495,24 +1495,56 @@ periodics:
   name: periodic-org-sync
   spec:
     containers:
-    - args:
-      - --config-path=./org/config.yaml
-      - --confirm=true
-      - --github-token-path=/etc/github/oauth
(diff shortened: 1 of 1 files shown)


bradmwilliams (commented):
/lgtm

PR #67172, opened 2025-07-17
Remove self and my old team

baijum:


baijum:
@pmacik @sbose78 Can you add a '/lgtm' comment here?

PR #67177, opened 2025-07-17
sync main config with release-1.0

cpmeadors:
We are restarting work on main.  This moves all the work we did for release-1.0 to main so we have the same testing.


ci-operator/config/openshift/kueue-operator/openshift-kueue-operator-main.yaml
```diff
@@ -66,6 +59,140 @@ tests:
...
+        REVISION=$(git log --oneline -1 | awk '{print $4}' | tr -d "'")
+        echo "Current Git branch:"
+        git branch --show-current
+
+        echo "Latest Git commits:"
+        git log --oneline -5
+
+        echo "Git status:"
+        git status
+        export OPERATOR_IMAGE=quay.io/redhat-user-workloads/kueue-operator-tenant/kueue-operator-1-0:on-pr-${REVISION}
```
kannon92: This would need to change.

This would probably need to be the konflux main images that you create.

kannon92: There are a few items like this that will change once you have the konflux images.

kannon92: can you update this?

ci-operator/config/openshift/kueue-operator/openshift-kueue-operator-main.yaml
```diff
@@ -66,6 +59,140 @@ tests:
...
```
kannon92: Use the main konflux images.

kannon92:
/hold

Now that we have our konflux images for main, let's use those for the e2e tests.

ci-operator/config/openshift/kueue-operator/openshift-kueue-operator-main.yaml
```diff
@@ -66,6 +59,140 @@ tests:
   skip_if_only_changed: ^\.tekton/|\.md$|^(LICENSE|OWNERS)$
   steps:
     workflow: openshift-ci-security
+- as: test-e2e-4-17
```
kannon92, line 82: Can you replace 4-17 with 4.19?

I have a PR to make 4.18 and 4.19 required jobs and 4.17 optional?

Could you update this also?

MaysaMacedo, line 82: If we should be testing the operator changes, shouldn't it be similar to what was done on release 1-0 for the operator repo?
https://github.com/openshift/release/blob/master/ci-operator/config/openshift/kueue-operator/openshift-kueue-operator-release-1.0.yaml#L81-L94

MaysaMacedo (changes requested):

cpmeadors:
Looks like the step "kueue-operator-image-env-setup" needs to be parameterized to take value to make the images names match the branch.

PR #67225, opened 2025-07-18
Moved skip_if_only_changed from job do config

hector-vido:


2uasimojo:
I can't load this up in the gh UI, but assuming the back end has been updated accordingly to support it, I love this plan!

PR #67228, opened 2025-07-18
SREP-1170: Delete boilerplate presubmits that moved to Konflux

joshbranham:
We are migrating the only remaining CI job `make pr-check` to Konflux, where we also build the image for `boilerplate`. We run the test in the context of the image, so it makes more sense to not have divergent CI/CD running.


joshbranham:
Depends on https://github.com/openshift/boilerplate/pull/575

PR #67229, opened 2025-07-18
make 4.18 and 4.19 jobs required and 4.17 jobs optional

kannon92:
Update kueue-operator and kueue to make 4.18 and 4.19 required and 4.17 jobs optional.


kannon92:
@MaysaMacedo can you add your lgtm when you get a chance?

PR #67235, opened 2025-07-18
KSVM: Use the merge base of base and PR refs as commitchecker's start.

benluddy:
When the base isn't a direct ancestor of the PR head, verify-commits passes without looking at any commits. Use the merge base of the two refs instead.


benluddy:
I think most verify-commits jobs will want to do this. Check out this test PR to openshift/kubernetes (https://github.com/openshift/kubernetes/pull/2368) from a branch that isn't a direct descendant of the current master.

/pj-rehearse

Expecting to see something like "no merge base found" in the rehearsal since PULL_BASE_SHA is an openshift/release SHA.

sanchezl (approved):
/lgtm

PR #67243, opened 2025-07-18
step-registry/openshift-extended-test: allow to print the output of successful tests

wainersm:
Added the TEST_INCLUDE_SUCCESS parameter to the openshift-extended-test step that when setting to 'yes' will print the output of successful tests.


beraldoleal (approved):
LGTM, thanks

PR #67247, opened 2025-07-19
update: make rhoai branch run with 4.18 , keep main run with 4.17, so we can cover both cases

zdtsw:
related to https://github.com/opendatahub-io/opendatahub-operator/pull/2191

as long as we run e2e and pass on both "main" and "rhoai" branch, we are fine with test coverage
use main with 4.17 can make the test loop shorter for now

in long term, we need to define what to do with logic on different ocp with external operator avaiablility


zdtsw:
cc @ugiordan

CFSNM (approved):
/approved /lgtm

PR #67260, opened 2025-07-21
Fix attribute architecture in ReportPortal for automated release jobs

liangxia:
automated release jobs does not contain `architecture` in its names, so we need to fetch that info from its versions which is stored in `release-images-*` files.


liangxia:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.16-automated-release-aws-ipi-disruptive-f999

PR #67273, opened 2025-07-21
Adjust resource requests for junit2reportportal

liangxia:


droslean (approved):
Are these resources requests being used already? Why do we need to change them since we have a mechanism to auto-tune them on the runtime?

liangxia:
> Are these resources requests being used already? Why do we need to change them since we have a mechanism to auto-tune them on the runtime?

We send test results to ReportPortal from both Jenkins CI and Prow CI. but we got some strange behaviors on the ReportPortal side. I'd like to minimize the differences from them to help the Data and Operation team to find the root cause of the issue.

droslean:
> > Are these resources requests being used already? Why do we need to change them since we have a mechanism to auto-tune them on the runtime?
> 
> We send test results to ReportPortal from both Jenkins CI and Prow CI. but we got some strange behaviors on the ReportPortal side. I'd like to minimize the differences from them to help the Data and Operation team to find the root cause of the issue.

example: https://pod-scaler-ui-ci.apps.ci.l2s4.p1.openshiftapps.com/usage/steps?branch=main&container=test&org=openshift&repo=verification-tests&step=junit2reportportal&target=aws-ipi-shared-phz-sts-f14&variant=installation-nightly-4.16

I mean that your change in this PR won't reflect anything.

PR #67278, opened 2025-07-21
add e2e test config for common templates v0.34

fossedihelm:
add e2e test config for common templates v0.34


fossedihelm:
> /hold

Can we unhold this?

PR #67283, opened 2025-07-21
fix main kueue finding wrong image

kannon92:
kubernetes main is not kueue-main but kubernetes-sigs-kueue-main.


kannon92:
cc @MaysaMacedo @cpmeadors

PR #67286, opened 2025-07-21
Fixup containers/container-selinux config

lsm5:


psalajova (approved):

lsm5:
@psalajova good to go. PTAL

PR #67292, opened 2025-07-21
NO-ISSUE: start fast-forward for ocm-2.16 instead of ocm-2.15

linoyaslan:
As part of the creating of ocm-2.16 branch, we want to switch between 2.15 and 2.16, so 2.16 will follow master now.
/cc @danmanor


linoyaslan:
NOTE: it's a test. will be closed once all tests pass

PR #67294, opened 2025-07-21
acm - business continuity CI checks update

tesshuflower:
- release 2.10 will no longer be built with konflux, so remove requirements for the konflux 2.10 check
- main is ffwding to release-2.15 which is now in konflux, so add CI requirement for the konflux 2.15 check


birsanv (approved):
lgtm

PR #67305, opened 2025-07-21
don't branch for rhel-coreos-config repo

dustymabe:
- also clean up 4.20/4.21 related config for rhel-coreos-config
- also add testing for rhel10


dustymabe:
Looks like everything passed. When I click through tide it says:

```
Only merges with author openshift-bot are currently allowed
```

dustymabe:
Now it says in merge pool so I think we;re good

PR #67306, opened 2025-07-21
Channel group fix

Amarthya-v:
This PR updates the rosa-cluster-provision-commands.sh script to enhance version resolution logic for OpenShift cluster provisioning.

Instead of only checking the nightly channel group, the script now:

Checks across all three channel groups: nightly, candidate, and stable

Resolves the correct channel group dynamically based on the user-specified OpenShift version

Currently, the rosa-classic-sts-conformance jobs fail when the desired OpenShift version (e.g., 4.19.5) is only available in candidate or stable, but the script assumes nightly.
This update prevents such failures by dynamically finding where the version is available.


Amarthya-v:
Closing in favor of a cleaner version:

PR #67308, opened 2025-07-21
Support resolving OpenShift version across nightly, candidate, and stable channel groups

Amarthya-v:
This PR updates the rosa-cluster-provision-commands.sh script to resolve the OpenShift version by checking all three channel groups - nightly, candidate, and stable.
Previously, the script would only check one channel group, which led to failures when versions existed only in a different channel.

Resolves: [SREP-1210](https://issues.redhat.com//browse/SREP-1210) https://issues.redhat.com/browse/SREP-1210

### DIFF
--- a/ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
+++ b/ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
@@ -10,7 +10,7 @@ STS=${STS:-true}
 HOSTED_CP=${HOSTED_CP:-false}
 COMPUTE_MACHINE_TYPE=${COMPUTE_MACHINE_TYPE:-"m5.xlarge"}
 OPENSHIFT_VERSION=${OPENSHIFT_VERSION:-}
-CHANNEL_GROUP=${CHANNEL_GROUP}
+EC_BUILD=${EC_BUILD:-false}
 MULTI_AZ=${MULTI_AZ:-false}
 EC2_METADATA_HTTP_TOKENS=${EC2_METADATA_HTTP_TOKENS:-"optional"}
 ENABLE_AUTOSCALING=${ENABLE_AUTOSCALING:-false}
@@ -26,7 +26,6 @@ FIPS=${FIPS:-false}
 PRIVATE=${PRIVATE:-false}
 PRIVATE_LINK=${PRIVATE_LINK:-false}
 PRIVATE_SUBNET_ONLY="false"
-CLUSTER_TIMEOUT=${CLUSTER_TIMEOUT}
 ENABLE_SHARED_VPC=${ENABLE_SHARED_VPC:-"no"}
 CLUSTER_SECTOR=${CLUSTER_SECTOR:-}
 ADDITIONAL_SECURITY_GROUP=${ADDITIONAL_SECURITY_GROUP:-false}
@@ -158,20 +157,6 @@ if [[ ! -z "$OLD_CLUSTER" ]]; then
   exit 1
 fi
 
-# Get the openshift version
-version_cmd="rosa list versions --channel-group ${CHANNEL_GROUP} -o json"
-if [[ "$HOSTED_CP" == "true" ]]; then
-  version_cmd="$version_cmd --hosted-cp"
-fi
-if [[ ${AVAILABLE_UPGRADE} == "yes" ]] ; then
-  # shellcheck disable=SC2089
-  version_cmd="$version_cmd | jq -r '.[] | select(.available_upgrades!=null) .raw_id'"
-else
-  version_cmd="$version_cmd | jq -r '.[].raw_id'"
-fi
-versionList=$(eval $version_cmd)
-echo -e "Available cluster versions:\n${versionList}"
-
 # If OPENSHIFT_VERSION is set to "release:latest", look at the environment variable
 # supplied by CI for the payload to use. This only really works for nightlies. ROSA
 # SRE has a job that polls the release controller and syncs new nightlies every 15 minutes,
@@ -186,10 +171,10 @@ if [[ "$OPENSHIFT_VERSION" = "release:latest" ]]; then
   start_time=$(date +%s)
 
   while true; do
-    versionList=$(eval $version_cmd)
-    echo -e "Looking for $PAYLOAD_TAG in available cluster versions:\n${versionList}"
+    version_list=$(eval $version_cmd)
+    echo -e "Looking for $PAYLOAD_TAG in available cluster versions:\n${version_list}"
     # Check if image has been synced yet
-    if echo "$versionList" | grep -q "$PAYLOAD_TAG"; then
+    if echo "$version_list" | grep -q "$PAYLOAD_TAG"; then
       echo "$PAYLOAD_TAG is available from ROSA, continuing..."
       OPENSHIFT_VERSION=$PAYLOAD_TAG
       break
@@ -217,32 +202,76 @@ if [[ "$OPENSHIFT_VERSION" = "release:latest" ]]; then
   done
 fi
 
-if [[ -z "$OPENSHIFT_VERSION" ]]; then
-  if [[ "$EC_BUILD" == "true" ]]; then
-    OPENSHIFT_VERSION=$(echo "$versionList" | grep -i ec | head -1 || true)
-  else
-    OPENSHIFT_VERSION=$(echo "$versionList" | head -1)
-  fi
-elif [[ $OPENSHIFT_VERSION =~ ^[0-9]+\.[0-9]+$ ]]; then
-  if [[ "$EC_BUILD" == "true" ]]; then
-    OPENSHIFT_VERSION=$(echo "$versionList" | grep -E "^${OPENSHIFT_VERSION}" | grep -i ec | head -1 || true)
-  else
-    OPENSHIFT_VERSION=$(echo "$versionList" | grep -E "^${OPENSHIFT_VERSION}" | head -1 || true)
-  fi
-else
-  # Match the whole line
-  OPENSHIFT_VERSION=$(echo "$versionList" | grep -x "${OPENSHIFT_VERSION}" || true)
+# Get the openshift version
+if [[ -z "${CHANNEL_GROUP}" ]]; then
+  log "Error: CHANNEL_GROUP variable is empty"
+  exit 1
 fi
+IFS=',' read -r -a CHANNEL_GROUP_ARRAY <<< "${CHANNEL_GROUP// /}"
 
-if [[ -z "$OPENSHIFT_VERSION" ]]; then
-  echo "Requested cluster version not available!"
+CHANNEL_GROUP=""
+
+for GROUP in "${CHANNEL_GROUP_ARRAY[@]}"; do
+  log "Checking ${GROUP} channel..."
+
+  version_cmd="rosa list versions --channel-group ${GROUP} -o json"
+  if [[ "${HOSTED_CP}" == "true" ]]; then
+    version_cmd="${version_cmd} --hosted-cp"
+  fi
+
+  if [[ "${AVAILABLE_UPGRADE}" == "yes" ]]; then
+    version_list=$(eval "${version_cmd}" | jq -r '.[] | select(.available_upgrades!=null) .raw_id')
+  else
+    version_list=$(eval "${version_cmd}" | jq -r '.[].raw_id')
+  fi
+
+  if [[ -n "${version_list}" ]]; then
+    log "Found versions in ${GROUP} channel"
+
+    if [[ -z "$OPENSHIFT_VERSION" ]]; then
+      if [[ "$EC_BUILD" == "true" ]]; then
+        temp_version=$(echo "$version_list" | grep -i ec | head -1 || true)
+      else
+        temp_version=$(echo "$version_list" | head -1 || echo "")
+      fi
+    elif [[ $OPENSHIFT_VERSION =~ ^[0-9]+\.[0-9]+$ ]]; then
+      if [[ "$EC_BUILD" == "true" ]]; then
+        temp_version=$(echo "$version_list" | grep -E "^${OPENSHIFT_VERSION}" | grep -i ec | head -1 || true)
+      else
+        temp_version=$(echo "$version_list" | grep -E "^${OPENSHIFT_VERSION}" | head -1 || true)
+      fi
+    else
+      temp_version=$(echo "$version_list" | grep -x "${OPENSHIFT_VERSION}" || true)
+    fi
+    
+    if [[ -n "$temp_version" ]]; then
+      OPENSHIFT_VERSION="$temp_version"
+      CHANNEL_GROUP="${GROUP}"
+      break
+    else
+      log "Target version not available in ${GROUP} channel"
+    fi
+  else
+    log "No versions found in ${GROUP} channel"
+  fi
+done
+
+if [[ -z "${CHANNEL_GROUP}" ]]; then
+  log "Error: No versions found in any of the specified channel groups: ${CHANNEL_GROUP}"
   exit 1
 fi
 
-# if [[ "$CHANNEL_GROUP" != "stable" ]]; then
-#   OPENSHIFT_VERSION="${OPENSHIFT_VERSION}-${CHANNEL_GROUP}"
-# fi
-log "Choosing openshift version ${OPENSHIFT_VERSION}"
+log "Proceeding with ${CHANNEL_GROUP} channel group"
+
+if [[ -z "$OPENSHIFT_VERSION" ]]; then
+  echo "Requested cluster version not available in any of the specified channel groups: ${CHANNEL_GROUP}"
+  exit 1
+fi
+
+if [[ "$CHANNEL_GROUP" != "stable" ]]; then
+  OPENSHIFT_VERSION="${OPENSHIFT_VERSION}-${CHANNEL_GROUP}"
+fi
+log "Choosing OpenShift version ${OPENSHIFT_VERSION}"
 
 # Add USER_TAGS to help with cloud cost and make sure to remove
 # invalid tags so that CLI usage doesn't fail later.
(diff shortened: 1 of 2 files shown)


ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
```diff
@@ -159,19 +161,36 @@ if [[ ! -z "$OLD_CLUSTER" ]]; then
...
+    versionList=$(eval $version_cmd | jq -r '.[].raw_id')
+  fi
+
+  if echo "$versionList" | grep -q "^${TARGET_VERSION}$"; then
+    echo "Found ${TARGET_VERSION} in $GROUP channel"
+    RESOLVED_CHANNEL_GROUP="$GROUP"
+    break
+  fi
+
+  sleep 1  # small delay between calls for safety
```
btoll: Is this needed?

ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
```diff
@@ -159,19 +161,36 @@ if [[ ! -z "$OLD_CLUSTER" ]]; then
...
-fi
-if [[ ${AVAILABLE_UPGRADE} == "yes" ]] ; then
-  # shellcheck disable=SC2089
-  version_cmd="$version_cmd | jq -r '.[] | select(.available_upgrades!=null) .raw_id'"
-else
-  version_cmd="$version_cmd | jq -r '.[].raw_id'"
+for GROUP in "${CHANNEL_GROUPS[@]}"; do
+  echo "Checking $GROUP for ${TARGET_VERSION}..."
+  version_cmd="rosa list versions --channel-group ${GROUP} -o json"
+  if [[ "${HOSTED_CP:-}" == "true" ]]; then
```
btoll: Why this syntax (using a default value if var is empty)?

ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
```diff
@@ -159,19 +161,36 @@ if [[ ! -z "$OLD_CLUSTER" ]]; then
...
+    version_cmd="$version_cmd --hosted-cp"
+  fi
+
+  if [[ "${AVAILABLE_UPGRADE:-}" == "yes" ]]; then
```
btoll: Why this syntax (using a default value if var is empty)?

ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
```diff
@@ -159,19 +161,36 @@ if [[ ! -z "$OLD_CLUSTER" ]]; then
...
+    versionList=$(eval $version_cmd | jq -r '.[] | select(.available_upgrades!=null) .raw_id')
+  else
+    versionList=$(eval $version_cmd | jq -r '.[].raw_id')
+  fi
+
+  if echo "$versionList" | grep -q "^${TARGET_VERSION}$"; then
+    echo "Found ${TARGET_VERSION} in $GROUP channel"
+    RESOLVED_CHANNEL_GROUP="$GROUP"
```
btoll: Why not just use `CHANNEL_GROUP` here instead of creating a new var (`RESOLVED_CHANNEL_GROUP`)?

ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
```diff
@@ -10,7 +10,9 @@ STS=${STS:-true}
 HOSTED_CP=${HOSTED_CP:-false}
 COMPUTE_MACHINE_TYPE=${COMPUTE_MACHINE_TYPE:-"m5.xlarge"}
 OPENSHIFT_VERSION=${OPENSHIFT_VERSION:-}
-CHANNEL_GROUP=${CHANNEL_GROUP}
+TARGET_VERSION="${VERSION}"
+RESOLVED_CHANNEL_GROUP=""
```
btoll: This probably isn't needed.

ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
```diff
@@ -159,19 +161,36 @@ if [[ ! -z "$OLD_CLUSTER" ]]; then
...
+  sleep 1  # small delay between calls for safety
+
+done
+
+if [[ -z "$RESOLVED_CHANNEL_GROUP" ]]; then
+  echo "Version ${TARGET_VERSION} not found in any channel group (nightly/candidate/stable)"
+  exit 1
 fi
-versionList=$(eval $version_cmd)
-echo -e "Available cluster versions:\n${versionList}"
```
btoll, line 173: Curious why this is being removed?

btoll:
@Amarthya-v Here are some comments that apply to the whole PR:
- If you continue to use `CHANNEL_GROUP` then a lot of the churn isn't needed (i.e., replacing `CHANNEL_GROUP` with `RESOLVED_CHANNEL_GROUP`).
- Could you please shellcheck the code that you've changed?  there are a lot of `shellcheck` warnings in the entire file, and fixing the ones in the lines of code that you've changed would help chip away at that.
- Do you think it would be worth allowing a specific channel to be searched and, if not, then defaulting to all channels?
- There is a mixing of variable syntax, both `I_AM_VARIABLE` and `{I_AM_VARIABLE}`?  Could you just use one for consistency?  Maybe try and determine which is used the most throughout the script.

btoll:
Also, it looks like that if the same version string exists in both `candidate` and `stable` that the `candidate` channel will always be used.

ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
Amarthya-v: this was added as a precaution but the impact is minimal so I removed it.

ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
Amarthya-v: I simplified the condition now

ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
Amarthya-v: I simplified the condition now

ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
Amarthya-v: I removed the original single-channel logic (eval $version_cmd) because the script now supports iterating through multiple channel groups (CHANNEL_GROUPS[@]). 
now I kept the echo -e "Available cluster versions..." log inside the loop for visibility during each iteration

ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
```diff
@@ -10,7 +10,8 @@ STS=${STS:-true}
 HOSTED_CP=${HOSTED_CP:-false}
 COMPUTE_MACHINE_TYPE=${COMPUTE_MACHINE_TYPE:-"m5.xlarge"}
 OPENSHIFT_VERSION=${OPENSHIFT_VERSION:-}
-CHANNEL_GROUP=${CHANNEL_GROUP}
+TARGET_VERSION="${VERSION}"
```
btoll: Why do you need a new variable `TARGET_VERSION` here?  Can't you just use `VERSION`?

ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
```diff
@@ -10,7 +10,8 @@ STS=${STS:-true}
...
+IFS=' ' read -r -a CHANNEL_GROUPS <<< "${CHANNEL_GROUPS_OVERRIDE:-nightly candidate stable}"
```
btoll: Where would the value for `CHANNEL_GROUPS_OVERRIDE` be coming from?  Also, this could lead to the same situation as before, where a `channel` value is given but then nothing is found.

I think what would need to happen is that if a `channel` value was given that the other channels would then still need to be searched if the desired `channel` doesn't contain the version.

ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
```diff
@@ -159,19 +160,35 @@ if [[ ! -z "$OLD_CLUSTER" ]]; then
...
+    version_cmd="${version_cmd} --hosted-cp"
+  fi
+
+  if [[ "${AVAILABLE_UPGRADE}" == "yes" ]]; then
+    versionList=$(eval "${version_cmd}" | jq -r '.[] | select(.available_upgrades!=null) .raw_id')
+  else
+    versionList=$(eval "${version_cmd}" | jq -r '.[].raw_id')
+  fi
+
+  echo -e "Available cluster versions:\n${versionList}"
```
btoll: Should the versions be listed for every channel or just the one that contains the version?

btoll:
https://github.com/openshift/release/pull/67308#issuecomment-3099912231

@Amarthya-v This situation could still occur.

ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
Amarthya-v: not actually, I have changed to only display available versions when it is found in a channel.

Amarthya-v: not actually, I have changed to only display available versions when it is found in a channel.

ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
Amarthya-v: I have updated the logic to address both concerns:

If CHANNEL_GROUPS_OVERRIDE is set, it’s treated as a preferred channel and prepended to the default list (stable candidate nightly).

I have used awk '!seen[$0]++' to deduplicate any repeated channels, ensuring all fallbacks are still checked.

if a version isn’t found in the preferred channel, it will continue checking the others in order.

ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
Amarthya-v: I have addressed it, thank you!

ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
```diff
@@ -159,19 +157,42 @@ if [[ ! -z "$OLD_CLUSTER" ]]; then
...
 
 # Get the openshift version
-version_cmd="rosa list versions --channel-group ${CHANNEL_GROUP} -o json"
-if [[ "$HOSTED_CP" == "true" ]]; then
-  version_cmd="$version_cmd --hosted-cp"
-fi
-if [[ ${AVAILABLE_UPGRADE} == "yes" ]] ; then
-  # shellcheck disable=SC2089
-  version_cmd="$version_cmd | jq -r '.[] | select(.available_upgrades!=null) .raw_id'"
+PREFERRED_CHANNEL="${CHANNEL_GROUPS_OVERRIDE}"
```
btoll: Where is this value coming from?

Amarthya-v: CHANNEL_GROUPS_OVERRIDE is intended as an optional env var for CI jobs where we want to prioritize a specific channel. If it’s not defined, the logic falls back to the default search order: stable  candidate nightly. This way we maintain backward compatibility while giving flexibility when needed.

btoll: What about just using the same env var as before?

Amarthya-v: Yeah, we can use the same env var, but there was a naming collision,  the script was using CHANNEL_GROUP both as an input (preferred channel) and later as an output (the channel where the version was found). To avoid confusion, I renamed the output variable to something like SELECTED_CHANNEL in the updated version.

btoll: It's easier to make changes to this script than to have to additionally make changes to the job definition.

ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
```diff
@@ -159,19 +157,49 @@ if [[ ! -z "$OLD_CLUSTER" ]]; then
 fi
 
 # Get the openshift version
-version_cmd="rosa list versions --channel-group ${CHANNEL_GROUP} -o json"
-if [[ "$HOSTED_CP" == "true" ]]; then
-  version_cmd="$version_cmd --hosted-cp"
+if [[ -z "${CHANNEL_GROUPS}" ]]; then
```
btoll, line 160: I still think we should keep this as `CHANNEL_GROUP` singular.  The step-registry ref may not need to be changed.

ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
```diff
@@ -159,19 +157,49 @@ if [[ ! -z "$OLD_CLUSTER" ]]; then
...
+  exit 1
 fi
-if [[ ${AVAILABLE_UPGRADE} == "yes" ]] ; then
-  # shellcheck disable=SC2089
-  version_cmd="$version_cmd | jq -r '.[] | select(.available_upgrades!=null) .raw_id'"
-else
-  version_cmd="$version_cmd | jq -r '.[].raw_id'"
+IFS=',' read -r -a CHANNEL_GROUPS_ARRAY <<< "${CHANNEL_GROUPS// /}"
+
+if [[ ${#CHANNEL_GROUPS_ARRAY[@]} -eq 0 ]]; then
```
btoll: This check is unnecessary, the var is already being checked above.

ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-ref.yaml
```diff
@@ -53,8 +53,8 @@ ref:
   - name: OPENSHIFT_VERSION
     default: ""
     documentation: The openshift version for rosa to install (e.g. "4.10.12"). Specify a major/minor (e.g. "4.10") to get the latest version from that stream.
-  - name: CHANNEL_GROUP
-    default: "stable"
+  - name: CHANNEL_GROUPS
+    default: "stable,candidate,nightly"
```
btoll: Should this be "nightly,candidate, stable"?

btoll:
@Amarthya-v There's a `log` function https://github.com/Amarthya-v/release/blob/channel-group-fix-clean2/ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh#L35-L37, you probably want to replace most or all of your `echo` statements with this to follow the script's convention.

ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
```diff
@@ -159,19 +157,49 @@ if [[ ! -z "$OLD_CLUSTER" ]]; then
...
+  fi
+done
+
+if [[ -z "${CHANNEL_GROUP}" ]]; then
+  echo "Error: No versions found in any of the specified channel groups: ${CHANNEL_GROUPS}"
+  exit 1
+fi
+
+echo "Proceeding with ${CHANNEL_GROUP} channel group"
+echo -e "Available cluster versions:\n${versionList}"
```
btoll: This is back?  :)

ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
```diff
@@ -159,19 +157,49 @@ if [[ ! -z "$OLD_CLUSTER" ]]; then
...
+if [[ -z "${CHANNEL_GROUPS}" ]]; then
+  echo "Error: CHANNEL_GROUPS variable is empty"
+  exit 1
 fi
-if [[ ${AVAILABLE_UPGRADE} == "yes" ]] ; then
-  # shellcheck disable=SC2089
-  version_cmd="$version_cmd | jq -r '.[] | select(.available_upgrades!=null) .raw_id'"
-else
-  version_cmd="$version_cmd | jq -r '.[].raw_id'"
+IFS=',' read -r -a CHANNEL_GROUPS_ARRAY <<< "${CHANNEL_GROUPS// /}"
```
btoll: This isn't handling AC2:
```
AC2: If no suitable version is found in nightly, the script correctly moves to the candidate channel and uses a version from there if available.
```

ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
```diff
@@ -159,19 +157,65 @@ if [[ ! -z "$OLD_CLUSTER" ]]; then
...
+  if [[ "${AVAILABLE_UPGRADE}" == "yes" ]]; then
+    versionList=$(eval "${version_cmd}" | jq -r '.[] | select(.available_upgrades!=null) .raw_id')
+  else
+    versionList=$(eval "${version_cmd}" | jq -r '.[].raw_id')
+  fi
+
+  if [[ -n "${versionList}" ]]; then
+    log "Found versions in ${GROUP} channel"
+
+    temp_version=""
```
btoll: This doesn't need to be declared and defined.  It can be removed.

ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
```diff
@@ -159,19 +157,65 @@ if [[ ! -z "$OLD_CLUSTER" ]]; then
...
+    if [[ -z "$OPENSHIFT_VERSION" ]]; then
+      if [[ "$EC_BUILD" == "true" ]]; then
+        temp_version=$(echo "$versionList" | grep -i ec | head -1 || true)
```
btoll: All the logical ORs in these blocks are unnecessary.

ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
```diff
@@ -159,19 +157,65 @@ if [[ ! -z "$OLD_CLUSTER" ]]; then
...
+for GROUP in "${CHANNEL_GROUPS_ARRAY[@]}"; do
+  log "Checking ${GROUP} channel..."
+
+  version_cmd="rosa list versions --channel-group ${GROUP} -o json"
+  if [[ "${HOSTED_CP}" == "true" ]]; then
+    version_cmd="${version_cmd} --hosted-cp"
+  fi
+
+  if [[ "${AVAILABLE_UPGRADE}" == "yes" ]]; then
+    versionList=$(eval "${version_cmd}" | jq -r '.[] | select(.available_upgrades!=null) .raw_id')
```
btoll: All the `versionList` vars should match the same format as the others, i.e., `version_list`.

ci-operator/step-registry/rosa/cluster/provision/rosa-cluster-provision-commands.sh
Amarthya-v: You're right about removing ORs true in most cases, but I need to keep them specifically in the channel selection logic.
The Issue is when searching for a version across multiple channels, grep will fail (exit code 1) if the version doesn't exist in that channel. With set -o errexit enabled, this causes the script to exit immediately instead of trying the next channel.

PR #67315, opened 2025-07-21
AGENT-1260: Add rhel-9-base-nodejs-openshift-4.19 base image

rwsu:
Make the rhel-9-base-nodejs-openshift-4.19 base image available for assisted-installer-ui image builds. It is the standard product base image for nodejs. It contains the git binary and doesn't require another step to install it using microdnf.


rwsu:
Has been replaced by https://github.com/openshift/release/pull/67480.

PR #67321, opened 2025-07-22
Add support for ppc64le in junit2reportportal

liangxia:
Currently the step `junit2reportportal` is only enabled on aws jobs, we will apply it to the other platform as well. Add support for ppc64le to get prepared for that.


liangxia:
/pj-rehearse skip

The ppc64le jobs haven't enabled the step to send test results to ReportPortal, it can not be rehearsed, so skip.

PR #67326, opened 2025-07-22
remove lint jobs for ovnk in 4.18+

jluhrsen:
it turns out our lint job was a NOOP anyway, so we can remove them now that a new change [0] has come in that would cause this to actually fail these jobs.

[0] https://github.com/ovn-kubernetes/ovn-kubernetes/commit/414e29623


jluhrsen:
/close
this is not what the team wants to do. they would rather fix the lint jobs to properly run

PR #67349, opened 2025-07-22
telco5g: Remove old releases from CI

sshnaidm:


ffromani:
/lgtm

let's save some CI resources

PR #67361, opened 2025-07-22
NO-ISSUE: Set always_run: false for unused MCO jobs

djoshy:
When https://github.com/openshift/release/pull/67244 and https://github.com/openshift/release/pull/67253 landed, a number of new jobs were added to the MCO presubmits. This change was done to prioritize the config fields in  `ci-operator/config/openshift` over the fields in `ci-operator/jobs/openshift`. This caused any jobs with an unspecified always_run value in `config/` to be set to true in `jobs/`.

This change adds an `always_run:false` for these jobs in the `ci-operator/config/openshift` & `ci-operator/config/openshift-priv` folders, restoring the MCO presubmits to old behavior.


djoshy:
Note to reviewer: I left `e2e-aws-serial` enabled on 4.19+, since we do have some tests in the serial suite and I thought it was useful signal. I left the files for <4.12 untouched as we don't support those versions anymore.

pablintino (approved):
/lgtm

PR #67412, opened 2025-07-23
Fix 4.18 promotion for sriov projects

SchSeba:


SchSeba:
Hi @jupierce can you please take a look on this PR?

PR #67429, opened 2025-07-24
CORENET-6207: fix ovnk lint job

jluhrsen:
the job had been silently aborting for possibly a very long time because the src image that runs it does not have docker or podman. However the exit was code 0 and that did not fail the job. recently, u/s ovnk modified it's Makefil to exit 1 in this case which now causes the prow lint job to fail.

this adds some steps to allow 'make lint' to actually run the golangci-lint command on the PR code.


jluhrsen:
/pj-rehearse pull-ci-openshift-ovn-kubernetes-release-4.21-lint

jluhrsen:
/pj-rehearse pull-ci-openshift-ovn-kubernetes-release-4.21-lint

jluhrsen:
/pj-rehearse pull-ci-openshift-ovn-kubernetes-release-4.21-lint

jluhrsen:
/pj-rehearse pull-ci-openshift-ovn-kubernetes-release-4.21-lint

jluhrsen:
/pj-rehearse pull-ci-openshift-ovn-kubernetes-release-4.21-lint

jluhrsen:
/pj-rehearse pull-ci-openshift-ovn-kubernetes-release-4.21-lint

jluhrsen:
/pj-rehearse pull-ci-openshift-ovn-kubernetes-release-4.21-lint

jluhrsen:
/pj-rehearse pull-ci-openshift-ovn-kubernetes-release-4.21-lint

jluhrsen:
/pj-rehearse pull-ci-openshift-ovn-kubernetes-release-4.21-lint

jluhrsen:
/pj-rehearse pull-ci-openshift-ovn-kubernetes-release-4.21-lint

jluhrsen:
/pj-rehearse pull-ci-openshift-ovn-kubernetes-release-4.21-lint


/pj-rehearse pull-ci-openshift-ovn-kubernetes-release-4.20-lint

ci-operator/step-registry/ovn/lint/ovn-lint-commands.sh
```diff
@@ -0,0 +1,46 @@
...
+set -o pipefail
+set -x
+
+# there is no container runtime in the src image so we need to run golangci-lint natively
+# we can still pull the same golangci-lint VERSION and run command from whatever is currently
+# in go-controller/hack/lint.sh and run exactly what would be run using 'make lint'
+# below is (pulling VERSION, grabbing that specific golangci-lint, parsing the run command from
+# the docker cmdline, and executing it)
+cd go-controller
+VERSION=$(grep '^VERSION=' hack/lint.sh | cut -d= -f2)
```
kyrtapz, line 14: Instead of adding a downstream specific alternative to running the linter without using container why can't we add a similar logic upstream?
We are already doing that sort of things for unit tests and a bunch of other things:
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/414e29623852c97be3743d29a5022db59947ff7c/go-controller/Makefile#L62-L67

With this approach there is a chance we are going to break this logic by introducing some innocent change to lint.sh.

kyrtapz:
@jluhrsen have you considered using the linter image we have available in CI and specifying the version that matches what ovn-k requires in a current release?
This seems like the easy way out but would require us to update the version to match upstream once in a while.

ci-operator/step-registry/ovn/lint/ovn-lint-commands.sh
jluhrsen: our lint.sh script bails if there is no container runtime which is not there in CI. maybe there is a way to get it, but when I first started looking at this I landed on this approach, and now it's sat for so long I can't remember why I ended up
here.

let me see if I can make upstream work without a container then.

PR #67438, opened 2025-07-24
oadp: change filename to indicate it is no-interop

tareqalayan:


tareqalayan:
@Sau1506mya this file name change is to distingueh between interop files and qe files

Sau1506mya:
@tareqalayan , Could you please also update the config for OCP4.19 with the old workflow  (used in -lp-interop for AWS-IPI testing). Asking you for the future z-stream testing.
Thanks!

tareqalayan:
> @tareqalayan , Could you please also update the config for OCP4.19 with the old workflow (used in -lp-interop for AWS-IPI testing). Asking you for the future z-stream testing. Thanks!

@Sau1506mya Yes. i will do it a different PR!

PR #67454, opened 2025-07-24
Reduce the frequency of the  cgroups v1 bare metal jobs to weekly

harche:
This PR reduces the cgroups v1 jobs on baremetal to weekly.


harche:
Since this only changes the frequency. 

/pj-rehearse ack

PR #67480, opened 2025-07-24
AGENT-1254: Create agent-installer-ui image

rwsu:
ABI OVE currently uses the image named assisted-install-ui. It will be transitioned to the productized name in this patch.

The image is used to run the assisted-disconnected-ui in a container on the ABI OVE live ISO.
github-repository:
https://github.com/openshift-assisted/assisted-installer-ui/tree/master/apps/assisted-disconnected-ui

To transition the ABI OVE UI service, an image using new name is being added in this patch. After the newly named imaged is published, the service will be updated to use the new image name. Finally the old image named assisted-install-ui will be removed.

The base image to create the frontend code is also being changed to rhel-9-base-nodejs-openshift-4.19 from nodejs-18-minimal. The base image is added here first. Then the Dockerfile will be updated to the use the base image. Finally nodejs-18-minimal will be removed in a future patch.

### DIFF
--- a/ci-operator/config/openshift-assisted/assisted-installer-ui/openshift-assisted-assisted-installer-ui-master.yaml
+++ b/ci-operator/config/openshift-assisted/assisted-installer-ui/openshift-assisted-assisted-installer-ui-master.yaml
@@ -3,6 +3,10 @@ base_images:
     name: nodejs-18-minimal
     namespace: edge-infrastructure
     tag: latest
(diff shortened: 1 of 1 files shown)


rwsu:
/cc @mrunalp @jupierce @sdodson 
Hi OpenShift mirroring approvers, can you review the image name? It follows the existing agent installer image names: 
- agent-installer-api-server
- agent-installer-csr-approver
- agent-installer-orchestrator
- agent-installer-node-agent
- agent-installer-utils

PR #67484, opened 2025-07-25
LOG-7422: use index image built by konflux in logging 6.2 and 6.3 jobs

QiaolingTang:
Update some logging 6.2 and 6.3 jobs to use step `openshift-logging-install-konflux-catalogsource`.

/cc @anpingli @kabirbhartiRH


anpingli:
Can you update the job cron to 999?  I think no job  need to be run periodically  when use  openshift-logging-install-konflux-catalogsource.

liangxia (approved):

PR #67526, opened 2025-07-25
rehearse service-status again

deads2k:


deads2k:
sweet, it runs and fails now.

PR #67561, opened 2025-07-28
ARO-HCP: remove trigger plugin

jfchevrette:
We do not use this plugin

Example: https://github.com/Azure/ARO-HCP/pull/2269#issuecomment-3124816786


jfchevrette:
Is there a way we can avoid/disable the bot's ok-to-test and do-not-merge/work-in-progress labeling and comments (ex: https://github.com/Azure/ARO-HCP/pull/2247#issuecomment-3108781883) while keeping the trigger plugin which appears to be needed due to having jobs configured?

PR #67571, opened 2025-07-28
feat: setting up prow job for llm-inference-service

bartoszmajsak:
For upcoming llm-inference-service component in KServe.


andresllh (approved):
looks good, I'll let a second reviewer leave the lgtm

PR #67586, opened 2025-07-28
add premerge job to check demo scripts

deads2k:


deads2k:
tagging for merge so we can test in https://github.com/Azure/ARO-HCP/pull/2274

PR #67587, opened 2025-07-28
Re-introduce dropped Installer pre-submit jobs

sadasu:


sadasu:
/pj-rehearse pull-ci-openshift-installer-release-4.19-shellcheck pull-ci-openshift-installer-release-4.19-yaml-lint

sadasu:
/hold

yaml-lint is currently failing in the installer repo.

ci-operator/jobs/openshift/installer/openshift-installer-main-presubmits.yaml
```diff
@@ -8112,3 +8166,86 @@ presubmits:
         secret:
           secretName: result-aggregator
     trigger: (?m)^/test( | .* )verify-vendor,?($|\s.*)
+  - agent: kubernetes
+    always_run: false
+    branches:
+    - ^main$
+    - ^main-
+    cluster: build01
+    context: ci/prow/yaml-lint
```
patrickdillon, line 8180: hm so the test is complaining because there are duplicate contexts

ci-operator/jobs/openshift/installer/openshift-installer-main-presubmits.yaml
```diff
@@ -8112,3 +8166,86 @@ presubmits:
...
+        secret:
+          secretName: result-aggregator
+    trigger: (?m)^/test( | .* )yaml-lint,?($|\s.*)
+  - agent: kubernetes
+    always_run: false
+    branches:
+    - ^main$
+    - ^main-
+    cluster: build01
+    context: ci/prow/yaml-lint
```
patrickdillon: here's the second one.

let's see if we can figure out why they're duplicates...

patrickdillon:
@sadasu when I pull this pr locally, remove all the jobs and generate them from the configs with `make update` the main presubmit jobs seem to be generated correctly, without duplicate contexts. Can you try that?

sadasu:
/pj-rehearse pull-ci-openshift-installer-release-4.20-shellcheck

sadasu:
/pj-rehearse ci/rehearse/openshift/installer/release-4.18/shellcheck

sadasu:
/pj-rehearse pull-ci-openshift-installer-release-4.19-shellcheck

sadasu:
/pj-rehearse pull-ci-openshift-installer-release-4.20-yaml-lint

patrickdillon:
To use the images, I believe we also need to mirror them to the quay ci proxy: https://docs.ci.openshift.org/docs/how-tos/external-images/

patrickdillon:
Hm, actually, it looks like the jobs may already be back up and running again: here's yaml lint on 4.19 for example: https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-openshift-installer-release-4.19-yaml-lint

patrickdillon:
/reopen

Oh, actually, I think the images have only been fixed on the release branches, so we still do need to re-add the jobs on the main branch

patrickdillon:
So perhaps you can just copy the shellcheck and yamllint jobs from release branch to the main branch. I'm not sure if we can use these images through a config, but it seems to work fine just creating the job directly.

sadasu:
/pj-rehearse pull-ci-openshift-installer-release-4.18-yaml-lint

ci-operator/config/openshift/installer/openshift-installer-main.yaml
```diff
@@ -87,6 +87,10 @@ base_images:
     name: builder
     namespace: ocp
     tag: rhel-9-golang-1.24-openshift-4.20
+  shellcheck:
+    name: shellcheck
+    namespace: ci-public
+    tag: shellcheck-alpine-v0.5.0
```
patrickdillon: Not sure if this was clear in my earlier https://github.com/openshift/release/pull/67587#issuecomment-3255415980 but I don't think you can specify the image needed for these jobs in the config.

I would try copying the pattern for the release branches: don't add config for the jobs, just add the jobs directly. Here's the example from the master branch:

https://github.com/openshift/release/blob/15fdd62a094c29d4ec9b35b6786d79556a04a0ea/ci-operator/jobs/openshift/installer/openshift-installer-master-presubmits.yaml#L3-L31

We would need to adjust that accordingly for the main branch

sadasu: I thought we don't directly update the files in jobs/. So, I hadn't thought of that as an alternative. Wouldn't a future update to the config files, overwrite those "manual" updates to the jobs files?

patrickdillon: Definitely better to use the config, but I've never had any issues editing jobs. Looks like you figured out the config, so that's great!

tthvo (commented):
LGTM!

Rehearsals passed :D Not sure if there is anything else we need to do here...

PR #67608, opened 2025-07-29
OCPBUGS-59907: Skip egressIP openstack test for 4.14-e2e-openstack-dualstack-techpreview jobs

MiguelCarpio:
Skip egressIP openstack test for 4.14-e2e-openstack-dualstack-techpreview jobs

### DIFF
--- a/ci-operator/config/shiftstack/ci/shiftstack-ci-release-4.14.yaml
+++ b/ci-operator/config/shiftstack/ci/shiftstack-ci-release-4.14.yaml
@@ -156,6 +156,9 @@ tests:
     cluster_profile: openstack-vh-bm-rhos
     env:
       FEATURE_SET: TechPreviewNoUpgrade
+      OPENSTACK_TEST_SKIPS: An egressIP attached to a floating IP should be kept after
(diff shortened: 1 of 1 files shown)


mandre:
/hold

What is the rational for skipping the test? Have we looked at why it is failing?

mandre:
> > /hold
> > What is the rational for skipping the test? Have we looked at why it is failing?
> 
> Hi @mandre, I propose skipping the egressIP openstack test until fixed https://issues.redhat.com/browse/OCPBUGS-59907. This is a permanent failure test, for a while, and could be hiding other bugs that we can't see if we consider that `periodic-ci-shiftstack-ci-release-4.14-e2e-openstack-dualstack-techpreview` job will always fail. The last work on solving this was taken by @MaysaMacedo here [openshift/openstack-test#200](https://github.com/openshift/openstack-test/pull/200), but it is already closed. I'll work on a solution, but I prefer to skip the test while I run that challenge.

Well, it seems to me like we want to re-open https://github.com/openshift/openstack-test/pull/200 and get it merged if that fixes our issue, instead of silencing the failure.

PR #67619, opened 2025-07-29
Bump ACM version to 2.15

dhaiducek:
Seems like we'll at least have 2.15 releases to work with, so bumping before I forget about it. :smile:


dhaiducek:
/pj-rehearse ack

The test failed, but the installation ran successfully, which is more what I was looking for.

PR #67624, opened 2025-07-29
CORS-4027: Updating feature gate name to GCPClusterHostedDNSInstall

sadasu:
GCP Cluster Hosted DNS feature has renamed its associated feature gate to GCPClusterHostedDNSInstall. Updating existing pre-submit and periodic jobs to use this new feature gate name.

PR #67649, opened 2025-07-30
refine gcp private and custom-dns presubmit jobs

jianlinliu:
1. Add `run_if_changed: (gcp|google)` on gcp custom-dns jobs, to help us catch regression early, e.g: https://github.com/openshift/release/pull/67624#issuecomment-3140150659
2. Overwrite `test` to drop e2e testing from installer pre-submit gcp custom-dns jobs to save cloud cost.
3. Update `gcp-private` job together to drop the useless steps.

### DIFF
--- a/ci-operator/config/openshift-priv/installer/openshift-priv-installer-main.yaml
+++ b/ci-operator/config/openshift-priv/installer/openshift-priv-installer-main.yaml
@@ -588,14 +588,17 @@ tests:
     workflow: openshift-e2e-gcp
   timeout: 6h0m0s
 - always_run: false
-  as: e2e-gcp-custom-dns
+  as: gcp-custom-dns
   optional: true
+  run_if_changed: (gcp|google)
   steps:
     cluster_profile: gcp-qe
     env:
       FEATURE_GATES: '["GCPClusterHostedDNSInstall=true"]'
       FEATURE_SET: CustomNoUpgrade
       USER_PROVISIONED_DNS: "yes"
+    test:
+    - chain: cucushift-installer-check-cluster-health
     workflow: openshift-e2e-gcp-custom-dns
   timeout: 6h0m0s
 - as: e2e-gcp-secureboot
@@ -1305,7 +1308,13 @@ tests:
     post:
     - chain: cucushift-installer-rehearse-gcp-ipi-private-deprovision
     pre:
-    - chain: cucushift-installer-rehearse-gcp-ipi-private-provision
+    - ref: gcp-provision-vpc
+    - ref: ignition-bastionhost
+    - ref: gcp-provision-bastionhost
+    - ref: proxy-config-generate
+    - chain: ipi-conf-gcp
+    - chain: ipi-install
+    - ref: cucushift-installer-check-gcp-private
     test:
     - chain: cucushift-installer-check-cluster-health
   timeout: 6h0m0s
--- a/ci-operator/config/openshift-priv/installer/openshift-priv-installer-release-4.19.yaml
+++ b/ci-operator/config/openshift-priv/installer/openshift-priv-installer-release-4.19.yaml
@@ -572,14 +572,17 @@ tests:
     workflow: openshift-e2e-gcp
   timeout: 6h0m0s
 - always_run: false
-  as: e2e-gcp-custom-dns
+  as: gcp-custom-dns
   optional: true
+  run_if_changed: (gcp|google)
   steps:
     cluster_profile: gcp-qe
     env:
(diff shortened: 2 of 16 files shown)


jianlinliu:
/pj-rehearse periodic-ci-openshift-release-master-nightly-4.20-e2e-gcp-custom-dns

jianlinliu:
/pj-rehearse pull-ci-openshift-installer-release-4.19-gcp-private pull-ci-openshift-installer-release-4.20-gcp-custom-dns

jianlinliu:
/pj-rehearse pull-ci-openshift-installer-release-4.20-gcp-custom-dns

ci-operator/config/openshift-priv/installer/openshift-priv-installer-release-4.19.yaml
```diff
@@ -1255,7 +1258,13 @@ tests:
     post:
     - chain: cucushift-installer-rehearse-gcp-ipi-private-deprovision
     pre:
-    - chain: cucushift-installer-rehearse-gcp-ipi-private-provision
+    - ref: gcp-provision-vpc
```
gpei, line 1261: @jianlinliu we don't want to keep the `gcp-provision-minimal-permission` step here right?

jianlinliu: I did not see the job enabled mini permission on the var setting, so I removed it. So shall we add it back ?

jianlinliu: Or when we decide to enable mini permission, we can add it back on that time.

gpei: ok to me

gpei: Ah I just realized that it's a bit different for GCP to run the minimal-permission test, it requires the cluster profile to have a pre-configured SA with minimal-permission granted, so we don't have any minimal-permission test involved for GCP in the pre-submit.

ci-operator/config/openshift-priv/installer/openshift-priv-installer-main.yaml
```diff
@@ -588,14 +588,17 @@ tests:
...
   optional: true
+  run_if_changed: (gcp|google)
   steps:
     cluster_profile: gcp-qe
     env:
       FEATURE_GATES: '["GCPClusterHostedDNSInstall=true"]'
       FEATURE_SET: CustomNoUpgrade
       USER_PROVISIONED_DNS: "yes"
+    test:
+    - chain: cucushift-installer-check-cluster-health
```
patrickdillon, line 601: So this will stop e2e tests from running on the custom DNS job? 

Are we doing this because the e2e tests are permafailing? I am ok with that, but not sure if there are other options? 

@sadasu & I were discussing this today and there were roughly ~10 tests that failed on both of the runs for the custom DNS job. I believe @gpei mentioned this before, and that it would not make sense to add ignores for each individual test.

jianlinliu: > because the e2e tests are permafailing

This is one of the reasons. Some others:
1. We set a public domain which resides in QE project, so it depends on QE's cluster profile. While the QE cloud project have a small number of budget. So skip the e2e testing to save the running time for saving the cloud cost.
2. The job is for installer presubmits jobs, most of time, we are focus on installation results, so I drop the e2e to avoid the noise from unexpected e2e failure.

gpei: The most ideal solution in practice would be to fix the tests in the origin repo, either by making them pass or skipping them, so that we can see the most accurate and reasonable e2e test results.
But I believe that having each team update their test code to adapt all test cases to every new cluster type we introduced would be a big workload for them, and potentially with a very lengthy timeline.
During our initial testing of custom DNS, I informed the Ingress team that certain tests were consistently failing. They submitted PR https://github.com/openshift/origin/pull/29985 to fix these issues, but the PR is still under review to this day.

sadasu: @jianlinliu and @gpei thanks for that explanation. I agree that installer pre-submit jobs should can be limited to verifying a successful install. 

We will be able to catch other errors during the periodic e2e custom-dns jobs. I am concerned about scenarios like openshift/origin#29985  and how that will affect the health of our periodic job within CR.

PR #67662, opened 2025-07-30
User newer golang base image for configmap-scale tests

jsafrane:
go 1.15 image is based on RHEL 7 and it cannot run recent `oc` binary.

This should fix permafailing pull-ci-openshift-kubernetes-master-configmap-scale job that [fails](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_kubernetes/2380/pull-ci-openshift-kubernetes-master-configmap-scale/1950462855719424000) with:
```
++ oc get route -n openshift-monitoring prometheus-k8s -o 'jsonpath={.spec.host}'
oc: /lib64/libc.so.6: version `GLIBC_2.33' not found (required by oc)
oc: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by oc)
oc: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by oc)
```

### DIFF
--- a/ci-operator/step-registry/openshift-perfscale/configmap-scale/openshift-perfscale-configmap-scale-ref.yaml
+++ b/ci-operator/step-registry/openshift-perfscale/configmap-scale/openshift-perfscale-configmap-scale-ref.yaml
@@ -3,7 +3,7 @@ ref:
   from_image:
     namespace: openshift
     name: release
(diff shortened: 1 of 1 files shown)


jsafrane:
This PR is not enough to fix the job
/close

PR #67690, opened 2025-07-31
Apply junit2reportportal to agent install workflows

liangxia:
So that we can send test results to ReportPortal directly from Prow.


jianlinliu:
cc @mhanss to review

PR #67699, opened 2025-07-31
fix: set always_run to false

tisutisu:
this value when set as `true`, causing it to run in e2e-tests PRs always, setting it to `false`


jinqi7 (approved):
/lgtm

PR #67710, opened 2025-07-31
rhdh-test-instance-job-update

subhashkhileri:
https://issues.redhat.com/browse/RHIDP-8098

### DIFF
--- a/ci-operator/jobs/redhat-developer/rhdh-test-instance/redhat-developer-rhdh-test-instance-main-presubmits.yaml
+++ b/ci-operator/jobs/redhat-developer/rhdh-test-instance/redhat-developer-rhdh-test-instance-main-presubmits.yaml
@@ -71,4 +71,4 @@ presubmits:
       - name: result-aggregator
         secret:
           secretName: result-aggregator
-    trigger: (?m)^/test( | .* )(deploy|remaining-required),?($|\s.*)
+    trigger: (?m)^/test( | .* )deploy,?($|\s.*)


ci-operator/step-registry/redhat-developer/rhdh-test-instance/redhat-developer-rhdh-test-instance-commands.sh
```diff
@@ -130,7 +134,7 @@ if [[ -n "$comment_body" && "$comment_body" != "null" ]]; then
         echo "Parsed arguments: $install_type $rhdh_version"
         echo "Time duration: $time"
         
-        source ./install.sh "$install_type" "$rhdh_version"
```
zdrapela, line 133: Hi, in this PR: https://github.com/redhat-developer/rhdh-test-instance/pull/10/files#diff-043df5bdbf6639d7a77e1d44c5226fd7371e5259a1e4df3a0dd5d64c30dca44f 
I see that you are still using the `./install.sh`. 
Do you use `./deploy.sh` here by accident, or not?

subhashkhileri: Not by accident, I wanted to rename it to deploy as it makes more sense for this script.

zdrapela: Okay, you got my approval then

PR #67721, opened 2025-07-31
rename user-provisioned-dns to custom-dns

jianlinliu:


jianlinliu:
cc @jianli-wei @gpei @yunjiang29 to review

PR #67757, opened 2025-08-01
Read hypervisor_nic_interface only on hybrid deployments

mcornea:


ci-operator/step-registry/openshift-qe/installer/bm/deploy/openshift-qe-installer-bm-deploy-commands.sh
```diff
@@ -61,6 +59,9 @@ if [[ $PUBLIC_VLAN == "false" ]]; then
 fi
 
 if [[ ! -z "$NUM_HYBRID_WORKER_NODES" ]]; then
+  HV_NIC_INTERFACE=$(cat "${CLUSTER_PROFILE_DIR}/hypervisor_nic_interface")
+  export HV_NIC_INTERFACE
+
 cat <<EOF >>/tmp/all.yml
```
josecastillolema: Ident both `cat` pls, only the first line of each

mcornea: Updated

PR #67770, opened 2025-08-03
oadp-qe: use team profile aws-oadp-qe

tareqalayan:


Sau1506mya:
@tareqalayan , There might be some issue with `cluster_profile` and `BASE_DOMAIN` that you can check later for the fips job for OADP 1.5 on OCP 4.19.
The PR looks good to me for the change of the cluster profile as it is NON-INTEROP job.

PR #67782, opened 2025-08-04
oadp: explicitly define the TEST_FOCUS variable

tareqalayan:
We must define the var in order to be able to control what cases to run. Also we must define it becuase in some runs there are 2 ginkgo sessions started one for kubevirt and one for e2e.
Exposing this file make it more clear what the workflow will be executing.


dfrazzette:
Rehearsal failure on lp-interop may have been due to use of smaller AWS cluster node (where recent production tests passed). Will observe weekly interop job runs post-merge.

dfrazzette (approved):

PR #39618, opened 2023-05-23
4.14 repos to rhel9.2

joepvd:

PR #67789, opened 2025-08-04
Move 4.13 repos to RHEL 9.2 out of EoL 9.0.

igsilya:
4.13 release images are based on RHEL 9.2 since https://github.com/openshift-eng/ocp-build-data/pull/2965.

But for some reason CI images are still using RHEL 9.0, which is EoL and is not supported.  While 4.13 itself is EoL, we still need it to have functional CI, as backporting patches into 4.13 is required in order to backport them further to 4.12.  And this is currently broken for ovn-kubernetes repository as we can't install newer openvswitch packages that are using 9.2 build root as so require newer libraries that are not available in 9.0: https://github.com/openshift/ovn-kubernetes/pull/2697#issuecomment-3136954655

Upgrading repositories to 9.2 to match release images and resolve the package installation problem for ovn-kubernetes.

This PR follows the pattern of the same change in 4.14: https://github.com/openshift/release/pull/39618


sdodson:
/label approved

We'll probably need to reach out to OWNERS but maybe the bot lets me do this?

PR #67806, opened 2025-08-04
NO-JIRA: fix: update tnf lanes for 4.19

eggfoobar:
updated tnf lanes for 4.19 to use DevPreview featureset since TNF will be in DevPreview in 4.19 as we promote the feature.


eggfoobar:
/pj-rehearse ack

This PR resolves the incorrect featureset, the errors are known and are being tracked for TNF

PR #67837, opened 2025-08-05
WIP: CNTRLPLANE-945: cao: move external oidc tests from presubmits to periodics

everettraven:


ci-operator/config/openshift/cluster-authentication-operator/openshift-cluster-authentication-operator-release-4.20__periodics.yaml
```diff
@@ -0,0 +1,142 @@
...
+  steps:
+    cluster_profile: aws-3
+    env:
+      FEATURE_SET: TechPreviewNoUpgrade
+      OPENSHIFT_SKIP_EXTERNAL_TESTS: "True"
+      TEST_SUITE: openshift/auth/external-oidc
+    test:
+    - ref: openshift-e2e-test-8h
+    workflow: openshift-e2e-aws
+  timeout: 8h0m0s
```
everettraven: For now, setting this to the highest possible timeout because they are long running jobs. That being said, it probably makes sense to consider splitting these tests up where possible so we can have much shorter runtimes for the set of tests.

everettraven:
Rehearsals will fail until https://github.com/openshift/origin/pull/30040 is included in a payload

ci-operator/config/openshift/cluster-authentication-operator/openshift-cluster-authentication-operator-release-4.20__periodics.yaml
everettraven: https://github.com/openshift/release/pull/67837/commits/2963c9c225fe42907088eb9d090ae3a56970094f does just this, splitting into logical chunks of tests for the feature, effectively reducing each job runtime to 1/3 of this previous configuration, but introduces 3 times the amount of jobs that run.

I'm not sure which approach is preferred so I'll let others chime in here.

ci-operator/config/openshift/cluster-authentication-operator/openshift-cluster-authentication-operator-release-4.20__periodics.yaml
```diff
@@ -0,0 +1,515 @@
...
+  steps:
+    cluster_profile: gcp
+    env:
+      OPENSHIFT_SKIP_EXTERNAL_TESTS: "True"
+      TEST_SKIPS: (\[OCPFeatureGate:ExternalOIDCWithUIDAndExtraClaimMappings\])|(\[OCPFeatureGate:ExternalOIDC\]
+        reverting to IntegratedOAuth)
+      TEST_SUITE: openshift/auth/external-oidc
+    workflow: openshift-e2e-gcp
+- as: e2e-vsphere-external-oidc-configure
+  interval: 24h
```
xingxingxia: Is it possible to choose some jobs to cover other ARCH like multi-arch or amd64? I intended to cover it to be tracked in https://issues.redhat.com/browse/AUTH-574 . Happened to see you're adding the vsphere platform jobs.

everettraven: We would have to add those in addition to the ones configured here.

I'm certainly open to it, but this PR is specifically scoped to adding the jobs required to satisfy the OCP feature promotion criteria which does not include `arm64` or `multi-arch` jobs.

I'll do some further exploration and if it is straightforward to add arm64 and/or multi-arch jobs I'll add them. Otherwise, I'll suggest that we add those in a follow-up PR so we can continue making progress towards feature promotion.

everettraven:
/pj-rehearse periodic-ci-openshift-cluster-authentication-operator-release-4.20-periodics-e2e-metal-ovn-ipv4-external-oidc-configure-techpreview

everettraven:
Obsolete now. Closing.

/close

PR #67865, opened 2025-08-05
SPLAT-2384: Reduced jobs that will run when vsphere changes detected.

vr4manta:
[SPLAT-2384](https://issues.redhat.com//browse/SPLAT-2384)

### Changes
- Reduced which jobs auto run when vsphere changes are detected.  The jobs will still exist but will be optional to run.

### Notes
The hopes are that in the future, we can make a more dynamic detection of what areas of code cause which jobs to be affected.  For now, we will reduce the special case tests that have periodicals to reduce CI build load to reduce false failures.

### DIFF
--- a/ci-operator/config/openshift/installer/openshift-installer-main.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-main.yaml
@@ -912,7 +912,8 @@ tests:
     cluster_profile: vsphere-elastic
     workflow: openshift-e2e-vsphere
   timeout: 6h0m0s
-- as: e2e-vsphere-externallb-ovn
+- always_run: false
+  as: e2e-vsphere-externallb-ovn
   optional: true
   steps:
     cluster_profile: vsphere-elastic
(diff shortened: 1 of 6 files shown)


ci-operator/config/openshift/installer/openshift-installer-main.yaml
```diff
@@ -1009,9 +1001,9 @@ tests:
     cluster_profile: vsphere-elastic
     workflow: openshift-e2e-vsphere-zones
   timeout: 6h0m0s
-- as: e2e-vsphere-ovn-zones-techpreview
+- always_run: false
+  as: e2e-vsphere-ovn-zones-techpreview
```
jcpowermac, line 1005: this probably shouldn't be named techpreview right?

vr4manta: I think the intent was for it to run for all TechPreview to see if anything fails.  If it not needed, i can remove.  The other jobs do not use multiple FDs.

PR #67866, opened 2025-08-05
add a check to ensure that PRs branch don't contain merge commits

deads2k:


deads2k:
the SHA for the rehearse is from openshift/release, not ARO-HCP.  We'll need to merge this and test on an ARO HCP PR.

We'll do that tomorrow.

PR #67890, opened 2025-08-06
PERFSCALE-4087: Run hammer cli commands inside a container

mcornea:


ci-operator/step-registry/openshift-qe/installer/bm/deploy/openshift-qe-installer-bm-deploy-commands.sh
```diff
@@ -165,7 +165,19 @@ if [[ "$PRE_PXE_LOADER" == "true" ]]; then
   echo "Modifying PXE loaders ..."
   for i in $HOSTS; do
     echo "Modifying PXE loader of server $i ..."
-    hammer -c /root/.hammer/cli.modules.d/foreman_$LAB.yml --verify-ssl false -u $LAB_CLOUD -p $PWD host update --name $i --operatingsystem "$FOREMAN_OS" --pxe-loader "PXELinux BIOS" --build 1
+    podman run \
+      -v /root/.hammer/cli.modules.d:/opt/hammer \
```
josecastillolema: I dont like to depend on files that we assume to be in the bastion host, couldnt we generate dynamically this files on `/tmp/foreman_$LAB.yml` when running the test instead?

josecastillolema: Wdyt?
The whole idea of moving hammer cli to the container is to reduce the dependency to the state of the bastion host.

mcornea: Sure, we can render it when running the test. I'll update it.

mcornea:
/pj-rehearse pull-ci-redhat-performance-jetlag-main-deploy-mno-hybrid

PR #67903, opened 2025-08-06
CORENET-5066: add openshift-pr-manager bot as trusted for ovnk

jluhrsen:
the bot is used to create PRs already, but we need it to be able to trigger actions with /comments to mark PRs ok-to-test and run payload jobs


martinkennelly:
@jluhrsen once this merges, want to validate it?

jluhrsen:
> @jluhrsen once this merges, want to validate it?

yeah, the next time I run the automation I will check that it works. but that won't happen until our current d/s merge makes it in and we need a new one

PR #67940, opened 2025-08-07
add additional checks for Rust code

vparfonov:
This PR adds additional optional checking command for Vector codebase. 

- `cargo clippy --workspace --all-targets -- -D warnings` - clippy can detect patterns that are technically correct but often lead to runtime bugs. This includes things like integer overflow, incorrect use of iterators, and precision loss in floating-point calculations. Finding these before they make it into production can save you a lot of time and trouble.
- `cargo fmt --check` - check is code well formatted 

/assign @jcantrill


vparfonov:
/pj-rehearse pull-ci-ViaQ-vector-v0.47.0-rh-cargo-clippy-check

PR #67941, opened 2025-08-07
Disable build04 and set the presubmit optional

deepsm007:
/cc @openshift/test-platform


deepsm007:
cluster compact and defrag complete, we should be back to normal

PR #67943, opened 2025-08-07
no-jira: Add allowed users to whereabouts_cni

bpickard22:
Add myself, marcelo, and peng to allowed users so we can provide backport-risk-assessed label on whereabouts backports

### DIFF
--- a/core-services/prow/02_config/openshift/whereabouts-cni/_pluginconfig.yaml
+++ b/core-services/prow/02_config/openshift/whereabouts-cni/_pluginconfig.yaml
@@ -21,6 +21,8 @@ label:
       - openshift-staff-engineers
       allowed_users:
       - maiqueb
+      - bpickard22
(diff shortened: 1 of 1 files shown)


core-services/prow/02_config/openshift/whereabouts-cni/_pluginconfig.yaml
```diff
@@ -21,6 +21,9 @@ label:
       - openshift-staff-engineers
       allowed_users:
       - maiqueb
+      - bpickard22
+      - pliu-rh
```
zenghui-shi:
```suggestion
      - pliurh
```

bpickard22: we dont need to add peng here actually since he is a member of the allowed team

PR #67953, opened 2025-08-07
Fix registry url for vsphere02

liangxia:


rvanderp3:
cc @jcpowermac @vr4manta @sgaoshang @WenXinWei

PR #67956, opened 2025-08-07
SPLAT-2398: Set vsphere dualstack-primaryv6 to not always run

vr4manta:
[SPLAT-2398](https://issues.redhat.com//browse/SPLAT-2398)

### Changes
- Made all vsphere dualstack-primaryv6 not auto run since the CI vsphere env does not currently support this

### Notes
Leaving jobs in instead of removing them so that if in the future we enhance CI env to support primaryv6, we can just return to always run.

Jobs haven't passed ever from what I can see.  Example: https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-openshift-cluster-network-operator-master-e2e-vsphere-ovn-dualstack-primaryv6


kyrtapz:
cc: @cybertron @mkowalski

mkowalski:
/lgtm

OK from on-prem networking team. Not super happy about lack of working job but agree there is no point in keeping the perma-failing job. We don't seem to have resources to fix it right now.

PR #68002, opened 2025-08-08
Adding initial list of non-product repositories

bradmwilliams:
This is my initial pass at defining the non-product repositories that will be excluded from the pre-merge verification logic.  This list was generated by scraping the files from the `ocp-build-data` project where `for_payload: false` was set.


bradmwilliams:
/hold
Want a blessing from @jupierce before this merges.

PR #68008, opened 2025-08-11
DPTP-4486: Write attribute cloud_provider when it's valid

liangxia:
When we are unable to determine the cloud_provider, we skip this attribute ( rather than write a attribute `cloud_provider: unknow` as this unknown attribute does not help.


liangxia:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.18-amd64-nightly-oci-agent-vm-compact-fips-f14

PR #68065, opened 2025-08-12
Updates cloud-provider-gcp CI Jobs

theobarberbany:
- Use openshift-hack/verify-vendor.sh for verifying vendoring (this repo is a special case)
- Run unit tests with openshift-hack/test-unit.sh

### DIFF
--- a/ci-operator/config/openshift/cloud-provider-gcp/openshift-cloud-provider-gcp-main.yaml
+++ b/ci-operator/config/openshift/cloud-provider-gcp/openshift-cloud-provider-gcp-main.yaml
@@ -133,11 +133,13 @@ tests:
       PROJECT_NAME: cloud-provider-gcp
     workflow: openshift-ci-security
 - as: verify-deps
-  steps:
-    env:
-      CHECK_MOD_LIST: "false"
-    test:
-    - ref: go-verify-deps
+  commands: openshift-hack/verify-vendor.sh
+  container:
+    from: src
+- as: unit
+  commands: openshift-hack/test-unit.sh
+  container:
+    from: src
 zz_generated_metadata:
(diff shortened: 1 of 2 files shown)


theobarberbany:
`k8s.io/cloud-provider-gcp/pkg/controller/nodeipam/ipam` is flakey by the looks of it - but running units is probably worth it.

ci-operator/config/openshift/cloud-provider-gcp/openshift-cloud-provider-gcp-main.yaml
```diff
@@ -133,11 +133,13 @@ tests:
       PROJECT_NAME: cloud-provider-gcp
     workflow: openshift-ci-security
 - as: verify-deps
-  steps:
-    env:
-      CHECK_MOD_LIST: "false"
-    test:
-    - ref: go-verify-deps
```
mdbooth, line 140: What does `go-verify-deps` do differently?

theobarberbany: Doesnt account for the symlink :) Which means our merges / rebases are always more complex, as we're undoing the symlink from the upstream merge in order to get our deps job to pass. Which often also means we're not building the vendored code in our fork of `cloud-provider-gcp`, so tests fail 😂

theobarberbany: https://github.com/openshift/cloud-provider-gcp/blob/main/openshift-hack/update-vendor.sh#L29-L34 

This approach is lifted from upstream, but with bazel removed.

PR #68094, opened 2025-08-13
app.ci/gangway: add static token for oadp-dev group

tareqalayan:
- Add gangway token assets under clusters/app.ci/gangway-tokens/oadp-dev
- Update clusters/app.ci/prow/03_deployment/gangway.yaml to mount/use the new token
- Enable oadp-dev group access via gangway with appropriate scopes

Generated by: Claude (AI Assistant)

### DIFF
--- a/clusters/app.ci/gangway-tokens/oadp-dev/admin_rbac.yaml
+++ b/clusters/app.ci/gangway-tokens/oadp-dev/admin_rbac.yaml
@@ -1,0 +1,52 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: oadp-dev
+  annotations:
+    openshift.io/description: Service Accounts for oadp-dev
+    openshift.io/display-name: oadp-dev CI
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: periodic-job-bot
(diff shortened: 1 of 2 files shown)


tareqalayan:
> PR sounds about right, if you could have in the comment info how it's going to be used? e.g. do you plan to integrate this to trigger jobs/get list of jobs etc...
> 
> /lgtm

yes. running periodics jobs, extremely improtant becuase it gives a way to modify job environment variables !

mpryc:
> > PR sounds about right, if you could have in the comment info how it's going to be used? e.g. do you plan to integrate this to trigger jobs/get list of jobs etc...
> > /lgtm
> 
> yes. running periodics jobs, extremely improtant becuase it gives a way to modify job environment variables !

So there will be periodic jobs configured with the env variables and each job will have option to be triggered with different variables?
1. Why is this better then defining separate jobs with those variables set as job definitions
2. Isn't this a problem with comparing different job runs? Same job name may be br8ken by different variable being set?

E.g. here is the status stating the job passed, which may be not the case for the job that had different arguments:
https://github.com/openshift/oadp-operator/blob/164bf422802686aa66e3cacf94afb19634081d9d/QE_PROW.md

PR #68111, opened 2025-08-13
Updated supplemental build images secrets and shellcheck destination

hector-vido:


clusters/app.ci/supplemental-ci-images/validation-images/shellcheck/shellcheck.yaml
```diff
@@ -1,38 +1,43 @@
...
-        imageOptimizationPolicy: SkipLayers
-      type: Docker
+  strategy:
+    dockerStrategy:
+      pullSecret:
+        name: registry-push-credentials-ci-images-mirror
+    type: Docker
+---
+apiVersion: batch/v1
+kind: CronJob
```
droslean, line 24: One of the reasons that we introduced the multi-arch build controller is to maintain all images that we build in the clusters with a manifest list and both amd64 and arm64 architectures. See https://github.com/openshift/release/tree/master/clusters/build-clusters/build09/supplemental-ci-images

droslean:
/hold cancel

Let's do a follow-up to migrate them all.

clusters/app.ci/supplemental-ci-images/validation-images/shellcheck/shellcheck.yaml
```diff
@@ -1,38 +1,43 @@
-apiVersion: v1
-kind: List
```
droslean, line 2: also, deleting the list and using the `---` yaml approach its not an ideal change.

PR #68114, opened 2025-08-14
METAL-1525 Increase disk size to avoid DiskPressure

tdomnesc:
For:
baremetalds-e2e-upgrade-ovn-ipv4-workflow
baremetalds-e2e-upgrade-ovn-ipv6-workflow


tdomnesc:
@liangxia - can you please override it again?

PR #68136, opened 2025-08-14
Fixed some scripts to be in compliance with new shellcheck

hector-vido:
Since we updated the shellcheck version, the test start to find other problems.

### DIFF
--- a/ci-operator/step-registry/upi/conf/vsphere/upi-conf-vsphere-commands.sh
+++ b/ci-operator/step-registry/upi/conf/vsphere/upi-conf-vsphere-commands.sh
@@ -322,13 +322,13 @@ cat >"${SHARED_DIR}/variables.ps1" <<-EOF
 \$control_plane_memory = 16384
 \$control_plane_num_cpus = 4
 \$control_plane_count = ${CONTROL_PLANE_REPLICAS}
-\$control_plane_ip_addresses = $(echo ${control_plane_ip_addresses} | tr -d [])
+\$control_plane_ip_addresses = $(echo ${control_plane_ip_addresses} | tr -d '[]')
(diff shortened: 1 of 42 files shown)


ci-operator/step-registry/upi/conf/vsphere/upi-conf-vsphere-commands.sh
```diff
@@ -322,13 +322,13 @@ cat >"${SHARED_DIR}/variables.ps1" <<-EOF
...
 \$control_plane_count = ${CONTROL_PLANE_REPLICAS}
-\$control_plane_ip_addresses = $(echo ${control_plane_ip_addresses} | tr -d [])
+\$control_plane_ip_addresses = $(echo ${control_plane_ip_addresses} | tr -d '[]')
 \$control_plane_hostnames = $(printf "\"%s\"," "${control_plane_hostnames[@]}" | sed 's/,$//')
 
 \$compute_memory = 16384
 \$compute_num_cpus = 4
 \$compute_count = ${COMPUTE_NODE_REPLICAS}
-\$compute_ip_addresses = $(echo ${compute_ip_addresses} | tr -d [])
+\$compute_ip_addresses = $(echo ${compute_ip_addresses} | tr -d '[]')
```
pruan-rht, line 331: @hector-vido @droslean I'm not a bash expert, but I think we can avoid `echo` altogether with something like 
```
compute_ip_addresses=$(tr -d '[]' <<< "$compute_ip_addresses")
```

anpingli:
Thanks

pruan-rht:
@jstuever the error you mentioned above all seems to be related to `oc` rather than `shellcheck`.

PR #68151, opened 2025-08-14
PRE-BRANCHING: Populate generated 4.21 release-gating jobs

hector-vido:
I hope this is the last time we do that manual intervention since @pruan-rht figure out a way to extract the version automatically.

Maybe we'll update the binary, or create a script to patch what was left.

### DIFF
--- a/ci-operator/config/openshift/release/openshift-release-master__ci-4.21-upgrade-from-stable-4.20-from-stable-4.19.yaml
+++ b/ci-operator/config/openshift/release/openshift-release-master__ci-4.21-upgrade-from-stable-4.20-from-stable-4.19.yaml
@@ -1,0 +1,49 @@
+releases:
+  initial:
+    candidate:
+      product: ocp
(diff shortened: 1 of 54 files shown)


ci-operator/config/openshift/release/openshift-release-master__ci-4.21-upgrade-from-stable-4.20-from-stable-4.19.yaml
pruan-rht, line 1: @hector-vido, are we only supporting upgrade chain of 3 hops?  I see there is a 
`openshift-release-master__ci-4.20-upgrade-from-stable-4.19-from-stable-4.18.yaml`, so I'm wondering if there needs to be a `openshift-release-master__ci-4.21-upgrade-from-stable-4.20-upgrade-from-stable-4.19-from-stable-4.18.yaml`... I hope not, but just want to confirm.

PR #68159, opened 2025-08-15
CNTRLPLANE-1257:[OAS]Call OTE in the ci in order to ensure that the tests will be executed always

wangke19:


wangke19:
/hold

Until we get merged: https://github.com/openshift/cluster-openshift-apiserver-operator/pull/623

PR #68187, opened 2025-08-18
DPTP-4486: Add handler for large xml files

liangxia:
When upload data to Datarouter/ReportPortal, it has limits for file sizes, but some of the tests generate large xml files that's larger than the limits. We need some process for those files to make the files smaller to fall into the limits.

### DIFF
--- a/ci-operator/step-registry/junit2reportportal/junit2reportportal-commands.sh
+++ b/ci-operator/step-registry/junit2reportportal/junit2reportportal-commands.sh
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-set -o pipefail
-
 set -x
 if [[ "${NO_REPORTPORTAL,,}" = 'true' ]]
 then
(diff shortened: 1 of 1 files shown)


liangxia:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.17-arm64-nightly-aws-ipi-f14-longduration-workload

ci-operator/step-registry/junit2reportportal/junit2reportportal-commands.sh
```diff
@@ -357,13 +355,25 @@ function fix_xmls() {
...
     fi
+
+    # when process openshift-extended-test-longduration/artifacts/junit/import-Workloads.xml
+    # we got: 413 Request Entity Too Large
+    large_xml_files="$(find "$LOCAL_DIR_RST" -size +10M)" || true
+    if [[ -n "$large_xml_files" ]]
+    then
+      for file in $large_xml_files
+      do
+        grep -B 10 -A 10 -E '<testcase|</testcase>|<failure|</failure>|<system-out|</system-out>' "$file" > "${file}.tmp"
```
pruan-rht, line 373: @liangxia is 10 lines Before and After enough to capture relevant information?

pruan-rht: @liangxia the datarouter limit is 300MB total,  so we could up the current limit check of > 10 MB

liangxia: Actually for this job, the file size is `61046428`, but it failed with `413 Request Entity Too Large`

liangxia: There is a link to the original logs in ReportPortal, so the job owner can always go to Prow to check the full logs for the root cause.

liangxia: BTW, the larger or smaller values can be adjusted later if needed. We need a initial value to get it started.

liangxia:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.17-arm64-nightly-aws-ipi-f14-longduration-workload

PR #68222, opened 2025-08-18
Update validate rover groups script

psalajova:
The current validation script had poor error handling and confusing output when configuration formatting issues occurred.

Depends on: https://github.com/openshift/ci-tools/pull/4651

### DIFF
--- a/hack/validate-rover-groups.sh
+++ b/hack/validate-rover-groups.sh
@@ -7,18 +7,25 @@ set -o pipefail
 RELEASE_REPO=${RELEASE_REPO:-.}
 
 echo >&2 "$(date -u +'%Y-%m-%dT%H:%M:%S%z') Executing sync-rover-groups validation"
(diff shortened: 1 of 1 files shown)


hack/validate-rover-groups.sh
```diff
@@ -7,18 +7,25 @@ set -o pipefail
...
-  echo "ERROR: comments and ordering the keys alphabetically."
+# Run sync-rover-groups and capture output
+if ! out=$(sync-rover-groups --manifest-dir=$RELEASE_REPO/clusters --config-file=$RELEASE_REPO/core-services/sync-rover-groups/_config.yaml --log-level=debug --print-config 2>&1); then
+    echo "ERROR: Config file has syntax errors or invalid fields"
+    echo "$out"
+    exit 1
+fi
+
+# Check if normalized output differs from current file
+if ! echo "$out" | diff - $RELEASE_REPO/core-services/sync-rover-groups/_config.yaml -U10 >/dev/null; then
```
hector-vido, line 19: ~I wonder if this will work if `diff` returns values different than `0`, maybe we need a `|| true`?~
TIL, if I knew this before it would saved me a lot of time s2

PR #68276, opened 2025-08-19
OCPBUGS-60548: remove e2e-rosa-sts-hypershift-ovn job

jbpratt:
this job is permafailing and was never intended to be monitored - the intentions of this job was to run conformance tests against an ROSA HCP cluster but there doesn't exist a suite currently that is suitable for doing it (or requires additional work). Remove these for now!


sdodson:
/retitle OCPBUGS-60548: remove e2e-rosa-sts-hypershift-ovn job

PR #68127, opened 2025-08-14
Add techPreview job with skipped tests for GCP custom-dns

sadasu:

PR #68299, opened 2025-08-20
Update the aws custom-dns pre-submit job as techPreview and also with skipped tests

gpei:
To keep consistent with the GCP changes in https://github.com/openshift/release/pull/68127


gpei:
since there's a bug blocking the AWS custom-dns installation - https://issues.redhat.com//browse/OCPBUGS-57364, so skip the rehearsal test

PR #68344, opened 2025-08-21
Add new job to run openshift-test-private cases

kasturinarra:
Adding new job to run gingko cases on microshift

### DIFF
--- a/ci-operator/config/openshift/microshift/openshift-microshift-release-4.20__periodics.yaml
+++ b/ci-operator/config/openshift/microshift/openshift-microshift-release-4.20__periodics.yaml
@@ -3,6 +3,10 @@ base_images:
     name: brew
     namespace: microshift
     tag: latest
+  tests-private:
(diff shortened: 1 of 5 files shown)


kasturinarra:
> I've reviewed it and learn about this possibility, no idea about it, tbh. I have few questions, though. 
How or where the microshift version is defined? I cannot find it.
[Microshift is being built from source and it is present in one of the commands.sh file, it defines how the payload comes]
step: openshift-microshift-infra-rpm-install-aws, corresponding sh file is in ./ci-operator/step-registry/openshift/microshift/infra/rpm-install-aws/openshift-microshift-infra-rpm-install-aws-commands.sh
Can we install from ostree commit or bootc image with this approcah? 
[I think so, we just need to write new steps to achieve this i believe]
How can we upgrade from previous MicroShift version (Y-1) and then run these tests?
[New steps needs to be written if not existing already or we can utilize existing steps]

PR #68349, opened 2025-08-21
Build latest image for opendatahub-operator in main branch

davidebianchi:


grdryn:
Is this something that we want/need? What is the effect of it?

zdtsw:
> /assign @zdtsw can you PTAL?

we already have image built as "main" on every commit land on main branch
so this change is to another job that doing the same but tag as "latest"
right?

PR #68351, opened 2025-08-21
Introduce rebasebot for the migtools/kopia repository

mpryc:
Initial jobs to introduce rebasebot into oadp repositories.

Full Rebase graph for oadp components:
  https://github.com/oadp-rebasebot/oadp-rebase/blob/oadp-dev/README.md


mpryc:
@weshayutin @sseago FYI

This is slightly different then other rebasebot project, because we have much more repositories and branches and as such there are konfig files that allows to control the upstream branches easily. I've added kopia first, but for the oadp-dev branch we discover the upstream latest branch, where for the oadp-1.5 we have full static control from which branch we are rebasing:
 https://github.com/oadp-rebasebot/oadp-rebase/tree/oadp-dev/rebase-configs

weshayutin (approved):

PR #68209, opened 2025-08-18
cao: move external oidc tests from presubmits to periodics

liouk:
Another attempt of #68029.

PR #68358, opened 2025-08-21
CNTRLPLANE-945: cao: add a metal presubmit for testing OIDC functionality

everettraven:
We have been running into some issues with the metal jobs in https://github.com/openshift/release/pull/68209 

To help us better test different solutions to hostname resolution, this PR adds one of the metal jobs from #68209 as a presubmit job.


everettraven:
Rehearsal should fail fairly quickly, but I want to make sure it fails for the expected reason before merging

everettraven:
Rehearsal failed with expected failures. Removing hold so that this can merge and we can begin testing the fix in https://github.com/openshift/origin/pull/30148

/hold cancel

PR #68372, opened 2025-08-21
vSphere 4.11 periodics to yearly via cron

jcpowermac:
In a previous pr I made changes to these periodics to weekly via
cron. We don't need these jobs to run, only yearly. Using
a cron to define exactly when they will run.


vrutkovs (approved):
/approve

PR #68376, opened 2025-08-21
vSphere - installer repo, reduce variants

jcpowermac:
This commit changes the existing vsphere variants
to always_run false and optional true.


jcpowermac:
/hold

@vr4manta for after recharge, ptal

PR #68377, opened 2025-08-21
add ovnk specific payload plugin config for pr-mgr

jluhrsen:
the openshift-pr-manager[bot] needs to be able to run payload commands. this adds that App to trusted_apps for the payload-testing-prow-plugin, only for
openshift/ovn-kubernetes

### DIFF
--- a/core-services/prow/02_config/openshift/ovn-kubernetes/payload-testing-prow-plugin.yaml
+++ b/core-services/prow/02_config/openshift/ovn-kubernetes/payload-testing-prow-plugin.yaml
@@ -1,0 +1,2 @@
+trusted_apps:
+- openshift-pr-manager


jluhrsen:
> Why do we need to run payload commands?

because we need to vet our d/s merges and branch sync PRs against the payload blocking jobs. Right now the bot can't run those commands so we need a human to first notice the PR and then remember to run the commands.

> How can we confirm its now trusted ?

I don't know without just merging this and checking. I am not really sure this is even right and I have a slack thread to ask.

martinkennelly:
Ack. Understand now especially after reviewing the bots ds merge and payload commands ignored.

jluhrsen:
/revert
this had no effect. the work needs to be done in [ci-tools](https://github.com/openshift/ci-tools/pull/4668) and then [add a config flag in the plugin yaml](https://github.com/openshift/release/pull/68576)

PR #68378, opened 2025-08-21
Consolidate CO periodic jobs on OCP 4.12

rhmdnd:
Instead of running all supported profiles on a single version of
OpenShift, we refactored the ocp4e2e test suite so that it consolidates
all rules of the same type in to one tailored profile for testing.

This makes it easier to run a single job periodically to know if the
rules are stable for a given version, regardless of the profile they're
included in. There is still a need for profile testing, but it's
reserved for testing PRs directly that make changes to the profile.

The approach we're using today requires one job for each profile per
OpenShift version per architecture, which is closing in on 100 different
permutations. This approach let's us DRY up the testing by consolidating
rules to a few TailoredProfiles.

This PR needs to land before the rehearsals will work:
https://github.com/ComplianceAsCode/ocp4e2e/pull/54


ci-operator/config/ComplianceAsCode/content/ComplianceAsCode-content-master__4.12.yaml
```diff
@@ -510,96 +509,12 @@ tests:
...
-      BASE_DOMAIN: quay.devcluster.openshift.com
-      FIPS_ENABLED: "true"
-    test:
-    - as: test
-      cli: latest
-      commands: |
-        export PROFILE=moderate-node
-        export PRODUCT=ocp4
+        set -x
         export component=ocp4-content-ds
```
yuumasato: `component` doesn't seem used anywhere and could be removed.

ci-operator/config/ComplianceAsCode/content/ComplianceAsCode-content-master__4.12.yaml
```diff
@@ -510,96 +509,12 @@ tests:
...
         export ROOT_DIR=$PWD
         git clone https://github.com/ComplianceAsCode/ocp4e2e.git ocp4e2e
         pushd ocp4e2e
-        go test -v -timeout 120m github.com/ComplianceAsCode/ocp4e2e -profile="$PROFILE" -product="$PRODUCT" -content-image="$CONTENT_IMAGE"
+        make e2e-node
       dependencies:
       - env: CONTENT_IMAGE
         name: ocp4-content-ds
```
yuumasato: We can drop this this dependency on `CONTENT_IMAGE` in all weekly jobs, since `ocp4e2e` will default to using `quay.io/redhat-user-workloads/ocp-isc-tenant/compliance-operator-content-dev:master`.

rhmdnd:
Issue with the scanner?

```
 2025/09/05 06:55:19 ComplianceSuite node-scan-binding is not DONE: there was an unexpected error in the scan 'rhcos-node-master': --- Starting Evaluation ---
Title
	Prefer to use a 64-bit Operating System when supported
Rule
	xccdf_org.ssgproject.content_rule_prefer_64bit_os
Result
	pass
```

/pj-rehearse periodic-ci-ComplianceAsCode-content-master-4.12-e2e-aws-openshift-node-compliance-weekly

PR #68380, opened 2025-08-22
gcp dedicated dns zone: unset static base domain

patrickdillon:
Unsets the env var which always points to the regional qe public base domain. Instead, we should use the base domain from the cluster profile.

The presubmits running against the feature PR are failing to find the public base domain and I think--but am not certain--this may be the fix. See [example failure run](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_installer/9792/pull-ci-openshift-installer-main-e2e-gcp-xpn-dedicated-dns-project/1958674590913794048): 

> level=error msg=failed to fetch Master Machines: failed to load asset "Install Config": failed to create install config: baseDomain: Not found: "Public DNS Zone (XXXXXXXXXXXXXXXXXXXXXX/regional.XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX)"


tthvo:
/pj-rehearse ack

Nothing to rehearse :D cc @barbacbd

PR #68272, opened 2025-08-19
BuildConfigs: output all to `QCI` instead of `registry.ci.openshift.org`

Prucek:
Issue: [DPTP-3877](https://issues.redhat.com/browse/DPTP-3877)

We have been using `QCI` for a long time now; still, a lot of things rely on our internal registry. With this change, we aim to remove unnecessary `ImageStreams` from the `app.ci`'s registry and push images built by `BuildConfigs` directly to `QCI`.

/hold

PR #68404, opened 2025-08-25
Enable Nested Containers

danilo-gemoli:
As detailed in https://issues.redhat.com/browse/DPTP-4471, this PR creates the environment to enable `ci-operator` to run test payloads that need to use podman within a container.

This works on `build02` only as it's the only cluster being on `TechPreview`.
What follows is a list of the critical changes that we are going to introduce:
1. ~Use `crun` in place of `runc` for these class of pods: `ci-scheduler-runtime-builds`, `ci-scheduler-runtime-tests`, `ci-scheduler-runtime-longtests`, `ci-scheduler-runtime-prowjobs`. This is perhaps the most critical one even though I don't expect any major disruption.~ Due to technical issues, `crun` has been introduced with https://github.com/openshift/release/pull/68423
2. Bind a privileged security context to the `ci-operator` SA.
3. Build and push an image that contains what's needed to run `podman`.

/cc @openshift/test-platform
/cc @deads2k @haircommander 

/label tide/merge-method-squash

### DIFF
--- a/clusters/app.ci/supplemental-ci-images/nested-podman.yaml
+++ b/clusters/app.ci/supplemental-ci-images/nested-podman.yaml
@@ -1,0 +1,107 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: nested-podman-data
+  namespace: ci
+data:
+  entrypoint.sh: |
+    #!/usr/bin/env bash
+
+    # Create Home directory
+    if [ ! -d "${HOME}" ]
+    then
+      mkdir -p "${HOME}"
+    fi
+
+    # Create Podman configuration
+    if [ ! -d "${HOME}/.config/containers" ]; then
+      mkdir -p ${HOME}/.config/containers
+      (echo 'unqualified-search-registries = [';echo '  "registry.access.redhat.com",';echo '  "registry.redhat.io",';echo '  "docker.io"'; echo ']'; echo 'short-name-mode = "permissive"') > ${HOME}/.config/containers/registries.conf
+      if [ -c "/dev/fuse" ] && [ -f "/usr/bin/fuse-overlayfs" ]; then
+        (echo '[storage]';echo 'driver = "overlay"';echo 'graphroot = "/tmp/graphroot"';echo '[storage.options.overlay]';echo 'mount_program = "/usr/bin/fuse-overlayfs"') > ${HOME}/.config/containers/storage.conf
+      else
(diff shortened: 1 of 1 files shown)


clusters/app.ci/supplemental-ci-images/nested-podman.yaml
```diff
@@ -0,0 +1,111 @@
...
+apiVersion: build.openshift.io/v1
+metadata:
+  namespace: ci
+  name: nested-podman
+spec:
+  successfulBuildsHistoryLimit: 1
+  failedBuildsHistoryLimit: 2
+  output:
+    to:
+      kind: ImageStreamTag
```
Prucek: Could you please output to QCI directly?
https://github.com/openshift/release/pull/68272

core-services/image-mirroring/_config.yaml
```diff
@@ -975,6 +975,10 @@ supplementalCIImages:
     image: quay.io/opendatahub/opendatahub-tests:2.19
   ci/opendatahub-tests:2.22:
     image: quay.io/opendatahub/opendatahub-tests:2.22
+  ci/nested-podman:latest:
```
Prucek: Then you could remove this

clusters/app.ci/supplemental-ci-images/nested-podman.yaml
danilo-gemoli: Nice, thanks! I'll try that

haircommander: ~oh wait, we're on 4.19 not 4.20 still right (hence the requirement for TP?) it's in 4.20 not 4.19~
wrong thread

clusters/app.ci/supplemental-ci-images/nested-podman.yaml
```diff
@@ -0,0 +1,113 @@
...
+    # Create subuid/gid entries for the user
+    USER=$(whoami)
+    START_ID=$(( $(id -u)+1 ))
+    END_ID=$(( 65536-${START_ID} ))
+    echo "${USER}:${START_ID}:${END_ID}" > /etc/subuid
+    echo "${USER}:${START_ID}:${END_ID}" > /etc/subgid
+
+    exec "$@"
+---
+kind: ImageStream
```
Prucek: I think this IS is redundant now, right?

danilo-gemoli: Yes I do think so, as we are pushing to QCI without storing anything within the local registry.

PR #68481, opened 2025-08-26
fix all optional-operators-ci-operator-sdk-* workflows

nunnatsa:
Make all the post steps in these workflows to be best-effort, so the test won't failed on post, if the test itself passed.


nunnatsa:
Test failure is because of ghe tedt environment.

/pj-rehearse ack

PR #68496, opened 2025-08-26
CMP-3385: Consolidate CaC/ocp4e2e testing and extend timeout

rhmdnd:
This commit updates the job configs for ComplianceAsCode/ocp4e2e so that
it runs all the platform and node tests, and extends the timeout. This
is so we exercise all the rules instead of just a couple of profiles.

This PR will need to land before this PR will work:

  https://github.com/ComplianceAsCode/ocp4e2e/pull/54

### DIFF
--- a/ci-operator/config/ComplianceAsCode/ocp4e2e/ComplianceAsCode-ocp4e2e-main.yaml
+++ b/ci-operator/config/ComplianceAsCode/ocp4e2e/ComplianceAsCode-ocp4e2e-main.yaml
@@ -19,7 +19,7 @@ resources:
       cpu: 200m
       memory: 400Mi
 tests:
-- as: e2e-aws-ocp4-moderate
+- as: e2e-aws-openshift-platform-compliance
   steps:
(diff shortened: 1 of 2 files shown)


ci-operator/config/ComplianceAsCode/ocp4e2e/ComplianceAsCode-ocp4e2e-main.yaml
```diff
@@ -19,7 +19,7 @@ resources:
       cpu: 200m
       memory: 400Mi
 tests:
-- as: e2e-aws-ocp4-moderate
+- as: e2e-aws-openshift-platform-compliance-weekly
```
yuumasato: Are these tests run weekly?
I thought these tests only run when we make PRs to `ocp4e2e`.

@rhmdnd  Let's keep the profile tests.
Although they won't be the default use case, keeping these tests will ensure they still work when we want to trigger them manually.

yuumasato (approved):
/lgtm

PR #68498, opened 2025-08-26
no-jira: Fix the GCP job for DNS Private Zone in a separate project

barbacbd:
**The install config contains the same data but under a DNS struct, so there is a need to add the DNS struct then the privateZone information.

### DIFF
--- a/ci-operator/step-registry/ipi/conf/gcp/private-dns-zone/ipi-conf-gcp-private-dns-zone-commands.sh
+++ b/ci-operator/step-registry/ipi/conf/gcp/private-dns-zone/ipi-conf-gcp-private-dns-zone-commands.sh
@@ -21,9 +21,10 @@ PATCH="${SHARED_DIR}/install-config-patch.yaml"
 cat > "${PATCH}" << EOF
 platform:
   gcp:
-    privateZone: 
-      zone: ${private_zone_name}
-      projectID: ${PRIVATE_ZONE_PROJECT}
+    dns:
+      privateZone: 
+        name: ${private_zone_name}
+        projectID: ${PRIVATE_ZONE_PROJECT}
 EOF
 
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/ipi/conf/gcp/private-dns-zone/ipi-conf-gcp-private-dns-zone-commands.sh
```diff
@@ -21,9 +21,10 @@ PATCH="${SHARED_DIR}/install-config-patch.yaml"
 cat > "${PATCH}" << EOF
 platform:
   gcp:
-    privateZone: 
-      zone: ${private_zone_name}
-      projectID: ${PRIVATE_ZONE_PROJECT}
+    dns:
+      privateZone: 
+        zone: ${private_zone_name}
```
tthvo: This should be `name`?

tthvo:
From the rehearsal's [log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_release/68498/rehearse-68498-pull-ci-openshift-installer-release-4.20-e2e-gcp-xpn-dedicated-dns-project/1960411547868073984/artifacts/e2e-gcp-xpn-dedicated-dns-project/ipi-install-install/artifacts/.openshift_install-1756241473.log):

> time="2025-08-26T20:02:33Z" level=warning msg="failed to parse first occurrence of unknown field: failed to unmarshal install-config.yaml: error unmarshaling JSON: while decoding JSON: json: unknown field \"dns\""
> time="2025-08-26T20:02:33Z" level=info msg="Attempting to unmarshal while ignoring unknown keys because strict unmarshaling failed"

It is setting the right field. Also, there is no error related to the public DNS zone anymore.

PR #68502, opened 2025-08-26
kueue-operator: check for on-pr and particular revision for bundles

rphillips:


rphillips:
Looks promising... Retesting the 4.19 periodic.

PR #68414, opened 2025-08-25
[PERFSCALE-4126] Test virt workflow on a SNO

josecastillolema:
This PRs also removes the warm-up kb runs :warning:

PR #68523, opened 2025-08-27
Disable warmup

rsevilla87:
Disabling useless warmup in the workloads


josecastillolema:
I just did the same in https://github.com/openshift/release/pull/68414

PR #68538, opened 2025-08-27
Use repository configuration files rather than URLs

rsevilla87:
Use repository config files rather than URLs pointing to files in the main branch


rsevilla87:
Closing in favor of https://github.com/openshift/release/pull/68544

PR #68564, opened 2025-08-27
reduce the churn in udn tests

mohit-sheth:
Reduce the churn back to its normal values as this is resolved https://issues.redhat.com/browse/OCPBUGS-59738


jtaleric:
yes - not sure why it wasn't merged earlier?

PR #68567, opened 2025-08-27
ci: decrease hypershift AKS periodic test frequency from 2h to once per day

bryan-cox:
- Change e2e-aks periodic test from every 2 hours to once per day

Signed-off-by: Bryan Cox <brcox@redhat.com>


bryan-cox:
/pj-rehearse ack

This is just changing the frequency of the test run. No changes in test setup or functionality.

PR #68589, opened 2025-08-28
[DPTP-4519] remove 'op' that has no reference in certificate.yaml

pruan-rht:
The error occurred because the JSON patches in $RELEASE_REPO//clusters/_cluster-install/build02.yaml were trying to remove the aws-project label from Certificate resources that didn't have this label in it https://github.com/openshift/release/blob/master/clusters/build-clusters/build02/cert-manager/certificate.yaml

### DIFF
--- a/clusters/_cluster-install/build02.yaml
+++ b/clusters/_cluster-install/build02.yaml
@@ -24,7 +24,6 @@ onboard:
       - kind: Certificate
         name: apps-tls
       inline: [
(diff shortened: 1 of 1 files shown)


pruan-rht:
@danilo-gemoli PTAL when you have a chance.

danilo-gemoli:
/hold
Removing the patch may not be ideal, let's see if [this](https://github.com/openshift/ci-tools/pull/4673) works.

PR #68602, opened 2025-08-28
do not upload any other junit files to reportportal when install failed

jianlinliu:


jianlinliu:
cc @liangxia to review.

PR #68614, opened 2025-08-28
ETCD-316: turn off target version check for etcd repo

tjungblu:
We want to quiet the target version errors that the jira bot is throwing for us over at openshift/etcd:
https://github.com/openshift/etcd/pull/339#issuecomment-3233513621

Which demands us to target version 4.21.0, which does not even exist in JIRA yet as a target version. As far as I understood, this setting will turn the check off entirely for good - meaning we don't have to create surrogate bug tickets on releases that do not actually exist yet.

### DIFF
--- a/core-services/jira-lifecycle-plugin/config.yaml
+++ b/core-services/jira-lifecycle-plugin/config.yaml
@@ -290,6 +290,40 @@ orgs:
             release_notes_default_text: "Cause: What actions or circumstances cause this bug to present.\r\nConsequence: What happens when the bug presents.\r\nFix: What was done to fix the bug.\r\nResult: Bug doesn’t present anymore."
   openshift:
     repos:
+      etcd:
+        branches:        
+          '*':
+            validate_by_default: true
+          release-4.21:
+            validate_by_default: true
+            skip_target_version_check: true
+          release-4.22:
+            validate_by_default: true
+            skip_target_version_check: true
+          release-4.23:
+            validate_by_default: true
+            skip_target_version_check: true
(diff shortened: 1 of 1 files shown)


core-services/jira-lifecycle-plugin/config.yaml
```diff
@@ -290,6 +290,11 @@ orgs:
             release_notes_default_text: "Cause: What actions or circumstances cause this bug to present.\r\nConsequence: What happens when the bug presents.\r\nFix: What was done to fix the bug.\r\nResult: Bug doesn’t present anymore."
   openshift:
     repos:
+      etcd:
+        branches:
+          '*':
```
sdodson, line 295: Can you change this to release-4.20? And then when we branch change it to release-4.21?

As it is now it's going to forego that logic on all branches and we wouldn't want that.

sdodson (changes requested):

core-services/jira-lifecycle-plugin/config.yaml
tjungblu: how about adding it for all future branches, and we drop off the last released version after GA?

sdodson: sure that works

tjungblu: thanks, done. I hope by 4.30 we finally merge into main like any other team...

@jupierce is there an easy way to chop a release off after branching? I assume there is some automation for the default branches, in that case we could also do it as Scott intended

PR #68653, opened 2025-08-29
Add doc about adding new regions to e2e test suite

roivaz:
A small document that provides guidance for the task of adding new regions to the prod e2e test suite.


mmazur:
@pruan-rht can you lift the hold?

PR #68262, opened 2025-08-19
USHIFT-5982: Trigger MicroShift release tests

agullon:
Add 4 new CI testing jobs to run when microshift repo `releases` scenarios ([ostree](https://github.com/openshift/microshift/tree/main/test/scenarios/releases) and [bootc](https://github.com/openshift/microshift/tree/main/test/scenarios-bootc/releases)) are modified:
- e2e-aws-tests-release
- e2e-aws-tests-arm-release
- e2e-aws-tests-bootc-release
- e2e-aws-tests-arm-bootc-release

Add 4 new periodics jobs to run every Sunday, Tuesday and Thursday:
 - e2e-aws-tests-release-periodic
 - e2e-aws-tests-arm-release-periodic
 - e2e-aws-tests-bootc-release-periodic
 - e2e-aws-tests-arm-bootc-release-periodic

These new 8 CI jobs has been only added to `release-4.20` branch because the purpose is to run Release Testing and releases won't be created from `main` branch.

PR #68665, opened 2025-08-29
NO-ISSUE: Add microshift presubmit releases dir

agullon:
This change was added yesterday in #68262 PR, but reverted by the bot because it was added in `release-4.20` file instead of `main` file.


pacevedom (approved):
/lgtm

PR #68741, opened 2025-09-02
CNTRLPLANE-946:Add Prow jobs for aks hypershift external oidc e2e tests

xiuwang:
Job [4.21-periodics-e2e-azure-aks-external-oidc ](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/68741/rehearse-68741-periodic-ci-openshift-hypershift-release-4.21-periodics-e2e-azure-aks-external-oidc/1991508687327334400) passed
Job [4.21-periodics-e2e-azure-aks-oidc-techpreview](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/68741/rehearse-68741-periodic-ci-openshift-hypershift-release-4.21-periodics-e2e-azure-aks-oidc-techpreview/1986035717419044864) passed
Job [4.20-periodics-e2e-azure-aks-external-oidc](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/68741/rehearse-68741-periodic-ci-openshift-hypershift-release-4.20-periodics-e2e-azure-aks-external-oidc/1991858979214462976) passed
**Changes:**
 - Implement step registry entries for deploying and configuring Keycloak server in AKS cluster
 - Add DNS configuration step for Keycloak integration
 - Update Azure e2e test runner to support OIDC testing scenarios
 - Add new periodic jobs for HyperShift 4.20 and 4.21 releases to test Azure AKS with external OIDC
 - Create a new `hypershift-azure-aks-oidc-workflow` that includes Keycloak setup
**Components:**
 - New step registry chain: `idp-external-oidc-keycloak-aks` with server deployment and DNS configuration steps
 - Enable keycloak provider as external oidc in Azure Aks


xiuwang:
/pj-rehearse periodic-ci-openshift-hypershift-release-4.20-periodics-e2e-azure-aks-oidc

xiuwang:
/pj-rehearse periodic-ci-openshift-hypershift-release-4.20-periodics-e2e-azure-aks-oidc

xiuwang:
/pj-rehearse periodic-ci-openshift-hypershift-release-4.20-periodics-e2e-azure-aks-oidc

xiuwang:
/pj-rehearse periodic-ci-openshift-hypershift-release-4.20-periodics-e2e-azure-aks-oidc

xiuwang:
/pj-rehearse periodic-ci-openshift-hypershift-release-4.20-periodics-e2e-azure-aks-oidc

ci-operator/config/openshift/hypershift/openshift-hypershift-release-4.20__periodics.yaml
```diff
@@ -225,6 +225,16 @@ tests:
   steps:
     cluster_profile: hypershift
     workflow: hypershift-azure-aks-conformance
+- as: e2e-azure-aks-oidc-techpreview
```
xingxingxia: All job names include "external-oidc" so that we can consitently filter with "external-oidc"? I ask this, because I know there are "-oidc-" appearing in other jobs / other steps et al but not related to authentication external oidc.

ci-operator/config/openshift/hypershift/openshift-hypershift-release-4.21__periodics.yaml
```diff
@@ -233,6 +233,25 @@ tests:
   steps:
     cluster_profile: hypershift
     workflow: hypershift-azure-aks-conformance
+- as: e2e-azure-aks-oidc
+  cron: 0 9 1,10,20 * *
```
xingxingxia: Is the frequency too less? Compared with others of the same YAML file.

ci-operator/config/openshift/hypershift/openshift-hypershift-release-4.21__periodics.yaml
```diff
@@ -233,6 +233,25 @@ tests:
...
+- as: e2e-azure-aks-oidc-techpreview
+  cron: 0 20 9,25 * *
+  steps:
+    cluster_profile: hypershift
+    env:
+      AUTH_THROUGH_CERTS: "true"
+      CI_TESTS_RUN: TestExternalOIDC
+      ENABLE_HYPERSHIFT_CERT_ROTATION_SCALE: "true"
+      TECH_PREVIEW_NO_UPGRADE: "true"
+    workflow: hypershift-azure-aks-oidc
```
xingxingxia: Same as https://github.com/openshift/release/pull/68741#discussion_r2512688595 , suggest (but not enforce) workflow name also includes "external-oidc"?

ci-operator/step-registry/idp/external-oidc/keycloak/aks/server/idp-external-oidc-keycloak-aks-server-ref.yaml
```diff
@@ -0,0 +1,22 @@
+ref:
+  as: idp-external-oidc-keycloak-aks-server
+  from_image:
+    namespace: ocp
+    name: "4.19"
+    tag: upi-installer
+  commands: idp-external-oidc-keycloak-aks-server-commands.sh
+  env:
+  - name: HYPERSHIFT_BASE_DOMAIN
+    default: "ci.azure.devcluster.openshift.com"
```
xingxingxia, line 10: ci-operator/step-registry/hypershift/azure/aks/oidc/hypershift-azure-aks-oidc-workflow.yaml also defines `HYPERSHIFT_BASE_DOMAIN` with same `documentation` words, but the default is different. Can they be consistent? Otherwise it reads indeed confusing at first glance. If their defaults indeed have to be different, how about updating the "documentation" clearer so that readers can read clearly at first glance?

ci-operator/step-registry/idp/external-oidc/keycloak/aks/set-dns/idp-external-oidc-keycloak-aks-set-dns-commands.sh
```diff
@@ -0,0 +1,70 @@
...
+az login --service-principal -u "${AZURE_AUTH_CLIENT_ID}" -p "${AZURE_AUTH_CLIENT_SECRET}" --tenant "${AZURE_AUTH_TENANT_ID}" --output none
+
+if [ ! -f ${SHARED_DIR}/keycloak-prefix ] ; then
+    echo "The keycloak-prefix file must be provided by a previous step!"
+    exit 1
+fi
+export KUBECONFIG="${SHARED_DIR}/kubeconfig"
+KEYCLOAK_PREFIX="$(cat ${SHARED_DIR}/keycloak-prefix)"
+KEYCLOAK_HOST="${KEYCLOAK_PREFIX}.${HYPERSHIFT_BASE_DOMAIN}"
+DNS_ZONE_RG_NAME="os4-common"
```
xingxingxia: Given ci-operator/step-registry/idp/external-oidc/keycloak/aks/set-dns/idp-external-oidc-keycloak-aks-set-dns-ref.yaml already sets the default of `DNS_ZONE_RG_NAME`, line 23 here uses hard code instead of trying to check the `DNS_ZONE_RG_NAME` set there. Is this expected?

xingxingxia: And, how about dynamically using the RG of the cluster instead of the team's preserved RG? Using the preserved RG requires us add one ref step to remove the dns record.

ci-operator/step-registry/idp/external-oidc/keycloak/aks/server/idp-external-oidc-keycloak-aks-server-commands.sh
```diff
@@ -0,0 +1,395 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+export KUBECONFIG="${SHARED_DIR}/kubeconfig"
```
xingxingxia: This seems by-design automatically propagated, as https://docs.ci.openshift.org/docs/architecture/step-registry/#a-note-on-kubeconfig . So this is not needed?

ci-operator/step-registry/idp/external-oidc/keycloak/aks/server/idp-external-oidc-keycloak-aks-server-commands.sh
```diff
@@ -0,0 +1,395 @@
...
+# Insall helm binary
+wget https://get.helm.sh/helm-v3.17.2-linux-amd64.tar.gz -O /tmp/helm.tar.gz
+tar -xvf /tmp/helm.tar.gz -C /tmp/
+chmod +x /tmp/linux-amd64/helm
+
+KEYCLOAK_ADMIN_TEST_USER="admin-$(< /dev/urandom tr -dc 'a-z0-9' | fold -w 6 | head -n 1 || true)"
+KEYCLOAK_ADMIN_TEST_PASSWORD="$(< /dev/urandom tr -dc 'a-z0-9' | fold -w 12 | head -n 1 || true)"
+KEYCLOAK_PREFIX="keycloak-$(< /dev/urandom tr -dc 'a-z0-9' | fold -w 9 | head -n 1 || true)"
+KEYCLOAK_HOST="$KEYCLOAK_PREFIX.$HYPERSHIFT_BASE_DOMAIN"
+CLUSTER_CONSOLE="$(oc whoami --show-console)"
```
xingxingxia, line 21: To avoid others and future updating guys being confused, how about adding a `#` comment to clarify something, like This is just a placeholder; If one hosted cluster to be created later will use this keycloak server as external oidc provider, this CLUSTER_CONSOLE is not right?

ci-operator/step-registry/idp/external-oidc/keycloak/aks/server/idp-external-oidc-keycloak-aks-server-commands.sh
```diff
@@ -0,0 +1,395 @@
...
+cat > /tmp/keycloak-values.yaml << EOF
+nameOverride: ""
+fullnameOverride: ""
+namespace: keycloak
+
+auth:
+  adminUser: ${KEYCLOAK_ADMIN_TEST_USER}
+  adminPassword: "${KEYCLOAK_ADMIN_TEST_PASSWORD}"
+image:
+  registry: docker.io
```
xingxingxia, line 94: docker.io has limit rate on anonymous pulls. Pulls might occasionally fail, right? If so, is there a way to resolve it?

ci-operator/step-registry/idp/external-oidc/keycloak/aks/server/idp-external-oidc-keycloak-aks-server-commands.sh
```diff
@@ -0,0 +1,395 @@
...
+  repository: bitnamilegacy/keycloak
+  tag: 26.3.3-debian-12-r0
```
xingxingxia, line 96: So the Keycloak version to cover is hard coded. Is it possible to let us cover the latest always?

ci-operator/step-registry/idp/external-oidc/keycloak/aks/server/idp-external-oidc-keycloak-aks-server-commands.sh
```diff
@@ -0,0 +1,395 @@
...
+  enabled: true
+  fsGroup: 1001
+  runAsUser: 1001
+  runAsNonRoot: true
+
+envFrom:
+  - configMapRef:
+      name: "keycloak-env-vars"
+  - secretRef:
+      name: "keycloak"
```
xingxingxia: Seems I do not search out `keycloak-env-vars` else where. Do you know where the configmap is used / what it is for? And where is the usage of secretRef name "keycloak"? (I can't search to check where the secret name "keycloak" is used, because Ctrl + F shows 130+ matches)

ci-operator/step-registry/idp/external-oidc/keycloak/aks/server/idp-external-oidc-keycloak-aks-server-commands.sh
```diff
@@ -0,0 +1,395 @@
...
+          timeout 10m bash -c 'while true; do
+            if curl -s http://localhost:${KC_HTTP_PORT:-8080}/health/ready; then
+              echo "Keycloak server is ready. Starting initialization..."
+              break
+            fi
+            sleep 5
+          done' || {
+            echo "ERROR: Timeout waiting for Keycloak server"
+            exit 1
+          }
```
xingxingxia: If there is failure in these lines, it is hard to debug where the failure is. Because we do not have way to examine the echo outputs of these lines. The `oc logs` way doesn't examine these outputs. The only way is: we have to put all outputs to /tmp/postStart.log and examine postStart.log instead.

ci-operator/step-registry/idp/external-oidc/keycloak/aks/server/idp-external-oidc-keycloak-aks-server-commands.sh
```diff
@@ -0,0 +1,395 @@
...
+        export PATH=\$PATH:/opt/bitnami/keycloak/bin
+        KCADM_CONFIG=/tmp/.keycloak-kcadm.config
+
+        echo "Authenticating with admin user: \${KC_BOOTSTRAP_ADMIN_USERNAME}"
+        kcadm.sh config credentials \
+          --server http://localhost:\${KC_HTTP_PORT:-8080} \
+          --realm master \
+          --user "\${KC_BOOTSTRAP_ADMIN_USERNAME}" \
+          --password "\$(cat "\${KC_BOOTSTRAP_ADMIN_PASSWORD_FILE}")" \
+          --config=\${KCADM_CONFIG}
```
xingxingxia: As far as I know from previous test logs in OCP openshift-tests-private keycloak Prow jobs, one trying might fail, this needs to use retrying execution until success.

ci-operator/config/openshift/hypershift/openshift-hypershift-release-4.20__periodics.yaml
```diff
@@ -225,6 +225,16 @@ tests:
...
+  cron: 0 10 4,14,24 * *
+  steps:
+    cluster_profile: hypershift
+    env:
+      AUTH_THROUGH_CERTS: "true"
+      CI_TESTS_RUN: TestExternalOIDC
+      ENABLE_HYPERSHIFT_CERT_ROTATION_SCALE: "true"
+      TECH_PREVIEW_NO_UPGRADE: "true"
```
xingxingxia: For 4.20, given now Default is backported in, we can remove the TPNU now.

ci-operator/step-registry/idp/external-oidc/keycloak/aks/server/idp-external-oidc-keycloak-aks-server-commands.sh
```diff
@@ -0,0 +1,395 @@
...
+    enabled: true
+  https:
+    enabled: true
+  hostname:
+    hostname: ${KEYCLOAK_HOST}
+    strict: false
+  jvm:
+    memory:
+      initial: 512m
+      max: 1024m
```
xingxingxia: Can there be any inline comment for the high mem resource? When cluster is in high workloads, either cluster components or the keycloak pod may hit OOM issue which will cause  job weird problem.

ci-operator/step-registry/idp/external-oidc/keycloak/aks/server/idp-external-oidc-keycloak-aks-server-commands.sh
```diff
@@ -0,0 +1,395 @@
...
+    repository: bitnamilegacy/postgresql
+    tag: 17.6.0-debian-12-r0
+
+resources:
+  requests:
+    cpu: 500m
+    memory: 1Gi
+  limits:
+    cpu: 1000m
+    memory: 2Gi
```
xingxingxia, line 329: Same here for any inline rational comments? The resource requirment seems high which can easily cause weird isssue for cluster components or keycloak running.

ci-operator/step-registry/idp/external-oidc/keycloak/aks/server/idp-external-oidc-keycloak-aks-server-commands.sh
```diff
@@ -0,0 +1,395 @@
...
+
+# Install Keycloak from helm chart
+/tmp/linux-amd64/helm install keycloak bitnami/keycloak \
+  --namespace keycloak \
+  --create-namespace \
+  --values /tmp/keycloak-values.yaml \
+  --timeout 30m \
+  --wait
+
+oc get pods,svc,ing -n keycloak
```
xingxingxia: If last command times out, will this debugging info still be printed? If no, we need improve to ensure once timing out and failing, we're easy to see the oc get pods status debugging info.

ci-operator/step-registry/idp/external-oidc/keycloak/aks/server/idp-external-oidc-keycloak-aks-server-commands.sh
```diff
@@ -0,0 +1,395 @@
...
+        {
+          "componentName": "console", "componentNamespace": "openshift-console", "clientID": "${CONSOLE_CLIENT_ID}",
+          "clientSecret": {"name": "${CONSOLE_CLIENT_SECRET_NAME}"}
+        }
+      ]
+    }
+  ],
+  "type": "OIDC"
+}
+EOF
```
xingxingxia, line 439: Just to confirm, will the "${SHARED_DIR}"/oidcProviders.json created by this script be used somewhere else?

ci-operator/step-registry/hypershift/azure/run-e2e/hypershift-azure-run-e2e-commands.sh
```diff
@@ -87,6 +87,34 @@ if [[ "${AZURE_MULTI_ARCH:-}" == "true" ]]; then
   AZURE_MULTI_ARCH_PARAMS="--e2e.azure-multi-arch=true"
 fi
 
+OAUTH_EXTERNAL_OIDC_PARAM=""
+if [ -f ${SHARED_DIR}/keycloak-provider ] ; then
```
xingxingxia: Can the file name have a common name without "keycloak" so that others in future can just `cat > "the same file name"` without adding some `if [ -f ${SHARED_DIR}/<another>-provider ] ; then` again in the file here?

ci-operator/step-registry/idp/external-oidc/keycloak/aks/set-dns/idp-external-oidc-keycloak-aks-set-dns-commands.sh
```diff
@@ -0,0 +1,70 @@
...
+# Check keycloak host connection
+sleep 60
+curl -sSI --cacert /tmp/router-ca/tls.crt https://${KEYCLOAK_HOST}/realms/master/.well-known/openid-configuration | grep -Eq 'HTTP/[^ ]+ 200'
+
+if [ $? -ne 0 ]; then
+  echo "ERROR: Failed to access Keycloak OpenID configuration at https://${KEYCLOAK_HOST}"
+  echo "Possible reasons: TLS certificate invalid, Keycloak not running, or URL incorrect"
+  exit 1
+fi
+echo "Success: Keycloak OpenID configuration is accessible"
```
xingxingxia: These lines, as well as the following lines till the `export KEYCLOAK_CA_BUNDLE_FILE=$SHARED_DIR/oidcProviders-ca.crt` line, are better moved to the keycloak server deployment script so that if the keycloak server deployment hits problem the job can fail in an earlier and original step?

ci-operator/step-registry/idp/external-oidc/keycloak/aks/server/idp-external-oidc-keycloak-aks-server-commands.sh
```diff
@@ -0,0 +1,395 @@
...
+oc wait --for=condition=ready pod -l app.kubernetes.io/instance=cert-manager -n cert-manager --timeout=300s
+# Create Let's Encrypt ClusterIssuer
+cat <<EOF | oc apply -f -
+apiVersion: cert-manager.io/v1
+kind: ClusterIssuer
+metadata:
+  name: letsencrypt-prod
+spec:
+  acme:
+    server: https://acme-v02.api.letsencrypt.org/directory
```
xingxingxia: Per my experience when I owned the cert-manager component before (now it transfers to OpenShift App Platform QE subteam), `acme-v02.api.letsencrypt.org/directory` is better used for end users' Product envs. It has rate limit to be called. https://letsencrypt.org/docs/rate-limits/ is about the rate limit. So I guess jobs may flaky fail due to the rate limit. And it says "We recommend running tests against our [staging API](https://letsencrypt.org/docs/staging-environment/) first". OAP QE subteam also use https://github.com/openshift/openshift-tests-private/blob/master/test/extended/oap/certmanager.go#L33 the staging.

xingxingxia: BTW, FYI, the certificate issued by the Product API https://acme-v02.api.letsencrypt.org/directory is a trusted certificate, i.e. it is like the certificate of https://google.com . `curl` on it without `--cacert` can succeed directly. Certificate issued by the staging is not a trusted, it is like our OpenShift's self-signed route's certificate.
Suggest use the staging API.

ci-operator/config/openshift/hypershift/openshift-hypershift-release-4.20__periodics.yaml
xingxingxia: `grep -nr '^  *as: .*-oidc'` under github.com/openshift/release can show many other external oidc irrelevant "-oidc-" definitions. So we'd better distinguish from just glancing at names.

xingxingxia:
@xiuwang , sorry for late clicking "Submit" for the cached Pending comments, due to only finishing review this PR < 90% before today. Thanks again for this PR and the associated great efforts! Now done review of all. Generally no big issue. Just some improvements comments.

ci-operator/step-registry/idp/external-oidc/keycloak/aks/set-dns/idp-external-oidc-keycloak-aks-set-dns-commands.sh
xiuwang: As before we discussed, I split the setup keycloak server into two steps, in case that we could re-use the [idp-external-oidc-keycloak-aks-server](https://github.com/openshift/release/pull/68741/files#diff-c4888a71c36f18f0a4a3dc3c83b65f6eb70af0c50ce4e17da02056b1b38a0400) step in other platforms, but as my test, seem it's hard to reuse idp-external-oidc-keycloak-aks-server in other platforms, such as aws .Since you brought it up, I'd like to combine these two steps together again.

ci-operator/step-registry/idp/external-oidc/keycloak/aks/server/idp-external-oidc-keycloak-aks-server-ref.yaml
xiuwang: I just removed the HYPERSHIFT_BASE_DOMAIN in ypershift-azure-aks-external-oidc workflow, let it use the default value

ci-operator/step-registry/idp/external-oidc/keycloak/aks/server/idp-external-oidc-keycloak-aks-server-commands.sh
xiuwang: "${SHARED_DIR}"/oidcProviders.json is not used in this test, but if we plan to add qe profiles to cover aks hcp with keycloak oidc, it could be used at that time.

xiuwang:
/pj-rehearse periodic-ci-openshift-hypershift-release-4.21-periodics-e2e-azure-aks-external-oidc

xiuwang:
/pj-rehearse periodic-ci-openshift-hypershift-release-4.20-periodics-e2e-azure-aks-external-oidc

xiuwang:
/pj-rehearse periodic-ci-openshift-hypershift-release-4.21-periodics-e2e-azure-aks-external-oidc-techpreview

ci-operator/step-registry/idp/external-oidc/keycloak/aks/server/idp-external-oidc-keycloak-aks-server-ref.yaml
xiuwang: `USE_HYPERSHIFT_AZURE_CREDS` set to true in `hypershift-azure-aks-oidc-workflow`, it used use hypershift azure credentials not qe ci azure credentials, this account don't set `ci.azure.devcluster.openshift.com` in `os4-common` RG. Change to use `hypershift.azure.devcluster.openshift.com` in  `hypershift-azure-aks-oidc-workflow`

xingxingxia: OK

ci-operator/step-registry/idp/external-oidc/keycloak/aks/server/idp-external-oidc-keycloak-aks-server-ref.yaml
```diff
@@ -0,0 +1,32 @@
...
+    documentation: "The cluster's FQDN will be a subdomain of the base domain."
+  - name: DNS_ZONE_RG_NAME
+    default: "os4-common"
+    documentation: "The Azure resource group containing the DNS zone for HYPERSHIFT_BASE_DOMAIN. Defaults to os4-common where ci.azure.devcluster.openshift.com is hosted."
```
xingxingxia, line 14: The `HYPERSHIFT_BASE_DOMAIN` variable name seems to mean "base domain for **hypershift cluster**". But this variable seems only used in idp-external-oidc-keycloak-aks-server and idp-external-oidc-keycloak-aks-cleanup steps, which are not related to  **hypershift cluster** at all? To be precise and to avoid reader's confusion, how about renaming it more correctly in both ref.yamls:
```suggestion
  - name: DNS_BASE_DOMAIN
    default: "ci.azure.devcluster.openshift.com"
    documentation: "A base domain for the keycloak server's dns record"
  - name: DNS_ZONE_RG_NAME
    default: "os4-common"
    documentation: "The Azure resource group containing the DNS zone for DBS_BASE_DOMAIN. Defaults to os4-common where ci.azure.devcluster.openshift.com is hosted."
```

ci-operator/step-registry/idp/external-oidc/keycloak/aks/server/idp-external-oidc-keycloak-aks-server-commands.sh
```diff
@@ -0,0 +1,492 @@
...
+          done' || {
+            echo "ERROR: Timeout waiting for Keycloak server after 10 minutes. Check /tmp/postStart.log for details" >> /tmp/postStart.log
+            exit 1
+          }
+
+          echo "Executing setup-script.sh..." >> /tmp/postStart.log
+          /opt/bitnami/keycloak/setup-scripts/setup-script.sh &>> /tmp/postStart.log || {
+            echo "ERROR: setup-script.sh failed. Check /tmp/postStart.log for details"
+            exit 1
+          }
```
xingxingxia: Like https://github.com/openshift/release/blob/master/ci-operator/step-registry/idp/external-oidc/keycloak/server/idp-external-oidc-keycloak-server-commands.sh#L71 , put all shell scripts into setup-script.sh. We can huddle for why.

ci-operator/step-registry/idp/external-oidc/keycloak/aks/server/idp-external-oidc-keycloak-aks-server-commands.sh
```diff
@@ -0,0 +1,492 @@
...
+  --timeout 30m \
+  --wait || {
+  echo "ERROR: Helm install failed or timed out. Collecting debugging information..."
+  echo "Keycloak pods,services and ingresses:"
+  oc get pods,svc,ing -n keycloak || true
+  echo ""
+  echo "Keycloak pod events:"
+  oc get events -n keycloak --sort-by='.lastTimestamp' || true
+  exit 1
+}
```
xingxingxia: Like https://github.com/openshift/release/blob/master/ci-operator/step-registry/idp/external-oidc/keycloak/server/idp-external-oidc-keycloak-server-commands.sh#L81-L85 , if `postStart` fails, we need print the `postStart` logs using below method:
```suggestion
if ! /tmp/linux-amd64/helm install keycloak bitnami/keycloak \
  --namespace keycloak \
  --create-namespace \
  --values /tmp/keycloak-values.yaml \
  --timeout 30m \
  --wait; then
  echo "Keycloak pods,services and ingresses:"
  oc get pods,svc,ing -n keycloak 
  exit 1
fi

echo "Keycloak setup logs:"
oc rsh -n keycloak dc/keycloak cat /tmp/postStart.log |& tee /tmp/postStart.log
if ! grep -q "Keycloak setup done" /tmp/postStart.log; then
    echo "Keycloak setup not done!"
    exit 1
fi
```

xiuwang:
/pj-rehearse periodic-ci-openshift-hypershift-release-4.20-periodics-e2e-azure-aks-external-oidc

xiuwang:
/pj-rehearse periodic-ci-openshift-hypershift-release-4.20-periodics-e2e-azure-aks-external-oidc

ci-operator/config/openshift/hypershift/openshift-hypershift-release-4.21__periodics.yaml
```diff
@@ -246,6 +246,25 @@ tests:
...
+- as: e2e-azure-aks-external-oidc
+  cron: 0 9 * * *
+  steps:
+    cluster_profile: hypershift
+    env:
+      AUTH_THROUGH_CERTS: "true"
+      CI_TESTS_RUN: TestExternalOIDC
+      ENABLE_HYPERSHIFT_CERT_ROTATION_SCALE: "true"
+    workflow: hypershift-azure-aks-external-oidc
+- as: e2e-azure-aks-external-oidc-techpreview
```
sjenning, line 258: Is there any difference in test assertion between the `Default` and `TechPreviewNoUpgrade` test runs?

xiuwang: Hmm, currently, there is no difference between the Default and TechPreviewNoUpgrade test runs,
`ExternalOIDC` and `ExternalOIDCWithUIDAndExtraClaimMappings` are GA in 4.21 and 4.20. Keeping `external-oidc-techpreview` is in case there are new features future... .
If we deem `external-oidc-techpreview` test too redundant, we can consider removing it.
CC @xingxingxia

PR #68781, opened 2025-09-02
CNTRLPLANE-1370: ci(hypershift): move presubmits and release deps to 4.20

bryan-cox:
- ocp_4.20_base-rhel9.name: 4.21 -> 4.20
- refresh release aliases: add latest-420; update initial-420, latest-418/419; set n1/n2=4.20
- rename e2e jobs: e2e-aks-4-19 -> e2e-aks-4-20, e2e-aws-4-19 -> e2e-aws-4-20
- update job contexts, triggers, and targets to 4-20

### DIFF
--- a/ci-operator/config/openshift/hypershift/openshift-hypershift-release-4.21.yaml
+++ b/ci-operator/config/openshift/hypershift/openshift-hypershift-release-4.21.yaml
@@ -74,42 +74,42 @@ releases:
       product: ocp
       stream: ci
       version: "4.21"
-  initial-419:
+  initial-420:
     candidate:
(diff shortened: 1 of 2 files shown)


ci-operator/config/openshift/hypershift/openshift-hypershift-main.yaml
```diff
@@ -95,27 +95,32 @@ releases:
     candidate:
       product: ocp
       stream: ci
-      version: "4.21"
+      version: "4.17"
```
sjenning: we can remove `latest-417` entirely

PR #68807, opened 2025-09-03
build k8sgpt image

jianzhangbjz:
As title, build the k8sgpt image in the Prow job.

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.21__images.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.21__images.yaml
@@ -144,6 +144,9 @@ images:
       as:
       - src
   to: tests-private-ui
+- dockerfile_path: images/Dockerfile.k8sgpt
+  from: root
+  to: k8sgpt
 promotion:
   to:
   - namespace: ci


jianzhangbjz:
/pj-rehearse pull-ci-openshift-openshift-tests-private-release-4.21-images-images

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.21__images.yaml
```diff
@@ -144,6 +144,9 @@ images:
       as:
       - src
   to: tests-private-ui
+- dockerfile_path: images/Dockerfile.k8sgpt
+  from: root
+  to: k8sgpt
```
liangxia, line 149: What's the relationship between `k8sgpt` and `openshift-tests-private` ?   Adding it here means every PR in `openshift-tests-private` for branch release-4.21 will need to build this image, which is not expected. Maybe put it in a more appropriate repo/folder.

jianzhangbjz: We're going to use the `k8sgpt` to identify the OCP issue, which means it belongs to the QE functional test. So, we put it here.

liangxia: We use the image != we need to build the image every time we have a PR

liangxia: I would suggest put the dockerfile and image build process in a separate repo/folder, we still can use `k8sgpt` to identify OCP issue, and we don't need to build that image every time when we have a PR and/or merge a PR.

jianzhangbjz: Yeah, I understand. However, regardless of the `repo/folder`, this image will be built for every new PR—not just in this QE private repo. Since k8sgpt is used not only by OLM but also by other sub-teams, which `repo/folder` would you recommend? Thanks!

jianzhangbjz: Also, building its image for every new PR will keep the k8sgpt to the latest version, that's as expected.

jianzhangbjz:
Hi @pruan-rht and @mgencur, could you help approve it since Liang is on PTO this week? Thanks!

mgencur:
I think that building this image with every PR is not necessary. Looking at the remote repo https://github.com/k8sgpt-ai/k8sgpt/commits/main/, there is maybe one commit per day. Sometimes no commits for a few days. It is also an external repo, not part openshift-tests-private test suite that would require building code from the PR under test.
In my opinion, the image should be built periodically once a day. And the CI for openshift-tests-private would just pull it from the location where it was built.

jianzhangbjz:
> In my opinion, the image should be built periodically once a day. And the CI for openshift-tests-private would just pull it from the location where it was built.

Yes, I understand. However, since [k8sgpt-ai/k8sgpt](https://github.com/k8sgpt-ai/k8sgpt) is an upstream repo without an officially published image yet, we have to build its image periodically in our Prow CI. Regardless of the repo/folder, this image will be built for every new PR—not just in this QE private repo. If we don’t place it here, which repo/folder would be more appropriate?

mgencur:
By custom job, I was thinking having a periodic job in openshift-tests-private that would build the image using the given Dockerfile and push it to quay.io. Then it could be mirrored to the cluster using https://docs.ci.openshift.org/docs/how-tos/external-images/
But I understand this is a bit complicated even though in the long term it makes sense.
Building with each PR in openshift-tests-private sounds a bit excessive. And if the image needs to be used in other repositories as well it could be sometimes out-of-date because it would be built in openshift-tests-private, not in the other repo on-demand.
Building periodically seems like the best option. But in the short term it could be built with PRs in openshift-tests-private.

mgencur:
> so it won’t be triggered for every PR
If this is possible, it would be nice.

I don't have a strong opinion. As a temporary solution, we can be building the image with PRs. There are 5-10 PRs every day. It might be fine.

PR #67997, opened 2025-08-08
Add access-transparency-service bootstrap configuration

tiwillia:
The included files were generated with `make new-repo`. 

The following options where provided:
```
Enter the organization for the repository: openshift-online
Enter the repository to initialize: access-transparency-service
Enter the development branch for the repository: [default: main]

Now, let's determine how the repository builds output artifacts...
Does the repository build and promote container images?  [default: no] yes
Does the repository promote images as part of the OpenShift release?  [default: no] no
Do any images build on top of the OpenShift base image?  [default: no]
Do any images build on top of the CentOS base image?  [default: no]

Now, let's configure how the repository is compiled...
What version of Go does the repository build with? [default: 1.13] 1.23.0
[OPTIONAL] Enter the Go import path for the repository if it uses a vanity URL (e.g. "k8s.io/my-repo"):
[OPTIONAL] What commands are used to build binaries in the repository? (e.g. "go install ./cmd/...") make install
[OPTIONAL] What commands are used to build test binaries? (e.g. "go install -race ./cmd/..." or "go test -c ./test/...") make install

Now, let's configure test jobs for the repository...

First, we will configure simple test scripts. Test scripts
execute unit or integration style tests by running a command
from your repository inside of a test container. For example,
a unit test may be executed by running "make test-unit" after
checking out the code under test.

Are there any test scripts to configure?  [default: no] yes
What is the name of this test (e.g. "unit")?  unit
Does this test require built binaries?  [default: no]
Does this test require test binaries?  [default: no]
What commands in the repository run the test (e.g. "make test-unit")?  make test
Are there any more test scripts to configure?  [default: no] yes
What is the name of this test (e.g. "unit")?  integration
Does this test require built binaries?  [default: no]
Does this test require test binaries?  [default: no]
What commands in the repository run the test (e.g. "make test-unit")?  make test-integration
Are there any more test scripts to configure?  [default: no] yes
What is the name of this test (e.g. "unit")?  e2e
Does this test require built binaries?  [default: no]
Does this test require test binaries?  [default: no]
What commands in the repository run the test (e.g. "make test-unit")?  make test-e2e
Are there any more test scripts to configure?  [default: no]

Next, we will configure end-to-end tests. An end-to-end test
executes a command from your repository against an ephemeral
OpenShift cluster. The test script will have "cluster:admin"
credentials with which it can execute no other tests will
share the cluster.

Are there any end-to-end test scripts to configure?  [default: no]

Repository configuration options loaded!
```

PR #68808, opened 2025-09-03
remove the access-transparency-service repo from the openshift-priv whitelist

droslean:
/cc @openshift/test-platform @danilo-gemoli 

It seems this repo was introduced in the whitelist in https://github.com/openshift/release/pull/68738


droslean:
@tiwillia FYI. This was introduced in https://github.com/openshift/release/pull/67997.

PR #68853, opened 2025-09-03
Set kubeadmin password for console tests

memodi:
Set kubeadmin password for console tests

### DIFF
--- a/ci-operator/step-registry/openshift-extended/web-tests/openshift-extended-web-tests-ref.yaml
+++ b/ci-operator/step-registry/openshift-extended/web-tests/openshift-extended-web-tests-ref.yaml
@@ -34,7 +34,7 @@ ref:
     default: ""
     documentation: Specific to Network Observability, to specify whether to SKIP NOO Install during the tests
   - name: USE_KUBEADMIN_IDP
-    default: false
+    default: "false"
     documentation: Set to true if kube:admin IDP should be used
   documentation: |-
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/openshift-extended/web-tests/openshift-extended-web-tests-commands.sh
```diff
@@ -8,6 +8,12 @@ pwd && ls -ltr
 cd frontend || exit 0
 pwd && ls -ltr
 
+#set env for kubeadmin
+if [ -f "${SHARED_DIR}/kubeadmin-password" ]; then
+    QE_KUBEADMIN_PASSWORD=$(cat "${SHARED_DIR}/kubeadmin-password")
+    export QE_KUBEADMIN_PASSWORD
```
yapei, line 14: how about just use `KUBEADMIN_PASSWORD`?

memodi: I could change it, but was just trying to be consistent [openshift-extended-test-commands.sh](https://github.com/openshift/release/blob/master/ci-operator/step-registry/openshift-extended/test/openshift-extended-test-commands.sh#L107)

yapei: okay...

PR #68854, opened 2025-09-03
OCPBUGS-60839: Specify an InfraID for hypershift clusters created for conformance tests

csrwng:
If not specified, a random infra-id is generated for the cluster by default, causing its DNS records to be cleaned up while in use.


csrwng:
/pj-rehearse periodic-ci-openshift-hypershift-release-4.20-periodics-e2e-aws-ovn-conformance

PR #68879, opened 2025-09-04
Remove verified label requirement for main branch in bpfman

frobware:
Fixes the configuration conflict where the main branch was both included in the verified label requirement and excluded from it via excludedBranches. This resolves similar issues to those fixed in #68836 for the bpfman-operator.

The main branch should not require manual verification as it's used for development work that doesn't need the same verification requirements as release branches.

## Problem
The main branch has conflicting tide configuration:
- includedBranches section requires `verified` label for main
- excludedBranches section excludes main from requirements

## Solution  
Remove the `verified` label requirement from the main branch's includedBranches configuration, making it consistent with the excludedBranches configuration.

This change allows dependency update PRs and other main branch PRs to merge without manual verification whilst preserving verification requirements for release branches.

## Related
This change follows the same pattern as #68836 which fixed the identical issue for bpfman-operator.

### DIFF
--- a/core-services/prow/02_config/openshift/bpfman-operator/_prowconfig.yaml
+++ b/core-services/prow/02_config/openshift/bpfman-operator/_prowconfig.yaml
@@ -43,7 +43,6 @@ tide:
     labels:
     - approved
     - lgtm
-    - verified
     missingLabels:
     - backports/unvalidated-commits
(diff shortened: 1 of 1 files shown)


frobware:
Related: https://github.com/openshift/release/pull/68881

PR #68885, opened 2025-09-04
Adjust the order of steps

sunzhaohua2:
Adjust the order of several steps.

Sometimes even if all the case are passed, but the cluster is not recovered to normal, this caused [post step failed.](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_machine-api-operator/1394/pull-ci-openshift-machine-api-operator-main-regression-clusterinfra-aws-ipi-mapi/1946215039862050816) 
This is because case `When cluster-wide proxy is configured, Machine API cluster operator should  create machines when configured behind a proxy` caused cluster unstable. 

I tried to fix it, even though I confirmed node ready and co stable [twice](https://github.com/openshift/cluster-api-actuator-pkg/compare/master...sunzhaohua2:proxy1?expand=1), it still doesn't work. 

So now I will change the order of these 3 steps to make sure this case is not the last case. and in step cucushift-e2e we just check machine running, no case checking co or other node status. so no effect to cases. 

I tried the job passed. 
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/68885/rehearse-68885-pull-ci-openshift-machine-api-operator-main-regression-clusterinfra-aws-ipi-mapi/1963629803529572352

@huali9 @miyadav PTAL

### DIFF
--- a/ci-operator/step-registry/openshift/e2e/test/clusterinfra-qe/regression/openshift-e2e-test-clusterinfra-qe-regression-chain.yaml
+++ b/ci-operator/step-registry/openshift/e2e/test/clusterinfra-qe/regression/openshift-e2e-test-clusterinfra-qe-regression-chain.yaml
@@ -40,9 +40,9 @@ chain:
   - chain: cucushift-installer-check-cluster-health
   - ref: idp-htpasswd
   - ref: cucushift-pre
-  - ref: openshift-extended-test
-  - ref: cucushift-e2e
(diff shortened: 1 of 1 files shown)


sunzhaohua2:
/pj-rehearse pull-ci-openshift-machine-api-operator-main-regression-clusterinfra-aws-ipi-mapi

PR #68889, opened 2025-09-04
OCPBUGS-61006: cloud-provider-gcp: update CI jobs

theobarberbany:
Part reverts #68065:

- Changes the use of openshift-hack/verify-vendor.sh, to verifying go work
- Adds back the `verify-deps` job, as that is required to check if the repo can be built by konflux.


theobarberbany:
@sunzhaohua2 this PR needs no QE :)

PR #68894, opened 2025-09-04
Pushing cluster-init mabc to QCI and made the builds cluster agnostic

hector-vido:
A follow up of https://github.com/openshift/ci-tools/pull/4659

This commit removed all `mabc` manifests from specific clusters folders and put them inside `clusters/build-clusters/common_except_app.ci/supplemental-ci-images/`.
The script used by the job `periodic-trigger-mabc-images` was also modified to use the manifests from the new location.

The credential inside `build10`, where the job actually run, was modified to include QCI credentials.

**TODO:** Modify `cluster-init` to avoid creating this builds on all clusters and add the QCI credential on each cluster specific `mabc` push secret.

/cc @danilo-gemoli @droslean


droslean (changes requested):
You need to remove the app.ci registry push. Also, this will break the controller since the credentials are missing.

PR #68914, opened 2025-09-04
Add new module for modular architecture MR

lucferbux:
## Description

We are onboarding a new module for odh-dashboard in rhoai. The outcome of this PR would be onboarding this module to have pre-submit and post-submit builds.


ci-operator/config/opendatahub-io/odh-dashboard/opendatahub-io-odh-dashboard-main.yaml
```diff
@@ -12,10 +12,15 @@ images:
...
   to: odh-dashboard-image
+- context_dir: .
+  dockerfile_path: ./packages/model-registry/Dockerfile.workspace
+  to: odh-mod-arch-model-registry-image
 promotion:
   to:
   - namespace: opendatahub-io
     tag: latest
+  - namespace: opendatahub-io
+    tag: main
```
andrewballantyne, line 23: We use this tag, no?

lucferbux: This is the tag that is used by the openshift ci registry that then is pushed to ours (quay) and retagged to whatever we tell in the `promote` pipeline. So funny thing is that this was something suggested by AI but it makes sense, it prevents race conditions agains `latest`. It's not mandatory but it can help.

andrewballantyne: Alrighty -- I guess the name of the image is different too -- so it won't conflict with our main images ... sorry for the dumb question -- this stuff still confuses me to this day lol

PR #68934, opened 2025-09-05
openshift-logging: Update Loki version used for testing releases

xperimental:
Updates Loki versions used for testing:
- `main` uses the latest release 3.5.4
- all other releases were switched to 3.4.3 which is what we will be using for the upcoming z-streams

Refs [LOG-7660](https://issues.redhat.com//browse/LOG-7660)

/cc @JoaoBraveCoding


JoaoBraveCoding (approved):
/lgtm

PR #68939, opened 2025-09-05
Add benluddy to the openshift/kubernetes labelers list.

benluddy:
I'm an approver in OWNERS, but this is insufficient for reviews that involve managing carries. I promise to be responsible.


bertinatto (commented):
/lgtm

PR #68986, opened 2025-09-08
config: set ephemeral storage requests for rhcos

jbtrystram:
We have been hitting issues where the pod was evicted during the test due to ephemeral storage limit being hit.
Request 40Gi as the tests seem to fail at 18Gi.


HuijingHei (approved):
LGTM

PR #69010, opened 2025-09-08
telco: NROP: skip running CI on k5x related changes

shajmakh:
Prow CI is not configured (at time of posting this) to consume konflux resources (namely images) and this enablement is not yet planned. Keeping prow CI running on all filed PRs creates slows down and sometimes blocks merging konflux-related updates, thus it's pointless to keep running the required jobs.
This commit skips the mandatory jobs if the PR contains only konflux-related updates.
See the following for more details: https://docs.prow.k8s.io/docs/jobs/


shajmakh:
Rehearsals are pointless in this case

PR #69014, opened 2025-09-08
Set paramiko version for ncclient

sgoveas:
Latest paramiko does not have `DSSKey` attribute, but ncclient fails trying to access it. This leads to job failure as the switches fail to configure.


sgoveas:
Overridden by https://github.com/openshift-eng/baremetal-qe-infra/pull/91

PR #69030, opened 2025-09-09
telcov10n: update orion config

abraham2512:
Renames orion test to track nightly cpu-test prow job


abraham2512:
cc: @ccardenosa

PR #69084, opened 2025-09-09
CNTRLPLANE-267: chore(hypershift): introduce an order convention for tests

devguyio:
This PR introduces an order convention for tests to improve test identification, IDE folding visibility and make it easier to quickly understand when and how the job runs without examining the full test configuration.


devguyio:
seems there's an order convention in place already by the ci-operator tools 😁

PR #69085, opened 2025-09-09
CNTRLPLANE-267: ensure hypershift jobs don't get automatically triggered

devguyio:
HyperShift jobs are costly jobs, this is why we want the action of triggering them becomes a deliberate action that takes place according to the judgement of the PR authors on the best timing. 

In some cases in early feedback and review stages of a change, triggering those jobs with every change that is pushed is costly.

### DIFF
--- a/ci-operator/jobs/openshift/hypershift/openshift-hypershift-main-presubmits.yaml
+++ b/ci-operator/jobs/openshift/hypershift/openshift-hypershift-main-presubmits.yaml
@@ -15,7 +15,6 @@ presubmits:
       pj-rehearse.openshift.io/can-be-rehearsed: "true"
     name: pull-ci-openshift-hypershift-main-e2e-aks
     rerun_command: /test e2e-aks
-    skip_if_only_changed: (^(\.tekton|\.github|docs|examples|enhancements|contrib|\.cursor)/)|(^[A-Z]+\.md$)|((^|/)OWNERS$)|(/overrides\.yaml$)|(^renovate\.json$)|(/\.testcoverage\.yml$)|(^\.gitlint$)|(^\.gitignore$)
     spec:
       containers:
       - args:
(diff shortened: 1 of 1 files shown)


ci-operator/config/openshift/hypershift/openshift-hypershift-main.yaml
```diff
@@ -398,7 +398,8 @@ tests:
       REQUEST_SERVING_COMPONENT_TEST: "true"
       TECH_PREVIEW_NO_UPGRADE: "true"
     workflow: hypershift-aws-e2e-external
-- as: verify-deps
+- always_run: false
+  as: verify-deps
```
mgencur, line 402: Maybe `verify-deps` could still be run always? The test just takes one minute and doesn't spin a new cluster.

devguyio: would there be a value behind that? and what's the difference between that and our own make deps?

mgencur: Good question. Our `make deps` should do more or less the same. But we don't run it in our CI?
I thought we wanted to remove time/resource consuming checks. This "verify-deps" seems like ething that could quickly help you fix dependencies before running full E2E.

devguyio: ack, but wouldn't such deps problem get caught during image building?

mgencur: No. I don't think so. From the go-verify-deps-ref.yaml:

>  This steps makes sure that "vendor/" directory of the tested repo is correct.
    It is useful for OCP forks of upstream repost that don't have
    "make verify-vendor" or similar make target and adding it in OCP
    would introduce a <carry> patch.

Anyway, not sure it is very useful for Hypershift.

devguyio: > ack, but wouldn't such deps problem get caught during image building?

I double checked our Dockerfile and indeed the make target used there build the binaries directly.  Ok I see your point now. changing this to always run.

devguyio:
/pj-rehearse skip

Not needed for such a change

PR #69099, opened 2025-09-10
OCPBUGS-61498: cluster-api-provider-vsphere: add units

damdo:
Add units job to the cluster-api-provider-vsphere branches

### DIFF
--- a/ci-operator/config/openshift/cluster-api-provider-vsphere/openshift-cluster-api-provider-vsphere-master.yaml
+++ b/ci-operator/config/openshift/cluster-api-provider-vsphere/openshift-cluster-api-provider-vsphere-master.yaml
@@ -55,6 +55,10 @@ resources:
       cpu: 100m
       memory: 200Mi
 tests:
+- as: unit
+  commands: openshift/unit-tests.sh
+  container:
+    from: src
 - as: e2e-vsphere
   optional: true
   skip_if_only_changed: ^docs/|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|PROJECT|LICENSE)$
--- a/ci-operator/config/openshift/cluster-api-provider-vsphere/openshift-cluster-api-provider-vsphere-release-4.11.yaml
+++ b/ci-operator/config/openshift/cluster-api-provider-vsphere/openshift-cluster-api-provider-vsphere-release-4.11.yaml
@@ -41,6 +41,10 @@ resources:
       cpu: 100m
       memory: 200Mi
 tests:
+- as: unit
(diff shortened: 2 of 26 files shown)


damdo:
/pj-rehearse pull-ci-openshift-cluster-api-provider-vsphere-master-unit

ci-operator/config/openshift/cluster-api-provider-vsphere/openshift-cluster-api-provider-vsphere-master.yaml
```diff
@@ -54,7 +54,12 @@ resources:
     requests:
       cpu: 100m
       memory: 200Mi
+test_binary_build_commands: openshift/unit-tests.sh
```
mdbooth: Why is this required? (And throughout)

damdo: It is present on the other openshift CAPI provider jobs so I added it here too for consistency.
I don't think it is required, it sounds like it provides some build artifact for further steps https://docs.ci.openshift.org/docs/architecture/ci-operator/#building-artifacts

mdbooth: Lets not add it if we don't think we need it. That'll also reduce the likelihood of us cargo-culting it somewhere else later.

mdbooth: It looks like it's creates a new image layer by executing `test_binary_build_commands` and tagging the result as `test-bin`. As unit-tests.sh doesn't produce any binary artifacts and in fact runs the full test suite, I think we just end up running the unit tests an extra time during the build step. I'm pretty sure we don't want that.

damdo: Ok cool. Updated.

damdo:
/pj-rehearse pull-ci-openshift-cluster-api-provider-vsphere-master-unit

mdbooth (approved):
/lgtm

mdbooth:
I'm assuming the fact that the unit tests fail is to be addressed, btw?

damdo:
They are expected to fail until https://github.com/openshift/cluster-api-provider-vsphere/pull/66 merges

damdo:
Hm it looks running the units locally works, but they fail here. I wonder why.

PR #69113, opened 2025-09-10
CNTRLPLANE-1399: Move HyperShift AKS e2e to centralus instead of eastus2

bryan-cox:
This PR:
- Continues defaulting AKS cluster location to eastus2 as before, but removes the LOCATION environment variable from the AKS provision workflow so other jobs can override the location as needed. No behavior change for existing pipelines; enables per-job location overrides.
- Move Hypershift AKS end-to-end job(s) to Azure Central US (centralus) region. Other jobs remain unchanged.

ref: https://redhat-external.slack.com/archives/C075PHEFZKQ/p1757507227903919


bryan-cox:
The test failed due to https://github.com/openshift/hypershift/pull/6782. The concern in this PR was getting AKS back to provisioning which this PR does.

PR #69120, opened 2025-09-10
CMP-3385: Enable platform and node testing on all content PRs

rhmdnd:
Now that we consolidated all rules into two profiles for testing
purposes, it's easier to run just two jobs on each PR to catch any
changes that break RHCOS or OCP rules.

### DIFF
--- a/ci-operator/config/ComplianceAsCode/content/ComplianceAsCode-content-master.yaml
+++ b/ci-operator/config/ComplianceAsCode/content/ComplianceAsCode-content-master.yaml
@@ -30,6 +30,46 @@ resources:
       cpu: 200m
       memory: 400Mi
 tests:
+- always_run: true
+  as: e2e-aws-openshift-platform-compliance
+  steps:
+    cluster_profile: quay-aws
+    env:
+      BASE_DOMAIN: quay.devcluster.openshift.com
+    test:
+    - as: test
+      cli: latest
+      commands: |
+        set -x
+        git clone https://github.com/ComplianceAsCode/ocp4e2e.git ocp4e2e
+        pushd ocp4e2e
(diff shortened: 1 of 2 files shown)


ci-operator/config/ComplianceAsCode/content/ComplianceAsCode-content-master.yaml
```diff
@@ -30,6 +30,50 @@ resources:
...
+  steps:
+    cluster_profile: quay-aws
+    env:
+      BASE_DOMAIN: quay.devcluster.openshift.com
+    test:
+    - as: test
+      cli: latest
+      commands: |
+        set -x
+        export component=ocp4-content-ds
```
yuumasato: It doesn't look like `component` is being used anywhere.
I think you need to pass it through `-content-image` flag, but there is no wiring in the make target to pass the image to be used.

Basically it is not clear what content image is being used for the e2e test, the one from `quay.io/redhat-user-workloads`, or the one just built as `ocp4-content-ds`.

If the purpose is test contents from the PR we need to ensure that `ocp4-content-ds` is being used.

rhmdnd: Good call - ripping out the make target so that we can be more specific in how we pass variables into the test suite for this use case.

Specifically, we want to make sure we're using the right content image and we're also passing in the content directory from the PR.

yuumasato (approved):
I think test are failing because `-content-dir` is being set to this PR's path...
`-content-directory=/go/src/github.com/ComplianceAsCode/content/ocp4e2e`

Let's test this on content PRs and see how they run.
/lgtm
/pj-rehearse ack

PR #69140, opened 2025-09-10
NO-JIRA: fix: increase disk size for tnf

eggfoobar:
during upgrade tnf clusters fase disk pressure, this is due to the default of 60 being too low, bumping to 120 disk size as this is the typical size we recommend for compact clusters


eggfoobar:
/pj-rehearse periodic-ci-openshift-release-master-nightly-4.20-e2e-metal-ovn-two-node-fencing-upgrade

PR #69179, opened 2025-09-11
NO-JIRA: explicit parallelism

neisw:
[origin#30226](https://github.com/openshift/origin/pull/30226) looks to adjust the default parallelism value of 30 based on the number of worker nodes to mitigate CPU overload on smaller clusters.  Microshift is skipping the majority of the high CPU consuming tests and currently runs cleanly at the default setting.  We don't want to increase job run time due to the lower value in this case.


neisw:
/assign @pacevedom 
Some jobs with fewer than the expected 3 worker nodes can struggle with the default parallelism value of 30. Your jobs handle it pretty well since you likely skip the heavier cpu hitting tests.  We don't want to increase your job run times as noted in [#30226](https://github.com/openshift/origin/pull/30226#discussion_r2334297263) so we are setting it explicitly here.

PR #69209, opened 2025-09-12
STOR-2297: remove oVirt storage jobs and references

RomanBednar:


jsafrane:
I think it would be better to remove just 4.21 jobs and keep the old ones there. Just in case we need to build something for the older versions.

jsafrane:
I think they will create `*-4.23.yaml` for any `*-master.yaml` that they find, but better ask the testplatform folks

PR #69216, opened 2025-09-12
OCPBUGS-60009: Skip aws lb controller CredentialsRequest creation in CI pre-install step on main

grzpiotrowski:
AWS Load Balancer Controller credentials request is now created in the setup of the aws-load-balancer-operator e2e tests, eliminating the race condition where the target namespace for the request did not exist yet when the credentials
request was created as part of the CI pre-install steps.

The CredentialsRequest creation in the pre-install step is still used by CI for previous releases of ALBO: 1.0 and 1.1 and the pre-install step remains for these two versions.

Also in a previous change #60039 the operator's CredentialRequest was removed from this pre-install step as it was not needed anymore, since the operator provisions the credentials for itself.

At this stage all the credentials necessary for the operator tests are requested either by the operator itself or in the operator e2e initialization.

Example failure prior to this change:
> ```--- FAIL: TestAWSLoadBalancerControllerWithCredentialsSecret (920.10s) === RUN TestAWSLoadBalancerControllerWithCustomIngressClass operator_test.go:406: Creating test namespace "aws-load-balancer-test-custom-ing" operator_test.go:412: Creating a custom ingress class operator_test.go:422: Creating aws load balancer controller instance with custom ingress class util.go:72: failed to get deployment aws-load-balancer-controller-cluster: deployments.apps "aws-load-balancer-controller-cluster" not found (retrying) util.go:72: failed to get deployment aws-load-balancer-controller-cluster: deployments.apps "aws-load-balancer-controller-cluster" not found (retrying) util.go:72: failed to get deployment aws-load-balancer-controller-cluster: deployments.apps "aws-load-balancer-controller-cluster" not found (retrying) ```

This PR depends on the merge of https://github.com/openshift/aws-load-balancer-operator/pull/201

### DIFF
--- a/ci-operator/step-registry/aws-load-balancer/pre-install/aws-load-balancer-pre-install-commands.sh
+++ b/ci-operator/step-registry/aws-load-balancer/pre-install/aws-load-balancer-pre-install-commands.sh
@@ -18,9 +18,7 @@ cp ${AWS_LOAD_BALANCER_OPERATOR_SRC_DIR}/hack/controller/controller-credentials-
 /tmp/yq w -i "${AWS_CREDENTIALS_REQUEST}" 'spec.secretRef.namespace' ${OO_INSTALL_NAMESPACE}
 /tmp/yq w -i "${AWS_CREDENTIALS_REQUEST}" 'spec.secretRef.name' ${OPERATOR_SECRET}
 
-if [ "${OO_APPLY_RESOURCES}" = "true" ] || [ "${OO_APPLY_RESOURCES}" = "controlleronly" ]; then
-    if [ "${OO_APPLY_RESOURCES}" != "controlleronly" ]; then
-        oc apply -f "${AWS_CREDENTIALS_REQUEST}"
-    fi
+if [ "${OO_APPLY_RESOURCES}" = "true" ]; then
+    oc apply -f "${AWS_CREDENTIALS_REQUEST}"
     oc apply -f "${AWS_CONTROLLER_CREDENTIALS_REQUEST}"
 fi
--- a/ci-operator/config/openshift/aws-load-balancer-operator/openshift-aws-load-balancer-operator-main.yaml
+++ b/ci-operator/config/openshift/aws-load-balancer-operator/openshift-aws-load-balancer-operator-main.yaml
@@ -62,7 +62,7 @@ tests:
   steps:
     cluster_profile: aws
     env:
-      OO_APPLY_RESOURCES: controlleronly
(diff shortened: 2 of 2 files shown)


ci-operator/step-registry/aws-load-balancer/pre-install/aws-load-balancer-pre-install-commands.sh
alebedev87, line 1: Do we still need this step? The operator provisions the credentials for itself, the controller will be handled at the e2e setup. I think that it can be removed entirely.

grzpiotrowski:
/retitle OCPBUGS-60009: Remove aws-load-balancer-pre-install step

ci-operator/step-registry/aws-load-balancer/pre-install/aws-load-balancer-pre-install-commands.sh
grzpiotrowski: Done. Now after removing this pre-install step, to satisfy the config check test, I removed the `OO_INSTALL_NAMESPACE` env vars from the release-1.0 and 1.1 workflows, as the `make ci-operator-checkconfig` recognized this env is not used anymore.
Though I don't think I fully understand the setup for multiple releases now.
Am I messing with the tests of the previous releases, which don't have the credentials set in the operator's e2e?

grzpiotrowski: I reverted the deletion of the pre-install step as it is needed by the 1.0 and 1.1 tests, since we don't backport the change in ALBO repository which handles the CredentialsRequest creation for the controller: https://github.com/openshift/aws-load-balancer-operator/pull/201

Since the [OO_APPLY_RESOURCES ](https://github.com/openshift/release/blob/11dd491e28994525c47cd561f2305a498006fcc5/ci-operator/step-registry/aws-load-balancer/pre-install/aws-load-balancer-pre-install-ref.yaml#L19-L20) var used by the pre-install script is `true` by default, the only tweak I did was to set it to false for the CI using the ALBO's main.

And reverted the script logic to the state before the optional `controlleronly` flow for the credentials request, as now the only two variants would be either create all cred requests or none at all and leave it for the operator and its e2e.

alebedev87:
`TestAWSLoadBalancerControllerWithWAFRegional` should be skipped on `release-1.0` and `release-1.1` branches due to https://issues.redhat.com/browse/OCPBUGS-57320 (the skip is already done on `main`).

/pj-rehearse ack

PR #69222, opened 2025-09-12
Further fixing the multiarch periodic timings

tvardema:
After all the changes for periodic timings, we missed updating many of the s390x jobs back to our regular cadences.

### DIFF
--- a/ci-operator/config/openshift/multiarch/openshift-multiarch-master__nightly-4.16.yaml
+++ b/ci-operator/config/openshift/multiarch/openshift-multiarch-master__nightly-4.16.yaml
@@ -375,7 +375,7 @@ tests:
 - as: ocp-e2e-ovn-remote-libvirt-s390x
   capabilities:
   - sshd-bastion
-  cron: 4 14 18 12 *
+  cron: 0 8 * * 1-4
   steps:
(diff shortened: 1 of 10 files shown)


ci-operator/config/openshift/multiarch/openshift-multiarch-master__nightly-4.16.yaml
```diff
@@ -375,7 +375,7 @@ tests:
 - as: ocp-e2e-ovn-remote-libvirt-s390x
   capabilities:
   - sshd-bastion
-  cron: 4 14 18 12 *
+  cron: 0 8 * * 1,3
```
amrutp-redhat: This should be "0 8 * * 1-4" as per the planning sheet.

amrutp-redhat (commented):
I couldn't find the change for 4.19 nightly single payload and multi payload. Could you please help me.

ci-operator/config/openshift/multiarch/openshift-multiarch-master__nightly-4.16.yaml
tvardema: Good catch!  Updating.  Thanks :D

PR #69230, opened 2025-09-13
make one serial job required for cluster network operator

xueqzhan:
Given the recent payload breakage, we recommend adding one required serial job to cno presubmit


danwinship:
/lgtm
/approve

but you ought to fix 4.20 and 4.19 too; I'll re-lgtm if you update that

danwinship:
/hold cancel
ok, i'll submit another PR for 4.20 and 4.19...

PR #69242, opened 2025-09-15
Update OWNER_ALIASES for sdn-approvers and sdn-reviewers

pliurh:
Remove former team members and add new ones.

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -221,28 +221,32 @@ aliases:
   - jaypoulz
   - cs-zhang
   - dharaneeshvrd
-  sdn-approvers:
+  core-networking-approvers:
   - abhat
   - danwinship
-  - dcbw
-  - dougbtv
-  - jluhrsen
-  - knobunc
-  - trozet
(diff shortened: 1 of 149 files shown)


OWNERS_ALIASES
```diff
@@ -224,25 +224,29 @@ aliases:
   sdn-approvers:
```
tssurya: where do we use sdn-approvers and sdn-reviewers aliases from?

tssurya (approved):
/lgtm

ci-operator/step-registry/baremetalds/e2e/ovn/OWNERS
```diff
@@ -1,4 +1,4 @@
 approvers:
 - sdn-approvers
```
liangxia: Looks like `sdn-approvers` is renamed in `OWNERS_ALIASES`, but still used here ?

pliurh: fixed

liangxia (approved):

PR #69273, opened 2025-09-15
DNM cdv2 job for kube-burner-ai

venkataanil:
kube-burner AI needs CI runs for cdv2 job with different config options. We want cdv2 to create variable number of cluster resources in each test run instead of predefined number of them. Each iteration in the test run, choses a random number between the min and max provided for the resource.

Please don't review or merge this change as it is a temporary change for our CI runs.


ci-operator/step-registry/openshift-qe/cluster-density-v2/openshift-qe-cluster-density-v2-commands.sh
```diff
@@ -52,7 +52,8 @@ fi
 REPO_URL="https://github.com/cloud-bulldozer/e2e-benchmarking";
 LATEST_TAG=$(curl -s "https://api.github.com/repos/cloud-bulldozer/e2e-benchmarking/releases/latest" | jq -r '.tag_name');
 TAG_OPTION="--branch $(if [ "$E2E_VERSION" == "default" ]; then echo "$LATEST_TAG"; else echo "$E2E_VERSION"; fi)";
-git clone $REPO_URL $TAG_OPTION --depth 1
+#git clone $REPO_URL $TAG_OPTION --depth 1
+git clone -b kb_ai https://github.com/venkataanil/e2e-benchmarking
```
vishnuchalla, line 56: Looks like we mixed up things with existing jobs here. We need an isolated job for this work.

PR #69299, opened 2025-09-16
Benchmark-Runner project step-registry

ebattat:
1. Adding Benchmark-Runner step-registry

### DIFF
--- a/ci-operator/step-registry/redhat-performance/benchmark-runner/redhat-performance-benchmark-runner-commands.sh
+++ b/ci-operator/step-registry/redhat-performance/benchmark-runner/redhat-performance-benchmark-runner-commands.sh
@@ -1,0 +1,3 @@
+
+#!/bin/bash
+python3.12 /benchmark_runner/main/main.py
--- a/ci-operator/step-registry/redhat-performance/benchmark-runner/OWNERS
+++ b/ci-operator/step-registry/redhat-performance/benchmark-runner/OWNERS
@@ -1,0 +1,9 @@
+approvers:
+  - ebattat
+  - RobertKrawitz
+  - jeniferh
+  - gqlo
+reviewers:
+  - RobertKrawitz
+  - jeniferh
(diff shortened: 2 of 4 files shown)


ci-operator/step-registry/redhat-performance/benchmark-runner/redhat-performance-benchmark-runner-ref.yaml 
```diff
@@ -0,0 +1,7 @@
+  ref:
```
paigerube14: I think if you run "make update" this will create another ref file ending in *-ref.metadata.json 
think this missing file is part of why your checks are failing

paigerube14: make sure "ref" is all the way to the left

ci-operator/step-registry/redhat-performance/benchmark-runner/redhat-performance-benchmark-runner-commands.sh
```diff
@@ -0,0 +1,7 @@
+ref:
```
paigerube14: copy this content into the ref file. The commands file should have the python run...

ci-operator/step-registry/redhat-performance/benchmark-runner/redhat-performance-benchmark-runner-commands.sh
```diff
@@ -0,0 +1 @@
+ python3.12 /benchmark_runner/main/main.py
```
paigerube14: Think you should add the type of script at the top like :

`#!/bin/bash`
https://github.com/openshift/release/blob/master/ci-operator/step-registry/redhat-chaos/application-outages/console/redhat-chaos-application-outages-console-commands.sh

ebattat:
Open new PR

PR #69316, opened 2025-09-16
Nutanix: Setting up a nutanix capi techpreview e2e job

abhay-nutanix:
Setting up a nutanix capi techpreview e2e job for the cluster-capi-operator


damdo (commented):
Looks good, but no point in merging until we merge the support for nutanix in cluster-capi-operator and add the first E2Es there.

So let's keep this open until then

PR #69321, opened 2025-09-16
Azure 4.21 jobs to use Standard_Das_v5 instances

sdodson:
COMPONENT | OLD |   NEW
--------------------|---------|--------
Control Plane | Standard_D8as_v3 (8vCPU/32GiB) 3rd generation| Standard_D4as_v5 (4vCPU/16GiB) 5th generation
Workers | Standard_D4as_v3 (4vCPU/16GiB) 3rd generation| Standard_D4as_v5 (4vCPU/16GiB) 5th generation

I have reviewed CPU consumption in the rehearsals and found that the control plane resize down to 4 CPUs seems to produce CPU consumption graphs that are inline with AWS, basically we were always over provisioning the control plane in terms of CPU count and almost always we had 4 cores idle. See graphs below for before/after and comparison to aws jobs.

### DIFF
--- a/ci-operator/config/openshift/release/openshift-release-master__ci-4.21-upgrade-from-stable-4.20.yaml
+++ b/ci-operator/config/openshift/release/openshift-release-master__ci-4.21-upgrade-from-stable-4.20.yaml
@@ -91,9 +91,10 @@ tests:
     cluster_profile: azure-2
     env:
       BASE_DOMAIN: ci2.azure.devcluster.openshift.com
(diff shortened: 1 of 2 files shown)


sdodson:
/retest-required

I've reviewed the jobs that passed, those that failed, and no obvious issues. I've also reviewed the CPU consumption and both control plane and workers seem down 10-20%. Also, specifically  on the control plane where we're using 8 CPUs where 4 is the norm on AWS and 6 is the norm on GCP I believe we can likely revert back to 4 CPU. I will need to look into what led us to use 8 CPU instances in the first place for control plane.

PR #69337, opened 2025-09-16
Search e2e use 2.15 integration pipeline for images

smcavey:
For Search e2e test runs use integration snapshot pipeline as no snapshots exist in dev, leading to consistent job failures and runs not happening: https://github.com/stolostron/search-e2e-test/pull/407


jlpadilla (approved):
/lgtm

PR #69344, opened 2025-09-16
OCPBUGS-61792: Collect coredumps on all nodes during CI runs

rfredette:
During the investigation for [OCPBUGS-61224](https://issues.redhat.com/browse/OCPBUGS-61224), it was found that haproxy in the router containers occasionally segfaulted. However, the coredumps from those segfaults weren't saved, so a reason for the segfault couldn't be accurately determined. This change configures all nodes to save all coredumps, and any coredumps that were saved are copied to the artifacts directory at the end of the CI run. The router pod runs in the host namespace, so any crashes should be captured this way.

While this is intended to help debug router issues, it should also help in cases where crashes occur in any privileged pod.

### DIFF
--- a/ci-operator/config/openshift/router/openshift-router-master.yaml
+++ b/ci-operator/config/openshift/router/openshift-router-master.yaml
@@ -61,6 +61,12 @@ tests:
   skip_if_only_changed: ^docs/|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|OWNERS_ALIASES|PROJECT|LICENSE)$
   steps:
     cluster_profile: gcp-3
+    post:
+    - chain: gather-core-dump
+    - chain: ipi-gcp-post
+    pre:
+    - ref: enable-node-coredumps
+    - chain: ipi-gcp-pre
     workflow: openshift-e2e-gcp
 - as: e2e-metal-ipi-ovn-ipv6
   cluster: build05
@@ -72,6 +78,12 @@ tests:
       DEVSCRIPTS_CONFIG: |
         IP_STACK=v6
         NETWORK_TYPE=OVNKubernetes
+    post:
+    - chain: gather-core-dump
+    - chain: baremetalds-ofcir-post
+    pre:
+    - ref: enable-node-coredumps
+    - chain: baremetalds-ofcir-pre
     workflow: baremetalds-e2e
 - as: e2e-metal-ipi-ovn-dualstack
   cluster: build05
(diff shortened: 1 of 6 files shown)


candita:
@rfredette do you need to split this into 2 PRs?  I see an error in ci-operator-config that indicates it doesn't yet recognize the coredump-service you created within this PR:

> time="2025-09-16T21:16:03Z" level=error error="failed to validate configuration openshift/release/openshift-release-master__ci-4.20-upgrade-from-stable-4.19.yaml: Failed resolve MultiStageTestConfiguration: test/e2e-azure-ovn-upgrade: workflow/openshift-upgrade-azure-ovn: invalid step reference: coredump-service"
> time="2025-09-16T21:16:03Z" level=fatal msg="error validating configuration files"
>

ci-operator/step-registry/coredump/coredump-ref.yaml
```diff
@@ -0,0 +1,17 @@
+ref:
+  as: coredump
```
candita: Does this need to be named `coredump-service` to match https://github.com/openshift/release/pull/69344/files#diff-973041f4ccef0bb535e8a0d0b5d6f517a5999b3fb7a770676a5a071a8319f906R93?

rfredette: Yes, the names were mismatched. I think `coredump` or `coredump-service` were a bit vague, so I standardized on `collect-coredumps` instead.

ci-operator/config/openshift/router/openshift-router-master.yaml
alebedev87, line 66: Do we have to specify the post chain from the given workflow? `gather-core-dump` alone adde to `post` is not enough? Same question is for `enable-node-coredumps` pre step.

rfredette: I did initially try just putting `enable-node-coredumps` in `pre` and `gather-core-dump` in `post`, but the tests failed in about 5 minutes with an error about missing some credentials. As I understand it, if you don't specify `pre` or `post` steps, there's a default based on the platform you choose, but specifying anything overrides that default.

alebedev87: Ack. It would be hard for me to access whether additional pre and post chains are correct for every job as I would have go through all of them. So I trust Ryan on this one. LGTM

alebedev87:
/approve

I let Jame give the LGTM.

neisw:
/lgtm

Based upon https://github.com/openshift/release/pull/69344#issuecomment-3457521290 and @jluhrsen being out on pto

PR #69362, opened 2025-09-17
Add secret name length check

psalajova:


hack/gcp-secret-manager/util.py
pruan-rht: @psalajova please consider rewording the error message to 
`raise click.BadParameter(f"Secret name must be less than {SECRET_NAME_MAX_LENGTH} characters.")`

psalajova: Updated in the new commit

PR #69388, opened 2025-09-17
Remove duplicate version orion env 

memodi:
* Remove duplicate version orion env 
* revert previous change to include `.0` since now quotes will added in config: https://github.com/openshift-eng/ocp-qe-perfscale-ci/pull/808


paigerube14 (approved):
/lgtm

PR #69400, opened 2025-09-17
AGENT-1209: Add OVE ISO periodic job for new releases

pawanpinjarkar:


ci-operator/config/openshift/agent-installer-utils/openshift-agent-installer-utils-main.yaml
```diff
@@ -92,6 +92,18 @@ tests:
     cluster_profile: equinix-ocp-metal
     workflow: agent-e2e-compact-ipv4-iso-no-registry
   timeout: 6h0m0s
+- as: e2e-agent-compact-ipv4-iso-no-registry-periodic
```
andfasano, line 95: I'm not sure about the benefits of having a periodic job now that we've already added presubmit ones (for the iso-no-registry)? Also, without using a workflow

pawanpinjarkar: The long term goal was to also enable the openshift conformance tests on these jobs and eventually create an informing job. If this is not something needed, I can close this PR.

andfasano: I think we'll have to support something like that for the feature promotion in future (but running the whole job with all the TP features enabled + conformance tests). If the current job remains identical to the presubmit I think it will be safe to close the PR and then readdress it in future

PR #69463, opened 2025-09-19
OCPQE-29421:add first custom prows for olmv1

kuiwang02:
This is an experiment — if it succeeds, more jobs will be added later.


kuiwang02:
/hold
till we get the cases in payload.

kuiwang02:
/unhold

because we get case in payload and the cases pass

PR #69468, opened 2025-09-19
CNTRLPLANE-1454: Remove MCE < 2.4 logic and update default version to 2.4

LiangquanLi930:
- Remove conditional logic for MCE versions below 2.4
- Update default MCE_VERSION from 2.2 to 2.4
- Standardize to use hcp CLI instead of hypershift
- Remove manual ManagedCluster creation for old versions
- Update documentation to reflect >=2.4 support only

### DIFF
--- a/ci-operator/step-registry/hypershift/mce/kubevirt/destroy/hypershift-mce-kubevirt-destroy-chain.yaml
+++ b/ci-operator/step-registry/hypershift/mce/kubevirt/destroy/hypershift-mce-kubevirt-destroy-chain.yaml
@@ -10,12 +10,7 @@ chain:
           source "${SHARED_DIR}/proxy-conf.sh"
         fi
 
-        MCE_VERSION=$(oc get "$(oc get multiclusterengines -oname)" -ojsonpath="{.status.currentVersion}" | cut -c 1-3)
         HYPERSHIFT_NAME=hcp
-        if (( $(awk 'BEGIN {print ("'"$MCE_VERSION"'" < 2.4)}') )); then
-          echo "MCE version is less than 2.4, use hypershift command as default"
-          HYPERSHIFT_NAME=hypershift
-        fi
 
         arch=$(arch)
         if [ "$arch" == "x86_64" ]; then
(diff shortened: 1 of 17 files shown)


ci-operator/step-registry/hypershift/mce/kubevirt/destroy/hypershift-mce-kubevirt-destroy-chain.yaml
```diff
@@ -12,10 +12,6 @@ chain:
 
         MCE_VERSION=$(oc get "$(oc get multiclusterengines -oname)" -ojsonpath="{.status.currentVersion}" | cut -c 1-3)
```
zhfeng: Remove `MCE_VERSION` as well since we don't use it anymore.

ci-operator/step-registry/hypershift/mce/power/create/hcp/hypershift-mce-power-create-hcp-commands.sh
```diff
@@ -21,10 +21,6 @@ export PATH=$PATH:/tmp/bin
 
 MCE_VERSION=$(oc get "$(oc get multiclusterengines -oname)" -ojsonpath="{.status.currentVersion}" | cut -c 1-3)
```
zhfeng, line 22: Same here.

LiangquanLi930: https://github.com/openshift/release/blob/b136c0192e815d8f0e7a63e10ec67e561243d29d/ci-operator/step-registry/hypershift/mce/power/create/hcp/hypershift-mce-power-create-hcp-commands.sh#L142-L144
this step use MCE env

jparrill (commented):
Looks good to me, I will left the tag for @zhfeng. 

Just one question, are we still creating the env using `--render --render-sensitive`?

LiangquanLi930:
> Looks good to me, I will left the tag for @zhfeng.
> 
> Just one question, are we still creating the env using `--render --render-sensitive`?

Currently, IBM P/Z uses this approach, and the same approach is also used for testing different versions of HostedCluster and NodePool.

zhfeng:
Thanks @LiangquanLi930 - lgtm

PR #69545, opened 2025-09-22
Add very basic CLAUDE.md for non-OCP users

mmazur:
Most external users of this repo just want to manager their jobs. This is a very small context file that helps them out with that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,5 +12,6 @@ jenkins/controllers/.idea/*
 clusters/app.ci/.actual_diff
 *.pyc
 hack/gcp-secret-manager/gcp-secrets-venv
+CLAUDE.local.md
 ### macOS ###
 .DS_Store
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -12,5 +12,6 @@ jenkins/controllers/.idea/*
 clusters/app.ci/.actual_diff
 *.pyc
 hack/gcp-secret-manager/gcp-secrets-venv
+CLAUDE.local.md
```
pruan-rht, line 15: @mmazur I don't understand why you want to add the file you added to .gitignore

mmazur: That is a different file.
* CLAUDE.md lives in the repo and provides high level info about what's where.
* CLAUDE.local.md is additional info you specifically want to provide, but isn't universal.

My `CLAUDE.local.md` says:

```
$ cat CLAUDE.local.md 
You work on ARO-HCP and therefore are interested only in jobs under `ci-operator/config/Azure/ARO-HCP/` unless specified otherwise.
```

pruan-rht: /lgtm
/approve

PR #69560, opened 2025-09-22
stackrox: rosa classic fips test

davdhacs:
ROSA classic with fips-mode enabled. This simulates the critical differences of FedRAMP from the perspective of stackrox, and we want to have regular testing to find if we begin failing.

There are some timing issues causing intermittent failures in the groovy tests. However, the failures are not caused by the change of running on rosa with fips, and so I want to enable these tests to begin getting coverage.


ci-operator/config/stackrox/stackrox/stackrox-stackrox-master.yaml
```diff
@@ -246,6 +246,16 @@ tests:
       COLLECTION_METHOD: core_bpf
     workflow: stackrox-automation-flavors-rosa-e2e
   timeout: 5h0m0s
+- always_run: false
+  as: rosa-fips-qa-e2e-tests
+  optional: true
+  steps:
+    env:
+      COLLECTION_METHOD: core_bpf
```
tommartensen, line 254: Do we still need to specify this by the way?

davdhacs: No, I've left most of the tests and configs for the teams to adjust or clean-up, but I think we can drop this at 4.8 and newer as we're not testing with COLLECTION_METHOD:ebpf anymore from 4.8.

davdhacs: core_bpf is:
* default in Collector: https://github.com/stackrox/collector/blob/master/collector/lib/CollectorConfig.h#L34
* default for e2e tests
* default in the deploy/install

And always used since 4.5 even if configured differently: "If EBPF is configured, it will be automatically converted to CORE_BPF" (https://issues.redhat.com/browse/ROX-23155)

PR #69587, opened 2025-09-23
Update istio builder image on Istio ci jobs

fjglira:
Update the builder image for the Istio jobs


ci-operator/config/openshift-service-mesh/istio/openshift-service-mesh-istio-master.yaml
```diff
@@ -3,6 +3,10 @@ base_images:
     name: "4.18"
     namespace: ocp
     tag: cli
+  maistra-builder:
```
FilipB: This is not used anywhere so it can be removed.

fjglira: Ok

FilipB (approved):

fjglira:
Non needed anymore

PR #69593, opened 2025-09-23
MABC: configs only in build10 and pushing to QCI

Prucek:
This PR:
 - moves the MABController configs to a new folder `clusters/build-clusters/multiarch_builds/multi-arch-builder-controller`
 - moves the MABConfigs configs to a new folder `clusters/build-clusters/multiarch_builds/supplemental-ci-images` :
   `managed-clonerefs` and `ci-tools-build-root` (both of them are now directly pushing to QCI)
 - the `clusters/build-clusters/multiarch_builds` is pinned to `build10` with a symlink, and this will be, for now, the primary multiarch build cluster (This can be easily changed by moving the symlink, but we want to build the images only on a single cluster)
 - the configs in cluster-init were updated not to include MABC 
 - the `managed-clonerefs` image IS no longer needed: https://github.com/openshift/ci-tools/pull/4710


deepsm007:
FYI another one @jcpowermac @rvanderp3

PR #69597, opened 2025-09-23
rm permafailing hypershift presubmits from ovnk

jluhrsen:
e2e-aws-ovn-hypershift-conformance-techpreview and e2e-aws-ovn-hypershift-kubevirt are permafailing. nobody cares. remove them.


martinkennelly:
@jluhrsen thank you jamo - this is the right approach if we cannot get folks to maintain CI jobs otherwise its toil reviewing those jobs.

PR #68540, opened 2025-08-27
Run k8sgpt as part of hypershift-dump chain

mgencur:
Adds [k8sgpt](https://github.com/k8sgpt-ai/k8sgpt) integration to the hypershift-dump workflow by:
- Adding k8sgpt step to the dump chain configuration
- Creating new k8sgpt step registry files (commands, metadata, yaml config, and OWNERS)

The step produces files under `hypershift-k8sgpt/artifacts/`:
* namespaces/hypershift/result.json (errors from the management cluster)
* namespaces/clusters/result.json (errors from the management cluster)
* namespaces/clusters-<hc_name>/result.json (errors from the management cluster)
* hostedcluster/result.json (errors from all namespace in hosted/guest cluster)

See [example with errors](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_release/68540/rehearse-68540-periodic-ci-openshift-hypershift-release-4.21-periodics-e2e-aws-ovn-conformance/1962434203760988160/artifacts/e2e-aws-ovn-conformance/hypershift-k8sgpt/artifacts/) from the latest run. It shows an error in the HCP namespace and no errors in the hosted cluster.

The step currently doesn't run the analysis through AI (via `--explain` flag for k8sgpt) as it is currently tied to an OpenShift OpenAI account that reached its limits. The plan is to change this to use Gemini (see [slack thread](https://redhat-internal.slack.com/archives/CH76YSYSC/p1756285954529649)). When this account is available we can enable AI analysis. Until that time, please see https://github.com/openshift/hypershift/pull/6720 which adds a custom Claude command to analyze the output of these errors (instead of doing it directly by k8sgpt by calling AI).

Note: The current usage of k8sgpt doesn't gather errors from Pod logs as it doesn't explicitly use the `Log` analyzer (available through `--filter Log`). This needs further investigation as there are usually errors that are "OK" such as some temporary reconciler error).

PR #69606, opened 2025-09-23
hypershift: reorder kubevirt presub flow to fix break

sjenning:
https://github.com/openshift/release/pull/68540 regressed the kubevirt presub job because the test cluster is destroyed before the dump stage.

https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-openshift-hypershift-main-e2e-kubevirt-aws-ovn-reduced

This reorders the flow so that the k8sgpt script no longer errors and we get meaningful dump artifacts from the test cluster.

### DIFF
--- a/ci-operator/step-registry/hypershift/kubevirt/run-e2e-external/hypershift-kubevirt-run-e2e-external-chain.yaml
+++ b/ci-operator/step-registry/hypershift/kubevirt/run-e2e-external/hypershift-kubevirt-run-e2e-external-chain.yaml
@@ -107,7 +107,3 @@ chain:
         cpu: 100m
         memory: 100Mi
     timeout: 1h0m0s
-  - chain: hypershift-kubevirt-destroy
-    env:
-      - default: "/tmp/secret/kubeconfig"
-        name: KUBECONFIG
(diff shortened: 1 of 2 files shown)


ci-operator/step-registry/hypershift/kubevirt/run-e2e-external/hypershift-kubevirt-run-e2e-external-chain.yaml
```diff
@@ -107,7 +107,3 @@ chain:
         cpu: 100m
         memory: 100Mi
     timeout: 1h0m0s
-  - chain: hypershift-kubevirt-destroy
-    env:
-      - default: "/tmp/secret/kubeconfig"
```
devguyio, line 112: is this env var not needed anymore?

sjenning: That's what the rehearsal is for :sweat_smile:.  Other chains like `hypershift-dump` do not specify a `KUBECONFIG` so I think it is already set.

devguyio: good point 😁 tagged and 🤞🏽 for the rehearsal

PR #69609, opened 2025-09-23
NE-1334: Add CI job to test gatewayAPI e2e tests for pre-release ossm images

rhamini3:
Added a 1 year valid token to access a konflux cluster which will be used to pull the latest ossm image and test NID teams gateway API e2e tests

### DIFF
--- a/ci-operator/config/openshift/cluster-ingress-operator/openshift-cluster-ingress-operator-master.yaml
+++ b/ci-operator/config/openshift/cluster-ingress-operator/openshift-cluster-ingress-operator-master.yaml
@@ -143,6 +143,28 @@ tests:
   steps:
     cluster_profile: aws
     workflow: openshift-e2e-aws-single-node
+- as: e2e-aws-pre-release-ossm
+  optional: true
+  steps:
+    cluster_profile: aws
+    test:
+    - as: test
+      cli: latest
+      commands: |
+        export TOKEN=$(cat /var/run/nid-ossm-token/secrets/konflux-cluster-token)
+        export AUTHBREW=$(cat /var/run/nid-ossm-token/secrets/brew-secret)
+        export AUTHSTAGE=$(cat /var/run/nid-ossm-token/secrets/stage-secret)
+        make test-pre-release-ossm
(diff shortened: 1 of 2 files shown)


ci-operator/config/openshift/cluster-ingress-operator/openshift-cluster-ingress-operator-master.yaml
```diff
@@ -143,6 +143,25 @@ tests:
...
+  steps:
+    cluster_profile: aws
+    test:
+    - as: test
+      cli: latest
+      commands: |
+        export TOKEN=/selfservice/nid-ossm-token/secrets/konflux-cluster-token
+        make test-pre-release-ossm
+      credentials: 
+        mount_path: /selfservice/nid-ossm-token/secrets/konflux-cluster-token
```
alebedev87: As discussed, the mounted secret will create a file for each key, you need to `cat` it into `TOKEN` variable. Plus `/var/run/` is more conventional for mount points.
```suggestion
        export TOKEN=$(cat /var/run/nid-ossm-token/secrets/konflux-cluster-token)
        make test-pre-release-ossm
      credentials: 
        mount_path: /var/run/nid-ossm-token/secrets/
```

ci-operator/config/openshift/cluster-ingress-operator/openshift-cluster-ingress-operator-master.yaml
```diff
@@ -143,6 +143,25 @@ tests:
...
+        name: konflux-cluster-token
```
alebedev87: The name of the secret should match `secretsync/target-name` key from CI Vault. I think it does here but just in case.

alebedev87:
/lgtm
/approve
/hold

Holding for future rehearsals once https://github.com/openshift/cluster-ingress-operator/pull/1283 is merged.

alebedev87:
/lgtm

Keeping the hold until we can get a successful rehearsal of https://github.com/openshift/cluster-ingress-operator/pull/1291 (it has to be merged first, of course).

PR #69612, opened 2025-09-23
Creating external-cluster workflow.

etirta:
Some times there is a need to run the test agains an external cluster. External cluster is a cluster that has been deployed outside the CI.

The content of KUBECONFIG need to be stored in the HashiCorp Vault as `kubeconfig` Secret Key and the Web Console password for `kubeadmin` user also may be stored as `kubeadmin-password` Secret Key.

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -311,6 +311,7 @@ aliases:
   - dfrazzette
   - danalanerh
   - sg-rh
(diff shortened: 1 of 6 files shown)


etirta:
Ping @josecastillolema @paigerube14 @LibinLiu @Userweiwei @dfrazzette, please review. Thx.

PR #69635, opened 2025-09-24
Allow overriding the default Pod CIDR

Sandeepyadav93:
For large clusters or environments, the default Pod CIDR may not be sufficient.
- Introduce POD_CIDR variable to override the rosa default.
- Add logic to pass --pod-cidr to the cluster creation command.
- Log the custom CIDR to the console and the cluster-config file.


chentex (approved):
/lgtm

PR #69641, opened 2025-09-24
Adding a `releaseControllerContainerWaiting` alert

bradmwilliams:
In reference to: https://redhat-internal.slack.com/archives/CNHC2DK2M/p1758673026045799

### DIFF
--- a/clusters/app.ci/openshift-user-workload-monitoring/mixins/_prometheus/release_controller_alerts.libsonnet
+++ b/clusters/app.ci/openshift-user-workload-monitoring/mixins/_prometheus/release_controller_alerts.libsonnet
@@ -66,6 +66,23 @@
           },
         ],
       },
+      {
+        name: 'release-controller-container-waiting',
+        rules: [
+          {
+            alert: 'releaseControllerContainerWaiting',
+            expr: 'kube_pod_container_status_waiting_reason{reason!="CrashLoopBackOff", namespace="ci", pod=~"release-controller.*"} > 0',
+            'for': '1h',
+            labels: {
+              severity: 'warning',
+              team: 'crt',
+            },
+            annotations: {
+              message: 'pod/{{ $labels.pod }} in namespace {{ $labels.namespace }} on container {{ $labels.container}} has been in waiting state for longer than 1 hour. (reason: "{{ $labels.reason }}").',
+            },
+          },
+        ],
+      },
     ],
   },
 }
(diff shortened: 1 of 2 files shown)


clusters/app.ci/openshift-user-workload-monitoring/mixins/_prometheus/release_controller_alerts.libsonnet
```diff
@@ -66,6 +66,23 @@
           },
         ],
       },
+      {
+        name: 'release-controller-container-waiting',
+        rules: [
+          {
+            alert: 'releaseControllerContainerWaiting',
+            expr: 'kube_pod_container_status_waiting_reason{reason!="CrashLoopBackOff", namespace="ci", deployment=~"release-controller.*"} > 0',
```
wking: nit: this is a Pod/container-scoped metric, so you want `pod=~"release-controller.*"` and not `deployment=~"release-controller.*"`.

Other than that, this looks a lot like [the stock `KubeContainerWaiting`][1], so makes sense to me.  And again, we need to copy this into a per-workload PrometheusRule to get [user-workload monitoring coverage][2], because the stock `KubeContainerWaiting` limits itself to `namespace=~"(openshift-.*|kube-.*|default)"` to be platform-focused and avoid distracting cluster admins with user-workload issues.

[1]: https://github.com/openshift/cluster-monitoring-operator/blob/e7abc6a1574d8949b6ff2b8ec8bfba8561d4463c/assets/control-plane/prometheus-rule.yaml#L147-L155
[2]: https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/monitoring/configuring-user-workload-monitoring#configuring-alerts-and-notifications-uwm

bradmwilliams: Updated!

bradmwilliams: > And again, we need to copy this into a per-workload PrometheusRule to get user-workload monitoring coverage, because the stock KubeContainerWaiting limits itself to namespace=~"(openshift-.*|kube-.*|default)" to be platform-focused and avoid distracting cluster admins with user-workload issues.

Just want to confirm, that's what I've done here, but it's literally only for the `release-controller`'s pods.  Correct?

PR #69646, opened 2025-09-24
Updating the ReleaseController's deployments to keel

bradmwilliams:


clusters/app.ci/release-payload-controller/deployment.yaml
Prucek, line 1: Here, the image wasn't changed

hack/generators/release-controllers/content/release_payload_controller_resources.py
Prucek, line 1: Here the image is not updated

deepsm007 (approved):

PR #69697, opened 2025-09-25
Update deprovision to handle vpc and iam users

deepsm007:
/cc @openshift/test-platform

### DIFF
--- a/core-services/ipi-deprovision/aws.sh
+++ b/core-services/ipi-deprovision/aws.sh
@@ -3,90 +3,145 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-trap finish TERM QUIT
+PROFILE="${1:-default}"
+CUTOFF="$(date -d '72 hours ago' --iso-8601=seconds)"
+MAX_RETRIES=5
 
-function finish {
-	CHILDREN=$(jobs -p)
-	if test -n "${CHILDREN}"; then
-		kill ${CHILDREN} && wait
-	fi
-	exit # since bash doesn't handle SIGQUIT, we need an explicit "exit"
+log() { echo "[$(date -u +%Y-%m-%dT%H:%M:%SZ)] $*"; }
+
+retry() {
+  local attempt=0
+  until "$@"; do
+    attempt=$(( attempt + 1 ))
+    if (( attempt >= MAX_RETRIES )); then
+      log "ERROR: command failed after $MAX_RETRIES attempts: $*"
+      return 1
+    fi
+    sleep_time=$(( attempt * 5 ))
+    log "WARN: retrying ($attempt/$MAX_RETRIES) after $sleep_time seconds: $*"
+    sleep "$sleep_time"
+  done
 }
 
-function queue() {
-	local LIVE="$(jobs | wc -l)"
-	while [[ "${LIVE}" -ge 2 ]]; do
-		sleep 1
-		LIVE="$(jobs | wc -l)"
-	done
-	echo "${@}"
-	"${@}" &
+# IAM User Cleanup
+cleanup_iam_users() {
+  aws --profile "$PROFILE" iam list-users \
+    --query "Users[?starts_with(UserName, 'ci-op-') && CreateDate < \`$CUTOFF\`].UserName" \
+    --output text | tr '\t' '\n' | while read -r user; do
+      [[ -z "$user" ]] && continue
+      log "Processing IAM user: $user"
+
+      # detach managed policy
+      for policy in $(aws --profile "$PROFILE" iam list-attached-user-policies \
+        --user-name "$user" --query 'AttachedPolicies[].PolicyArn' --output text 2>/dev/null); do
+          retry aws --profile "$PROFILE" iam detach-user-policy --user-name "$user" --policy-arn "$policy" || true
+      done
+
+      # delete inline policy
+      for policy in $(aws --profile "$PROFILE" iam list-user-policies \
+        --user-name "$user" --query 'PolicyNames[]' --output text 2>/dev/null); do
+          retry aws --profile "$PROFILE" iam delete-user-policy --user-name "$user" --policy-name "$policy" || true
+      done
+
+      # delete access key
+      for key in $(aws --profile "$PROFILE" iam list-access-keys \
+        --user-name "$user" --query 'AccessKeyMetadata[].AccessKeyId' --output text 2>/dev/null); do
+          retry aws --profile "$PROFILE" iam delete-access-key --user-name "$user" --access-key-id "$key" || true
+      done
+
+      # remove from group
+      for group in $(aws --profile "$PROFILE" iam get-groups-for-user \
+        --user-name "$user" --query 'Groups[].GroupName' --output text 2>/dev/null); do
+          retry aws --profile "$PROFILE" iam remove-user-from-group --user-name "$user" --group-name "$group" || true
+      done
+
+      # delete user
+      if retry aws --profile "$PROFILE" iam delete-user --user-name "$user"; then
+        log "✓ Deleted IAM user: $user"
+      else
+        log "✗ Failed to delete IAM user: $user"
+      fi
+  done
 }
 
-function deprovision() {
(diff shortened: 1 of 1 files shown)


core-services/ipi-deprovision/aws.sh
```diff
@@ -5,88 +5,197 @@ set -o pipefail
 
 trap finish TERM QUIT
 
+AWS_PROFILE="${AWS_PROFILE:-default}"
+CLUSTER_TTL="${CLUSTER_TTL:-12 hours}" # TTL for IAM users and clusters
```
bear-redhat: i think its safer to set to larger than 72hr+buffer

core-services/ipi-deprovision/aws.sh
```diff
@@ -5,88 +5,197 @@ set -o pipefail
...
+        set -e
+
+        if [[ $STATUS -eq 0 ]]; then
+            echo "✓ VPC ${VPC_ID} deleted successfully"
+            break
+        else
+            echo "⚠ Attempt $i failed for VPC ${VPC_ID}, retrying in ${delay}s ..."
+            sleep $delay
+            delay=$((delay*2))
+        fi
```
bear-redhat: missing Route53 records and zones

core-services/ipi-deprovision/aws.sh
```diff
@@ -5,88 +5,197 @@ set -o pipefail
...
-		timeout --signal=SIGQUIT 60m openshift-install --dir "${WORKDIR}" --log-level error destroy cluster && touch "${WORKDIR}/success" || touch "${WORKDIR}/failure"
-	fi
+    local WORKDIR="${1}"
+    local REGION
+    local INFRA_ID
+    REGION=$(jq -r .aws.region "${WORKDIR}/metadata.json")
+    INFRA_ID=$(jq -r '.aws.identifier[0]|keys[0]' "${WORKDIR}/metadata.json" | cut -d '/' -f3)
+
+    set +e
+    if [[ -n ${HYPERSHIFT_PRUNER:-} ]]; then
```
bear-redhat: can we reuse the existing hypershift pruner?  or just relies on the hypershift pruner and focus on deprovisioning ipi clusters

bear-redhat (changes requested):

core-services/ipi-deprovision/aws.sh
liangxia: If a job runs for X hours, will those IAM users remove after 12 hours or X+12 hours ?

deepsm007: i updated with timestamp to avoid confusion

core-services/ipi-deprovision/aws.sh
```diff
@@ -3,90 +3,145 @@ set -o errexit
...
-	REGION="$(cat ${WORKDIR}/metadata.json|jq .aws.region -r)"
-	INFRA_ID="$(cat ${WORKDIR}/metadata.json|jq '.aws.identifier[0]|keys[0]' -r|cut -d '/' -f3|tr -d '\n')"
-	if [[ -n ${HYPERSHIFT_PRUNER:-} ]]; then
-		HYPERSHIFT_BASE_DOMAIN="${HYPERSHIFT_BASE_DOMAIN:-origin-ci-int-aws.dev.rhcloud.com}"
-		timeout --signal=SIGQUIT 30m hypershift destroy infra aws --aws-creds "${AWS_SHARED_CREDENTIALS_FILE}" --infra-id "${INFRA_ID}" --base-domain "${HYPERSHIFT_BASE_DOMAIN}" --region "${REGION}" || touch "${WORKDIR}/failure"
-		timeout --signal=SIGQUIT 30m hypershift destroy iam aws --aws-creds "${AWS_SHARED_CREDENTIALS_FILE}" --infra-id "${INFRA_ID}" --region "${REGION}" || touch "${WORKDIR}/failure"
-	else
-		timeout --signal=SIGQUIT 60m openshift-install --dir "${WORKDIR}" --log-level error destroy cluster && touch "${WORKDIR}/success" || touch "${WORKDIR}/failure"
-	fi
+# VPC cleanup
```
patrickdillon, line 94: the vpc deletion I saw was blocked on firewalls, so you should probably check those too.

core-services/ipi-deprovision/aws.sh
```diff
@@ -3,90 +3,145 @@ set -o errexit
...
-function deprovision() {
-	WORKDIR="${1}"
-	REGION="$(cat ${WORKDIR}/metadata.json|jq .aws.region -r)"
-	INFRA_ID="$(cat ${WORKDIR}/metadata.json|jq '.aws.identifier[0]|keys[0]' -r|cut -d '/' -f3|tr -d '\n')"
-	if [[ -n ${HYPERSHIFT_PRUNER:-} ]]; then
-		HYPERSHIFT_BASE_DOMAIN="${HYPERSHIFT_BASE_DOMAIN:-origin-ci-int-aws.dev.rhcloud.com}"
-		timeout --signal=SIGQUIT 30m hypershift destroy infra aws --aws-creds "${AWS_SHARED_CREDENTIALS_FILE}" --infra-id "${INFRA_ID}" --base-domain "${HYPERSHIFT_BASE_DOMAIN}" --region "${REGION}" || touch "${WORKDIR}/failure"
-		timeout --signal=SIGQUIT 30m hypershift destroy iam aws --aws-creds "${AWS_SHARED_CREDENTIALS_FILE}" --infra-id "${INFRA_ID}" --region "${REGION}" || touch "${WORKDIR}/failure"
-	else
-		timeout --signal=SIGQUIT 60m openshift-install --dir "${WORKDIR}" --log-level error destroy cluster && touch "${WORKDIR}/success" || touch "${WORKDIR}/failure"
```
patrickdillon, line 35: I have a lot of concerns about removing the installer from this cleanup. This script does not comprehensively delete all resources created in a cluster. If a job is interrupted before deprovisioning with the installer, then all of those cluster resources are sitting around, and we would expect them to be deprovisioned by this script.  

But because this doesn't handle all resources, we're going to leak a lot...

patrickdillon:
/hold

I could be misunderstanding here, but if this is the only backstop for jobs that get installed but interrupted before deprovisioning, I'm very concerned about all of the resources we're going to leak. I'm surprised in testing that you were able to delete the VPCs as this doesn't handle VMs or load balancers? 

If I'm misunderstanding, please correct me.

core-services/ipi-deprovision/aws.sh
deepsm007: yeah, makes sense. This is a temp fix and will improve it later on

deepsm007:
/unhold

I'll follow up with more things in separate PR and need to look at logs on what else is left over that can be removed properly

core-services/ipi-deprovision/aws.sh
hector-vido: > But because this doesn't handle all resources, we're going to leak a lot...

Certainly not more than already leak today.
This is a first step, we gonna merge this, watch the logs and keep adding stuff next days.

wking: I agree that you want to keep the installer in this loop, because there are timeouts on Prow jobs as a whole and also on the deprovision step specifically, and there are also no guarantees that the build-cluster Node running a deprovision step survives through the whole run.  So having this script trying to detect leaked clusters and use the installer to clean them up is covering that leakage and giving you the benefit of all the time and effort the installer folks sink into "how to delete an AWS cluster".

However, as the rest of this script shows, there are some reasources that leak where it no longer looks enough like "a cluster" for the installer to delete them, or where the installer can struggle to apply its usual deletion logic.  IAM users are a good example, where the installer's "walk all users and check labels to find the ones owned by _my cluster_" breaks down under AWS throttling when there are too many users to look through.  In those cases, _supplementing_ the installer with more direct deletion logic (where we can make CI-specific assumptions like "all user names like `$THIS` that are older than `$DURATION` are leaked, and we can delete them") will help keep the account clean so the installer can get back to working efficiently.

PR #69717, opened 2025-09-25
Cleanup IAM user

deepsm007:
/cc @openshift/test-platform

### DIFF
--- a/core-services/ipi-deprovision/aws.sh
+++ b/core-services/ipi-deprovision/aws.sh
@@ -82,6 +82,27 @@ done
 
 wait
 
+# IAM user cleanup (ci-op-* older than 72h)
+cutoff="$(date -u -d '72 hours ago' --iso-8601=seconds)"
+aws iam list-users --query "Users[?starts_with(UserName, 'ci-op-') && CreateDate < '${cutoff}'].UserName" --output text | tr '\t' '\n' | while read -r user; do
+	if [[ -n "$user" ]]; then
+		echo "Cleaning IAM user: $user"
+		aws iam list-attached-user-policies --user-name "$user" --query 'AttachedPolicies[].PolicyArn' --output text | tr '\t' '\n' | while read -r policy; do
+			[[ -n "$policy" ]] && aws iam detach-user-policy --user-name "$user" --policy-arn "$policy" || true
+		done
+		aws iam list-user-policies --user-name "$user" --query 'PolicyNames[]' --output text | tr '\t' '\n' | while read -r policy; do
+			[[ -n "$policy" ]] && aws iam delete-user-policy --user-name "$user" --policy-name "$policy" || true
+		done
+		aws iam list-access-keys --user-name "$user" --query 'AccessKeyMetadata[].AccessKeyId' --output text | tr '\t' '\n' | while read -r key; do
+			[[ -n "$key" ]] && aws iam delete-access-key --user-name "$user" --access-key-id "$key" || true
+		done
+		aws iam get-groups-for-user --user-name "$user" --query 'Groups[].GroupName' --output text | tr '\t' '\n' | while read -r group; do
+			[[ -n "$group" ]] && aws iam remove-user-from-group --user-name "$user" --group-name "$group" || true
+		done
+		aws iam delete-user --user-name "$user" && echo "✓ Deleted: $user"
+	fi
+done
+
 FAILED="$(find ${clusters} -name failure -printf '%H\n' | sort)"
 if [[ -n "${FAILED}" ]]; then
 	echo "Deprovision failed on the following clusters:"


core-services/ipi-deprovision/aws.sh
```diff
@@ -44,6 +127,26 @@ if [[ -n ${HYPERSHIFT_PRUNER:-} ]]; then
 	done
 	# Exit here if we had errors, otherwise we destroy the OIDC providers for the hostedclusters and deadlock deletion as cluster api creds stop working so it will never be able to remove machine finalizers
 	if [[ $had_failure -ne 0 ]]; then exit $had_failure; fi
+	# IAM user cleanup (ci-op-* older than 72h)
+	cutoff="$(date -u -d '72 hours ago' --iso-8601=seconds)"
+	aws iam list-users --query "Users[?starts_with(UserName, 'ci-op-') && CreateDate < '${cutoff}'].UserName" --output text | tr '\t' '\n' | while read -r user; do
```
wking: You're currently putting this in the `HYPERSHIFT_PRUNER` block.  But leaked IAM users don't seem like a HyperShift-specific thing.  Maybe you want to shift this to the main level of the script?  And you may want to move it to after the `queue deprovision` step, so that it doesn't have to get involved with the users that the installer is able to delete itself?

core-services/ipi-deprovision/aws.sh
```diff
@@ -23,6 +23,82 @@ function queue() {
 	"${@}" &
 }
 
+function retry() {
```
mtulio: are we using version 1 or 2 of cli? There is a native retry on CLI, have you tried it?

https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-retries.html

deepsm007: https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zi

core-services/ipi-deprovision/aws.sh
```diff
@@ -34,6 +110,13 @@ function deprovision() {
 	else
 		timeout --signal=SIGQUIT 60m openshift-install --dir "${WORKDIR}" --log-level error destroy cluster && touch "${WORKDIR}/success" || touch "${WORKDIR}/failure"
 	fi
+
+	# Failsafe: if cluster destroy failed, clean VPCs and Route53
+    if [[ -f "${WORKDIR}/failure" ]]; then
+        cluster="${WORKDIR##*/}"
+        force_cleanup_vpc "${REGION}" "${cluster}"
+        force_cleanup_route53 "${cluster}"
```
mtulio: Another suggestion for tagged resources (may no apply to R53 records):

I would suggest to [query the resource API](https://docs.aws.amazon.com/tag-editor/latest/userguide/find-resources-to-tag.html) to gather all resources with the cluster owned tag, so you can save listings API calls across all services, focusing only on resources that is "leaked". 

Here is one example/idea:
```sh
aws resourcegroupstaggingapi get-resources --tag-filters Key=kubernetes.io/cluster/"${cluster}",Values=owned  --region "${REGION}"
```

Returns:

```json
{
    "ResourceTagMappingList": [
        {
            "ResourceARN": "arn:aws:ec2:us-east-1:726924432237:subnet/subnet-0e95cea2c4061ddb7",
            "Tags": [
                {
                    "Key": "sigs.k8s.io/cluster-api-provider-aws/cluster/mrb-sg-v47-nd8kf",
                    "Value": "owned"
                },
                {
                    "Key": "sigs.k8s.io/cluster-api-provider-aws/role",
                    "Value": "public"
                },
                {
                    "Key": "kubernetes.io/role/elb",
                    "Value": "1"
                },
                {
                    "Key": "Name",
                    "Value": "mrb-sg-v47-nd8kf-subnet-public-us-east-1a"
                },
                {
                    "Key": "kubernetes.io/cluster/mrb-sg-v47-nd8kf",
                    "Value": "owned"
                }
            ]
        },
....
```

Then:
- Save the output to later check to ensure this script is not missing resources
- Parse the resource type from the ARN
- Call delete the delete function by service, be carefully to the dependencies (or maybe run in parallel with a soft exponential backoff/retry?)

core-services/ipi-deprovision/aws.sh
```diff
@@ -78,6 +453,26 @@ openshift-install version
 clusters=$( find "${logdir}" -mindepth 1 -type d )
 for workdir in $(shuf <<< ${clusters}); do
 	queue deprovision "${workdir}"
+	# IAM user cleanup (ci-op-* older than 72h)
+	cutoff="$(date -u -d '72 hours ago' --iso-8601=seconds)"
+	aws iam list-users --query "Users[?starts_with(UserName, 'ci-op-') && CreateDate < '${cutoff}'].UserName" --output text | tr '\t' '\n' | while read -r user; do
```
wking: This doesn't need to be in the `for workdir ...` block, because it's not cluster-specific.  I'd expect IAM pruning to run _after_ the per-cluster loop, like:

```shell
for workdir in $(shuf <<< ${clusters}); do
	queue deprovision "${workdir}"
done

wait  # until the queued deprovision calls all wrap up.

# IAM user cleanup (ci-op-* older than 72h)
...your IAM user pruning

FAILED="$(find ${clusters} -name failure -printf '%H\n' | sort)"
if [[ -n "${FAILED}" ]]; then
	echo "Deprovision failed on the following clusters:"
	xargs --max-args 1 basename <<< $FAILED
	exit 1
fi

echo "Deprovision finished successfully"
```
```

File: core-services/ipi-deprovision/aws.sh (hunk 3 above)
Comment (wking) [reply core-services/ipi-deprovision/aws.sh]: I suggest you avoid directly deleting resources that `resourcegroupstaggingapi` turns up with the `kubernetes.io/cluster/${cluster}=owned` label.  I agree that those can be deleted, but that's [how the bulk of the installer's AWS cluster deletion logic works][1], so it doesn't seem worthwhile to me to reproduce that in shell.  To find an example, https://prow.ci.openshift.org/job-history/gs/test-platform-results/logs/periodic-ipi-deprovision-aws -> [example run][2] with build-logs:

```
deprovision /logs/artifacts/deprovision/ci-op-ti0bt08p-312d2-x588d
Deprovision finished successfully
```

Dropping into [gathered artifacts for that cluster's deletion][3]:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/periodic-ipi-deprovision-aws/1971600804922003456/artifacts/deprovision/ci-op-ti0bt08p-312d2-x588d/.openshift_install.log | head
time="2025-09-26T16:02:54Z" level=debug msg="OpenShift Installer unreleased-master-11950-g249d7428c51fe85cabdab6394c4aed86fc24d398-dirty"
time="2025-09-26T16:02:54Z" level=debug msg="Built from commit 249d7428c51fe85cabdab6394c4aed86fc24d398"
time="2025-09-26T16:02:54Z" level=info msg="Credentials loaded from the \"default\" profile in file \"/aws/.awscred\""
time="2025-09-26T16:02:54Z" level=debug msg="search for matching resources by tag in us-west-2 matching aws.Filter{\"kubernetes.io/cluster/ci-op-ti0bt08p-312d2-x588d\":\"owned\"}"
time="2025-09-26T16:02:54Z" level=debug msg="search for matching resources by tag in us-east-1 matching aws.Filter{\"kubernetes.io/cluster/ci-op-ti0bt08p-312d2-x588d\":\"owned\"}"
time="2025-09-26T16:02:54Z" level=debug msg="search for IAM roles"
time="2025-09-26T16:02:55Z" level=debug msg="iterating over a page of 100 IAM roles"
time="2025-09-26T16:03:01Z" level=debug msg="iterating over a page of 100 IAM roles"
time="2025-09-26T16:03:07Z" level=debug msg="iterating over a page of 100 IAM roles"
time="2025-09-26T16:03:14Z" level=debug msg="iterating over a page of 100 IAM roles"
```

It also records the resources it deleted:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/periodic-ipi-deprovision-aws/1971600804922003456/artifacts/deprovision/ci-op-ti0bt08p-312d2-x588d/.openshift_install.log | grep -v level=debug | head
time="2025-09-26T16:02:54Z" level=info msg="Credentials loaded from the \"default\" profile in file \"/aws/.awscred\""
time="2025-09-26T16:09:39Z" level=info msg=Deleted id=net/ci-op-ti0bt08p-312d2-x588d-ext/c416ba9d6217a196
time="2025-09-26T16:09:41Z" level=info msg=Deleted classic load balancer=adc933c1d529546f9ad0920e13ae72d2 id=vpc-09624557af6898db8 resourceType=vpc
time="2025-09-26T16:09:41Z" level=info msg=Deleted id=vpc-09624557af6898db8 load balancer=loadbalancer/net/ci-op-ti0bt08p-312d2-x588d-int/a87f0e9728b7e3bc resourceType=vpc
time="2025-09-26T16:09:42Z" level=info msg=Deleted NAT gateway=nat-0a00aaf5a930e6717 id=vpc-09624557af6898db8 resourceType=vpc
time="2025-09-26T16:09:42Z" level=info msg=Deleted NAT gateway=nat-06e3af7d4860b80b4 id=vpc-09624557af6898db8 resourceType=vpc
time="2025-09-26T16:09:43Z" level=info msg=Deleted id=ci-op-ti0bt08p-312d2-openshift-machine-api-aws-88c7n policy=ci-op-ti0bt08p-312d2-openshift-machine-api-aws-88c7n-policy
time="2025-09-26T16:09:43Z" level=info msg=Deleted id=ci-op-ti0bt08p-312d2-openshift-machine-api-aws-88c7n
time="2025-09-26T16:09:44Z" level=info msg=Deleted id=vol-0460097ba31181eb0 resourceType=volume
time="2025-09-26T16:09:45Z" level=info msg=Deleted id=ci-op-ti0bt08p-312d2-x588d-master-role name=ci-op-ti0bt08p-312d2-x588d-master-role policy=ci-op-ti0bt08p-312d2-x588d-master-policy
grep: write error: Broken pipe
```

I think it's fine to add logic to delete leaked resources in new, CI-specific ways that the installer cannot delete (e.g. the installer needs to be generic, and cannot hard-code "all `ci-op-` IAM users over `$DURATION` hours old are leaked, and should be deleted", but this CI-specific script absolutely can hard-code that context).  But I'd recommend against this script encoding generic stuff that the installer already covers, because that seems like significant maintenance work without a clear value.

[1]: https://github.com/openshift/installer/blob/1fbf7f2a9f47714adb832ed657dcfd423bbae4a9/pkg/destroy/aws/aws.go#L258-L322
[2]: https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ipi-deprovision-aws/1971600804922003456
[3]: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/periodic-ipi-deprovision-aws/1971600804922003456/artifacts/deprovision/ci-op-ti0bt08p-312d2-x588d/

core-services/ipi-deprovision/aws.sh
patrickdillon: The installer already performs tag-based destroy. If we think there are enhancements that can be made to tag-based destroy, it would be best to incorporate those changes into the installer.

The tactical advantage you have in this case is that you can use assumptions about the testing account to implement more aggressive destroy code, for example you can delete everything older than a certain timestamp, like you're doing for IAM.

PR #69722, opened 2025-09-26
Add lgtm label for QE assignment

duanwei33:
It will help notify QE for pre-merge testing and add the /verify label.


duanwei33:
cc @jsafrane

PR #69728, opened 2025-09-26
Fix: add 'MCE_VERSION' env for mce-dump

LiangquanLi930:


veera-damisetti:
Looks good @LiangquanLi930 
Thanks for the fix

veera-damisetti (approved):

PR #69748, opened 2025-09-26
ovnk, virt, periodics: add single stack IPv4/v6 jobs

maiqueb:


maiqueb:
@JoelSpeed adding the lanes you've requested in https://github.com/openshift/api/pull/2496

@everettraven FYI

ci-operator/config/openshift/ovn-kubernetes/openshift-ovn-kubernetes-release-4.21__periodics.yaml
```diff
@@ -81,6 +81,90 @@ tests:
...
+  steps:
+    cluster_profile: equinix-ocp-metal
+    env:
+      CNV_PRERELEASE_LATEST_CHANNEL: "true"
+      DEVSCRIPTS_CONFIG: |
+        IP_STACK=v4
+      EXTRA_MG_ARGS: --host-network
+      EXTRA_NETWORK_CONFIG: |
+        EXTRA_NETWORK_NAMES="extranet"
+        EXTRANET_NETWORK_SUBNET_V4='192.168.221.0/24'
```
qinqon, line 96: We are not testing localnet so we don't need any of those, let's remove it so we have less stuff going on at the udnerlay

ci-operator/config/openshift/ovn-kubernetes/openshift-ovn-kubernetes-release-4.21__periodics.yaml
```diff
@@ -81,6 +81,90 @@ tests:
...
```
qinqon, line 118: ditto

ci-operator/config/openshift/ovn-kubernetes/openshift-ovn-kubernetes-release-4.21__periodics.yaml
```diff
@@ -81,6 +81,90 @@ tests:
...
+  steps:
+    cluster_profile: equinix-ocp-metal
+    env:
+      CNV_PRERELEASE_LATEST_CHANNEL: "true"
+      DEVSCRIPTS_CONFIG: |
+        IP_STACK=v6
+      EXTRA_MG_ARGS: --host-network
+      EXTRA_NETWORK_CONFIG: |
+        EXTRA_NETWORK_NAMES="extranet"
+        EXTRANET_NETWORK_SUBNET_V6='fd2e:6f44:5dd8:ca56::/120'
```
qinqon: Ditto also the error that we have with installing openshift virt can be related to the fact that an extra default gw is being configured if we do this.

qinqon (commented):
We don't need the extra interface at nodes since we are not testing localnet at origin, let's remove it can cause connectivity issues since as part of it a extra default gw and dns server is configured, libvirt provide those as part of DHCP answer.

ci-operator/config/openshift/ovn-kubernetes/openshift-ovn-kubernetes-release-4.21__periodics.yaml
```diff
@@ -81,6 +81,48 @@ tests:
       FEATURE_SET: TechPreviewNoUpgrade
     workflow: baremetalds-e2e-ovn-bgp-virt-dualstack
   timeout: 6h0m0s
+- as: e2e-metal-ipi-ovn-bgp-virt-single-stack-ipv4-techpreview
```
kyrtapz: NIT: I don't think we usually name the jobs `single-stack` as it is the default, same with ipv4.
What do you think about going with `e2e-metal-ipi-ovn-bgp-virt-techpreview` and `e2e-metal-ipi-ovn-bgp-virt`?

maiqueb: I'm confused since this is the only comment I got, but is a nit. Should I do comply with this request or ignore it ? 

I don't care that much about the name, and I'd rather do my little bit to save the rain forest a lil' bit by not requiring another CI round on this, but I'll blindly trust the maintainers judgement .

kyrtapz: Let's just rename it please to keep it consistent.
We don't have to wait for another rehearsal.

PR #69753, opened 2025-09-26
Network edge: skip tests when no code is changed

rikatz:
This change skips e2e tests on NI&D repos, when all of the changes are only not-code change (eg.: OWNERS, docs, etc).


rikatz:
the error is related with timeout. We can confirm that the job per se is running fine

/pj-rehearse ack

PR #69754, opened 2025-09-26
volsync - update to use go 1.24 builder

tesshuflower:


JohnStrunk (approved):
/lgtm

tesshuflower:
/cc @JohnStrunk 

Sorry John, do you mind reviewing again? It did that thing where it automatically removed the old replace builder section (for golang 1.23) since there was no longer a match in our Dockerfile - so I had to rebase to fix the conflict.

PR #69756, opened 2025-09-26
Fix: Fix e2e-test-image-stream-import-mode Presubmit Job to include TechPreviewNoUpgrade

jatinsu:
[This previous PR](https://github.com/openshift/release/pull/69661) did not enable TechPreviewNoUpgrade which is a requirement as it enables the ImageStreamImportMode featuregate, allowing the test suite to run.

### DIFF
--- a/ci-operator/config/openshift/origin/openshift-origin-main.yaml
+++ b/ci-operator/config/openshift/origin/openshift-origin-main.yaml
@@ -281,13 +281,13 @@ tests:
       - observers-resource-watch
     workflow: openshift-upgrade-aws
 - always_run: false
-  as: e2e-test-image-stream-import-mode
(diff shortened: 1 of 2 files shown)


ci-operator/config/openshift/origin/openshift-origin-main.yaml
```diff
@@ -286,6 +286,7 @@ tests:
   steps:
     cluster_profile: aws
     env:
+      FEATURE_SET: TechPreviewNoUpgrade
```
neisw, line 289: You should update the name to include `-techpreview` as well then

jatinsu: Makes sense! Just added it.

PR #68875, opened 2025-09-04
Fix issue for internal registry post check

jinyunma:
Failed jobs:
1. https://qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/qe-private-deck/logs/periodic-ci-openshift-openshift-tests-private-release-4.19-multi-nightly-azure-ipi-public-to-private-amd-f28-destructive/1961799616546476032
```
 privateEndpointName check...
ERROR: Get unexpected value! expected value: imageregistryciopz9wm7xl; actual value: null
subnetName check...
ERROR: Get unexpected value! expected value: ci-op-z9inz982-09ee5-h7dvw-master-subnet; actual value: null
vnetName check...
ERROR: Get unexpected value! expected value: ci-op-z9inz982-09ee5-h7dvw-vnet; actual value: null 
```

2. Even first issue has been fixed, still encounter the second issue:
https://qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/qe-private-deck/pr-logs/pull/openshift_release/67002/rehearse-67002-periodic-ci-openshift-openshift-tests-private-release-4.20-multi-nightly-azure-ipi-public-to-private-arm-f28/1963484566916173824
```
 storage account name check...
ERROR: Get unexpected value! expected value: imageregistryciopcfcjgw8
imageregistryciopcfphhvb; actual value: imageregistryciopcfcjgw8 
```

PR #69772, opened 2025-09-28
Fix post check failure for job azure-ipi-sharednetwork-public-to-private

jinyunma:
Continued https://github.com/openshift/release/pull/68875

Issue should be fixed in https://github.com/openshift/release/pull/68875, but forgot to update installation chain `azure-ipi-sharednetwork-public-to-private`


jinyunma:
also cc @gpei to review

PR #69808, opened 2025-09-29
[content-mirror] Purging outdated deployments

bradmwilliams:
These older releases are no longer being produced and therefore the existence of the associated content-mirrors are not necessary.  

rh-pre-commit.version: 2.3.2
rh-pre-commit.check-secrets: ENABLED


Prucek:
Hey @jcpowermac @rvanderp3! We would need your approval here

PR #69809, opened 2025-09-29
stackrox: add ocp420 to nightlies

davdhacs:
* nightlies will move to ocp4.20 as the latest supported ocp

The tests are all passing: https://prow.ci.openshift.org/pr-history/?org=stackrox&repo=stackrox&pr=17048 (tested on master)

### DIFF
--- a/ci-operator/config/stackrox/stackrox/stackrox-stackrox-release-4.8__ocp-4-19.yaml
+++ b/ci-operator/config/stackrox/stackrox/stackrox-stackrox-release-4.8__ocp-4-19.yaml
@@ -25,7 +25,7 @@ tests:
   steps:
     env:
       COLLECTION_METHOD: core_bpf
-      OCP_VERSION: ocp/candidate-4.20
+      OCP_VERSION: ocp/candidate-4.19
       TEST_SUITE: ocp-qa-e2e-tests
     workflow: stackrox-automation-flavors-ocp-4-e2e
   timeout: 5h0m0s
@@ -36,7 +36,7 @@ tests:
   steps:
     env:
       COLLECTION_METHOD: core_bpf
-      OCP_VERSION: ocp/candidate-4.20
+      OCP_VERSION: ocp/candidate-4.19
       TEST_SUITE: ocp-scanner-v4-install-tests
     workflow: stackrox-automation-flavors-ocp-4-e2e
   timeout: 5h0m0s
@@ -45,7 +45,7 @@ tests:
   run_if_changed: ^((generated|image|operator|pkg|make|deploy|scripts)/.*|Makefile|status\.sh|[A-Z_]+_VERSION)
   steps:
     env:
-      OCP_VERSION: ocp/candidate-4.20
+      OCP_VERSION: ocp/candidate-4.19
       TEST_SUITE: ocp-operator-e2e-tests
     workflow: stackrox-automation-flavors-ocp-4-e2e
   timeout: 5h0m0s
@@ -55,7 +55,7 @@ tests:
   steps:
     env:
       COLLECTION_METHOD: core_bpf
-      OCP_VERSION: ocp/candidate-4.20
+      OCP_VERSION: ocp/candidate-4.19
       TEST_SUITE: ocp-sensor-integration-tests
     workflow: stackrox-automation-flavors-ocp-4-e2e
   timeout: 5h0m0s
@@ -65,7 +65,7 @@ tests:
   steps:
     env:
       COLLECTION_METHOD: core_bpf
-      OCP_VERSION: ocp/candidate-4.20
+      OCP_VERSION: ocp/candidate-4.19
       TEST_SUITE: ocp-ui-e2e-tests
     workflow: stackrox-automation-flavors-ocp-4-e2e
   timeout: 5h0m0s
@@ -75,7 +75,7 @@ tests:
   steps:
     env:
       COLLECTION_METHOD: core_bpf
-      OCP_VERSION: ocp/candidate-4.20
+      OCP_VERSION: ocp/candidate-4.19
       TEST_SUITE: ocp-nongroovy-e2e-tests
     workflow: stackrox-automation-flavors-ocp-4-e2e
   timeout: 5h0m0s
@@ -86,7 +86,7 @@ tests:
     env:
       COLLECTION_METHOD: core_bpf
       FIPS_ENABLED: "true"
-      OCP_VERSION: ocp/candidate-4.20
+      OCP_VERSION: ocp/candidate-4.19
       TEST_SUITE: ocp-qa-e2e-tests
     workflow: stackrox-automation-flavors-ocp-4-e2e
   timeout: 5h0m0s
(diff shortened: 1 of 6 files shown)


ci-operator/config/stackrox/stackrox/stackrox-stackrox-nightlies__ocp-4-20.yaml
```diff
@@ -0,0 +1,89 @@
...
+      cpu: 2000m
+      memory: 4000Mi
+test_binary_build_commands: .openshift-ci/dispatch.sh test-binary-build-commands
+tests:
+- as: qa-e2e-tests
+  postsubmit: true
+  steps:
+    env:
+      COLLECTION_METHOD: core_bpf
+      OCP_VERSION: ocp/candidate-4.20
```
msugakov, line 27: I tried diffing this file with `ci-operator/config/stackrox/stackrox/stackrox-stackrox-nightlies__ocp-4-19.yaml` and found that the latter has `OCP_VERSION: ocp/candidate-4.19`. Shouldn't that be `OCP_VERSION: ocp/stable-4.19`?

If yes, `ocp/candidate-4.19` is not just in that file, it's also mentioned in all release branches.

ci-operator/config/stackrox/stackrox/stackrox-stackrox-nightlies__ocp-4-20.yaml
```diff
@@ -0,0 +1,89 @@
...
+    name: apollo-ci
+    namespace: stackrox
+    tag: stackrox-ui-test-0.4.9
+resources:
+  '*':
+    requests:
+      cpu: 2000m
+      memory: 4000Mi
+test_binary_build_commands: .openshift-ci/dispatch.sh test-binary-build-commands
+tests:
```
msugakov, line 21: Also, when diffing to `ci-operator/config/stackrox/stackrox/stackrox-stackrox-nightlies__ocp-4-19.yaml`, I noticed that the latter has `24nodes-scale-test`, `24nodes-scale-scannerv4-test` while this new `ci-operator/config/stackrox/stackrox/stackrox-stackrox-nightlies__ocp-4-20.yaml` does not.

Assuming `stackrox-stackrox-nightlies__ocp-4-20.yaml` becomes "main" at some point, I fear if `stackrox-stackrox-nightlies__ocp-4-19.yaml` simply gets deleted and those 24 nodes scale tests are gone. What's the process of moving these around?

davdhacs: 🤔 I think I'll move the perf&scale to a separate config file. A few weeks ago, @mtodor and I discussed that we do want to manually pin and track the openshift version for the perf&scale tests. So having it in a separate file will prevent it from getting lost or bumped as part of the overall support change.

ci-operator/config/stackrox/stackrox/stackrox-stackrox-nightlies__ocp-4-20.yaml
davdhacs: You're right, I have left them as "candidate" from the pre-GA stage like where 4.20 is right now (no stable-4.20 exists yet). I'm justifying leaving it using the candidate-x.y builds because they are patches that are unlikely to change at GA and so we're seeing the "latest" version a few days or a week before it reaches customers ("faster-fast").
What do you think? I think I really just forgot to bump 4.19 to stable after GA, but I like the idea of fewer scheduled steps to update the config files.

msugakov: I see 4.19.15 in [ocp/candidate-4.19](https://mirror.openshift.com/pub/openshift-v4/clients/ocp/candidate-4.19/) and 4.19.15 in [ocp/fast-4.19](https://mirror.openshift.com/pub/openshift-v4/clients/ocp/fast-4.19/), and the checksums are identical. So you're right that `ocp/candidate` seems to be updated after the release.

I tried finding a confirmation in docs but the latest doc on channels I found is [for 4.13](https://docs.redhat.com/en/documentation/openshift_container_platform/4.13/html/updating_clusters/understanding-upgrade-channels-releases#understanding-upgrade-channels_understanding-upgrade-channels-releases), and it doesn't make the matter clear referring to GA just like `ocp/candidate-x.y` will be populated only before the first `x.y` GA release and then abandoned.

> The candidate-4.13 channel offers unsupported early access to releases as soon as they are built. Releases present only in candidate channels may not contain the full feature set of eventual GA releases or features may be removed prior to GA. Additionally, these releases have not been subject to full Red Hat Quality Assurance and may not offer update paths to later GA releases. Given these caveats, the candidate channel is only suitable for testing purposes where destroying and recreating a cluster is acceptable. 

Do you know if there's any newer source that confirms we can rely on `ocp/candidate` to be "faster-fast"? I'm in general ok to rely on what we factually see but in either case it will be good to make it more explicit in [our doc](https://spaces.redhat.com/spaces/StackRox/pages/272005264/Making+changes+to+OpenShift+CI) and link the confirming info if it exists.

ci-operator/config/stackrox/stackrox/stackrox-stackrox-nightlies__ocp-4-20.yaml
msugakov: Do you plan this in the same or a separate PR?

ci-operator/config/stackrox/stackrox/stackrox-stackrox-nightlies__ocp-4-20.yaml
davdhacs: Well, I found where it is configured but not in the docs yet. Example for the promotion of the latest patch to ocp4.19 to `fast` includes a link to the pr for promotion to `candidate`: https://github.com/openshift/cincinnati-graph-data/pull/8126

davdhacs: Here is a newer doc with an explanation matching candidate is "faster-fast": https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/updating_clusters/understanding-openshift-updates-1#update-availability_understanding-openshift-updates
"""
A new release is initially added to the candidate channel.
After successful final testing, a release on the candidate channel is promoted to the fast channel, an errata is published, and the release is now fully supported.
"""

ci-operator/config/stackrox/stackrox/stackrox-stackrox-nightlies__ocp-4-20.yaml
davdhacs: I'd like to do that in a separate PR.

ci-operator/config/stackrox/stackrox/stackrox-stackrox-nightlies__ocp-4-20.yaml
davdhacs: To our doc, I added an opinionated explanation of the stream promotion and why I think we should primarily use the "candidate" stream. And I updated the link to the ocp4.19 stream doc.

ci-operator/config/stackrox/stackrox/stackrox-stackrox-nightlies__ocp-4-20.yaml
davdhacs: I created a PR to move the perf&scale test out of the ocp config file: https://github.com/openshift/release/pull/70292

ci-operator/config/stackrox/stackrox/stackrox-stackrox-nightlies__ocp-4-20.yaml
msugakov: Looks good, thank you!

PR #69810, opened 2025-09-29
Fix misspelling: {HYPER|OPEN}SHFIT -> SHIFT

lunarwhite:
A boring patch


lunarwhite:
This is a no-op change 

/pj-rehearse ack

PR #69849, opened 2025-09-30
Prow jobs need new mirrors

tbuskey:
[KATA-4225](https://issues.redhat.com/browse/KATA-4225)

Use the mirrors in git for OSC and trustee directly
Update openshift test private image in prowjobs to latest.  Otherwise we don't get any of our changes


littlejawa (approved):
/lgtm

PR #69861, opened 2025-09-30
NE-2142: ExternalDNS Operator - Update OCP to 4.21

alebedev87:


alebedev87:
operator-sdk image didn't change since `4.17` that's _probably_ why it's not available in `4.21` image stream.

PR #69864, opened 2025-09-30
DPTP-4552: Periodic: Prune zombie ClusterPools

2uasimojo:
Heretofore, when hive ClusterPool users would rename or delete a ClusterPool configuration from the release repository, it would remain on the hosted-mgmt cluster.

This introduces a periodic job to scrub such zombie ClusterPools. Here's how it works:
- Find ClusterPool config files in the release repo and extract a list of their namespace+name.
- Query ClusterPools on the hosted-mgmt cluster, listing their namespace+name.
- Compare the two lists, reporting the entries which exist in the second list but not the first.
- Delete those ClusterPools.

Note that if the first discovery/filtering operation fails to produce a complete/correct list, we could end up deleting ClusterPools we oughtn't. We're guarding against this by
- Running the script under `-e` and `pipefail`; and
- Aborting if that file winds up empty...

...but it's still a risk.

### DIFF
--- a/ci-operator/jobs/infra-build-farm-periodics.yaml
+++ b/ci-operator/jobs/infra-build-farm-periodics.yaml
@@ -496,3 +496,68 @@ periodics:
     - name: hive
       secret:
         secretName: hypershift-workload-credentials
+- agent: kubernetes
+  cluster: build04
+  cron: '@daily'
+  decorate: true
+  decoration_config:
+    timeout: 1h0m0s
+  labels:
+    pj-rehearse.openshift.io/can-be-rehearsed: "true"
+  name: periodic-openshift-dptp-4552-clusterpool-pruner
+  reporter_config:
+    slack:
+      channel: '#team-hive-alert'
+      job_states_to_report:
+      - failure
+      - error
+      - success
+      report_template: 'Job {{.Spec.Job}}: <{{.Status.URL}}|View logs>'
+  spec:
+    containers:
+    - command:
+      - sh
+      - -c
+      - |
+        set -e
+        set -o pipefail
+        echo "Discovering configured ClusterPools..."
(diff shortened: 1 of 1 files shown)


ci-operator/jobs/infra-build-farm-periodics.yaml
```diff
@@ -496,3 +496,65 @@ periodics:
...
+    slack:
+      channel: '#team-hive-alert'
+      job_states_to_report:
+      - failure
+      - error
+      - success
+      report_template: 'Job {{.Spec.Job}}: <{{.Status.URL}}|View logs>'
+  spec:
+    containers:
+    - command:
```
deepsm007, line 518:
```suggestion
    - command:
      - sh
      - -c
      - |
```

2uasimojo: Thanks, trying...

ci-operator/jobs/infra-build-farm-periodics.yaml
```diff
@@ -496,3 +496,67 @@ periodics:
...
+      - sh
+      - -c
+      - |
+        set -x
```
deepsm007:
```suggestion
```

ci-operator/jobs/infra-build-farm-periodics.yaml
```diff
@@ -496,3 +496,67 @@ periodics:
...
+        set -e
+        set -o pipefail
+        find clusters/hosted-mgmt/hive/pools -type f -name '*.yaml' | xargs grep 'kind: ClusterPool' | awk -F: '{print $1}' | xargs /tmp/yq e '[.metadata.namespace, .metadata.name] | join(" ")' | sort > /tmp/clusterpools.configured
+        if ! [[ -s /tmp/clusterpools.configured ]]; then
+          echo "Discovered no configured ClusterPools. This probably means something is wrong. Aborting."
```
deepsm007:
```suggestion
          echo "ERROR: Discovered no configured ClusterPools. This probably means something is wrong. Aborting."
```

PR #69891, opened 2025-10-01
Kubevirt chaos

paigerube14:


ci-operator/config/redhat-chaos/lp-chaos/OWNERS
```diff
@@ -0,0 +1,10 @@
+approvers:
+  - etirta 
+  - LibinLiu 
+  - Userweiwei 
+reviewers:
+  - paigerube14 
+  - chaitanyaenr
+  - etirta 
+  - LibinLiu 
+  - Userweiwei
```
etirta: Since our group has [alias](https://github.com/openshift/release/blob/b24f0fa4ff6d2fdbc47a5384f81f2aba70b4de4a/OWNERS_ALIASES#L491) defined already, please use the alias instead.
```suggestion
approvers:
  - ieng-chaos
reviewers:
  - ieng-chaos
  - paigerube14
  - chaitanyaenr
```

Or if both approvers & reviewers should be the same (I don't really get the idea why they are different 😁).
```suggestion
approvers:  &owners
  - ieng-chaos
  - paigerube14
  - chaitanyaenr
reviewers: *owners
```

ci-operator/config/redhat-chaos/lp-chaos/redhat-chaos-lp-chaos-main__ocp4.19.yaml
```diff
@@ -0,0 +1,52 @@
...
+    namespace: chaos
+    tag: latest
+  ocp-qe-perfscale-ci:
+    name: ocp-qe-perfscale-ci
+    namespace: ci
+    tag: latest
+  upi-installer:
+    name: "4.19"
+    namespace: ocp
+    tag: upi-installer
```
etirta: Do we really need these keys? Who uses those images? The steps are using `from_image` directly.

Can we try to use the minimal config? I would think the more we define in here, the longer the job will take as it will need to mirror the image into the test namespace. Also generally speaking, it's always the best not putting something unrelated to the Job, my 2 cents.

ci-operator/config/redhat-chaos/lp-chaos/redhat-chaos-lp-chaos-main__ocp4.19.yaml
```diff
@@ -0,0 +1,52 @@
...
+  capabilities:
+  - intranet
+  steps:
+    allow_best_effort_post_steps: true
+    allow_skip_on_success: true
+    cluster_profile: metal-redhat-gs
+    env:
+      KUBE_VIRT_FAILURES: "True"
+      TARGET_NAMESPACE: benchmark-runner
+    pre:
```
etirta: I would prefer us to use the `workflow`, so it's much easier to switch it when the BM Assisted Installation workflow is available. Unless there is a reason that make sense for not using workflow.

ci-operator/step-registry/redhat-chaos/kubevirt-outage/redhat-chaos-kubevirt-outage-commands.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+set -o errexit
+
+console_url=$(oc get routes -n openshift-console console -o jsonpath='{.spec.host}')
+export HEALTH_CHECK_URL=https://$console_url
+oc get vmis -A
+set -o nounset
+set -o pipefail
+set -x
```
etirta: I think it's preferable to have this on top, so we get the full script "line-echo" (line 4-5).

ci-operator/step-registry/redhat-chaos/kubevirt-outage/redhat-chaos-kubevirt-outage-commands.sh
```diff
@@ -0,0 +1,42 @@
...
+telemetry_password=$(cat "/secret/telemetry/telemetry_password")
+export TELEMETRY_PASSWORD=$telemetry_password
+
+export NAMESPACE=$TARGET_NAMESPACE 
+
+oc get vmis -A 
+
+export KUBE_VIRT_NAMESPACE=$TARGET_NAMESPACE
+./kubevirt-outage/prow_run.sh
+rc=$?
```
etirta: Since the script is running with `set -e`, non-0 exit code from execution of `./kubevirt-outage/prow_run.sh` will stop the script. So really printing on line 42 will never occurs in case of that script execution exit with error. If it's printed, it will always be `0` (successful).

If the execution of line 38-40 is desired regardless of the exit status of `./kubevirt-outage/prow_run.sh`, then do the following instead:
```suggestion
./kubevirt-outage/prow_run.sh || rc=$?
```
The above will not stop the script and assign the var. `rc` to the exit status of `./kubevirt-outage/prow_run.sh` execution (and this particular assignment will always be successful). Then line 38-40 will be executed, even the `./kubevirt-outage/prow_run.sh` return an error and it make sense to print the `rc` on line 42.

I'm also a strong advocate to use explicit `exit` (or `return` in function) at the end for script that the parent monitor its exit status.

A POSIX Shell, have a caveat that absent of explicit `exit` or `return` statement, the exit status of the script or function execution is actually the exit status of the LAST executed line (not necessarily the actual physical last line, for example in case of the end of the script is a `case` statement, then every last line of the branch potentially become the last executed line of the script or function).

So if the last executed line validly yield non-0 (normal case in logical compound statement), while the execution is actually completed successfully, the exit status may actually non-0. For example, if the last line is copying an optional file:
```
 ...
[ -e /some/optional/file ] && cp /some/optional/file /target/location/
```
In case of that optional file actually doesn't exist, that last line yield a non-0 exit status and become the exit status of the script execution and will kill the Job.

Since in CI Operator the Step must yield 0 exit status for it to continue, to avoid unexpected failure, it's better to always put an explicit `exit` statement at the end of the script, so in case of there is future modification like the above example, it doesn't kill the Job.

Not really necessary in this particular step per se, but I find it safer to avoid future headache if we are discipline in enforcing an explicit `exit` statement at the end of every Step Shell Script. Just my 2 cents.

ci-operator/config/redhat-chaos/lp-chaos/redhat-chaos-lp-chaos-main__ocp4.19.yaml
paigerube14: when a ref has just "from: <image>" it doesn't always pull the needed image. Having these images listed verifies it has the correct images during the run

ci-operator/step-registry/redhat-chaos/kubevirt-outage/redhat-chaos-kubevirt-outage-commands.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+set -o nounset
+set -o errexit
+
+console_url=$(oc get routes -n openshift-console console -o jsonpath='{.spec.host}')
+export HEALTH_CHECK_URL=https://$console_url
+oc get vmis -A
+
+set -o pipefail
+set -x
```
etirta, line 10: I meant the whole `set` series should be on top. Unless you really don't want the "line-echo" mode (`set -c`) for commands in line 5-7 in the above.

Right now, by simply checking the log, we won't see what is the `console_url` (hence the `HEALTH_CHECK_URL`)  is set to. If this is desired, then this is fine, but if you want those to be seen in the log, then please move it on top.

Generally speaking, all the `set` and `shopt` should be the first lines of the script, unless there is specific reason for it, say we just want to "temporarily" turn of/of some feature/setting for just a block of lines.

ci-operator/config/redhat-chaos/lp-chaos/redhat-chaos-lp-chaos-main__ocp4.19.yaml
```diff
@@ -0,0 +1,39 @@
+base_images:
+  cli:
+    name: "4.19"
+    namespace: ocp
+    tag: cli
```
etirta, line 5: I got this idea for listing it in here. But I'm just curious (it's more for me to actually understand how CI Operator works, then anything else), which Step that actually uses this image?

Correct me if I'm wrong, but in my understanding, there are basically only 2 Step executed by this Job. The `openshift-qe-installer-bm-load-kubeconfig` (via `external-cluster` workflow) and `redhat-chaos-kubevirt-outage` (the actual test Step). They both use `from_image` that if I understand it correctly, it doesn't even use this `base_images` settings, instead it just directly go to the `Namepace/ImageStreamTag` as defined by the `from_image` key.

I was able to run some simple Job (with `external-cluster` workflow) without even having `base_images` key at all.

CI Operator unfortunately is a very complex application, and I have to admit it wasn't easy for me to understand it. So one way to help I try to be discipline and create Job configuration with minimal configuration. I believe this approach help future engineer to have a better understanding on what or which settings are actually matters / needed.

Again, this is just my 2 cents. If this is really needed of course I'm fine with it, but I just want to know the exact reason. Please bare with me as I'm trying to reach the `know-why` rather than just the `know-how`, so I will ask a lot of "annoying" questions during my onboarding 😁.

ci-operator/step-registry/redhat-chaos/kubevirt-outage/redhat-chaos-kubevirt-outage-commands.sh
paigerube14: I have seen the set and pipefail give false failures when performing some oc commands so thats why it is after getting the health check url

ci-operator/config/redhat-chaos/lp-chaos/redhat-chaos-lp-chaos-main__ocp4.19.yaml
paigerube14: its good to specify the cli image specific to the version you are running, there have been version/go errors in the past when the cli is one version against a different versioned cluster. This helps specify the correct versioned with the same cluster version

The must gather uses the cli image and want to be sure it matches to get the correct must-gather when needing to look into failure runs

https://github.com/openshift/release/blob/3bde7aa6c7d8557c0c942df9375915e89b707bb1/ci-operator/step-registry/gather/must-gather/gather-must-gather-ref.yaml#L5

ci-operator/step-registry/redhat-chaos/kubevirt-outage/redhat-chaos-kubevirt-outage-commands.sh
etirta: If after setting `set -o pipefail` the `oc` commands fails, don't you think we want it to be really fail? What do you mean by "false" failure?
The current command is: `console_url=$(oc get routes -n openshift-console console -o jsonpath='{.spec.host}')`, there is no pipe involved, so the `set -o pipefail` will not be the reason why it failed. So for this particular script, I don't think you need to worry about it, since there is no pipe involved in setting the `console_url`. And we definitely want to have `set -x` before this command, so if this `oc` command indeed failed, we wan't to see in in the log for post-mortem analysis.

For other piped compound command, next time you see the "false" failure due to `set -o pipefail` please let me know. I think the pipe commands are not done properly and I will be able to suggest the proper solution. I honestly never seen "false" failure in my experience with various POSIX Shell.

If what you meant by "false" failure is the commands in prev. pipe may be have non-0 exit status, but it should not cause the script to exit with error, then the proper solution is:
```
set -e
set -o pipefail
 ...
{ command1 | commmand2; } || true
 ...
```
Say `command1` may give a non-0 exit status, but it's not considered a failure that warrant the script to be stopped, then just put `|| true` to ignore it. There are indeed some cases that validly justified to do this.

ci-operator/config/redhat-chaos/lp-chaos/redhat-chaos-lp-chaos-main__ocp4.19.yaml
etirta: Cool!! Thank you for this explanation. This is exactly what I'm looking for when I ask these kind of question 🎉.

ci-operator/step-registry/redhat-chaos/kubevirt-outage/redhat-chaos-kubevirt-outage-commands.sh
paigerube14: that one command doesn't matter too much to get the output on, I print the variable later on that I don't need it. This set up aligns with our 10 other scripts and I'd like to keep them consistent. We can edit in the future but for this starting PR can we get this merged?

etirta: Okay that's reasonable enough.

PR #69928, opened 2025-10-02
hypershift: add a default and techpreview job for external oidc testing

everettraven:
This updates the hypershift testing for the External OIDC feature(s) to run both default and tech preview jobs.

Necessary for automatically verifying https://github.com/openshift/hypershift/pull/6025

### DIFF
--- a/ci-operator/config/openshift/hypershift/openshift-hypershift-release-4.21__periodics.yaml
+++ b/ci-operator/config/openshift/hypershift/openshift-hypershift-release-4.21__periodics.yaml
@@ -163,7 +163,15 @@ tests:
   steps:
     cluster_profile: hypershift
     env:
-      CI_TESTS_RUN: TestExternalOIDCTechPreview
+      CI_TESTS_RUN: TestExternalOIDC
+      OAUTH_EXTERNAL_OIDC_PROVIDER: keycloak
+    workflow: hypershift-aws-e2e-external-oidc
+- as: e2e-aws-external-oidc-techpreview
+  minimum_interval: 12h
+  steps:
+    cluster_profile: hypershift
+    env:
+      CI_TESTS_RUN: TestExternalOIDC
       OAUTH_EXTERNAL_OIDC_PROVIDER: keycloak
       TECH_PREVIEW_NO_UPGRADE: "true"
     workflow: hypershift-aws-e2e-external-oidc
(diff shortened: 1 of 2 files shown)


ci-operator/config/openshift/hypershift/openshift-hypershift-release-4.21__periodics.yaml
```diff
@@ -163,7 +163,15 @@ tests:
...
-      CI_TESTS_RUN: TestExternalOIDCTechPreview
+      CI_TESTS_RUN: TestExternalOIDC
+      OAUTH_EXTERNAL_OIDC_PROVIDER: keycloak
+    workflow: hypershift-aws-e2e-external-oidc
+- as: e2e-aws-external-oidc-techpreview
+  minimum_interval: 12h
+  steps:
+    cluster_profile: hypershift
+    env:
+      CI_TESTS_RUN: TestExternalOIDC
```
xingxingxia, line 174: @everettraven , @xiuwang , should tp job test tp cases as below? Otherwise the uid/extra cases would be run twice in the tp job?
```suggestion
      CI_TESTS_RUN: TestExternalOIDCTechPreview
```

/lgtm cancel

xiuwang: +1

everettraven: I do see where we are ending up running them twice in the TP job.

I think for the short-term this is acceptable. I am planning to remove the `TestExternalOIDCTechPreview` test method in a PR to openshift/hypershift after this PR merges. I couldn't do this beforehand or else the existing job before this change would be broken.

I'm wanting to get us to a state where we are using a singular entrypoint for the tests to run so we don't have to maintain multiple test entrypoints moving forward. 

The purpose of this is to ensure that all the tests are gated on their feature-gate enablement status so that when we promote them, they automatically get included in the set of tests run by default instead of us also needing to go update the test code to move them from the `*TechPreview` test method to the "default" method.

xingxingxia: Understood
/lgtm
@sjenning , help take a look / approve

PR #69932, opened 2025-10-02
Adding chaos profiles to get general pull secret

paigerube14:
Need to add chaos cluster secrets to be able to get the general pull-secret and have access to registry to install red hat operators


pruan-rht (commented):
/lgtm
/approve

PR #69934, opened 2025-10-02
feat: run periodic jobs every 24h instead of every 4h

flacatus:


psturc (approved):
/lgtm
/pj-rehearse ack
/retest

PR #69947, opened 2025-10-03
RHIDP-9114: Change from single varibles to tuple format for backend and db replicas.

ghost:


ci-operator/config/redhat-performance/backstage-performance/redhat-performance-backstage-performance-main.yaml
```diff
@@ -427,8 +416,7 @@ tests:
       SCALE_CPU_REQUESTS_LIMITS: ':'
       SCALE_DB_STORAGES: 2Gi
       SCALE_MEMORY_REQUESTS_LIMITS: ':'
-      SCALE_RHDH_DB_REPLICAS: 1 2 3 4 5
-      SCALE_RHDH_REPLICAS: 1 2 3 4 5
+      SCALE_REPLICAS: "1:1"
```
pmacik: @skestwal This should not be reduced to `1:1` but rather to `1:1 2:2 3:3 4:4 5:5'
```suggestion
      SCALE_REPLICAS: "1:1 2:2 3:3 4:4 5:5"
```

ci-operator/config/redhat-performance/backstage-performance/redhat-performance-backstage-performance-main.yaml
```diff
@@ -448,8 +436,7 @@ tests:
       SCALE_CPU_REQUESTS_LIMITS: ':'
       SCALE_DB_STORAGES: 2Gi
       SCALE_MEMORY_REQUESTS_LIMITS: ':'
-      SCALE_RHDH_DB_REPLICAS: 6 7 8 9 10
-      SCALE_RHDH_REPLICAS: 6 7 8 9 10
+      SCALE_REPLICAS: "1:1"
```
pmacik: @skestwal This should not be reduced to `1:1` but rather to `6:6 7:7 8:8 9:9 10:10'
```suggestion
      SCALE_REPLICAS: "6:6 7:7 8:8 9:9 10:10"
```

PR #69955, opened 2025-10-03
Fix e2e-metal-ipi-ovn-bgp-virt-ipv4[-techpreview]

kyrtapz:
Add a workflow for ipv4 and reuse it in the TechPreview and GA lanes.


maiqueb (approved):

kyrtapz:
https://github.com/openshift/origin/pull/30338

PR #69960, opened 2025-10-03
arohcp: run frontend simulator on every PR in ARO-HCP

bennerv:
In Azure github org land, we want to make required CI status checks.  One of these we want to be is frontend-simulator, but unfortunately if a job doesn't trigger because of a path check, it will not run and the CI will stall merging the PR since it's a required check.

There are painful ways around this, but this is the least impactful and will ensure we don't merge frontend/api changes unless our integration tests are passing. 

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks

### DIFF
--- a/ci-operator/config/Azure/ARO-HCP/Azure-ARO-HCP-main.yaml
+++ b/ci-operator/config/Azure/ARO-HCP/Azure-ARO-HCP-main.yaml
@@ -49,12 +49,11 @@ resources:
       memory: 200Mi
 test_binary_build_commands: make test-compile
 tests:
-- always_run: false
+- always_run: true
   as: frontend-simulation
   capabilities:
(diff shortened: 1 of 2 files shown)


ci-operator/config/Azure/ARO-HCP/Azure-ARO-HCP-main.yaml
```diff
@@ -54,7 +54,6 @@ tests:
   capabilities:
   - nested-podman
   nested_podman: true
-  run_if_changed: ^(frontend)|^(internal)
```
deads2k, line 57: modify the always_run above

PR #69970, opened 2025-10-03
Remove baseDomain for azure

deepsm007:
https://redhat.enterprise.slack.com/archives/CBN38N3MW/p1759526570267999?thread_ts=1759502756.084789&channel=CBN38N3MW&message_ts=1759526570.267999

https://issues.redhat.com/browse/DPTP-4566
/cc @openshift/test-platform @wking

### DIFF
--- a/ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh
+++ b/ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh
@@ -43,6 +43,13 @@ echo "get ocp version: ${version}"
 rm pull-secret
 popd
 
+if [[ ! -r "${CLUSTER_PROFILE_DIR}/baseDomain" ]]; then
+  echo "Using default value: ${BASE_DOMAIN}"
+  AZURE_BASE_DOMAIN="${BASE_DOMAIN}"
+else
+  AZURE_BASE_DOMAIN=$(< ${CLUSTER_PROFILE_DIR}/baseDomain)
+fi
+
 CONFIG="${SHARED_DIR}/install-config.yaml"
 
 REGION="${LEASED_RESOURCE}"
@@ -80,7 +87,7 @@ echo "Using compute node replicas: ${workers}"
 echo "Using controlPlane node replicas: ${master_replicas}"
 
 cat >> "${CONFIG}" << EOF
-baseDomain: ${BASE_DOMAIN}
+baseDomain: ${AZURE_BASE_DOMAIN}
 platform:
   azure:
     region: ${REGION}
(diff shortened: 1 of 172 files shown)


deepsm007:
fixing the config-metadata check

ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh
```diff
@@ -43,6 +43,13 @@ echo "get ocp version: ${version}"
 rm pull-secret
 popd
 
+if [[ ! -r "${CLUSTER_PROFILE_DIR}/baseDomain" ]]; then
+  echo "Using default value: ${BASE_DOMAIN}"
+  AZURE_BASE_DOMAIN="${BASE_DOMAIN}"
```
wking, line 48: This feels backwards to me.  I'd expect most jobs that requested a custom base domain to have that ask respected, regardless of a cluster-profile-level default.  And if the job didn't have a custom preference, it would fall back to the cluster-profile-level default.  And that most (all?) jobs wouldn't have a custom preference.  If that sounds good, something like:

```bash
if [[ -n "${BASE_DOMAIN}" ]]; then
  echo "Using default value: ${BASE_DOMAIN}"
  AZURE_BASE_DOMAIN="${BASE_DOMAIN}"
elif [[ -f "${CLUSTER_PROFILE_DIR}/baseDomain" ]; then
  AZURE_BASE_DOMAIN=$(< ${CLUSTER_PROFILE_DIR}/baseDomain)
else
  echo "Neither BASE_DOMAIN nor cluster profile baseDomain set, but this step requires a configured base domain." >&2
  exit 1
fi

File: ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-ref.yaml [hunk 2]
```diff
@@ -52,7 +52,7 @@ ref:
     default: ci.azure.devcluster.openshift.com
```
wking, line 52: If you buy [this argument][1], you'll want to drop the default here, or set it equal to the empty string like `SIZE_VARIANT` and some of the other variables in this file.

[1]: https://github.com/openshift/release/pull/69970#discussion_r2403400956

deepsm007: I set similar expectations for AWS, we can discuss more on this next week. If this isnt blocking we can merge hoping azure issue will be relieved

ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh
bear-redhat: i would actually suggest to use `-n` than `-r` since `not readable` is another error (it should not happen)

PR #69972, opened 2025-10-03
lower quota slices due to Network LoadBalancer quotas being too low

pruan-rht:
for ref: https://redhat-internal.slack.com/archives/CBN38N3MW/p1759524934407769?thread_ts=1758630821.871019&cid=CBN38N3MW


core-services/prow/02_config/generate-boskos.py
```diff
@@ -16,28 +16,28 @@
         'us-east-1': 10,
     },
     'aws-2-quota-slice': {
-        'us-east-1': 50,
-        'us-east-2': 35,
-        'us-west-1': 35,
-        'us-west-2': 35,
+        'us-east-1': 25,
```
wking: I don't understand why we'd drop this, with a current NLB-per-region quota in AWS-2 of 200, and [a requirement for 2 NLBs per cluster][1], it doesn't seem like the NLB quota would be a limiting factor in this account.

[1]: https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/installing_on_aws/installing-aws-account#installation-aws-limits_installing-aws-account

pruan-rht: readjusted to match aws-1 quotas

PR #70002, opened 2025-10-06
Onboard openshift-online/rosa-enhancements to prow

tiwillia:
The goal is only to enable prow plugins/automation. There are not any tests or builds that need to be performed today.

It is currently unclear to me if this will work without a ci-operator config.


tiwillia:
@smg247 @hector-vido this should be ready for review, passing all tests and fairly simple assuming my understanding is correct.

smg247:
You don't need either of us to approve this, just needs an lgtm from someone on your team.

PR #70194, opened 2025-10-13
Upgrade Golang version to 1.24 to fix CI build failure

xiaoyu74:
This PR updates the build root image to use Go 1.24 in order to resolve the CI build error observed in [openshift/backplane-cli#800](https://github.com/openshift/backplane-cli/pull/800)


feichashao:
Updating the linter in my local solved the problem:
```
go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest
```

PR #70212, opened 2025-10-13
Remove gathering of loadbalancers from openstack-gather step

stephenfin:
We are currently seeing failures [in this step](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_csi-operator/438/pull-ci-openshift-csi-operator-main-e2e-openstack-cinder-csi/1977529467593035776). Based on [what we are able to collect](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_csi-operator/438/pull-ci-openshift-csi-operator-main-e2e-openstack-cinder-csi/1977529467593035776/artifacts/e2e-openstack-cinder-csi/openstack-gather/artifacts/), it seems the gathering of load balancers is the issue. We do not have enough context to figure out what is going on, so remove the steps for now so we can debug later.

### DIFF
--- a/ci-operator/step-registry/openstack/gather/openstack-gather-commands.sh
+++ b/ci-operator/step-registry/openstack/gather/openstack-gather-commands.sh
@@ -180,21 +180,5 @@ for fip in $(jq -r '.[].ID' "${ARTIFACT_DIR_JSON}/openstack_fip_list.json"); do
 	openstack floating ip show "$fip"             >> "${ARTIFACT_DIR}/openstack_fips.log"
 done
 
-# load balancers
-
-openstack loadbalancer list -f json \
-    | jq --arg CLUSTER_NAME "$CLUSTER_NAME" 'map(select(.name | test($CLUSTER_NAME)))' \
-    > "${ARTIFACT_DIR_JSON}/openstack_loadbalancer_list.json"
-
-for lb in $(jq -r '.[].ID' "${ARTIFACT_DIR_JSON}/openstack_loadbalancer_list.json"); do
-	openstack loadbalancer show "$lb" -f json
-done | jq --slurp '.' > "${ARTIFACT_DIR_JSON}/openstack_loadbalancer_show.json"
-
-touch "${ARTIFACT_DIR}/openstack_loadbalancers.log"
-for lb in $(jq -r '.[].ID' "${ARTIFACT_DIR_JSON}/openstack_loadbalancer_list.json"); do
-	echo -e "\n$ openstack loadbalancer show $lb" >> "${ARTIFACT_DIR}/openstack_loadbalancers.log"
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/openstack/gather/openstack-gather-commands.sh
```diff
@@ -180,21 +180,5 @@ for fip in $(jq -r '.[].ID' "${ARTIFACT_DIR_JSON}/openstack_fip_list.json"); do
 	openstack floating ip show "$fip"             >> "${ARTIFACT_DIR}/openstack_fips.log"
 done
 
-# load balancers
-
-openstack loadbalancer list -f json \
-    | jq --arg CLUSTER_NAME "$CLUSTER_NAME" 'map(select(.name | test($CLUSTER_NAME)))' \
```
mandre, line 186: Should it have been `map(select(.Name | test($CLUSTER_NAME)))` with uppercase Name? For when we restore the additional log gathering.

Let's drop the change for now since it's causing the CI to fail.

stephenfin: No, I think `name` is correct. This works locally for me.

```
❯ openstack loadbalancer list -f json | jq --arg CLUSTER_NAME "phbbp-hchsv" 'map(select(.name | test($CLUSTER_NAME)))'
[
  {
    "id": "1dc263eb-0505-4a2b-af2c-c5a1d4dae968",
    "name": "kube_service_autoscaling-phbbp-hchsv_openshift-ingress_router-default",
    "project_id": "0693e2bb538c42b79a49fe6d2e61b0fc",
    "vip_address": "10.0.0.167",
    "provisioning_status": "ERROR",
    "operating_status": "OFFLINE",
    "provider": "amphora"
  }
]
```

ci-operator/step-registry/openstack/gather/openstack-gather-commands.sh
```diff
@@ -180,21 +180,5 @@ for fip in $(jq -r '.[].ID' "${ARTIFACT_DIR_JSON}/openstack_fip_list.json"); do
...
-    > "${ARTIFACT_DIR_JSON}/openstack_loadbalancer_list.json"
-
-for lb in $(jq -r '.[].ID' "${ARTIFACT_DIR_JSON}/openstack_loadbalancer_list.json"); do
```
mandre, line 189: then the problem appears to be that we should be using lowercase `.id` here, and on line 194 too.

stephenfin: Yup, that's likely the one :facepalm: Will repropose once this has merged. More important to get things passing first.

stephenfin: Revert with these fixes included proposed here https://github.com/openshift/release/pull/70248

PR #70219, opened 2025-10-13
ART-14167: Remove cluster-api-operator

mdbooth:


damdo:
@mdbooth anything we need to do here/ping anyone? Thanks!

PR #70239, opened 2025-10-14
OCPEDGE-2154: feat: update two node steps to use ofcir

eggfoobar:
with equinix sun setting the metal offering, we need to move away from using those services, this starts us off by moving our two node jobs over to using the ofcir infra for ci


eggfoobar:
/pj-rehearse ack

The credential change has gone through, and some of those failures are not related.

PR #70252, opened 2025-10-14
Set sysType to s1022 for PowerVS nodes in fran04

Neha-dot-Yadav:
Switching fra04 to use the s1022 system type to test a broader set of systems. 
Update the install-config generation script to set sysType to s1022 for both compute and worker nodes when deploying to this leased resource.


prb112 (approved):
Based on the code, this makes sense and force lease "fran-powervs-8-quota-slice-0 to use s1022.

PR #70256, opened 2025-10-14
Using standard OCP instead of hypershift for OSSM jobs

FilipB:


dgn (approved):
LGTM but we need to make sure we keep those OCI references up to date

PR #70271, opened 2025-10-14
scanner 2.38

daynewlee:


ci-operator/config/stackrox/scanner/stackrox-scanner-release-2.38.yaml
```diff
@@ -0,0 +1,89 @@
+base_images:
+  ubi-minimal:
+    name: ubi-minimal
+    namespace: ocp
+    tag: "8"
+build_root:
+  image_stream_tag:
+    name: apollo-ci
+    namespace: stackrox
+    tag: scanner-test-0.4.8
```
dcaravel: Why the downgrade from `scanner-test-0.4.9`?

BradLugo:
Looks like we need to update the test data. The important thing is that the tests are running as expected.
/pj-rehearse skip

PR #70281, opened 2025-10-14
adding pf5 build to the release

blublinsky:
adding console-pf5 to build


raptorsun:
how about we exclude the changes in operator repo from this PR and add the operator changes into a follow up PR?
I think we need to merge the changes on lightspeed-console in order to start pushing `lightspeed-console-plugin-pf5` images into the image stream `ols`

PR #70348, opened 2025-10-16
OCPBUGS-57482: OpenStack: Bump LVM operator and set namespace

mandre:


stephenfin:
@mandre I'll let you leave rehearsals-ack, assuming there's nothing else you want to check

PR #70356, opened 2025-10-16
vSphere serial techpreview failing for timeout

jcpowermac:
This commit extends 4.21 and 4.20 techpreview serial to 7 hours.

Something changed around 10/8 - 10/10.


neisw:
/approve

Adding a note that vshpere is not a highly scalable environment so sharding serial jobs like these isn't a good option.  Overall the desire is not to have jobs run this long but other than breaking the job up we don't have a better option currently.

PR #70358, opened 2025-10-16
RHDH configure prowgen

zdrapela:
Configure reporters in `.config.prowgen` with the new filtering using job names: https://docs.ci.openshift.org/docs/how-tos/notification/. With this, we no longer need to define the reporters for `auth-providers` and `upgrade` jobs manually after creating jobs for a new release branch.

This ensures the reporter config is generated for jobs after a new file is created in `ci-operator/config/redhat-developer/rhdh` (e.g., new `release-1.9` jobs).

Configure all jobs to send a report in case of error (e.g., OpenShift CI issue), and the fine-grained report will not trigger.

https://issues.redhat.com/browse/RHIDP-10976

### DIFF
--- a/ci-operator/jobs/redhat-developer/rhdh/redhat-developer-rhdh-main-periodics.yaml
+++ b/ci-operator/jobs/redhat-developer/rhdh/redhat-developer-rhdh-main-periodics.yaml
@@ -193,6 +193,13 @@ periodics:
     ci.openshift.io/generator: prowgen
     pj-rehearse.openshift.io/can-be-rehearsed: "true"
   name: periodic-ci-redhat-developer-rhdh-main-e2e-aks-helm-nightly
+  reporter_config:
+    slack:
+      channel: '#rhdh-e2e-test-alerts'
+      job_states_to_report:
+      - error
+      report_template: '<!subteam^S07BMJ56R8S> <@U08UP0REWG1> :failed: `{{.Spec.Job}}`
+        ended with *{{.Status.State}}*, :scroll: <{{.Status.URL}}| logs>'
   spec:
     containers:
     - args:
@@ -253,6 +260,13 @@ periodics:
     ci.openshift.io/generator: prowgen
     pj-rehearse.openshift.io/can-be-rehearsed: "true"
   name: periodic-ci-redhat-developer-rhdh-main-e2e-aks-operator-nightly
+  reporter_config:
+    slack:
+      channel: '#rhdh-e2e-test-alerts'
+      job_states_to_report:
+      - error
+      report_template: '<!subteam^S07BMJ56R8S> <@U08UP0REWG1> :failed: `{{.Spec.Job}}`
+        ended with *{{.Status.State}}*, :scroll: <{{.Status.URL}}| logs>'
   spec:
     containers:
     - args:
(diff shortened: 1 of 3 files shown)


ci-operator/jobs/redhat-developer/rhdh/redhat-developer-rhdh-main-periodics.yaml
```diff
@@ -193,6 +193,13 @@ periodics:
     ci.openshift.io/generator: prowgen
     pj-rehearse.openshift.io/can-be-rehearsed: "true"
   name: periodic-ci-redhat-developer-rhdh-main-e2e-aks-helm-nightly
+  reporter_config:
+    slack:
+      channel: '#rhdh-e2e-test-alerts'
+      job_states_to_report:
+      - error
```
subhashkhileri, line 200: will it only report the failures?

zdrapela: This should report only failures caused by OpenShift CI. In that case, the new fine-grained reporting won't have a chance to send any report. This should be like a backup

subhashkhileri: since you already have these : 
  job_name_patterns:
  - "e2e-.*-helm-nightly$"
  - "e2e-.*-operator-nightly$"
  
  in .config.prowgen, does it still require a manully defination here as well?

zdrapela: It works like this:
1. you configure it in `.config.prowgen`
2. you create a new config without the `ci-operator/config/redhat-developer/rhdh/redhat-developer-rhdh-release-1.9.yaml` for example
3. you run `make jobs`
4. the job in `ci-operator/jobs/redhat-developer/rhdh/redhat-developer-rhdh-release-1.9-presubmits.yaml` is generated **with** the `reporter_config`.

So it's generated, not manually added now.

**Without** the `.config.prowgen`, you do steps 2. and 3. and you get the job generated **without** the `reporter_config`.

subhashkhileri: Okay

subhashkhileri: Nightly test Alerts app also reports right, I am thinking if it would not duplicate the alerts for same job.

zdrapela: Nightly test Alerts app works only if the test goes to the post step `redhat-developer-rhdh-send-alert`. To my knowledge, the state error should mostly happen when the job couldn't run and the post step is not executed. Without this, we sometimes didn't get any message when the job failed because of OpenShift CI issues.
Sometimes we might get two reports, but I think it should be rare.

zdrapela: If the test just fails, it should end up in state failed.

PR #70368, opened 2025-10-16
trt-2255: remove skips for optional jobs

neisw:
Cleanup 2 remaining `always_run: false` jobs with `skip_if_only_changed` enabled presubmits.

@petr-muller tells me I only have to update main and 4.21 && 4.22 will propagate automatically


ci-operator/jobs/openshift/origin/openshift-origin-main-presubmits.yaml
```diff
@@ -145,7 +144,7 @@ presubmits:
       - name: result-aggregator
         secret:
           secretName: result-aggregator
-    trigger: (?m)^/test( | .* )e2e-aws-csi,?($|\s.*)
+    trigger: (?m)^/test( | .* )(e2e-aws-csi|remaining-required),?($|\s.*)
```
smg247, line 147: I think you are unintentionally enabling the manual 2-stage pipelines here.

PR #70459, opened 2025-10-20
run ACSCS e2e tests against ROSA HCP again

johannes94:
We changed the ACSCS e2e tests to run against infra OCP again, because we had consistent test failures with ROSA HCP due to DNS issues.

This PR reverts that change of #70220, once a rehearse succeed we should merge it back to profit from the shorter creation times of ROSA HCP again.


johannes94:
/pj-rehearse

Failed because of an unrelated issue retrying.

PR #70464, opened 2025-10-20
Always run false

paigerube14:
Want to turn off always run: true to false to be able to push directly to main on benchmark-runner versioning steps

### DIFF
--- a/ci-operator/jobs/redhat-performance/benchmark-runner/redhat-performance-benchmark-runner-main-postsubmits.yaml
+++ b/ci-operator/jobs/redhat-performance/benchmark-runner/redhat-performance-benchmark-runner-main-postsubmits.yaml
@@ -1,7 +1,7 @@
 postsubmits:
   redhat-performance/benchmark-runner:
   - agent: kubernetes
-    always_run: true
+    always_run: false
     branches:
     - ^main$
     cluster: build05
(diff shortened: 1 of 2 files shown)


ci-operator/jobs/redhat-performance/benchmark-runner/redhat-performance-benchmark-runner-main-postsubmits.yaml
```diff
@@ -1,7 +1,7 @@
 postsubmits:
   redhat-performance/benchmark-runner:
   - agent: kubernetes
-    always_run: true
+    always_run: false
```
liangxia, line 4: It's suggest to make the change in config (ci-operator/config/org_name/repo_name), rather than in jobs.

paigerube14: @liangxia do you know how I could edit my config for that value to be false instead of setting it in the job itself? 

config: https://github.com/openshift/release/blob/master/ci-operator/config/redhat-performance/benchmark-runner/redhat-performance-benchmark-runner-main.yaml

liangxia: Just found this is presubmit/postsubmit jobs.  In this case, you can remove the whole file `ci-operator/config/redhat-performance/benchmark-runner/redhat-performance-benchmark-runner-main.yaml` since there are no others tests defined in it.

PR #70469, opened 2025-10-20
WINC-1420: winc: add optional AWS test with WinServer2025

jrvaldes:
this PR add tests to test in 
- https://github.com/openshift/windows-machine-config-operator/pull/3459

### DIFF
--- a/ci-operator/config/openshift/windows-machine-config-operator/openshift-windows-machine-config-operator-master.yaml
+++ b/ci-operator/config/openshift/windows-machine-config-operator/openshift-windows-machine-config-operator-master.yaml
@@ -103,6 +103,20 @@ tests:
     workflow: ipi-aws-ovn-hybrid
   timeout: 5h0m0s
 - always_run: false
+  as: aws-e2e-operator-windows-server-2025
+  optional: true
+  skip_if_only_changed: ^(?:docs|\.github|\.tekton)/|\.md$|^(?:\.gitignore|renovate\.json|OWNERS|PROJECT|LICENSE|Containerfile|Containerfile.bundle)$
+  steps:
+    cluster_profile: aws-2
+    env:
+      COMPUTE_NODE_REPLICAS: "1"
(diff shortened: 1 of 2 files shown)


jrvaldes:
ci/prow/ci-operator-config-metadata job  failures are not related.

ci-operator/config/openshift/windows-machine-config-operator/openshift-windows-machine-config-operator-master.yaml
```diff
@@ -102,6 +102,20 @@ tests:
     - ref: windows-e2e-operator-test
     workflow: ipi-aws-ovn-hybrid
   timeout: 5h0m0s
+- always_run: false
+  as: aws-e2e-operator-windows-server-2025
+  optional: true
+  skip_if_only_changed: ^(?:docs|\.github|\.tekton)/|\.md$|^(?:\.gitignore|renovate\.json|OWNERS|PROJECT|LICENSE|Containerfile|Containerfile.bundle)$
```
sebsoto, line 108: I think this should be removed, otherwise it will still run on PRs as normal.
I may be misremembering, but there shouldn't be a downside to removing it while this remains optional.

jrvaldes: `always_run: false + skip_if_only_changed + optional` is a supported and allowed combination

Behavior:
Test only runs manually (/test aws-e2e-operator-windows-server-2025) 
Test skips if only docs/config files changed
Test won't block merges if it fails




`run_if_changed`, `skip_if_only_changed`, and `always_run: true` are the ones mutually exclusive, not this case.

sebsoto: Okay, if you are confident this will not run unless specifically asked to, thats fine with me

PR #70487, opened 2025-10-21
supplemental-images: delete unused IS

Prucek:
I tried to look for owners, but they are no longer in Red Hat.
- `ibmz-eco` org no longer exists or is private
- `opm-builder` not many usages, if someone is missing a new image, they need to add it to the mirroring config
- `redhat-operator-index` apparently added by us, so again, if someone is missing a new image, they need to add it to the mirroring config
- `openshift/release` - https://redhat-internal.slack.com/archives/CB95J6R4N/p1759534820435319?thread_ts=1759221171.768279&cid=CB95J6R4N


deepsm007:
/lgtm
/hold
We will remove hold on Thursday UTC morning time

PR #70489, opened 2025-10-21
telco: automerge trusted k5x PRs for numaresources-operator and scheduler

shajmakh:
Avoid waiting for prow CI and configure Tide to auto merge PRs by
konflux bot that are labeled with `auto-merge`.

### DIFF
--- a/core-services/prow/02_config/openshift-kni/scheduler-plugins/_prowconfig.yaml
+++ b/core-services/prow/02_config/openshift-kni/scheduler-plugins/_prowconfig.yaml
@@ -1,5 +1,25 @@
 tide:
   queries:
+  - author: red-hat-konflux
+    labels:
+    - auto-merge
+    missingLabels:
+    - do-not-merge
+    - do-not-merge/hold
+    - do-not-merge/work-in-progress
+    - needs-rebase
+    repos:
+    - openshift-kni/scheduler-plugins
+  - author: red-hat-konflux[bot]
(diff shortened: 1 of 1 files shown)


ffromani (commented):
/lgtm

for numaresources-operator and scheduler-plugins

shajmakh:
TLDR; this change won't do the goal because branch protection will be applied on the branch without considering the author or the queries.
Explanation for future reference: 
 we want to immediately merge bot-trusted PR for konflux updates, which means also skipping any CI job. (first 2 queries)
 OTOH all other PRs need to submit to the CI configuration and require "lgtm" and "approve" labels. (last query)
 In reality both PR types are meant to target the main/master branch. The problem lies in that tide's branch protection rules takes affect per branch and not per query, so checking the rules under https://github.com/kubernetes-sigs/prow/blob/main/pkg/config/prow-config-documented.yaml it is not possible to achieve the goal of automerging bot PRs with no policy protection and at the same time to protect the branch upon non-bot PRs.

possible option to check forward could be:
make Konflux CI optional for specific PRs (bot ones) - somewhere [here](https://github.com/openshift/release/pull/62780/files) perhaps- && ensure all jobs in https://github.com/openshift/release/blob/master/ci-operator/config/openshift-kni/numaresources-operator/openshift-kni-numaresources-operator-main.yaml and all affected branches are optional/skipped for such PRs.

/close

PR #70495, opened 2025-10-21
Block build04 to bump up os image of control plane node

liangxia:


liangxia:
/hold
so that I can work on it tomorrow.

PR #70499, opened 2025-10-21
CMP-3766: Adjust periodic tests frequencies for older OCP versions

Anna-Koudelkova:
1. Checked all our periodic jobs (for CO, SPO and FIO) running on OCP versions 4.12 - 4.15 (4.15 included)
2. Adjust jobs that were not running with frequency f60 + generate new cronjob entry for these jobs

Only jobs in OCP v 4.14 for CO (on amd and arm) and SPO (on amd) were affected, the rest was already wired to be triggered every 60 days.

Jira ticket: https://issues.redhat.com/browse/CMP-3766


xiaojiey:
@liangxia Could you please help to review/merge? Thanks.

PR #70504, opened 2025-10-21
Update network flow matrix nightly lanes cron

shirmoran:
In order to investigate nightly failures in network flow matrix tests, we updated the job to run twice a day.

### DIFF
--- a/ci-operator/config/openshift/release/openshift-release-master__nightly-4.16.yaml
+++ b/ci-operator/config/openshift/release/openshift-release-master__nightly-4.16.yaml
@@ -725,7 +725,7 @@ tests:
     workflow: telco5g-ptp
 - as: telco5g-network-flow-matrix-bm
   cluster: build05
-  cron: 0 23 * * 0,3
+  cron: 0 23 * * *
   restrict_network_access: false
   steps:
     env:
@@ -752,7 +752,7 @@ tests:
     workflow: telco5g-network-flow-matrix-bm
 - as: telco5g-network-flow-matrix-single-node-bm
   cluster: build05
-  cron: 0 23 * * 0,3
+  cron: 0 23 * * *
   restrict_network_access: false
   steps:
     env:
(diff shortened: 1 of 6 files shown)


neisw:
Thanks @shirmoran.  The reason I am asking is that for the older releases the changes come in less frequently.  I wouldn't expect you to need to run the tests against the same payload repeatedly to identify a failure in this case.  We are trying to balance the number of jobs we run and focus heavier testing on the development releases.   It doesn't seem like you need to run the older releases twice daily in this case.

Who monitors these jobs when there are failures?

neisw:
If you are planning to have alerting setup for these jobs and monitor the health of them then daily sounds good.  Having alerting and monitoring setup is a requirement for becoming informing so that would align well.

ci-operator/config/openshift/release/openshift-release-master__nightly-4.16.yaml
```diff
@@ -725,7 +725,7 @@ tests:
     workflow: telco5g-ptp
 - as: telco5g-network-flow-matrix-bm
   cluster: build05
-  cron: 0 23 * * 0,3
+  cron: 0 23 * * *
```
neisw, line 728: It looks like you are going to kick off 6 jobs each night at 11 across all releases.  In the past I know some of the telco resources were limited.  Are you sure they have capacity for this, have you considered staggering the schedule across releases?

ci-operator/config/openshift/release/openshift-release-master__nightly-4.19.yaml
```diff
@@ -969,7 +969,7 @@ tests:
     workflow: telco5g-mno-ztp-cnf
 - as: telco5g-network-flow-matrix-bm
   cluster: build05
-  cron: 0 23 * * 2,5
+  cron: 0 11,23 * * *
```
neisw: I'm still a bit hesitant about running these twice a day.  I think where you really want to get to is running these on each payload to catch issues as they come in.  I'm less concerned about running them 2 times a day now but more just letting them keep running on this schedule a year from now, etc.  I feel like you would still catch your issues running once a day.

shirmoran: Okay, let's start with once a day and if needed we'll consider increasing the runs.

PR #70225, opened 2025-10-13
RHDH hardcode platform env vars & establish OCP 4.20

zdrapela:
https://issues.redhat.com/browse/RHDHBUGS-2129

This hardcodes the `IS_OPENSHIFT` and `CONTAINER_PLATFORM` variables to the jobs themselves. The identification of OCP vs K8s dynamically was not reliable, and the `CONTAINER_PLATFORM` had to be decided from the job name.

The `CONTAINER_PLATFORM_VERSION` variable is moved into the jobs too, but is parsed from the `oc`/`kubectl` output.

Move the GKE cluster login, initial setup, and service account creation to the job itself to align with other jobs (which use ephemeral clusters).

Unify CI job configuration and structure it into blocks.

Fix getting the `STATUS_URL_REPORTPORTAL` from `droute`.

This PR will be followed by another one in https://github.com/redhat-developer/rhdh repo.

PR #70527, opened 2025-10-21
RHIDP-9038 - Refactor CI configuration for Red Hat Developer RHDH

gustavolira:
- Updated periodic jobs to include cleanup tasks for AKS and EKS clusters.
- Introduced new periodic jobs for OSD GCP cleanup and operator tasks.
- Refactored existing presubmit jobs to align with new naming conventions and targets.
- Enhanced job specifications with improved resource requests and volume mounts.
- Adjusted cron schedules for various jobs to optimize execution times.

### DIFF
--- a/ci-operator/config/redhat-developer/rhdh/redhat-developer-rhdh-main.yaml
+++ b/ci-operator/config/redhat-developer/rhdh/redhat-developer-rhdh-main.yaml
@@ -11,7 +11,7 @@ releases:
       channel: fast
       version: "4.15"
 resources:
-  '*':
+  "*":
     limits:
       cpu: "2"
       memory: 5Gi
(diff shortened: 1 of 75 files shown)


zdrapela (commented):
If you do want to go this way, I suggest setting the new jobs cron to different time slots than the existing ones. Maybe at 3 am or so. Too many jobs at the same time would eat up the whole cluster pool and cause rate limiting.

Cleanup jobs for EKS and AKS now have the purpose of removing the hanging spot clusters. The current clean up that I see you created would increase the run time of the job without a real benefit.

For configuring owners, you can use symlinks, that will be automatically created by running:
```
hack/dedup-owners.sh ci-operator/step-registry/redhat-developer
```
See https://github.com/openshift/release/pull/70221 for context.

There were updates to the CI scripts to hardcode the platform and move all K8s cluster setup to this repo to align with OCP jobs. Please see this PR https://github.com/openshift/release/pull/70225 and possibly update to reflect those changes.

PR #70540, opened 2025-10-21
Using generic oc cli image to mirror

hector-vido:
The `buildConfig` that creates the image was removed.
The script will be downloaded from git instead of reside inside the image.
The custom image will be replaced for the official openshift 4 cli one.
The image will only be downloaded if not present, since there is no special tooling inside it.


deepsm007:
/pj-rehearse ack
@hector-vido will monitor the results once this merges. It eliminates dependency of mirroring image which embeds the script and one less image to maintain

PR #70554, opened 2025-10-22
Remove useless 4.19&4.20 GCP custom-dns periodic jobs

gpei:
The GCP custom-dns feature will be GA in 4.21, so dropping the 4.19&4.20 e2e-gcp-custom-dns-techpreview periodic jobs.


sadasu:
/hold

I have a question about whether we need the techpreview job for 4.20.

PR #70644, opened 2025-10-24
remove some OWNERS for stackrox/acs-fleet-manager

johannes94:
Removing people who are not working in the CS team from the OWNERS file for stackrox/acs-fleet-manager CI jobs.


mclasmeier (approved):
Fine with me, but note that @porridge isn't working today.

PR #70664, opened 2025-10-24
NO-JIRA: CVO: Specify a concrete golangci-lint version for main

DavidHurta:
Specify a concrete golangci-lint version in main so that the CVO PRs will not start randomly failing during development when new versions are rolled out.

I do not think that it is necessary to address lint issues that urgently.


DavidHurta:
/pj-rehearse pull-ci-openshift-cluster-version-operator-main-lint

PR #70750, opened 2025-10-28
Add periodic CI configuration for MetalLB 4.20

karampok:
This adds periodic CI testing for MetalLB on OCP 4.20 release on bare metal with Slack notifications to #team-metallb-ci-periodic channel.

The periodic job runs every 72 hours on the equinix-ocp-metal cluster with IPv4 stack, OVNKubernetes network type, and frr-k8s-cno BGP type.

Assisted-By: Claude <noreply@anthropic.com>


fedepaol:
does it make sense to add for 4.21 too?

karampok:
not sure if we should add -main or 4.21, but yes it makes sense.
But my previous effort to add 4.20 periodics never passed CI, so step by step 

https://github.com/openshift/release/pull/67284

PR #70756, opened 2025-10-28
Add periodic OLMv1 job for observability-operator on OCP 4.21 TechPreview

danielmellado:
This PR adds a new step for OLMv1 installation and creates periodic
jobs to test that on 4.21 nightly with TPNU (Tech Preview No Upgrade)

Signed-off-by: Daniel Mellado <dmellado@fedoraproject.org>


lihongyan1:
code lgtm, may better give a specific image if no parameter is passed in and run prerehearse.

PR #70759, opened 2025-10-28
Ignore exit code 3 with Orion

jtaleric:
This just means there was no data found with the query. This shouldn't cause failures.

### DIFF
--- a/ci-operator/step-registry/openshift-qe/orion/openshift-qe-orion-commands.sh
+++ b/ci-operator/step-registry/openshift-qe/orion/openshift-qe-orion-commands.sh
@@ -117,7 +117,7 @@ set -e
 
 cp *.csv *.xml *.json *.txt "${ARTIFACT_DIR}/" 2>/dev/null || true
 
-if [ $orion_exit_status -eq 3]; then
+if [ $orion_exit_status -eq 3 ]; then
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/openshift-qe/orion/openshift-qe-orion-commands.sh
```diff
@@ -117,4 +117,10 @@ set -e
 
 cp *.csv *.xml *.json *.txt "${ARTIFACT_DIR}/" 2>/dev/null || true
 
+if $orion_exit_status -eq 3; then
```
rsevilla87: Missing square brackets?

```suggestion
if [ $orion_exit_status -eq 3]; then
```

ci-operator/step-registry/openshift-qe/orion/openshift-qe-orion-commands.sh
```diff
@@ -117,4 +117,10 @@ set -e
 
 cp *.csv *.xml *.json *.txt "${ARTIFACT_DIR}/" 2>/dev/null || true
 
+if [ $orion_exit_status -eq 3]; then
```
rsevilla87:
```suggestion
if [ $orion_exit_status -eq 3 ]; then
```

chentex (approved):
/lgtm

PR #70792, opened 2025-10-29
use westus for azure so we launch arm cluster

anpingli:
use westus as there isn't arm profile in centralus
az vm list-usage --location centralus  --output table  |grep DPS

set the number to 3, as the Limit 350 can only allow about 3 clusters.
```
az vm list-usage --location westus --output table |grep DPS
Standard DPSv5 Family vCPUs               0               350
Standard Dpsv6 Family vCPUs               0               350
```


anpingli:
@liangxia  can you approve it?

PR #70805, opened 2025-10-29
OCPBUGS-63622: (hot fix) Change to use TestExternalOIDC test suite

xiuwang:
Update 4.20 HyperShift periodic test configuration to use the stable TestExternalOIDC test suite instead of TestExternalOIDCTechPreview.


xingxingxia:
@jparrill , could you help approve? It is just a one line update but a hot fix. Thanks!

xingxingxia:
@csrwng , could you PTAL / approve?

PR #70810, opened 2025-10-29
No managed chaos

paigerube14:
Removing managed chaos tests

### DIFF
--- a/ci-operator/config/redhat-chaos/prow-scripts/redhat-chaos-prow-scripts-main__rosa-4.17-nightly.yaml
+++ b/ci-operator/config/redhat-chaos/prow-scripts/redhat-chaos-prow-scripts-main__rosa-4.17-nightly.yaml
@@ -46,8 +46,8 @@ resources:
       cpu: 100m
       memory: 200Mi
 tests:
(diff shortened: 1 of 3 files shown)


ci-operator/config/redhat-chaos/prow-scripts/redhat-chaos-prow-scripts-main__rosa-4.15-nightly.yaml
```diff
@@ -1,192 +0,0 @@
-base_images:
```
chaitanyaenr, line 1: @paigerube14 maybe we should disable the job runs by setting run_always to false instead to be able to enable them back in the future. Thoughts?

paigerube14: yes, think that will be better. Good catch

shahsahil264: +1

PR #70811, opened 2025-10-29
Add vSphere and Azure tech preview, disk setup job

jcpowermac:
This commit adds two additional jobs
to test the disk setup e2e before merging.

### DIFF
--- a/ci-operator/jobs/openshift/machine-config-operator/openshift-machine-config-operator-main-presubmits.yaml
+++ b/ci-operator/jobs/openshift/machine-config-operator/openshift-machine-config-operator-main-presubmits.yaml
@@ -1538,6 +1538,81 @@ presubmits:
     - ^main$
     - ^main-
     cluster: build10
+    context: ci/prow/e2e-azure-ovn-multidisk-techpreview
+    decorate: true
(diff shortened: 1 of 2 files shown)


ci-operator/jobs/openshift/machine-config-operator/openshift-machine-config-operator-main-presubmits.yaml
```diff
@@ -3892,6 +3967,81 @@ presubmits:
         secret:
           secretName: result-aggregator
     trigger: (?m)^/test( | .* )e2e-vsphere,?($|\s.*)
+  - agent: kubernetes
+    always_run: false
+    branches:
+    - ^main$
+    - ^main-
+    cluster: vsphere02
```
yuqi-zhang, line 3975: (question for my own understanding) I didn't know we had a vsphere02 cluster. Is there a specific reason this gets autogenerated as the default vsphere job cluster? I was under the impression that our main buildxx clusters spin up resources in the different clouds based on the profile and is itself platform agnostic

yuqi-zhang (approved):
Generally seems fine to me, just had a question

PR #70624, opened 2025-10-23
MPIIT: Add ACS cr job for OCP 4.21

oharan2:

PR #70831, opened 2025-10-29
[INTEROP-8523] Add stable ACS Component readiness config for OCP 4.21

sg-rh:
ACS interop update:
1. Added stable version of ACS
2. This job will be running compliance tests


ci-operator/config/stackrox/stackrox/stackrox-stackrox-release-4.9__ocp-4-21-lp-interop-cr.yaml
```diff
@@ -0,0 +1,54 @@
...
+            [
+              {"step": "stackrox-*", "failure_type": "all", "classification": "ACS Test Failure", "jira_assignee": "!default", "jira_additional_labels": ["!default","interop-tests"]}
+            ]
+        }
+      FIREWATCH_DEFAULT_JIRA_ADDITIONAL_LABELS: '["4.21-lp","self-managed-lp","acs-lp"]'
+      FIREWATCH_DEFAULT_JIRA_ASSIGNEE: dahouse@redhat.com
+      FIREWATCH_DEFAULT_JIRA_PROJECT: ROX
+      MAP_TESTS: "true"
+      OCP_VERSION: "4.21"
+      REPORTPORTAL_CMP: ACS-lp-interop
```
davdhacs, line 44:
```suggestion
      REPORTPORTAL_CMP: ACS-lp-interop
      TEST_SUITE: ocp-compliance-e2e-tests
```

davdhacs: Could we add the test name here and keep the default? The default is used by the weekly test run (and you could add the env var for that one instead of adding it here. but it will need changed if the default is changed).

davdhacs (changes requested):
looks great! One change, or need to specify the "ocp-qa-e2e-tests" for the weekly test run instead.

ci-operator/config/stackrox/stackrox/stackrox-stackrox-release-4.9__ocp-4-21-lp-interop-cr.yaml
sg-rh: @davdhacs we can use the test name here, but as we mentioned there is not going to be the weekly test run. The default test suite can run in other Job ([PR](https://github.com/openshift/release/pull/70624))

sg-rh:
/pj-rehearse periodic-ci-stackrox-stackrox-release-4.9-ocp-4-21-lp-interop-cr-acs-tests-aws

sg-rh:
/pj-rehearse periodic-ci-stackrox-stackrox-release-4.9-ocp-4-21-lp-interop-cr-acs-tests-aws

sg-rh:
/pj-rehearse periodic-ci-stackrox-stackrox-release-4.9-ocp-4-21-lp-interop-cr-acs-tests-aws

davdhacs:
/pj-rehearse periodic-ci-stackrox-stackrox-release-4.9-ocp-4-21-lp-interop-cr-acs-tests-aws

sg-rh:
/pj-rehearse periodic-ci-stackrox-stackrox-release-4.9-ocp-4-21-lp-interop-cr-acs-tests-aws

ci-operator/step-registry/stackrox/compliance-e2e/OWNERS
etirta, line 1: Is this file based on the parent's OWNER file? If so, please follow the Best Practice to create a link pointing to the parent's OWNER file instead.

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
```diff
@@ -0,0 +1,100 @@
...
+# This ensures the script exits immediately if any command fails, including
+# commands in pipelines and sub-shells. xtrace helps with post-mortem analysis.
+set -euxo pipefail; shopt -s inherit_errexit
+
+# Define global variables with defaults using typeset
+# ARTIFACT_DIR and SHARED_DIR are injected by CI, but we provide defaults for safety
+typeset artifactDir="${ARTIFACT_DIR:-/tmp/artifacts}"
+typeset sharedDir="${SHARED_DIR:-/tmp/shared}"
+typeset jobNameSafe="${JOB_NAME_SAFE:-}"
+typeset testSuite="${TEST_SUITE:-}"
```
etirta: Some of these env. var. are guaranteed exist in CI Operator and other are configured in the `Step`, so those are also guaranteed set. So there is no reason to assign default value.

Also, any particular reason why we create the mirror shell var. instead of just use the env. var. directly?

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
```diff
@@ -0,0 +1,100 @@
...
+
+    # Verify installation
+    # Using || true to prevent script exit if yq is not available
+    typeset cmdYq=""
+    cmdYq="$(/tmp/bin/yq --version || true)"
+
+    if [ -n "${cmdYq}" ]; then
+        : "yq installed: ${cmdYq}"
+    else
+        export MAP_TESTS="false"
```
etirta: I'm can't figure out the use of this env. var. `MAP_TESTS`, but in top the mirror shell var. is created and used all over the place in this scripts. So changing the env. var. in here without synching it to the corresponding mirror shell var., is this intentional?

This illustrate the danger of using mirror shell var., instead of just use the env. var. directly. Generally speaking, it's much easier and less prone to error if we just use single source of truth, hence it's recommended to use the env. var. directly instead of create the mirror shell var. Unless there is strong reason why this approach is used???

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
```diff
@@ -0,0 +1,100 @@
...
+
+        # Send modified files to shared dir
+        cp -r "${artifactDir}"/junit-* "${sharedDir}" || : "Warning: couldn't copy files to SHARED_DIR"
+    fi
+
+    true
+}
+
+# Set trap for cleanup on exit
+trap 'CleanupCollect' SIGINT SIGTERM ERR EXIT
```
etirta: trapping the `EXIT` should suffice as it basically cover all exit scenario.

```suggestion
trap 'CleanupCollect' EXIT
```

etirta (changes requested):

ci-operator/config/stackrox/stackrox/stackrox-stackrox-release-4.9__ocp-4-21-lp-interop-cr.yaml
```diff
@@ -0,0 +1,55 @@
...
+    candidate:
+      product: ocp
+      stream: nightly
+      version: "4.21"
+resources:
+  '*':
+    requests:
+      cpu: "8"
+      memory: 8000Mi
+test_binary_build_commands: .openshift-ci/dispatch.sh test-binary-build-commands
```
oharan2, line 22: Is this something we need to also configure in the latest job?

ci-operator/config/stackrox/stackrox/stackrox-stackrox-release-4.9__ocp-4-21-lp-interop-cr.yaml
```diff
@@ -0,0 +1,55 @@
...
+            [
+              {"step": "stackrox-*", "failure_type": "all", "classification": "ACS Test Failure", "jira_assignee": "!default", "jira_additional_labels": ["!default","interop-tests"]}
+            ]
+        }
+      FIREWATCH_DEFAULT_JIRA_ADDITIONAL_LABELS: '["4.21-lp","self-managed-lp","acs-lp"]'
+      FIREWATCH_DEFAULT_JIRA_ASSIGNEE: dahouse@redhat.com
+      FIREWATCH_DEFAULT_JIRA_PROJECT: ROX
+      MAP_TESTS: "true"
+      OCP_VERSION: "4.21"
+      REPORTPORTAL_CMP: ACS-lp-interop
```
oharan2, line 44: Should be something like `ACS-Stable-lp-interop`. Do you have this thread where we covered this discussion? I can check about that

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
```diff
@@ -0,0 +1,100 @@
...
+    if [ ! -d "stackrox" ]; then
+        git clone https://github.com/stackrox/stackrox.git
+    fi
+    cd stackrox
+fi
+
+# Execute dispatch script
+.openshift-ci/dispatch.sh "${job}"
+
+true
```
oharan2: What is this true stands for?

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
```diff
@@ -0,0 +1,100 @@
...
+# commands in pipelines and sub-shells. xtrace helps with post-mortem analysis.
+set -euxo pipefail; shopt -s inherit_errexit
+
+# Define global variables with defaults using typeset
+# ARTIFACT_DIR and SHARED_DIR are injected by CI, but we provide defaults for safety
+typeset artifactDir="${ARTIFACT_DIR:-/tmp/artifacts}"
+typeset sharedDir="${SHARED_DIR:-/tmp/shared}"
+typeset jobNameSafe="${JOB_NAME_SAFE:-}"
+typeset testSuite="${TEST_SUITE:-}"
+typeset mapTests="${MAP_TESTS:-true}"
```
oharan2: The default value should be false (already set in the step ref file), since we should address this script as a non CR related test step as a default, only as an additional layered condition. 
It is assumed that it will also be used for other platform-related jobs.

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
sg-rh: I agree, I overlooked these changes to env. var. done via agent. I will review the script again and update

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
sg-rh: it was to cover all cases, but yes `EXIT` is sufficient

ci-operator/config/stackrox/stackrox/stackrox-stackrox-release-4.9__ocp-4-21-lp-interop-cr.yaml
sg-rh: This has been there for all ACS configs, likely used to build test binary, dependencies, test env env.

ci-operator/config/stackrox/stackrox/stackrox-stackrox-release-4.9__ocp-4-21-lp-interop-cr.yaml
sg-rh: I dont think there is slack thread since there were meetings and docs, I can send you that doc for review.
But basically, existing `ACSLatest-lp-interop` is running as PoC because it runs latest version of ACS. I have not distinguished or renamed this job like because it's like other interop jobs we run.

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
etirta: It's Shell Best Practice when `errexit` is enabled. See [section 2.2](https://spaces.redhat.com/spaces/MPEXIENG/pages/728700228/Best+Practices#bash-shell-script) for explanation.

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
sg-rh: Here `MAP_TESTS `is a fallback if yq installation failed. I will actually do a check if yq exists before using it.

Now reviewing again, I had mirror shell var to avoid duplicating, agreed it still causes redundancy and some confusion so i will revert to using `MAP_TESTS `directly

ci-operator/config/stackrox/stackrox/stackrox-stackrox-release-4.9__ocp-4-21-lp-interop-cr.yaml
etirta: @oharan2 The `test_binary_build_commands` is to build test executables. Some test is created using program language that need to be compiled, instead of using scripting language.

ci-operator/config/stackrox/stackrox/stackrox-stackrox-release-4.9__ocp-4-21-lp-interop-cr.yaml
etirta, line 1: Please use Component Readiness (CR) Job Naming convention, we need `-aws-` somewhere in the variant portion of this file name, so the CR can set the `platform` parameter properly.

ci-operator/step-registry/stackrox/compliance-e2e/OWNERS
etirta: @sg-rh This file should be a symbolic link pointing to `../OWNERS`, not a text file.

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
```diff
@@ -0,0 +1,105 @@
...
+    # Verify installation
+    # Using || true to prevent script exit if yq is not available
+    typeset cmdYq=""
+    cmdYq="$(/tmp/bin/yq --version || true)"
+
+    if [ -n "${cmdYq}" ]; then
+        : "yq installed: ${cmdYq}"
+    else
+        : "Warning: yq installation failed, test mapping will be skipped"
+    fi
```
etirta: Can we use a more simplified statement? Having 10 lines just to verify installation is just overkill.

```suggestion
    /tmp/bin/yq --version || MAP_TESTS=false    # Skip test mapping if yq installation failed.        
```

First of all, if the installation failed due to call to `curl` failed, the script will stop before executing this. This will be the most possible outcome. And generally speaking, if this occurs we don't want to continue anyway because skipping test mapping essentially have negative effect towards Component Readiness. I personally would prefer to exit out.

Second, if it fails due to any other reason and we would like to continue the test by disabling the test mapping (knowing it will have negative effect towards CR), then we should explicitly set the `MAP_TESTS` to `false`, because this env. var. can be set to `true` by the `Job` Configuration, so we should not rely on the default value set by the `Step` Configuration.

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
```diff
@@ -0,0 +1,105 @@
...
+
+    true
+}
+
+function MapTestsForComponentReadiness() {
+    typeset resultsFile="${1:-}"
+
+    if [[ "${MAP_TESTS}" == "true" && -n "${resultsFile}" ]]; then
+        if [ -f "${resultsFile}" ]; then
+            # Check if yq is available before attempting to use it
```
etirta: Why we check for `yq` again in here? We already called `InstallYq` at this point. Please try to avoid duplicated check.

etirta (changes requested):

ci-operator/config/stackrox/stackrox/stackrox-stackrox-release-4.9__ocp-4-21-lp-interop-cr.yaml
sg-rh: Thanks I have updated the filename

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
sg-rh: Thanks for suggestion, i have updated it

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
sg-rh: okay yes its updated now

ci-operator/config/stackrox/stackrox/stackrox-stackrox-release-4.9__ocp-4-21-lp-interop-cr.yaml
etirta: @sg-rh I'm really sorry that I gave you bad information. I misunderstood Forrest statement on how CR set the `platform` parameter. It turned out that CR use the Job Name. The Job Name is constructed from the Job Configuration file name PLUS test name (from `.test[].as`).

So it's enough to have `-aws` either in the file name OR in the test name.

In your case the test name actually is `acs-tests-aws`, so it's already have `-aws-` in it and this will propagate properly to the Job Name.

I was wrong asking you to change the name, because now there is 2 `aws` words in the Job Name. Sorry for extra work. Please choose one, either have it in the file name or in the test name, but not both. Looking at other Job, I would suggest be consistent with other and have the `aws` in the test name.

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
```diff
@@ -0,0 +1,93 @@
...
+function CleanupCollect() {
+    # Disable exit on error in cleanup to be resilient to timeouts/interruptions
+    # This ensures cleanup completes even if individual operations fail
+    set +e
+    
+    if [[ "${MAP_TESTS}" == "true" ]]; then
+        InstallYq || : "Warning: yq installation failed, skipping test mapping"
+
+        typeset originalResults="${ARTIFACT_DIR}/original_results"
+        mkdir -p "${originalResults}" 2>/dev/null || true
```
etirta: Do you have any justification on silencing the potential error?

Generally speaking we shouldn't to this. If it fails, we want to know and fail.

Or better question, why we are preserving original result to begin with. I wonder about this but I don't think it's a big deal to preserve it, but since now you introduce this silencing in the last commit, make me wonder whether this is actually necessary or just waste of storage. Do you know the retention policy from DPP for CI Operator log? How long it's being kept?

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
```diff
@@ -0,0 +1,93 @@
...
+
+        # Keep a copy of all the original JUnit files
+        cp -r "${ARTIFACT_DIR}"/junit-* "${originalResults}" 2>/dev/null || : "Warning: couldn't copy original files"
+
+        # First, copy all XML files to SHARED_DIR (do this before yq processing)
+        # Process files in a single pass: copy first, then modify with yq
+        typeset resultFile
+        while IFS= read -r -d '' resultFile; do
+            # Copy file to SHARED_DIR
+            cp -- "${resultFile}" "${SHARED_DIR}/$(basename "${resultFile}")" 2>/dev/null || : "Warning: couldn't copy ${resultFile} to SHARED_DIR"
```
etirta: Now it make me wonder if this is the best implementation. Every new file put in `SHARED_DIR`, a secret is created. The more secret is created, the longer the execution time is, as at the start of next `step` execution the CI Operator need to create the file again in that Container's `SHARED_DIR`.

What is this actually for? If there is a good reason doing this, please put comment so we know why this is done.

Another thing to consider:
 1. How many files do we expect to be copied over in this loop? May be compress and packing those files into a single compressed file is better to reduce the no. of files under `SHARED_DIR`?
 2. Have we consider on possible name collision? The files put under `SHARED_DIR` will carry on to the next step. And currently we can't have sub-directory. So there is a potential file name collision in here.

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
sg-rh: I dont recall as I was following some other products like pipelines, oadp, quay for preserving the originals, but I believe, its for debugging & troubleshooting  if yq breaks structure @oharan2 ?

The retention is controlled via GCS Bucket rules, and I don't have access, but for periodic jobs I believe its 6 months to 1 year.

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
sg-rh: @etirta thanks for insight, yes I agree , with more files, there is time & performance impact
1.  I think this compliance tests generate 10-20 XML files, I can install & run `junitparser ` to merge them. Data Router expects XML files not compressed file, so merging might be better, which wil also eliminate name collisions. So it will create a single merged xml in SHARED_DIR.

2. I am using basename to avoid path conflicts, but name collisions are still possible if multiple files have same basename from different directories. Currrently, I would still need to handle collision with additioanal logic. Using merge can avoid that

sg-rh:
/pj-rehearse periodic-ci-stackrox-stackrox-release-4.9-ocp-4-21-lp-interop-aws-cr-acs-tests-aws

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
oharan2: Yes, I think at least one of the product teams asked us to keep the original results for debugging purposes. (Out mapping overrides their original test suite names)

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
oharan2: Thanks @sg-rh.
@etirta `What is this actually for?`: This is the only way to pass result files throughout the pipeline to the DR reporting container step, and this is the way the dno team told me is expected.

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
etirta: Do we know how the Prow Dashboard behave when there is jUnit XML files under sub-dir. of the Step's Artifact dir?

In my understanding, if there is jUnit XML files under Step's Artifact dir., it will be auto-collected. So if this auto-collection traverse down to sub-dir., then keeping the original jUnit XML files will have a negative consequences, by doubling the amount of TCs reported under this Step.

A potential workaround is to compress the original files into zip or tgz. file.

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
etirta: My point is can we modify the DR Reporting step to take the compress file instead.

There are 2 reasons to consider this, IMO:
 1. Potential file name collision (by file name, I do mean basename, not full path name).
 2. Degradation of execution performance when SHARE_DIR is populated with a lot of files.

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
```diff
@@ -0,0 +1,142 @@
...
+                ((fileCount++))
+            else
+                : "Warning: ${resultFile} does not appear to be a valid JUnit XML file, skipping" >&2
+            fi
+        done < <(find "${ARTIFACT_DIR}" -type f -iname "*.xml" -print0 2>/dev/null || true)
+        
+        if [[ ${fileCount} -eq 0 ]]; then
+            : "Warning: No XML files found to process for Data Router" >&2
+        else
+            typeset mergedFile="${SHARED_DIR}/junit-compliance-e2e-merged.xml"
```
etirta: Good solution to merge all into 1, make it more manageable.

However please use name that reduce the risk of collision with other step. My recommendation is use full relative path (to `step-registry`) by replacing `/` with `__` (the likely hood double-underscore appear as directory name is extremely low) of the `Step`.


```suggestion
            typeset mergedFile="${SHARED_DIR}/junit--stackrox__compliance-e2e__stackrox-compliance-e2e.xml"
```

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
oharan2: @etirta If Prow Dashboard means CR, then TCs aren't duplicated, since the original files aren't mapped and doesn't get picked up by the Component Matcher associating them into our board.
If there is such thing, can you share? Are we tracking it?

etirta: @oharan2 No, I meant like in https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/69455/rehearse-69455-pull-ci-redhat-chaos-lp-chaos-main-ocp4.21--edttj--poc-0-step-1/2016618651548913664 for example, where it listed `fail/Tot` and `pass/Tot`. I think the `Tot` is collecting all `xml` files under `artifacts` directories, including the ones under sub-dirs.

If this is true, then the `Tot` will be increased as some `Step` we copy over the original jUnit as well. Not really a big deal as we don't really do metric against these, but those metrics indeed shows in CR as well. So if oiginal is wanted, I think we should zip it and put the zipped file in the `artifact` directory instead.

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
```diff
@@ -0,0 +1,142 @@
...
+        # How it works:
+        # 1. Files are modified with yq to set test suite name (required for CR)
+        # 2. All XML files are merged into one file using junitparser
+        # 3. Merged file is copied to SHARED_DIR for Data Router Reporter step
+        #
+        # Why merge: Combining multiple XML files into one reduces Kubernetes secret overhead
+        # (each file becomes a secret), improving step transition performance and avoiding filename collisions
+        typeset resultFile
+        typeset fileCount=0
+        typeset xmlFiles=()
```
etirta: Please declare with the correct type. Even though shell is very forgiving, but it's good practice to delcare properly.


```suggestion
        typeset resultFile
        typeset -i fileCount=0
        typeset -a xmlFiles=()
```

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
etirta: @sg-rh Let's use `junit--stackrox__compliance-e2e__stackrox-compliance-e2e.xml` as file name. This is manual selection setting, but we can't wait until DPTP deliver the requested Env. Var. if even they agree to implement it. Thx.

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
sg-rh: ok i will update it

sg-rh:
/pj-rehearse periodic-ci-stackrox-stackrox-release-4.9-ocp-4-21-lp-interop-cr-acs-tests-aws

sg-rh:
/pj-rehearse periodic-ci-stackrox-stackrox-release-4.9-ocp-4-21-lp-interop-cr-acs-tests-aws

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
```diff
@@ -0,0 +1,82 @@
...
+    curl -L "https://github.com/mikefarah/yq/releases/latest/download/yq_linux_${arch}" \
+        -o /tmp/bin/yq && chmod +x /tmp/bin/yq
+    /tmp/bin/yq --version || export MAP_TESTS=false
+}
+
+# One JUnit XML for Data Router: collect XMLs, one yq eval-all (root/name/recalc), write and copy to SHARED_DIR.
+# All env (e.g. MAP_TESTS, MAP_TESTS_SUITE_NAME) must be defined in step ref or script.
+function CleanupCollect() {
+    typeset mergedFN="${1:-junit.xml}"
+    (($#)) && shift
```
etirta: I have a small suggestion for readability here. When parsing multiple parameters in sequence, grouping the assignment and the shift on a single line can help keep the block concise and easier to scan.

```suggestion
    typeset mergedFN="${1:-junit.xml}"; (($#)) && shift
```

Please consider the following scenario:
```bash
    typeset par1="${1:-}"
    (($#)) && shift
    typeset par2="${1:-}"
    (($#)) && shift
    typeset par3="${1:-}"
    (($#)) && shift
    typeset par4="${1:-}"
    (($#)) && shift
```
VS
```bash
   typeset par1="${1:-}"; (($#)) && shift
   typeset par2="${1:-}"; (($#)) && shift
   typeset par3="${1:-}"; (($#)) && shift
   typeset par4="${1:-}"; (($#)) && shift
```
It reduces the vertical footprint of the boilerplate code, allowing the actual logic of the function to stand out more. My 2 cents.

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
```diff
@@ -0,0 +1,82 @@
...
+    typeset resultFile=''
+    typeset -a xmlFiles=()
+
+    set +e
```
etirta: I'm not fully convinced we should disable it globally.

We are already using || true in specific places where we want to ignore failures, which implies that for other commands, we do want the script to exit on an error.

Could you share your train of thought on this change? I'd like to make sure we aren't accidentally masking critical failures while trying to handle specific non-zero exits. Thx.

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
```diff
@@ -0,0 +1,82 @@
...
+    InstallYq || true
+
+    while IFS= read -r -d '' resultFile; do
+        grep -qE '<testsuite(s)?' "${resultFile}" && xmlFiles+=("${resultFile}")
+    done < <(find "${ARTIFACT_DIR}" -type f -iname "*.xml" -print0 2>/dev/null || true)
```
etirta: The `|| true` isn't needed here. Process substitution (and command substitution) doesn't propagate exit codes to the main shell, so it won't trigger `errexit`. Removing `2>/dev/null` also helps with debugging without affecting the loop's stability. If `find` fails, the loop simply won't run. Just a small detail to keep the code a bit cleaner.

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
```diff
@@ -0,0 +1,82 @@
...
+
+    ((${#xmlFiles[@]})) || {
+        : "Warning: No JUnit XML file found to process" >&2
```
etirta: Since `:` is a no-op command, it doesn't produce any output. The `>&2` at the end doesn't actually have anything to redirect, so it can be removed to keep the line clean.

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
```diff
@@ -0,0 +1,82 @@
...
+        true
+        return
+    }
+
+    yq eval-all -px -ox -I2 '
+        {
+            "+p_xml": "version=\"1.0\" encoding=\"UTF-8\"",
+            "testsuites": {"testsuite": [
+                .[] |
+                (.testsuites.testsuite? // .testsuite? // .) |
```
etirta: I noticed a small optimization opportunity in this pipe. Because we perform `.[]` prior to this pipe, the context is already scoped to the inner content of the `<testsuites>` wrapper or the inner content of the `<testsuite>` (if it's not wrapped with `<testsuites>`). Consequently, the first part of that expression will always evaluate to false.

```suggestion
                (.testsuite // .) |
```
With the above proposal:
  * Case A (Wrapper exists): `.testsuite` correctly extracts the `<testsuite>` content.
  * Case B (No wrapper): `.testsuite` returns null, and the alternate `.` yields the root content (the `<testsuite>` itself).

In both scenarios, we achieve a consistent state for the subsequent pipes. Please stick with this specific `yq` script, as it was designed to handle both XML structures with maximum efficiency.

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
```diff
@@ -0,0 +1,82 @@
...
+                ([] + .)[] |
+                select(kind == "map") |
+                (select(env(MAP_TESTS) == "true") | ."+@name" = env(MAP_TESTS_SUITE_NAME) | .) // . |
+                ([] + (.testcase?)//[]) as $tc |
+                ."+@tests" = ($tc | length | tostring) |
+                ."+@failures" = ([$tc[] | select(.failure)] | length | tostring) |
+                ."+@errors" = ([$tc[] | select(.error)] | length | tostring)
+            ]}
+        }
+    ' "${xmlFiles[@]}" 1> "${ARTIFACT_DIR}/${mergedFN}" 2>/dev/null || true
```
etirta: I’m concerned about the `2>/dev/null || true` here. If this command fails, the job will proceed without any jUnit results, and we might not even realize it failed because the error is silenced. I believe we should let the error surface so we can debug it if the merge fails. This aligns with a 'fail-fast' strategy—if we have no results to process, the pipeline should probably let us know.

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
```diff
@@ -0,0 +1,82 @@
...
+
+    [[ -f "${ARTIFACT_DIR}/${mergedFN}" ]] && cp "${ARTIFACT_DIR}/${mergedFN}" "${SHARED_DIR}/" || true
```
etirta: Building on the previous point: if we remove the `|| true` from the `yq` command above, the `-f` check here might become redundant.

Regarding the `|| true` at the end of the `cp` command, we should answer the same question: if the copy fails, do we actually want the pipeline to keep moving? If the artifact doesn't make it to the `${SHARED_DIR}/`, subsequent steps that rely on it will likely fail anyway ([@oharan](https://redhat.enterprise.slack.com/team/U03SB1UAE3T) please confirm whether the Data Router Step will fail or not). I'd suggest we keep the error visible so we know exactly where the chain broke.

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
```diff
@@ -0,0 +1,82 @@
...
+    typeset arch=""
+    arch="$(uname -m | sed 's/aarch64/arm64/;s/x86_64/amd64/')"
+    curl -L "https://github.com/mikefarah/yq/releases/latest/download/yq_linux_${arch}" \
+        -o /tmp/bin/yq && chmod +x /tmp/bin/yq
+    /tmp/bin/yq --version || export MAP_TESTS=false
+}
+
+# One JUnit XML for Data Router: collect XMLs, one yq eval-all (root/name/recalc), write and copy to SHARED_DIR.
+# All env (e.g. MAP_TESTS, MAP_TESTS_SUITE_NAME) must be defined in step ref or script.
+function CleanupCollect() {
```
etirta, line 25: As point of reference, I'm putting the recommended clean up function that works for any Step we have.

High level workflow:
  * Merge multiple jUnit XML files into one, to avoid potential naming conflict with other Steps.
  * If warranted, map the Test Suite Name to CR (Component Readiness) known pattern.
  * It will archive the original jUnit XML files, so the Prow do not see duplicated result.

```bash
function CleanupCollect() {
    typeset mergedFN="${1:-jUnit.xml}"; (($#)) && shift
    typeset resultFile=''
    typeset -a xmlFiles=()

    InstallYq || true

    while IFS= read -r -d '' resultFile; do
        grep -qE '<testsuites?\b' "${resultFile}" && xmlFiles+=("${resultFile}")
    done 0< <(find "${ARTIFACT_DIR}" -type f -iname "*.xml" ! -name "${mergedFN}" -print0)

    ((${#xmlFiles[@]})) || {
        : "Warning: No jUnit XML file found to process"
        return
    }

    # Prepare one jUnit XML: collect -> map suite name and merge
    yq eval-all -px -ox -I2 '
        {
            "+p_xml": "version=\"1.0\" encoding=\"UTF-8\"",
            "testsuites": {"testsuite": [
                .[] |
                (.testsuite // .) |
                ([] + .)[] |
                select(kind == "map") | (
                    select(env(MAP_TESTS) == "true") |
                    ."+@name" = env(MAP_TESTS_SUITE_NAME)
                )//. |
                ([] + (.testcase // [])) as $tc |
                ."+@tests" = ($tc | length | tostring) |
                ."+@failures" = ([$tc[] | select(.failure)] | length | tostring) |
                ."+@errors" = ([$tc[] | select(.error)] | length | tostring)
            ]}
        }
    ' "${xmlFiles[@]}" 1> "${ARTIFACT_DIR}/${mergedFN}"

    # Archive the original jUnit XMLs.
    tar zcf "${ARTIFACT_DIR}/jUnit-original.tgz" -C "${ARTIFACT_DIR}/" "${xmlFiles[@]#${ARTIFACT_DIR}/}"
    rm -f "${xmlFiles[@]}"

    # The Data Router Step need to be able to fetch the merged jUnit XML.
    cp "${ARTIFACT_DIR}/${mergedFN}" "${SHARED_DIR}/"

    true
}

trap 'CleanupCollect junit--stackrox__compliance-e2e__stackrox-compliance-e2e.xml' EXIT
```

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
sg-rh: Yes, it was added so the EXIT trap wouldn’t exit on the failure for the rest of this scope,  we have `yq, grep, cp` commands there so was to avoid masking the original exit status as you mentioned.
I can remove and update those to ` || true` so we don’t mask other errors.

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
sg-rh: I have updated it

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
sg-rh: Thanks makes sense

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
sg-rh: Okay thanks, yes thats better handle for both cases, I can update it

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
sg-rh: Yes its better to have the error surfaced if there is no results, I have it updated.

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
```diff
@@ -0,0 +1,88 @@
...
+    typeset mergedFN="${1:-jUnit.xml}"; (($#)) && shift
+    typeset resultFile=''
+    typeset -a xmlFiles=()
+
+    InstallYq || true
+
+    while IFS= read -r -d '' resultFile; do
+        [[ "$(basename "${resultFile}")" == "${mergedFN}" ]] && continue
+        grep -qE '<testsuites?\b' "${resultFile}" && xmlFiles+=("${resultFile}") || true
+    done < <(find "${ARTIFACT_DIR}" -type f -iname "*.xml" -print0)
```
etirta: Kudos for avoiding to merge the merged file. Even though it should not be an issue since this is only called once, but I like the idea of this kind of prevention. It create a much better code!

However, it's better if we filter it in `find` as it's already doing the filter anyway, so it will be much more efficient if doing it from the source of the loop.


```suggestion
    while IFS= read -r -d '' resultFile; do
        [[ "$(basename "${resultFile}")" == "${mergedFN}" ]] && continue
        grep -qE '<testsuites?\b' "${resultFile}" && xmlFiles+=("${resultFile}") || true
    done < <(find "${ARTIFACT_DIR}" -type f -iname "*.xml" ! -name "${mergedFN}" -print0)
```

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
```diff
@@ -0,0 +1,88 @@
...
+
+    ((${#xmlFiles[@]})) || {
+        echo "Warning: No JUnit XML file found to process" >&2
```
etirta: No, let's avoid `echo` here as it will cause duplicated printout, because we already have `xtrace` active, and per our Best Practice documentation, the `xtrace` ensures every line is printed to STDERR before execution.
```suggestion
        : 'Warning: No JUnit XML file found to process'
```
Since `:` is a no-op that doesn't produce its own output, the `>&2` is redundant (the string itself will already be captured in the logs by `xtrace`). Let's keep it clean and just let the `xtrace` do its job.

etirta (changes requested):

ci-operator/step-registry/stackrox/compliance-e2e/stackrox-compliance-e2e-commands.sh
sg-rh: ok i will remove the loop check. find excludes the merged file.

PR #69966, opened 2025-10-03
hack: improve validate-boskos.sh with better error handling  - Add dy…

pgstorm148:
…namic path resolution to find repo root - Add color-coded output for better readability in CI - Improve error handling for missing files and failed generation - Restore original file on failure to prevent workspace pollution - Add proper cleanup and backup mechanisms - Use function-based structure for better maintainability  This makes the validation script more robust and user-friendly.

PR #70836, opened 2025-10-29
Revert "hack: improve validate-boskos.sh with better error handling  - Add dy…"

pruan-rht:
Reverts openshift/release#69966


pruan-rht:
@pgstorm148 I think if you keep the original way of setting the basedir then it should work

PR #70878, opened 2025-10-30
Updating the timeouts for Power and Z parallel jobs

tvardema:
There has been some concerns about needing more time for testing of the conformance parallel test suite on Power and Z, so we're extending the timeout to 5 hours in hopes this removes some false positives.


prb112 (approved):
/lgtm

PR #70882, opened 2025-10-30
TRT-2388: Migrate 4.21 periodics openshift ovn kubernetes

neisw:


neisw:
@jcaamano @tssurya - we are looking to make sure periodics are lined up with releases so we can analyze the signal properly. Let me know if this update looks right to you or you feel it belongs in master.

PR #70913, opened 2025-10-31
TRT-2388: migrate 4.21 periodic definitions

neisw:


ci-operator/config/openshift/kubernetes-sigs-karpenter/openshift-kubernetes-sigs-karpenter-release-4.21__periodics.yaml
```diff
@@ -0,0 +1,65 @@
...
+  '*':
+    limits:
+      memory: 4Gi
+    requests:
+      cpu: 100m
+      memory: 200Mi
+tests:
+- as: e2e-kwok
+  cron: 0 5 * * 1
+  presubmit: true
```
petr-muller, line 45: @neisw these cases where a single definition generates both a presubmit and a periodic are interesting.

I think `config-brancher` has a gap here because it skips this job entirely (I guess because it is _also_ a periodic) in its [release-4.22](https://github.com/openshift/release/blob/master/ci-operator/config/openshift/kubernetes-sigs-karpenter/openshift-kubernetes-sigs-karpenter-release-4.22.yaml) target. I think this means we are leaving some presubmits behind when we branch. I will follow up on this with TP.

The job exists in the [release-4.21](https://github.com/openshift/release/blob/cf7aa59262ca45cc025ff2205e1b7dfb1c48e0e0/ci-operator/config/openshift/kubernetes-sigs-karpenter/openshift-kubernetes-sigs-karpenter-release-4.21.yaml#L118-L135) target but I think this is because of the previous config-brancher fix that we want to revert.

We probably want to drop the `presubmit: true` here. The presubmit part of the job will be copied over from main. We handle the periodic part in this file. The unfortunate consequence is that now we have duplicate configuration where we'd normally have one. 🤔

PR #70918, opened 2025-10-31
TRT-2388: Migrate 4.21 periodics openshift cluster node tuning operator

neisw:


jmencak:
Thank you for the PR, does this need to be backported to 4.20 and below as well?

neisw:
@jmencak - the main issue we are trying to address is lack of ocp release attribution from `main` config and duplication of testing when the current dev branch like `4.21` is really a FF of main.  So no backport needed we are just trying to get periodics lined up with specific releases and reduce duplication.

PR #70919, opened 2025-10-31
TRT-2388: Migrate 4.21 periodics openshift cluster network operator

neisw:


neisw:
@jcaamano @tssurya - we are looking to make sure periodics are lined up with releases so we can analyze the signal properly. Let me know if this update looks right to you or you feel it belongs in master.

jcaamano:
@neisw honestly I would like to understand the difference. Isn't 4.21 and master the same thing now, so both definitions here would be in sync and equal?

neisw:
@jcaamano - There is no difference in the code being tested.  But for the tooling we use like Sippy && Component Readiness there is a difference between main/master and 4.21.  We require a specific OCP release to collect and analyze the data.  Currently we are running both main/master and 4.21 jobs which effectively duplicates job runs and doubles the frequency.

That said there have been branching issues with the periodic variant configuration not getting promoted at branch time.  Part of the what led to finding these master configuration is looking for those gaps to try to get that issue resolved.

> don't we need the master definitions to be able to use things like /payload-job-with-prs

I don't believe so.  You are looking to run the 4.21 version job, not the master.  Also the jobs for the current dev branch have been disabled previously so that only the master jobs were runnning.  This we turned off while we work to resolve the missing periodic configuration branching step but will likely be the case again soon.

No rush on this PR, it was something I found while investigating how we can fix the gap with periodic variant config branching.  If you want to hold off for now you can.

PR #70967, opened 2025-11-03
fix the cluster-secrets-gcp-confidential-qe secret configuration

droslean:
fixes https://github.com/openshift/release/pull/69411

/cc @openshift/test-platform @yalzhang


droslean:
@yalzhang Just for the heads up, the misconfiguration was supposed to be caught by our CI  which it did but the job passed because of a bug in our tool. There is https://github.com/openshift/ci-tools/pull/4801 opened to fix it. Sorry for the noise.

PR #70970, opened 2025-11-03
retester: telco: TAS: configure CI retries limit

shajmakh:
Configure max retests for **required** failing CI jobs to 2 per base branch for each commit sha. That applies to TAS repos which are numaresources-operator and scheduler-plugins.

### DIFF
--- a/core-services/retester/_config.yaml
+++ b/core-services/retester/_config.yaml
@@ -13,6 +13,14 @@ retester:
       enabled: true
     openshift-knative:
       enabled: true
+    openshift-kni:
+      repos:
+        numaresources-operator:
+          enabled: true
+          max_retests_for_sha_and_base: 2
+        scheduler-plugins:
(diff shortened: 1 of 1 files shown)


core-services/retester/_config.yaml
```diff
@@ -13,6 +13,17 @@ retester:
...
+      repos:
+        numaresources-operator:
+          enabled: true
+          max_retests_for_sha_and_base: 2
+        scheduler-plugins:
+          enabled: true
+          max_retests_for_sha_and_base: 2
+        cnf-features-deployer:
+          enabled: true
+          max_retests_for_sha_and_base: 2
```
ffromani: is this part of TAS?
once clarified LGTM

shajmakh: no, forgot to save the new updates by mistake. should be cleaned up now

PR #70971, opened 2025-11-03
[art] add konflux component validation

ashwindasr:
There was a z-stream release where the absence of a component in konflux caused us to miss the image from being released. With this PR, we should be able to identify the absence of such mapping at a much earlier stage


joepvd (approved):
/lgtm

PR #71030, opened 2025-11-04
Remove gather_loop logic

rsevilla87:
The output of the node-density jobs is unreadable, lets remove this logic as it's not currently used

/cc @vishnuchalla @jtaleric @josecastillolema


ci-operator/step-registry/openshift-qe/node-density/openshift-qe-node-density-commands.sh
rsevilla87: Do we need this command as well?

PR #71049, opened 2025-11-05
[draft] DNM: aws-gpu cluster up

swghosh:
Hack to quickly bring up an AWS cluster with GPU nodes.


empovit:
@swghosh why would you need to hack someone else's job, and use their AWS account to get a GPU cluster? Let alone creating unnecessary noise for an engineering team, without even coordinating with them?

PR #71108, opened 2025-11-06
OKD-279: Add e2e-vsphere-ovn test

kenjpais:
This test is required to merge the PR [#9959](https://github.com/openshift/installer/pull/9959)
Please refer the related PR for more details.

### DIFF
--- a/ci-operator/jobs/openshift/installer/openshift-installer-main-presubmits.yaml
+++ b/ci-operator/jobs/openshift/installer/openshift-installer-main-presubmits.yaml
@@ -7989,6 +7989,92 @@ presubmits:
     branches:
     - ^main$
     - ^main-
+    cluster: vsphere02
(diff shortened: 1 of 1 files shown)


Prashanth684:
@jcpowermac could you approve this so we can run the presubmit on https://github.com/openshift/installer/pull/10063

Prashanth684:
> > @jcpowermac could you approve this so we can run the presubmit on [openshift/installer#10063](https://github.com/openshift/installer/pull/10063)
> 
> Shouldn't the segfault issue be fixed before merging a ci job that we know doesn't pass?

ah..right the segfault. I went back and looked at the nightlies and it seems to happen there as well - although it doesn't get flagged as a failure (it happens first boot and when MCO overlays the node image, it goes away). I'm going to file a separate bug to look at that if you don't mind approving this

PR #71116, opened 2025-11-06
OCPKUEUE-362: Kueue: install jobset operator

MaysaMacedo:


MaysaMacedo:
The step has installed as expected:
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_release/71116/rehearse-71116-periodic-ci-openshift-kueue-operator-main-test-e2e-4-20/1986500395844767744/artifacts/test-e2e-4-20/kueue-operator-jobset-install-operator/build-log.txt

MaysaMacedo:
/pj-rehearse periodic-ci-openshift-kueue-operator-main-test-e2e-4-19

ci-operator/step-registry/kueue-operator/jobset-install-operator/kueue-operator-jobset-install-operator-commands.sh
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+
+echo "Installing Jobset Operator"
```
kannon92: In this case, I think you are fine but there is a precondition for jobset that cert manager must be installed.

MaysaMacedo: this step will only run after cert-manager and kueue are up

ci-operator/step-registry/kueue-operator/test/e2e/kueue-operator-test-e2e-chain.yaml
```diff
@@ -14,6 +14,20 @@ chain:
         cpu: 200m
         memory: 300Mi
   - ref: kueue-operator-image-patch
+  - as: kueue-operator-jobset-install-operator
+    cli: latest
+    commands: |
+      CURRENT_BRANCH=$(git branch --show-current)
+      if [[ "$CURRENT_BRANCH" == "release-1.0" || "$CURRENT_BRANCH" == "release-1.1" ]]; then
```
kannon92, line 21: This would only work on periodics when the branch is release-1.0 / releaes-1.1.

I guess this would always run on presubmits.

Actually that is probably fine but this does mean that on PRs open to release-1.1 or release-1.0 we would still run this block of code.

TBH I don't even really think we need this code. What is the bad thing about installing jobset-operator on release-1.1 or release-1.0.

We don't have to run the e2e tests but I don't see a reason why we can't install the operator anyway just so we don't need to carry this code.

kannon92: Oh. This target doesn't exist on release branches based on release-1.1 or release-1.0.

MaysaMacedo:
/pj-rehearse pull-ci-openshift-kueue-operator-release-1.0-test-e2e-4-17

kannon92 (commented):
/lgtm
/approve
/pj-rehearse ack

PR #71132, opened 2025-11-06
stackrox(ci): finish ocp 4-20 testing setup

davdhacs:
* Enable OCP 4.20 tests on master as always-run/run-if-changed
* Remove OCP 4.18 and 4.19 test configs


ci-operator/config/stackrox/stackrox/stackrox-stackrox-master__ocp-4-18.yaml
```diff
@@ -103,7 +103,7 @@ tests:
       TEST_SUITE: ocp-compliance-e2e-tests
     workflow: stackrox-automation-flavors-ocp-4-e2e
   timeout: 5h0m0s
-- always_run: true
+- always_run: false
```
msugakov: Why do we need 4.18 on `master` at all?
Should we just delete it as well as 4.19 after making 4.20 the default for tests?

davdhacs: Keeping it (as only run on-demand) allows us to run CI tests against a supported version of OCP from a PR.
I think it would be useful to keep the range of OCP versions available to test while we're supporting running on them. What do you think? I think the downsides are A: the extra files in this directory (too many / confusing for the wider team to approach the config?) and B: the long `/test ?` list (but that is already un-usably long with only the oldest+latest versions).

msugakov: Yes, `/test ?` goes out of hand and that's the main reason for me. Whenever I ask Konflux to `/retest <something>`, OSCI replies with longer and longer list.

Also, we don't cover all versions so this looks inconsistent and as an engineer not involved in OSCI I would doubt before attempting anything on 4.18.

Finally, these files add mental overhead which means more expensive maintenance.

If I have a case to test my change on older OCP version, I'd spin up an Infra cluster and test my change there, without e2e tests. In addition, will trust e2e tests on the "newest" and the "oldest" versions to do their part and provide safety from regressions.

Therefore, if there's no active/persistent situation to use them, I'd vote to delete instead of having them "in case of need" because the need may never materialize.

davdhacs: > as an engineer not involved in OSCI I would doubt before attempting anything on 4.18.

> Finally, these files add mental overhead which means more expensive maintenance.

I think these are reason enough to remove the extra version tests.

> will trust e2e tests on the "newest" and the "oldest" versions to do their part and provide safety from regressions.

👍

ci-operator/jobs/infra-periodics.yaml
```diff
@@ -60,7 +60,7 @@ periodics:
           cpu: 500m
     serviceAccountName: upgrader
 - agent: kubernetes
-  cluster: build07
+  cluster: app.ci
```
msugakov: Is this `cluster: build07` -> `cluster: app.ci` change intended and/or automatic? Asking because I never saw `app.ci` before.

msugakov (approved):
/lgtm

but

/hold

due to the question

ci-operator/jobs/infra-periodics.yaml
davdhacs: This was automatic. I'll investigate.

davdhacs: It was resolved in master: https://github.com/openshift/release/pull/71166/files

PR #71187, opened 2025-11-10
fix(OCPQE-31091): i18next-parser modeule requires node version 22 but…

jimma:
… it's 24


jimma:
@yapei Please review.

PR #71190, opened 2025-11-10
Add cluster metadata when KUBECONFIG_ORIGIN=vault

tanfengshuang:
Some steps require metadata.json, such as https://steps.ci.openshift.org/reference/acm-tests-observability#line32

The code has been verified in rehearse job [ocp4.20-interop-opp-aws-RHCOS10](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/70935/rehearse-70935-periodic-ci-stolostron-policy-collection-main-ocp4.20-interop-opp-aws/1986663452638187520)

### DIFF
--- a/ci-operator/step-registry/openshift-qe/installer/bm/load-kubeconfig/openshift-qe-installer-bm-load-kubeconfig-commands.sh
+++ b/ci-operator/step-registry/openshift-qe/installer/bm/load-kubeconfig/openshift-qe-installer-bm-load-kubeconfig-commands.sh
@@ -63,7 +63,7 @@ PROXY_EOF
     for tgtFile in kube{admin-password,config} metadata.json; do
         srcFile="${tgtFile}${KCFG_SRC_SFX:+"--${KCFG_SRC_SFX}"}"
         if [ -r "${CLUSTER_PROFILE_DIR}/${srcFile}" ]; then
-            cp "${CLUSTER_PROFILE_DIR}/${srcFile}" "${SHARED_DIR}/${tgtFile}"
+            cp -f "${CLUSTER_PROFILE_DIR}/${srcFile}" "${SHARED_DIR}/${tgtFile}"
             [ "${tgtFile}" = kubeconfig ] && cp "${CLUSTER_PROFILE_DIR}/${srcFile}" "${SHARED_DIR}/${tgtFile}-minimal"
         fi
     done


ci-operator/step-registry/openshift-qe/installer/bm/load-kubeconfig/openshift-qe-installer-bm-load-kubeconfig-commands.sh
```diff
@@ -60,7 +60,7 @@ PROXY_EOF
   (vault)
     typeset srcFile='' tgtFile=''
 
-    for tgtFile in kube{admin-password,config}; do
+    for tgtFile in kube{admin-password,config} metadata.json; do
```
etirta, line 63: Please consider the proposed solution instead of hard-coding user files in here.

ci-operator/step-registry/openshift-qe/installer/bm/load-kubeconfig/openshift-qe-installer-bm-load-kubeconfig-commands.sh
```diff
@@ -60,7 +60,7 @@ PROXY_EOF
...
         srcFile="${tgtFile}${KCFG_SRC_SFX:+"--${KCFG_SRC_SFX}"}"
         if [ -r "${CLUSTER_PROFILE_DIR}/${srcFile}" ]; then
             cp "${CLUSTER_PROFILE_DIR}/${srcFile}" "${SHARED_DIR}/${tgtFile}"
```
etirta: If you don't mind please add `-f` so it doesn't fail if for some reason the target exist.
```suggestion
            cp -f "${CLUSTER_PROFILE_DIR}/${srcFile}" "${SHARED_DIR}/${tgtFile}"
```

etirta (changes requested):
May I got clarification whether this `metadata.json` is a file that is integral to CI Operator, like the `KUBECONFIG` file.

For context, the `KUBECONFIG` and `KUBEADMIN_PASSWORD_FILE` is [integral](https://docs.ci.openshift.org/docs/architecture/step-registry/#available-environment-variables) to CI Operator, so it's guaranteed to always available. I'm not sure if this is the case with the `metadata.json`.

If it's not, then I would propose to develop a generic solution, because:
 1. The `openshift-qe-installer-bm-load-kubeconfig` step is specific to `KUBECONFIG`.
 2. We should not hard-code user supplied file.

My proposal is to create a generic step that allow user to sync. user secret under Cluster Profile, say `external-cluster-cluster-profile-file-picker`:
```shell
#!/bin/bash
set -eux -o pipefail; shopt -s inherit_errexit

typeset srcFile= tgtFile=
typeset EC__CPFP__SFX="${EC__CPFP__SFX:-${KCFG_SRC_SFX}}"
eval "typeset -A fileMap=(${EC__CPFP__FILES})"

for tgtFile in "${!fileMap[@]}"; do
    srcFile="${fileMap[${tgtFile}]"
    if [ -r "${CLUSTER_PROFILE_DIR}/${srcFile}" ]; then
        cp -f "${CLUSTER_PROFILE_DIR}/${srcFile}" "${SHARED_DIR}/${tgtFile}"
    fi
done

exit 0
```

```yaml
 ...
  documentation: |-
    This Step is meant to be used in `external-cluster` Workflow to override
    files from main Vault Secret Path.

    As per CI Operator documentation
    (https://docs.ci.openshift.org/docs/how-tos/adding-a-cluster-profile/#providing-credentials)
    it is possible for multiple Vault Secret Paths to be synchronized to the
    same Secret Name. In this case the content of `${CLUSTER_PROFILE_DIR}/` will
    be the resulting merged of all Keys from these Vault Secrets group.

    The problem is we can not control Vault Secret precedence, so the Keys with
    the same name will override each other and the final content of the
    corresponding files is not predictable. To overcome this situation, we
    should never have more than one Key with the same name across these Vault
    Secrets groups.

    One way to avoid name collision is to applied suffix. For ex., if there is a
    Key name `secrets.json` in the main Vault Secret and we want to use
    different Key from different Vault Secret (that is also synchronized to the
    same Cluster Profile Secret), then we should name the other Key as
    `secrets--my-secret.json`.

    So now under `${CLUSTER_PROFILE_DIR}/` there will be files named
    `secrets.json` and `secrets--my-secret.json`. This step can be configured to
    pick one of these files and copy it to `${SHARED_DIR}/`

    Other use case is, there might be some extra files that some `step`, say an
    OCP Deployment `step`, put under `${SHARED_DIR}/`, for ex. `metadata.json`.
    Since the `external-cluser` workflow do not call any OCP Deployment `step`,
    then these extra files will be missing. This `step` can also be used to
    transfer these missing files that correspond to the associated external
    cluster. In this case, as long as there is no name conflict, the use of
    suffix is optional.
  env:
    - name: EC__CPO__FILES
      default: ''
      documentation: |-
        List of files under to be copied from `${CLUSTER_PROFILE_DIR}/` to
        `${SHARED_DIR}/`, in BASH Associative Array value format.

        The Associative Array will be created via
        `eval typeset -A fileMap=(${EC__CPFP__FILES})`.

        Syntax: [<tgtFile>]='<srcFile>' ...
        Example: [secrets.json]='secrets--my-secret.json' [creds.yaml]='creds--my-cred.yaml'
 ...
```

ci-operator/step-registry/openshift-qe/installer/bm/load-kubeconfig/openshift-qe-installer-bm-load-kubeconfig-commands.sh
tanfengshuang: File metadata.json is generated after OCP cluster installation like kubeadmin-password and kubeconfig.

kubeadmin-password, kubeconfig-minimal, kubeconfig and metadata.json are stored under $SHARED_DIR after cluster insallation.

I have shared more info with you via Slack, and we can talk more about it there. Thank you!

PR #71224, opened 2025-11-10
OCPBUGS-61376: Configure a new HPA periodic job

maxcao13:
This commit configures a new weekly periodic job for o/k that only focuses on the HPA. It runs these through the new kubernetes/autoscaling/hpa OTE suite that was created here: https://github.com/openshift/kubernetes/pull/2526
 
There is a long tail of history for this change and relevant slack conversations.

See: https://issues.redhat.com//browse/OCPBUGS-61376
Upstream issue: https://github.com/kubernetes/kubernetes/issues/135026

### DIFF
--- a/ci-operator/config/openshift/kubernetes/openshift-kubernetes-master.yaml
+++ b/ci-operator/config/openshift/kubernetes/openshift-kubernetes-master.yaml
@@ -568,6 +568,13 @@ tests:
       enable:
       - observers-resource-watch
     workflow: openshift-e2e-external-aws
+- as: k8s-e2e-aws-hpa
+  cron: 0 5 * * 1
+  steps:
+    cluster_profile: aws-2
+    env:
+      TEST_SUITE: kubernetes/autoscaling/hpa
+    workflow: openshift-e2e-aws
 zz_generated_metadata:
   branch: master
   org: openshift
--- a/ci-operator/jobs/openshift/kubernetes/openshift-kubernetes-master-periodics.yaml
+++ b/ci-operator/jobs/openshift/kubernetes/openshift-kubernetes-master-periodics.yaml
@@ -70,3 +70,74 @@ periodics:
     - name: result-aggregator
       secret:
         secretName: result-aggregator
+- agent: kubernetes
+  cluster: build01
+  cron: 0 5 * * 1
+  decorate: true
+  extra_refs:
+  - base_ref: master
+    org: openshift
+    path_alias: k8s.io/kubernetes
+    repo: kubernetes
+  labels:
+    ci-operator.openshift.io/cloud: aws
+    ci-operator.openshift.io/cloud-cluster-profile: aws-2
+    ci.openshift.io/generator: prowgen
+    pj-rehearse.openshift.io/can-be-rehearsed: "true"
+  name: periodic-ci-openshift-kubernetes-master-k8s-e2e-aws-hpa
+  spec:
+    containers:
+    - args:
+      - --gcs-upload-secret=/secrets/gcs/service-account.json
+      - --image-import-pull-secret=/etc/pull-secret/.dockerconfigjson
+      - --lease-server-credentials-file=/etc/boskos/credentials
+      - --report-credentials-file=/etc/report/credentials
+      - --secret-dir=/secrets/ci-pull-credentials
+      - --target=k8s-e2e-aws-hpa
+      command:
+      - ci-operator
+      image: quay-proxy.ci.openshift.org/openshift/ci:ci_ci-operator_latest
+      imagePullPolicy: Always
+      name: ""
+      resources:
+        requests:
+          cpu: 10m
+      volumeMounts:
+      - mountPath: /etc/boskos
+        name: boskos
+        readOnly: true
+      - mountPath: /secrets/ci-pull-credentials
+        name: ci-pull-credentials
+        readOnly: true
+      - mountPath: /secrets/gcs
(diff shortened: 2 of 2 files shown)


maxcao13:
This PR requires https://github.com/openshift/kubernetes/pull/2492 to be merged.

/hold

maxcao13:
I'll rehearse a different job to show that changing the step doesn't mess up the defaults and therefore the tests.

/pj-rehearse pull-ci-openshift-kubernetes-master-k8s-e2e-aws

ci-operator/config/openshift/kubernetes/openshift-kubernetes-master.yaml
```diff
@@ -568,6 +568,14 @@ tests:
...
       - observers-resource-watch
     workflow: openshift-e2e-external-aws
+- as: k8s-e2e-aws-hpa
+  optional: true
+  steps:
+    cluster_profile: aws-2
+    env:
+      KUBE_E2E_TEST_ARGS: -focus=\[Feature:HPA\]
+      NODES: "3"
+    workflow: openshift-kubernetes-e2e-aws
```
bertinatto: In the spirit of saving CI resources, would it be possible to create a periodic job instead?

I'm not sure if it makes sense to run these tests/job every time we create a PR against o/k. IMO a periodic job would make more sense.

CC @jacobsee

maxcao13: Sure, totally makes sense to me. I'll change it to a periodic with weekly frequency.

bertinatto: @maxcao13 would it be possible to use a regular e2e job (i.e., a job that uses `openshift-tests` rather than the upstream ginkgo-based binary)?

maxcao13: Looking over everything again, I think there's a way, but this would involve going back to openshift/kubernetes and changing some configuration again, if you don't mind reviewing that.

I think what we can do is:
1. Remove HPA as a disabled spec
```diff
# k8s-test-ext/disabled_tests.go
		"Flaky": {
			"Job should run a job to completion when tasks sometimes fail and are not locally restarted", // seems flaky, also may require too many resources
		},
---		// tests that will be configured to run manually through their own dedicated prow jobs
---		"DedicatedJob": {
---		"[Feature:HPA]", // we run these tests directly using test-kubernetes-e2e.sh and ginkgo --focus through a separate prow job per https://issues.redhat.com/browse/OCPBUGS-61376
		},
```
2. Add the `DedicatedJob` label to the HPA tests so the filter can catch it.
```diff
# k8s-test-ext/labels.go
		"[sig-arch]": {
			// not run, assigned to arch as catch-all
			"[Feature:GKELocalSSD]",
			"[Feature:GKENodePool]",
		},
+++		 "[DedicatedJob]": {
+++		 	"[Feature:HPA]",
+++		},
```
3. include `DedicatedJob` in the `withExcludedTestsFilter`, so that these tests don't get run by default in the `k8s-test-ext` suites.
```diff
# k8s-test-ext/k8s-tests.go
func withExcludedTestsFilter(baseExpr string) string {
	excluded := []string{
		"[Disabled:",
		"[Disruptive]",
		"[Skipped]",
		"[Slow]",
		"[Flaky]",
		"[Local]",
+++		"[DedicatedJob]",
	}

	filter := ""
	for i, s := range excluded {
		if i > 0 {
			filter += " && "
		}
---		filter += fmt.Sprintf("!name.contains('%s')", s)
+++		filter += fmt.Sprintf("!name.contains('%s') && !labels.exists(l, l == '%s')", s, s)

	}

	if baseExpr != "" {
		return fmt.Sprintf("(%s) && (%s)", baseExpr, filter)
	}
	return filter
}

```
4. Add a _new_ suite to this OTE called `kubernetes/autoscaling/hpa`:
```diff
# k8s-test-ext/k8s-tests.go
	kubeTestsExtension.AddSuite(e.Suite{
		Name:       "kubernetes/autoscaling/hpa",
		Qualifiers: []string{"name.contains('[Feature:HPA]')"},
		Parallelism: 3,
	})
```

And then in this PR, add that suite to the corresponding `openshift-e2e-test` ref as a suite, so we can reference it in our new job.

Before changing anything, would this path be better than hacking this in as currently presented?

bertinatto: Thank you so much for investigating this, @maxcao13! This seems reasonable to me. Running these tests with `openshift-tests` is useful since it give us helpful features like monitor tests or other similar mechanisms. I think it's worth the effort.

maxcao13: Opened https://github.com/openshift/kubernetes/pull/2526

ci-operator/config/openshift/kubernetes/openshift-kubernetes-master.yaml
```diff
@@ -568,6 +568,13 @@ tests:
       enable:
       - observers-resource-watch
     workflow: openshift-e2e-external-aws
+- as: k8s-e2e-aws-hpa
```
bertinatto, line 571: This could be done as a follow-up PR, but I suggest you don't use the `k8s-` prefix because that implies the job is using ginkgo.

bertinatto (commented):
/lgtm

PR #71280, opened 2025-11-12
Add DNS verification after creating GCP Cloud DNS records

gpei:
## Summary

  Add DNS verification after creating GCP Cloud DNS records to ensure records are resolvable before proceeding with
  cluster installation.

  ## Motivation

  Currently, the `gcp-provision-public-dns-zone-record-sets` step creates DNS records in GCP Cloud DNS but does not
  verify that they are actually resolvable. While GCP typically propagates DNS records to its authoritative
  nameservers within seconds, DNS caching and resolver behavior can introduce delays. Proceeding with cluster
  installation before DNS records are fully propagated can lead to failures in subsequent steps that depend on these
   records.
   Example failure job: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/70591/rehearse-70591-pull-ci-openshift-installer-main-e2e-aws-custom-dns-techpreview/1988412658591731712

### DIFF
--- a/ci-operator/config/openshift/installer/openshift-installer-main.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-main.yaml
@@ -439,7 +439,6 @@ tests:
     cluster_profile: aws-qe
     env:
       AWS_INSTALL_USE_MINIMAL_PERMISSIONS: "yes"
-      CUSTOM_OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE: registry.build10.ci.openshift.org/ci-ln-3lp2yn2/release:latest
       FEATURE_SET: TechPreviewNoUpgrade
       TEST_SKIPS: DNS should provide DNS for the cluster \| Aggregator Should be able
         to support the 1.17 Sample API Server using the current Aggregator \| Ensure


gpei:
@yunjiang29 for the review, thanks

ci-operator/step-registry/gcp/provision/public-dns-zone/record-sets/gcp-provision-public-dns-zone-record-sets-commands.sh
```diff
@@ -48,3 +48,40 @@ do
 gcloud --project ${project_id} dns record-sets delete ${name} --type ${record_type} --zone ${base_domain_zone_name}
 EOF
 done
+
+echo "Verifying the created DNS records can be resolved ..."
```
yunjiang29, line 52: How about put this verification in `cucushift-installer-check-user-provisioned-dns`, so it could cover all public custom DNS clusters.

gpei: Oh it's actually not a post check point, it's similar to https://steps.ci.openshift.org/reference/aws-provision-route53-public-record-sets#line94 to make sure the public DNS created on GCP are ready before running the subsequent steps, like `nodes-readiness`, etc. 
For the custom DNS clusters using AWS Route53 to provision the public DNS, seems `aws route53 wait resource-record-sets-changed ` is enough.

PR #71282, opened 2025-11-12
TRT-2337: update stable release controller to use registry-based workflow for minor upgrade

jianlinliu:


core-services/release-controller/_releases/release-ocp-4.y-stable.json
```diff
@@ -10,24 +10,24 @@
...
       "maxRetries": 2,
-      "prowJob":{"name":"release-openshift-origin-installer-e2e-aws-upgrade"}
+      "prowJob":{"name":"periodic-ci-openshift-release-master-stable-4.y-e2e-aws-ovn-upgrade"}
     },
     "upgrade-minor":{
       "upgrade":true,
       "maxRetries": 2,
       "upgradeFrom": "PreviousMinor",
-      "prowJob":{"name":"release-openshift-origin-installer-e2e-aws-upgrade"}
+      "prowJob":{"name":"periodic-ci-openshift-release-master-stable-4.y-e2e-aws-ovn-upgrade"}
```
neisw, line 19: If you don't have a good way to test this via rehearsal I would suggest only changing this upgrade-minor job first.  Since it is already [broken](https://amd64.ocp.releases.ci.openshift.org/releasestream/4-stable/release/4.20.3) it shouldn't cause any issues by changing it.  If we validate it works then we can change the others.

jianlinliu: Make sense.

PR #71298, opened 2025-11-12
Fix clear_partition_disk_table to handle missing devices field in LVM cluster status

ccardenosa:
The `clear_partition_disk_table` function was failing when the LVM cluster status didn't have a 'devices' field, which occurs when the volume group creation fails due to partitioned disks.

Changes:
- Add fallback to extract device names from failure reason messages
- Implement device exclusion logic using LVM cluster's excluded devices list
- Add block device discovery to identify partition-bearing devices
- Improve device wiping with multiple fallback methods (sgdisk/wipefs/dd)
- Add partprobe to refresh partition tables after wiping
- Enhanced error handling and logging for better debugging

This ensures the function works in both scenarios:
1. When devices are successfully added to VG (devices field exists)
2. When VG creation fails due to partitioned devices (parse from error)

The function now properly identifies devices that need partition table cleaning while respecting the exclusion list from LVM cluster status.

### DIFF
--- a/ci-operator/step-registry/telcov10n/metal-single-node-spoke/site-config/deploy/telcov10n-metal-single-node-spoke-site-config-deploy-commands.sh
+++ b/ci-operator/step-registry/telcov10n/metal-single-node-spoke/site-config/deploy/telcov10n-metal-single-node-spoke-site-config-deploy-commands.sh
@@ -64,27 +64,90 @@ function clear_partition_disk_table {
 
   set -x
   node_name="$(oc get node -oname)"
-  devices_to_be_wiped=$(oc -n openshift-storage get lvmclusters.lvm.topolvm.io lvmcluster -ojson | \
-    jq -c '
-      .status.deviceClassStatuses[]
-      | select(.name == "vg1").nodeStatus[].devices')
+
+  # Try to get devices from status (if LVM cluster is working)
+  devices_from_status=$(oc -n openshift-storage get lvmclusters.lvm.topolvm.io lvmcluster -ojson | \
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/telcov10n/metal-single-node-spoke/site-config/deploy/telcov10n-metal-single-node-spoke-site-config-deploy-commands.sh
```diff
@@ -64,27 +64,90 @@ function clear_partition_disk_table {
...
+    echo "Wiping device ${node_dev}..."
     echo
+    set -x
     oc debug ${node_name} -n default -- chroot /host bash -c "
      set -x ;
-     lsblk --fs ${node_dev} ;
-     sgdisk --zap-all ${node_dev} ;
+     echo 'Before wiping:' ;
+     lsblk --fs ${node_dev} || true ;
+     sgdisk --zap-all ${node_dev} || wipefs -af ${node_dev} || dd if=/dev/zero of=${node_dev} bs=1M count=100 ;
```
eifrach: just to make it faster and simpler 
you don't need to wipe the disk or write zero's to the disk  -- it takes too long and the only value is for security 

maybe it's better just to drop the partition table and recreate it -- it will remove all data ( or pointers to the data ) 

```
sfdisk --delete /dev/sdb
```

ccardenosa: You are right. Good catch!

PR #71322, opened 2025-11-12
increase timeout for upgrade jobs in CNO 4.21

jluhrsen:
this is becoming common across projects now. we use the same in OVNK as well


ricky-rav:
let's see if we can get the aws upgrade job to pass
/pj-rehearse

ci-operator/config/openshift/cluster-network-operator/openshift-cluster-network-operator-master__4.21-upgrade-from-stable-4.20.yaml
```diff
@@ -45,6 +45,7 @@ tests:
       EXTRA_MG_ARGS: --host-network
       TEST_TYPE: upgrade-conformance
     workflow: openshift-upgrade-aws-ovn
+  timeout: 6h0m0s
```
arkadeepsen, line 48: Details of the default decoration config timeout can be found here:

- https://docs.ci.openshift.org/docs/architecture/timeouts/#prow-test-process-timeouts
- https://github.com/openshift/release/blob/9238ee8b89c861793f9b8e8f7f6509abe33bd0b8/core-services/prow/02_config/_config.yaml#L559

PR #50219, opened 2024-03-26
enhance chain upgrade process

jianlinliu:
1. Remove loop in upgrade-toimage step, we only use it in one-hop upgrade as a step.
2. Introduce `cucushift-upgrade-drop-last-hop`  to drop the last hop upgrade edge, so that we can easily opt-in some other steps before the last hop upgrade.
3. Update chainupgrade-toimage step to make it possible to run the whole upgrade edge.
4. Sync functions codes between upgrade-toimage and chainupgrade-toimage.
5. Enhance e2e log reporter in chainupgrade-toimage to meet the same e2e test suite is executed for multiple times.
6. Move rollback ci jobs from release folder to QE's ci folder.

PR #71323, opened 2025-11-12
ci-operator/config/openshift/hypershift: Add e2e-azure-upgrade-rollback-oldest-supported

wking:
Similar to the CI we used to run for standalone before a9af8fdebf (#50219) shifted those into
`ci-operator/config/openshift/openshift-tests-private`.  I'm using Azure, because [the Hosted ARO folks are interested in seeing this CI passing][1].

I'm using the `fast` channel because the Hosted ARO folks would like to see successful rollbacks from new candidate releases back to an older generally-available release.  Ideally we can leave the actual version pinned to 4.20.0 for the life of 4.20, but if we need to land a change that breaks rollbacks to 4.20.0, this allows us to raise the `oldest-supported-z-rollback` to any other generally-available 4.20.z.

[1]: https://issues.redhat.com/browse/OCPSTRAT-1499

### DIFF
--- a/ci-operator/config/openshift/hypershift/openshift-hypershift-release-4.20__periodics.yaml
+++ b/ci-operator/config/openshift/hypershift/openshift-hypershift-release-4.20__periodics.yaml
@@ -57,6 +57,10 @@ releases:
       product: ocp
       stream: nightly
       version: "4.20"
+  oldest-supported-z-rollback:
+    release:
+      channel: fast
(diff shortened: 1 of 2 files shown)


wking:
/pj-rehearse periodic-ci-openshift-hypershift-release-4.20-periodics-e2e-azure-upgrade-rollback-oldest-supported

PR #71324, opened 2025-11-12
adding best_effort for int gather_extra runs

sclarkso:
@raelga Per the slack discussion, I attempted this.  I have not made prow changes before, so this could be the wrong direction.  I added a chain file so that best_effort could be set for specific runs and not affect all runs by modifying the gather_extra step directly.

### DIFF
--- a/ci-operator/config/Azure/ARO-HCP/Azure-ARO-HCP-main__periodic.yaml
+++ b/ci-operator/config/Azure/ARO-HCP/Azure-ARO-HCP-main__periodic.yaml
@@ -49,12 +49,13 @@ tests:
 - as: integration-e2e-parallel
   cron: 0 */4 * * *
   steps:
+    allow_best_effort_post_steps: true
     env:
       ARO_HCP_SUITE_NAME: integration/parallel
       CUSTOMER_SUBSCRIPTION: ARO SRE Team - INT (EA Subscription 3)
       LOCATION: uksouth
     post:
-    - ref: aro-hcp-provision-aro-hcp-gather-extra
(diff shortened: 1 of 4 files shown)


ci-operator/config/Azure/ARO-HCP/Azure-ARO-HCP-main__periodic.yaml
```diff
@@ -49,12 +49,13 @@ tests:
...
   cron: 0 */4 * * *
   steps:
+    allow_best_effort_post_steps: true
     env:
       ARO_HCP_SUITE_NAME: integration/parallel
       CUSTOMER_SUBSCRIPTION: ARO SRE Team - INT (EA Subscription 3)
       LOCATION: uksouth
     post:
-    - ref: aro-hcp-provision-aro-hcp-gather-extra
+    - chain: aro-hcp-provision-aro-hcp-gather-extra-best-effort
```
venkateshsredhat, line 58: this should be a best effort in all cases where we use them IMO .

sclarkso: I asked in the thread and was waiting for a response.  If it can be for EVERY run of that test, then we can just modify the config to allow, and then add the best_effort: true to the step-registry I think, instead of adding the chain.

venkateshsredhat: Ahh i missed that . IMO yes we should .

PR #71346, opened 2025-11-13
Fix automation issues on AWS Custom  DNS cluster

yunjiang29:
Fixed:
* Even though the route53 is a global resources, but if no region was specified on GovCloud, will hit `InvalidClientTokenId` errors while performing `list-hosted-zones-by-name` command, specify region explicitly now.
* Hardcode waiting 180s for DNS get ready after new record created.


yunjiang29:
@gpei PTAL

PR #71348, opened 2025-11-13
PTP CI: output podman build logs even when it fails

nocturnalastro:


ci-operator/step-registry/telco5g/ptp/tests/telco5g-ptp-tests-commands.sh
```diff
@@ -62,8 +62,20 @@ spec:
           set -xe
           yum install jq git wget podman-docker -y
           yum group install "development-tools" -y
+
```
aneeshkp: just add compression=none and try 
wget --compression=none

aneeshkp (commented):
wget --compression=none

ci-operator/step-registry/telco5g/ptp/tests/telco5g-ptp-tests-commands.sh
lack: Or maybe use `curl -LO $URL` because it keeps the destination the same as the source?

PR #71383, opened 2025-11-14
CNTRLPLANE-1137: Add ci for cilium e2e test

wewang58:
Add ci for e2e test codes: https://github.com/openshift/hypershift/pull/7077

### DIFF
--- a/ci-operator/config/openshift/hypershift/openshift-hypershift-main.yaml
+++ b/ci-operator/config/openshift/hypershift/openshift-hypershift-main.yaml
@@ -176,6 +176,26 @@ tests:
       HYPERSHIFT_AZURE_MARKETPLACE_IMAGE_SKU: ""
       HYPERSHIFT_AZURE_MARKETPLACE_IMAGE_VERSION: ""
     workflow: hypershift-azure-aks-e2e
+- as: e2e-aks-cilium
+  skip_if_only_changed: (^(\.tekton|\.github|\.claude|docs|examples|enhancements|contrib|\.cursor)/)|(^[A-Z]+\.md$)|((^|/)OWNERS$)|(/overrides\.yaml$)|(^renovate\.json$)|(/\.testcoverage\.yml$)|(^\.gitlint$)|(^\.gitignore$)
+  steps:
+    cluster_profile: hypershift-aks
+    dependencies:
+      OCP_IMAGE_N1: release:n1minor
+      OCP_IMAGE_N2: release:n2minor
+      OCP_IMAGE_N3: release:n3minor
+      OCP_IMAGE_N4: release:n4minor
+    env:
+      AUTH_THROUGH_CERTS: "true"
+      CILIUM_VERSION: 1.15.1
+      ENABLE_HYPERSHIFT_CERT_ROTATION_SCALE: "true"
+      EXTERNAL_CNI_PROVIDER: cilium
+      HYPERSHIFT_AZURE_LOCATION: centralus
+      HYPERSHIFT_AZURE_MARKETPLACE_IMAGE_OFFER: ""
+      HYPERSHIFT_AZURE_MARKETPLACE_IMAGE_PUBLISHER: ""
+      HYPERSHIFT_AZURE_MARKETPLACE_IMAGE_SKU: ""
+      HYPERSHIFT_AZURE_MARKETPLACE_IMAGE_VERSION: ""
+    workflow: hypershift-azure-aks-e2e
 - always_run: false
   as: e2e-aks-4-21
   pipeline_run_if_changed: ^(test/e2e|hypershift-operator)
--- a/ci-operator/step-registry/hypershift/azure/run-e2e/hypershift-azure-run-e2e-commands.sh
+++ b/ci-operator/step-registry/hypershift/azure/run-e2e/hypershift-azure-run-e2e-commands.sh
@@ -102,6 +102,10 @@ MARKETPLACE_IMAGE_PARAMS=""
 if [[ -n "${HYPERSHIFT_AZURE_MARKETPLACE_IMAGE_PUBLISHER:-}" && -n "${HYPERSHIFT_AZURE_MARKETPLACE_IMAGE_OFFER:-}" && -n "${HYPERSHIFT_AZURE_MARKETPLACE_IMAGE_SKU:-}" && -n "${HYPERSHIFT_AZURE_MARKETPLACE_IMAGE_VERSION:-}" ]]; then
   MARKETPLACE_IMAGE_PARAMS="--e2e.azure-marketplace-publisher ${HYPERSHIFT_AZURE_MARKETPLACE_IMAGE_PUBLISHER} --e2e.azure-marketplace-offer ${HYPERSHIFT_AZURE_MARKETPLACE_IMAGE_OFFER} --e2e.azure-marketplace-sku ${HYPERSHIFT_AZURE_MARKETPLACE_IMAGE_SKU} --e2e.azure-marketplace-version ${HYPERSHIFT_AZURE_MARKETPLACE_IMAGE_VERSION}"
 fi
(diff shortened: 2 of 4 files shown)


ci-operator/config/openshift/hypershift/openshift-hypershift-main.yaml
```diff
@@ -178,6 +178,25 @@ tests:
...
+  skip_if_only_changed: (^(\.tekton|\.github|\.claude|docs|examples|enhancements|contrib|\.cursor)/)|(^[A-Z]+\.md$)|((^|/)OWNERS$)|(/overrides\.yaml$)|(^renovate\.json$)|(/\.testcoverage\.yml$)|(^\.gitlint$)|(^\.gitignore$)
+  steps:
+    cluster_profile: hypershift
+    dependencies:
+      OCP_IMAGE_N1: release:n1minor
+      OCP_IMAGE_N2: release:n2minor
+      OCP_IMAGE_N3: release:n3minor
+      OCP_IMAGE_N4: release:n4minor
+    env:
+      AUTH_THROUGH_CERTS: "true"
```
mgencur, line 189: I suppose EXTERNAL_CNI_PROVIDER needs to be set here. Otherwise it won't run with Cilium.

ci-operator/step-registry/hypershift/azure/run-e2e/hypershift-azure-run-e2e-commands.sh
```diff
@@ -103,28 +104,41 @@ if [[ -n "${HYPERSHIFT_AZURE_MARKETPLACE_IMAGE_PUBLISHER:-}" && -n "${HYPERSHIFT
...
+    --e2e.latest-release-image="${OCP_IMAGE_LATEST}" \
+    --e2e.previous-release-image="${OCP_IMAGE_PREVIOUS}" \
+    ${cni_provider_arg} &
+  wait $!
+}
+
+# Run test with external CNI provider if specified
+if [[ -n "${EXTERNAL_CNI_PROVIDER:-}" ]]; then
+  run_e2e_test "--e2e.external-cni-provider=${EXTERNAL_CNI_PROVIDER}"
+fi
```
mgencur, line 136: Could you use the same approach like the rest of variables? For example, there's this (see below). The N1_NP_VERSION_TEST_ARGS is always passed to the command but sometimes it is empty, sometimes not.
``` 
N1_NP_VERSION_TEST_ARGS=""
if [[ ${OCP_IMAGE_N1} != "${OCP_IMAGE_LATEST}" ]]; then
  N1_NP_VERSION_TEST_ARGS="--e2e.n1-minor-release-image=${OCP_IMAGE_N1}"
fi
```
In our new case, it would be something like this:
```
EXTERNAL_CNI_PROVIDER_ARGS=""
if [[ -n ${EXTERNAL_CNI_PROVIDER:-} ]]; then
  EXTERNAL_CNI_PROVIDER_ARGS="--e2e.external-cni-provider=${EXTERNAL_CNI_PROVIDER}"
fi
```
And then it would be used like this:
```
hack/ci-test-e2e.sh -test.v \
  -test.run=${CI_TESTS_RUN:-} \
  -test.parallel=20 \
  --e2e.platform=Azure \
  --e2e.azure-credentials-file=/etc/hypershift-ci-jobs-azurecreds/credentials.json \
  --e2e.pull-secret-file=/etc/ci-pull-credentials/.dockerconfigjson \
  --e2e.base-domain=hypershift.azure.devcluster.openshift.com \
  --e2e.azure-location=${HYPERSHIFT_AZURE_LOCATION} \
  --e2e.oidc-issuer-url=${AZURE_OIDC_ISSUER_URL} \
  --e2e.sa-token-issuer-private-key-path=${AZURE_SA_TOKEN_ISSUER_KEY_PATH} \
    ${EXTERNAL_CNI_PROVIDER_ARGS:-} \ //<--------------------- HERE -------
    ${EXTERNAL_DNS_ARGS:-} \
    ${AKS_ANNOTATIONS:-} \
    ${N1_NP_VERSION_TEST_ARGS:-} \
    ${N2_NP_VERSION_TEST_ARGS:-} \
    ${N3_NP_VERSION_TEST_ARGS:-} \
    ${N4_NP_VERSION_TEST_ARGS:-} \
    ${MI_ARGS:-} \
    ${DP_ARGS:-} \
    ${AZURE_MULTI_ARCH_PARAMS:-} \
  --e2e.azure-encryption-key-id=${AKS_KMS_KEY} \
  --e2e.azure-kms-credentials-secret-name=${AKS_KMS_CREDENTIALS_SECRET} \
  ${MARKETPLACE_IMAGE_PARAMS} \
  --e2e.latest-release-image="${OCP_IMAGE_LATEST}" \
  --e2e.previous-release-image="${OCP_IMAGE_PREVIOUS}" &
```

```

**Review by mgencur — COMMENTED** [review]
Two minor comments (see inline). Otherwise I think it looks good.

With this job we'd be able to run Hypershift E2E test suite on AKS. That's fine. But there would be a mismatch between AKS and other platforms (AWS, Baremetal, Kubevirt) where we run Conformance tests for Cilium and Calico.
I will leave this to someone else to consider if this is ok. But later, if we want to get to the same level of support for AKS and other platforms we'd need to run the conformance tests as well. Thanks

File: ci-operator/config/openshift/hypershift/openshift-hypershift-main.yaml (hunk 1 above)
Comment (wewang58) [reply ci-operator/config/openshift/hypershift/openshift-hypershift-main.yaml:189]: Updated.

File: ci-operator/step-registry/hypershift/azure/run-e2e/hypershift-azure-run-e2e-commands.sh (hunk 2 above)
Comment (wewang58) [reply ci-operator/step-registry/hypershift/azure/run-e2e/hypershift-azure-run-e2e-commands.sh:136]: Updated, thanks reviewing.

File: ci-operator/step-registry/hypershift/azure/run-e2e/hypershift-azure-run-e2e-commands.sh [hunk 3]
```diff
@@ -102,6 +102,10 @@ MARKETPLACE_IMAGE_PARAMS=""
 if [[ -n "${HYPERSHIFT_AZURE_MARKETPLACE_IMAGE_PUBLISHER:-}" && -n "${HYPERSHIFT_AZURE_MARKETPLACE_IMAGE_OFFER:-}" && -n "${HYPERSHIFT_AZURE_MARKETPLACE_IMAGE_SKU:-}" && -n "${HYPERSHIFT_AZURE_MARKETPLACE_IMAGE_VERSION:-}" ]]; then
   MARKETPLACE_IMAGE_PARAMS="--e2e.azure-marketplace-publisher ${HYPERSHIFT_AZURE_MARKETPLACE_IMAGE_PUBLISHER} --e2e.azure-marketplace-offer ${HYPERSHIFT_AZURE_MARKETPLACE_IMAGE_OFFER} --e2e.azure-marketplace-sku ${HYPERSHIFT_AZURE_MARKETPLACE_IMAGE_SKU} --e2e.azure-marketplace-version ${HYPERSHIFT_AZURE_MARKETPLACE_IMAGE_VERSION}"
 fi
+EXTERNAL_CNI_PROVIDER_ARGS=""
+if [[ -n ${EXTERNAL_CNI_PROVIDER:-} ]]; then
```
mgencur, line 106: I suppose this also needs to atomatically set `--network-type=Other` ?
Otherwise the condition `if platform != hyperv1.NonePlatform && hostedCluster.Spec.Networking.NetworkType != hyperv1.Other {` here won't match and it will try to "validate" the cluster even though the CNI is Cilium  https://github.com/openshift/hypershift/pull/7423/files#diff-65fe136e12a7c2756931254c4f076ef0b5bc990b0fc13a8407647f421ddfa328R128

wewang58:
/pj-rehearse pull-ci-openshift-hypershift-main-e2e-aks-cilium

PR #71404, opened 2025-11-14
`ship-help-bot`: change build_root base image to version with python

smg247:
The image built in ci is having issues being used in the beta deployment. Hoping this will resolve it.


jupierce:
Looks like you will need `USER 0` before dnf . Alternatively, one of the ART build base images should work.

PR #71430, opened 2025-11-17
update ash job to make sure it can be ran in pre-submit jobs

sunzhaohua2:


ci-operator/config/openshift/cloud-provider-azure/openshift-cloud-provider-azure-main.yaml
```diff
@@ -123,10 +123,10 @@ tests:
   steps:
     cluster_profile: azure4
     workflow: openshift-e2e-azure
-- always_run: false
-  as: e2e-azurestack-ipi
+- as: e2e-azurestack-ipi
   cluster: build01
   optional: true
+  run_if_changed: ^(go\.mod|go\.sum)$
```
theobarberbany: Do we want to mirror what we do for the above azure e2e?? 

```yaml
skip_if_only_changed: ^docs/|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|PROJECT|LICENSE)$
```

sunzhaohua2: sure, if we want to test on ash as much as possible, just ash quota is small, the failure rate may be high, need us to check frequently.

theobarberbany: Hmm - I wasn't aware of this. Let's see if this is problematic with your most recent change, and revisit it if we hit regular issue :)

theobarberbany:
/pj-rehearse skip 

this is just changing when the job is run.

PR #71476, opened 2025-11-18
update installer fips presubmit jobs trigger

jianlinliu:
Sometimes coreos update impacts the fips install, e.g: https://redhat-internal.slack.com/archives/C01CQA76KMX/p1763388147238089,  so update the fips presubmit jobs trigger.


jianlinliu:
cc @gpei @yunjiang29 @patrickdillon to review.

PR #71504, opened 2025-11-18
MCO-1917: create prow steps to configure osimagestream in day1 and day2

sergiordlr:
Prow steps to configure the osImageStream in the cluster for both day1 and day2

It is not decided yet how day1 config will work. We only created and scaffold that will be modified later.

### DIFF
--- a/ci-operator/step-registry/mco/conf/day1/os-image-stream/mco-conf-day1-os-image-stream-commands.sh
+++ b/ci-operator/step-registry/mco/conf/day1/os-image-stream/mco-conf-day1-os-image-stream-commands.sh
@@ -1,0 +1,85 @@
+#!/bin/bash
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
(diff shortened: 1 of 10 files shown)


ci-operator/step-registry/mco/conf/day1/os-image-stream/mco-conf-day1-os-image-stream-commands.sh
```diff
@@ -0,0 +1,85 @@
...
+  echo "MCO_CONF_DAY1_OS_IMAGE_STREAM_NAME is empty. No MachineConfigPool manifests will be created."
+  exit 0
+fi
+
+function create_mcp_manifest() {
+  local manifests_dir="${1}"
+  local mcp_name="${2}"
+  local stream_name="${3}"
+
+  local manifest_file="${manifests_dir}/manifest_mcp-${mcp_name}-osimagestream.yaml"
```
pablintino, line 22: Without day-zero **installer** support the file name is critical. Please check https://gist.github.com/pablintino/15b6e900b47aead5983a161cfb9b6ed9

PR #71517, opened 2025-11-18
Limit e2e suite parallel runs to 1 given environment constraints.

roivaz:
The ARO-HCP integration environment is limited to 20 HCP clusters; in practice, this means only one e2e suite can run simultaneously.

### DIFF
--- a/core-services/prow/02_config/_boskos.yaml
+++ b/core-services/prow/02_config/_boskos.yaml
@@ -42,6 +42,10 @@ resources:
 - max-count: 1
   min-count: 1
   state: free
+  type: aro-hcp-int-e2e-quota-slice
+- max-count: 1
+  min-count: 1
+  state: free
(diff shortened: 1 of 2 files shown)


core-services/prow/02_config/_boskos.yaml
stevekuznetsov, line 1: @roivaz ptal

roivaz: As per the docs https://docs.ci.openshift.org/docs/architecture/quota-and-leases/#adding-a-new-type-of-resource my understanding is that this is generated by the user and checked in.

stevekuznetsov: Ah sorry I missed that you changed the generator config too

stevekuznetsov (approved):

PR #71519, opened 2025-11-18
mao - basic vsphere jobs to non-optional

jcpowermac:
With the addition of ote in mao we should test
vsphere as non-optional.


mtulio (commented):
/lgtm

PR #71528, opened 2025-11-18
TRT-2388: Add variant-periodic commands

neisw:
Adds slash commands to help 

- /find-missing-variant-periodics: to compare the defined periodic variant configurations between two releases and identify any that are missing.
- /migrate-variant-periodics: to migrate any periodic variant configurations from one release to another that do not already have an existing configuration for the target release

### DIFF
--- a/.claude/commands/find-missing-variant-periodics.md
+++ b/.claude/commands/find-missing-variant-periodics.md
@@ -17,7 +17,7 @@ You are helping users identify OpenShift periodic CI job configurations that exi
 
 ## Context
 
-Periodic jobs are CI tests that run on a schedule (via cron) rather than on pull requests. These are defined in files following the pattern: `*-release-{major.minor}__periodics.yaml` in the `ci-operator/config/` directory.
+Periodic jobs are CI tests that run on a schedule (via cron) or after a specified interval since last execution rather than on pull requests. These are defined in files following the pattern: `*-release-{major.minor}__periodics.yaml` in the `ci-operator/config/` directory.
 
 This is a **read-only analysis command** - it doesn't modify any files. Use it before `/migrate-variant-periodics` to understand the scope of work.
 
@@ -146,8 +146,6 @@ All periodic configurations for {from_version} have corresponding {to_version} c
 # Find missing periodics for an organization
 /find-missing-variant-periodics 4.19 4.20 ci-operator/config/openshift
 
-# Check openshift-priv repositories
-/find-missing-variant-periodics 4.17 4.18 ci-operator/config/openshift-priv
 ```
 
 ## Example Output
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -44,7 +44,7 @@ This runs: `make jobs`, `make ci-operator-config`, `make prow-config`, `make reg
 ### CI Job Configuration
 When modifying CI jobs in `ci-operator/config/`:
 ```bash
-make jobs
+make update
 ```
 This validates config, generates Prow job configs, and sanitizes job definitions.
 
@@ -84,7 +84,7 @@ Regenerates release controller configurations.
    - `tests`: Test definitions (can reference step-registry workflows)
    - `promotion`: Where to push successful builds
    - `releases`: OpenShift releases to test against
-3. Running `make jobs` generates Prow job configs in `ci-operator/jobs/`
+3. Running `make update` generates Prow job configs in `ci-operator/jobs/`
 4. Prow triggers jobs based on PR/merge/periodic events
 5. Jobs execute ci-operator with the config, which orchestrates builds and tests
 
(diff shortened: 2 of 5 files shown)


neisw:
/hold
for https://github.com/openshift/release/pull/70987

CLAUDE.md
```diff
@@ -115,6 +140,44 @@ Steps are referenced in `ci-operator/config/` test definitions using multi-stage
 3. Define `pre`, `test`, and `post` phases using existing steps/chains
 4. Run `make validate-step-registry` to check correctness
 
+### Creating Variant Periodic Configurations
+For repositories with many periodic tests, separate them from the main config:
```
petr-muller: This is not a good general guidance. Simply having many periodic tests is not a good reason to have a split config; configs are not just `tests` and having a split config means you likely have a duplicate content which typically needs to be kept in sync. So it's mostly a cost, not a benefit.

There are two use cases for having a separate `branch__periodics.yaml` variant file:

1. The periodic jobs somehow use a different CI config than the base config. Stuff like different base images, different releases involved
2. Some automation owns the base configuration file and therefore you cannot add your desired periodics there, because your changes would be overwritten.

The latter is the TRT relevant case. People cannot add jobs to standard `release-4.21.yaml` files because they would be overwritten with content derived from `main.yaml`. So we invented a crutch where people maintain their TRT-releavant periodics in `release-4.21__periodics.yaml` because nothing will overwrite them there.

(side note, an interesting thing is that after the branch config *stops* being managed by the automation -- post branch cut -- we could even merge the two config files again to make repo content simpler.)

CLAUDE.md
```diff
@@ -115,6 +140,44 @@ Steps are referenced in `ci-operator/config/` test definitions using multi-stage
...
+1. Create `ci-operator/config/<org>/<repo>/<org>-<repo>-<branch>__periodics.yaml`
+2. Include `base_images`, `build_root`, `images`, `promotion`, `releases` sections (copy from main config)
+3. Add only `tests:` entries with `interval:` (periodic scheduling)
+4. Ensure `zz_generated_metadata` has `variant: periodics`
```
petr-muller: I'd recommend telling the bot to never touch `zz_generated_metadata`. It will just fight with `make update`

CLAUDE.md
```diff
@@ -115,6 +140,44 @@ Steps are referenced in `ci-operator/config/` test definitions using multi-stage
...
+5. Run `make jobs` to generate Prow jobs in separate `-periodics.yaml` file
```
petr-muller: Shall we generally tell all bots to always run full `make update` to never run into random "content out of sync" problems?

CLAUDE.md
```diff
@@ -115,6 +140,44 @@ Steps are referenced in `ci-operator/config/` test definitions using multi-stage
...
+6. This pattern keeps main config focused on PR/postsubmit jobs while isolating scheduled tests
```
petr-muller: Same as https://github.com/openshift/release/pull/71528/files#r2542016330

CLAUDE.md
```diff
@@ -115,6 +140,44 @@ Steps are referenced in `ci-operator/config/` test definitions using multi-stage
...
+
+## OpenShift Release Versioning
+
+### Release Branches
+OpenShift uses semantic versioning with minor releases (4.18, 4.19, 4.20, 4.21, 4.22, etc.):
+- **Branch names**: `release-4.21`, `release-4.20`, etc. (or `main`/`master`)
+- **Config files**: One per branch per repository (e.g., `openshift-oauth-server-release-4.21.yaml`)
+- **Release cycle**: New minor versions branch quarterly
```
petr-muller: quarterly is not correct, maybe we do not need to make any assumptions here or just very general ones

CLAUDE.md
```diff
@@ -115,6 +140,44 @@ Steps are referenced in `ci-operator/config/` test definitions using multi-stage
...
+
+### Config Brancher Tool
+The `config-brancher` tool automates creating CI configs for new releases:
+```bash
+# Example from recent commits
+config-brancher --config-dir ./ci-operator/config --current-release 4.21 --future-release 4.22 --confirm
+```
+This copies configs from current to future release, updating version references.
```
petr-muller: I can imagine giving more context about the tool may help the bot in understanding the ci config structure.

The source content for `config-brancher` is always the current development branch, such as master/main. The `--current-release 4.21` flag tells the tool "main development branch currently targets 4.21". It means the tool takes the content from the main config (assumed to be managed by humans) and copies it to the the release-4.21 config (assumed to not be maintained by humans at this time because there is also an assumption that the release git branch is being fast-forwarded to the main branch content). The `--future-release 4.21` param tells it to *also* maintain an additional config for another future branch. The tool makes sure the versions in the copied configs are modified appropriately.

CLAUDE.md
```diff
@@ -115,6 +140,44 @@ Steps are referenced in `ci-operator/config/` test definitions using multi-stage
...
+
+Use `--skip-periodics` to avoid branching periodic jobs (they're often managed separately in `__periodics.yaml` files).
```
petr-muller, line 167: Here we could tell the agent that teams typically only have periodics set up for master, if they actually want to run jobs for master code. The code in both managed release branches is identical so there is no point in having copies of these periodics; they build and test the same code. That's why `--skip-periodics` exists.

The difference with TRT-facing periodics is that they should _only_ exist in release branch CI configs (we need them this way to be able to recognize they are tied to a specific release in TRT tools). We cannot put them in `main` config so that config-brancher puts them to release branch config, because we actually do not want them to run for `main`. And we cannot put them directly in release branch CI configs, because that's owned by config-brancher enforcing that the release branch config is derived from main config.

.claude/commands/find-missing-variant-periodics.md
```diff
@@ -0,0 +1,186 @@
...
+  - name: path
+    description: Optional directory path to search for periodic files (default is ci-operator/config/)
+    required: false
+---
+
+You are helping users identify OpenShift periodic CI job configurations that exist for one release but are missing for another.
+
+## Context
+
+Periodic jobs are CI tests that run on a schedule (via cron) rather than on pull requests. These are defined in files following the pattern: `*-release-{major.minor}__periodics.yaml` in the `ci-operator/config/` directory.
```
petr-muller: Maybe this also needs to mention `interval`-based jobs

```suggestion
Periodic jobs are CI tests that run on a schedule (via cron) or after a specified interval since last execution rather than on pull requests. These are defined in files following the pattern: `*-release-{major.minor}__periodics.yaml` in the `ci-operator/config/` directory.
```

CLAUDE.md
neisw: > So we invented a crutch where people maintain their TRT-releavant periodics in release-4.21__periodics.yaml because nothing will overwrite them there

This is so informative... Thank you!

neisw: Added updates about our CI analytical tooling requiring release specific config and how that conflicts with default config main fast-fowarding.  Really insightful feedback here, thanks again!

PR #71531, opened 2025-11-18
ci-operator: Azure/ARO-HCP: update jobs

stevekuznetsov:
This commit combines a few changes:
- after chatting with David Eads, we realized the create-functional-cluster job and the associated steps are deprecated and can be removed, along with the job to run e2e without cleanup
- we should have a workflow for the e2e runs, so we don't copy-paste the chains we're using. Officially, if we wanted to use a cluster_profile to inject leases, credentials/env, and all that, we could add our profiles to the ci-tools repository, but that seems like a lot of work and still would not support our complex leasing setup
- new quota slices are added - one for the future test tenant, one for the environments we're testing, and one for the Red Hat tenant. We expect this to work as such:
  - a job touching a MSFT environment will aquire two quota slices - one for the environment and one for the test tenant. This is how we can ensure that e.g. INT supports only one concurrent test at the same time as we can ensure that the test tenant in which we run ARO HCP clusters is not overloaded
  - a job for local dev will use the Red Hat tenant quota slice, we might want to separate this out if some local e2e require a lot more quota than others, but it can be one for now - the size of this slice will depend on how many concurrent ARO HCP test regions + ARO HCP test clusters the Red Hat tenant can support
- NOTE: all quota slice quantities are wild-ass-guesses. We can adapt as we need to
- removed the periodic jobs, as we run a job in every Ev2 rollout now


core-services/prow/02_config/generate-boskos.py
```diff
@@ -246,6 +246,18 @@
...
+    'aro-hcp-stg-e2e-quota-slice': {
+        'default': 1,
+    },
+    'aro-hcp-prod-e2e-quota-slice': {
+        'default': 1,
+    },
+    'aro-hcp-test-tenant-quota-slice': {
+        'default': 10,
+    },
+    'aro-hcp-redhat-tenant-quota-slice': {
```
venkateshsredhat: does this needs to match https://github.com/openshift/ci-tools/pull/4830/files#diff-dfee74d02fd350fb1ee1ffbbbc32acf4cb075c9ae42b6e31c8c7acefe7fed623R2366  . "aro-hcp-dev-quota-slice" .

stevekuznetsov:
/pj-rehearse pull-ci-Azure-ARO-HCP-main-integration-e2e-parallel

stevekuznetsov:
/pj-rehearse pull-ci-Azure-ARO-HCP-main-integration-e2e-parallel pull-ci-Azure-ARO-HCP-main-e2e-parallel

weherdh:
Hey Steve, I have been requested to add myself as the owner. Could you please also add me as the necessary approvers & owners?

ci-operator/config/Azure/ARO-HCP/Azure-ARO-HCP-main__periodic.yaml
```diff
@@ -12,176 +12,55 @@ images:
...
-  steps:
-    allow_best_effort_post_steps: true
-    env:
-      ARO_HCP_SUITE_NAME: prod/parallel
-      CUSTOMER_SUBSCRIPTION: ARO Hosted Control Planes (EA Subscription 1)
-      LOCATION: uksouth
-    post:
-    - ref: aro-hcp-provision-aro-hcp-gather-extra
-    - chain: aro-hcp-provision-teardown-cluster
+    cluster_profile: aro-hcp-stg
```
weherdh, line 55: I checked the vault, so you have created the vault for the staging and prod? So later if I want to move staging and prod e2e test to the new tenant, I just need to update the vault?

weherdh: Or I can assign the secret name which allows me to test it first

stevekuznetsov: Exactly, yes!

PR #71560, opened 2025-11-19
Fix partition table clearing to wipe all non-excluded devices

ccardenosa:
**Root cause discovered:**
    - /dev/sdb had hidden partition table signatures that lsblk couldn't detect
    - vgcreate could detect these signatures and refused to use the disk
    - The script was only wiping devices with visible partitions in lsblk
    - This caused vg-manager to stay in 0/1 Ready state for days
    
**Solution:**
    1. Wipe ALL non-excluded block devices (not just those with visible partitions)
       - vgcreate can detect partition signatures that lsblk misses
       - Better to be aggressive with wiping since LVM excludes what it needs
    2. Use dd to zero first 1MB (in addition to sfdisk --delete)
       - Ensures all partition table signatures are cleared
       - **_sfdisk --delete alone wasn't sufficient in this case_**
    3. Reduced sleep time from 3s to 2s (minor optimization)
    
**Testing:**
      - Manually wiped /dev/sdb on production cluster
      - vg-manager immediately created volume group
      - CSI plugin registered successfully
      - LVMCluster became Ready within 15 seconds
    
This fix ensures the script will properly clean disks even when partition table signatures are hidden from standard tools.

### DIFF
--- a/ci-operator/step-registry/telcov10n/metal-single-node-spoke/site-config/deploy/telcov10n-metal-single-node-spoke-site-config-deploy-commands.sh
+++ b/ci-operator/step-registry/telcov10n/metal-single-node-spoke/site-config/deploy/telcov10n-metal-single-node-spoke-site-config-deploy-commands.sh
@@ -102,21 +102,14 @@ function clear_partition_disk_table {
       devices_to_wipe="${devices_to_wipe} ${dev}"
     done
   else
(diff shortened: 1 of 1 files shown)


ccardenosa:
/pj-rehearse periodic-ci-openshift-kni-eco-ci-cd-main-nightly-4.21-telcov10n-metal-single-node-spoke

PR #69775, opened 2025-09-28
TRT-2358: enhancement for Symptom Detection testing

jianlinliu:
1. move the content of symptom.sh from gist to the gather-extra shell script for easy maintenance.
2. update Infrastructure quota/limit detection testing

PR #71610, opened 2025-11-20
OCPBUGS-65670: loosen infrastructure check in symptoms detecting

jianlinliu:
A side effect introduced by https://github.com/openshift/release/pull/69775, in some case, when cluster operators hit rate limiting, later it recovered from it, the infrastructure detect will catch the rate limiting, and fail the test. 
Now introduce a condition, only when install fail, the detected infrastructure issue would fail the test.


jianlinliu:
I am using a fake pattern to test the condition, looks it works well, refer to [here](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/71610/rehearse-71610-periodic-ci-openshift-release-master-ci-4.21-e2e-aws-ovn-techpreview/1991368057288658944).

PR #71614, opened 2025-11-20
TRT-2337: drop old release controllers for EOL OCP and the referenced old template workflows

jianlinliu:
Dropping the old enough release controllers (4.1 - 4.6, they were EOL a long time ago) that are referencing old template-style install workflows.


petr-muller (approved):
Thanks! I was mostly just interested whether we're going with gut feeling about really really old OCPs, or if we have some known reason to keep e.g. 4.7. No problem with the chosen approach, just wanted to see it written down.

PR #71624, opened 2025-11-20
OVNK-BGP e2e tests: wait for all the frrk8s deployments

fedepaol:
Instead of waiting specifically for the webhook deployment to be available, we wait for all the deployments in the namespace. This allows the script to be future proof and to work with both the current implementaion (frr-k8s-webhook-server deployment) and the new upstream implementation where the webhook is hosted in the node status cleaner pod.

### DIFF
--- a/ci-operator/step-registry/baremetalds/e2e/ovn/bgp/pre/baremetalds-e2e-ovn-bgp-pre-commands.sh
+++ b/ci-operator/step-registry/baremetalds/e2e/ovn/bgp/pre/baremetalds-e2e-ovn-bgp-pre-commands.sh
@@ -272,8 +272,8 @@ until oc rollout status daemonset -n openshift-frr-k8s frr-k8s --timeout 2m &> /
   sleep 5
 done
 
-echo "Waiting for deploy 'frr-k8s-webhook-server' to be created..."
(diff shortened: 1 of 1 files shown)


fedepaol:
@jcaamano ptal, the 4.20 rehearse succeded

ci-operator/step-registry/baremetalds/e2e/ovn/bgp/pre/baremetalds-e2e-ovn-bgp-pre-commands.sh
```diff
@@ -272,8 +272,8 @@ until oc rollout status daemonset -n openshift-frr-k8s frr-k8s --timeout 2m &> /
   sleep 5
 done
 
-echo "Waiting for deploy 'frr-k8s-webhook-server' to be created..."
-until oc wait -n openshift-frr-k8s deployment frr-k8s-webhook-server --for condition=Available --timeout 2m &> /dev/null; do
+echo "Waiting for all deployments in openshift-frr-k8s namespace to be created..."
+until oc wait -n openshift-frr-k8s deployment --all --for condition=Available --timeout 2m &> /dev/null; do
```
jcaamano, line 276: So my concern here is that it won't wait for deployments that are to be added and don't yet exist.

But I guess we can assume that by the time the daemonset has been rolled out, deployments should already have been added.

PR #71666, opened 2025-11-21
frontend and backend image override

venkateshsredhat:


ci-operator/step-registry/aro-hcp/test/local/aro-hcp-test-local-commands.sh
```diff
@@ -23,7 +23,7 @@ REMOTE_PORT=8443
 
 wait_for_service() {
     for i in {1..5}; do
-        if kubectl get svc -n "$NAMESPACE" "$SERVICE" >/dev/null 2>&1; then
+        if oc get svc -n "$NAMESPACE" "$SERVICE" >/dev/null 2>&1; then
```
stevekuznetsov, line 26: Mind explaining why you made the switch?

ci-operator/step-registry/aro-hcp/provision/environment/aro-hcp-provision-environment-commands.sh
```diff
@@ -2,7 +2,6 @@
 set -o errexit
 set -o nounset
 set -o pipefail
-set -o xtrace
```
stevekuznetsov, line 5: You will regret this - remove at your own risk :)

venkateshsredhat: it kind of leaked a token from one of the az command while testing , then had the Release team remove the log . 
Though its masked now but removed because I added it to debug something .

venkateshsredhat: I understand the concern though .

ci-operator/step-registry/aro-hcp/test/local/aro-hcp-test-local-commands.sh
venkateshsredhat: we use oc registry login for creds during mirroring process . So makes sense to keep one tool . OC extends kubectl so access to both worlds .

ci-operator/step-registry/aro-hcp/provision/environment/aro-hcp-provision-environment-commands.sh
venkateshsredhat: I will add this back if required , I have another PR coming for the base image .

stevekuznetsov: We wrote code that made sure that any secret logged by the jobs is changed to `XXX...` - it should in theory be impossible to leak anything ....

venkateshsredhat: but unfortunately it leaked and it was an access token :/ , but Release team added it the list and it was masked after that .

PR #71688, opened 2025-11-21
Update ODH operator to Go 1.25

RomanFilip:
Related to https://github.com/opendatahub-io/opendatahub-operator/pull/2885


zdtsw:
it might be the chicken-egg case here @grdryn

PR #71706, opened 2025-11-21
openshift-test-private master branch is now 4.22

tbuskey:
The tests-private.tag should be what master is because OSC always uses the master branch.  We cannot set it to "master" so we advance the version number from 4.21 to 4.22

Without this, OSC does not run our latest tests


tbuskey:
I think it's not switched yet.

PR #71736, opened 2025-11-24
Move the gcp-ipi-custom-endpoints-proxy-whitelist test as individual installation test

gpei:
Because there isn't a dedicated URL allowlist specifically for the e2e test section, so move the testing as dedicated installation test.


gpei:
cc @liangxia for the approval, thanks.

liangxia (approved):

PR #71746, opened 2025-11-24
telco: NROP: build Dockerfile.mustgather as is

shajmakh:
ubi-minimal doesn't have rsync and other binaries required by must-gather command (namely `oc adm inspect`). However the dedicated dockerfile runs with openshift/origin-cli base which contains these bins so use the same u/s configuration without replacing the base.


shajmakh:
it's likely chicken-egg problem, we need to have https://github.com/openshift-kni/numaresources-operator/pull/2572

Tal-or:
/pj-rehearse ack
Took the liberty to add this label, we'll evaluate the test on NROP repository side.

PR #71817, opened 2025-11-25
USHIFT-6343: microshift build-cache jobs fail to download brew RPMs

agullon:


ci-operator/step-registry/openshift/microshift/infra/iso-build/openshift-microshift-infra-iso-build-commands.sh
```diff
@@ -42,33 +42,43 @@ chmod +x /tmp/iso.sh
...
+
+    # Check if if manage_brew_rpms.sh script exists
+    if ! [ -e ./test/bin/manage_brew_rpms.sh ] ; then
+        echo "ERROR: ./test/bin/manage_brew_rpms.sh not found - RPM download from brew is not possible"
+        exit 1
+    fi
+    # Check if manage_brew_rpms.sh script supports the version_type option
+    if ! ./test/bin/manage_brew_rpms.sh -h | grep -q 'version_type' ; then
+        echo "ERROR: ./test/bin/manage_brew_rpms.sh does not support the version_type option - RPM download from brew is not possible"
+        exit 1
```
pacevedom: I think we should exit with 0 instead. This repo is branch-less, which means this script impacts every job that executes it. In case of MicroShift, this includes old releases where we didnt have the `version_type` option yet.
If we exit with 1 then the whole step fails and nothing will be cached.

agullon: I agree,

```suggestion
        exit 0
```

agullon:
/pj-rehearse pull-ci-openshift-microshift-main-e2e-aws-tests-cache-arm pull-ci-openshift-microshift-main-e2e-aws-tests-cache pull-ci-openshift-microshift-release-4.20-e2e-aws-tests-cache-arm pull-ci-openshift-microshift-release-4.20-e2e-aws-tests-cache

ci-operator/step-registry/openshift/microshift/infra/iso-build/openshift-microshift-infra-iso-build-commands.sh
```diff
@@ -42,33 +42,43 @@ chmod +x /tmp/iso.sh
...
-            LATEST_RELEASE_TYPE="ec"
-        fi
-        # Warn on failures, as there are periods of time where a specific release type RPM for a release may not be available.
-        bash -x ./test/bin/manage_brew_rpms.sh download "4.${y_version}" "${out_path}" "${LATEST_RELEASE_TYPE}" || echo "WARNING: Failed to download ${LATEST_RELEASE_TYPE} RPMs for 4.${y_version}"
-        bash -x ./test/bin/manage_brew_rpms.sh download "4.${y_version}" "${out_path}" "nightly" || echo "WARNING: Failed to download nightly RPMs for 4.${y_version}"
-        bash -x ./test/bin/manage_brew_rpms.sh download "4.$((${y_version} - 1))" "${out_path}" "zstream" || echo "WARNING: Failed to download zstream RPMs for 4.$((${y_version} - 1))"
-        bash -x ./test/bin/manage_brew_rpms.sh download "4.$((${y_version} - 2))" "${out_path}" "zstream"
+
+    # Check if if manage_brew_rpms.sh script exists
+    if ! [ -e ./test/bin/manage_brew_rpms.sh ] ; then
```
pacevedom: Just realized this is also an exit 0 condition. In those branches where the file does not exist we need to skip all this entirely. The previous approach was only doing it if the file existed and was ready to act based on `version_type` in help.
Can we do early exit 0 here too?

agullon: Yes, it makes sense, change it in 34431c2650c3ed937bc2187078e1eaf7fc1b180b commit

PR #71819, opened 2025-11-25
Add configs for build05

liangxia:
Configs for the newly provisioned build05


droslean (approved):

liangxia:
The job `build05-dry` won't pass until we merge this PR, because it will not sync the `sa.config-updater.build05.config` which is added in this PR until PR is merged.

PR #71868, opened 2025-11-26
add a better retry mechanism for port-forwarding

venkateshsredhat:


ci-operator/step-registry/aro-hcp/test/local/aro-hcp-test-local-commands.sh
```diff
@@ -15,53 +14,104 @@ az account set --subscription "${SUBSCRIPTION_ID}"
...
+            echo "Service $SERVICE found"
+            return 0
+        else
+            echo "Service $SERVICE not found (attempt $i/5)"
+            [[ $i -lt 5 ]] && sleep 10
+        fi
+    done
+    echo "Service not available after 5 attempts, exiting"
+    exit 1
+}
```
roivaz, line 36: Waiting just for the Service to exist doesn't guarantee that it actually has something behind ready to accept connections. We should either check the Endpoint resources associated with the Service

```bash
  for i in {1..30}; do
    if kubectl get svc -n "${NAMESPACE}" "${SERVICE}" >/dev/null 2>&1 &&
       [[ -n "$(kubectl get endpoints -n "${NAMESPACE}" "${SERVICE}" \
                 -o jsonpath='{.subsets[*].addresses[*].ip}')" ]]; then
      echo "Service ${SERVICE} has ready endpoints"
      return 0
    fi
    echo "Waiting for ${SERVICE} endpoints (attempt ${i}/30)..."
    sleep 5
```

or just directly wait for the deployment to be available (might be easier)

```bash
kubectl wait -n "${NAMESPACE}" \
  --for=condition=Available \
  deployment/${DEPLOYMENT} \
  --timeout=300s
```

roivaz: Merging this in its current state. We can add this at a later time, not super important.

roivaz (approved):

PR #71870, opened 2025-11-26
OTA-1721 Adding Claude Subagent for creating stream upgrade jobs

JianLi-RH:
This PR added a new sub-agent for creating stream upgrade jobs.

We can go to ci-operator/config/openshift/openshift-tests-private folder then ask Claude to create stream upgrade jobs, for example:
```
> cd ci-operator/config/openshift/openshift-tests-private
> Claude
> Please create Y stream and Z stream upgrade jobs for OCP 4.22
```

Subagents reference: https://code.claude.com/docs/en/sub-agents

**Advantage** of sub-agent is that we don't need to load it like loading skills, we can use them directly in a project.

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/.claude/scripts/create_stream_upgrades.sh
+++ b/ci-operator/config/openshift/openshift-tests-private/.claude/scripts/create_stream_upgrades.sh
@@ -1,0 +1,51 @@
+#!/bin/bash
+
+set -e
+
+TARGET_VERSION="$1"
+
+
+PRIOR_VERSION=$(.claude/scripts/get_prior_minor_version.sh "$TARGET_VERSION")
+Y_PRIOR_VERSION=$(.claude/scripts/get_prior_minor_version.sh "$PRIOR_VERSION")
+
+# Find all upgrade files for prior version
+for file in openshift-openshift-tests-private-release-"$PRIOR_VERSION"__*-"$PRIOR_VERSION"-upgrade-from-stable-*.yaml; do
+  # Get the initial version for Y stream and Z stream upgrade configuration files
+  if echo "$file" | grep "${PRIOR_VERSION}.yaml" > /dev/null; then
+    initial_version=${PRIOR_VERSION}
+  elif echo "$file" | grep "${Y_PRIOR_VERSION}.yaml" > /dev/null; then
+    initial_version=${Y_PRIOR_VERSION}
+  else
+    continue
+  fi
+
+  # Extract the initial version from filename (e.g., 4.20, 4.21)
+  initial_major=$(echo "$initial_version" | cut -d. -f1)
+  initial_minor=$(echo "$initial_version" | cut -d. -f2)
+
+  # Calculate new initial version (add 1 to minor)
+  new_initial_minor=$((initial_minor + 1))
+  new_initial_version="${initial_major}.${new_initial_minor}"
+
+  # Create new filename (replace 4.21 with 4.22, and initial version with new initial version)
+  new_file=$(echo "$file" | 
+    sed "s/release-${PRIOR_VERSION}/release-${TARGET_VERSION}/g" | 
+    sed "s/-${PRIOR_VERSION}-upgrade/-${TARGET_VERSION}-upgrade/g" | 
+    sed "s/stable-${initial_version}/stable-${new_initial_version}/g"
+  )
+
+  # Extract architecture/stream from original filename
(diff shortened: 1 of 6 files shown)


ci-operator/config/openshift/openshift-tests-private/.claude/agents/create_YZ_stream_upgrade_jobs.md
```diff
@@ -0,0 +1,33 @@
+---
+name: create_YZ_stream_upgrade_jobs
+description: Prow CI expert maintainer. Can find existing Prow CI configuration files and copy existing Y stream and Z stream upgrade job configuration files for the next version. Updates new file names and content for the next version.
+tools: Bash,Edit,Glob,Grep,Read,Write
```
jhou1: nit: leave a space after every comma

ci-operator/config/openshift/openshift-tests-private/.claude/agents/create_YZ_stream_upgrade_jobs.md
```diff
@@ -0,0 +1,33 @@
+---
+name: create_YZ_stream_upgrade_jobs
```
jhou1: suggestion, what about a simple name instead? The upgrade jobs are Y/Z streams so naming as `YZ_stream` is redundant. what about `create_upgrade_jobs`?

JianLi-RH: I will create three more Subagents: `create_cpou_upgrade_jobs`, `create_chain_upgrade_jobs`, `create_rollback_jobs`.
So `create_YZ_stream_upgrade_jobs` is more accurate, WDTY?

ci-operator/config/openshift/openshift-tests-private/.claude/agents/create_YZ_stream_upgrade_jobs.md
JianLi-RH: done

ci-operator/config/openshift/openshift-tests-private/.claude/agents/create_YZ_stream_upgrade_jobs.md
jhou1: OK, nit: be consistent with your naming pattern, following snake cases. YZ -> yz. They don't have to be capitalized in file names

ci-operator/config/openshift/openshift-tests-private/.claude/scripts/create_stream_upgrades.sh
```diff
@@ -0,0 +1,51 @@
...
+
+  # Copy and update file content
+  sed -e "s/\"${PRIOR_VERSION}\"/\"${TARGET_VERSION}\"/g" \
+      -e "s/\"${initial_version}\"/\"${new_initial_version}\"/g" \
+      -e "s/branch: release-${PRIOR_VERSION}/branch: release-${TARGET_VERSION}/g" \
+      -e "s/variant: .*-${PRIOR_VERSION}-upgrade-from-stable-${initial_version}/variant: ${arch_stream}-${TARGET_VERSION}-upgrade-from-stable-${new_initial_version}/g" \
+      "$file" > "$new_file"
+
+  # Update cron settings
+  tools/update-cron-entries.py --backup no "$new_file"
```
jhou1, line 48: Let's ensure the subagent handle the python pip dependency as well, because your script is calling this file here.

jhou1:
It's functional, looking good. I left a few suggestions.

ci-operator/config/openshift/openshift-tests-private/.claude/scripts/create_stream_upgrades.sh
JianLi-RH: I ran the script manually and compared it with the AI ​​results, they are identical

jhou1: Because your env has installed everything, does not work for someone who has not configured `ruamel.yaml`
Would be useful to tell subagent to run `pip install` in case the script fails and it determines to solve the problem other way round

JianLi-RH: ah, good catch.
I think simply reporting an error is sufficient because only the CI maintainers will run this sub-agents.
Let me try it without the env var.

PR #71877, opened 2025-11-26
fix: add dependency to e2e job

davidebianchi:


CFSNM (approved):
/approved /lgtm

ci-operator/config/opendatahub-io/opendatahub-operator/opendatahub-io-opendatahub-operator-rhoai.yaml
```diff
@@ -55,6 +55,8 @@ tests:
     allow_best_effort_post_steps: true
     allow_skip_on_success: true
     cluster_profile: gcp-opendatahub
+    dependencies:
+      OO_INDEX: opendatahub-operator-rhoai-bundle
```
zdtsw, line 59: OO_BUNDLE ?

davidebianchi: I see that for the workflow we are using here (`optional-operators-ci-gcp`), it depends on [`OO_INDEX`](https://github.com/openshift/release/blob/master/ci-operator/step-registry/optional-operators/subscribe/optional-operators-subscribe-ref.yaml#L36-L38), instead of [`OO_BUNDLE`](https://github.com/openshift/release/blob/master/ci-operator/step-registry/optional-operators/operator-sdk/optional-operators-operator-sdk-ref.yaml#L34-L36) dependency of the `optional-operators-ci-operator-sdk-gcp` used by main.

If I well interpreted that 😄 

We could also move to use the same workflow used in main, what do you think?

zdtsw: do i understand it for now:
- odh main, we build with bundle+srccat
-  odh rhoai, we build with bundle+fbc

but in about 2 months, when we shift odh main to konflux, that need to be switch to
- odh main and odh rhoai, both bundle+fbc

PR #71880, opened 2025-11-26
adding workflow for ztunnel

zmiklank:
Aim of this workflow:
1) build ztunnel on PR
2) push to google cloud
3) use in istio midstream testing.

Depends on: 
https://github.com/openshift-service-mesh/ztunnel/pull/14
https://github.com/maistra/test-infra/pull/430

### DIFF
--- a/ci-operator/config/openshift-service-mesh/ztunnel/openshift-service-mesh-ztunnel-release-1.27.yaml
+++ b/ci-operator/config/openshift-service-mesh/ztunnel/openshift-service-mesh-ztunnel-release-1.27.yaml
@@ -23,7 +23,6 @@ resources:
 tests:
 - as: cargo-build
   steps:
-    cluster_profile: ossm-aws
     test:
     - as: build
       cli: latest
@@ -39,7 +38,6 @@ tests:
 - as: cargo-build-and-push
   postsubmit: true
   steps:
-    cluster_profile: ossm-aws
     test:
     - as: build-and-push
       cli: latest
--- a/ci-operator/jobs/openshift-service-mesh/ztunnel/openshift-service-mesh-ztunnel-release-1.27-postsubmits.yaml
+++ b/ci-operator/jobs/openshift-service-mesh/ztunnel/openshift-service-mesh-ztunnel-release-1.27-postsubmits.yaml
@@ -10,8 +10,6 @@ postsubmits:
       skip_cloning: true
       timeout: 2h0m0s
     labels:
-      ci-operator.openshift.io/cloud: aws
-      ci-operator.openshift.io/cloud-cluster-profile: ossm-aws
       ci.openshift.io/generator: prowgen
     max_concurrency: 1
     name: branch-ci-openshift-service-mesh-ztunnel-release-1.27-cargo-build-and-push
@@ -20,7 +18,6 @@ postsubmits:
       - args:
         - --gcs-upload-secret=/secrets/gcs/service-account.json
         - --image-import-pull-secret=/etc/pull-secret/.dockerconfigjson
-        - --lease-server-credentials-file=/etc/boskos/credentials
         - --report-credentials-file=/etc/report/credentials
         - --secret-dir=/secrets/ci-pull-credentials
         - --target=cargo-build-and-push
@@ -33,9 +30,6 @@ postsubmits:
           requests:
             cpu: 10m
         volumeMounts:
-        - mountPath: /etc/boskos
-          name: boskos
-          readOnly: true
         - mountPath: /secrets/ci-pull-credentials
           name: ci-pull-credentials
           readOnly: true
@@ -53,12 +47,6 @@ postsubmits:
           readOnly: true
       serviceAccountName: ci-operator
       volumes:
-      - name: boskos
-        secret:
-          items:
-          - key: credentials
-            path: credentials
-          secretName: boskos-credentials
       - name: ci-pull-credentials
         secret:
           secretName: ci-pull-credentials
--- a/ci-operator/jobs/openshift-service-mesh/ztunnel/openshift-service-mesh-ztunnel-release-1.27-presubmits.yaml
+++ b/ci-operator/jobs/openshift-service-mesh/ztunnel/openshift-service-mesh-ztunnel-release-1.27-presubmits.yaml
@@ -12,8 +12,6 @@ presubmits:
       skip_cloning: true
       timeout: 1h30m0s
     labels:
-      ci-operator.openshift.io/cloud: aws
-      ci-operator.openshift.io/cloud-cluster-profile: ossm-aws
       ci.openshift.io/generator: prowgen
       pj-rehearse.openshift.io/can-be-rehearsed: "true"
     name: pull-ci-openshift-service-mesh-ztunnel-release-1.27-cargo-build
@@ -23,7 +21,6 @@ presubmits:
       - args:
         - --gcs-upload-secret=/secrets/gcs/service-account.json
         - --image-import-pull-secret=/etc/pull-secret/.dockerconfigjson
-        - --lease-server-credentials-file=/etc/boskos/credentials
         - --report-credentials-file=/etc/report/credentials
         - --secret-dir=/secrets/ci-pull-credentials
         - --target=cargo-build
@@ -36,9 +33,6 @@ presubmits:
           requests:
             cpu: 10m
         volumeMounts:
-        - mountPath: /etc/boskos
-          name: boskos
-          readOnly: true
         - mountPath: /secrets/ci-pull-credentials
           name: ci-pull-credentials
           readOnly: true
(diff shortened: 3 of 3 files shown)


ci-operator/config/openshift-service-mesh/ztunnel/openshift-service-mesh-ztunnel-release-1.27.yaml
```diff
@@ -0,0 +1,153 @@
...
+      memory: 200Mi
+tests:
+- as: unit
+  steps:
+    cluster_profile: ossm-aws
+    env:
+      CI: "true"
+      COMPUTE_NODE_REPLICAS: "1"
+      COMPUTE_NODE_TYPE: m5d.2xlarge
+      MAISTRA_BUILDER_IMAGE: quay.io/maistra-dev/maistra-builder:3.2
```
jwendell: Is ztunnel's build hungry on resources like envoy/proxy? If not, I'd recommend using the sail or istio jobs as a template rather than proxy.

zmiklank: No, ztunnel is not that resource hungry.
I used proxy as a template, but lowered resource limits where applicable.
I have just compared with istio job, and it seems similar to the settings in this PR.

zmiklank: Hi @jwendell. Sorry it it took me so long. However I checked how istio handles ztunnel builds in CI, and it seems to me, that ztunnel builds are also pre-built and then just downloaded from gdrive.

ci-operator/config/openshift-service-mesh/ztunnel/openshift-service-mesh-ztunnel-release-1.27.yaml
```diff
@@ -0,0 +1,150 @@
...
+      from: src
+      resources:
+        requests:
+          cpu: 100m
+          memory: 100Mi
+      timeout: 1h0m0s
+    workflow: servicemesh-ztunnel-e2e-aws
+  timeout: 1h30m0s
+- as: copy-artifacts-gcs
+  postsubmit: true
```
yxun, line 39: You may temporary remove this `postsubmit true` line and test the job run from this PR pj-rehearse comment. 
When that works, you can commit this line back.

If the copy and upload script is simple, we may not need a workflow to provision a remote cluster.

ci-operator/config/openshift-service-mesh/ztunnel/openshift-service-mesh-ztunnel-release-1.27.yaml
```diff
@@ -0,0 +1,150 @@
...
+      env:
+      - name: MAISTRA_NAMESPACE
+      - name: MAISTRA_SC_POD
+      from: src
+      resources:
+        requests:
+          cpu: 100m
+          memory: 100Mi
+      timeout: 20m0s
+    - as: unit
```
yxun: I checked the build log of this job. Is it just doing a cargo build and compiling all source code ?
I didn't see any test from the log. Should we rename this job step as "cargo-build" ?

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_release/71880/rehearse-71880-pull-ci-openshift-service-mesh-ztunnel-release-1.27-unit/2013243019339239424/artifacts/unit/unit/build-log.txt

ci-operator/config/openshift-service-mesh/ztunnel/openshift-service-mesh-ztunnel-release-1.27.yaml
```diff
@@ -0,0 +1,150 @@
...
+      - name: MAISTRA_NAMESPACE
+      - name: MAISTRA_SC_POD
+      - name: CI
+      from: src
+      resources:
+        requests:
+          cpu: 100m
+          memory: 100Mi
+      timeout: 1h0m0s
+    workflow: servicemesh-ztunnel-e2e-aws
```
yxun: The reason of using a workflow is for running scripts which requires additional privilege in a remote pod.
It takes time for waiting a remote cluster pod ready.

If the cargo scripts can be run as a regular user, we may not need a workflow in that case.
For example , the gencheck part of istio job is not using workflow. The integration tests need a workflow for running KinD:
https://github.com/openshift/release/blob/master/ci-operator/config/openshift-service-mesh/istio/openshift-service-mesh-istio-release-1.24.yaml#L48

zmiklank: Hi @yxun Thank you very much for your feedback. As you can see from the state of the PR I am still quite confused from how these prow jobs work :).
 
 > If the cargo scripts can be run as a regular user, we may not need a workflow in that case.

 ztunnel build does not need any additional privileges and I was thinking I will do build only (no tests).
 But when you are mentioning it... ztunnel tests needs to run in privileged rootful container (the user running the container must be root). Is something like this possible in prow environment? 
 If no, it is not a problem, as we have ztunnel tests running downstream.

yxun: >  Is something like this possible in prow environment?

yes, your current PR is using a workflow. That workflow is creating a remote cluster and then copying source code to that remote pod. 
Because we have full privilege access to that remote cluster pod, all following steps can run in that remote rootful pod.

If we don't use a workflow, a job will be much faster by picking a prow environment default cluster pod which does not grant rootful privilege.

I can check your scripts details later and then provide some suggestion.

What I suggested earlier was trying to run your build scripts without using a workflow. The following `gencheck` job example is much clear and simpler than using a workflow one:
https://github.com/openshift/release/blob/master/ci-operator/config/openshift-service-mesh/istio/openshift-service-mesh-istio-release-1.24.yaml#L37

zmiklank: > If we don't use a workflow, a job will be much faster by picking a prow environment default cluster pod which does not grant rootful privilege.

then I think it is more useful to do build only (no tests) and do it without workflow. We can still add the tests later if really needed. I will rework the PR. Thanks.

zmiklank: that would mean I can leave out the whole `- as: copy-src` sections?

yxun: Yes, I think so.

https://github.com/openshift/release/blob/master/ci-operator/config/openshift-service-mesh/istio/openshift-service-mesh-istio-release-1.24.yaml#L47

line 47 "container.from: src" will git clone source code to a default cluster pod. The github repo is referring to the line 163 metadata below:
https://github.com/openshift/release/blob/master/ci-operator/config/openshift-service-mesh/istio/openshift-service-mesh-istio-release-1.24.yaml#L163

ci-operator/config/openshift-service-mesh/ztunnel/openshift-service-mesh-ztunnel-release-1.27.yaml
zmiklank: Renamed.

ci-operator/config/openshift-service-mesh/ztunnel/openshift-service-mesh-ztunnel-release-1.27.yaml
```diff
@@ -0,0 +1,78 @@
...
+      memory: 200Mi
+tests:
+- as: unit
+  steps:
+    cluster_profile: ossm-aws
+    test:
+    - as: cargo-build
+      cli: latest
+      commands: |
+        ./ossm/ci/pre-submit.sh"
```
jwendell: extra `"`

zmiklank:
/pj-rehearse pull-ci-openshift-service-mesh-ztunnel-release-1.27-cargo-build-and-push

ci-operator/config/openshift-service-mesh/ztunnel/openshift-service-mesh-ztunnel-release-1.27.yaml
```diff
@@ -0,0 +1,84 @@
...
+        name: gcs-credentials
+        namespace: test-credentials
+      from: src
+      resources:
+        requests:
+          cpu: 100m
+          memory: 100Mi
+      timeout: 1h0m0s
+  timeout: 2h0m0s
+- as: update-istio
```
yxun, line 58: The `cargo-build-and-push` and `update-istio` steps here are all post-submit jobs. 

I think the `update-istio` will also run when the `cargo-build-and-push` job failed. Will that cause an issue ? cc @jwendell

zmiklank: I think it is done the same in all our other components. At least in proxy it is like that for sure

PR #71892, opened 2025-11-26
USHIFT-6348: cache job will fail when brew rpm download fail

agullon:


ci-operator/step-registry/openshift/microshift/infra/iso-build/openshift-microshift-infra-iso-build-commands.sh
```diff
@@ -84,7 +84,7 @@ if [[ "${JOB_NAME}" =~ .*-cache.* ]] ; then
         bash -x ./test/bin/manage_brew_rpms.sh download "4.${y_version}" "${out_path}" "${LATEST_RELEASE_TYPE}" || echo "WARNING: Failed to download ${LATEST_RELEASE_TYPE} RPMs for 4.${y_version}"
         bash -x ./test/bin/manage_brew_rpms.sh download "4.${y_version}" "${out_path}" "nightly" || echo "WARNING: Failed to download nightly RPMs for 4.${y_version}"
         bash -x ./test/bin/manage_brew_rpms.sh download "4.$((${y_version} - 1))" "${out_path}" "zstream" || echo "WARNING: Failed to download zstream RPMs for 4.$((${y_version} - 1))"
-        bash -x ./test/bin/manage_brew_rpms.sh download "4.$((${y_version} - 2))" "${out_path}" "zstream"
+        bash -x ./test/bin/manage_brew_rpms.sh download "4.$((${y_version} - 2))" "${out_path}" "zstream" || ( echo "WARNING: Failed to download zstream RPMs for 4.$((${y_version} - 2))" && brew_rpm_download_failed=true )
     fi
     popd &>/dev/null
```
ggiguash: Can we refactor this code into a function?
This way was can return non-zero code when something breaks without the complex flag-based logic.

agullon: yes I agree, but where should I put the new function? in the same file? in `openshift-microshift-includes-commands.sh`? in microshift repo?
I always wondered why this was not a function in the first place

agullon: done, exported to a function in the same script, please review

ci-operator/step-registry/openshift/microshift/infra/iso-build/openshift-microshift-infra-iso-build-commands.sh
```diff
@@ -42,51 +42,52 @@ chmod +x /tmp/iso.sh
 #
 ci_clone_src
 
-# Attempt downloading latest MicroShift RPMs from brew.
-# This requires VPN access, which is only enabled for the cache jobs.
-skip_brew_rpm_download=false
-if [[ "${JOB_NAME}" =~ .*-cache.* ]] ; then
+download_brew_rpms() {
```
ggiguash, line 45: The return values in this function seem to be mixed
- 0 should be returned on success
- 1 should be returned on failure

Then it should allow code like `if ! download_brew_rpms ; then echo ERROR ; fi`

Am I missing something?

ci-operator/step-registry/openshift/microshift/infra/iso-build/openshift-microshift-infra-iso-build-commands.sh
```diff
@@ -42,51 +42,52 @@ chmod +x /tmp/iso.sh
...
+
+    # Download the latest RPMs from brew: latest release (ec, rc or zstream), nightly, Y-1 zstream and Y-2 zstream
+    y_version="$(cut -d'.' -f2 "${src_path}/Makefile.version.$(uname -m).var")"
+    bash -x ./test/bin/manage_brew_rpms.sh download "4.${y_version}" "${out_path}" "${LATEST_RELEASE_TYPE}" || echo "WARNING: Failed to download ${LATEST_RELEASE_TYPE} RPMs for 4.${y_version}"
+    bash -x ./test/bin/manage_brew_rpms.sh download "4.${y_version}" "${out_path}" "nightly" || echo "WARNING: Failed to download nightly RPMs for 4.${y_version}"
+    bash -x ./test/bin/manage_brew_rpms.sh download "4.$((${y_version} - 1))" "${out_path}" "zstream" || echo "WARNING: Failed to download zstream RPMs for 4.$((${y_version} - 1))"
+    bash -x ./test/bin/manage_brew_rpms.sh download "4.$((${y_version} - 2))" "${out_path}" "zstream" || ( echo "WARNING: Failed to download zstream RPMs for 4.$((${y_version} - 2))" && return 1 )
+
     popd &>/dev/null
+}
```
ggiguash, line 83: Let's add an explicit `return 0` here.

ci-operator/step-registry/openshift/microshift/infra/iso-build/openshift-microshift-infra-iso-build-commands.sh
ggiguash: Ah, I see. We do not consider line 52 and 57 an error.
Why?

agullon: because on old branches (release-4.19 and before) this script does not exist, so we don't want to fail, just skip it

ci-operator/step-registry/openshift/microshift/infra/iso-build/openshift-microshift-infra-iso-build-commands.sh
agullon: done

ci-operator/step-registry/openshift/microshift/infra/iso-build/openshift-microshift-infra-iso-build-commands.sh
agullon: added comments to explain this: d96094d4f61898ae08ee804804e01dcb559c3ef2

PR #71901, opened 2025-11-26
Fix the ilo and pdu reset

sgoveas:
Fix the ilo reset for ipxe with vmedia hosts. These arm64 hosts have
intermittent kernel crash failures in IPI installs, and sometimes host
hangs during booting up.

PDU reboot has not been working for a long time, this may have
contributed to the kernel crash issues. The PDU ssh server uses old
algorithims causing the ssh step to fail siliently. It works from
bastion host as it has some params to use old algorithms already setup.

Depends on 
PR for a [script to be added in bastion host](https://github.com/openshift-eng/baremetal-qe-infra/pull/94) and 
PR to [Update disk cleanup](https://github.com/openshift-eng/baremetal-qe-infra/pull/93)


sgoveas:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.21-multi-nightly-baremetal-ipi-ovn-ipv4-arm-f7

PR #71932, opened 2025-11-27
Remove oadp-vm-file-restore repo from openshift-priv configs

psalajova:
to fix https://github.com/openshift/release/pull/71787


mpryc:
@psalajova without this entry the images for `quay.io/konveyor/oadp-vm-file-restore` will not get synced form prow build job to quay?

PR #71938, opened 2025-11-27
Add presubmits job for hyperfleet-broker

yingzhanredhat:
Add the presubmit job for hyperfleet-api repo
-make lint
-make test-integration

Other:
-make test-unit //Will add it until https://issues.redhat.com/browse/HYPERFLEET-274 fixed

### DIFF
--- a/ci-operator/config/openshift-hyperfleet/hyperfleet-broker/openshift-hyperfleet-hyperfleet-broker-main__presubmits.yaml
+++ b/ci-operator/config/openshift-hyperfleet/hyperfleet-broker/openshift-hyperfleet-hyperfleet-broker-main__presubmits.yaml
@@ -1,0 +1,54 @@
+base_images:
+  jasee_openshift-build_rhel-9-golang-1.25:
+    name: openshift-build
+    namespace: jasee
+    tag: rhel-9-golang-1.25
+  nested-podman:
+    name: nested-podman
+    namespace: ci
+    tag: latest
+build_root:
+  image_stream_tag:
+    name: openshift-build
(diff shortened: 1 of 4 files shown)


ci-operator/config/openshift-hyperfleet/hyperfleet-broker/openshift-hyperfleet-hyperfleet-broker-main__presubmits.yaml
```diff
@@ -0,0 +1,54 @@
+base_images:
```
yasun1, line 1: I have some concerns:
1. Unit test is mock test, I believe it should take very few resources. So it could be in the `-main.yaml` directly.
2. In my personal preferences, presubmit should be used for some jobs needs to highlight out. For integration test, it is always `presubmit`, I do not think we need the extra world again.

yingzhanredhat: Some unit still requires Test container and the base image requires nested-podman. 
@xueli181114 Any advice here?

xueli181114: +1 to Yanming's points.

yingzhanredhat: Updated it

PR #71961, opened 2025-11-28
use gcp service account prefix instead

jianli-wei:
- update step `ipi-conf-gcp-service-account` to support GCP service account email prefix
- update related tests to give email prefix only, so that the GCP project ID is not hard-coded


ci-operator/step-registry/ipi/conf/gcp/service-account/ipi-conf-gcp-service-account-commands.sh
```diff
@@ -4,34 +4,50 @@ set -o nounset
 set -o errexit
 set -o pipefail
 
+GOOGLE_PROJECT_ID="$(< ${CLUSTER_PROFILE_DIR}/openshift_gcp_project)"
```
gpei, line 11: can we add a check if `GOOGLE_PROJECT_ID` is null, in case it's not actually well set.

PR #72011, opened 2025-12-01
no-jira: Update the installer golint to v2.3.1

barbacbd:
The current version (1.64.8) is insufficient on all platforms to find lint errors. A PR in the openshift installer (https://github.com/openshift/installer/pull/10121) will update the golangci-lint yaml file to version v2.

### DIFF
--- a/ci-operator/config/openshift/installer/openshift-installer-main.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-main.yaml
@@ -30,7 +30,7 @@ base_images:
   golangci-lint:
     name: golangci-lint
     namespace: ci
-    tag: latest
+    tag: v2.3.1
   nested-environment-builder:
     name: nested-environment-builder
     namespace: ci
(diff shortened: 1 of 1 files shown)


ci-operator/config/openshift/installer/openshift-installer-main.yaml
```diff
@@ -30,7 +30,7 @@ base_images:
   golangci-lint:
     name: golangci-lint
     namespace: ci
-    tag: v1.64.8
+    tag: latest
```
tthvo: My only concern is that the `latest` might be updated to point to a _newer_ golangci-lint version. For example, it was updated to point to a v2 version, which is breaking. Maybe, even a minor version can be breaking.

WDYT of creating a `v2.3.1` tag similar to https://github.com/openshift/release/pull/61018? I think everyone else might benefit from that too.

```yaml
# In file: core-services/image-mirroring/_config.yaml
ci/golangci-lint:v2.3.1:
  image: quay.io/openshiftci/golangci-lint:v2.3.1
ci/golangci-lint:latest:
  image: quay.io/openshiftci/golangci-lint:v2.3.1
```

tthvo (commented):
/approve

PR #72031, opened 2025-12-02
(ocm) chore: remove Prow configurations (GRC)

dhaiducek:
Remove all Prow configurations aside from 2.16 in favor of Konflux and GitHub Actions

ref: https://issues.redhat.com/browse/ACM-25785


dhaiducek:
Okay, I've got this down to just `publish` jobs (and have also restored all the framework tests as a result). Should be good to merge now.

PR #72041, opened 2025-12-02
Fix ORIGINAL_RELEASE_IMAGE_LATEST unbound variable in rosa-sts-account-roles-create

wangke19:
## Summary

Fixes the `periodic-ci-openshift-release-master-nightly-4.21-e2e-rosa-sts-ovn` job that has been failing since Nov 22, 2025 with:
```
/bin/bash: line 57: ORIGINAL_RELEASE_IMAGE_LATEST: unbound variable
```

## Root Cause

The `rosa-sts-account-roles-create` step script uses `set -o nounset` and attempts to access `ORIGINAL_RELEASE_IMAGE_LATEST` without checking if it's set. This variable should be automatically provided by ci-operator when `releases.latest` is configured, but it appears to not be set for variant configurations due to a ci-operator regression.

## Changes

Added a defensive check before using `ORIGINAL_RELEASE_IMAGE_LATEST`:
- Only attempts to parse the variable if it's set
- Falls back to existing ROSA CLI version query logic if the variable is missing
- Maintains backward compatibility

## Testing

The fix makes the script more robust by:
1. Preventing the unbound variable error
2. Gracefully falling back to the version query logic that already exists in the script
3. Not changing behavior when the variable IS properly set

/cc @openshift/openshift-team-ci

### DIFF
--- a/ci-operator/step-registry/rosa/sts/account-roles/create/rosa-sts-account-roles-create-commands.sh
+++ b/ci-operator/step-registry/rosa/sts/account-roles/create/rosa-sts-account-roles-create-commands.sh
@@ -52,7 +52,9 @@ VERSION_SWITCH=""
 if [[ "$CHANNEL_GROUP" != "stable" ]]; then
   # Get the X.Y from the release payload pullspec if we're using the one from CI
   if [[ "$OPENSHIFT_VERSION" == "release:latest" ]]; then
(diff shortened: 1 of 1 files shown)


wangke19:
/payload-job periodic-ci-openshift-release-master-nightly-4.21-e2e-rosa-sts-ovn

PR #72059, opened 2025-12-02
CMP-4002: Update Dockerfile path for security-profiles-operator to be correct

Vincent056:
Previously, security-profiles-operator was using the wrong Docker image; we need to fix the path to the correct one.


ci-operator/config/openshift/security-profiles-operator/openshift-security-profiles-operator-konflux.yaml
```diff
@@ -5,7 +5,7 @@ build_root:
     tag: rhel-9-golang-1.24-openshift-4.20
 canonical_go_repository: sigs.k8s.io/security-profiles-operator
 images:
-- dockerfile_path: Dockerfile.openshift
```
yuumasato, line 8: @Vincent056 I though this change was intentional in #70768 as you changed `Docker.openshift` in https://github.com/openshift/security-profiles-operator/pull/107/files#diff-f20a7cf8ba7e8b20f3bd1963e18191e6ba4474bb0bf92bdd745bb62841e4a6c3 too.

PR #72078, opened 2025-12-02
AGENT-1383: Add agent-iso-builder dependency

pawanpinjarkar:


ci-operator/step-registry/baremetalds/devscripts/setup/baremetalds-devscripts-setup-commands.sh
```diff
@@ -395,6 +395,12 @@ then
   echo "export APPLIANCE_IMAGE=${APPLIANCE_IMAGE}" >> /root/dev-scripts/config_root.sh
 fi
 
+# Add AGENT_ISO_BUILDER_IMAGE only for OVE ISOBuilder e2e tests
+if [ "${AGENT_E2E_TEST_BOOT_MODE}" == "ISO_NO_REGISTRY" ];
```
rwsu: Is the if clause needed? I don't think there is any consequence if the export is always added.

rwsu (changes requested):

ci-operator/step-registry/baremetalds/devscripts/setup/baremetalds-devscripts-setup-commands.sh
pawanpinjarkar: I think the if clause makes it more obvious to understand when is this env var is needed.

pawanpinjarkar: Similar to how appliance has been using above.

rwsu: I think the value for AGENT_ISO_BUILDER_IMAGE can always be written out. AGENT_E2E_TEST_BOOT_MODE will determine if it is used later in dev-scripts.

ci-operator/step-registry/baremetalds/devscripts/setup/baremetalds-devscripts-setup-commands.sh
```diff
@@ -395,6 +395,8 @@ then
   echo "export APPLIANCE_IMAGE=${APPLIANCE_IMAGE}" >> /root/dev-scripts/config_root.sh
 fi
 
+echo "export AGENT_ISO_BUILDER_IMAGE=${AGENT_ISO_BUILDER_IMAGE}" >> /root/dev-scripts/config_root.sh
```
andfasano: we don't want to set this config var always, isn't it? But only when testing the OVE ISO

pawanpinjarkar: This was discussed here https://github.com/openshift/release/pull/72078#discussion_r2590511138

andfasano: Since this step is shared by many different jobs (ie, not only agent jobs but also metal ones), it would be better to pollute other teams configuration with unwanted/not requested configuration (similarly to what already appliance does at line [393](https://github.com/openshift/release/pull/72078/files#diff-f891b30b93f62c85827a6dbf087d3deb2ce2f35e6c73cafeed7e4787d1b7b386R393)) - given that this is a specific agent config, and not a general one.

pawanpinjarkar: I agree 🙂 Adding it back

PR #71773, opened 2025-11-24
NO-ISSUE: Configure onprem test to run on assisted_amd64_el9

yoavsc0302:
For testing purposes, configure the e2e-metal-assisted-onprem-4-20 test to run on the assisted_amd64_el9 cluster type (AWS-based).

This is for testing https://issues.redhat.com/browse/MGMT-21418. Once merged, this will allow testing of
https://github.com/openshift/assisted-test-infra/pull/2751 in CI. After successful verification, this current PR will be reverted.

PR #72104, opened 2025-12-03
NO-ISSUE: Add assisted_amd64_el9 mapping to ofcir-acquire

yoavsc0302:
The assisted_amd64_el9 CLUSTERTYPE was added in #71773 but the ofcir-acquire script was not updated to map it to the corresponding CIRTYPE, causing tests to fail with 404 errors from OFCIR.

Add the mapping to allow tests using assisted_amd64_el9 to acquire the correct CI resource.


danmanor (approved):

yoavsc0302:
@tdomnesc  it didn't seem to have an affect. maybe /approve instead of /approved? or just /lgtm? or any other idea?

PR #72110, opened 2025-12-03
openperouter: Onboarding

zeeke:
Create configuration files to enable plugins and tide merging for the repository https://github.com/openshift-kni/openperouter

cc @oribon @fedepaol @maiqueb


maiqueb (approved):
I don't have much experience with reviewing openshift CI related changes. But I think this is OK.

PR #72122, opened 2025-12-03
CNTRLPLANE-2123:  hypershift: fix version reference from 4.19 to 4.18

jparrill:
Fix version reference in hypershift release-4.19 configuration.

Updates the version from "4.19" to "4.18" in the initial-418 release section to maintain proper version alignment.

### DIFF
--- a/ci-operator/config/openshift/hypershift/openshift-hypershift-release-4.19.yaml
+++ b/ci-operator/config/openshift/hypershift/openshift-hypershift-release-4.19.yaml
@@ -63,7 +63,7 @@ releases:
       product: ocp
       relative: 1
       stream: ci
-      version: "4.19"
+      version: "4.18"
   latest:
     integration:
       include_built_images: true


jparrill:
/hold

Until branching happens

ci-operator/config/openshift-priv/hypershift/openshift-priv-hypershift-release-4.21.yaml
```diff
@@ -180,9 +180,9 @@ tests:
     dependencies:
       OCP_IMAGE_LATEST: release:latest-420
```
bryan-cox, line 181: Why is this not 421, 420, 419, 418, 417?

ci-operator/config/openshift/hypershift/openshift-hypershift-main.yaml
```diff
@@ -116,26 +116,31 @@ releases:
...
+    candidate:
+      product: ocp
+      stream: ci
+      version: "4.21"
   n1minor:
     candidate:
       product: ocp
       stream: ci
-      version: "4.20"
+      version: "4.21"
```
bryan-cox: Shouldn't this be 4.20 and then each of the others are one version less right after this line?

ci-operator/config/openshift/hypershift/openshift-hypershift-main.yaml
```diff
@@ -178,17 +183,17 @@ tests:
...
-      OCP_IMAGE_LATEST: release:latest-420
-      OCP_IMAGE_N1: release:latest-419
-      OCP_IMAGE_N2: release:latest-418
-      OCP_IMAGE_N3: release:latest-417
-      OCP_IMAGE_N4: release:latest-416
-      OCP_IMAGE_PREVIOUS: release:initial-420
+      OCP_IMAGE_LATEST: release:latest-421
+      OCP_IMAGE_N1: release:latest-420
+      OCP_IMAGE_N2: release:latest-419
+      OCP_IMAGE_N3: release:latest-419
```
sjenning: should be 418 and then 417

ci-operator/config/openshift/hypershift/openshift-hypershift-release-4.19.yaml
```diff
@@ -227,7 +222,7 @@ tests:
       OCP_IMAGE_N1: release:latest-417
       OCP_IMAGE_N2: release:latest-416
       OCP_IMAGE_N3: release:latest-415
-      OCP_IMAGE_N4: release:latest-414
+      OCP_IMAGE_N4: release:latest-415
```
sjenning: why this change?

jparrill: 4.14 is not supported anymore after branching AFAIK.

ci-operator/config/openshift/hypershift/openshift-hypershift-main.yaml
jparrill: AKS is not supported in versions bellow 4.19, even with that, should I put that value there?

ci-operator/config/openshift-priv/hypershift/openshift-priv-hypershift-release-4.21.yaml
jparrill: This is for `- as: e2e-aks-4-20`, so 4.21 should not be there (?). Regarding others, AKS is not supported under 4.19, even with that, should I put those versions?

ci-operator/config/openshift/hypershift/openshift-hypershift-release-4.19.yaml
sjenning: This will not work because the NodePool test expects `SupportedVersionSkew` to be `False` on N-4 case, but it will be `True`

jparrill: ok, I will keep the 4.14 then

ci-operator/config/openshift/hypershift/openshift-hypershift-release-4.19.yaml
```diff
@@ -63,7 +63,7 @@ releases:
       product: ocp
       relative: 1
       stream: ci
-      version: "4.19"
+      version: "4.18"
```
sjenning, line 66: `ci-operator/config/openshift-priv` is automatically sync'ed from `ci-operator/config/openshift` on a regular basis.  This change is the only one that is really required.

PR #72124, opened 2025-12-03
OPRUN-4315: Add upgrade-e2e CI to OLMv1

tmshort:


tmshort:
I would expect the `pj-rehearse` to fail...

PR #72127, opened 2025-12-03
Use Firewatch base config file for CR related jobs

oharan2:
In new Firewatch Base Config file for CR jobs:

- Data Router Reporter step will be submitting issues in the LPTINEROP board
- Ignore all other steps by default, i.e. provisioning, nodes readiness, must gather etc.


rabin-io (approved):
/lgtm

PR #72140, opened 2025-12-04
Fix version

liangxia:
Point `master/release-4.21/release-4.22` to `rhel-9-golang-1.24-openshift-4.21`. We don't have openshift-4.22 yet, that's why release-4.22 also points to openshift-4.21. We will use follow up PR to fix that once we have openshift-4.22 image.


liangxia:
/pj-rehearse pull-ci-openshift-openshift-tests-private-release-4.21-images pull-ci-openshift-openshift-tests-private-release-4.21-images-images

PR #72151, opened 2025-12-04
Remove swift-operator kuttl test configuration

tosky:
This patch updates the swift-operator tests and removes the kuttl that are handled via zuul since
https://github.com/openstack-k8s-operators/swift-operator/pull/419 (main branch only, not 18.0-fr4).


fmount:
@stuggi @abays we also need to land this one when you get the chance to approve! (thanks a lot!)

stuggi (approved):
/lgtm

PR #72157, opened 2025-12-04
Add channels and allow X.Y.Z OCP versions

tbuskey:
Allow OCP versions like 4.20.6 to be installed
Allow channels: prerelease, fast, stable
defaults to fast which will have all versions.  stable channel only has versions that are recommended.
If you specify an X.Y version, you will get latest in that channel

Signed-off-by: Tom Buskey <tbuskey@redhat.com>

regenerate prowjobs & ran make


tbuskey:
Needed to merge @ldoktor 's ARO changes

PR #72175, opened 2025-12-04
OSASINFRA-3960: Stop using hack script for CAPO e2e tests in CCAPIO

stephenfin:
This aligns us with other platforms and lets us get rid of an [otherwise unnecessary bash script][1].

[1]: https://github.com/openshift/cluster-capi-operator/blob/cafd8d0e1b1e197449a5c402b5bcbd2429ee4cf6/hack/e2e-openstack.sh#L3-L5


mandre (commented):
/lgtm
/approve

PR #72179, opened 2025-12-04
OSASINFRA-3960: Rework e2e-techpreview job for CAPO

stephenfin:
Start calling the script directly to allow us to remove the otherwise useless Makefile target. Also rename the job to e2e-openstack-capi-techpreview to better capture its intent and align with CAPI providers.


stephenfin:
/hold

We need to merge https://github.com/openshift/cluster-api-provider-openstack/pull/385 first

PR #72181, opened 2025-12-04
NO-JIRA: bump expiration dates for microshift and rosa payload job skips

smg247:
microshift should be fixed prior to expiration with [USHIFT-6363](https://issues.redhat.com//browse/USHIFT-6363)

### DIFF
--- a/clusters/app.ci/prow/03_deployment/payload-testing-prow-plugin.yaml
+++ b/clusters/app.ci/prow/03_deployment/payload-testing-prow-plugin.yaml
@@ -136,11 +136,11 @@ items:
           - name: SKIP_JOB_REGEX_1
             value: microshift
           - name: SKIP_JOB_EXPIRE_1
-            value: "2025-11-27T00:00:00Z"
+            value: "2026-05-27T00:00:00Z"
           - name: SKIP_JOB_REGEX_2
             value: rosa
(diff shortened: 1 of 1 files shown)


clusters/app.ci/prow/03_deployment/payload-testing-prow-plugin.yaml
```diff
@@ -136,11 +136,11 @@ items:
           - name: SKIP_JOB_REGEX_1
             value: microshift
           - name: SKIP_JOB_EXPIRE_1
-            value: "2025-11-27T00:00:00Z"
+            value: "2026-05-27T00:00:00Z"
```
petr-muller, line 139:
```suggestion
            value: "2026-02-27T00:00:00Z"
```

maybe shorten microshift because [USHIFT-6363](https://issues.redhat.com/browse/USHIFT-6363) claims to get this working _fairly_ fast? lgtm+hold for you to decide

/hold
/lgtm

smg247: I thought about doing that, but I don't think it does much good. I think people are interested in this now, and `microshift` skipping will be disabled (hopefully) before either of those dates anyways. Now that we have a jira to track, we aren't likely to need to be reminded of these failures if it is not complete by 2/27.

/shrug

PR #72049, opened 2025-12-02
Bring back Cilium/Calico jobs on 4.19

mgencur:
This is a follow up to https://github.com/openshift/release/pull/69591 . The current PR will need a rebase once the other one is merged.

[CNTRLPLANE-1463](https://issues.redhat.com/browse/CNTRLPLANE-1463)

PR #72229, opened 2025-12-05
Use lvm-catalogsource as default for baremetal Agent installations

mgencur:
This is a follow-up to https://github.com/openshift/release/pull/72198 . It also changes the default for Agent workflows.

I was hitting this problem in my other https://github.com/openshift/release/pull/72049


mgencur:
@LiangquanLi930 Thank you so much for the quick turnaround!

PR #72282, opened 2025-12-08
Update require_self_approval from false to true for openshift-hyperfleet repos

yingzhanredhat:
1.update require_self_approval from false to true
2.Add prow config for hyperfleet-api-spec repo


yingzhanredhat:
@liangxia Would you help approve it? Thanks

PR #72296, opened 2025-12-08
installer: Skip OpenStack jobs when only docs are changed

eshulman2:
Add skip_if_only_changed pattern to OpenStack e2e jobs to prevent unnecessary job execution when only documentation files are modified. This affects e2e-openstack-ovn, e2e-openstack-proxy, e2e-openstack-nfv-intel, and openstack-manifests jobs across all active release branches (4.12-4.22 and main).

Also remove docs/user/openstack/install_upi.md from run_if_changed patterns in e2e-openstack-dualstack-upi jobs to avoid conflicts with the skip_if_only_changed filter.

assited-by claude

### DIFF
--- a/ci-operator/config/openshift/installer/openshift-installer-main.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-main.yaml
@@ -848,7 +848,7 @@ tests:
     cluster_profile: equinix-ocp-metal
     workflow: baremetalds-e2e-virtualmedia
 - as: e2e-openstack-ovn
-  run_if_changed: openstack|(^data/.*/rhcos\.json)
+  run_if_changed: (^(pkg|data|images|upi|cluster-api|vendor)/.*openstack)|(^data/.*/rhcos\.json)
   steps:
     cluster_profile: openstack-vexxhost
     env:
@@ -880,14 +880,14 @@ tests:
 - always_run: false
   as: e2e-openstack-proxy
   optional: true
-  run_if_changed: openstack
+  run_if_changed: ^(pkg|data|images|upi|cluster-api|vendor)/.*openstack
   steps:
     cluster_profile: openstack-vh-mecha-central
     workflow: openshift-e2e-openstack-proxy
 - always_run: false
   as: e2e-openstack-nfv-intel
   optional: true
-  run_if_changed: openstack
+  run_if_changed: ^(pkg|data|images|upi|cluster-api|vendor)/.*openstack
   steps:
     cluster_profile: openstack-nfv
     workflow: openshift-e2e-openstack-nfv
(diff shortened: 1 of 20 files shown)


ci-operator/config/openshift/installer/openshift-installer-main.yaml
```diff
@@ -858,7 +858,7 @@ tests:
 - always_run: false
   as: e2e-openstack-dualstack-upi
   optional: true
-  run_if_changed: (^images/openstack/)|(^docs/user/openstack/install_upi\.md$)|(^upi/openstack/)
+  run_if_changed: (^images/openstack/)|(^upi/openstack/)
```
mandre: We want to keep exercising the CI for changes to the `docs/user/openstack/install_upi.md` file because it has markers that are interpreted during the CI run (grep for `e2e-openstack-upi` in [the file](https://github.com/openshift/installer/blob/main/docs/user/openstack/install_upi.md?plain=1)).

It's being used [here](https://github.com/openshift/release/blob/f9ede5b5906641af383275342da45689583acba7/ci-operator/step-registry/upi/install/openstack/upi-install-openstack-commands.sh#L117-L120).

ci-operator/config/openshift/installer/openshift-installer-main.yaml
```diff
@@ -848,7 +848,7 @@ tests:
     cluster_profile: equinix-ocp-metal
     workflow: baremetalds-e2e-virtualmedia
 - as: e2e-openstack-ovn
-  run_if_changed: openstack|(^data/.*/rhcos\.json)
+  run_if_changed: (^(pkg|data|images|upi|cluster-api|vendor)/.*openstack)|(^data/.*/rhcos\.json)
```
mandre, line 851: I understand the intent, but aren't we at risk of NOT running the CI on important changes if the structure of the installer changes?

ci-operator/config/openshift/installer/openshift-installer-main.yaml
eshulman2: fixed

ci-operator/config/openshift/installer/openshift-installer-main.yaml
eshulman2: I agree this is a risk we need to be aware of. IMO this is a risk we need to decide if we want to take. I would assume a major structure change would lead to re-working some of the CI in any case but I agree this is something we need to take into account when considering the pros and cons of this change.

mandre: I'm leaning towards rejecting the change because the gain is minimal, for the rare changes to openstack docs, and introduces a risk of missing CI feedback on important changes. I'll let @stephenfin weight in.

PR #72335, opened 2025-12-09
OTA-1766 Adding stable e2e config files for OCP 4.21

JianLi-RH:
Adding stable e2e config files for OCP 4.21

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.21__amd64-stable.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.21__amd64-stable.yaml
@@ -1,0 +1,679 @@
+base_images:
+  ansible:
+    name: "4.21"
+    namespace: ocp
+    tag: ansible
+  aws-efs-csi-operator-create-efs:
+    name: "4.21"
+    namespace: ocp
+    tag: aws-efs-csi-operator-create-efs
+  cli:
+    name: "4.21"
+    namespace: ocp
+    tag: cli
+  dev-scripts:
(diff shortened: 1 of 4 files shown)


ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.21__amd64-stable.yaml
```diff
@@ -0,0 +1,679 @@
...
+      E2E_RUN_TAGS: '@rosa and @critical'
+      OCM_LOGIN_ENV: production
+      OPENSHIFT_VERSION: "4.21"
+      REGION: us-east-1
+      TEST_FILTERS: ~ChkUpgrade&;~NonPreRelease&;~Serial&;~Disruptive&;~DisconnectedOnly&;~HyperShiftMGMT&;~MicroShiftOnly&;~NonHyperShiftHOST&;ROSA&
+      TEST_IMPORTANCE: Critical
+    test:
+    - chain: openshift-e2e-test-hypershift-qe
+    workflow: rosa-aws-sts-hcp-byo-kms-oidc-auditlog
+- as: aws-rosa-hcp-sector-byo-kms-oidc-auditlog-prod-critical-f7
```
jhou1, line 177: We don't maintain these `rosa-hcp` tests, do we? Why do we create them here?

JianLi-RH: It would be better to move these jobs out of this file, keep normal OpenShift jobs and rosa jobs in their own files. But not urgent.

PR #72337, opened 2025-12-09
add timeout to align with e2e

sunzhaohua2:
we defined [e2e timeout](https://github.com/openshift/cluster-api-actuator-pkg/blob/master/hack/ci-integration.sh#L8) as 180 minutes, this modifies the job's test timeout to 3 hours.


damdo (commented):
/approve

PR #72345, opened 2025-12-09
steps.osc: Move cucushift-wait to post

ldoktor:
we moved the cucushift-wait job to test to ensure cleanup runs even on
job timeout, the problem is it doesn't run when e2e errors-out. Another
problem is clusterbot, which replaces the test step with it's own step,
which again ignores our timeout. Lets move it back to post and attempt
to signal to prow that this step might fail (to run the post-steps when
it fails)


ldoktor:
Changes:

* Unified the workflow for all providers to have the same experience on all clouds (aro/azure/aws)

PR #72417, opened 2025-12-10
Network edge: skip tests when no code is changed on releases from 4.20 to 4.16

davidesalerno:
This change is porting on  releases from 4.20 to 4.16 the changes introduced with #69753.

In particular we will avoid CI to execute e2e tests when no code is changed because this is in general not needed.

NOTE: if the job has got `always_run: false` and `optional: true ` it is not needed the `skip_if_only_changed` rule since the test is not executed automatically every time and it won't block the merge.


alebedev87 (approved):
/lgtm
/approve

PR #72472, opened 2025-12-11
Add cspi-qe to cluster profiles

oharan2:


ci-operator/step-registry/cluster-profiles/cluster-profiles-config.yaml
```diff
@@ -911,3 +911,40 @@
 - profile: aro-hcp-dev
   owners:
     - org: Azure
+- profile: aws-cspi-qe
```
etirta, line 914: Can you do more specific to repo level please. Thx.

PR #72483, opened 2025-12-11
Fix cyborg CI test command

thiagoalessio:
The PR <https://github.com/openshift-eng/cyborg/pull/230> changed the location of files in order to make cyborg installable by PIP.


hoxhaeris:
/hold
We need to coordinate the merges for both PRs.

PR #72493, opened 2025-12-11
kueue: remove yq and use jq from the package repo

rphillips:


kannon92:
Looking at https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/72493/rehearse-72493-pull-ci-openshift-kueue-operator-main-test-e2e-downstream-4-20/1999182716603469824,

it seems that this is working. 

It got past the build steps so I think this is fine to merge.

/lgtm

PR #72498, opened 2025-12-11
chore: check for [skip-e2e] commit comments in the PR title of RHDH helm e2e tests (RHDHBUGS-2431)

nickboldt:
chore: check for [skip-build] commit comments in the PR title of RHDH helm e2e tests ([RHDHBUGS-2431](https://issues.redhat.com//browse/RHDHBUGS-2431))

Signed-off-by: Nick Boldt <nboldt@redhat.com>

Assisted-by: Cursor

### DIFF
--- a/ci-operator/step-registry/redhat-developer/rhdh/ocp/helm/redhat-developer-rhdh-ocp-helm-commands.sh
+++ b/ci-operator/step-registry/redhat-developer/rhdh/ocp/helm/redhat-developer-rhdh-ocp-helm-commands.sh
@@ -1,10 +1,10 @@
 #!/bin/bash
 
-# Check for [skip-build] commit comments in the PR title
+# Check for [skip-e2e] commit comments in the PR title
 if [ "$JOB_TYPE" == "presubmit" ] && [[ "$JOB_NAME" != rehearse-* ]]; then
     PR_TITLE=$(curl -s "https://api.github.com/repos/${GITHUB_ORG_NAME}/${GITHUB_REPOSITORY_NAME}/pulls/${GIT_PR_NUMBER}" | jq -r '.title')
-    # only skip e2e tests for a [skip-build] PR that has been auto-approved by github-actions[bot]
-    if [[ "$PR_TITLE" == *"[skip-build]"* ]]; then
+    # only skip e2e tests for a [skip-e2e] PR that has been auto-approved by github-actions[bot]
+    if [[ "$PR_TITLE" == *"[skip-e2e]"* ]]; then
         echo "PR_TITLE: $PR_TITLE"
         APPROVALS=$(curl -s "https://api.github.com/repos/${GITHUB_ORG_NAME}/${GITHUB_REPOSITORY_NAME}/pulls/${GIT_PR_NUMBER}/reviews") # json array of approvals
         # iterate through the approvals and check if the approval is from the .user.login = "github-actions[bot]" and if the .state = "APPROVED" 


ci-operator/step-registry/redhat-developer/rhdh/ocp/helm/redhat-developer-rhdh-ocp-helm-commands.sh
```diff
@@ -1,4 +1,15 @@
 #!/bin/bash
+
+# Check for [skip-build] commit comments in the PR title
+if [ "$JOB_TYPE" == "presubmit" ] && [[ "$JOB_NAME" != rehearse-* ]]; then
```
subhashkhileri, line 4: I am trying to understand, if we want to skip the build then we also want the e2e to not run? why not use downstream image in this case?

nickboldt: because if the change didn't require a build then there's nothing to test. For example if I update some doc, I don't have to rebuild the image and re-run all the tests just to fix a typo in some markdown file. 

Or if we bump versions like in https://github.com/redhat-developer/rhdh/pull/3819/files there's zero need to spend hours and resources re-testing a no-op change.

nickboldt: if preferred I can rework this to use an auto-approve comment like https://github.com/redhat-developer/rhdh/pull/3819#pullrequestreview-3553130148 as the gate for "this PR doesn't need to run the e2e prow tests"

ci-operator/step-registry/redhat-developer/rhdh/ocp/helm/redhat-developer-rhdh-ocp-helm-commands.sh
```diff
@@ -1,4 +1,25 @@
...
+        # iterate through the approvals and check if the approval is from the .user.login = "github-actions[bot]" and if the .state = "APPROVED" 
+        approval_count=$(echo "$APPROVALS" | jq 'length')
+        for ((i=0; i<approval_count; i++)); do
+            user_login=$(echo "$APPROVALS" | jq -r ".[$i].user.login")
+            state=$(echo "$APPROVALS" | jq -r ".[$i].state")
+            if [[ "$user_login" == "github-actions[bot]" ]] && [[ "$state" == "APPROVED" ]]; then
+                echo "Auto-approved by github-actions[bot], so no need to run tests; skip all test execution with exit code 0"
+                exit 0
+            fi
+        done
```
subhashkhileri, line 19:
```suggestion
        done
        exit 1
```

nickboldt: I don't understand -- you wanted to ONLY skip the e2e testing if `"$PR_TITLE" == *"[skip-build]"*` and  `"$user_login" == "github-actions[bot]"` and `"$state" == "APPROVED"` ... but you want to fail the e2e testing here for all cases where the `[skip-build]` appears in the commit message?

That seems like the opposite of what you want here. I thought the point was to simply skip over this check if all three conditions are not met... and therefore DO run the tests.

subhashkhileri: The goal here is to fail fast and exit early when the PR title contains [skip-build], unless it satisfies the additional condition of being auto-approved by github-actions[bot].

This suggestion is based understanding that tests should not run when [skip-build] is present in the PR title. The checks for "$user_login" == "github-actions[bot]" and "$state" == "APPROVED" are included purely as a verification step, to ensure that the PR was auto-approved before marking it green.

It's just a suggestion. I was okay even if we don't want to fail fast. I had approved the PR.

ci-operator/step-registry/redhat-developer/rhdh/ocp/helm/redhat-developer-rhdh-ocp-helm-commands.sh
zdrapela: I don't think we want to use the `[skip-build]`, because it is now regularly used also in cases where you want the e2e tests to run while using the next image, but you don't need to wait for the build of the PR image (you are only skipping build). If we want to go this way, I'd at least choose a label that fits (like `[skip-e2e]`). But I consider having a label like this dangerous (it's giving everyone the right to skip e2e). You can always manually override the e2e-tests with `/override ci/prow/e2e-ocp-helm` if you have the rights to do that (this requires conscious manual approval, but it will work just fine for https://github.com/redhat-developer/rhdh/pull/3819). Or maybe a bot account from `auto-approve workflow` can use that?

But I still think that using the `skip_if_only_changed` is a safer option. It accepts golang-style regex, and you can configure it to target specific files in specific directories... See the docs: https://docs.prow.k8s.io/docs/jobs/#triggering-jobs. It can work perfectly well for PRs like this one, you mentioned initially: https://github.com/redhat-developer/rhdh/pull/3783, or at least use a combination.
Another benefit is that the e2e tests will be skipped instantly. You won't be waiting for a cluster claim, less wasted resources overall.

ci-operator/step-registry/redhat-developer/rhdh/ocp/helm/redhat-developer-rhdh-ocp-helm-commands.sh
zdrapela: I've written my opinion in the other comment, see: https://github.com/openshift/release/pull/72498/files#r2618732364

ci-operator/step-registry/redhat-developer/rhdh/ocp/helm/redhat-developer-rhdh-ocp-helm-commands.sh
nickboldt: OK so I'm hearing two things:
* use a new keyword `[skip-e2e]` for obvious no-op changes created by an approved bot (like bumping versions in package.json files after we release a .z update)
* have the auto-approver for these no-op changes created by the bot also use the `/override ci/prow/e2e-ocp-helm` command so that we don't bother to provision a cluster when we don't need one

> skip_if_only_changed

this only works for **file paths**. Are you suggesting I **add all the package.json files** into that regex so that any change to any package.json would be auto-skipped? that's not a good solution either because we only want to skip on version bumps, not actual changes. This `skip_if_only_changed` regex isn't smart enough to look at what the changes are, only what the filenames are. 

> It can work perfectly well for PRs like this one, you mentioned initially: https://github.com/redhat-developer/rhdh/pull/3783, or at least use a combination.

Again, this would mean any REAL changes to the dockerfiles would be autoskipped which is BAD. We only want version bumps to get the e2e-skip treatment... not ALL changes to dockerfiles.

ci-operator/step-registry/redhat-developer/rhdh/ocp/helm/redhat-developer-rhdh-ocp-helm-commands.sh
nickboldt: > if we don't want to fail fast.

I don't  want to fail fast. I want to pass fast / skip doing anything / return a zero code (pass, not fail).

nickboldt:
/pj-rehearse pull-ci-redhat-developer-rhdh-release-1.8-e2e-ocp-helm

ci-operator/step-registry/redhat-developer/rhdh/ocp/helm/redhat-developer-rhdh-ocp-helm-commands.sh
zdrapela: Thank you for taking the time to explain it. Now I see the full picture and see that `skip_if_only_changed` is not a solution for any changes that this targets. The solution in this PR is good.

PR #72512, opened 2025-12-12
fix setting about attach-default-network

zhouying7780:


zhouying7780:
@heliubj18 Please help review , thanks .

PR #72518, opened 2025-12-12
OpenStack: drop periodics for unmaintained releases

mandre:
OCP 4.15 dropped out of support on August 27 2025, and OCP 4.12, while
technically still covered by `Extended Update Support Add-On - Term 2`
until January 17, 2026 doesn't receive any updates.

Also reduce the frequency of 4.14 jobs.


stephenfin:
This looks good. However, should we keep the image references part of the jobs? I don't imagine any of these test images are making their way into the payload, but I'm cautious after https://github.com/openshift/release/pull/72233

stephenfin:
Thought that might be the case but happy to have asked all the same. Thanks :ok_hand: 

/lgtm
/approve

PR #72545, opened 2025-12-12
Remove wait for Azure cluster in peerpods-param-cm

tbuskey:
There was a wait for an Azure cluster that looked for a non-existant condition.  This caused all peer-pods and coco cluster build to timeout and fail


ci-operator/step-registry/sandboxed-containers-operator/peerpods/param-cm/sandboxed-containers-operator-peerpods-param-cm-commands.sh
```diff
@@ -162,7 +162,6 @@ handle_azure() {
     else
         # On normal cluster wait for it to be Available
         echo "Waiting for OpenShift infrastructure to be ready..."
```
wainersm: Hi @tbuskey ! Let's get rid of that `echo "Waiting ..."` too. Otherwise, lgtm.

PR #72547, opened 2025-12-12
feat: parse arch for updates

dhaiducek:
Currently running `make update` on an M1 Mac requires overriding `CONTAINER_ENGINE_OPTS`, so this adds logic to parse the arch.


dhaiducek:
Something in the images changed (maybe CGO was shut off?), but `make update` started working without changing the arch.

PR #72573, opened 2025-12-14
Upating to rename models as a service

jland-redhat:
We recently updated the `model-billing` repo to `models-as-a-service`, which seems to have broken some of our integrations


jland-redhat:
Covered by this:
https://github.com/openshift/release/pull/72318/files

PR #72592, opened 2025-12-15
OSASINFRA-3978: Increase openstack application credentials timeout

eshulman2:
Increase application credentials time to live to make sure they do not expire in cluster-control-plane-machine-set-operator which im some cases are longer than 6 hours.


mandre (commented):
I wouldn't be opposed to bump the global app creds timeout to 8 hours the same way @stephenfin did in https://github.com/openshift/release/pull/72129.

eshulman2:
@mandre sure, will bump the default to 8H

PR #72611, opened 2025-12-15
METAL-1651: Add periodic sync for IPE with rebasebot

elfosardo:


elfosardo:
/cc @chrischdi @damdo @elmiko @joelspeed @mdbooth @nrb @racheljpg @radekmanak

please have a look when you have a moment, thanks

PR #72618, opened 2025-12-15
ofcir: remove lab-small server type

tdomnesc:


tdomnesc:
I forgot intranet network access is required for bmctest. Closing this for now as I need to try and move the script to the CI pod instead of a server.

/close

PR #72637, opened 2025-12-16
Extend timeout to 24h for openshift-extended-test-disruptive

liangxia:
`openshift-extended-test-disruptive` need longer execution time.

### DIFF
--- a/ci-operator/step-registry/openshift-extended/test/disruptive/openshift-extended-test-disruptive-ref.yaml
+++ b/ci-operator/step-registry/openshift-extended/test/disruptive/openshift-extended-test-disruptive-ref.yaml
@@ -4,7 +4,7 @@ ref:
   grace_period: 10m
   commands: openshift-extended-test-disruptive-commands.sh
   cli: latest
-  timeout: 16h0m0s
+  timeout: 24h0m0s
   env:
   - name: TEST_SCENARIOS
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/openshift-extended/test/disruptive/openshift-extended-test-disruptive-ref.yaml
```diff
@@ -4,7 +4,7 @@ ref:
   grace_period: 10m
   commands: openshift-extended-test-disruptive-commands.sh
   cli: latest
-  timeout: 16h0m0s
+  timeout: 36h0m0s
```
jianlinliu: Is it a bit aggressive ? How about 24 or 32 hours?

liangxia: We don't know how long that job needs. So this is a initial value, a little bit large to make sure the job can finish (no interruption due to timeout again), then we can tune again base on the job execution time.

liangxia:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.21-multi-nightly-gcp-ipi-ovn-ipsec-amd-mixarch-f28-destructive

PR #72644, opened 2025-12-16
DPTP-4634: Allow gitops-cluster-* in ci-secret-bootstrap

liangxia:
Changes in this PR,

- Add `gitops-cluster-build05` and `gitops-cluster-core-ci` to `bw-allow-unused` list.
- Sort the `bw-allow-unused` list.

So that we will not get an unused error which fails [job](https://deck-internal-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/origin-ci-private/logs/periodic-ci-secret-bootstrap/2000375525905797120) `ci-secret-bootstrap`


danilo-gemoli:
Why not removing them entirely if they are not used?

PR #72663, opened 2025-12-16
OSSM: upload test results from Istio int tests to report portal

FilipB:


ci-operator/config/openshift-service-mesh/istio/openshift-service-mesh-istio-master.yaml
```diff
@@ -154,9 +154,13 @@ tests:
     cluster_profile: ossm-aws
     env:
       BASE_DOMAIN: servicemesh.devcluster.openshift.com
+      INSTALLATION_METHOD: helm
```
fjglira, line 157: Yeah, technically this is installed using Helm by the integration test, but I think it worth to mention in some way that this is using directly upstream framework. Idk something like: `INSTALLATION_METHOD: upstream-testing-framework`

ci-operator/config/openshift-service-mesh/istio/openshift-service-mesh-istio-master.yaml
```diff
@@ -440,9 +476,13 @@ tests:
     cluster_profile: ossm-aws
     env:
       BASE_DOMAIN: servicemesh.devcluster.openshift.com
+      INSTALLATION_METHOD: converter
```
fjglira: Yeah, we are using the converter to run this test but at the end the installation method is sail-operator and for downstream for this cases we should set ossm-operator. I think is more explicit and accurate because the istio resources are going to be installed using those operators in those cases

FilipB: INSTALLATION_METHOD is used for sail operator e2e where we can have values `helm` and `olm`, then for Istio int tests where we need to cover: 

- upstream framework which is using helm to install the control plane (no sail operator)
- converter which is installing sail operator via helm and control plane via operator
- olm in downstream (old job) + predefined Istio CRs
- olm in downstream + converter

I think for sail operator tests it's ok what we have. Not sure what to use for istio int tests for those 4 variants.

fjglira: Ok. We should in a doc the use to allow folks to understand  how it work, later when we apply some filters in the report portal dashboards we should be able to understand the differences between the test run easily. I'm think more in how to present the changes for the release progress

FilipB: Maybe we can ignore old downstream job and  use `helm` and `sail`. It will be clear with combination with other attributes. `midstream + helm`, `midstream + sail`, `downstream + sail`

fjglira: Yes, I prefer this the combination of stage + installation will be easy to understand and be clear about where and How

FilipB:
/pj-rehearse pull-ci-openshift-service-mesh-istio-master-istio-integration-helm

fjglira (approved):
Lgtm

PR #72710, opened 2025-12-17
rhdh set pre-submit job max concurrency and move cron 3h early

subhashkhileri:
https://issues.redhat.com/browse/RHDHBUGS-2440 

move all the cron jobs by 3 hours.

add max concurrency for pre-submit jobs.

### DIFF
--- a/ci-operator/jobs/redhat-developer/rhdh/redhat-developer-rhdh-main-presubmits.yaml
+++ b/ci-operator/jobs/redhat-developer/rhdh/redhat-developer-rhdh-main-presubmits.yaml
@@ -141,7 +141,7 @@ presubmits:
     labels:
       ci.openshift.io/generator: prowgen
       pj-rehearse.openshift.io/can-be-rehearsed: "true"
-    max_concurrency: 1
+    max_concurrency: 4
     name: pull-ci-redhat-developer-rhdh-main-e2e-aks-helm-nightly
     optional: true
     rerun_command: /test e2e-aks-helm-nightly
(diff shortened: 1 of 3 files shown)


ci-operator/jobs/redhat-developer/rhdh/redhat-developer-rhdh-main-presubmits.yaml
```diff
@@ -77,6 +77,7 @@ presubmits:
     labels:
       ci.openshift.io/generator: prowgen
       pj-rehearse.openshift.io/can-be-rehearsed: "true"
+    max_concurrency: 1
```
zdrapela, line 80: You can leave the AKS and EKS jobs without limiting the `max_concurrency`, as they use the ephemeral clusters where each cluster is created just for the job run (no pool). If you want to limit it somehow to avoid rate limits, you can set it similarly to `e2e-ocp-helm`

subhashkhileri: Yup, we can revert that if we’re confident about it, and if concurrent runs won’t cause any quota issues on the cloud providers.

subhashkhileri: Updated and set it to 4. I’ll leave cleanup jobs at max 1, since it doesn’t really make sense to run them in parallel.

rm3l (approved):
/lgtm

PR #72739, opened 2025-12-17
Adding documentation for performance job in Sail

fjglira:


ci-operator/config/openshift-service-mesh/sail-operator/README.md
```diff
@@ -0,0 +1,577 @@
+# OpenShift Service Mesh CI Documentation
```
FilipB, line 1: This document should be located one level higher in the dir structure.

ci-operator/config/openshift-service-mesh/sail-operator/README.md
```diff
@@ -0,0 +1,577 @@
...
+- **pilot** - Core traffic management and service discovery
+- **telemetry** - Metrics, tracing, and logging functionality
+- **security** - mTLS, authorization policies, certificates
+- **ambient** - Ambient mesh mode with ztunnel
+- **helm** - Installation and upgrade testing
+
+**Environment Variables:**
+```yaml
+INSTALLATION_METHOD: helm|sail    # Report Portal attribute
+CONTROL_PLANE_SOURCE: sail        # Report Portal attribute
```
FilipB: This is not used by report portal. I think we should not duplicate description of the ENV variables as they are already described in the step registry. We should probably just point there.

ci-operator/config/openshift-service-mesh/sail-operator/README.md
fjglira: Yeah, the gen check fails if I try to place this in the main openshift-service-mesh folder. I'm looking for another option

fjglira:
> The idea is very good but this AI generated doc will probably be hard to understand for new users. I'm not sure who is expected consumer of this doc. For maintainers it's too long and contains unnecessary information, for new users, I don't think they will understand the AI generated content. I'm not sure if we should just keep it as simple as possible and keep here really only things useful for maintainers, things which are not obvious.

Yeah, I think we can do that, only leave there in the files information that can be useful for maintainers. Let me simpoify everything

PR #72630, opened 2025-12-16
TRT-2452: Create missing 4.22 periodic variants

neisw:
Post 4.21 branching to migrate periodic variant configurations

Generated with Claude: `/migrate-variant-periodics --skip-existing 4.21 4.22`

This is an alternative to [DPTP-4435](https://issues.redhat.com/browse/DPTP-4435) while we wait to see what is possible there.

PR #72740, opened 2025-12-17
cao: remove unused built image from variant configs

petr-muller:
The `periodics` variant configs do not need to contain `images` stanza (unless something in the jobs themselves needs the built image). The periodics are configured to install/update to `releases.latest` unmodified, but the jobs will still build the image from whatever the corresponding branch HEAD at the trigger time. This takes time, a bit of compute and also causes a redundant -images job to exist. None of this is necessary.


petr-muller:
https://github.com/openshift/release/pull/72630 should go first (and then this needs to be updated for 4.22 stuff)

/hold

PR #72766, opened 2025-12-18
Add jianzhangbjz to OLM openshift-patch-managers

jianzhangbjz:
The OTE framework has largely stabilized, and an increasing number of QE test cases have been migrated to the downstream repositories. Given this progress, it would be helpful to add a QE lead to help handle automation backport issues.


jianzhangbjz:
Hi @kuiwang02 @Xia-Zhao-rh @bandrade , could you help approve it? Thanks!

PR #72788, opened 2025-12-18
Set default LVM_OPERATOR_SUB_CHANNEL to match current cluster version if not specified

gaol:
* Unset the values in the workflows to decide based on current cluster version

Now the `lvms-catalogsource` step installs the catalog source based on current cluster version, so I think it would be good to align the subscription side as well by default.

Users still can override in the job config level (`ci-operator/config/`), but it would be good to have the default value align with current cluster version.


gaol:
@mgencur  WDYT ?

gaol:
/pj-rehearse periodic-ci-openshift-hypershift-release-4.19-periodics-e2e-kubevirt-metal-conformance-calico

PR #72793, opened 2025-12-18
Add per environment resources for MSI reuse in aro-hcp e2e tests

roivaz:
Each `aro-hcp-test-msi-containers-<env>` holds resources that match pre-created identity containers (resource groups) in Azure. Each container provides the identities required to create an ARO HCP cluster. An e2e job will requests N leases to support all the tests that it needs to run.


raelga (approved):
/lgtm

PR #72801, opened 2025-12-18
Copy timing info over, needed for generating links

janboll:


ci-operator/step-registry/aro-hcp/test/local/aro-hcp-test-local-commands.sh
stevekuznetsov, line 122: This won't work. `$SHARED_DIR` is a configmap with no directories. The timing metadata itself already is in here - can we improve the generation of that artifact to also record other metadata we care about?

raelga:
/remove-lgtm

(as per @stevekuznetsov comment)

ci-operator/step-registry/aro-hcp/test/local/aro-hcp-test-local-commands.sh
janboll: Why is it called like that :sob: 
That just makes no sense.

I'll find another way then

janboll:
Just discovered, the data is already present in shared dir

PR #72805, opened 2025-12-18
adding support to pass cred that enables scale-from-zero test in hype…

jhjaggars:
…rshift


ci-operator/step-registry/hypershift/install/hypershift-install-commands.sh
```diff
@@ -55,6 +55,10 @@ if [ "${TEST_CPO_OVERRIDE}" == "1" ]; then
   EXTRA_ARGS="${EXTRA_ARGS} --enable-cpo-overrides"
 fi
 
+if [ "${ENABLE_SCALE_FROM_ZERO}" == "true" ]; then
+  EXTRA_ARGS="${EXTRA_ARGS} --scale-from-zero-provider aws --scale-from-zero-creds=/etc/hypershift-pool-aws-credentials/credentials"
+fi
+
```
LiangquanLi930: IMHO, there’s no need to make this an option; making it the default would be better.

PR #72807, opened 2025-12-18
cnto: make periodics test payloads to expose the results to sippy

petr-muller:
These periodics were migrated in https://github.com/openshift/release/pull/70918 from the main config but the migrated configs kept their "build images + ephemeral release to test" configuration, but the payload+component readiness jobs should actually test ci or nightly candidate payloads, not ephemeral one. I have fixed up the config, this also causes prowgen to correctly label the jobs with `job-release` labels which make the jobs visible to TRT machinery.


jmencak:
Thank you for the PR.
/approve
/lgtm

PR #72808, opened 2025-12-18
Update test execution timeout for release jobs

kasturinarra:
Update test_execution_timeout for release jobs to 60m and remove the same from periodics.


ci-operator/config/openshift/microshift/openshift-microshift-release-4.21.yaml
```diff
@@ -224,6 +224,7 @@ tests:
       EC2_INSTANCE_TYPE: c5.metal
       MICROSHIFT_OS: rhel-9.6
       SCENARIO_TYPE: releases
+      TEST_EXECUTION_TIMEOUT: 60m
```
ggiguash, line 227: We have the `release` test execution scheduled in periodics too. This means that we need to update `__periodic` files, or they will be perma-failing with timeouts.

PR #72819, opened 2025-12-18
change 4.21 to be a sync'd branch

jluhrsen:
now that 4.22 is master we no longer want to do a downstream merge from it (4.22 will handle that), but we do want to sync with whatever ends up in 4.22. replace the d/s merge job with the branch sync job


jluhrsen:
> /lgtm
> 
> did they automatically copy over the jobs and is that why we don't need to do anything for master and we got a free copy at 4.21 is it?

yep!

PR #72823, opened 2025-12-19
AGENT-1383: Add agent-iso-builder job for appliance

pawanpinjarkar:


ci-operator/config/openshift/appliance/openshift-appliance-master.yaml
```diff
@@ -66,6 +66,16 @@ tests:
     allow_skip_on_success: true
     cluster_profile: packet-assisted
     workflow: agent-e2e-appliance-ipv4
+- always_run: false
```
danielerez, line 69: No need to include in master until the iso builder supports it...

PR #72845, opened 2025-12-19
prevent regex backtracking timeout in dev-scripts log redaction

bfournie:
The sed pattern 's/.*auths.*/...' causes catastrophic backtracking when processing long lines in log files or streaming output. The greedy '.*' prefix forces the regex engine to backtrack character-by- character when matching fails, resulting in O(n²) performance that can add significant delays during the 175-minute dev-scripts installation.

This issue can manifest as timeouts during CI runs, particularly in the log redaction phase or during streaming output filtering from SSH sessions.

Changes:
- Replace 's/.*auths.*/.../' with 's/auths.*$/.../' in both locations:
  1. Static log file processing in the finished() function
  2. Streaming output filter during dev-scripts installation
- Remove '/g' flag as only one match per line is needed
- Add comments explaining the pattern choice

The new pattern matches from 'auths' to end-of-line without the leading '.*', eliminating backtracking and providing O(n) performance.

Help with timeout issues in baremetalds dev-scripts setup jobs.

### DIFF
--- a/ci-operator/step-registry/baremetalds/devscripts/setup/baremetalds-devscripts-setup-commands.sh
+++ b/ci-operator/step-registry/baremetalds/devscripts/setup/baremetalds-devscripts-setup-commands.sh
@@ -35,8 +35,9 @@ finished()
   echo "dev-scripts setup completed, fetching logs"
   ssh "${SSHOPTS[@]}" "root@${IP}" tar -czf - /root/dev-scripts/logs | tar -C "${ARTIFACT_DIR}" -xzf -
   echo "Removing REDACTED info from log..."
+  # Use '/auths/ s/.*/' instead of 's/.*auths.*/' to avoid regex backtracking on long lines
   sed -i '
-    s/.*auths.*/*** PULL_SECRET ***/g;
+    /auths/ s/.*/*** PULL_SECRET ***/;
     s/password: .*/password: REDACTED/;
     s/X-Auth-Token.*/X-Auth-Token REDACTED/;
     s/UserData:.*,/UserData: REDACTED,/;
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/baremetalds/devscripts/setup/baremetalds-devscripts-setup-commands.sh
```diff
@@ -35,8 +35,9 @@ finished()
   echo "dev-scripts setup completed, fetching logs"
   ssh "${SSHOPTS[@]}" "root@${IP}" tar -czf - /root/dev-scripts/logs | tar -C "${ARTIFACT_DIR}" -xzf -
   echo "Removing REDACTED info from log..."
+  # Use 's/auths.*$/' instead of 's/.*auths.*/' to avoid regex backtracking on long lines
   sed -i '
-    s/.*auths.*/*** PULL_SECRET ***/g;
+    s/auths.*$/*** PULL_SECRET ***/;
```
zaneb: Previously this redacted the whole line, but now only the bit starting with `auths`. I think an efficient equivalent to the previous command would be:

```
/auths/ s/.*/*** PULL_SECRET ***/
```

The whole thing is weird though, because a pull secret contains `auth` not `auths`.

bfournie: Good point. Let me try that.

Pull-secret format is:
```
{
  "auths": {
    "https://your.private.registry.com": {
      "username": "<your-username>",
      "password": "<your-password>",
      "email": "your.email@example.com",
      "auth": "<base64-encoded-auth-string>"
    },
    "quay.io": {
      "auth": "<base64-encoded-auth-string-for-quay>",
      "email": "your.email@example.com"
    }
  }
}
```

andfasano:
> /pj-rehearse pull-ci-openshift-agent-installer-utils-release-4.21-e2e-agent-compact-ipv4-iso-no-registry
> /pj-rehearse ack

In general the patch looks fine to me, but since the setup step is very general and shared by many different team jobs, I'd feel more confident in running a more extensive rehearsal to test also non-agent jobs (rather than covering just agent jobs - we share only some targets with the other jobs)

PR #72855, opened 2025-12-19
steps.osc: Add default TEST_FILTERS to workflow

ldoktor:
The default TEST_FILTERS include ~Serial, but our "Install operator" test is marked as serial now. This works well in the defined workflows as we defined the filters there, but it's failing in clusterbot unless one specifies the TEST_FILTERS manually, which is currently likely impossible as slack corrutps the & character. Let's move the TEST_FILTERS from workflow to step to have them our way by default while reflecting user-defined values.


wainersm:
I always thought about removing TEST_FILTERS from the jobs definition to somewhere in the steps. It's a boilerplate that we rarely want to customize.

PR #72876, opened 2025-12-22
CMP-4041: Set a default value true for the INSTALL_OPERATOR variable

xiaojiey:
The default value for INSTALL_OPERATOR variable is missing, which will leading to a parse error.  The line INSTALL_OPERATOR=${INSTALL_OPERATOR:-true} sets a default value of true for the INSTALL_OPERATOR variable if it's not already set in the environment. This prevents the error you were seeing:                                                                                                                                                                                                  
`  invalid boolean value "" for -install-operator: parse error `


xiaojiey:
/pj-rehearse periodic-ci-ComplianceAsCode-content-master-4.16-e2e-aws-openshift-platform-compliance-weekly

PR #72882, opened 2025-12-22
NTO: adding must gather for day0 installation

eifrach:


ci-operator/step-registry/telcov10n/functional/compute-nto/ocp-setup-sno-day0/telcov10n-functional-compute-nto-ocp-setup-sno-day0-workflow.yaml
```diff
@@ -6,6 +6,7 @@ workflow:
     - ref: telcov10n-functional-compute-nto-process-inventory
     - ref: telcov10n-functional-compute-nto-config
     - ref: telcov10n-functional-compute-nto-ocp-deploy-sno
+    - ref: telcov10n-functional-compute-nto-must-gather
```
shaior, line 9: Shouldn't the mustgather be at the post? just in case of a failure, if happens, just wondering...

eifrach: it need the must gather for the gotest -- so if there isn t one the test will fail
this is mandatory

PR #72926, opened 2025-12-23
Remove unused Kueue jobs

MaysaMacedo:
Kueue does not release with OCP, it follows it's own release process and the branches are release-1.2, release-1.1, main, and all of them should be supported on various OCP releases. Let's remove the jobs that do not exercise that, to avoid needing to maintain those files.


MaysaMacedo:
/cc @kannon92 @sohankunkerkar 
Let me know if it's ok from your perspective to remove these jobs

PR #72947, opened 2025-12-25
OCPQE-31327: Remove aks hcp jobs from low version upgrade profiles

xiuwang:
Those jobs used 4.19.0 and 4.18.z as original version to setup aks hcp, all vesions which don't include https://github.com/openshift/hypershift/pull/6376/files should have below issue.
hcp failed with error：
```
  - lastTransitionTime: "2025-11-24T08:49:06Z"
    message: 'failed to read certificate: read /mnt/certs/: is a directory'
    observedGeneration: 2
    reason: InsufficientClusterCapabilities
    status: "False"
    type: ValidHostedControlPlaneConfiguration
```


JianLi-RH:
LGTM

PR #72970, opened 2025-12-28
Update jobs for the NVIDIA GPU monitoring console plugin

empovit:
* Publish main branch image with tag `main` instead of `latest`.
* Skip image builing if only OWNERS, README or the Helm chart change.
* Add image build jobs for releases 0.2.3 through 0.2.5.


yakovbeder (commented):
/lgtm

PR #72999, opened 2025-12-30
OCPQE-31370:Fix azure lease auth mode

wewang58:


wewang58:
@heliubj18 @xiuwang  please review it,thanks

PR #73031, opened 2026-01-01
ARO-23201 | Migrate ci-observability GitHub action to Prow

oriAdler:
1. Add the required tools to the base image.
2. Add new jobs for Promethues & Grafana.

### DIFF
--- a/ci-operator/config/Azure/ARO-HCP/Azure-ARO-HCP-main.yaml
+++ b/ci-operator/config/Azure/ARO-HCP/Azure-ARO-HCP-main.yaml
@@ -89,6 +89,18 @@ images:
         mv prometheus-3.2.1.linux-amd64/promtool /usr/local/bin/promtool && \
         chmod +x /usr/local/bin/promtool && \
         rm -rf prometheus-3.2.1.linux-amd64
+    RUN rpm --import https://packages.microsoft.com/keys/microsoft.asc
+    RUN dnf install -y https://packages.microsoft.com/config/rhel/9.0/packages-microsoft-prod.rpm
+    RUN dnf install -y azure-cli libicu
+    RUN az bicep install
+    # Move bicep to a system-wide location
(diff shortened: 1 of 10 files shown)


oriAdler:
/pj-rehearse pull-ci-Azure-ARO-HCP-main-observability-prometheus

oriAdler:
@geoberle @stevekuznetsov PTAL

ci-operator/config/Azure/ARO-HCP/Azure-ARO-HCP-main.yaml
```diff
@@ -89,6 +89,18 @@ images:
...
+    RUN dnf install -y azure-cli libicu
+    RUN az bicep install
+    # Move bicep to a system-wide location
+    RUN mv /root/.azure/bin/bicep /usr/local/bin/bicep
+    RUN dnf install -y wget
+    RUN curl -sfLo /usr/local/bin/yq https://github.com/mikefarah/yq/releases/download/v4.44.1/yq_linux_amd64 && \
+        chmod +x /usr/local/bin/yq
+    # Install Python for grafana tests
+    RUN dnf install -y python3 python3-pip
+    RUN pip3 install black
```
deads2k, line 103: This path seems likely to lead to future conflicts amongst tools.  I'll allow it, but if steveK decides we need a different approach, I'll allow that too.

Similarly, this image required to run various targets is getting pretty long and also points to problems for future developers too as they work on getting their tool chains up.  Again, I'll allow it, but if we later decide to split the configuration from the development, be ready for it.

stevekuznetsov: We could also use LLM to migrate this tool from Python to Go (and keep this as a non-prow task until then)- it has been tech debt x100 that it's Python this entire time - we struggle to manage the Python dependencies that this actually gets run with in the rollout as well

PR #73088, opened 2026-01-05
NE-2380: Add gatewayapi conformance test on baremetalds

lihongan:


ci-operator/step-registry/ingress/baremetalds-metallb/OWNERS
```diff
@@ -0,0 +1,12 @@
+approvers:
+  - Miciah
```
rikatz: we are missing the other new team members: https://github.com/openshift/cluster-ingress-operator/blob/master/OWNERS#L27-L28

ci-operator/step-registry/ingress/baremetalds-metallb/conf/OWNERS
```diff
@@ -0,0 +1,12 @@
+approvers:
```
rikatz: AFAIK you don't need another OWNERS here, it will inherit from the parent directory

ci-operator/step-registry/ingress/baremetalds-metallb/conf/ingress-baremetalds-metallb-conf-commands.sh
```diff
@@ -0,0 +1,202 @@
...
+  if oc get packagemanifest metallb-operator -n openshift-marketplace &>/dev/null; then
+    PACKAGE_EXISTS=true
+    echo "Found metallb-operator package in catalog"
+    break
+  fi
+  echo "Retry ${retry}/${CATALOG_RETRIES}: metallb-operator not found, waiting for catalog sync..."
+  sleep 10
+done
+
+if [[ "${PACKAGE_EXISTS}" == "false" ]]; then
```
rikatz, line 24: interesting. Is there some way from metallb team (or a way for us) to install metallb even on pre-release instead of upstream? 

My concern here is that we start to pass (or fail) based on upstream, and may not reflect the final release with metallb.

ci-operator/step-registry/ingress/baremetalds-metallb/conf/ingress-baremetalds-metallb-conf-commands.sh
```diff
@@ -0,0 +1,202 @@
...
+cat <<EOF | oc apply -f -
+apiVersion: metallb.io/v1beta1
+kind: IPAddressPool
+metadata:
+  name: ip-pool
+  namespace: metallb-system
+spec:
+  autoAssign: true
+  addresses:
+  - 192.168.111.30-192.168.111.35
```
rikatz: just double checking, I am not sure how many parallel tests we run for Gateway API (I know conformance ones are parallel and may create at least 5 gateways), so should we have a bigger range here?

ci-operator/step-registry/ingress/baremetalds-metallb/conf/ingress-baremetalds-metallb-conf-commands.sh
lihongan: I think using upstream is more simple and stable than pre-release. It is not easy to configure pre-release since it needs catalog, image digest mirror set and especially pull secret, and job might be failing/flaky due to these configs.
Until the release to be GAed we cannot install metallb via the default catalogsource.

ci-operator/step-registry/ingress/baremetalds-metallb/conf/ingress-baremetalds-metallb-conf-commands.sh
lihongan: Oh, during my testing I just observed 4 loadbalancer services of gateway in the same time.
And no worries, it is safe and easy to expand the range. I could update the range from `30` to `39`,

ci-operator/step-registry/ingress/baremetalds-metallb/conf/OWNERS
lihongan: I think you mean use the symlink pattern, like 
```
ln -s ../OWNERS OWNERS
```
right?
I will update the parent `ingress/OWNERS` and symlinks in `ingress/baremetalds-metallb/`, other sub folders could be updated in follow up.

ci-operator/step-registry/ingress/baremetalds-metallb/conf/ingress-baremetalds-metallb-conf-ref.yaml
```diff
@@ -0,0 +1,14 @@
+ref:
+  as: ingress-baremetalds-metallb-conf
```
rikatz, line 2: maybe if we want to reuse this approach on vsphere, can this be just something like ingress-metallb-conf and we use on vsphere, vcf9 and onprem?

The only change needed would be to pass the range of the L2Advertisement on the script by some other step (like on vsphere you get it from the  "static" step, on equinix you hardcode that 192.168.xxxx)

rikatz:
Interesting, your test executed almost successfully but something deleted the CRDs during it?

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_release/73088/rehearse-73088-pull-ci-openshift-cluster-ingress-operator-release-4.21-e2e-baremetalds-metallb-gatewayapi-conformance/2009222953094352896/artifacts/e2e-baremetalds-metallb-gatewayapi-conformance/ingress-baremetalds-metallb-gatewayapi-conformance-test/build-log.txt

lihongan:
> Interesting, your test executed almost successfully but something deleted the CRDs during it?
> 
> 
@rikatz sorry, yesterday I just added `sleep 30m` to the scripts and rsh to the host then did some manual testing (e2e) to check something, I believe one e2e test deleted the CRD, please ignore this.
And I will remove `sleep 30m` and retest today.

ci-operator/step-registry/ingress/baremetalds-metallb/conf/ingress-baremetalds-metallb-conf-ref.yaml
lihongan: The approach for vpshere is still not clear to me, I don't want to consider vsphere at this moment.
And yes, we might reuse it when implementing vsphere, if that is true then we could refactor this, that is the plan.

lihongan:
/pj-rehearse pull-ci-openshift-cluster-ingress-operator-master-e2e-baremetalds-metallb-gatewayapi-conformance pull-ci-openshift-cluster-ingress-operator-release-4.21-e2e-baremetalds-metallb-gatewayapi-conformance pull-ci-openshift-cluster-ingress-operator-release-4.20-e2e-baremetalds-metallb-gatewayapi-conformance

rikatz:
/pj-rehearse ack
tests worked, I could verify that metallb is being used on on-prem
Thanks!

lihongan:
> /pj-rehearse pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-gatewayapi-conformance
> 
> this is weird, I will try one more time otherwise hold and we can investigate later

Looks just the step `ipi-install-install` failed, not related to conformance test

PR #73112, opened 2026-01-05
Fixing some of the names of the day-0 mixed-arch jobs

tvardema:
Some of the jobs were configured differently than their names suggested.  This PR fixes those jobs.

### DIFF
--- a/ci-operator/config/openshift/multiarch/README.md
+++ b/ci-operator/config/openshift/multiarch/README.md
@@ -27,10 +27,10 @@ The architectures will be represented in single character arrangements, with the
 ### Job Name Examples
 * `periodic-ci-openshift-multiarch-master-nightly-4.17-ocp-e2e-gcp-ovn-multi-x-ax`
   * e2e tests run in gcp using the multi payload with `x86_64` control plane and mixed `arm64` and `x86_64` compute nodes.
-* `periodic-ci-openshift-multiarch-master-nightly-4.17-ocp-e2e-gcp-ovn-multi-day-0-x-ax`
-  * e2e tests run in gcp using the multi payload with `x86_64` control plane and mixed `arm64` and `x86_64` compute nodes at deployment time (nodes not added later).
+* `periodic-ci-openshift-multiarch-master-nightly-4.17-ocp-e2e-gcp-ovn-multi-day-0-a-x`
+  * e2e tests run in gcp using the multi payload with `arm64` control plane and `x86_64` compute nodes at deployment time (no additional nodes added later).
 * `periodic-ci-openshift-multiarch-master-nightly-4.17-ocp-e2e-ibmcloud-ovn-multi-x-xz`
-  * e2e tests run in ibmcloud using the multi payload with `x86_64` control plane and mixed `s390x` and `x86_64` compute nodes.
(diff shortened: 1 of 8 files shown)


ci-operator/config/openshift/multiarch/README.md
```diff
@@ -27,7 +27,7 @@ The architectures will be represented in single character arrangements, with the
 ### Job Name Examples
 * `periodic-ci-openshift-multiarch-master-nightly-4.17-ocp-e2e-gcp-ovn-multi-x-ax`
   * e2e tests run in gcp using the multi payload with `x86_64` control plane and mixed `arm64` and `x86_64` compute nodes.
-* `periodic-ci-openshift-multiarch-master-nightly-4.17-ocp-e2e-gcp-ovn-multi-day-0-x-ax`
+* `periodic-ci-openshift-multiarch-master-nightly-4.17-ocp-e2e-gcp-ovn-multi-day-0-a-x`
```
sherine-k, line 30: The description of the job in the line below doesn't match anymore with the periodic job. 
Maybe we can update that too?

ci-operator/config/openshift/multiarch/openshift-multiarch-master__nightly-4.17.yaml
```diff
@@ -696,7 +696,7 @@ tests:
     - ref: ipi-install-heterogeneous
     - ref: openshift-e2e-test
     workflow: openshift-e2e-gcp-ovn
-- as: ocp-e2e-gcp-ovn-multi-day-0-x-ax
+- as: ocp-e2e-gcp-ovn-multi-day-0-a-x
```
sherine-k, line 699: just checking if there was ever a plan that this job be a day0 installation of  a mixed arch (ax) compute cluster?
ie. is there a mistake in the job configuration? or in naming?

sherine-k: Since the AWS test ocp-e2e-aws-ovn-multi-day-0-a-x covers the arm as Control plane, is it worth it to switch this one to test day-0-x-a?

ci-operator/config/openshift/multiarch/README.md
tvardema: Good catch.  I relied too much on the Cursor prompt.  I'll double check the workflow and deployment of the original job to see if we need to update anything... otherwise I will change this line back to how it originally was in the README, as the README is just a series of example job names, not a list of existing jobs.

ci-operator/config/openshift/multiarch/openshift-multiarch-master__nightly-4.17.yaml
tvardema: As mentioned above, I'll review to make sure.  I looked at the workflow at a high-level, but I'll crawl through to verify whether or not it was supposed to be mixed or not.

tvardema: Sorry, didn't see the second message.  First, the workflow used for this job does not include mixed-architecture options, so it is in fact deploying based off the control and compute variables in the job definition.  Second, if we want to test `day-0-x-a` in GCP here, we should do that in a separate PR and simply fix the names here for now.  That way we can iterate with any issues in a PR dedicated to that work.

PR #73113, opened 2026-01-05
Updating scheduling for 4.16 jobs on P/Z for the sustainability team

tvardema:


tvardema:
> @tvardema Hi, For 4.16, we should consider to change config file of 4.16 i.e `openshift-multiarch-master__nightly-4.16-upgrade-from-nightly-4.15.yaml ` instead of `openshift-multiarch-master__nightly-4.17-upgrade-from-nightly-4.16.yaml` Please correct me if I'm wrong, I am referring old PR.

You're absolutely correct, I changed the wrong upgrade job.  Sorry about that!  Will update right now.

PR #73135, opened 2026-01-06
All regions test: fixes and enhancement for AZs

yunjiang29:
* Fixed the issue that edge zones may be listed while checking instance type offerings, now only regular AZ is returned.
* Added `AZ_INDEX` env, now AZ can be selected by particular index rather than default `0`.

### DIFF
--- a/ci-operator/step-registry/cucushift/installer/rehearse/aws/cases/clusters/provision/cucushift-installer-rehearse-aws-cases-clusters-provision-commands.sh
+++ b/ci-operator/step-registry/cucushift/installer/rehearse/aws/cases/clusters/provision/cucushift-installer-rehearse-aws-cases-clusters-provision-commands.sh
@@ -243,28 +243,47 @@ if [[ ${COMPUTE_INSTANCE_TYPE} != "" ]]; then
   yq-v4 eval -i '.compute[0].platform.aws.type = env(COMPUTE_INSTANCE_TYPE)' "${CONFIG}"
 fi
 
-# Patch AZ, only use one AZ to reduce EIP usage
-if [ "$TEST_OBJECT" == "InstanceTypes" ];  then
-  # check the instance type offering
-  
+AZ_NAME=""
+
+case "$TEST_OBJECT" in
+InstanceTypes)
   t1=$(mktemp)
   t2=$(mktemp)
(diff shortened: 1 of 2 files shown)


ci-operator/step-registry/cucushift/installer/rehearse/aws/cases/clusters/provision/cucushift-installer-rehearse-aws-cases-clusters-provision-commands.sh
```diff
@@ -243,28 +243,37 @@ if [[ ${COMPUTE_INSTANCE_TYPE} != "" ]]; then
...
+    AZ_NAME=$(jq -r '[.AvailabilityZones[] | select(.ZoneType=="availability-zone")] | .[0].ZoneName' "${ZONES_JSON}")
+    echo "AZ configuration: Auto select $AZ_NAME for TEST_OBJECT: $TEST_OBJECT"
+    ;;
+  *)
+    echo "AZ configuration: Skip for TEST_OBJECT: $TEST_OBJECT"
+    ;;
+  esac
 else
-  # select the first one
-  ZONE_NAME=$(jq -r '[.AvailabilityZones[] | select(.ZoneType=="availability-zone")] | .[0].ZoneName' "${ZONES_JSON}")
```
jianlinliu, line 261: So before this change, when testing LocalZones/ WavelengthZones, we also configured AZ for it, right ?

yunjiang29: No, LZ/WL's behavior is not changed, we don't set AZ for them.

yunjiang29: btw, it's a little complicated to let user provide ZONE_NAME, as REGION may be mismatch.
In the commit, it's replaced by AZ_INDEX, this may be a good option, user can config it rather than always set to `0`

jianlinliu: >  LZ/WL's behavior is not changed, we don't set AZ for them

Ah, I got it, because the line 264 skip it.

PR #73138, opened 2026-01-06
ARO-23203 | Migrate config-change-detection to Prow Job

oriAdler:


oriAdler:
@stevekuznetsov @deads2k PTAL

PR #73140, opened 2026-01-06
All regions test: Remove old tests

yunjiang29:
The new tests for all regions are available, remove the old workflow and jobs:
 
* Removed the following jobs:
  ```
  aws-ipi-all-local-zones-f60
  aws-ipi-all-regions-set-a-arm-f60
  aws-ipi-all-regions-set-a-f60
  aws-ipi-all-regions-set-b-arm-f60
  aws-ipi-all-regions-set-b-f60
  aws-ipi-all-wavelength-zones-f60
  ```
* Removed `cucushift-installer-rehearse-aws-cases-all-regions`


yunjiang29:
@jianlinliu @gpei @jinyunma @jianli-wei PTAL

PR #73141, opened 2026-01-06
aws upi proxy install enhancement

jianlinliu:
1. Filter out sensitive info from the ci output to avoid `This file contained potentially sensitive information and has been removed` by gitleaks tools.
2. Enhance trap setting to avoid overriding and ensure all logs are collected.
3. Update the conditions of creating the proxy server, e.g: Directories in `SHARED_DIR` is not available across multi steps.
4. When a region (e.g: us-west-1) has <3 AZs, vpc cf stack would fail to create because of missing subnets, add a logic to judge available zones, provide a proper `ZONES_COUNT` parameter for vpc cf template.
5. Update proxy server ignition version and AMI selection to ensure the server running, not broken because of the incompatibility between them.
6. When bootstrap failed, also collect proxy server instance console output to help debugging.
7. Update proxy cf template to use higher version of ignition and IAM role to get s3 bucket data.
8. Fix several proxy server podman deployment issues.
9. Use https proxy instead of http proxy
10. remove useless `ipi-deprovision-proxy` step from `upi-aws-proxy-post` chain.

### DIFF
--- a/ci-operator/step-registry/upi/conf/aws/OWNERS
+++ b/ci-operator/step-registry/upi/conf/aws/OWNERS
@@ -3,3 +3,5 @@ approvers:
 - rna-afk
 - patrickdillon
 - jhixson74
+- jianlinliu
+- yunjiang29
--- a/ci-operator/step-registry/upi/conf/aws/proxy/upi-conf-aws-proxy-ref.metadata.json
+++ b/ci-operator/step-registry/upi/conf/aws/proxy/upi-conf-aws-proxy-ref.metadata.json
@@ -5,7 +5,9 @@
 			"staebler",
 			"rna-afk",
 			"patrickdillon",
-			"jhixson74"
+			"jhixson74",
+			"jianlinliu",
+			"yunjiang29"
 		]
 	}
 }


jianlinliu:
/pj-rehearse pull-ci-openshift-installer-main-e2e-aws-upi-proxy

jianlinliu:
/pj-rehearse pull-ci-openshift-installer-main-e2e-aws-upi-proxy

jianlinliu:
/pj-rehearse periodic-ci-openshift-release-master-nightly-4.19-e2e-aws-ovn-upi

jianlinliu:
/pj-rehearse pull-ci-openshift-installer-main-e2e-aws-upi-proxy

jianlinliu:
/pj-rehearse pull-ci-openshift-installer-main-e2e-aws-upi-proxy

jianlinliu:
/pj-rehearse pull-ci-openshift-installer-main-e2e-aws-upi-proxy

jianlinliu:
/pj-rehearse pull-ci-openshift-installer-main-e2e-aws-upi-proxy

jianlinliu:
/pj-rehearse pull-ci-openshift-installer-main-e2e-aws-upi-proxy

ci-operator/step-registry/upi/conf/aws/proxy/upi-conf-aws-proxy-commands.sh
```diff
@@ -313,44 +321,33 @@ http_port 3128
...
 EOF
 )"
 
-PROXY_IMAGE=registry.ci.openshift.org/origin/4.5:egress-http-proxy
-
 # define squid.sh
 SQUID_SH="$(base64 -w0 << EOF
 #!/bin/bash
-podman run --entrypoint='["bash", "/squid/proxy.sh"]' --expose=3128 --net host --volume /tmp/squid:/squid:Z ${PROXY_IMAGE}
+podman run --entrypoint='["bash", "/squid/proxy.sh"]' --expose=3128 --net host --volume /srv/squid:/squid:Z --volume /srv/squid/log:/var/log/squid:Z ${PROXY_IMAGE}
```
gpei: will 3130 port be exposed? or in Host network mode, no need to specify `--expose` parameter

ci-operator/step-registry/upi/conf/aws/proxy/upi-conf-aws-proxy-commands.sh
```diff
@@ -1,11 +1,26 @@
...
+
+pushd ${SHARED_DIR}
+USER_NAME=$(yq-go r "install-config.yaml" "metadata.name")
+base_domain=$(yq-go r "install-config.yaml" "baseDomain")
+PROXY_DNS="squid.${USER_NAME}.${base_domain}"
+popd
+
 echo "Generating proxy certs..."
-ROOTCA=${SHARED_DIR}/CA
+tmpDir=$(mktemp -d)
```
gpei, line 19: shall we remove the entire ${tmpDir} by the end of the script to avoid the CA/key/pem files were left over?

ci-operator/step-registry/upi/conf/aws/proxy/upi-conf-aws-proxy-commands.sh
```diff
@@ -286,22 +297,19 @@ KEY_PASSWORD="$(base64 -w0 << EOF
 echo ${PROXY_KEY_PASSWORD}
 EOF
 )"
-PROXY_DNS="squid.${USER_NAME}.${base_domain}"
 
 export PROXY_URL="http://${USER_NAME}:${PASSWORD}@${PROXY_DNS}:3128/"
 export TLS_PROXY_URL="https://${USER_NAME}:${PASSWORD}@${PROXY_DNS}:3130/"
 
 echo ${PROXY_URL} > ${SHARED_DIR}/PROXY_URL 
 echo ${TLS_PROXY_URL} > ${SHARED_DIR}/TLS_PROXY_URL
```
gpei: seems the two files were not being used anymore?

- ${SHARED_DIR}/PROXY_URL
- ${SHARED_DIR}/TLS_PROXY_URL

ci-operator/step-registry/upi/conf/aws/proxy/upi-conf-aws-proxy-commands.sh
jianlinliu: Per my testing, if not explicitly expose the port in the parameters,  it still works on the default port mapping. E.g: if the port is listening on 3130 inside the container, even not explicitly expose the port in, the 3130 port is still accessible from outside.

ci-operator/step-registry/upi/conf/aws/proxy/upi-conf-aws-proxy-commands.sh
jianlinliu: Before this PR, the two files are referenced, and printed out the contents in the log when executing `upi-install-aws-cluster-commands.sh`, that lead sensitive info exposed, gitleak will detect it, and remove the whole log file, so in this PR, I just printed out the file path, but not the contents. I prefer to keep it, so that easier to get the value when doing online debug, or future reference when developing new steps.

gpei:
The final step `upi-deprovision-aws-check-destroy` step https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_release/73141/rehearse-73141-pull-ci-openshift-installer-main-e2e-aws-upi-proxy/2009624908442636288/artifacts/e2e-aws-upi-proxy/upi-deprovision-aws-check-destroy/build-log.txt doesn't have a good check logic, as aws cloudformation wait will fail if stack doesn't exist or is in DELETE_FAILED state. 
Do we want to fix it in this PR?

ci-operator/step-registry/upi/conf/aws/proxy/upi-conf-aws-proxy-commands.sh
gpei: got it, maybe we can make it more clear

FYI, here's the reference answer from AI
```
  The command uses --net host (host network mode), which means:

  1. The container shares the host's network namespace directly
  2. All ports the application binds to inside the container are automatically available on the host
  3. Port mapping flags (-p, --expose) are ignored/have no effect

  What Actually Happens

  When squid starts inside the container (via /squid/proxy.sh), it reads the config which defines:

  http_port 3128
  https_port 3130 cert=/squid/tls.crt key=/squid/tls.key cafile=/squid/ca-chain.pem

  Since --net host is used:
  - Squid binds to 0.0.0.0:3128 → directly accessible on host:3128
  - Squid binds to 0.0.0.0:3130 → directly accessible on host:3130

  The --expose=3128 Flag is Misleading

  This flag is actually redundant and confusing here because:

  1. ✅ With --net host, you don't need --expose or -p flags
  2. ❌ It suggests only 3128 is exposed (but 3130 is too!)
  3. ⚠️ --expose only adds metadata when using bridge networking (not host mode)

  Correct Command

  The command should be simplified to:

  podman run --entrypoint='["bash", "/squid/proxy.sh"]' \
    --net host \
    --volume /srv/squid:/squid:Z \
    --volume /srv/squid/log:/var/log/squid:Z \
    ${PROXY_IMAGE}

  Or, if you want to be explicit about both ports (even though it's unnecessary with --net host):

  podman run --entrypoint='["bash", "/squid/proxy.sh"]' \
    --net host \
    --expose=3128 --expose=3130 \
    --volume /srv/squid:/squid:Z \
    --volume /srv/squid/log:/var/log/squid:Z \
    ${PROXY_IMAGE}
```

jianlinliu: I did some research, actually `expose` did not anything, just something like a label, accessing it depends on the network topology. In this case, because of `--net host`, all the ports listening inside the containers can be accessed by default even without `--expose` and `--publish`.

To make it less confused, I am going to change the parameter to use `--publish` to replace it.

ci-operator/step-registry/upi/conf/aws/proxy/upi-conf-aws-proxy-commands.sh
gpei: OK, how about renaming the two files to more meaningful names.
Currently our steps are using 
```
proxy_public_url="http://${proxy_credential}@${bastion_public_ip}:3128"
proxy_private_url="http://${proxy_credential}@${bastion_private_ip}:3128"
echo "${proxy_public_url}" > "${SHARED_DIR}/proxy_public_url"
echo "${proxy_private_url}" > "${SHARED_DIR}/proxy_private_url"
```
for this one, maybe 
```
  echo "${PROXY_URL}" > "${SHARED_DIR}/http_proxy_url"
  echo "${TLS_PROXY_URL}" > "${SHARED_DIR}/https_proxy_url"
```
?

ci-operator/step-registry/gcp/provision/bastionhost/gcp-provision-bastionhost-commands.sh
jianlinliu, line 111: cc @jianli-wei to be aware of the removal, because is it not referenced anywhere.

ci-operator/step-registry/upi/conf/aws/proxy/upi-conf-aws-proxy-commands.sh
```diff
@@ -313,44 +321,33 @@ http_port 3128
...
 EOF
 )"
 
-PROXY_IMAGE=registry.ci.openshift.org/origin/4.5:egress-http-proxy
-
 # define squid.sh
 SQUID_SH="$(base64 -w0 << EOF
 #!/bin/bash
-podman run --entrypoint='["bash", "/squid/proxy.sh"]' --expose=3128 --net host --volume /tmp/squid:/squid:Z ${PROXY_IMAGE}
+podman run --entrypoint='["bash", "/squid/proxy.sh"]' -p 3128:3128 -p 3129:3129 --net host --volume /srv/squid:/squid:Z --volume /srv/squid/log:/var/log/squid:Z ${PROXY_IMAGE}
```
gpei: typo here?
```suggestion
podman run --entrypoint='["bash", "/squid/proxy.sh"]' -p 3128:3128 -p 3130:3130 --net host --volume /srv/squid:/squid:Z --volume /srv/squid/log:/var/log/squid:Z ${PROXY_IMAGE}
```

jianlinliu: done

ci-operator/step-registry/upi/install/aws/cluster/upi-install-aws-cluster-commands.sh
yunjiang29: missing `--region ${AWS_REGION}`

ci-operator/step-registry/upi/conf/aws/proxy/upi-conf-aws-proxy-commands.sh
yunjiang29, line 305: it seems we don't need these shared info.

yunjiang29: is it for the following e2e test?

yunjiang29: Oh, I just saw https://github.com/openshift/release/pull/73141#discussion_r2680719693, let's keep it.

jianlinliu:
@yunjiang29 @gpei I just updated owner list in the last commit, pls help review again.

PR #73150, opened 2026-01-06
Add config to access QCI programmatic

praveenkumar:
We need long term token to have our script run in the konflux smoothly which pull the images from registry.ci.openshift.org, currently we are using personal token which expire in 30 days.


droslean:
@praveenkumar Can you add some description on the PR to justify why you need this?

PR #66981, opened 2025-07-11
WIP: CNTRLPLANE-947: cluster-authentication-operator: add external-oidc conformance periodic

liouk:

PR #73209, opened 2026-01-07
OCPBUGS-71206: OIDC conformance manual presubmits

liouk:
Currently, the jobs added with this PR will fail, because Openshift's conformance suites are not made to be run on external OIDC enabled clusters. Adding manually triggered jobs will facilitate development/fixes in the origin repo because we currently have no jobs to run to validate changes such as https://github.com/openshift/origin/pull/30292.

Once all necessary fixes are in place in origin, the goal is to convert these jobs to periodics; hence this PR replaces https://github.com/openshift/release/pull/66981.


liouk:
Skipping rehearsals as we expect the new jobs to fail.

/pj-rehearse skip

PR #73236, opened 2026-01-08
fixing incorrect flag in cleanup script

sclarkso:
This cleanup job has been broken for 3 weeks.  Digging into it, it is using the flag "--env-type" which appears to be incorrect.  I have changed it to "--mode" which seems correct based on the code.
```
Running: ./test/aro-hcp-tests cleanup resource-groups --expired --env-type no-rp --include-location westus3
+ eval './test/aro-hcp-tests cleanup resource-groups --expired --env-type no-rp --include-location westus3'
++ ./test/aro-hcp-tests cleanup resource-groups --expired --env-type no-rp --include-location westus3
Error: unknown flag: --env-type
unknown flag: --env-type
```


venkateshsredhat (approved):
/lgtm

PR #73269, opened 2026-01-08
console-plugin-nvidia-gpu: Fix multi-arch image mirroring

empovit:
Replace assisted-baremetal-images-publish step with custom inline mirroring step that uses --keep-manifest-list=true flag to properly mirror multi-arch manifest lists to Quay.

The assisted-baremetal-images-publish step doesn't support multi-arch manifest lists and fails with: 'error: the image is a manifest list and contains multiple images - use --filter-by-os to select from'.

This is a temporary workaround. The assisted-baremetal-images-publish step should be fixed to support multi-arch by adding the --keep-manifest-list=true flag to its oc image mirror command.


yakovbeder (commented):
/lgtm

PR #73276, opened 2026-01-08
stackrox(perfscale): pin ocp version

davdhacs:
We want to test with an un-changing OCP version so that differences in metrics are because of stackrox changes and not openshift changes.

https://docs.ci.openshift.org/docs/architecture/ci-operator/#testing-with-an-existing-openshift-release


davdhacs:
/pj-rehearse ack
it is running, and orion is reporting a possible regression
/cc @mtodor
This pins to stable ocp

ci-operator/config/stackrox/stackrox/stackrox-stackrox-nightlies__perf-scale.yaml
```diff
@@ -14,10 +14,9 @@ build_root:
     tag: stackrox-ui-test-0.4.9
 releases:
   latest:
-    candidate:
-      product: ocp
-      stream: nightly
-      version: "4.19"
+    release:
+      channel: stable
+      version: "4.20"
```
mtodor: tbh. I would pin it to strict patch. If we use `stable` channel we could still get instability from `4.20.z` updates. @davdhacs What do you think?

davdhacs: good idea. How about 4.20.8 (Dec 12, then 4.20.9 was released Jan 9)?

davdhacs:

PR #73286, opened 2026-01-08
avoid attempting to overwrite GH branch protection settings for an upstream repo

grokspawn:
Remove branch protection enforcement from prow config.  This would overwrite the set of required checks for the master branch, which makes no sense. 
Instead, this prow config should merely execute the configured workflows with the existing GH required checks.


rashmigottipati (approved):
/lgtm

PR #73302, opened 2026-01-09
Remove CONFIG variable

rsevilla87:
This variable is redundant and not used in the script. Also simplifying the logic a little bit by using the basename command rather than bash variable expansion


jtaleric:
@rsevilla87 while we do not use this var, I do think other teams are using it to point the configs to a URL.

rsevilla87:

PR #73343, opened 2026-01-12
CORENET-6697: add konflux shared pull secret into qe-catalogsource to make it as global

zhaozhanqi:
as almost all olm operators already move to konflux,  here is trying to add konflux shared pull secret as global pull secret of cluster


rioliu-rh:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.21-automated-release-aws-ipi-f999

zhaozhanqi:
> > should `enable-qe-catalogsource-disconnected` be updated as well?
> 
> I think current implementation does not support disconnected env. FBC index image are not mirrored. @zhaozhanqi correct me if I'm wrong

Yes,  I just reply that.  we can use a separate PR for disconnect cluster so that it does not block level0 testing

and I saw all operators cases passed from the CI job

PR #73359, opened 2026-01-12
OLS-2452: Add OBS 4.21 OCP claim

onmete:
Add OBS 4.21 OCP claim


JoaoFula:
@onmete the cluster pool will not be created until this PR is merged. This means any rehearsal test that will result from running make jobs (there should be two affected by it), will not pass because they won't be able to do any cluster claim. 
Either separate it to run the rehearsal tests or we're gonna have to wing it and hope it doesn't break when we do merge it.

PR #73378, opened 2026-01-12
Add TLS scanner CI configuration and step registry

richardsonnick:
Creates a new CI step for running the [tls-scanner](https://github.com/openshift/tls-scanner) against OpenShift.

This is a long running job (can take upwards of 3 hours) that ultimately emits an artifact detailing the TLS settings present for OpenShift components. The tls-scanner iterates through all open ports of each component and attempts to connect using a set of TLS parameters. It records which parameters were successfully negotiated. This gives a picture of the TLS configuration for all externally accessible routes.


richardsonnick:
/pj-rehearse pull-ci-openshift-tls-scanner-main-run-scanner-on-cluster

PR #73423, opened 2026-01-13
ART-14327: Add verify-deps to prometheus-alertmanager 4.18 and lower

adobes1:
Add `verify-deps` job to openshift/prometheus-alertmanager 4.12 - 4.18 to catch issues with incomplete `vendor/` early before it has a chance to cause issues for hermetic builds

### DIFF
--- a/ci-operator/config/openshift/prometheus-alertmanager/openshift-prometheus-alertmanager-release-4.12.yaml
+++ b/ci-operator/config/openshift/prometheus-alertmanager/openshift-prometheus-alertmanager-release-4.12.yaml
@@ -57,6 +57,12 @@ tests:
         requests:
           cpu: 100m
     workflow: ipi-azure
+- as: verify-deps
+  steps:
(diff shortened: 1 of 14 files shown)


jan--f:
Just wondering: how is this different from the `vendor` test? https://github.com/adobes1/release/blob/10cb128ed29bf62240e270ed0b837108fdf5174b/ci-operator/config/openshift/prometheus-alertmanager/openshift-prometheus-alertmanager-release-4.14.yaml#L40

jan--f:
For future readers:
[`verify-deps`](https://github.com/openshift/release/blob/311c828951353f48116decae2389074e2c2bca1c/ci-operator/step-registry/go/verify-deps/go-verify-deps-commands.sh) does more then the existing `vendor` check. Most critically it ignores `.gitignore` to mirror the behavior of cachi2/hermeto during builds. Hence this
/lgtm

PR #73424, opened 2026-01-13
NE-2417: Add e2e-vsphere-static-metallb-gatewayapi-conformance

melvinjoseph86:


ci-operator/step-registry/ingress/metallb-conf/ingress-metallb-conf-workflow.yaml
```diff
@@ -1,17 +1,17 @@
 workflow:
-  as: ingress-baremetalds-metallb
+  as: ingress-metallb-conf
```
lihongan, line 2: Please do not change the workflow name

ci-operator/step-registry/ingress/metallb-conf/ingress-metallb-conf-workflow.yaml
```diff
@@ -1,17 +1,17 @@
...
   steps:
     env:
       DEVSCRIPTS_CONFIG: |
         IP_STACK=v4
         NETWORK_TYPE=OVNKubernetes
     pre:
     - chain: baremetalds-ofcir-pre
-    - ref: ingress-baremetalds-metallb-conf
+    - ref: ingress-metallb-conf
```
lihongan, line 10: Yes, this is right. We just update this step to use new `ref: ingress-metallb-conf` but the workflow name is unchanged.

PR #73428, opened 2026-01-13
Enable FIPS mode on amd clusters

Anna-Koudelkova:
While porting downstream test cases upstream we find out the rule `enable-fips-mode` is currently failing on all our upstream weekly test runs, even though some of the clusters has parameter  `FIPS_ENABLED: true`. 
Manual tests runs confirmed the rule is behaving as it should and the problem is most likely in the CI config of our arm clusters. 

The easiest solution right now is to add the `FIPS_ENABLED: true` to the CI config of some of the amd clusters because we already know it works (FIPS enabled amd cluster is periodically testing the rule downstream).

### DIFF
--- a/ci-operator/config/openshift/origin/openshift-origin-main.yaml
+++ b/ci-operator/config/openshift/origin/openshift-origin-main.yaml
@@ -15,18 +15,6 @@ base_images:
     name: hypershift-operator
     namespace: hypershift
     tag: latest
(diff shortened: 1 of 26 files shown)


Anna-Koudelkova:
/pj-rehearse periodic-ci-ComplianceAsCode-content-master-4.21-e2e-aws-openshift-node-compliance-weekly

PR #73447, opened 2026-01-13
Add new claim-failure-analysis workflow with a failure-analysis post step

calebevans:


ci-operator/step-registry/redhat-appstudio/claim-failure-analysis/redhat-appstudio-claim-failure-analysis-workflow.yaml
```diff
@@ -0,0 +1,28 @@
...
+    - ref: konflux-ci-unregister-sprayproxy
+      best_effort: true
+    - chain: gather
+      optional_on_success: true
+    - ref: redhat-appstudio-gather
+      optional_on_success: true
+    - ref: redhat-appstudio-report
+      best_effort: true
+    - ref: redhat-appstudio-prow-failure-analysis
+      best_effort: true
```
psturc: hmm what if we set this to `optional_on_success: true` instead? i.e. when the job succeeds, don't run the analysis?

calebevans: Updated it

PR #73486, opened 2026-01-14
Increasing timeout for hypershift-hostedcluster-destroy-hostedcluster

FilipB:
Increasing the timeout to 1h as an attempt to fix https://issues.redhat.com/browse/DPTP-4648


fjglira:
@FilipB it's still needed this change?

PR #73487, opened 2026-01-14
Download kind in tmp

alicefr:
In order to avoid any conflicts with existing directories, let download kind in tmp and then move it to the installation directory.

### DIFF
--- a/ci-operator/step-registry/trusted-execution-clusters/ref/operator/beaker-kind-provision/trusted-execution-clusters-ref-operator-beaker-kind-provision-commands.sh
+++ b/ci-operator/step-registry/trusted-execution-clusters/ref/operator/beaker-kind-provision/trusted-execution-clusters-ref-operator-beaker-kind-provision-commands.sh
@@ -410,26 +410,27 @@ echo "[SUCCESS] kubectl installed"
 echo "[INFO] Installing Kind..."
 KIND_VERSION="v0.30.0"
 KIND_PATH="/usr/local/bin/kind"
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/trusted-execution-clusters/ref/operator/beaker-kind-provision/trusted-execution-clusters-ref-operator-beaker-kind-provision-commands.sh
```diff
@@ -410,26 +410,27 @@ echo "[SUCCESS] kubectl installed"
...
 
-    if file ./kind | grep -q "ELF.*executable"; then
+    if file "${OUT_FILE}" | grep -q "ELF.*executable"; then
         echo "[INFO] Download verified: ELF executable"
     else
         echo "[ERROR] Downloaded file is not an ELF executable"
         exit 1
     fi
 
     chmod +x kind
```
Jakob-Naucke: OUT_FILE too?

Jakob-Naucke (approved):

alicefr:
@yalzhang it seems now that the kind cluster takes too long to be boostrapped. Not quite sure if the timeout comes from the CI.

alicefr:
mmh I see. TBH, it will be better to install docker, kind and kubectl only once and not on every PR. We could also cache the images on the node which are pulled from the external quay. We created a [script](https://github.com/trusted-execution-clusters/operator/blob/main/scripts/pre-pull-images.sh) that does this

PR #73489, opened 2026-01-14
cfg: set image updater verbosity to 0

raelga:
This pull request makes a minor change to the default debug level for the automation image update step. The default value for the `VERBOSITY` environment variable is now set to "0" (no debug output), instead of "2" (verbose debug output).

* Set the default value of `VERBOSITY` to "0" to disable debug output by default in `aro-hcp-automation-image-update-ref.yaml`.


hbhushan3:
Superseded by https://github.com/openshift/release/pull/73525

PR #73491, opened 2026-01-14
[DNM][testonly] see if we can tune memory requests and GOMEMLIMIT to not crash with OOM

maschmid:
Do not merge,

testing only,

trying memory requests to see if we can get rid of the flakiness caused by OOMs on the "go test" process.


maschmid:
/pj-rehearse periodic-ci-openshift-knative-serverless-operator-main-420-mesh-e2e-c

PR #73525, opened 2026-01-14
Fix ACM helm-chart commit logic

hbhushan3:
Removed redundant conditional check in the ACM helm-chart rendering workflow that was unreachable due to an earlier exit condition. Here are debug logs when there are no acm changes, yet we commit them - https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/73525/rehearse-73525-periodic-ci-Azure-ARO-HCP-main-image-updater-image-updater-tooling/2011521527270346752

In line 572 and line 84. In line 84 there is no acm/mce update, commiting them can be seen in line 572.

Changes
Now unconditionally runs yamlfmt and update-helm-fixtures after helm chart rendering. Commits only when there are actual changes after formatting/fixtures

### DIFF
--- a/ci-operator/step-registry/aro-hcp/automation/image-update/aro-hcp-automation-image-update-ref.yaml
+++ b/ci-operator/step-registry/aro-hcp/automation/image-update/aro-hcp-automation-image-update-ref.yaml
@@ -46,7 +46,7 @@ ref:
     documentation: |-
       The URL of the Prow job.
   - name: VERBOSITY
-    default: "3"
+    default: "0"
     documentation: |-
       The debug level. 0 for no debug, 1 for debug, 2 for debug with verbose output.
   credentials:


hbhushan3:
/pj-rehearse periodic-ci-Azure-ARO-HCP-main-image-updater-image-updater-tooling

ci-operator/step-registry/aro-hcp/automation/image-update/aro-hcp-automation-image-update-commands.sh
```diff
@@ -156,16 +156,17 @@ git commit --all --quiet --message "chore: execute image-updater for all compone
 info "acm: rendering ACM helm-charts"
 run make -C acm helm-charts
```
raelga: I don't get this change, the update-helm-fixture, the yaml format and the commit should run only if the ACM `helm-carts` target generates changes. So, the outcome should be the same running inside or outside the conditional block.

PR #73543, opened 2026-01-15
Adding shard to avoid ci job timeout

gangwgr:
https://github.com/openshift/cluster-kube-apiserver-operator/pull/1993#issuecomment-3750621322
e2e-gcp-operator-serial-ote-openshift-e2e-test took 4h10m16s, considering that tests adding shard ci job

### DIFF
--- a/ci-operator/jobs/openshift/cluster-kube-apiserver-operator/openshift-cluster-kube-apiserver-operator-main-presubmits.yaml
+++ b/ci-operator/jobs/openshift/cluster-kube-apiserver-operator/openshift-cluster-kube-apiserver-operator-main-presubmits.yaml
@@ -1253,22 +1253,23 @@ presubmits:
     - ^main$
     - ^main-
     cluster: build08
-    context: ci/prow/e2e-gcp-operator-serial-ote
+    context: ci/prow/e2e-gcp-operator-serial-ote-1of2
     decorate: true
     labels:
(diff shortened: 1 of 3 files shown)


wangke19:
Better to exactly 2 commits following this structure:
1. Commit 1: Infrastructure code changes only                                                                                                                                                                                 
2. Commit 2: Automatically produced code/Dependency updates (go.mod, go.sum) and vendor/ directory changes

ci-operator/config/openshift/cluster-kube-apiserver-operator/openshift-cluster-kube-apiserver-operator-main.yaml
```diff
@@ -453,6 +453,7 @@ tests:
     workflow: openshift-e2e-cert-rotation-short-azure
 - as: e2e-gcp-operator-serial-ote
   optional: true
+  shard_count: 2
```
p0lyn0mial, line 456: this job used to work in the past just fine.

i suspect https://github.com/openshift/cluster-kube-apiserver-operator/pull/1997#pullrequestreview-3661598179 might have cause it.

`WaitForAPIServerToStabilizeOnTheSameRevision` is very expensive in terms of time, for sure will delay any test that use it.


could you check if that method is absolutely required ? again in the past the operator tests just worked.

PR #73552, opened 2026-01-15
Remove loaded upgrade jobs

rsevilla87:
If Im not mistaken, nobody is looking at these jobs, at some point we'll have to revamp and add them back


ci-operator/config/openshift-eng/ocp-qe-perfscale-ci/openshift-eng-ocp-qe-perfscale-ci-main__rosa-4.21-candidate-x86-loaded-upgrade-from-4.20.yaml
mukrishn, line 1: Instead of deleting them, can we just disable the cron schedules (for rosa and rosa-hcp - 4.21, 4.22 at least) we use it for release readiness tests.

PR #73571, opened 2026-01-15
cfg: bump prod slice quota

raelga:
When a prod rollout is executed, we ran 7 E2E suites in paralel (one against each region). Currently the E2E steps are timming out waiting for the lease.

Resource quota configuration updates:

* Increased both `max-count` and `min-count` for `aro-hcp-prod-quota-slice` from 3 to 7 in `core-services/prow/02_config/_boskos.yaml`, allowing for more concurrent resources in production.


raelga (commented):
/approve

PR #73609, opened 2026-01-16
Run dnf update

venkateshsredhat:


ci-operator/config/Azure/ARO-HCP/Azure-ARO-HCP-main__periodic.yaml
```diff
@@ -8,10 +8,11 @@ images:
 - dockerfile_literal: |
     FROM src
     RUN cat /etc/dnf/dnf.conf || true
+    RUN dnf update -y && dnf clean all
```
janboll: isn't clean all removing the cache again?

PR #73617, opened 2026-01-16
Remove ref related gcp pd service account

chao007:
`ref: ipi-conf-gcp-persistent-disk-workload-identity-federation` only works in the gcp cco env. 
Remove the job which not suitable


chao007:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.21-multi-nightly-gcp-ipi-custom-endpoint-mini-perm-arm-f14

jianli-wei (approved):

PR #73620, opened 2026-01-16
OCPBUGS-73907: hypershift: combine docs build and deploy into single step

celebdor:
## Summary
- Combine the hypershift-docs-build and hypershift-docs-deploy-cloudflare steps into a single step
- This avoids passing the built docs through SHARED_DIR which has a ~3MB limit (backed by K8s secret)
- The compressed docs tarball exceeds this limit, causing the job to fail

## Test plan
- [ ] Rehearsal job passes on the dependent PR

Depends-on: openshift/hypershift#7482

### DIFF
--- a/ci-operator/step-registry/hypershift/docs/build-and-deploy/hypershift-docs-build-and-deploy-commands.sh
+++ b/ci-operator/step-registry/hypershift/docs/build-and-deploy/hypershift-docs-build-and-deploy-commands.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+
+set -o nounset
+set -o errexit
+set -o pipefail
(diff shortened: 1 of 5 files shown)


ci-operator/step-registry/hypershift/docs/build-and-deploy/hypershift-docs-build-and-deploy-commands.sh
bryan-cox, line 1: Doesn't have to be in this PR but it would be nice if this job used an image that had the prereq things needed - mkdocs, node.js, wrangler

bryan-cox (commented):
/approve

PR #73678, opened 2026-01-19
CNTRLPLANE-2535: Add serial OTE CI job for cli-manager-operator

ropatil010:
Hi Team, 

Add new e2e-aws-operator-serial-ote job to run OpenShift Tests Extension (OTE) tests in serial mode for the cli-manager-operator.

This job supports the OTE migration implemented in https://github.com/openshift/cli-manager-operator/pull/977 and https://github.com/openshift/cli-manager-operator/pull/979.

The job uses the openshift-e2e-test step reference with serial execution enforced via TEST_ARGS=-c 1 and is marked as optional to allow PR merges while OTE tests are being developed.

hold till the PR: https://github.com/openshift/cli-manager-operator/pull/979 gets merged
/hold 
/assign @wangke19 @gangwgr @sandeepknd

### DIFF
--- a/ci-operator/config/openshift/cli-manager-operator/openshift-cli-manager-operator-main.yaml
+++ b/ci-operator/config/openshift/cli-manager-operator/openshift-cli-manager-operator-main.yaml
@@ -23,14 +23,21 @@ promotion:
   to:
   - name: "4.22"
     namespace: ocp
+releases:
+  initial:
+    integration:
+      name: "4.22"
+      namespace: ocp
+  latest:
+    integration:
+      include_built_images: true
+      name: "4.22"
+      namespace: ocp
 resources:
   '*':
     requests:
       cpu: 100m
       memory: 200Mi
-tag_specification:
-  name: "4.22"
-  namespace: ocp
 tests:
 - as: verify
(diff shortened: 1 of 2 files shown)


ardaguclu:
Would it make sense if we remove e2e-aws-operator?. Because e2e-aws-operator runs the same tests anyway?. So that we can remove e2e-operator-test from cli manager operator. WDYT?

ropatil010:
> Would it make sense if we remove e2e-aws-operator?. Because e2e-aws-operator runs the same tests anyway?. So that we can remove e2e-operator-test from cli manager operator. WDYT?

Agree, but currently checking our newly added job runs pass rate, once we conclude both are working same then we can remove this earlier job(e2e-operator-test )?

ardaguclu:
/reopen
I'm sorry about closing incorrect PR mistakenly

ropatil010:
/hold 
As of now openshift-tests does not support with OLM operators. 
as per discussion: https://redhat-internal.slack.com/archives/C07RDCVEYJG/p1768985999194229?thread_ts=1768547398.592969&cid=C07RDCVEYJG

PR #73691, opened 2026-01-19
kubevirt: add new version of prom-metrics-linter image

akrejcir:
Added new version of `quay.io/kubevirt/prom-metrics-linter`. 
It will be used in a future PR.


avlitman:
cc @0xFelix can you approve maybe?

PR #73731, opened 2026-01-20
feat(hypershift): add Azure scale-from-zero flags

LiangquanLi930:
Add scale-from-zero-provider and scale-from-zero-creds flags for Azure.


mgencur:
No JIRA associated. Closing. Feel free to re-open with more information if this is still valid.

/close

PR #73740, opened 2026-01-20
Cluster Health Analyzer Prepare Integration Tests

DavidRajnoha:
Starts Proxy for thanos querying and installs promtool, prerequisities for integration tests.

### DIFF
--- a/ci-operator/step-registry/cluster-health-analyzer/cluster-health-analyzer-commands.sh
+++ b/ci-operator/step-registry/cluster-health-analyzer/cluster-health-analyzer-commands.sh
@@ -2,70 +2,98 @@
 
 set -euo pipefail
 
-# Function to collect debug artifacts - runs on exit (success or failure)
-function collectArtifacts {
-  echo "=== Collecting debug artifacts ==="
-  
-  # Save pod descriptions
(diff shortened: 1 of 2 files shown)


ci-operator/step-registry/cluster-health-analyzer/cluster-health-analyzer-commands.sh
```diff
@@ -24,41 +49,54 @@ function collectArtifacts {
...
+make install-integration-test-tools
+export PATH="/tmp:${PATH}"
+end_step "Installing dependencies"
+
+log "=== Running make targets ==="
+
+start_step "Starting thanos proxy"
+make proxy &
+PROXY_PID=$!
+sleep 5
```
rioloc: This may be the cause of some flakyness. I'd suggest to use a for loop to check the status of the thanos proxy (you may use something like `pidof`). You could try n attempts, one every m seconds.

ci-operator/step-registry/cluster-health-analyzer/cluster-health-analyzer-commands.sh
rioloc, line 29: Have you verified netcat is installed by default?

DavidRajnoha: I did by running the rehearsal and the tests passing with nc being used there.

PR #73742, opened 2026-01-20
Add gcp-hcp-infra-base build image with UBI 10 + pip install uv

patjlm:
## Summary

Adds a new build image `gcp-hcp-infra-base` using UBI 10 with dnf + pip installation to support Python scripts in the gcp-hcp-infra repository.

## Changes

**New Build Image**: `gcp-hcp-infra-base`
- Base: UBI 10 (10.1-1767602397) with Python 3.12
- Terraform 1.14.3-1 from Hashicorp repo (dnf)
- uv 0.9.26 via pip install (no EPEL or binary downloads)
- Base tools: git, make, jq (dnf)

## Why UBI 10 + pip install uv

**UBI 10 provides Python 3.12** which is required by gcp-hcp-infra Python scripts (UBI 9 only has Python 3.9)

**pip install approach**:
- ✅ No EPEL repository needed
- ✅ No binary downloads from GitHub
- ✅ Versioned and reproducible (uv==0.9.26)
- ✅ Simple dnf-based installation via python3-pip package

## Implementation Details

- Single RUN statement for all dnf installs to optimize layer caching and avoid re-downloading repo metadata
- TF_PLUGIN_CACHE_DIR set to /tmp (not /go since this isn't a Go builder image)
- ConfigChange trigger for automatic rebuilds
- TODO: Will add imageChange trigger once ocp/ubi:10 imagestream is available (pending mirroring from PR #73753)

## Validation

Tested locally with actual gcp-hcp-infra repository:
```bash
make check-generated-files  # ✅ Passes
```

Both Python scripts work correctly:
- `argocd/scripts/render.py` (requires Python 3.13 - uv downloads automatically)
- `scripts/infra.py` (requires Python 3.11+ - uv manages dependencies)

## Next Steps

After this PR merges and the build image is rebuilt (~15-30 minutes):
1. Update gcp-hcp-infra CI configuration to use `gcp-hcp-infra-base` as build_root
2. Add `check-generated-files` presubmit test

## Related

- gcp-hcp-infra PR #300: https://github.com/openshift-online/gcp-hcp-infra/pull/300
- Jira: [GCP-354](https://issues.redhat.com//browse/GCP-354) - https://issues.redhat.com/browse/GCP-354
- UBI 10 mirroring: PR #73753

### DIFF
--- a/clusters/app.ci/supplemental-ci-images/hcm-gcp/gcp-hcp-infra-base.yaml
+++ b/clusters/app.ci/supplemental-ci-images/hcm-gcp/gcp-hcp-infra-base.yaml
@@ -1,0 +1,49 @@
+kind: BuildConfig
+apiVersion: build.openshift.io/v1
+metadata:
+  namespace: ci
+  name: gcp-hcp-infra-base
+spec:
+  successfulBuildsHistoryLimit: 1
+  failedBuildsHistoryLimit: 2
+  output:
+    pushSecret:
+      name: registry-push-credentials-ci-images-mirror
+    to:
+      kind: DockerImage
+      name: quay.io/openshift/ci:ci_gcp-hcp-infra-base_latest
+  source:
+    dockerfile: |
+      FROM quay-proxy.ci.openshift.org/openshift/ci:ocp_ubi_10
+      ARG TERRAFORM_VERSION
+      ARG UV_VERSION
+      ENV TF_PLUGIN_CACHE_DIR="/tmp/.terraform-cache"
+      RUN mkdir -p --mode=2775 ${TF_PLUGIN_CACHE_DIR}
+      # Install all tools in one RUN to avoid re-downloading repo metadata
+      RUN dnf config-manager --add-repo https://rpm.releases.hashicorp.com/RHEL/hashicorp.repo && \
+          dnf install -y \
+            git \
+            make \
+            jq \
+            python3-pip \
+            terraform-${TERRAFORM_VERSION} && \
+          pip install uv==${UV_VERSION} && \
+          dnf clean all
+    type: Dockerfile
+  strategy:
+    dockerStrategy:
+      buildArgs:
+      - name: TERRAFORM_VERSION
+        value: "1.14.3-1"
+      - name: UV_VERSION
+        value: "0.9.26"
+      imageOptimizationPolicy: SkipLayers
+    type: Docker
+  triggers:
+  - type: ConfigChange
+  - imageChange:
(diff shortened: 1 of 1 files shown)


clusters/app.ci/supplemental-ci-images/hcm-gcp/gcp-hcp-infra-base.yaml
```diff
@@ -0,0 +1,50 @@
...
+          pip install uv==${UV_VERSION} && \
+          dnf clean all
+    type: Dockerfile
+  strategy:
+    dockerStrategy:
+      buildArgs:
+      - name: TERRAFORM_VERSION
+        value: "1.14.3-1"
+      - name: UV_VERSION
+        value: "0.9.26"
```
billmvt, line 39: How will we keep these versions aligned with what gets deployed for dev/stage/prod? What about keeping up to date on CVE patching / z-stream updates? Probably not a blocker to solve for this PR but something to think about.

clusters/app.ci/supplemental-ci-images/hcm-gcp/gcp-hcp-infra-base.yaml
```diff
@@ -0,0 +1,50 @@
...
+      imageOptimizationPolicy: SkipLayers
+    type: Docker
+  triggers:
+  - type: ConfigChange
+  # TODO: Add imageChange trigger once ocp/ubi:10 ImageStream is created (image is mirrored, ImageStream pending)
```
cblecker: Can we try to add this in right now? It might not pull until first use. Everything should be in place as far as I can tell.

patjlm: I can add it but:
- it will be useless until the ubi:10 imagestreamtag appears
- it will be useless until this imagestream is updated regularly (by itself or by an extrnal process)

I'll uncomment

patjlm: uncommented

clusters/app.ci/supplemental-ci-images/hcm-gcp/gcp-hcp-infra-base.yaml
patjlm: +1 i'll try to think how we can sync these with values from the gcp-hcp-infra repo. Would be for an other PR as I think we should implement those missing tests asap

PR #73770, opened 2026-01-21
Dynamic AWS Credentials Injection for ACM CLC Tests

tanfengshuang:
### Summary

  Updates `acm-tests-clc-create` step to dynamically extract AWS credentials from cluster profile instead of using hardcoded credentials in secrets.

  ### Problem

  - AWS credentials currently hardcoded in `secret-options-yaml`
  - Manual updates required when credentials rotate
  - Cannot easily use different AWS accounts for different tests

  ### Solution

  Extract credentials from `${CLUSTER_PROFILE_DIR}/.awscred` and inject into `options.yaml`:

 **Benefits:**
  - Eliminates test failures during credential rotation
  - No manual vault secret updates required
  - Zero test downtime during credential rotation

### DIFF
--- a/ci-operator/step-registry/acm/tests/clc-create/acm-tests-clc-create-commands.sh
+++ b/ci-operator/step-registry/acm/tests/clc-create/acm-tests-clc-create-commands.sh
@@ -1,49 +1,64 @@
 #!/bin/bash
-set -o nounset
-set -o errexit
-set -o pipefail
+set -euxo pipefail; shopt -s inherit_errexit
 
-# The variables defined in this step come from files in the `SHARED_DIR` and credentials from Vault.
-SECRETS_DIR="/tmp/secrets"
+# --- Variables ---
+typeset secretsDir="/tmp/secrets"
+typeset optionFile="./options.yaml"
+typeset awsCredFile="${CLUSTER_PROFILE_DIR}/.awscred"
 
-# Get the creds from ACMQE CI vault and run the automation on pre-exisiting HUB
-SKIP_OCP_DEPLOY="false"
-if [[ $SKIP_OCP_DEPLOY == "true" ]]; then
-    echo "------------ Skipping OCP Deploy = $SKIP_OCP_DEPLOY ------------"
-    cp ${SECRETS_DIR}/ci/kubeconfig $SHARED_DIR/kubeconfig
-    cp ${SECRETS_DIR}/ci/kubeadmin-password $SHARED_DIR/kubeadmin-password
-fi 
+# Get the creds from ACMQE CI vault and run the automation on pre-existing HUB
+if [[ "${SKIP_OCP_DEPLOY:-false}" == "true" ]]; then
+    : "------------ Skipping OCP Deploy = ${SKIP_OCP_DEPLOY} ------------"
+    cp "${secretsDir}/ci/kubeconfig" "${SHARED_DIR}/kubeconfig"
+    cp "${secretsDir}/ci/kubeadmin-password" "${SHARED_DIR}/kubeadmin-password"
+fi
 
-######### added for debugging purposes #########
-# export DISPLAY=:99
-# Xvfb :99 &
-# # Xvfb :99 -screen 0 1024x768x16 2>/dev/null &
+cp "${secretsDir}/clc-interop/secret-options-yaml" "${optionFile}"
 
-# export CYPRESS_video=false
-# export CYPRESS_videoCompression=false
-###############################################
+# Update the AWS credentials in options.yaml from cluster profile
+if [[ -f "${awsCredFile}" ]]; then
+    typeset awsAccKeyID=
+    typeset awsAccKeyToken=
 
-export KUBECONFIG=${SHARED_DIR}/kubeconfig
+    set +x
+    awsAccKeyID="$(sed -nE 's/^\s*aws_access_key_id\s*=\s*//p;T;q' "${awsCredFile}")"
+    awsAccKeyToken="$(sed -nE 's/^\s*aws_secret_access_key\s*=\s*//p;T;q' "${awsCredFile}")"
 
-cp ${SECRETS_DIR}/clc-interop/secret-options-yaml ./options.yaml
+    [ -n "${awsAccKeyID}" ] && [ -n "${awsAccKeyToken}" ]
 
-# Set the dynamic vars based on provisioned hub cluster.
-CYPRESS_BASE_URL=$(oc whoami --show-console)
-export CYPRESS_BASE_URL
+    : "Updating credentials in ${optionFile}..."
+    yq -o json eval . "${optionFile}" |
+    jq -c \
+          --arg awsAccKeyID "${awsAccKeyID}" \
+          --arg awsAccKeyToken "${awsAccKeyToken}" \
+        '
+          .connections.apiKeys.aws|=(
+                .awsAccessKeyID=$awsAccKeyID |
+                .awsSecretAccessKey=$awsAccKeyToken
+            )
+        ' |
+    yq -p json -o yaml eval . > "${optionFile}.tmp"
+    mv -f "${optionFile}.tmp" "${optionFile}"
+    set -x
 
-CYPRESS_HUB_API_URL=$(oc whoami --show-server)
-export CYPRESS_HUB_API_URL
+    unset awsAccKeyID awsAccKeyToken
+fi
 
-CYPRESS_OPTIONS_HUB_PASSWORD=$(cat $SHARED_DIR/kubeadmin-password)
-export CYPRESS_OPTIONS_HUB_PASSWORD
-# Version of spoke cluster to be provisioned.
-CYPRESS_CLC_OCP_IMAGE_VERSION=$(cat $SECRETS_DIR/clc/ocp_image_version)
-export CYPRESS_CLC_OCP_IMAGE_VERSION
+: "Executing CLC interop commands..."
+set +x
+export CYPRESS_OPTIONS_HUB_PASSWORD=
+CYPRESS_OPTIONS_HUB_PASSWORD="$(cat "${SHARED_DIR}/kubeadmin-password")"
+set -x
 
-CLOUD_PROVIDERS=$(cat $SECRETS_DIR/clc/ocp_cloud_providers)
-export CLOUD_PROVIDERS
+CYPRESS_BASE_URL="$(oc whoami --show-console)" \
+CYPRESS_HUB_API_URL="$(oc whoami --show-server)" \
+CYPRESS_CLC_OCP_IMAGE_VERSION="$(cat "${secretsDir}/clc/ocp_image_version")" \
+CLOUD_PROVIDERS="$(cat "${secretsDir}/clc/ocp_cloud_providers")" \
+bash +x ./execute_clc_interop_commands.sh || :
 
-# run the test execution script
-bash +x ./execute_clc_interop_commands.sh
+unset CYPRESS_OPTIONS_HUB_PASSWORD
 
-cp -r reports $ARTIFACT_DIR/
+: "Copying artifacts..."
+cp -r reports "${ARTIFACT_DIR}/"
+
+true


ci-operator/step-registry/acm/tests/clc-create/acm-tests-clc-create-commands.sh
etirta:
```suggestion
set -euxo pipefail; shopt -s inherit_errexit
```

ci-operator/step-registry/acm/tests/clc-create/acm-tests-clc-create-commands.sh
```diff
@@ -27,6 +27,37 @@ export KUBECONFIG=${SHARED_DIR}/kubeconfig
...
+    fi
+
+    if ! yq -i ".connections.apiKeys.aws.awsSecretAccessKeyID = strenv(AWS_SECRET_ACCESS_KEY)" ./options.yaml; then
+        : "ERROR: Failed to update awsSecretAccessKeyID in options.yaml"
+        exit 1
+    fi
+else
+    : "ERROR: ${CLUSTER_PROFILE_DIR}/.awscred not found. Cannot proceed without valid AWS credentials from cluster profile."
+    exit 1
+fi
```
etirta, line 45: First, we can just export to Env. Var. from the get go, IF these var. really need to be exported.
Second, do we REALLY need it to be an env. var? As mentioned in the Best Practice doc., avoid using Env. Var. unless it's really necessary. We can use `yq + jq` to update the `options.yaml`.
With `xtrace` enabled, the secret setting may be exposed to the log unmasked (unfortunately we can't do Shell Process Substitution trick in here, so we should ).
For easier maintenance, if a file path is used several times, it's worth to assign it to a Shell Var., so if the file path change, only 1 place modification is needed.


```suggestion
# Update the AWS credentials in options.yaml from cluster profile
typeset awsCredFile="${CLUSTER_PROFILE_DIR}/.awscred"
typeset optionFile='./options'
if [ -f "${awsCredFile}" ]; then
    : "Extracting AWS credentials from cluster profile..."
    typeset awsAccKeyID= awsAccKeyToken=

    set +x
    awsAccKeyID="$(sed -nE 's/^\s*aws_access_key_id\s*=\s*//p;T;q' "${awsCredFile}")"
    awsAccKeyToken="$(sed -nE 's/^\s*aws_secret_access_key\s*=\s*//p;T;q' "${awsCredFile}")"
    set -x

    [ -n "${awsAccKeyID}" ] && [ -n "${awsAccKeyToken}" ]

    : "Updating AWS credentials in ${optionFile}..."
    ( set +x
        yq -o json eval . "${optionFile}" |
        jq -c \
              --arg awsAccKeyID "${awsAccKeyID}" \
              --arg awsAccKeyToken "${awsAccKeyToken}" \
            '
              .connections.apiKeys.aws|=(
                    .awsAccessKeyID=$awsAccKeyID |
                    .awsSecretAccessKey=$awsAccKeyToken
                )
            ' |
        yq -p json -o yaml eval . > "${optionFile}.tmp"
        mv -f "${optionFile}.tmp" "${optionFile}"
    true )
else
    : "ERROR: ${awsCredFile} not found. Cannot proceed without valid AWS credentials from cluster profile."
    exit 1
fi
```

ci-operator/step-registry/acm/tests/clc-create/acm-tests-clc-create-commands.sh
etirta: With `errexit` enabled, please add `true` at the end.


```suggestion
cp -r reports $ARTIFACT_DIR/

true
```

etirta, line 1: General note is please follow our [Best Practice](https://spaces.redhat.com/spaces/MPEXIENG/pages/728700228/Best+Practices):
  - Use Shell Var. instead of Env. Var. if possible.
  - Use established naming convention, so it's easier for user to distinguish Shell vs Env. Var.

ci-operator/step-registry/acm/tests/clc-create/acm-tests-clc-create-commands.sh
```diff
@@ -27,6 +27,37 @@ export KUBECONFIG=${SHARED_DIR}/kubeconfig
...
+
 # Set the dynamic vars based on provisioned hub cluster.
 CYPRESS_BASE_URL=$(oc whoami --show-console)
 export CYPRESS_BASE_URL
```
etirta: Best practice is to declare first, then assigned. Use double-quote on assignment.
Do this for all applicable variables through out the script.

```suggestion
export CYPRESS_BASE_URL=
CYPRESS_BASE_URL="$(oc whoami --show-console)"
```

If all these Env. Var. are only use during execution of `./execute_clc_interop_commands.sh`, then better way is to do inline Env. Var. assignment, except of password passing, because we need to protect it from being displayed.

```sh
set +x
export CYPRESS_OPTIONS_HUB_PASSWORD=
CYPRESS_OPTIONS_HUB_PASSWORD="$(cat $SHARED_DIR/kubeadmin-password)"
set -x
CYPRESS_BASE_URL="$(oc whoami --show-console)" \
    CYPRESS_CLC_OCP_IMAGE_VERSION="$(cat "${SECRETS_DIR}/clc/ocp_image_version")" \
    CLOUD_PROVIDERS="$(cat "${SECRETS_DIR}/clc/ocp_cloud_providers")" \
    bash +x ./execute_clc_interop_commands.sh
unset CYPRESS_OPTIONS_HUB_PASSWORD
```

tanfengshuang:
/pj-rehearse periodic-ci-stolostron-policy-collection-main-ocp4.20-interop-opp-aws periodic-ci-stolostron-policy-collection-main-ocp4.21-interop-opp-aws

ci-operator/step-registry/acm/tests/clc-create/acm-tests-clc-create-commands.sh
tanfengshuang: > With `errexit` enabled, please add `true` at the end.

Sure. I made AI use Best practice, and it helped add it at the end.

ci-operator/step-registry/acm/tests/clc-create/acm-tests-clc-create-commands.sh
tanfengshuang: Sorry, I was thinking about using Best Practice to improve my script, and it was not ready for review yesterday. But thanks for your review, I like your yq handle.

I've removed the code below to support non-AWS platforms:
else
    : "ERROR: ${awsCredFile} not found. Cannot proceed without valid AWS credentials from cluster profile."
    exit 1

ci-operator/step-registry/acm/tests/clc-create/acm-tests-clc-create-commands.sh
tanfengshuang: Done this improvement. Thank you!

Initially, I didn't plan to refactor the entire script. However, I really appreciate your review feedback - I‘ve learned a lot from it. I'll have another shell script to update soon, and I'd appreciate your review on that as well!

ci-operator/step-registry/acm/tests/clc-create/acm-tests-clc-create-commands.sh
tanfengshuang: Export first now.

For CYPRESS_OPTIONS_HUB_PASSWORD, this password is for the temporary OCP test cluster that gets destroyed immediately after the test run completes. Since this cluster is ephemeral and never exposed beyond the CI environment, I don't think the `set +x`/`unset` protection is necessary here.

However, if you prefer to keep the same pattern for consistency, I'm happy to add it.

ci-operator/step-registry/acm/tests/clc-create/acm-tests-clc-create-commands.sh
tanfengshuang: Thank you. Have reviewed all the Shell Env. and updated them to camelCase.

ci-operator/step-registry/acm/tests/clc-create/acm-tests-clc-create-commands.sh
```diff
@@ -1,49 +1,74 @@
...
+    : "------------ Skipping OCP Deploy = ${SKIP_OCP_DEPLOY} ------------"
+    cp "${secretsDir}/ci/kubeconfig" "${SHARED_DIR}/kubeconfig"
+    cp "${secretsDir}/ci/kubeadmin-password" "${SHARED_DIR}/kubeadmin-password"
+fi
 
-######### added for debugging purposes #########
-# export DISPLAY=:99
-# Xvfb :99 &
-# # Xvfb :99 -screen 0 1024x768x16 2>/dev/null &
+export KUBECONFIG="${SHARED_DIR}/kubeconfig"
```
etirta: I have question about this line. Why not use the original KUBECONFIG value that is set by the `Step`? For every `Step` it should set this env. var. (along with others such as `SHARED_DIR`, etc).

For earlier `Step`s that (before a `Step` store something at `${SHARED_DIR}/kubeconfig`, this env. var. will be set to a file that does not exist. For later `Step`, it will be set to a file that contain a copy of `${SHARED_DIR}/kubeconfig`. See https://docs.ci.openshift.org/docs/architecture/step-registry/#a-note-on-kubeconfig for understanding how this env. var. works in CI Operator.

ci-operator/step-registry/acm/tests/clc-create/acm-tests-clc-create-commands.sh
etirta: For this particular case it's true, but it make us have to think carefully. A better approach is to train ourselves to always unset env. var. containing secret. This practice reduce the risk of we forget to unset a longer term or even permanent secret.

BTW, unsetting shell var. is less significant, because it won't propagate to Child, unless sub-shell, and we control the sub-shell, but it may still be a good practice.

So please add it. And please consider to use inline setting of env. var. for the rest of non-secret values.

ci-operator/step-registry/acm/tests/clc-create/acm-tests-clc-create-commands.sh
tanfengshuang: @etirta, I agree with you. I also think it's not needed here. This script was not filed by me, but I can improve it now.

Done.

ci-operator/step-registry/acm/tests/clc-create/acm-tests-clc-create-commands.sh
tanfengshuang: Done. unset secret var. and use inline setting of env. var. now. Thank you!

tanfengshuang:
/pj-rehearse periodic-ci-stolostron-policy-collection-main-ocp4.20-interop-opp-aws periodic-ci-stolostron-policy-collection-main-ocp4.21-interop-opp-aws

ci-operator/step-registry/acm/tests/clc-create/acm-tests-clc-create-commands.sh
```diff
@@ -1,49 +1,67 @@
...
+CYPRESS_OPTIONS_HUB_PASSWORD="$(cat "${SHARED_DIR}/kubeadmin-password")" 
+set -x
 
-CLOUD_PROVIDERS=$(cat $SECRETS_DIR/clc/ocp_cloud_providers)
-export CLOUD_PROVIDERS
+CYPRESS_BASE_URL="$(oc whoami --show-console)" \
+CYPRESS_HUB_API_URL="$(oc whoami --show-server)" \
+CYPRESS_CLC_OCP_IMAGE_VERSION="$(cat "${secretsDir}/clc/ocp_image_version")" \
+CLOUD_PROVIDERS="$(cat "${secretsDir}/clc/ocp_cloud_providers")" \
+bash +x ./execute_clc_interop_commands.sh || :
```
etirta, line 57: Why not fail here if the execution of `execute_clc_interop_commands.sh` error out? Generally speaking we should fail fast, unless there is a good reason we should not do `... || true` (the `:` will always yield `true`).

ci-operator/step-registry/acm/tests/clc-create/acm-tests-clc-create-commands.sh
```diff
@@ -1,49 +1,67 @@
...
 
-# run the test execution script
-bash +x ./execute_clc_interop_commands.sh
+unset CYPRESS_OPTIONS_HUB_PASSWORD
 
-cp -r reports $ARTIFACT_DIR/
+: "Copying artifacts..."
+cp -r reports "${ARTIFACT_DIR}/" || :
```
etirta: Ditto here. I would think we would like to fail if line fail. Or?

ci-operator/step-registry/acm/tests/clc-create/acm-tests-clc-create-commands.sh
tanfengshuang: "|| true" here is to make sure this step won't exit the job(exit 1 will exit the whole prow job), and then the rest steps can go ahead.

The junit.xml is generated in this step, and Prow will read this file to figure out FAIL or PASS.

etirta: Got it, if this is deliberate then I OK with it.

ci-operator/step-registry/acm/tests/clc-create/acm-tests-clc-create-commands.sh
etirta: The only thing that you might consider is cy removing `xtrace` for the whole command, you may miss the parameters (via Env. Var.) of the actual command that is executed. Unfortunately my original suggestion seems to cause the whole log is snipped.

So my new suggestion is, back to what it was, but just disable `xtrace` on sensitive information.
```suggestion
set +x
export CYPRESS_OPTIONS_HUB_PASSWORD=
CYPRESS_OPTIONS_HUB_PASSWORD="$(cat $SHARED_DIR/kubeadmin-password)"
set -x
export CYPRESS_BASE_URL=
CYPRESS_BASE_URL="$(oc whoami --show-console)"
export CYPRESS_HUB_API_URL=
CYPRESS_HUB_API_URL="$(oc whoami --show-server)"
export CYPRESS_CLC_OCP_IMAGE_VERSION=
CYPRESS_CLC_OCP_IMAGE_VERSION="$(cat "${SECRETS_DIR}/clc/ocp_image_version")"
export CLOUD_PROVIDERS=
CLOUD_PROVIDERS="$(cat "${SECRETS_DIR}/clc/ocp_cloud_providers")"
bash +x ./execute_clc_interop_commands.sh
unset CYPRESS_OPTIONS_HUB_PASSWORD
```
You will see all parameter except the secret (a compromise that we must take because the secret is coming from non secret location, FYI, if the secret is coming from secret location, we don't need to worry, because the log will be sanitized).

If you don't think seeing these final parameter in the log is not giving any value, that it's fine the way you do it. But if you think it may give post-mortem troubleshooting value, then I suggest only turn of `xtrace` for setting the password.

etirta, line 42: > @etirta, could you please review this part again? I use yq to handle directly instead of yq+jq+file. Thank you!

Sorry @tanfengshuang I missed your comment in the above. Any reason why you go back to `yq`? My problem with `yq` is is can only deal with env. var. and as I mentioned setting secret in env. var. have higher risk because it's inherited to child.

I'm actually file a request to `yq` author to allow `--arg` argument like in `jq`, but as of now he's non-committal. So until then, I think we should use `jq` as intermediary, even though it's less efficient. But security is paramount and should take precedence than efficiency.

ci-operator/step-registry/acm/tests/clc-create/acm-tests-clc-create-commands.sh
tanfengshuang: I canceled "|| :" in my new update. 
Log reports can always be generated after testing. "|| :" here may hide other potential issues.

ci-operator/step-registry/acm/tests/clc-create/acm-tests-clc-create-commands.sh
tanfengshuang: Yes, agree with you, I have changed it back. I was trying to use the fewest lines, but seems it's not the best.

tanfengshuang: Yes, I was wondering why you use yq+jq instead of only yq, I've known your concern now.
I agree with you that using yq+jq to avoid the security risk. 
I have reverted the code, and I also put the following line in set +x/-x, or the log will be removed because of the suspicion of secrects leakage. 

- [ -n "${awsAccKeyID}" ] && [ -n "${awsAccKeyToken}" ]
- This file contained potentially sensitive information and has been removed.

tanfengshuang:
/pj-rehearse periodic-ci-stolostron-policy-collection-main-ocp4.20-interop-opp-aws

ci-operator/step-registry/acm/tests/clc-create/acm-tests-clc-create-commands.sh
etirta: Ahhhh.... That's why the log got removed. Good catch and yes you are right that line should have been inside code block where the `xtrace` is temporarily disabled.

Just FYI, other technique to temporarily disable this is use sub-shell (because we don't set any var. to be used later outside the block):
```
(set +x
    cp "${secretsDir}/clc-interop/secret-options-yaml" "${optionFile}"

    # Update the AWS credentials in options.yaml from cluster profile
    if [[ -f "${awsCredFile}" ]]; then
        typeset awsAccKeyID=
        typeset awsAccKeyToken=

        awsAccKeyID="$(sed -nE 's/^\s*aws_access_key_id\s*=\s*//p;T;q' "${awsCredFile}")"
        awsAccKeyToken="$(sed -nE 's/^\s*aws_secret_access_key\s*=\s*//p;T;q' "${awsCredFile}")"

        [ -n "${awsAccKeyID}" ] && [ -n "${awsAccKeyToken}" ]

        echo "Updating credentials in ${optionFile}..."
        yq -o json eval . "${optionFile}" |
        jq -c \
              --arg awsAccKeyID "${awsAccKeyID}" \
              --arg awsAccKeyToken "${awsAccKeyToken}" \
            '
              .connections.apiKeys.aws|=(
                    .awsAccessKeyID=$awsAccKeyID |
                    .awsSecretAccessKey=$awsAccKeyToken
                )
            ' |
        yq -p json -o yaml eval . 1> "${optionFile}.tmp"
        mv -f "${optionFile}.tmp" "${optionFile}"
    fi
true)
```
Advantage, we don't have to do `unset` and `set -x` (re-enable the `xtrace`). But make sure we use `echo` in this block instead of using `:`. The caveat on using this approach is, if there is some variables we need to set and need to be used later outside this block, it will requires some other tricks, as any changed inside sub-shell ONLY pertain to sub-shell. The parent shell remain unchanged.

This is just FYI for future options, no need to update this PR. It's safe enough, it's just a matter of preference.

PR #73777, opened 2026-01-21
Add notification for Core OVN triggered jobs to slack integration.

dceara:
PR #71890 added slack integration.  Extend it to report triggered jobs too.


dceara:
CC: @josecastillolema

PR #73782, opened 2026-01-21
CMP-3812: Enable FIPS mode on amd clusters

Anna-Koudelkova:
While porting downstream test cases upstream we find out the rule `enable-fips-mode` is currently failing on all our upstream weekly test runs, even though some of the clusters has parameter `FIPS_ENABLED: true`.
Manual tests runs confirmed the rule is behaving as it should and the problem is most likely in the CI config of our arm clusters.

The easiest solution right now is to add the `FIPS_ENABLED: true` to the CI config of some of the amd clusters because we already know it works (FIPS enabled amd cluster is periodically testing the rule downstream).

## Summary by CodeRabbit

This updates the OpenShift CI configuration for `ComplianceAsCode/content` to run the affected weekly AWS compliance jobs in FIPS mode on known-good amd clusters. In practical terms, the `e2e-aws-openshift-platform-compliance-weekly` and `e2e-aws-openshift-node-compliance-weekly` jobs were adjusted across the supported 4.16, 4.18, 4.20, and 4.22 configs so the upstream weekly runs use `FIPS_ENABLED=true` and can validate the `enable-fips-mode` rule reliably.

### DIFF
--- a/ci-operator/config/ComplianceAsCode/content/ComplianceAsCode-content-master__4.16.yaml
+++ b/ci-operator/config/ComplianceAsCode/content/ComplianceAsCode-content-master__4.16.yaml
@@ -557,6 +557,7 @@ tests:
     cluster_profile: quay-aws
     env:
       BASE_DOMAIN: quay.devcluster.openshift.com
+      FIPS_ENABLED: "true"
     test:
     - as: test
       cli: latest
@@ -582,6 +583,7 @@ tests:
     cluster_profile: quay-aws
     env:
       BASE_DOMAIN: quay.devcluster.openshift.com
+      FIPS_ENABLED: "true"
     test:
     - as: test
       cli: latest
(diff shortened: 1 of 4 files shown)


Anna-Koudelkova:
/pj-rehearse periodic-ci-ComplianceAsCode-content-master-4.18-e2e-aws-openshift-node-compliance-weekly

Anna-Koudelkova:
/pj-rehearse periodic-ci-ComplianceAsCode-content-master-4.18-e2e-aws-openshift-node-compliance-weekly

vickeybrown:
I noticed this only updates up to 4.20, will that be an issue for 4.22 tests? If so, is there a common spot for us to put this setting where all the versions can pull from or would we have to just remember to enable FIPS every time there's a new Openshift version?

Anna-Koudelkova:
> I noticed this only updates up to 4.20, will that be an issue for 4.22 tests? If so, is there a common spot for us to put this setting where all the versions can pull from or would we have to just remember to enable FIPS every time there's a new Openshift version?

It won't be an issue for 4.22, just the 4.22 file did not exist at the time of origin of this PR :slightly_smiling_face: I have updated it. 
Regarding the central location of the settings - I am not aware of anything like that, so far we have been doing it for all the releases manually AFAIK - Once new OCP is released, we create new testing jobs.  Sounds interesting though! 
I was aiming to enable FIPS on every other OCP version, so we could be fine for a while.

PR #73793, opened 2026-01-21
feat(rhdh): add rerun-failed-tests CI job configuration

gustavolira:
## Summary

Add CI configuration for the new `/test rerun-failed-tests` command for the RHDH project. This command re-executes only the tests that failed in the previous `e2e-ocp-helm` run, optimizing CI time and resources.

### New Step Registry
- `ci-operator/step-registry/redhat-developer/rhdh/ocp/rerun-failed-tests/`
  - `OWNERS`
  - `redhat-developer-rhdh-ocp-rerun-failed-tests-ref.yaml`
  - `redhat-developer-rhdh-ocp-rerun-failed-tests-commands.sh`

### Modified CI Config
- `ci-operator/config/redhat-developer/rhdh/redhat-developer-rhdh-main.yaml`
  - Added new test `rerun-failed-tests` (optional, manually triggered)

## New Command

After this PR is merged, the following command will be available in RHDH PRs:

```
/test rerun-failed-tests
```

## Dependencies

This PR depends on: https://github.com/redhat-developer/rhdh/pull/4037

🤖 Generated with [Claude Code](https://claude.ai/code)


ci-operator/config/redhat-developer/rhdh/redhat-developer-rhdh-main.yaml
```diff
@@ -38,6 +38,25 @@ tests:
     test:
     - ref: redhat-developer-rhdh-ocp-helm
     workflow: generic-claim
+- always_run: false
+  as: rerun-failed-tests
```
zdrapela, line 42: I would stick to the prefix `e2e-ocp-helm-` so we know which platform and install method it is.

ci-operator/step-registry/redhat-developer/rhdh/ocp/rerun-failed-tests/redhat-developer-rhdh-ocp-rerun-failed-tests-commands.sh
```diff
@@ -0,0 +1,187 @@
...
+    done
+fi
+
+echo "========== Current branch =========="
+echo "Current branch: $(git branch --show-current)"
+echo "Using Image: ${QUAY_REPO}:${TAG_NAME}"
+
+echo "========== Test Execution =========="
+echo "Executing openshift-ci-tests.sh for rerun-failed-tests"
+bash ./.ibm/pipelines/openshift-ci-tests.sh
```
zdrapela, line 187: Does it have any counterpart PR in rhdh repo?

ci-operator/step-registry/redhat-developer/rhdh/ocp/rerun-failed-tests/redhat-developer-rhdh-ocp-rerun-failed-tests-commands.sh
```diff
@@ -0,0 +1,187 @@
+#!/bin/bash
```
zdrapela, line 1: The result of this `rerun-failed-tests` job would be only informative, because we still require to have a green pass on `e2e-ocp-helm`.
Will this help much? The actual Playwright run is like 7 mins for the `showcase` project and 5 mins for the `showcase-rbac` project. And it will be even lower when we proceed with plugin test extraction.

ci-operator/step-registry/redhat-developer/rhdh/ocp/rerun-failed-tests/redhat-developer-rhdh-ocp-rerun-failed-tests-commands.sh
zdrapela: Sorry, see it now. But the PR is closed: https://github.com/redhat-developer/rhdh/pull/4037

ci-operator/step-registry/redhat-developer/rhdh/ocp/rerun-failed-tests/redhat-developer-rhdh-ocp-rerun-failed-tests-commands.sh
gustavolira: I believe the main issue isn’t the execution time itself. The problem is that when you rerun the full suite, you sometimes hit new failures due to flaky tests, and that turns the process into a real struggle to get the PR green.

It would likely be much easier to unblock the process by rerunning only the tests that previously failed, avoiding the risk of introducing new failures from tests that had already passed.

My idea is to find a way to allow the PR to be eligible for merge when rerun-failed-tests passes, without having to depend on e2e-ocp-helm when it gets triggered again.

Does that make sense?

zdrapela: If I skip the discussion whether we need this or whether it's the right idea...

Now we have the `e2e-ocp-helm` job marked as required. If this one doesn't pass, the PR is not allowed to merge. Even if `rerun-failed-tests` job passes. Do you have a plan to get over this?

gustavolira: The idea I have is: when rerun-failed-tests passes, it would use the GitHub Status API to update the ci/prow/e2e-ocp-helm commit status to success. This way, from Tide/Prow's perspective, the required check is green and the PR becomes eligible for merge.

The flow would be:

  1. e2e-ocp-helm runs and fails (e.g., due to flaky tests)
  2. Developer triggers /test rerun-failed-tests
  3. The job re-runs only the previously failed tests
  4. If they all pass, the script calls the GitHub Status API to mark ci/prow/e2e-ocp-helm as success on the commit
  5. PR is now mergeable

The main motivation is avoiding the "flaky test loop", when you /test e2e-ocp-helm again to retry, the full suite can introduce new flaky failures on tests that previously passed, making it very hard to get the PR green.

zdrapela: I think that this can be nice to have at times, but the occasions when this is needed would be becoming rare, as we keep investing in stabilizing the tests, and more importantly, with the plugin test extraction.

If this won't be too hard to achieve properly, ensuring we don't introduce broken tests to the codebase by doing this, it's okay to have it.

PR #73814, opened 2026-01-21
osde2e- protect main branch, update hypershift job to run only on core code change

ritmun:
protect main branch
update hypershift job to run only on core code change
also fix regex for md files

### DIFF
--- a/ci-operator/config/openshift/osde2e/openshift-osde2e-main.yaml
+++ b/ci-operator/config/openshift/osde2e/openshift-osde2e-main.yaml
@@ -24,7 +24,7 @@ tests:
     make check
   container:
     from: src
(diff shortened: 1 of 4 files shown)


ci-operator/config/openshift/osde2e/openshift-osde2e-main.yaml
```diff
@@ -41,14 +41,14 @@ tests:
     out/osde2e test --secret-locations ${SECRET_LOCATIONS} --configs ${CONFIGS}
   container:
     from: src
+  run_if_changed: ^(pkg\/common\/|pkg\/e2e\/)$
```
YiqinZhang: `run_if_changed: ^(pkg\/common\/|pkg\/e2e\/)` matches any path that starts with pkg/common/ or pkg/e2e/

PR #73835, opened 2026-01-22
telco: PP: configure the isolated and reserved cpus on gcp

shajmakh:
GCP cluster profile uses ipi-gcp flow which by default uses 6 vCPUs for compute machines (see `step-registry/ipi/conf/gcp/ipi-conf-ref.yaml`).The performance profile suites configures a profile with `reserved: "0"` and `isolated: "1-3"` (see https://github.com/openshift/cluster-node-tuning-operator/pull/909), unless environment vars are specificed. In general this is the good practice to include all node's cpus in the PP cpu section, but reason why we need this now is that we have some new tests that requires most the cpus to be all distributed using PP (see https://github.com/openshift/cluster-node-tuning-operator/pull/1432#issuecomment-3772349918).

In this commit we start updating only the affected job on which the test would run, later we will need to add this setting to all other jobs that consume ipi-gcp cluster configuration.
Note: this is subject to change should the CPU specifications on GCP get modified.

### DIFF
--- a/ci-operator/config/openshift/cluster-node-tuning-operator/openshift-cluster-node-tuning-operator-main.yaml
+++ b/ci-operator/config/openshift/cluster-node-tuning-operator/openshift-cluster-node-tuning-operator-main.yaml
@@ -150,7 +150,8 @@ tests:
     test:
     - as: test
       cli: latest
-      commands: CLUSTER=mcp-only make cluster-deploy-pao && make pao-functests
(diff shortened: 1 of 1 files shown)


shajmakh:
/pj-rehearse pull-ci-openshift-cluster-node-tuning-operator-main-e2e-gcp-pao-updating-profile

PR #73872, opened 2026-01-22
Add e2e job for osac-test-infra

trewest:


ci-operator/step-registry/osac-project/baremetal/test/osac-project-baremetal-test-ref.yaml
```diff
@@ -0,0 +1,12 @@
+ref:
+  as: osac-project-baremetal-test
+  from: dev-scripts
```
adriengentil, line 3: maybe we'll want to have our own image here, where we install our tooling (python, ansible, ...), but it's fine for the moment.

OWNERS_ALIASES
```diff
@@ -546,3 +546,9 @@ aliases:
   - afcollins
   - sraviteja-maker
   - HughNhan
+  osac-cicd:
+  - adriengentil
+  - jhernand
+  - akshaynadkarni
+  - larsks
+  - trewest
```
adriengentil, line 554:
```suggestion
  - trewest
  - tzvatot 
```

ci-operator/step-registry/osac-project/baremetal/test/osac-project-baremetal-test-ref.yaml
```diff
@@ -0,0 +1,12 @@
...
+  grace_period: 10m
+  timeout: 3h0m0s
+  commands: osac-project-baremetal-test-commands.sh
+  resources:
+    requests:
+      cpu: "3"
+      memory: 600Mi
+    limits:
+      memory: 2Gi
```
adriengentil:
```suggestion
    requests:
      cpu: 100m
      memory: 200Mi
```

let's start with the minimum, and increase as needed

OWNERS_ALIASES
trewest: Added but he is not in the openshift org so I'm not if he will be automatically removed or not

PR #73912, opened 2026-01-23
Prow change to get actual tag if catsrc has :latest

tbuskey:
If CATALOG_SOURCE_IMAGE == :latest, env-cm will lookup the latest X.Y.Z-unix_epoch tag and replace it. Any other tag will be left alone. If env-cm cannot find a tag, :latest will be left.

All the prowjobs now have :latest.  create-prowjob defaults to :latest and can be overridden.  Konflux already overrides the tag

By not changing the prowjob, the results can go to testgrid.

By substituting at runtime, the catsrc and tests can log the actual build and its timestamp instead of :latest

https://issues.redhat.com/browse/KATA-4617

### DIFF
--- a/ci-operator/step-registry/sandboxed-containers-operator/env-cm/sandboxed-containers-operator-env-cm-commands.sh
+++ b/ci-operator/step-registry/sandboxed-containers-operator/env-cm/sandboxed-containers-operator-env-cm-commands.sh
@@ -40,7 +40,7 @@ EOF
 latest_catsrc_image_tag() {
     local api_url="https://quay.io/api/v1/repository/redhat-user-workloads/ose-osc-tenant/osc-test-fbc/tag/"
     local page=1
-    local max_pages=60 # safety limit, typically exits much earlier
+    local max_pages=20 # safety limit, typically exits much earlier
 
     while [ "$page" -le "$max_pages" ]; do
         local resp
@@ -97,6 +97,10 @@ if [[ "$TEST_RELEASE_TYPE" == "Pre-GA" ]]; then
   fi
 
   create_catsrc "${CATALOG_SOURCE_NAME}" "${CATALOG_SOURCE_IMAGE}"
+
+  # Save resolved CATALOG_SOURCE_IMAGE for subsequent steps
+  echo "CATALOG_SOURCE_IMAGE=${CATALOG_SOURCE_IMAGE}" > "${SHARED_DIR}/catalog-source-image.env"
+  echo "Saved resolved CATALOG_SOURCE_IMAGE to ${SHARED_DIR}/catalog-source-image.env"
 else
   if [[ -n "$CATALOG_SOURCE_IMAGE" ]]; then
     echo "CATALOG_SOURCE_IMAGE can only be used when TEST_RELEASE_TYPE==Pre-GA ($CATALOG_SOURCE_IMAGE)"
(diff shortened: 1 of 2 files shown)


ci-operator/step-registry/sandboxed-containers-operator/create-prowjob/README.md
```diff
@@ -143,11 +145,28 @@ ci-operator/step-registry/sandboxed-containers-operator/create-prowjob/sandboxed
...
+   - If the tag is `:latest`, it queries the Quay API to find the most recent `X.Y.Z-unix_epoch` tag
+   - If the tag is anything else (specific version or SHA), it is passed through unchanged
+3. **Supported tag formats**:
+   - `latest` - resolved to newest `X.Y.Z-unix_epoch` tag at runtime
+   - `X.Y.Z-unix_epoch` (e.g., `1.11.1-1766149846`) - passed through unchanged
+   - SHA (e.g., `sha256:abc123...` or short SHA) - passed through unchanged
+4. **Source**: `quay.io/redhat-user-workloads/ose-osc-tenant/osc-test-fbc`
+5. **Method**: Quay API with pagination (max 5 pages)
+
+### Benefits
```
ldoktor: Not sure whether we need "benefits" section in our documentation.

ci-operator/step-registry/sandboxed-containers-operator/create-prowjob/README.md
```diff
@@ -236,13 +255,20 @@ The script includes comprehensive validation:
...
 ```bash
 # Test connectivity to Quay API
 curl -sf "https://quay.io/api/v1/repository/redhat-user-workloads/ose-osc-tenant/osc-test-fbc/tag/?limit=10&page=1"
 
-# Check for matching tags manually
+# Check for matching tags manually (X.Y.Z-unix_epoch format)
 curl -s "https://quay.io/api/v1/repository/redhat-user-workloads/ose-osc-tenant/osc-test-fbc/tag/" | \
-  jq '.tags[] | select(.name | test("^[0-9]+\\.[0-9]+(\\.[0-9]+)?-[0-9]+$")) | .name'
+  jq -r '.tags[] | select(.name | test("^[0-9]+\\.[0-9]+\\.[0-9]+-[0-9]+$")) | "\(.start_ts) \(.name)"' | \
+  sort -nr | head -5
```
ldoktor, line 259: While checking this it might be a good idea to add a note about many entries in different format and the necessity to use "?page=2" (3, 4, 5, 5, ...) to get any tags. Today it's page=5 to get the first one.

ci-operator/step-registry/sandboxed-containers-operator/env-cm/sandboxed-containers-operator-env-cm-commands.sh
```diff
@@ -41,16 +44,31 @@ latest_catsrc_image_tag() {
...
+          jq -r '.tags[]? | select(.name | test("^[0-9]+\\.[0-9]+\\.[0-9]+-[0-9]+$")) | "\(.start_ts) \(.name)"')
+
+        if [ -n "${page_tags}" ]; then
+            all_tags="${all_tags}${page_tags}"$'\n'
         fi
 
         ((page++))
+
+        # Limit to first 5 pages to avoid excessive API calls
+        if [ "$page" -gt 5 ]; then
```
ldoktor: It used to be 20, is 5 sufficient? (today there is 1 tag on 5th page so I'm afraid we might need more pages...)

PR #73914, opened 2026-01-23
add retries to git clone command to avoid transient failures

mohit-sheth:
---                                                                              
  Summary                                                                          
                                                                                   
  - Add retry logic for git clone operations to prevent transient network failures 
  from causing job failures                                                        
  - Create a shared retry library (retry-lib.sh) in SHARED_DIR that can be sourced 
  by subsequent steps                                                              
  - Implements exponential backoff with 3 retry attempts (5s, 10s, 20s delays)     
                                                                                   
  Details                                                                          
                                                                                   
  The workers-scale step now creates a shared retry library at                     
  ${SHARED_DIR}/retry-lib.sh containing a retry_git_clone function. All subsequent 
  perfscale test steps source this library and use retry_git_clone instead of git  
  clone for cloning e2e-benchmarking and orion repositories.                       
                                                                                   
  This approach mirrors how proxy-conf.sh is shared across steps, making the retry 
  logic available to all steps in the chain without requiring changes to container 
  images.                                                                          
                                                                                   
  Files Modified                                                                   
                                                                                   
  - 48 step-registry command scripts updated                                       
  - Affects: openshift-qe/, cloud-bulldozer/e2e-benchmarking/, and                 
  telcov10n/orion/*

### DIFF
--- a/ci-operator/step-registry/cloud-bulldozer/e2e-benchmarking/cluster-density-ms/cloud-bulldozer-e2e-benchmarking-cluster-density-ms-commands.sh
+++ b/ci-operator/step-registry/cloud-bulldozer/e2e-benchmarking/cluster-density-ms/cloud-bulldozer-e2e-benchmarking-cluster-density-ms-commands.sh
@@ -7,6 +7,14 @@ set -x
 # Source shared retry library if available
 if [[ -f "${SHARED_DIR}/retry-lib.sh" ]]; then
     source "${SHARED_DIR}/retry-lib.sh"
+else
+    echo "retry-lib.sh not found in ${SHARED_DIR}"
+fi
+
+# Guarantee fallback
+if ! declare -F retry_git_clone >/dev/null; then
+    echo "retry_git_clone not defined; falling back to plain git clone (no retries)"
+    retry_git_clone() { git clone "$@"; }
 fi
 
 cat /etc/os-release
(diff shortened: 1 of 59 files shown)


ci-operator/step-registry/cloud-bulldozer/e2e-benchmarking/cluster-density-ms/cloud-bulldozer-e2e-benchmarking-cluster-density-ms-commands.sh
```diff
@@ -3,14 +3,20 @@ set -o errexit
...
+
 cat /etc/os-release
 oc config view
 oc projects
 pushd /tmp
 
 
 if [[ "$JOB_TYPE" == "presubmit" ]] && [[ "$REPO_OWNER" = "cloud-bulldozer" ]] && [[ "$REPO_NAME" = "e2e-benchmarking" ]]; then
-    git clone https://github.com/${REPO_OWNER}/${REPO_NAME}
+    retry_git_clone https://github.com/${REPO_OWNER}/${REPO_NAME}
```
mcornea, line 27: The `retry_git_clone` function will be undefined when the library is missing.  We could add a fallback function something like:

```
if [[ -f "${SHARED_DIR}/retry-lib.sh" ]]; 
    source "${SHARED_DIR}/retry-lib.sh"   
else                                      
    retry_git_clone() { git clone "$@"; } 
fi                                        
```

mohit-sheth: makes sense
updated

PR #73918, opened 2026-01-23
OCPBUGS-74383: update openshift/ go.mod if needed for d/s merges

jluhrsen:
openshift/go.mod has some dependency on ../go-controller/go.mod and if those happen to change we need to update openshift/ as well. This will add a new commit on top of the downstream merge PR when it happens

### DIFF
--- a/ci-operator/step-registry/github/downstream-sync/github-downstream-sync-commands.sh
+++ b/ci-operator/step-registry/github/downstream-sync/github-downstream-sync-commands.sh
@@ -86,6 +86,31 @@ if ! git merge "upstream/${DEFAULT_BRANCH}"; then
   CONFLICT=true
 fi
 
+echo "🔧 Syncing openshift/go.mod with upstream dependencies…"
+pushd openshift > /dev/null
+if go mod tidy; then
+  popd > /dev/null
+  # Check if there are any changes to commit
+  if ! git diff --quiet openshift/go.mod openshift/go.sum; then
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/github/downstream-sync/github-downstream-sync-commands.sh
```diff
@@ -86,15 +86,37 @@ if ! git merge "upstream/${DEFAULT_BRANCH}"; then
...
 fi
 
+echo "🔧 Syncing openshift/go.mod with upstream dependencies…"
+pushd openshift > /dev/null
+if go mod tidy && go mod vendor; then
+  popd > /dev/null
+  # Check if there are any changes to commit
+  if ! git diff --quiet openshift/go.mod openshift/go.sum openshift/vendor/; then
+    echo "   📝 Changes detected in openshift/go.mod, committing…"
+    git add openshift/go.mod openshift/go.sum openshift/vendor/
```
jcaamano: But we are not vendoring, are we? Why the change to vendor now?

jluhrsen: My bad. `go mod vendor` is my own brain's muscle memory mistake. it's removed now

jluhrsen: @jcaamano [here](https://github.com/openshift/ovn-kubernetes/pull/2962) is the latest d/s merge made from this PR. look ok to you?

PR #17985, opened 2021-04-22
Move some platform agnostic e2e jobs from gcp to azure

petr-muller:

PR #73934, opened 2026-01-26
NO-JIRA: cvo: add a serial e2e test with TP-enabled

hongkailiu:
To run https://github.com/openshift/cluster-version-operator/pull/1306 correctly.

/hold

### DIFF
--- a/ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-main.yaml
+++ b/ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-main.yaml
@@ -107,6 +107,14 @@ tests:
     test:
     - chain: openshift-e2e-test-ota-qe
     workflow: cucushift-installer-rehearse-gcp-ipi
+- as: e2e-agnostic-ovn-techpreview-serial
+  skip_if_only_changed: ^docs/|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|PROJECT|LICENSE)$
+  steps:
+    cluster_profile: azure-2
+    env:
+      FEATURE_SET: TechPreviewNoUpgrade
+      TEST_SUITE: openshift/conformance/serial
+    workflow: openshift-e2e-azure
 - as: e2e-agnostic-ovn-upgrade-into-change
   skip_if_only_changed: ^docs/|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|PROJECT|LICENSE)$
   steps:
--- a/ci-operator/jobs/openshift/cluster-version-operator/openshift-cluster-version-operator-main-presubmits.yaml
+++ b/ci-operator/jobs/openshift/cluster-version-operator/openshift-cluster-version-operator-main-presubmits.yaml
@@ -297,6 +297,79 @@ presubmits:
     - ^main$
     - ^main-
     cluster: build07
+    context: ci/prow/e2e-agnostic-ovn-techpreview-serial
+    decorate: true
+    labels:
+      ci-operator.openshift.io/cloud: azure4
+      ci-operator.openshift.io/cloud-cluster-profile: azure-2
+      ci.openshift.io/generator: prowgen
+      pj-rehearse.openshift.io/can-be-rehearsed: "true"
+    name: pull-ci-openshift-cluster-version-operator-main-e2e-agnostic-ovn-techpreview-serial
+    rerun_command: /test e2e-agnostic-ovn-techpreview-serial
+    skip_if_only_changed: ^docs/|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|PROJECT|LICENSE)$
+    spec:
+      containers:
+      - args:
+        - --gcs-upload-secret=/secrets/gcs/service-account.json
+        - --image-import-pull-secret=/etc/pull-secret/.dockerconfigjson
+        - --lease-server-credentials-file=/etc/boskos/credentials
+        - --report-credentials-file=/etc/report/credentials
+        - --secret-dir=/secrets/ci-pull-credentials
+        - --target=e2e-agnostic-ovn-techpreview-serial
+        command:
+        - ci-operator
+        image: quay-proxy.ci.openshift.org/openshift/ci:ci_ci-operator_latest
+        imagePullPolicy: Always
+        name: ""
+        resources:
+          requests:
+            cpu: 10m
+        volumeMounts:
+        - mountPath: /etc/boskos
+          name: boskos
+          readOnly: true
+        - mountPath: /secrets/ci-pull-credentials
+          name: ci-pull-credentials
+          readOnly: true
+        - mountPath: /secrets/gcs
+          name: gcs-credentials
+          readOnly: true
+        - mountPath: /secrets/manifest-tool
+          name: manifest-tool-local-pusher
+          readOnly: true
+        - mountPath: /etc/pull-secret
+          name: pull-secret
+          readOnly: true
+        - mountPath: /etc/report
+          name: result-aggregator
+          readOnly: true
+      serviceAccountName: ci-operator
+      volumes:
+      - name: boskos
+        secret:
+          items:
+          - key: credentials
+            path: credentials
+          secretName: boskos-credentials
+      - name: ci-pull-credentials
+        secret:
+          secretName: ci-pull-credentials
+      - name: manifest-tool-local-pusher
+        secret:
+          secretName: manifest-tool-local-pusher
+      - name: pull-secret
+        secret:
+          secretName: registry-pull-credentials
+      - name: result-aggregator
+        secret:
+          secretName: result-aggregator
+    trigger: (?m)^/test( | .* )e2e-agnostic-ovn-techpreview-serial,?($|\s.*)
+  - agent: kubernetes
+    always_run: false
+    branches:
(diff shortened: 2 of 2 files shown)


ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-main.yaml
```diff
@@ -137,6 +137,16 @@ tests:
...
     workflow: openshift-e2e-aws
+- always_run: false
+  as: e2e-aws-ovn-techpreview-serial
+  optional: true
+  skip_if_only_changed: ^docs/|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|PROJECT|LICENSE)$
+  steps:
+    cluster_profile: aws-2
+    env:
+      FEATURE_SET: TechPreviewNoUpgrade
+      TEST_SUITE: openshift/cluster-version-operator/conformance/serial
```
DavidHurta: At the moment, we make the `openshift/cluster-version-operator/conformance/serial` test suite a child of the `openshift/conformance/serial` test suite [[1](https://github.com/openshift/cluster-version-operator/blob/e444b8e7400c3910440f574bbd9368a2f60a0a0f/cmd/cluster-version-operator-tests/main.go#L34)].

IIRC, this means that running `openshift/conformance/serial` will run the tests in the `openshift/cluster-version-operator/conformance/serial` suite. Here we are explicitly stating to only execute the CVO specific test suite.

All tests we add to the  `openshift/cluster-version-operator/conformance/serial` suite will run in any job executing the general  `openshift/conformance/serial` suite. My open question is - Why shouldn't we just execute the `openshift/conformance/serial` suite while we are at it? Is there a specific reason to only run the `openshift/cluster-version-operator/conformance/serial` suite in the CVO repository?

ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-main.yaml
```diff
@@ -137,6 +137,16 @@ tests:
     env:
       FEATURE_SET: TechPreviewNoUpgrade
     workflow: openshift-e2e-aws
+- always_run: false
+  as: e2e-aws-ovn-techpreview-serial
```
DavidHurta: If we are to execute the `openshift/cluster-version-operator/conformance/serial` test suite, we may rename the test to something like the following text to make the underlying logic of the job more explicit:

```suggestion
  as: e2e-aws-ovn-techpreview-cvo-serial
```

ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-main.yaml
```diff
@@ -137,6 +137,16 @@ tests:
...
+  optional: true
+  skip_if_only_changed: ^docs/|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|PROJECT|LICENSE)$
+  steps:
+    cluster_profile: aws-2
```
DavidHurta: A majority of our e2e jobs run using the `azure-2` cluster profile. Is there a specific reason to use `aws-2` here?

ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-main.yaml
DavidHurta: However, historically, there was no need to execute `openshift/conformance/serial` in the CVO repository.
There is a chance that we can create a test, which will not break `openshift/cluster-version-operator/conformance/serial`, but will break `openshift/conformance/serial`. When we are creating new tests, they should always respect their boundaries (for example, return the cluster to the previous state if needed), and we should always be mindful of these boundaries.
If we do not need the other conformance serial tests, and it's not worth running them just to be sure that our test logic will not break them, then it's okay to omit them, IMO.

ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-main.yaml
hongkailiu: Just a copy/paste from another test. No specific reason.
Maybe spread the workload among tests and extend the coverage on different clouds?
Ideally there should be a knob that gives DPTP's automation to choose (even randomly at the runtime) if we do not care but that knob does not exist AFAIK.

ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-main.yaml
hongkailiu: This is a good one.
I know why I do not want to run non-CVO tests here: they slow down the merging process which is already LOOOONG.

BUT you give me also a good reason to run with them: our CVO serials will be executed with others in TRT's payload testing. It is better to know early (before merge) if we are going to break their tests.

Dilemma it is!
Maybe safety is more important here than speed. I got enough reverted PR from TRT. Let us run everything.

ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-main.yaml
hongkailiu: This is not relevant as I choose the another path.

hongkailiu:
/pj-rehearse pull-ci-openshift-cluster-version-operator-main-e2e-aws-ovn-techpreview-serial

This one should work now as we run everything.

ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-main.yaml
DavidHurta: > Ideally there should be a knob that gives DPTP's automation to choose (even randomly at the runtime) if we do not care but that knob does not exist AFAIK.

Yep, I agree.

I don't see any documentation regarding what choose preferably, and the release folks seem to be manually rebalancing the cluster profiles as needed:

- https://www.github.com/openshift/release/pull/17985
- https://www.github.com/openshift/release/pull/69922

So I don't mind. Although, if we do not mind, I would suggest using the `agnostic` for the platform in the test name to explicitly indicate that we do not care, as we do with other tests:

```
$ grep -E "(^- |  )as: e2e" ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-main.yaml | sort
- as: e2e-agnostic-operator
  as: e2e-agnostic-operator-devpreview
  as: e2e-agnostic-operator-techpreview
- as: e2e-agnostic-ovn
- as: e2e-agnostic-ovn-upgrade-into-change
  as: e2e-agnostic-ovn-upgrade-into-change-techpreview
- as: e2e-agnostic-ovn-upgrade-out-of-change
  as: e2e-agnostic-ovn-upgrade-out-of-change-techpreview
- as: e2e-aws-ovn-techpreview
  as: e2e-extended-tests
- as: e2e-hypershift
- as: e2e-hypershift-conformance
```

ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-main.yaml
DavidHurta: Works for me, both options sound reasonable. We can always change things if needed :bow:

hongkailiu:
/pj-rehearse pull-ci-openshift-cluster-version-operator-main-e2e-agnostic-operator-techpreview-serial

ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-main.yaml
```diff
@@ -152,6 +152,14 @@ tests:
           cpu: 100m
           memory: 200Mi
     workflow: openshift-e2e-azure
+- always_run: false
+  as: e2e-agnostic-operator-techpreview-serial
```
DavidHurta: We use the `operator` substring in test naming when running the integration tests:

```
$ grep -E "e2e-agnostic-operator" -A 8 ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-main.yaml
- as: e2e-agnostic-operator
  skip_if_only_changed: ^docs/|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|PROJECT|LICENSE)$
  steps:
    cluster_profile: azure-2
    test:
    - as: e2e-test
      commands: make integration-test
      from: src
      resources:
--
  as: e2e-agnostic-operator-techpreview
  steps:
    cluster_profile: azure-2
    env:
    test:
    - as: e2e-test
      commands: make integration-test
      from: src
--
  as: e2e-agnostic-operator-devpreview
  optional: true
  steps:
    cluster_profile: azure-2
    env:
      FEATURE_SET: DevPreviewNoUpgrade
    test:
    - as: e2e-test
      commands: make integration-test

```

Here we are executing the serial test suite. I think it's worth making a distinction between the tests. To keep things somewhat consistent, I would propose something like:

```suggestion
  as: e2e-agnostic-techpreview-serial
```

or 

```suggestion
  as: e2e-agnostic-ovn-techpreview-serial
```

hongkailiu: It is lots of more difficult than I thought.

I go with `e2e-agnostic-ovn-techpreview-serial`.
I do not know  why `ovn` is there but it is consistent with other tests.

PR #73963, opened 2026-01-26
OSASINFRA-4098: openstack: Remove mecha-az0

stephenfin:
This host is underutilized and the jobs are not important enough to justify keeping it separate rather than folding it into the existing managed cloud.

- **Remove launch-kuryr cluster-bot cluster**
- **openstack: Remove mecha az0 + associated jobs**


ci-operator/config/openshift/cluster-network-operator/openshift-cluster-network-operator-release-4.14.yaml
tssurya, line 1: Is it that CNO doesn't have `e2e-openstack-kuryr` on later versions post 4.14? 
Just wondering why the removal stopped at 4.14 and if by doing this we are loosing coverage for something we should be testing

stephenfin: @tssurya Yeah, [it was removed in 4.15](https://docs.redhat.com/en/documentation/openshift_container_platform/4.15/html/release_notes/ocp-4-15-release-notes#networking-deprecated-and-removed-features). While we lose coverage on 4.12 and 4.14, both of those are in EUS stage now and we would need a very serious issue to justify any changes to kuryr (serious enough to also warrant spinning up _internal_ test capacity)

PR #73993, opened 2026-01-27
ART-14327: Add verify-deps to assisted-service 4.12-4.18

adobes1:
Add `verify-deps` job to `openshift/assisted-service` 4.12 - 4.18 to catch issues with incomplete `vendor/` early before it has a chance to cause issues for hermetic builds


danmanor (commented):
I have 2 questions / notes:

1. Why only 4.12-4.18 ?
2. We have a job that actually does that and more but for some reason we don't run it in these branches, we should do it instead

PR #74002, opened 2026-01-27
MPIIT: Remove call for non exists function

oharan2:
Update:
The `yq` installation is now handled in `cnv-ci`
https://github.com/openshift-cnv/cnv-ci/blob/master/hack/deploy-cnv.sh#L11

Removed old call for revoked function

### DIFF
--- a/ci-operator/step-registry/interop-tests/cnv-tests-e2e-deploy/interop-tests-cnv-tests-e2e-deploy-commands.sh
+++ b/ci-operator/step-registry/interop-tests/cnv-tests-e2e-deploy/interop-tests-cnv-tests-e2e-deploy-commands.sh
@@ -10,7 +10,7 @@ function mapTestsForComponentReadiness() {
         results_file="${1}"
         echo "Patching Tests Result File: ${results_file}"
         if [ -f "${results_file}" ]; then
-            install_yq_if_not_exists
+            # yq is installed within the container of openshift-cnv/cnv-ci/
             echo "Mapping Test Suite Name To: CNV-lp-interop"
             yq eval -px -ox -iI0 '.testsuite."+@name" = "CNV-lp-interop"' $results_file
         fi


ci-operator/step-registry/interop-tests/cnv-tests-e2e-deploy/interop-tests-cnv-tests-e2e-deploy-commands.sh
dfrazzette, line 7: @oharan2 - Is there value to keep or create the `install_yq_if_not_exists` function to be consistent with other cnv steps?
```suggestion
function install_yq_if_not_exists() {
    # Install yq manually if not found in image
    echo "Checking if yq exists"
    cmd_yq="$(yq --version 2>/dev/null || true)"
    if [ -n "$cmd_yq" ]; then
        echo "yq version: $cmd_yq"
    else
        echo "Installing yq"
        mkdir -p /tmp/bin
        export PATH=$PATH:/tmp/bin/
        curl -L "https://github.com/mikefarah/yq/releases/latest/download/yq_linux_$(uname -m | sed 's/aarch64/arm64/;s/x86_64/amd64/')" \
         -o /tmp/bin/yq && chmod +x /tmp/bin/yq
    fi
}
```

rabin-io: I don't mind, keep it. Or alternative way is to extract into a file or gist and have it referenced.
Some day we'll have `yq` by default on UBI and RHEL images, like with `jq` :)

oharan2: @dfrazzette we're actively discussing about getting `yq` into the image so that it'll apply for all CNV's step containers.

PR #74016, opened 2026-01-27
Adding network policies ci jobs

gangwgr:
Adding network policies ci jobs

### DIFF
--- a/ci-operator/config/openshift/cluster-authentication-operator/openshift-cluster-authentication-operator-master.yaml
+++ b/ci-operator/config/openshift/cluster-authentication-operator/openshift-cluster-authentication-operator-master.yaml
@@ -109,21 +109,12 @@ tests:
           cpu: 100m
       timeout: 4h0m0s
     workflow: ipi-gcp
-- as: e2e-operator-network-policies-parallel
-  optional: true
+- always_run: true
+  as: e2e-gcp-operator-disruptive
   steps:
     cluster_profile: gcp
     env:
-      TEST_SUITE: openshift/conformance/parallel
-    test:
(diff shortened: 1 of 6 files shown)


ci-operator/config/openshift/cluster-authentication-operator/openshift-cluster-authentication-operator-master.yaml
```diff
@@ -124,6 +124,22 @@ tests:
...
+    - ref: openshift-e2e-test
+    workflow: ipi-gcp
+- as: e2e-operator-network-policies-serial
+  steps:
+    cluster_profile: gcp
+    env:
+      TEST_SUITE: openshift/conformance/serial
+    test:
+    - ref: openshift-e2e-test
+    workflow: ipi-gcp
```
liouk: This configuration generates jobs that will always be run; my suggestion which we also discussed offline would be to add these presubmits only to be manually triggered. 

Also, IMHO the job names are misleading; these jobs might be used for network policy testing, but they're running the conformance suites and might potentially be used for other tests as well.

Here's a suggestion based on the above:
```suggestion
- always_run: false
  optional: true
  as: e2e-operator-conformance-parallel
  steps:
    cluster_profile: gcp
    env:
      TEST_SUITE: openshift/conformance/parallel
    test:
    - ref: openshift-e2e-test
    workflow: ipi-gcp
- always_run: false
  optional: true
  as: e2e-operator-conformance-serial
  steps:
    cluster_profile: gcp
    env:
      TEST_SUITE: openshift/conformance/serial
    test:
    - ref: openshift-e2e-test
    workflow: ipi-gcp
```

PR #74022, opened 2026-01-27
CNTRLPLANE-2602: Add openshift/aws-node-termination-handler

enxebre:


ci-operator/config/openshift/aws-node-termination-handler/openshift-aws-node-termination-handler-main.yaml
```diff
@@ -0,0 +1,36 @@
+build_root:
```
sjenning, line 1: This should be either
```
build_root_image:
  name: release
  namespace: openshift
  tag: rhel-9-release-golang-1.25-openshift-4.22
```
or, even better,
```
build_root:
  from_repository: true
```
then add a `.ci-operator.yaml` to the repo like this https://github.com/openshift/hypershift/blob/main/.ci-operator.yaml

This is the cause of the rehearsal failure.

enxebre: thanks! done https://github.com/openshift/aws-node-termination-handler/pull/2

enxebre:
/pj-rehearse pull-ci-openshift-aws-node-termination-handler-main-unit

PR #74033, opened 2026-01-27
release-controller: Add RHEL  10.2 version in 4.22

ravanelli:
- Add rhel-10.2


liangxia:
> @liangxia do you know if it is something at my side that needs to be fixed?

Looks like you already have it fixed in https://github.com/openshift/release/pull/74146

PR #74050, opened 2026-01-28
Update networking-console-plugin CI to use npm instead of yarn

upalatucci:
This PR updates the CI configuration for `openshift/networking-console-plugin` to use npm commands instead of yarn.

## Changes
- Replace `yarn install` with `npm ci` (more reliable for CI environments)
- Replace `yarn run lint` with `npm run lint`
- Replace `yarn build` with `npm run build`
- Replace `yarn i18n` with `npm run i18n`

This change standardizes the package manager used in CI and aligns with npm best practices for CI/CD pipelines.

### DIFF
--- a/ci-operator/config/openshift/networking-console-plugin/openshift-networking-console-plugin-main.yaml
+++ b/ci-operator/config/openshift/networking-console-plugin/openshift-networking-console-plugin-main.yaml
@@ -48,7 +48,7 @@ tests:
   steps:
     test:
     - as: lint
-      commands: yarn install && yarn run lint && yarn build
+      commands: npm ci && npm run lint && npm run build
       from: src
       resources:
         requests:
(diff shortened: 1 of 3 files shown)


ci-operator/config/openshift/networking-console-plugin/openshift-networking-console-plugin-main.yaml
```diff
@@ -48,15 +48,15 @@ tests:
   steps:
     test:
     - as: lint
-      commands: yarn install && yarn run lint && yarn build
+      commands: npm ci && npm run lint && npm run build
```
rszwajko: We can use `--ignore-scripts` here (and in other places). It's a security best-practice to disable unwanted post-install scripts.

upalatucci: we can't because `esbuild` uses a postbuild script https://github.com/evanw/esbuild/issues/4288

upalatucci: @rszwajko

PR #74066, opened 2026-01-28
Add YAML indentation validation with auto-fix, git-diff support, enhanced error messages, and pre-commit hook

deepsm007:
After PRs #73531 and #73537 had indentation issues that weren't caught until after merge, this adds a validation script to catch them earlier.

The script checks for:
- Indentation issues (tabs, mixed spacing, trailing whitespace)
- YAML syntax errors
- Common field name typos (like presubmit vs presubmits)
- Wrong value types (string "true" instead of boolean true)
- Includes a pre-commit hook that runs automatically, and a presubmit job that validates changed YAML files in CI. You can also run make check-yaml-indentation locally or use --fix to auto-fix indentation issues

/cc @openshift/test-platform

### DIFF
--- a/ci-operator/jobs/openshift/release/openshift-release-master-presubmits.yaml
+++ b/ci-operator/jobs/openshift/release/openshift-release-master-presubmits.yaml
@@ -2017,30 +2017,37 @@ presubmits:
             cpu: 10m
     trigger: (?m)^/test( | .* )sync-rover-groups,?($|\s.*)
   - agent: kubernetes
-    always_run: true
+    always_run: false
     branches:
     - ^master$
     - ^master-
-    cluster: build04
+    cluster: build01
     context: ci/prow/yamllint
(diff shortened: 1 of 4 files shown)


ci-operator/jobs/openshift/release/openshift-release-master-presubmits.yaml
```diff
@@ -1475,6 +1475,36 @@ presubmits:
...
+      - args:
+        - ./
+        command:
+        - /bin/bash
+        - -c
+        - |
+          export GIT_DIFF=1
+          hack/validate-yaml-indentation.sh .
+        image: quay.io/openshift/ci-public:ci_python-validation_latest
+        imagePullPolicy: Always
```
Prucek: I think it is not necessary to have the `imagePullPolicy: Always`, or?

PR #74069, opened 2026-01-28
NO-ISSUE: Align IBIO OCP versions

danmanor:
This PR also adds me to the OWNERS of the IBIO tests.

### DIFF
--- a/ci-operator/config/openshift/image-based-install-operator/openshift-image-based-install-operator-backplane-2.10.yaml
+++ b/ci-operator/config/openshift/image-based-install-operator/openshift-image-based-install-operator-backplane-2.10.yaml
@@ -12,7 +12,7 @@ base_images:
     namespace: edge-infrastructure
     tag: latest
   installer:
-    name: "4.19"
+    name: "4.20"
     namespace: ocp
     tag: installer
   ocp_4.20_base-rhel9:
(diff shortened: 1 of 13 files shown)


ci-operator/config/openshift/image-based-install-operator/openshift-image-based-install-operator-backplane-2.10.yaml
giladravid16: This should also be 4.20

danmanor: great thanks

ci-operator/config/openshift/image-based-install-operator/openshift-image-based-install-operator-main.yaml
```diff
@@ -31,12 +31,16 @@ images:
       as:
       - registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.24-openshift-4.20
   to: image-based-install-operator
+promotion:
+  to:
+  - name: "2.12"
```
carbonin, line 36: Do we usually promote from the main branch? I would have figured we'd only do this in the 2.12 branch.

danmanor: I am adding IBIO tests to `ib-orchestrate-vm`, `bip-orchestrate-vm` and `recert` as they have affect on the job, and we need to test it when we make changes there. Therefore I need to import the IBIO image so i need it promoted

PR #74071, opened 2026-01-28
ARO-23200 | Migrate the Bicep lint workflow to Prow

oriAdler:
Add a new job 'bicep-lint'.

https://issues.redhat.com/browse/ARO-23200

### DIFF
--- a/ci-operator/step-registry/aro-hcp/bicep-lint/aro-hcp-bicep-lint-commands.sh
+++ b/ci-operator/step-registry/aro-hcp/bicep-lint/aro-hcp-bicep-lint-commands.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+set -o errexit
+set -o nounset
+set -o pipefail
+set -o xtrace
+
+az config set bicep.use_binary_from_path=true
+az bicep version
+
+# Run lint from dev-infrastructure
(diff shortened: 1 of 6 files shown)


ci-operator/step-registry/aro-hcp/bicep-lint/aro-hcp-bicep-lint-commands.sh
```diff
@@ -0,0 +1,16 @@
...
+az config set bicep.use_binary_from_path=true
+az bicep version
+
+# Run lint from dev-infrastructure
+cd dev-infrastructure
+make fmt
+make lint
+
+# Check for uncommitted changes
+git diff --exit-code -- '*.bicep' || (echo "Uncommitted changes detected in bicep templates" && exit 1)
```
stevekuznetsov: Do you also need `.bicepparam`?

oriAdler: Good catch!
I've changed the pattern to match the one in the Github workflow https://github.com/Azure/ARO-HCP/blob/main/.github/workflows/bicep-lint.yml#L46
Thanks

PR #74084, opened 2026-01-28
Gather EndpointSlices rather than Endpoints

danwinship:
The Endpoints API is deprecated (https://kubernetes.io/blog/2025/04/24/endpoints-deprecation/). We should be gathering EndpointSlices, not Endpoints. (The EndpointSlices have all of the information that the Endpoints do, plus information about terminating endpoints, topology, and dual-stack endpoints.)


danwinship:
/pj-rehearse pull-ci-openshift-cloud-provider-gcp-main-e2e-gcp-ovn

Miciah:
The router and CoreDNS migrated to EndpointSlice in OpenShift 4.6 and 4.8, respectively (see https://github.com/openshift/router/pull/154 and https://github.com/openshift/coredns/pull/52; internally, the router converts endpointslices into endpoints objects, which can be confusing, but it is consuming endpointslices).  

@alebedev87, are aws-load-balancer-controller and external-dns using the Endpoints API?  It looks like the operands support endpointslices, but external-dns-operator configures RBAC only for endpoints, not endpointslices: https://github.com/openshift/external-dns-operator/blob/4abf935dbaf067db15b8418ca9b0ac7541cb256d/bundle/manifests/external-dns_rbac.authorization.k8s.io_v1_clusterrole.yaml#L18

rikatz:
/lgtm
/approve
The new PR now is collecting endpointslices additionally to endpoints, not excluding one in favor of the other.
Thanks!

PR #74086, opened 2026-01-28
HIVE-3059: Add .config.prowgen for slack alerts

suhanime:
Previously we had slack alerts configured for failing periodic jobs. This configuration was overwritten by a recent change. Since reporter_config is present only in the periodic yaml under hive jobs, it is very easy to lose the setting if any change triggers a `make jobs`. To ensure these alert configs are preserved, we now have a new way via a .config.prowgen file to be placed under the config subdirectory.

This commit essentially preserves the same slack alert we had for the same jobs, with an additional reporting for aborted jobs - and some minor formatting of the failure message.

### DIFF
--- a/ci-operator/config/openshift/hive/.config.prowgen
+++ b/ci-operator/config/openshift/hive/.config.prowgen
@@ -5,8 +5,14 @@ slack_reporter:
   - error
   - aborted
   report_template: 'Job *{{.Spec.Job}}* ended with *{{.Status.State}}*. <{{.Status.URL}}|View logs>'
+  job_name_patterns: ["e2e-.*-weekly"]
+  excluded_job_patterns: [".*openstack.*"]
+- channel: "#team-hive-alert"
+  job_states_to_report:
+  - failure
+  - error
+  - aborted
+  - success
+  report_template: 'Job *{{.Spec.Job}}* ended with *{{.Status.State}}*. <{{.Status.URL}}|View logs>'
   job_names: 
-  - e2e-azure-weekly
-  - e2e-gcp-weekly
-  - e2e-pool-weekly
-  - e2e-weekly
+  - e2e-openstack-weekly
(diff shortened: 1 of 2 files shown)


suhanime:
/pj-rehearse pull-ci-openshift-hive-master-periodic-images periodic-ci-openshift-hive-master-periodic-e2e-gcp-weekly periodic-ci-openshift-hive-master-periodic-e2e-weekly periodic-ci-openshift-hive-master-periodic-e2e-azure-weekly

pj rehearse seems to be having issues. Manually running periodic tests, skipping e2e-pool because it is similar enough to others and it had a recent successful run

ci-operator/config/openshift/hive/.config.prowgen
```diff
@@ -0,0 +1,18 @@
+slack_reporter:
+- channel: "#team-hive-alert"
+  job_states_to_report:
+  - failure
+  - error
+  - aborted
+  report_template: 'Job *{{.Spec.Job}}* ended with *{{.Status.State}}*. <{{.Status.URL}}|View logs>'
+  job_name_patterns: ["e2e-.*-weekly"]
+  excluded_job_patterns: [".*openstack.*"]
```
2uasimojo, line 9: Guess we could have done this simply by making the openstack stanza use only `success` in `job_states_to_report`. But whatevs :)

2uasimojo: Ah, I'm told the tooling isn't smart enough to union config stanzas that way, and you're not allowed to have multiple slack reporters for the same job.

suhanime: A job_name can only be present for one of the slack_reporter entries unfortunately, as per [docs](https://docs.ci.openshift.org/docs/how-tos/notification/#setting-up-slack-notifications-with-prowgen)

PR #74093, opened 2026-01-28
Filter out AI zones for standard machine types and randomize zone selection

deepsm007:
CI jobs are experiencing `ZONE_RESOURCE_POOL_EXHAUSTED_WITH_DETAILS` errors when creating control plane nodes in AI zones like `us-central1-ai1a`. This occurs because:

1. **Control plane nodes use standard machine types** (e.g., `e2-custom-6-16384`) but are being placed in AI zones
2. **AI zones are optimized for GPU/AI machine types** (e.g., `a2-highgpu-1g`) and have limited capacity for standard machine types
3. **When `CONTROL_PLANE_ZONES` is not set**, zones aren't populated in the install-config, causing the installer to auto-select zones which can include AI zones

/cc @patrickdillon


deepsm007:
This is a temp PR until https://github.com/openshift/installer/pull/10269 merges

PR #74094, opened 2026-01-28
imagebuilder: run unit tests with Go 1.23 instead of 1.21

nalind:
Run our unit tests with a not-archived-yet version of Go.  Version 1.26 is imminent, but for now, 1.23 is what we're about to start using when running conformance tests.


TomSweeneyRedHat:
LGTM

PR #74113, opened 2026-01-29
Hot fix removing unnecessary trace

xingxingxia:
@liangxia , could you PTAL?

Also found other files that might need to be handled as well. Checking and fixing other files may need separately more time. But others are not as serious as current fixed shell script file. So raising this PR first. https://url.corp.redhat.com/932f0f0 (RH internal) is an example that traced the unnecessary info to be hidden by the PR.  CC @chaitanya1731


xingxingxia:
@psalajova , thank you!
/pj-rehearse ack
The shell script update is simple and tested in my local host. So adding rehearsals ack directly

PR #74135, opened 2026-01-29
chore(fix): Change OCP version used by Orion for ACS scale tests

mtodor:
This PR is changing the OCP version used by Orion when it fetches data from OpenSearch.

We have changed to use 4.20 for running scale tests. Orion config has to be adjusted to fetch the correct data.

**Note:** This should be merged around 2026-02-04. Because at that point, we will have enough collected data for 4.20.

/cc @davdhacs


mtodor:

PR #74186, opened 2026-01-30
add back 4.19 metal jobs

mohit-sheth:
We need to keep the metal/virt one because unfortunately due to a bug we cannot run it still on OCP4.20/4.21


ci-operator/config/openshift-eng/ocp-qe-perfscale-ci/openshift-eng-ocp-qe-perfscale-ci-main__metal-4.19-nightly-multi.yaml
```diff
@@ -0,0 +1,83 @@
+base_images:
```
josecastillolema: This file you can delete, it was inherited from QE

PR #73587, opened 2026-01-15
ovn-kubernetes: Tweak config for OVN large/xlarge jobs.

dceara:
- Use larger service CIDRs for xlarge jobs.
- Disable pprof collection for large/xlarge jobs.
- Disable GC and GC metrics for large/xlarge jobs.
- Bump node-density-cni step timeout to 8h.

NOTES:
- most of this is just a rebase of #63340
- the ovncore-perfscale-aws-ovn-xlarge-node-density-cni is currently failing due to https://github.com/kube-burner/kube-burner/issues/1080; once that's fixed the job should succeed.

PR #74195, opened 2026-01-30
Rename EXTRA_FLAGS to ND_CNI_EXTRA_FLAGS

rsevilla87:
Since the EXTRA_FLAGS variable is used by other jobs such as cd-v2, we need to use a dedicated variable for each workload — in this case, for node-density-cni.

The changes introduced in `ci-operator/step-registry/openshift-qe/node-density-cni/openshift-qe-node-density-cni-commands.sh` by https://github.com/openshift/release/pull/73587/changes#diff-bc5fca558cf1d5f5e2ed4ccd1a977aa6e19c832bb0aa7b07bb717b8f6dcc8db3

Now the EXTRA_FLAGS variable is appended rather than declared.

```diff
export EXTRA_FLAGS="--gc-metrics=true --pods-per-node=$PODS_PER_NODE --namespaced-iterations=$NAMESPACED_ITERATIONS --iterations-per-namespace=$ITERATIONS_PER_NAMESPACE --profile-type=${PROFILE_TYPE} --pprof=${PPROF}"
EXTRA_FLAGS+=" --gc=$GC --gc-metrics=$GC_METRICS --pods-per-node=$PODS_PER_NODE --namespaced-iterations=$NAMESPACED_ITERATIONS --iterations-per-namespace=$ITERATIONS_PER_NAMESPACE --profile-type=${PROFILE_TYPE} --pprof=${PPROF}"
```

This is causing issues because other workloads such as cd-v2, run within the same control-plane-chain and use the same EXTRA_FLAGS variable
https://github.com/openshift/release/blob/master/ci-operator/step-registry/openshift-qe/control-plane/openshift-qe-control-plane-chain.yaml#L5-L6

https://github.com/openshift/release/blob/master/ci-operator/step-registry/openshift-qe/cluster-density-v2/openshift-qe-cluster-density-v2-ref.yaml#L17-L20

As a result, we are unintentionally running node-density-cni with churning enabled.

### DIFF
--- a/ci-operator/config/openshift-priv/ovn-kubernetes/openshift-priv-ovn-kubernetes-master.yaml
+++ b/ci-operator/config/openshift-priv/ovn-kubernetes/openshift-priv-ovn-kubernetes-master.yaml
@@ -633,9 +633,9 @@ tests:
       BASE_DOMAIN: core.ovn.devcluster.openshift.com
       COMPUTE_NODE_TYPE: m5.2xlarge
       CONTROL_PLANE_INSTANCE_TYPE: r5.8xlarge
-      ND_CNI_EXTRA_FLAGS: --churn-duration=20m --service-latency
       GC: "false"
       GC_METRICS: "false"
+      ND_CNI_EXTRA_FLAGS: --churn-duration=20m --service-latency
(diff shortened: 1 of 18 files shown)


tssurya:
no rehearsals seem to be run here?... how do we know this change works? I can't see a link to a succeeded job run

PR #74202, opened 2026-01-30
Fix condition failure in hypershift-hostedcluster-destroy-hostedcluster

xingxingxia:
Forgot `set -e` can make `[ -f xxx ] && xxxx` fail if [ -f xxx ] is false.


xingxingxia:
The change is very simple by adding `|| true`. So adding ack directly:
/pj-rehearse ack
@pruan-rht could you help merge? CC @jlost

PR #74218, opened 2026-02-01
ARO-23197 | Migrate api.yml to a Prow job

oriAdler:
Create a new light image with the dependencies for the job.

### DIFF
--- a/ci-operator/step-registry/aro-hcp/api-validation/aro-hcp-api-validation-commands.sh
+++ b/ci-operator/step-registry/aro-hcp/api-validation/aro-hcp-api-validation-commands.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+set -o errexit
+set -o nounset
+set -o pipefail
+set -o xtrace
+
+# Verify Node.js is available
+node --version
+npm --version
+
+# Unset VERSION to prevent CI environment variable from overriding Makefile default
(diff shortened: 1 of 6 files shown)


ci-operator/step-registry/aro-hcp/api-validation/aro-hcp-api-validation-commands.sh
```diff
@@ -0,0 +1,34 @@
...
+set -o nounset
+set -o pipefail
+set -o xtrace
+
+# Verify Node.js is available
+node --version
+npm --version
+
+# Unset VERSION to prevent CI environment variable from overriding Makefile default
+# The Makefile uses VERSION ?= v20251223preview, but CI may have VERSION set to something else
```
oriAdler, line 12: Reference https://github.com/Azure/ARO-HCP/blob/main/api/Makefile#L6
Another approach would be to change the VERSION to API_VERSION in the ARO-HCP Makefile.

roivaz: I assume VERSION contains the Openshift version, which we don't use.

oriAdler:
@raelga @roivaz thank you for the review, I rebased the PR, and am ready for another cycle of reviews.
As for the long list of tests affected by this PR, IIUC, it is caused by changes in **ci-operator/config/Azure/ARO-HCP/Azure-ARO-HCP-main.yaml**.

PR #74232, opened 2026-02-02
NO-JIRA: feat: increase the interval of tnf runs

eggfoobar:
Since we are getting close to promotion, we need to increase the number of runs to gather data for promotion

This will run the jobs twice a day with at least 8 hour gap between runs, cron's are randomized to avoid spikes.


eggfoobar:
/pj-rehearse ack

No runs needed, just a cron update

PR #74286, opened 2026-02-03
Triggering Jobs: Complete overhaul on the framework.

etirta:
## Related To
- [INTEROP-8526](https://issues.redhat.com/browse/INTEROP-8526): [GS CNV] Update weekly_trigger file job and watcher bot configs to include CNV GS BM job

## Description
The Interop team uses the `trigger-jobs` Step to trigger jobs. The current algorithm in this Step uses Job list name patterns, which is not flexible and may require very long Job list names when multiple scheduling conditions are needed.

Name-based scheduling requires multiple lists for variants of the same test suite, e.g., `fips` vs. non-`fips`, because each list must have a distinct name. If we have N groups of tests with `fips` variants, we will need 2N lists.

The algorithm is also not easy to extend. Each `if` block must evaluate several logical expressions. Whenever a new condition is needed, not only must a new `if` block be added, but new logical expressions must also be added to existing `if` blocks.

Currently there are 2 `if` blocks, each evaluating 3 logical expressions. Adding this new scheduling requirement for GS BM Cluster increases the number of logical expressions from 3x2 to 4x3. If another condition is required, it will become 5x4, and so on.

This O(N^2) complexity is hard to maintain. A new approach is needed to achieve O(1) complexity.

There are other Steps, such as
`trigger-jobs-interop-ocp-watcher-bot-send-message`, that largely duplicate the scheduling algorithm. This non-modular approach is problematic when multiple Steps with the same logic is needed.

Changes in `trigger-jobs` Step:
  - Introduces a new data input to explicitly configure the desired scheduling conditions. Removes dependency on Job list name.
  - Decouples the trigger condition check from the `trigger-jobs` Step, making it modular and reusable. Trigger condition Steps are executed as pre-Steps.
  - Introduces conditional post-tasks, which are also decoupled from the `trigger-jobs` Step, allowing scheduling algorithm reuse for different tasks such as configuring the watcher bot. Post-task Steps are executed as post-Steps.

Advantages of the new algorithm:
  - Configuration-driven: No code changes needed when using existing scheduling algorithms.
  - O(1) code changes when adding new scheduling condition algorithms (vs. O(N) changes in old approach).
  - Eliminates the wall of `if` blocks that grew with each different scheduling condition.
  - Modular, reusable, and configurable.
  - Reduces the number of required Job lists, e.g., enabling use of a single Job list name for both `fips` and non-`fips` variants.

For backward compatibility with current Jobs with old triggering Job list, some legacy code is kept (with update) and some backward compatible `jq` code to convert the old format to new format in this PR. Once all old triggering Jobs are updated to onboard with this new framework, those codes should be removed.


etirta:
/pj-rehearse periodic-ci-rhpit-interop-tests-main-weekly_trigger-bm-gs--lp-interop periodic-ci-rhpit-interop-tests-main-weekly_trigger-bm-gs--lp-interop--watcher-bot

etirta:
/unhold

Bug if fixed.
Testing done:
  - [test--07-trigger-jobs-post-interop-ocp-watcher-bot-send-message--backward-comp.log](https://github.com/user-attachments/files/25056934/test--07-trigger-jobs-post-interop-ocp-watcher-bot-send-message--backward-comp.log):
      - Testing execution decision when post task flag is set.
      - Testing execution decision when post task flag is not set.
  - [test--08-trigger-jobs-post-interop-ocp-watcher-bot-send-message--new.log](https://github.com/user-attachments/files/25056936/test--08-trigger-jobs-post-interop-ocp-watcher-bot-send-message--new.log):
      - Testing execution decision when post task flag is set.
      - Testing execution decision when post task flag is not set.

etirta:
/pj-rehearse periodic-ci-rhpit-interop-tests-main-weekly_trigger-rosa-sts-hypershift-layered-product-interop-416 periodic-ci-rhpit-interop-tests-main-weekly_trigger-rosa-sts-hypershift-lp-watcher-bot-message-416 periodic-ci-rhpit-interop-tests-main-weekly_trigger-rosa-classic-sts-layered-product-interop-416 periodic-ci-rhpit-interop-tests-main-weekly_trigger-rosa-classic-sts-lp-watcher-bot-message-416

ci-operator/config/rhpit/interop-tests/rhpit-interop-tests-main__weekly_trigger.yaml
```diff
@@ -30,131 +32,171 @@ tests:
...
+      JT__SKIP_TRIG_MAIN_JOBS: "1"
+      JT__TRIG_COND_EXEC_FLGS: "1"
+      JT__TRIG_JOB_LIST: opp-interop-jobs
+    post:
+    - ref: trigger-jobs-post-interop-ocp-watcher-bot-send-message
     test:
-    - ref: trigger-jobs-interop-ocp-watcher-bot-send-message
-- as: ocp-self-managed-layered-product-interop-420
+    - ref: trigger-jobs
+- as: ocp-self-managed-lp-interop--fips--421
```
oharan2, line 44:
```suggestion
- as: ocp-self-managed-lp-interop--fips-421
```

ci-operator/config/rhpit/interop-tests/rhpit-interop-tests-main__weekly_trigger.yaml
```diff
@@ -30,131 +32,171 @@ tests:
...
       FIREWATCH_CONFIG: |
         {
-          "failure_rules": [{"step": "trigger-jobs", "failure_type": "all", "classification": "self-managed Trigger Failure", "jira_additional_labels": ["4.20-lp","self-managed-lp","weekly-trigger-lp"]}]
+          "failure_rules": [{"step": "trigger-jobs", "failure_type": "all", "classification": "self-managed Trigger Failure", "jira_additional_labels": ["4.21-lp","self-managed-lp","fips","weekly-trigger-lp"]}]
         }
       FIREWATCH_DEFAULT_JIRA_PROJECT: LPINTEROP
-      JSON_TRIGGER_LIST: self-managed-lp-interop-jobs-420
+      JT__POST_TASK_EXEC_FLGS: "0"
+      JT__TRIG_COND_EXEC_FLGS: "1"
+      JT__TRIG_JOB_LIST: self-managed-lp-interop--fips--421
```
oharan2, line 56:
```suggestion
      JT__TRIG_JOB_LIST: self-managed-lp-interop--fips-421
```

ci-operator/config/rhpit/interop-tests/rhpit-interop-tests-main__weekly_trigger.yaml
```diff
@@ -30,131 +32,171 @@ tests:
...
     post:
     - ref: firewatch-report-issues
     test:
     - ref: trigger-jobs
-- as: ocp-self-managed-lp-watcher-bot-message-420
+- as: ocp-self-managed-lp-interop--fips--watcher-bot--421
```
oharan2, line 61:
```suggestion
- as: ocp-self-managed-lp-interop--fips--watcher-bot-421
```

ci-operator/config/rhpit/interop-tests/rhpit-interop-tests-main__weekly_trigger.yaml
```diff
@@ -30,131 +32,171 @@ tests:
...
   steps:
     cluster_profile: aws-cspi-qe
     env:
-      WATCHER_BOT_JOB_FILE: self-managed-lp-interop-jobs-420
-      WATCHER_BOT_JOB_GROUP_NAME: OCP Self Managed Layered Product Scenarios
+      JT__POST__WB__JOB_GROUP_NAME: OCP Self Managed Layered Product Scenarios - FIPS
+      JT__POST_TASK_EXEC_FLGS: "1"
+      JT__SKIP_TRIG_MAIN_JOBS: "1"
+      JT__TRIG_COND_EXEC_FLGS: "1"
+      JT__TRIG_JOB_LIST: self-managed-lp-interop--fips--421
```
oharan2, line 70:
```suggestion
      JT__TRIG_JOB_LIST: self-managed-lp-interop--fips-421
```

ci-operator/config/rhpit/interop-tests/rhpit-interop-tests-main__weekly_trigger.yaml
```diff
@@ -30,131 +32,171 @@ tests:
...
+      JT__SKIP_TRIG_MAIN_JOBS: "1"
+      JT__TRIG_COND_EXEC_FLGS: "1"
+      JT__TRIG_JOB_LIST: rosa-classic-sts-lp-interop-jobs
+    post:
+    - ref: trigger-jobs-post-interop-ocp-watcher-bot-send-message
     test:
-    - ref: trigger-jobs-interop-ocp-watcher-bot-send-message
-- as: ocp-self-managed-layered-product-interop-zstream-n-1
+    - ref: trigger-jobs
+- as: ocp-self-managed-lp-interop--zstream-n-1
```
oharan2, line 199:
```suggestion
- as: ocp-self-managed-lp-interop-zstream-n-1
```

ci-operator/config/rhpit/interop-tests/rhpit-interop-tests-main__weekly_trigger.yaml
```diff
@@ -164,12 +206,14 @@ tests:
...
-      JSON_TRIGGER_LIST: self-managed-lp-interop-jobs-zstream-n-1
+      JT__POST_TASK_EXEC_FLGS: "0"
+      JT__TRIG_COND_EXEC_FLGS: "1"
+      JT__TRIG_JOB_LIST: self-managed-lp-interop--zstream-n-1
     post:
     - ref: firewatch-report-issues
     test:
     - ref: trigger-jobs
-- as: ocp-self-managed-layered-product-interop-zstream-n-2
+- as: ocp-self-managed-lp-interop--zstream-n-2
```
oharan2, line 216:
```suggestion
- as: ocp-self-managed-lp-interop-zstream-n-2
```

ci-operator/config/rhpit/interop-tests/rhpit-interop-tests-main__weekly_trigger.yaml
```diff
@@ -164,12 +206,14 @@ tests:
           "failure_rules": [{"step": "trigger-jobs", "failure_type": "all", "classification": "self-managed Trigger Failure", "jira_additional_labels": ["self-managed-lp","weekly-trigger-lp"]}]
         }
       FIREWATCH_DEFAULT_JIRA_PROJECT: LPINTEROP
-      JSON_TRIGGER_LIST: self-managed-lp-interop-jobs-zstream-n-1
+      JT__POST_TASK_EXEC_FLGS: "0"
+      JT__TRIG_COND_EXEC_FLGS: "1"
+      JT__TRIG_JOB_LIST: self-managed-lp-interop--zstream-n-1
```
oharan2, line 211:
```suggestion
      JT__TRIG_JOB_LIST: self-managed-lp-interop-zstream-n-1
```

ci-operator/config/rhpit/interop-tests/rhpit-interop-tests-main__weekly_trigger.yaml
```diff
@@ -179,35 +223,48 @@ tests:
           "failure_rules": [{"step": "trigger-jobs", "failure_type": "all", "classification": "self-managed Trigger Failure", "jira_additional_labels": ["self-managed-lp","weekly-trigger-lp"]}]
         }
       FIREWATCH_DEFAULT_JIRA_PROJECT: LPINTEROP
-      JSON_TRIGGER_LIST: self-managed-lp-interop-jobs-zstream-n-2
+      JT__POST_TASK_EXEC_FLGS: "0"
+      JT__TRIG_COND_EXEC_FLGS: "1"
+      JT__TRIG_JOB_LIST: self-managed-lp-interop--zstream-n-2
```
oharan2, line 228:
```suggestion
      JT__TRIG_JOB_LIST: self-managed-lp-interop-zstream-n-2
```

ci-operator/config/rhpit/interop-tests/rhpit-interop-tests-main__weekly_trigger.yaml
```diff
@@ -179,35 +223,48 @@ tests:
...
     post:
     - ref: firewatch-report-issues
     test:
     - ref: trigger-jobs
-- as: rosa-classic-sts-layered-product-interop-416
-  cron: 0 23 31 2 *
+- as: bm-gs--lp-interop
```
oharan2, line 233:
```suggestion
- as: bm-gs-lp-interop
```

ci-operator/config/rhpit/interop-tests/rhpit-interop-tests-main__weekly_trigger.yaml
```diff
@@ -179,35 +223,48 @@ tests:
...
+      JT__TRIG_JOB_LIST: bm-gs--lp-interop
     post:
     - ref: firewatch-report-issues
+    pre:
+    - ref: trigger-jobs-trig-check-resource-owner
     test:
     - ref: trigger-jobs
-- as: rosa-classic-sts-lp-watcher-bot-message-416
-  cron: 0 23 31 2 *
+- as: bm-gs--lp-interop--watcher-bot
```
oharan2, line 252:
```suggestion
- as: bm-gs-lp-interop--watcher-bot
```

ci-operator/config/rhpit/interop-tests/rhpit-interop-tests-main__weekly_trigger.yaml
```diff
@@ -179,35 +223,48 @@ tests:
...
   steps:
     cluster_profile: aws-cspi-qe
     env:
-      WATCHER_BOT_JOB_FILE: rosa-classic-sts-lp-interop-jobs
-      WATCHER_BOT_JOB_GROUP_NAME: ROSA Classic STS Layered Product Scenarios
+      JT__POST__WB__JOB_GROUP_NAME: OCP Goldman Sachs Bare-Metal Tests
+      JT__POST_TASK_EXEC_FLGS: "1"
+      JT__SKIP_TRIG_MAIN_JOBS: "1"
+      JT__TRIG_COND_EXEC_FLGS: "1"
+      JT__TRIG_JOB_LIST: bm-gs--lp-interop
```
oharan2, line 261:
```suggestion
      JT__TRIG_JOB_LIST: bm-gs-lp-interop
```

ci-operator/config/rhpit/interop-tests/rhpit-interop-tests-main__weekly_trigger.yaml
```diff
@@ -179,35 +223,48 @@ tests:
...
       FIREWATCH_CONFIG: |
         {
-          "failure_rules":[{"step": "trigger-jobs", "failure_type": "all", "classification": "Rosa-classic-sts Trigger Failure"}]
+          "failure_rules": [{"step": "trigger-jobs", "failure_type": "all", "classification": "self-managed Trigger Failure", "jira_additional_labels": ["4.19-lp","gs-bm-lp","biweekly-trigger-lp"]}]
         }
       FIREWATCH_DEFAULT_JIRA_PROJECT: LPINTEROP
-      JSON_TRIGGER_LIST: rosa-classic-sts-lp-interop-jobs
+      JT__POST_TASK_EXEC_FLGS: "0"
+      JT__TRIG_COND_EXEC_FLGS: "1"
+      JT__TRIG_JOB_LIST: bm-gs--lp-interop
```
oharan2, line 245:
```suggestion
      JT__TRIG_JOB_LIST: bm-gs-lp-interop
```

ci-operator/step-registry/trigger-jobs/trigger-jobs-commands.sh
```diff
@@ -1,114 +1,251 @@
...
+typeset gangwayAPItoken; gangwayAPItoken="$(cat "${CLUSTER_PROFILE_DIR}/${JT__GW__API_TOKEN}")"
+
+### Legacy code start. To be replaced with custom trigger condition Step.    ###
+# Get the day of the month.
+typeset -i month_day; month_day="$(date +%-d)"
 
 # additional checks for self-managed fips and non-fips testing
 self_managed_string='self-managed-lp-interop-jobs'
 zstream_string='zstream'
 fips_string='fips'
```
tanfengshuang, line 31: @etirta, these 3 vars are not defined with typeset.

ci-operator/step-registry/trigger-jobs/trigger-jobs-commands.sh
```diff
@@ -1,114 +1,251 @@
...
+    # Triggering Jobs.
+    while IFS=$'\t' read -r jobType jobName; do
+        : "Issuing trigger for active job: ${jobName}"
+        jobExecType="${jobExecTypeMaps[${jobType}]:-}"
+        ((jobExecType)) || {
+            : "Invalid \`.jobType\` value: ${jobType}"
+            exit 1
+        }
+        for ((tryLeft=3; tryLeft; tryLeft--)); do
+            rsp="$(eval "$(cat - 0<<cmd1EOF
```
tanfengshuang, line 161: @etirta, a question here. Is rsp needed to define with typeset here?

ci-operator/config/rhpit/interop-tests/rhpit-interop-tests-main__weekly_trigger.yaml
etirta: Actually it's very useful to have `--` to separate component vs words.
The word is separated by `-`, and if we uses the same deliminator to separate component, it is impossible for a script to parse it, if there is a need.

So we must employ different character (or set-of-markers). I personally like `camelCaseComp1-camelCaseComp2`, but since CI operator naming convention is not using camel case, I don't feel like to disrupt too much.

CI Operator itself using double-underscore as variant deliminator on the Job Configuration file name.

So the choice are either single-underscore `_`  or double-dash `--`. The issue with underscore, it's harder to type and it's my cause issue with CI Operator parsing of variant.

We've been using `--` for a while is Chaos team, and so far we never encountered any issue.

Do you have any other character to suggest? We can't use use `-` and make the whole name not parse-able by script in generic manner because we can't distinguish with word separator.

I'm not concern with AI, they should be very good to detect character deviation, much more accurately than human. But reading `abc--foo-bar--v123`, `other-pfx--some-word--234`, is easier for human to understand that `abc`, `other-pfx` are prefix, and `v123`,  `234` are some kind of suffix or version. And more importantly, it's parse-able by a script in generic manner.

This is similar with our Best Practice for Env. Naming convention in a Step should be prefixeed with `PFX__SUB_PFX_1__SUB_PFX_2__ACTUAL_ENV_VAR_NAME`, it's very easy to identify their (sub) prefixes. So in testing, we can easily print the whole group, say `typeset -p ${!PFX__SUB_PFX_1__}` to print all related env. var. that correspond to a functionality associated with `SUB_PFX_1`. Or if necessary we can do efficient clean up via `unset ${!PFX__SUB_PFX_1__}`. As you can see it's easier for reader to understand the author intention of grouping.

I hope this explain the technical reasoning behind `--`. Let me know if you have more question regarding this. Or if there is indeed an issue with CI Operator using this scheme, as so far I haven't encountered it. But there might be issues that I'm not aware of. Thx.

ci-operator/config/rhpit/interop-tests/rhpit-interop-tests-main__weekly_trigger.yaml
etirta: Ditto the above.

ci-operator/config/rhpit/interop-tests/rhpit-interop-tests-main__weekly_trigger.yaml
etirta: Ditto the above.

ci-operator/config/rhpit/interop-tests/rhpit-interop-tests-main__weekly_trigger.yaml
etirta: Ditto the above.

ci-operator/step-registry/trigger-jobs/trigger-jobs-commands.sh
etirta: Yes I'm aware, but those are legacy codes that will be removes as soon as all trigger list has been converted to the new format.

I consciously only do minimal update the legacy code that-to-be-removed. Thx for seeing this though, it shows that you have adopted the Best Practice to your daily workflow 👍.

ci-operator/step-registry/trigger-jobs/trigger-jobs-commands.sh
etirta: Good catch!!! 👏 . As per our Best Practice, we should explicitly declare this. Thank you for spotting this! This also proof that even AI can miss, hence stress the importance of human in the loop.

etirta:
/pj-rehearse ack
Fixing missed explicit declaration of `rsp`. This do not affect the functionality. So no round of rehearsal is required.

ci-operator/step-registry/trigger-jobs/post/interop-ocp-watcher-bot-send-message/trigger-jobs-post-interop-ocp-watcher-bot-send-message-commands.sh
sg-rh, line 45:
```suggestion
    #### Collect triggered jobs for bulk notification ####
```

ci-operator/step-registry/trigger-jobs/trig/check-resource-owner/README.md
sg-rh, line 12:
```suggestion
owner of a resource and creates or deletes an ownership flag file under `${SHARED_DIR}/`, which the `trigger-jobs` Step checks/verifies to decide
whether to trigger specific Jobs.
```

sg-rh, line 28:
```suggestion
    "trigCond": [
      {
        "trigCondFlgs": 1,
        "trigCondStep": "trigger-jobs-trig-check-resource-owner",
        "trigCondPars": {"trigCondFlag": "cluster-owner", "expOwnerName": "Interop Team"}
      },
     {
       "trigCondFlgs": 2,
       "trigCondStep": "trigger-jobs-trig-check-aws-quota",
       "trigCondPars": { "region": "us-east-1", "min_cpu": 64 }
     }
    ],
```
Not needed, but reviewing this I saw we can have additional example for future use, we can add such trigger condition like the AWS Quota is sufficient.

ci-operator/step-registry/trigger-jobs/README.md
sg-rh, line 30: why is  `job_name` replaced with `jobName`

ci-operator/step-registry/trigger-jobs/trigger-jobs-ref.yaml
sg-rh: Is there a default?  And can you simplify or add more description?
Something like this variable acts as a Master switch to enable/disable specific checks (like the Resource Owner check or FIPS check) defined in JSON file.

Also to be more descriptive, maybe update in Readme about JT__TRIG_COND_EXEC_FLGS bitwise usage

sg-rh, line 72: is default value needed

ci-operator/step-registry/trigger-jobs/README.md
sg-rh, line 267:
```suggestion
## Process

The `trigger-jobs` Step follows a modular pipeline to ensure jobs are only triggered when all environment and ownership conditions are met.
```

sg-rh:
```suggestion
 2. **API Health Check (Optional)** Test the Gangway API to verify availability (skipped by default if  `JT__SKIP_HC=true`). When enabled, it will retry every minute for up to one hour before failing.
```

sg-rh:
```suggestion
 3. ** Condition Validation:** 
  - Iterates through `trigCond `entries in the JSON.
  - **Bitwise Filter**: A condition is only evaluated if (trigCondFlgs & JT__TRIG_COND_EXEC_FLGS) is non-zero.
  - **Flag Verification**: Checks for flag files in `${SHARED_DIR}` (created by pre-steps like `Check-resource-owner`). If any required flag is missing, the job group is skipped.

```

sg-rh:
```suggestion
    Note: If `JT__SKIP_TRIG_MAIN_JOBS`  is non-zero, this execution phase is bypassed
```

sg-rh:
```suggestion
 5. Automatically retries failed API calls up to 3 times. All successful triggers and final failures are recorded in the `build-log.txt`.
```

sg-rh, line 89: Somewhere in the readme or script you can explain the usage of execution flags

```suggestion
**How to use Execution Flags:**This system uses bitwise logic to enable specific checks. Think of each power of two as a unique toggle:
**1** = Condition A
**2** = Condition B
**4** = Condition C
To enable both **Condition A and C**, set your environment variable to 5 **($1 + 4$).** The script will then "match" those flags against the JSON and execute only those specific parts of the configuration.
```
or add short comment in script where we have trigger/ post task logics like `((trigCondFlgs & JT__TRIG_COND_EXEC_FLGS)) || continue`

ci-operator/config/rhpit/interop-tests/rhpit-interop-tests-main__weekly_trigger.yaml
sg-rh: If we are going to use `-- `as a delimiter for separate the Component, Variant, and Version, we should update our documentation to be consistent for all jobs

ci-operator/step-registry/trigger-jobs/trigger-jobs-commands.sh
sg-rh:
```suggestion
                    : 'The trigger condition is not met. Skipping this job group and moving to the next entry in JSON.''
```

sg-rh, line 4: as safety check, you can ensures variable is initialized as an integer to avoid errors
```
typeset -i JT__TRIG_COND_EXEC_FLGS=${JT__TRIG_COND_EXEC_FLGS:-0}
typeset -i JT__POST_TASK_EXEC_FLGS=${JT__POST_TASK_EXEC_FLGS:-0}
```

ci-operator/config/rhpit/interop-tests/rhpit-interop-tests-main__weekly_trigger.yaml
etirta: What I meant by component is not the same with OpenShift Component. I meant when a logical grouping in a name. May be I should said "grouping" instead so it doesn't get confused with OpenShift Compoenent. My bad.

For example, when the name: bm-gs--lp-interop--watcher-bot
It has 3 groups in that name, 1st is the target, `bm-gs`, next our BM cluster may be called `bm-other-cust`. The 2nd is the team `lp-interop`, `lp-chaos` is used by Chaos Team. And last is extra comment, so we know this is nudging the `watcher-bot`. This make it easier for reading to understand the grouping. Without it, it's very hard to read a very lengthy Job name with all having 1 level delimiter like we have now. 

Regarding to Test Component, Variant and Version, we can't change it as that is controlled by CI Operator. But inside Variant name, if we want to employ the grouping for consistency, I'm for it, but I wouldn't put it has high priority as we have other thing. What we can do is to update it when we do some work to an LP test.

ci-operator/step-registry/trigger-jobs/post/interop-ocp-watcher-bot-send-message/trigger-jobs-post-interop-ocp-watcher-bot-send-message-commands.sh
etirta: Will do. Except it's not only the triggered Job. It's simply submit the list, they way it has been. My guess the watcher-bot app. check the `active` flags and act accordingly.

@oharan2 please confirm.

ci-operator/step-registry/trigger-jobs/trig/check-resource-owner/README.md
etirta: The `trigger-jobs` is only check for existence, the future plan is the ownership flag will not exist if the current ownership is not the expected owner. I could have just create an empty file, but I put the name of the owner (only if the expected owner is owning it at the time of check) may be for subsequent post task may need it.

So the content is more for potential future post task, if ever needed, but not for the main `trigger-jobs` script.

etirta: No, this is README for `check-resource-owner` Step. For future Step, they will have their own README.

But you got the right idea 👍.

ci-operator/step-registry/trigger-jobs/README.md
etirta: So I can detect old format vs new, in order to provide backward compatibility.

etirta: Sorry, that's too specific. Not all triggering conditions perform ownership check. The process need to describe a high level workflow of this framework in generic manner.

It is possible to just have a list that will always triggered, hence in our current use it will just run weekly, no matter what.

etirta: I prefer to have the same style, like number 3, where the front statment only say `Optional` or `If applicable`, etc.

Will update to:
```
 2. **(Optional)** Perform Gangway API Health Check to verify availability (skipped by default via `JT__SKIP_HC=true`). When enabled, checks every minute for one hour, before failing the Step.
```

etirta: I think it's too detail to implementation if we describe about the flag. It's already explained in several places (I updating them to clearly states that this is bitwise flag / masking) how the trigger condition flag is expected to work. Also it's the common suggested approach on how to use the flag. I made it very flexible, that if needed we can have a special custom trigger condition Step and uses a totally different way for indication. So stick with generic statement, e.g. `trigger condition is met or not`,

But I like the statement:  If any required flag is missing, the job group is skipped.

I'm updating it to:
```
 3. **(If applicable)** Check trigger conditions written by pre-Steps (e.g., resource ownership). Trigger conditions are only checked when
    the bitwise mask `(trigCondFlgs & JT__TRIG_COND_EXEC_FLGS)` evaluates to non-0. If any checked trigger condition is not met, the rest
    of the processing is skipped.
```

etirta: Will update to (because I don't what to use a new terminology, `execution phase` that is not introduced earlier):
```
 4. **(If applicable)** Loop through the jobs-to-trigger JSON and trigger each Job that has `active: true` and matches the set trigger condition, if any.
    This Step is skipped if `JT__SKIP_TRIG_MAIN_JOBS` is non-zero.
```
To be consistent in styling.

etirta: Too detail of information. I should not have included it to begin with, I missed this when AI suggesting it. Removing this line.

ci-operator/step-registry/trigger-jobs/trigger-jobs-commands.sh
etirta: These variables are guaranteed to be set, one way or the other, by CI Operator.

etirta: Too detailed, but will update to:
```
: 'Trigger condition is not met, skipping associated Jobs.'
```

ci-operator/step-registry/trigger-jobs/trigger-jobs-ref.yaml
etirta: No default because I want to force a Test Author to explicitly set it in the Job Configuration.

etirta: Ditto the above.

ci-operator/step-registry/trigger-jobs/post/interop-ocp-watcher-bot-send-message/trigger-jobs-post-interop-ocp-watcher-bot-send-message-commands.sh
oharan2, line 28: Encapsulate logic and improve readability
```suggestion
# Ensure Requirements:
function ensureJqExists() {
    # Install jq manually if not found in image
    typeset binDir="/tmp/bin"
    mkdir -p "${binDir}"

    # Check for jq,
    jq --version 2>/dev/null || {
            wget -qO "${binDir}/jq" \
                "https://github.com/jqlang/jq/releases/latest/download/jq-$(
                    uname -s | tr '[:upper:]' '[:lower:]' | sed 's/darwin/macos/'
                )-$(
                    uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/'
                )" && \
            chmod a+x "${binDir}/jq" && \
            "${binDir}/jq" --version && \
            PATH="${binDir}:${PATH}"
        }
    true
    }
```

ci-operator/step-registry/trigger-jobs/README.md
etirta: We want to balance documentation depth with readability. Since our primary audience is Software Engineers, I have kept the explanation concise by assuming familiarity with standard bitwise operations to avoid over-explaining.

However, I have updated the wording to explicitly state that this uses bitwise masking to ensure there is no ambiguity.

The logic is just a standard `(A & B) != 0` check. It supports multi-bit masking on either variable:
  - Using `JT__TRIG_COND_EXEC_FLGS` as selector (e.g., is set to 5):
      - trigCondFlgs=1 -> checked
      - trigCondFlgs=2 -> skipped
      - trigCondFlgs=4 -> checked
  - Using `trigCondFlgs` as selector (e.g., set to 3):
      - JT__TRIG_COND_EXEC_FLGS=1 -> checked
      - JT__TRIG_COND_EXEC_FLGS=2 -> checked
      - JT__TRIG_COND_EXEC_FLGS=4 -> skipped
  - Using both:
      - JT__TRIG_COND_EXEC_FLGS=3, trigCondFlgs=5 -> checked
      - JT__TRIG_COND_EXEC_FLGS=6, trigCondFlgs=8 -> skipped

ci-operator/step-registry/trigger-jobs/post/interop-ocp-watcher-bot-send-message/trigger-jobs-post-interop-ocp-watcher-bot-send-message-commands.sh
oharan2, line 80: better to assign outside of the cli execution command/entrypoint
```suggestion
typeset slackGroupId
typeset webookUrl

slackGroupId=$(printf '%q' "$(cat "${secretDir}/${JT__POST__WB__MENTIONED_GROUP_ID_SECRET_NAME}")")
webookUrl=$(printf '%q' "$(cat "${secretDir}/${JT__POST__WB__WEBHOOK_URL_SECRET_NAME}")")

    interop-ocp-watcher-bot \
        --job_file_path=${botJobList@Q} \
        --mentioned_group_id="${slackGroupId}" \
        --webhook_url= "${webookUrl}" \
        --job_group_name=${JT__POST__WB__JOB_GROUP_NAME@Q}
```

ci-operator/step-registry/trigger-jobs/trig/check-resource-owner/trigger-jobs-trig-check-resource-owner-commands.sh
oharan2, line 8: ditto
```suggestion
#   Ensure Requirements:
function ensureJqExists() {
```

ci-operator/step-registry/trigger-jobs/README.md
oharan2, line 127: ditto
```suggestion

```

oharan2, line 211: ditto (ref and scripts duplicates again)

oharan2, line 273:
```suggestion
 1. Set vars based on path to `cluster_profile` secrets.
```

ci-operator/step-registry/trigger-jobs/trigger-jobs-commands.sh
oharan2, line 91: Can this get encapsuled into a function called triggerJob?
```suggestion

```

oharan2, line 184: ditto (call `triggerJob`)

ci-operator/config/rhpit/interop-tests/rhpit-interop-tests-main__weekly_trigger.yaml
oharan2: Where do we actually parse the `<test_name>--<platform_and_release>` info? couldn't find it.
I'm trying to understand _why_ we need it / what for.

ci-operator/step-registry/trigger-jobs/README.md
oharan2, line 124: Can we delete and alternaively maintain a template `trigger-jobs-trig-template-ref.yaml` and `trigger-jobs-trig-template-commands.sh`? This section may become outdated very soon if our flow is updated.
I also saw other step registries doing it this way.

```suggestion

```

ci-operator/step-registry/trigger-jobs/post/interop-ocp-watcher-bot-send-message/trigger-jobs-post-interop-ocp-watcher-bot-send-message-commands.sh
etirta: I was considering this, but ultimately decided not to do this. because if there is another 2nd or Nth utility to install, then we will keep adding the `binDir` to the `PATH`, unless we put logic to it, which is kind of too much work.

Can we then just call it `EnsureReqs()`? Yes, but I'm not a fond of calling a function for 1x use only, unless there are multiple blocks all over the place, then yes it can be ok to call 1x functions.

This is just 1 block in the beginning of the file, and I wrote it in a way it's clear where the block start and end. There is a reason why I choose to write it in "inline" sub-shell. This give a very clear block boundary because the sub-shell lines are idented. And the 2nd or Nth utility, all are inside the sub-shell.

Think the sub-shell as the function block body 😄 and it will look as readable as function.

ci-operator/config/rhpit/interop-tests/rhpit-interop-tests-main__weekly_trigger.yaml
etirta: I'm saying if there is a need for it. I'd rather to think ahead and prepare for it, rather than later have to update tons of files because it's not parse-able, when the need arise.

Remember changing something that has been published is not always easy. If we publish a Job or Step that is used by others, then changing it will break them. So that's why our philosophy should be prepare and get it right from the get go. Anticipate potential future need, even though we don't need to yet now. But of course, we need to balance this. It's impossible to predict everything.

At this point, I don't see any harm using `--` we have been using it and if it's the Test Name (or Step Name) CI Operator don't care.

ci-operator/step-registry/trigger-jobs/post/interop-ocp-watcher-bot-send-message/trigger-jobs-post-interop-ocp-watcher-bot-send-message-commands.sh
etirta: Why? I'm also not fond of 1x used variable either. With CI Operator enforcing `spellcheck`, we have to have separate lines between the declaration and assignment from command substitution. It's adding 4 more lines in this case.

Does `interop-ocp-watcher-bot` not peform a check on those parameter (sanitize it) before using it (so we need to fail if those var. are empty)?

If there is a good reason, I'm willing to do that.

ci-operator/step-registry/trigger-jobs/trig/check-resource-owner/trigger-jobs-trig-check-resource-owner-commands.sh
etirta: Ditto the above.

ci-operator/step-registry/trigger-jobs/README.md
etirta: Are you referring to the example?

It's expected to ALWAYS update the README to reflect the current code / workflow. So it doesn't matter whether we put in the README or a template files, either way it need to be updated.

But I'm intrigued on the approach and I may favor it. My current concern are:
  - The example need to be a working example, not something I can just use `...` for ex. or it will break the `spellcheck`, etc. Or even `make update`.
  - The template will go to official Step Registry documentation. I may be ok with this.

I'm not dismissing nor discounting this idea, but let me ponder it. We can always submit anather PR. Thank you for having a different idea and voice it!!! 👏 

But right now we need to get this PR merged so our FIPS Job can start triggering.

etirta: See the above response.

etirta: Ditto above response.

etirta: This will be removed once the legacy code are removed. Let's not do cosmetic only update at this point.

Good spotting though 👍, I would have accepted this fix, had it for non-deprecated information. Keep it up!!!

ci-operator/step-registry/trigger-jobs/trigger-jobs-commands.sh
etirta: Not for fuction that is called fom 1 place only. I don't see any benefit for it. Or?

etirta: Ditto the above.

etirta:
/pj-rehearse periodic-ci-rhpit-interop-tests-main-weekly_trigger-bm-gs--lp-interop periodic-ci-rhpit-interop-tests-main-weekly_trigger-bm-gs--lp-interop--watcher-bot periodic-ci-rhpit-interop-tests-main-weekly_trigger-ocp-self-managed-lp-interop--fips--421 periodic-ci-rhpit-interop-tests-main-weekly_trigger-ocp-self-managed-lp-interop--fips--watcher-bot--421 periodic-ci-rhpit-interop-tests-main-weekly_trigger-opp-interop-jobs-trigger periodic-ci-rhpit-interop-tests-main-weekly_trigger-opp-interop-jobs-watcher-bot-message

etirta:
/pj-rehearse ack

Some Jobs are re-rehearsed for sanity only.

ci-operator/step-registry/trigger-jobs/trig/check-resource-owner/README.md
sg-rh: okay yes could go in the general README or future additions accordingly

PR #74287, opened 2026-02-03
NTO: Fix e2e-hypershift-pao tests by setting CLUSTER_TYPE=hypershift

SargunNarula:
After commit 26615a37031, cluster_profile was changed from 'hypershift' to 'hypershift-aws'. This causes ci-operator to automatically set CLUSTER_TYPE=hypershift-aws. This behavior is documented in the step-registry [code](https://github.com/openshift/release/blob/master/ci-operator/step-registry/sandboxed-containers-operator/record-metadata/sandboxed-containers-operator-record-metadata-commands.sh#L138). The mapping occurs in [gen.go](https://github.com/openshift/ci-tools/blob/main/pkg/steps/multi_stage/gen.go#L493-L499), which then resolves to values defined in [types.go](https://github.com/openshift/ci-tools/blob/main/pkg/api/types.go#L1982-L1991),

The NTO test [code](https://github.com/openshift/cluster-node-tuning-operator/blob/main/test/e2e/performanceprofile/functests/utils/hypershift/hypershift.go#L16) checks for an exact match. With the new value 'hypershift-aws', the HyperShift detection fails and tests try to create PerformanceProfile CRDs instead of ConfigMaps.

This fix explicitly sets CLUSTER_TYPE=hypershift to override the automatic value set by ci-operator, restoring the expected behavior for:
- e2e-hypershift-pao
- e2e-pao-updating-profile-hypershift

### DIFF
--- a/ci-operator/config/openshift/cluster-node-tuning-operator/openshift-cluster-node-tuning-operator-main.yaml
+++ b/ci-operator/config/openshift/cluster-node-tuning-operator/openshift-cluster-node-tuning-operator-main.yaml
@@ -61,6 +61,7 @@ tests:
     - as: e2e-hypershift-pao
       cli: latest
       commands: |-
+        export CLUSTER_TYPE=hypershift
         export CLUSTER_NAME="$(echo -n $PROW_JOB_ID|sha256sum|cut -c-20)"
         export HYPERSHIFT_MANAGEMENT_CLUSTER_KUBECONFIG="${SHARED_DIR}/management_cluster_kubeconfig"
         export HYPERSHIFT_HOSTED_CONTROL_PLANE_NAMESPACE="$(${CLI_DIR}/oc get hcp -A -o=jsonpath="{.items[?(@.metadata.name==\"$CLUSTER_NAME\")].metadata.namespace}")"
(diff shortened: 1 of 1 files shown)


shajmakh:
Thanks for catching and handling this.
for more context, can you point to where "This causes ci-operator to automatically set CLUSTER_TYPE=hypershift-aws" happens? IOW where is the cluster_profile interpreted automatically to set the CLUSTER_TYPE?

SargunNarula:
> IOW where is the cluster_profile interpreted automatically to set the CLUSTER_TYPE?

The CLUSTER_TYPE env variable is automatically derived from cluster_profile by ci-operator. This behavior is documented in the step-registry [code](https://github.com/openshift/release/blob/master/ci-operator/step-registry/sandboxed-containers-operator/record-metadata/sandboxed-containers-operator-record-metadata-commands.sh#L138). 

The mapping occurs in [gen.go](https://github.com/openshift/ci-tools/blob/main/pkg/steps/multi_stage/gen.go#L493-L499), where profile.ClusterType() is called to set the CLUSTER_TYPE env var. This method resolves to hardcoded mappings defined in [types.go](https://github.com/openshift/ci-tools/blob/main/pkg/api/types.go#L1982-L1991), where cluster_profile: hypershift-aws maps to CLUSTER_TYPE=hypershift-aws.

PR #74311, opened 2026-02-03
Use direct access via Public IP

janboll:
This changes ARO-HCP E2E Tests to connect to the Frontend External IP/Loadbalancer instead of using tunneling. It is proposed to make connection more robust and remove related errors.

### DIFF
--- a/ci-operator/step-registry/aro-hcp/test/local/aro-hcp-test-local-commands.sh
+++ b/ci-operator/step-registry/aro-hcp/test/local/aro-hcp-test-local-commands.sh
@@ -23,7 +23,7 @@ az account set --subscription "${INFRA_SUBSCRIPTION_ID}"
 make frontend-grant-ingress DEPLOY_ENV=prow
 az account set --subscription "${SUBSCRIPTION_ID}"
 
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/aro-hcp/test/local/aro-hcp-test-local-commands.sh
```diff
@@ -21,100 +21,13 @@ SERVICE="aro-hcp-frontend"
...
-        echo "Port forward running (PID: $(cat "$PIDFILE"))"
-        return 0
-    else
-        echo "Port forward failed to start"
-        rm -f "$PIDFILE"
-        return 1
-    fi
-}
+FRONTEND_ADDRESS=$(kubectl get virtualservice -n aro-hcp aro-hcp-vs-frontend -o jsonpath='{.spec.hosts[0]}')
+ADMIN_API_ADDRESS=$(kubectl get virtualservice -n aro-hcp-admin-api admin-api-vs -o jsonpath='{.spec.hosts[0]}')
```
geoberle, line 20: FRONTEND_ADDRESS is expected by https://github.com/Azure/ARO-HCP/pull/3979 to be an URL with protocol prefix

geoberle: ah i see you handle it at the bottom of the file. all good

PR #74334, opened 2026-02-03
GCP-293/GCP-294: Add GKE provision/deprovision steps and e2e workflow

cristianoveiga:
## Summary

This PR implements [GCP-293](https://issues.redhat.com//browse/GCP-293) (GKE Provision/Deprovision Steps) and [GCP-294](https://issues.redhat.com//browse/GCP-294) (E2E Workflow and CI Configuration) for HyperShift GCP support.

**CI Rehearsal:** [Successful](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/74334/rehearse-74334-pull-ci-openshift-hypershift-main-e2e-gke/2019471435348578304) (~54 min total)

### GKE Provision/Deprovision Steps ([GCP-293](https://issues.redhat.com//browse/GCP-293))
- `ci-operator/step-registry/gke/provision/` - Creates GCP projects, VPC, and GKE Autopilot cluster
- `ci-operator/step-registry/gke/deprovision/` - Cleans up GKE cluster and GCP projects
- Added resilience for partial provision failures in deprovision step

### E2E Workflow ([GCP-294](https://issues.redhat.com//browse/GCP-294))
- `ci-operator/step-registry/hypershift/gcp/gke/e2e/` - E2E workflow definition
- `ci-operator/step-registry/hypershift/gcp/run-e2e/` - E2E test step
- Added GCP section to `hypershift-install-commands.sh` with:
  - gcloud CLI and GKE auth plugin installation
  - Prometheus operator CRDs (ServiceMonitors, PrometheusRules, PodMonitors)
  - OpenShift Route CRD and DNSEndpoint CRD
  - `--external-dns-provider=google --private-platform=GCP`
  - GCP_PROJECT and GCP_REGION env vars on operator deployment (workaround for [GCP-402](https://issues.redhat.com//browse/GCP-402))

### CI Configuration
- Added `e2e-gke` test to `openshift-hypershift-main.yaml` with safe introduction flags:
  - `always_run: false`
  - `optional: true`
  - `skip_report: true`
  - `run_if_changed` for GCP-related paths

## Workflow Structure

```
pre:
  - ipi-install-rbac   # Grant image-puller permissions
  - gke-provision      # Create GCP projects, VPC, GKE cluster
  - hypershift-install # Install HyperShift operator (GCP mode)
test:
  - hypershift-gcp-run-e2e  # Basic validation
post:
  - gke-deprovision    # Cleanup
```

## Design Decisions

1. **GKE Autopilot** - Managed node pools, similar to AKS approach
2. **Disabled webhooks** - GKE Autopilot compatibility (same as AKS)
3. **Dynamic projects** - Created/deleted per test run for isolation
4. **upi-installer image (4.22)** - Provides gcloud CLI pre-installed

## CI Rehearsal Results

| Phase | Duration |
|-------|----------|
| Image builds | ~33 min |
| gke-provision | ~10 min |
| hypershift-install | ~4 min |
| hypershift-gcp-run-e2e | ~1 min |
| gke-deprovision | ~6 min |
| **Total** | **~54 min** |

## Testing

Run rehearsal with:
```
/pj-rehearse pull-ci-openshift-hypershift-main-e2e-gke
```

## Follow-up Work

- [GCP-401](https://issues.redhat.com/browse/GCP-401) - HyperShift Operator GCP Setup:
  - **cert-manager** - Install cert-manager and configure ClusterIssuer for TLS certificates
  - **Workload Identity** - Configure GKE Workload Identity for PSC operations
- [GCP-402](https://issues.redhat.com/browse/GCP-402) - Add `--gcp-project` and `--gcp-region` flags to hypershift install command (upstream fix to remove workaround)
- [GCP-295](https://issues.redhat.com/browse/GCP-295) - Add GCP platform support to E2E test framework and implement cluster lifecycle tests in hypershift repo (enables creating hosted clusters in CI)

## Related

- Depends on: [GCP-378](https://issues.redhat.com//browse/GCP-378) (Cluster Profile) - ✅ Merged
- Follow-up: [GCP-401](https://issues.redhat.com/browse/GCP-401), [GCP-402](https://issues.redhat.com/browse/GCP-402), [GCP-295](https://issues.redhat.com/browse/GCP-295)
- JIRA: [GCP-293](https://issues.redhat.com/browse/GCP-293), [GCP-294](https://issues.redhat.com/browse/GCP-294)

### DIFF
--- a/ci-operator/step-registry/gke/provision/gke-provision-commands.sh
+++ b/ci-operator/step-registry/gke/provision/gke-provision-commands.sh
@@ -134,6 +134,7 @@ gcloud container clusters create-auto "${CLUSTER_NAME}" \
     --cluster-secondary-range-name="gke-pods" \
     --services-secondary-range-name="gke-services" \
     --release-channel="${RELEASE_CHANNEL}" \
+    --enable-private-nodes \
     --quiet
 
 # ============================================================================


ci-operator/step-registry/hypershift/gcp/gke/e2e/hypershift-gcp-gke-e2e-workflow.yaml
```diff
@@ -0,0 +1,30 @@
...
+    1. gke-deprovision: Delete GKE cluster and GCP projects
+  steps:
+    pre:
+    - ref: ipi-install-rbac
+    - ref: gke-provision
+    - ref: hypershift-install
+    test:
+    - ref: hypershift-gcp-run-e2e
+    post:
+    - ref: gke-deprovision
```
cblecker, line 27:
```suggestion
    post:
    - chain: gather
    - ref: gke-deprovision
```

We should collect artifacts on failure

ci-operator/step-registry/hypershift/install/hypershift-install-commands.sh
```diff
@@ -97,7 +97,87 @@ if [ "${CLOUD_PROVIDER}" == "Azure" ]; then
   ${EXTRA_ARGS}
 fi
 
-if [ "${CLOUD_PROVIDER}" != "Azure" ] && [ "${CLOUD_PROVIDER}" != "AWS" ]; then
+if [ "${CLOUD_PROVIDER}" == "GCP" ]; then
```
cblecker: Should this be a `case` switch on `${CLOUD_PROVIDER}`? We can collapse the `if [ "${AKS}" == "true" ]; then` under the `"${CLOUD_PROVIDER}" == "Azure"` section before the hypershift install commands, as we won't ever have an AKS cluster on a non-Azure CP

ci-operator/step-registry/gke/provision/gke-provision-commands.sh
```diff
@@ -0,0 +1,205 @@
...
+    --services-secondary-range-name="gke-services" \
+    --release-channel="${RELEASE_CHANNEL}" \
+    --quiet
+
+echo "Waiting for GKE cluster to be ready"
+gcloud container clusters describe "${CLUSTER_NAME}" \
+    --project="${MGMT_PROJECT_ID}" \
+    --region="${GCP_REGION}" \
+    --format="value(status)" | grep -q "RUNNING"
+
```
cblecker:
```suggestion
```

gcloud container clusters create-auto (line 170) is synchronous by default -- it blocks until the cluster reaches RUNNING status. This single-shot status check is therefore redundant. If create-auto succeeded
 (exited 0), the cluster is already RUNNING.

ci-operator/step-registry/gke/deprovision/gke-deprovision-commands.sh
```diff
@@ -0,0 +1,160 @@
...
+    if ! gcloud container clusters describe "${CLUSTER_NAME}" \
+        --project="${MGMT_PROJECT_ID}" \
+        --region="${GCP_REGION}" 2>/dev/null; then
+        echo "GKE cluster deleted successfully"
+        break
+    fi
+    echo "Waiting for GKE cluster deletion... (attempt $i/30)"
+    sleep 10
+done
+
```
cblecker: gcloud container clusters delete is synchronous by default -- it blocks until the cluster is fully deleted. The || true on line 43 suppresses errors (e.g., cluster already gone), which is correct for a cleanup
 script.

 The polling loop on lines 47-56 is therefore redundant when the delete command succeeds. However, it serves a purpose when the delete command fails (caught by || true) -- it polls to see if the cluster is
 actually gone. This is a reasonable defensive pattern.

 But there's a subtle issue: If gcloud container clusters delete fails for a reason other than "cluster not found" (e.g., permission error, network timeout) and returns non-zero (caught by || true), the polling
 loop will wait 5 minutes and then silently proceed. The cluster may still be running, and subsequent VPC cleanup will fail because GKE holds references to VPC resources.

 Recommendation: Consider checking the error from the delete command more explicitly:

 if ! gcloud container clusters delete "${CLUSTER_NAME}" \
     --project="${MGMT_PROJECT_ID}" \
     --region="${GCP_REGION}" \
     --quiet 2>&1; then
     echo "WARNING: Cluster delete command failed, checking if cluster exists..."
 fi

 Alternatively, use --async and then poll, which gives you explicit control:

 gcloud container clusters delete "${CLUSTER_NAME}" \
     --project="${MGMT_PROJECT_ID}" \
     --region="${GCP_REGION}" \
     --quiet --async || true

 # Now poll is the primary wait mechanism
 for i in {1..60}; do
     if ! gcloud container clusters describe "${CLUSTER_NAME}" \
         --project="${MGMT_PROJECT_ID}" \
         --region="${GCP_REGION}" 2>/dev/null; then
         echo "GKE cluster deleted successfully"
         break
     fi
     echo "Waiting for GKE cluster deletion... (attempt $i/60)"
     sleep 10
 done

 Using --async + polling is arguably cleaner because:
 - The delete returns immediately, and the poll loop is the single source of truth for completion
 - No double-waiting (sync delete blocking + then polling on top)
 - Better timeout control (60 attempts x 10s = 10 minutes max wait)

ci-operator/step-registry/gke/deprovision/gke-deprovision-commands.sh
```diff
@@ -0,0 +1,160 @@
...
+        gcloud compute networks delete "${vpc}" \
+            --project="${CUSTOMER_PROJECT_ID}" --quiet || true
+    done
+
+    # Delete customer project
+    echo "Deleting customer project: ${CUSTOMER_PROJECT_ID}"
+    gcloud projects delete "${CUSTOMER_PROJECT_ID}" --quiet || true
+else
+    echo "No customer project to clean up"
+fi
```
cblecker: Do we need to clean up individual things in the customer project if we are just going to delete it?

ci-operator/step-registry/gke/deprovision/gke-deprovision-commands.sh
```diff
@@ -0,0 +1,160 @@
...
+
+# ----------------------------------------------------------------------------
+# Step 4: Delete management project (now safe after explicit resource cleanup)
+# ----------------------------------------------------------------------------
+
+echo "Deleting management project: ${MGMT_PROJECT_ID}"
+gcloud projects delete "${MGMT_PROJECT_ID}" --quiet || true
+
+echo "Cleanup complete"
```
cblecker, line 82: Same here for the management project

ci-operator/step-registry/gke/deprovision/gke-deprovision-commands.sh
```diff
@@ -0,0 +1,160 @@
...
+# Check if provision completed - if not, nothing to clean up
+if [[ ! -f "${SHARED_DIR}/mgmt-project-id" ]]; then
+    echo "No mgmt-project-id file found - provision may not have completed"
+    echo "Nothing to deprovision, exiting successfully"
+    exit 0
+fi
+
+# Load cluster info from provision step
+MGMT_PROJECT_ID="$(<"${SHARED_DIR}/mgmt-project-id")"
+CUSTOMER_PROJECT_ID="${SHARED_DIR}/customer-project-id"
```
cblecker: Line 18 initially assigns the file path to CUSTOMER_PROJECT_ID, then on line 19 uses that variable value (a file path string) as the argument to -f. This works coincidentally because the path is valid and -f
 checks if that path is a file. However, this is confusing to read -- the variable holds a path on line 18, then gets reassigned to the file's contents on line 20. A clearer pattern:

 CUSTOMER_PROJECT_ID_FILE="${SHARED_DIR}/customer-project-id"
 if [[ -f "${CUSTOMER_PROJECT_ID_FILE}" ]]; then
     CUSTOMER_PROJECT_ID="$(<"${CUSTOMER_PROJECT_ID_FILE}")"
 else
     CUSTOMER_PROJECT_ID=""
 fi

 This isn't a functional bug, but it's a readability concern that could lead to future maintenance issues.

ci-operator/step-registry/hypershift/install/hypershift-install-commands.sh
```diff
@@ -97,7 +97,87 @@ if [ "${CLOUD_PROVIDER}" == "Azure" ]; then
...
+  GKE_AUTH_PLUGIN_INSTALL_DIR="${GCLOUD_INSTALL_DIR}/bin" "${INSTALL_SCRIPT}"
+  rm -f "${INSTALL_SCRIPT}"
+  export USE_GKE_GCLOUD_AUTH_PLUGIN=True
+
+  echo "gcloud version: $(gcloud version 2>/dev/null | head -1)"
+
+  # Authenticate gcloud with the service account from the cluster profile
+  # CLUSTER_PROFILE_DIR is available to all steps when cluster_profile is specified
+  echo "Authenticating gcloud with service account..."
+  gcloud auth activate-service-account --key-file="${CLUSTER_PROFILE_DIR}/credentials.json"
```
cblecker: Per GCP docs, you can construct a kubeconfig with a short-lived access token instead of the exec plugin. In gke-provision, after get-credentials, you could replace the exec-based kubeconfig with one that embeds
  a bearer token:
```
  # Get a short-lived access token (valid ~60 minutes, sufficient for CI)
  ACCESS_TOKEN=$(gcloud auth print-access-token)

  # Build a static kubeconfig using the cluster CA and endpoint
  CLUSTER_CA=$(gcloud container clusters describe "${CLUSTER_NAME}" \
      --project="${MGMT_PROJECT_ID}" --region="${GCP_REGION}" \
      --format="value(masterAuth.clusterCaCertificate)")
  CLUSTER_ENDPOINT=$(gcloud container clusters describe "${CLUSTER_NAME}" \
      --project="${MGMT_PROJECT_ID}" --region="${GCP_REGION}" \
      --format="value(endpoint)")

  # Write kubeconfig with embedded token (no exec plugin needed)
  cat > "${SHARED_DIR}/kubeconfig" << EOF
  apiVersion: v1
  kind: Config
  clusters:
  - cluster:
      certificate-authority-data: ${CLUSTER_CA}
      server: https://${CLUSTER_ENDPOINT}
    name: gke-cluster
  contexts:
  - context:
      cluster: gke-cluster
      user: gke-user
    name: gke-context
  current-context: gke-context
  users:
  - name: gke-user
    user:
      token: ${ACCESS_TOKEN}
  EOF
```
  This would eliminate the need to download gcloud in hypershift-install entirely. The token is valid for ~60 minutes, which should be sufficient given hypershift-install runs immediately after gke-provision.

cblecker: Per GCP docs, you can construct a kubeconfig with a short-lived access token instead of the exec plugin. In gke-provision, after get-credentials, you could replace the exec-based kubeconfig with one that embeds
  a bearer token:
```
  # Get a short-lived access token (valid ~60 minutes, sufficient for CI)
  ACCESS_TOKEN=$(gcloud auth print-access-token)

  # Build a static kubeconfig using the cluster CA and endpoint
  CLUSTER_CA=$(gcloud container clusters describe "${CLUSTER_NAME}" \
      --project="${MGMT_PROJECT_ID}" --region="${GCP_REGION}" \
      --format="value(masterAuth.clusterCaCertificate)")
  CLUSTER_ENDPOINT=$(gcloud container clusters describe "${CLUSTER_NAME}" \
      --project="${MGMT_PROJECT_ID}" --region="${GCP_REGION}" \
      --format="value(endpoint)")

  # Write kubeconfig with embedded token (no exec plugin needed)
  cat > "${SHARED_DIR}/kubeconfig" << EOF
  apiVersion: v1
  kind: Config
  clusters:
  - cluster:
      certificate-authority-data: ${CLUSTER_CA}
      server: https://${CLUSTER_ENDPOINT}
    name: gke-cluster
  contexts:
  - context:
      cluster: gke-cluster
      user: gke-user
    name: gke-context
  current-context: gke-context
  users:
  - name: gke-user
    user:
      token: ${ACCESS_TOKEN}
  EOF
```
  This would eliminate the need to download gcloud in hypershift-install entirely. The token is valid for ~60 minutes, which should be sufficient given hypershift-install runs immediately after gke-provision.

cblecker: Per GCP docs, you can construct a kubeconfig with a short-lived access token instead of the exec plugin. In gke-provision, after get-credentials, you could replace the exec-based kubeconfig with one that embeds
  a bearer token:
```
  # Get a short-lived access token (valid ~60 minutes, sufficient for CI)
  ACCESS_TOKEN=$(gcloud auth print-access-token)

  # Build a static kubeconfig using the cluster CA and endpoint
  CLUSTER_CA=$(gcloud container clusters describe "${CLUSTER_NAME}" \
      --project="${MGMT_PROJECT_ID}" --region="${GCP_REGION}" \
      --format="value(masterAuth.clusterCaCertificate)")
  CLUSTER_ENDPOINT=$(gcloud container clusters describe "${CLUSTER_NAME}" \
      --project="${MGMT_PROJECT_ID}" --region="${GCP_REGION}" \
      --format="value(endpoint)")

  # Write kubeconfig with embedded token (no exec plugin needed)
  cat > "${SHARED_DIR}/kubeconfig" << EOF
  apiVersion: v1
  kind: Config
  clusters:
  - cluster:
      certificate-authority-data: ${CLUSTER_CA}
      server: https://${CLUSTER_ENDPOINT}
    name: gke-cluster
  contexts:
  - context:
      cluster: gke-cluster
      user: gke-user
    name: gke-context
  current-context: gke-context
  users:
  - name: gke-user
    user:
      token: ${ACCESS_TOKEN}
  EOF
```
  This would eliminate the need to download gcloud in hypershift-install entirely. The token is valid for ~60 minutes, which should be sufficient given hypershift-install runs immediately after gke-provision.

cristianoveiga: Right... I actually considered this option but I was concerned about the 60 minutes limitation.

I will give it a try.

ci-operator/step-registry/hypershift/gcp/gke/e2e/hypershift-gcp-gke-e2e-workflow.yaml
cristianoveiga: The `gather` chain is OpenShift-specific - it uses `oc adm must-gather`, queries OpenShift CRDs (ClusterOperator, ClusterVersion, MachineConfigPool, etc.), and accesses OpenShift namespaces (openshift-monitoring, openshift-etcd, etc.). None of this works on a vanilla GKE cluster.

I used `hypershift-dump` instead, which is designed to work with HyperShift on any management cluster (GKE, AKS, or OpenShift). It runs `bin/hypershift dump cluster` to collect HyperShift-specific artifacts. This is the same pattern used by `hypershift-azure-aks-conformance` and other AKS-based workflows.

ci-operator/step-registry/hypershift/install/hypershift-install-commands.sh
cristianoveiga: Done - refactored to use case switch on `${CLOUD_PROVIDER}`.

ci-operator/step-registry/gke/provision/gke-provision-commands.sh
cristianoveiga: Done - removed the redundant status check.

ci-operator/step-registry/gke/deprovision/gke-deprovision-commands.sh
cristianoveiga: Done - implemented the `--async` + polling approach as suggested.

ci-operator/step-registry/gke/deprovision/gke-deprovision-commands.sh
cristianoveiga: Done - simplified to rely on project deletion. We still explicitly delete the GKE cluster first.

ci-operator/step-registry/gke/deprovision/gke-deprovision-commands.sh
cristianoveiga: Done.

ci-operator/step-registry/gke/deprovision/gke-deprovision-commands.sh
cristianoveiga: Done - now uses `CUSTOMER_PROJECT_FILE` for the path and `CUSTOMER_PROJECT_ID` for the contents.

ci-operator/step-registry/hypershift/install/hypershift-install-commands.sh
cristianoveiga: Done - implemented static kubeconfig with embedded access token. Also added `set +x` around the sensitive operations to prevent token/CA exposure in CI logs.

ci-operator/step-registry/gke/deprovision/gke-deprovision-ref.yaml
```diff
@@ -0,0 +1,30 @@
...
+  documentation: |-
+    This step destroys the GKE cluster and cleans up GCP projects created by gke-provision.
+
+    IMPORTANT: GCP project deletion does NOT automatically clean up all resources.
+    Some resources (VPC, GKE clusters) can block project deletion and must be
+    deleted explicitly first. This step handles the proper cleanup order:
+    1. Delete GKE cluster
+    2. Delete VPC resources (firewall rules, NAT, router, subnets, VPC)
+    3. Delete customer project resources
+    4. Delete both projects
```
cblecker: I don't think this comment is accurate anymore, right? Perhaps we need to, as a part of the deprovision, confirm there are no liens on the project, and then we don't even have to delete the GKE cluster manually: https://docs.cloud.google.com/resource-manager/docs/troubleshooting-project-deletion

ci-operator/step-registry/gke/provision/gke-provision-commands.sh
```diff
@@ -0,0 +1,198 @@
...
+    --project="${MGMT_PROJECT_ID}" \
+    --region="${GCP_REGION}" \
+    --format="value(masterAuth.clusterCaCertificate)")
+CLUSTER_ENDPOINT=$(gcloud container clusters describe "${CLUSTER_NAME}" \
+    --project="${MGMT_PROJECT_ID}" \
+    --region="${GCP_REGION}" \
+    --format="value(endpoint)")
+ACCESS_TOKEN=$(gcloud auth print-access-token)
+
+cat > "${SHARED_DIR}/kubeconfig" << EOF
```
cblecker, line 165: Could we be descriptive that this is the "Control Plane cluster" (aka management cluster)?

ci-operator/step-registry/gke/deprovision/gke-deprovision-ref.yaml
cristianoveiga: Done - updated documentation to reflect the simplified cleanup order (delete Hosted Cluster project, delete GKE cluster with polling, delete Control Plane project). Also added note about GKE clusters placing liens on projects.

ci-operator/step-registry/gke/provision/gke-provision-commands.sh
cristianoveiga: Done. I've updated the kubeconfig entry and added a comment about this being the Control Plane cluster (kept the name of kubeconfig file since it's used in other steps (including hypershift install)

PR #74343, opened 2026-02-03
NO-JIRA: feat: adding prow and plugin configs for new edge repos

jeff-roche:
Configuration for two new edge repos:
- `openshift-eng/edge-tooling`
- `openshift-eng/edge-hiring`


core-services/prow/02_config/openshift-eng/edge-tooling/OWNERS
```diff
@@ -0,0 +1,24 @@
...
+- suleymanakbas91
+options: {}
+reviewers:
+- eggfoobar
+- jaypoulz
+- jeff-roche
+- jerpeter1
+- pacevedom
+- qjkee
+- suleymanakbas91
```
brandisher: We have OWNERS defined in the individual repos. Why are they defined again here?

core-services/prow/02_config/openshift-eng/edge-hiring/OWNERS
```diff
@@ -0,0 +1,24 @@
...
+- suleymanakbas91
+options: {}
+reviewers:
+- eggfoobar
+- jaypoulz
+- jeff-roche
+- jerpeter1
+- pacevedom
+- qjkee
+- suleymanakbas91
```
brandisher: We have OWNERS defined in the individual repos. Why are they defined again here?

core-services/prow/02_config/openshift-eng/edge-hiring/_pluginconfig.yaml
```diff
@@ -0,0 +1,37 @@
...
+  repos:
+  - openshift-eng/edge-hiring
+  require_self_approval: false
+lgtm:
+- repos:
+  - openshift-eng/edge-hiring
+  review_acts_as_lgtm: true
+plugins:
+  openshift-eng/edge-hiring:
+    plugins:
```
brandisher, line 12: Question because I haven't worked with Prow in a while - are all of these plugins necessary? These repos are internal so we likely won't have public CI for them and really only need prow to handle PR merging.

jeff-roche: This is just the standard set, I figured it doesn't hurt to have them

core-services/prow/02_config/openshift-eng/edge-hiring/OWNERS
jeff-roche: I just copy/pasted this from the lvm-operator config. I believe this will be auto-updated by prow after merging to include the complete list (copied from the repo iirc). We have to have an owners file in this directory though, otherwise it will fail PR checks

brandisher: I'm confused. If Prow will respect the OWNERS files in the respective repos, why do we need to define a list here at all?

core-services/prow/02_config/openshift-eng/edge-hiring/_pluginconfig.yaml
brandisher: Can we reduce this to a minimal list of what we need? In line with least privileges, we probably want to prefer a curated set of plugins instead of a generic list that we're unlikely to use.

core-services/prow/02_config/openshift-eng/edge-hiring/OWNERS
jeff-roche: I believe it's for this folder specifically. It assumes the one in the repository should be duplicated here and have the same owners

brandisher: Ah okay. Let's take action on that assumption then and replace this list with the list used the currently defined OWNERS lists in the respective repos. Note that each repo has OWNERS and OWNERS_ALIAS; each list is different intentionally due to the subject matter in each repo.

core-services/prow/02_config/openshift-eng/edge-hiring/_pluginconfig.yaml
jeff-roche: Dropped it to `lgtm` and `approve` which I think is our min

core-services/prow/02_config/openshift-eng/edge-hiring/OWNERS
jeff-roche: Updated to match the lists for each repo

core-services/prow/02_config/openshift-eng/edge-tooling/OWNERS
brandisher, line 5: Nit but this is no longer true :-)

core-services/prow/02_config/openshift-eng/edge-hiring/OWNERS
brandisher, line 5: Nit but this is no longer true :-)

PR #74356, opened 2026-02-04
ACM: cut foundation repos to release 2.17

zhujian7:


ci-operator/config/stolostron/cluster-proxy/stolostron-cluster-proxy-main.yaml
```diff
@@ -23,7 +23,7 @@ images:
 promotion:
   to:
   - disabled: true
-    name: "2.11"
+    name: "2.12"
```
elgnay: There will be no 2.12 for MCE. The next MCE version will align with ACM version. They both use 2.17. See https://redhat-internal.slack.com/archives/C012L9Y9C79/p1770127046039489 for more details.

PR #74424, opened 2026-02-04
CONSOLE-5072: CI updates and prepare for node 22

logonoff:
- Backport https://github.com/openshift/release/pull/39470 to 4.12 and 4.13 (needed by https://github.com/openshift/console/pull/15988 and https://github.com/openshift/console/pull/15989)
- Give more CPU/RAM to console jobs due to slow runs and flakiness
- Use ART base image for console-plugin-demo target

### DIFF
--- a/ci-operator/config/openshift/console/openshift-console-main.yaml
+++ b/ci-operator/config/openshift/console/openshift-console-main.yaml
@@ -42,11 +42,11 @@ resources:
       memory: 14Gi
   console:
     requests:
-      cpu: "1"
-      memory: 8Gi
+      cpu: "4"
+      memory: 10Gi
   frontend:
     requests:
-      cpu: "1"
+      cpu: "2"
       memory: 10Gi
   test-bin:
     requests:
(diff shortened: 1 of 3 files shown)


ci-operator/config/openshift/console/openshift-console-release-4.12.yaml
```diff
@@ -70,7 +70,9 @@ tests:
   container:
     from: ocp_builder_rhel-8-golang-1.18-openshift-4.12
 - as: frontend
-  commands: ./test-frontend.sh
+  commands: |
+    git config --global --add safe.directory /go/src/github.com/openshift/console
```
jhadvig, line 74: not sure why this is needed to tag this directory as safe ?

logonoff: See https://github.com/openshift/console/pull/12828#issuecomment-1552772045 and https://github.com/openshift/release/pull/39470:

> The frontend tests run some git commands to make sure that yarn.lock and i18n namespaces are up-to-date. This fails in CI since the tests run as a random user who does not own the directory. Tell git that the console directory is safe.

PR #74439, opened 2026-02-05
CORENET-6614: Add skip_if_only_changed for CNO and OVNK

jluhrsen:
adds patterns to check and if only files matching those files were changed for all e2e jobs in OVNK and CNO.

no need to run full e2e tests on real clusters if the only files changed would have no real effect on the code running in the cluster

### DIFF
--- a/ci-operator/config/openshift/cluster-network-operator/openshift-cluster-network-operator-master.yaml
+++ b/ci-operator/config/openshift/cluster-network-operator/openshift-cluster-network-operator-master.yaml
@@ -50,6 +50,7 @@ tests:
   container:
     from: src
 - as: e2e-gcp-ovn-upgrade
+  skip_if_only_changed: ^(docs|\.vscode)/|\.md$|^(\.gitignore|\.gitattributes|\.golangci\.yaml|OWNERS|OWNERS_ALIASES|LICENSE|sample-.*\.yaml)$
   steps:
     cluster_profile: gcp
     env:
@@ -71,6 +72,7 @@ tests:
       EXTRA_MG_ARGS: --host-network
     workflow: openshift-e2e-aws-ovn
 - as: e2e-aws-ovn-upgrade
+  skip_if_only_changed: ^(docs|\.vscode)/|\.md$|^(\.gitignore|\.gitattributes|\.golangci\.yaml|OWNERS|OWNERS_ALIASES|LICENSE|sample-.*\.yaml)$
   steps:
     cluster_profile: aws-4
     env:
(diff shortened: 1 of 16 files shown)


jluhrsen:
/retitle [CORENET-6614](https://issues.redhat.com//browse/CORENET-6614): Add skip_if_only_changed for CNO and OVNK

ci-operator/config/openshift/cluster-network-operator/openshift-cluster-network-operator-master.yaml
```diff
@@ -50,6 +50,7 @@ tests:
   container:
     from: src
 - as: e2e-gcp-ovn-upgrade
+  skip_if_only_changed: ^(docs|\.vscode)/|\.md$|^(\.gitignore|\.ci-operator\.yaml|\.gitattributes|\.golangci\.yaml|OWNERS|OWNERS_ALIASES|LICENSE|sample-.*\.yaml)$
```
jcaamano: There's stuff I can't find, while there might be somethign else missing. Not sure if this is a fixed pattern you picked up from somewhere.

`.vscode`?
`.ci-operator\.yaml` wouldn't we want to test if this changes?
`.gitattributes`?
`\.golangci\.yaml`?
`OWNERS_ALIASES`?
`sample-.*\.yaml`?

I am ok if we add some stuff that it's not there today but might be there tomorrow, however not all on that list fits my view on it.

Stuff that might be missing, not sure if you have considered:

`CODEOWNERS`
`.github`
`contrib`
`etc`
`helm`
`.coderabbit.yaml`
`crd-docs-config.yaml`
`mkdocs.yml`
`requirements.txt`

jluhrsen: > There's stuff I can't find, while there might be somethign else missing. Not sure if this is a fixed pattern you picked up from somewhere.

@jcaamano ,

I started with some common pattern I found in other projects then built on top of that. I think the main
problem here could be that I used the same pattern for CNO and OVNK but technically they can/should
be different. You are right that some patterns don't match one or the other repo.

I've updated now it to try to match more exactly to what each repo has or doesn't.

> 
> `.vscode`? `.ci-operator\.yaml` wouldn't we want to test if this changes? `.gitattributes`? `\.golangci\.yaml`? `OWNERS_ALIASES`? `sample-.*\.yaml`?

.vscode is in CNO

you are right about .ci-operator.yaml. I removed it from the pattern

I don't know what .gitattributes is for, but claude tells me it controls something about language stats and
won't affect the production code

I don't think we want to run e2e if .golangci.yaml changes. the lint job will still run if that changes, AFAIK

> 
> I am ok if we add some stuff that it's not there today but might be there tomorrow, however not all on that list fits my view on it.
> 
> Stuff that might be missing, not sure if you have considered:
> 
> `CODEOWNERS` `.github` `contrib` `etc` `helm` `.coderabbit.yaml` `crd-docs-config.yaml` `mkdocs.yml` `requirements.txt`

right, I think most of these are in OVNK and should now be reflected in those patterns.

How does it look now?

jcaamano: oh sorry I wasn't actually looking at CNO

> .vscode is in CNO

I think it shouldn't be there but let's go with it for now.

jluhrsen:
/pj-rehearse ack
I will make some test PRs after this lands to double check it's working like we think

PR #74446, opened 2026-02-05
NO-ISSUE: Add IBIO dual-stack jobs

danmanor:
DS supported for version 4.20 and DS with primary ipv6 from OCP 4.21


ci-operator/config/openshift/image-based-install-operator/openshift-image-based-install-operator-main.yaml
```diff
@@ -132,6 +200,42 @@ tests:
       CLUSTERTYPE: assisted_medium_el9
       SEED_IMAGE_TAG: "4.22"
     workflow: image-based-install-operator-ofcir-reinstall
+- always_run: false
+  as: ibio-reinstall-v4v6
```
giladravid16, line 204: Does reinstall not need the `IP_STACK` env?

danmanor: nice catch

PR #74453, opened 2026-02-05
Fix self-signed cert issue

sgaoshang:
After Nutanix's upgrade, we need to export self-signed cert to install OCP


sgaoshang:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.22-amd64-nightly-nutanix-ipi-gpu-f28

PR #74483, opened 2026-02-05
Triggering Jobs: Add generic time-based trigger condition

etirta:
## Related To
- [INTEROP-8852](https://issues.redhat.com/browse/INTEROP-8852): Triggering Jobs: Add generic time-based trigger condition

## Description
Add generic time-based trigger condition that allows users to determine the appropriate time using user-supplied `date` command parameters and arithmetic expressions for trigger decisions.

Add debugging information to existing scripts.


etirta:
Documentation update:
  - Forgot to add this new trigger condition Step to the supported list.

PR #74499, opened 2026-02-06
[FIX] OSP FIPS runs

srivickynesh:


Sau1506mya:
@srivickynesh , rehearsal job is passing. Just run `make update` and check for the change in periodics and pre-submit files. After that, we are good with this one.

srivickynesh:
No changes in periodics and pre-submit files after running the `make update`

PR #66733, opened 2025-07-04
OCPQE-30100: Use equinix ssh key for arm64 hosts

sgoveas:
packet-ssh-key would be replaced by equinix-ssh-key if it exists and host required is c3.large.arm64

PR #66613, opened 2025-07-01
OCPQE-30100: Move baremetalds arm jobs back to equinix

sgoveas:

PR #74548, opened 2026-02-07
Stop baremetal arm64 dev script jobs

sgoveas:
Baremetal dev script jobs were set (openshift/release#66613 and openshift/release#66733, ticket [OCPQE-30100](https://issues.redhat.com//browse/OCPQE-30100)) to run in equinix as there are no arm64 hosts in ofcir provider. Equinix also stopped providing arm64 hosts a while ago an now we no longer use equinix services. Until we have arm64 hosts available via ofcir provider these jobs are stopped.


sgoveas:
@honza please approve

PR #74580, opened 2026-02-09
DPTP-4676: Remove multiarch builder on build02/08

liangxia:
GCP clusters ( build02/04/08 ) does not have arm64 nodes, so no need to run multiarch builder on them.

### DIFF
--- a/core-services/ci-secret-generator/_config.yaml
+++ b/core-services/ci-secret-generator/_config.yaml
@@ -257,12 +257,10 @@
   params:
     cluster:
     - build01
-    - build02
     - build03
     - build05
     - build06
     - build07
-    - build08
     - build09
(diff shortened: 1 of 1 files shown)


liangxia:
> It's not clear to me whether we want all clusters to be the same (same CPU arch, configurations, etc.) or not. If it's not the case I'm fine with this PR.

I have the same thoughts that we would like to make all clusters to be the same. I tried that, but multiarch build does not work on build02 ( more detail in https://redhat-internal.slack.com/archives/GB7NB0CUC/p1770306016773289 ) which make build02 show as degrade/OutOfSync in ArgoCD.  So now GCP/AWS clusters are managed by different ArgoCD ApplicationSets.

PR #74471, opened 2026-02-05
Using ART maintained golang build images

hector-vido:
We noticed a lot of non-standard golang `build_root` images being used in our CI environment.

This PR is an effort to standardize it and use more reliable images with a known build process, described on [OCP Builder Images](https://docs.ci.openshift.org/docs/architecture/images/#golang-build-images).

This should not affect the applications since `build_image` is used only to generate binaries and artifacts that will be moved inside the application images created in our CI environment.

Problems will be handled individually, since this is a huge change and a rehearsal on 7k jobs will use a lot of resources.

The updated tags are the following ones:

```rhel-8-release-golang-1.14-openshift-4.6
rhel-8-release-golang-1.14-openshift-4.7
rhel-8-release-golang-1.14-openshift-4.8
rhel-8-release-golang-1.15-openshift-4.6
rhel-8-release-golang-1.15-openshift-4.7
rhel-8-release-golang-1.15-openshift-4.8
rhel-8-release-golang-1.16-openshift-4.10
rhel-8-release-golang-1.16-openshift-4.8
rhel-8-release-golang-1.16-openshift-4.9
rhel-8-release-golang-1.17-openshift-4.10
rhel-8-release-golang-1.17-openshift-4.11
rhel-8-release-golang-1.18-openshift-4.11
rhel-8-release-golang-1.18-openshift-4.12
rhel-8-release-golang-1.19-openshift-4.12
rhel-8-release-golang-1.19-openshift-4.13
rhel-8-release-golang-1.19-openshift-4.14
rhel-8-release-golang-1.19-openshift-4.15
rhel-8-release-golang-1.20-openshift-4.14
rhel-8-release-golang-1.20-openshift-4.15
rhel-8-release-golang-1.20-openshift-4.16
rhel-8-release-golang-1.20-openshift-4.17
rhel-8-release-golang-1.21-openshift-4.14
rhel-8-release-golang-1.21-openshift-4.15
rhel-8-release-golang-1.21-openshift-4.16
rhel-8-release-golang-1.21-openshift-4.17
rhel-8-release-golang-1.21-openshift-4.18
rhel-8-release-golang-1.22-openshift-4.17
rhel-8-release-golang-1.22-openshift-4.18
rhel-8-release-golang-1.22-openshift-4.19
rhel-8-release-golang-1.23-openshift-4.16
rhel-8-release-golang-1.23-openshift-4.17
rhel-8-release-golang-1.23-openshift-4.18
[...]

PR #74618, opened 2026-02-09
Azure/ARO-RP: fix base image

stevekuznetsov:


ci-operator/config/Azure/ARO-RP/Azure-ARO-RP-master.yaml
```diff
@@ -2,7 +2,7 @@ build_root:
   image_stream_tag:
     name: release
     namespace: openshift
-    tag: golang-1.25
+    tag: rhel-9-release-golang-1.25-openshift-4.22
```
mociarain, line 10: Just to note I don't see the 4.22 tag in here: https://github.com/openshift/release/pull/74471

This has been on my radar but it's not clear to me yet. We (ARO-RP) possibly need to move to `rhel-9-release-golang-1.25-openshift-4.23`

PR #74686, opened 2026-02-10
[ossm] Fix empty string in HUB env for istio integration tests

mkralik3:
This PR https://github.com/openshift/release/pull/74603/changes#diff-8b512d792b5b41108d1a4b89fa8924774308ab8844678354f3a4cda26bef1879R113-R114  introduced `HUB` env, which is an empty string when it is not set. 

However, when it is an empty string, it causes an error in the Istio  Makefile:
```
HUB ?=istio
ifeq ($(HUB),)
  $(error "HUB cannot be empty")
endif
```

Added default value in the step to be sure the HUB env is always set.


fjglira (approved):
LGTM

PR #74689, opened 2026-02-10
monitoring-plugin increase CPU constraints

PeterYurkovich:


ci-operator/config/openshift/monitoring-plugin/openshift-monitoring-plugin-main.yaml
```diff
@@ -34,9 +34,10 @@ releases:
 resources:
   '*':
     limits:
+      cpu: "32"
```
deepsm007: this wont work, the pod needs more memory 8Gi

PR #74696, opened 2026-02-10
NO-ISSUE: shard metal mco-disruptive suites

djoshy:
The current metal mco-disruptive suites all hit a timeout, sharding might help mitigate this.


djoshy:
Note that the jobs are actually completing without hitting a timeout now, test failures were pre-existing

PR #74766, opened 2026-02-11
Enable Node Auto Provisioning (NAP) for AKS management clusters

bryan-cox:
## Summary
- Adds NAP (Karpenter-based) support to the `aks-provision` step to avoid `OverconstrainedZonalAllocationRequest` errors when a specific VM SKU (e.g., `Standard_D16_v3`) is unavailable in a zone
- When `ENABLE_NAP=true`, the cluster is created with `--node-provisioning-mode Auto` and Karpenter NodePool/AKSNodeClass CRDs are applied to allow flexible VM SKU selection within the specified family (D) and CPU count (16)
- Placeholder pods with topology spread constraints trigger node provisioning across all specified zones, then are cleaned up after nodes are ready
- Existing non-NAP behavior is completely unchanged when `ENABLE_NAP` is not set
- NAP is only enabled for the `e2e-aks` job; conformance workflows are unchanged

## Why placeholder pods?

Karpenter is reactive — it only provisions nodes when pods are pending and unschedulable. The [recommended overprovisioning pattern](https://deepwiki.com/aws-samples/karpenter-blueprints/4.2.4-overprovisioning) uses low-priority pause pods that get preempted by real workloads, keeping a standing capacity buffer.

For this CI use case we use a simplified variant: deploy a placeholder Deployment with resource requests sized to fill one D16-equivalent node per replica, wait for all nodes to reach Ready, then delete the Deployment before the test workload runs. This avoids the need for PriorityClass configuration and is sufficient since we only need a one-time warm-up, not an ongoing buffer.

## Test plan
- [ ] Rehearse `pull-ci-openshift-hypershift-main-e2e-aks` job to validate NAP cluster creation
- [ ] Verify nodes are provisioned across all 3 zones
- [ ] Verify placeholder deployment is cleaned up after nodes are ready
- [ ] Confirm HyperShift install succeeds on NAP-provisioned cluster

### DIFF
--- a/ci-operator/step-registry/aks/provision/aks-provision-commands.sh
+++ b/ci-operator/step-registry/aks/provision/aks-provision-commands.sh
@@ -331,6 +331,29 @@ EOF
         NAP_ELAPSED=$((NAP_ELAPSED + 30))
     done
 
+    # Wait for zone labels to propagate on NAP nodes.
+    # Karpenter-provisioned nodes are named aks-default-* while system pool
+    # nodes are aks-nodepool1-*.  Zone labels may not be present immediately
+    # after a node reaches Ready.
+    echo "Waiting for zone labels on NAP nodes"
+    ZONE_TIMEOUT=120
+    ZONE_ELAPSED=0
+    while true; do
+        UNLABELED=$(oc get nodes --no-headers -o custom-columns='NAME:.metadata.name,ZONE:.metadata.labels.topology\.kubernetes\.io/zone' 2>/dev/null \
+            | awk '/^aks-default-/ && $2 == "<none>" { count++ } END { print count+0 }')
+        if [[ "$UNLABELED" -eq 0 ]]; then
+            echo "All NAP nodes have zone labels"
+            break
+        fi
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/aks/provision/aks-provision-commands.sh
```diff
@@ -153,5 +166,154 @@ echo "Getting kubeconfig to the AKS cluster"
...
+
+    echo "Applying Karpenter NodePool"
+    echo "$NODEPOOL_YAML" | oc apply -f -
+
+    echo "Applying Karpenter AKSNodeClass"
+    echo "$NODECLASS_YAML" | oc apply -f -
+
+    # Create placeholder pods to trigger NAP node provisioning.
+    # Resource requests are set high enough to ensure one pod per D16-equivalent node.
+    PLACEHOLDER_YAML=$(cat <<EOF
```
enxebre, line 263: Wouldn't we want to enfore topologySpreadConstraints in the placeholder deployment so we ensure we have nodes across AZs?

PR #74776, opened 2026-02-12
installer: add devpreview jobs

patrickdillon:
Adds generic devpreview jobs for aws, azure, gcp & vsphere.


tthvo:
argh, looks like we need to run `make jobs` again :D

PR #74857, opened 2026-02-13
Add upgrade workflow for kueue operator testing

sohankunkerkar:


ci-operator/step-registry/kueue-operator/test/upgrade/workload/kueue-operator-test-upgrade-workload-ref.yaml
```diff
@@ -0,0 +1,13 @@
+ref:
+  as: kueue-operator-test-upgrade-workload
```
MaysaMacedo: instead of defining new tests, perhaps we could re-use other steps and run our downstream + upstream tests? or even just downstream.

ci-operator/step-registry/kueue-operator/test/upgrade/ocp/kueue-operator-test-upgrade-ocp-commands.sh
```diff
@@ -0,0 +1,92 @@
+#!/bin/bash
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+# Minimal OCP cluster upgrade step.
+# Triggers oc adm upgrade and waits for ClusterVersion to complete.
```
MaysaMacedo: I noticed there is an existent upgrade step-registry, I wonder if maybe we could re-use it https://github.com/openshift/release/blob/main/ci-operator/step-registry/openshift/e2e/test/openshift-e2e-test-ref.yaml#L28-L31

ci-operator/step-registry/kueue-operator/test/upgrade/workload/kueue-operator-test-upgrade-workload-ref.yaml
sohankunkerkar: Yeah, that's the ultimate plan but I'm just playing around to understand the upgrade workflow with a minimal test.

ci-operator/step-registry/kueue-operator/test/upgrade/ocp/kueue-operator-test-upgrade-ocp-commands.sh
sohankunkerkar: I tried that initially but openshift-e2e-test with TEST_TYPE: upgrade runs the full OCP upgrade conformance suite (~30 disruption/health tests) and when any of those flake, the kueue workload test never gets to run. I want to keep the minimal setup.

ci-operator/config/openshift/kueue-operator/openshift-kueue-operator-release-1.3__upgrade-from-4.18.yaml
```diff
@@ -0,0 +1,87 @@
...
+      product: ocp
+      stream: nightly
+      version: "4.19"
+resources:
+  '*':
+    requests:
+      cpu: "4"
+      memory: 4Gi
+tests:
+- as: e2e-upgrade-4-18-to-4-19-kueue-1-2
```
anahas-redhat: Hi @sohankunkerkar, I guess this should be e2e-upgrade-4-18-to-4-19-kueue-1-3, right?

ci-operator/config/openshift/kueue-operator/openshift-kueue-operator-release-1.3__upgrade-from-4.19.yaml
```diff
@@ -0,0 +1,87 @@
...
+      product: ocp
+      stream: nightly
+      version: "4.20"
+resources:
+  '*':
+    requests:
+      cpu: "4"
+      memory: 4Gi
+tests:
+- as: e2e-upgrade-4-19-to-4-20-kueue-1-2
```
anahas-redhat: I guess this one would be e2e-upgrade-4-19-to-4-20-kueue-1-3

ci-operator/config/openshift/kueue-operator/openshift-kueue-operator-release-1.3__upgrade-from-4.20.yaml
```diff
@@ -0,0 +1,87 @@
...
+      product: ocp
+      stream: nightly
+      version: "4.21"
+resources:
+  '*':
+    requests:
+      cpu: "4"
+      memory: 4Gi
+tests:
+- as: e2e-upgrade-4-20-to-4-21-kueue-1-2
```
anahas-redhat: e2e-upgrade-4-20-to-4-21-kueue-1-3

ci-operator/config/openshift/kueue-operator/openshift-kueue-operator-release-1.3__upgrade-from-4.18.yaml
sohankunkerkar: whoops, yeah.. copy paste error

PR #74865, opened 2026-02-13
post-branching 5.0: Release merge blocker

jmguzik:


ci-operator/jobs/infra-periodics.yaml
```diff
@@ -1830,6 +1830,7 @@ periodics:
       - --config-dir=./ci-operator/config/
       - --current-release=4.22
       - --future-release=4.23
+      - --future-release=5.0
```
liangxia, line 1833: Two future release, will it work ?

jmguzik: yes

liangxia: Ack

PR #74892, opened 2026-02-13
Improve script quality for base ibm-fusion-access step registry steps

amp-rh:
## Summary
- Standardize bash headers to `set -eux -o pipefail; shopt -s inherit_errexit` across all base `ibm-fusion-access/` step scripts
- Replace `echo` with `: 'marker'` where `set -x` is active, replace `sleep` with `oc wait`, add `FA__*` env var prefixes, and remove error suppression patterns
- Update `shared-lib` with JUnit helper functions (`add_test_result`, `generate_junit_xml`)
- Remove dead code from `custom-ibm-must-gather`: the credential mounting (`credentials:` block in ref.yaml) and `AUTHFILE` creation logic were never passed to `oc adm must-gather`, which uses the node's global pull secret instead. The step now relies on the global pull secret patched by the `create-pull-secrets` step

**Part 1 of 5** in the IBM Fusion Access CNV CI configuration series.

Made with [Cursor](https://cursor.com)

PR #74893, opened 2026-02-13
Improve script quality for interop-tests ibm-fusion-access steps

amp-rh:
## Summary
- Apply reviewer-requested improvements to all existing `interop-tests/ibm-fusion-access/` step scripts
- Standardize bash headers (`set -eux -o pipefail; shopt -s inherit_errexit`), replace `echo` with `: 'marker'`, use `oc wait` instead of `sleep`, add `FA__*` env var prefixes, remove error suppression, and update ref.yaml documentation
- Update existing operator CI configs to match the renamed `FA__*` env vars
- Fix SC2206 shellcheck warning in create-local-disks

No new steps, deleted steps, or chain changes are included.

**Part 2 of 5** in the IBM Fusion Access CNV CI configuration series.
Depends on: #74892

## Test plan
- [ ] Verify all CI checks pass (shellcheck, ci-operator-config, registry, yamllint)
- [ ] Review script patterns match reviewer preferences (etirta)
- [ ] Confirm operator configs are consistent with step env var renames


ci-operator/config/RedHatQE/interop-testing/RedHatQE-interop-testing-master__ibm-fusion-access-operator-ocp-4.21-lp-interop-cr.yaml
```diff
@@ -24,51 +28,45 @@ resources:
...
       MAP_TESTS: "true"
       OCP_VERSION: "4.21"
       REPORT_TO_DR: "true"
-      REPORTPORTAL_CMP: Fusion-access-lp-interop
-      STORAGE_SCALE_CLIENT_CPU: "2"
-      STORAGE_SCALE_CLIENT_MEMORY: 4Gi
-      STORAGE_SCALE_CLUSTER_NAME: ibm-spectrum-scale
-      STORAGE_SCALE_STORAGE_CPU: "2"
-      STORAGE_SCALE_STORAGE_MEMORY: 8Gi
+      REPORTPORTAL_CMP: ibm-fusion-access-operator-lp-interop
```
oharan2:
```suggestion
      REPORTPORTAL_CMP: Fusion-access-lp-interop
```

ci-operator/config/RedHatQE/interop-testing/RedHatQE-interop-testing-master__ibm-fusion-access-operator-ocp-4.21-lp-interop-cr.yaml
```diff
@@ -24,51 +28,45 @@ resources:
...
+    - ref: ipi-conf-aws-byo-ipv4-pool-public
+    - ref: ipi-install-monitoringpvc
+    - chain: aws-provision-iam-user-minimal-permission
+    - ref: rhcos-conf-osstream
+    - chain: ipi-install
     test:
     - chain: interop-tests-ibm-fusion-access-environment-setup-chain
     - chain: interop-tests-ibm-fusion-access-ebs-integration-chain
     - ref: interop-tests-ibm-fusion-access-verify-cluster
+    workflow: firewatch-ipi-aws-cr
```
oharan2: The workflow doesn't work when you specify pre,post, tests. I know it from the 4.22 expansion work, 
ref: https://github.com/openshift/release/pull/75692
```suggestion
```

ci-operator/config/RedHatQE/interop-testing/RedHatQE-interop-testing-master__ibm-fusion-access-operator-ocp-4.22-lp-interop.yaml
```diff
@@ -25,37 +25,37 @@ resources:
       memory: 400Mi
 tests:
 - as: cr-fusion-access-ibm
-  cron: 0 3,15 * * *
+  cron: 0 23 31 2 *
```
oharan2: Do we disable on 4.22?

ci-operator/config/RedHatQE/interop-testing/RedHatQE-interop-testing-master__ibm-fusion-access-operator-ocp-4.21-lp-interop-cr.yaml
oharan2: Also, revert deletion of:
`    - ref: mpiit-data-router-reporter`
from post (see above)

ci-operator/step-registry/interop-tests/ibm-fusion-access/configure-aws-security-groups/interop-tests-ibm-fusion-access-configure-aws-security-groups-commands.sh
```diff
@@ -1,149 +1,61 @@
...
-CLUSTER_NAME="fusion-access-test"
-INFRA_ID=""
-
-# Try to get cluster name from shared directory
-if [[ -f "${SHARED_DIR}/CLUSTER_NAME" ]]; then
-  CLUSTER_NAME="$(<"${SHARED_DIR}/CLUSTER_NAME")"
-fi
-
-# Get infrastructure ID from metadata.json
+typeset region="${LEASED_RESOURCE:-}"
```
oharan2: I'm afraid setting an empty default is not great here

ci-operator/config/RedHatQE/interop-testing/RedHatQE-interop-testing-master__ibm-fusion-access-operator-ocp-4.21-lp-interop-cr.yaml
amp-rh: Fixed. Changed to `Fusion-access-lp-interop` on the 4.21 CR. Also proactively fixed the 4.20 CR, which had the same wrong value on main.

ci-operator/config/RedHatQE/interop-testing/RedHatQE-interop-testing-master__ibm-fusion-access-operator-ocp-4.21-lp-interop-cr.yaml
amp-rh: Fixed. Removed `workflow: firewatch-ipi-aws-cr` from the 4.21 CR. Also proactively removed `workflow:` from the 4.20 CR, 4.20 IPI, and 4.21 IPI configs that had the same pattern (expanded inline pre + workflow).

amp-rh: Fixed. Restored `mpiit-data-router-reporter` to the 4.21 CR post section. Also proactively added it to the 4.20 CR, 4.20 IPI, and 4.21 IPI configs for consistency with 4.22/4.23.

ci-operator/config/RedHatQE/interop-testing/RedHatQE-interop-testing-master__ibm-fusion-access-operator-ocp-4.22-lp-interop.yaml
amp-rh: Re-enabled the 4.22 CR cron. The disable was overly cautious; since 4.23 CR was already kept enabled with the same type of `FA__` env var renames, 4.22 should be consistent.

ci-operator/step-registry/interop-tests/ibm-fusion-access/configure-aws-security-groups/interop-tests-ibm-fusion-access-configure-aws-security-groups-commands.sh
amp-rh: Fixed. Removed the empty default. `LEASED_RESOURCE` is CI-guaranteed, so nounset should catch the missing variable directly rather than letting it pass through as an empty region.

ci-operator/step-registry/interop-tests/ibm-fusion-access/check-crds/interop-tests-ibm-fusion-access-check-crds-commands.sh
```diff
@@ -1,117 +1,96 @@
...
-  echo "❌ CRDs not established within timeout"
-  test_message="CRDs not established within 600s timeout"
+  oc get crd clusters.scale.spectrum.ibm.com -o yaml --ignore-not-found
+  testMessage="CRDs not established within 600s timeout"
 fi
 
-test_duration=$((SECONDS - test_start))
-add_test_result "test_storage_scale_crds_established" "$test_status" "$test_duration" "$test_message"
+typeset -i testDuration=$((SECONDS - testStart))
+AddTestResult "test_storage_scale_crds_established" "${testStatus}" "${testDuration}" "${testMessage}"
```
sg-rh: Will the JUnit file be always generated ? even if the oc command fail? If it makes sense maybe wrap the test in a subshell or disable -e momentarily to capture failure
```
set +e 
if oc wait --for=condition=Established crd/clusters.scale.spectrum.ibm.com --timeout=600s; then
  testStatus="passed"
  testMessage=""
else
  testStatus="failed"
  testMessage="CRDs not established within 600s timeout"
  oc get crd clusters.scale.spectrum.ibm.com -o yaml --ignore-not-found || true
fi
set -e

testDuration=$((SECONDS - testStart))
AddTestResult "test_storage_scale_crds_established" "${testStatus}" "${testDuration}" "${testMessage}"
```

ci-operator/step-registry/interop-tests/ibm-fusion-access/check-nodes/interop-tests-ibm-fusion-access-check-nodes-commands.sh
```diff
@@ -1,122 +1,110 @@
...
+typeset nodesJson=''
+nodesJson="$(oc get nodes -l node-role.kubernetes.io/worker= -o json)"
+typeset -i workerNodeCount=0
+workerNodeCount=$(printf '%s' "${nodesJson}" | jq '.items | length')
 
-if [[ $WORKER_NODE_COUNT -lt 3 ]]; then
-  echo "⚠️  WARNING: Only $WORKER_NODE_COUNT worker nodes (minimum 3 required for quorum)"
-  test_message="Insufficient worker nodes: found $WORKER_NODE_COUNT, minimum 3 required for quorum"
+if [[ "${workerNodeCount}" -lt 3 ]]; then
+  testMessage="Insufficient worker nodes: found ${workerNodeCount}, minimum 3 required for quorum"
```
sg-rh: You are using set -e (exit on error). If oc get nodes fails or returns a non-zero exit code, the script exits immediately.

The Problem: The trap will trigger `GenerateJunitXml`. However, if the script crashes before `AddTestResult` is called, your XML will show 0 tests, which might make a failing CI run look like a "success" for reports.

Suggestion: maybe move testStatus="failed" logic inside a try/catch style block or use set +e specifically around the command execution.

ci-operator/step-registry/interop-tests/ibm-fusion-access/configure-aws-security-groups/interop-tests-ibm-fusion-access-configure-aws-security-groups-commands.sh
```diff
@@ -1,149 +1,61 @@
...
+  aws ec2 describe-security-groups \
+    --region "${region}" \
+    --filters \
+      "Name=tag:sigs.k8s.io/cluster-api-provider-aws/role,Values=node" \
+      "Name=tag:Name,Values=${infraId}-*" \
+    --query 'SecurityGroups' \
+    --output json |
+  jq -cr '(first(.[]) | .GroupId) // empty'
+)"
+[[ -z "${workerSgId}" ]] && false
```
sg-rh: If `describe-security-groups` returns an empty list, the script fails here
For debugging purpose you can add
```
if [[ -z "${workerSgId}" ]]; then
  echo "ERROR: Could not find worker Security Group for infraId ${infraId}" >&2
  exit 1
fi
```

ci-operator/step-registry/interop-tests/ibm-fusion-access/configure-kmm-registry/interop-tests-ibm-fusion-access-configure-kmm-registry-commands.sh
```diff
@@ -1,258 +1,194 @@
...
+typeset test1Message=''
 
-if oc get configmap kmm-image-config -n "${FUSION_ACCESS_NAMESPACE}" >/dev/null 2>&1; then
-  echo "  ✅ kmm-image-config already exists (will update if needed)"
+if oc get configmap kmm-image-config -n "${FA__NAMESPACE}"; then
+  test1Status='passed'
 else
-  echo "  ℹ️  kmm-image-config does not exist, will create"
+  test1Status='failed'
+  test1Message='kmm-image-config ConfigMap not found (idempotency pre-check)'
```
sg-rh: If this is a clean install, the ConfigMap shouldn't exist?  should it be marked "failed then?

ci-operator/step-registry/interop-tests/ibm-fusion-access/configure-kmm-registry/interop-tests-ibm-fusion-access-configure-kmm-registry-commands.sh
```diff
@@ -1,258 +1,194 @@
...
+typeset fullRepo=''
+if [[ -n "${FA__KMM__REGISTRY_ORG}" ]]; then
+  finalRegistryUrl="${FA__KMM__REGISTRY_URL:-quay.io}"
+  fullRepo="${FA__KMM__REGISTRY_ORG}/${FA__KMM__REGISTRY_REPO}"
 else
-  # Use OpenShift internal registry
-  FINAL_REGISTRY_URL="image-registry.openshift-image-registry.svc:5000"
-  FULL_REPO="ibm-spectrum-scale/${KMM_REGISTRY_REPO}"
-  echo "  Using internal OpenShift registry: ${FINAL_REGISTRY_URL}/${FULL_REPO}"
+  finalRegistryUrl='image-registry.openshift-image-registry.svc:5000'
```
sg-rh: Is the hardcode safe here? Maybe add some debug message if not available

ci-operator/step-registry/interop-tests/ibm-fusion-access/create-cluster/interop-tests-ibm-fusion-access-create-cluster-commands.sh
```diff
@@ -1,163 +1,154 @@
...
+      --argjson quorum "$(( workerCount >= 3 ? 1 : 0 ))" \
+      '
+        .metadata.name = $name |
+        .metadata.namespace = $ns |
+        .spec.daemon.roles[0].resources = { cpu: $clientCpu, memory: $clientMem } |
+        .spec.daemon.roles[1].resources = { cpu: $storageCpu, memory: $storageMem } |
+        if $quorum == 0 then del(.spec.quorum) else . end
+      '
+  } 0<<'SKELETON' | oc apply -f -
 apiVersion: scale.spectrum.ibm.com/v1beta1
```
sg-rh: This looks bit fragile as you are piping a SKELETON that has placeholders(so maybe not valid YAML yet) and oc create expects a valid object. 

It might be better to pipe the SKELETON directly into jq, let jq fix the fields, and then pipe the resulting JSON into oc apply

ci-operator/step-registry/interop-tests/ibm-fusion-access/create-cluster/interop-tests-ibm-fusion-access-create-cluster-commands.sh
```diff
@@ -169,91 +160,87 @@ spec:
...
-TEST4_MESSAGE=""
+typeset -i test4Start=0
+test4Start=$(date +%s)
+typeset test4Status="failed"
+typeset test4Message=''
 
-DEVICE_PATH=$(oc get cluster "${STORAGE_SCALE_CLUSTER_NAME}" -n "${STORAGE_SCALE_NAMESPACE}" \
-  -o jsonpath='{.spec.daemon.nsdDevicesConfig.localDevicePaths[0].devicePath}' 2>/dev/null || echo "")
+typeset bypassDiscovery=''
+if [[ -n "${clusterJson}" ]]; then
```
sg-rh: Since you already capture `clusterJson `after the creation attempt, you can reuse that for tests 3 and 4 without re-querying. 
`oc get cluster` gets called each time

ci-operator/step-registry/interop-tests/ibm-fusion-access/create-cluster/interop-tests-ibm-fusion-access-create-cluster-commands.sh
sg-rh:
```
# Refactored creation logic
if jq \
    --arg ns "${FA__SCALE__NAMESPACE}" \
    --arg name "${FA__SCALE__CLUSTER_NAME}" \
    --arg clientCpu "${FA__SCALE__CLIENT_CPU}" \
    --arg clientMem "${FA__SCALE__CLIENT_MEMORY}" \
    --arg storageCpu "${FA__SCALE__STORAGE_CPU}" \
    --arg storageMem "${FA__SCALE__STORAGE_MEMORY}" \
    --argjson quorum "$(( workerCount >= 3 ? 1 : 0 ))" \
    '
      .metadata.name = $name |
      .metadata.namespace = $ns |
      (.spec.daemon.roles[] | select(.name == "client")).resources = {cpu: $clientCpu, memory: $clientMem} |
      (.spec.daemon.roles[] | select(.name == "storage")).resources = {cpu: $storageCpu, memory: $storageMem} |
      if $quorum == 0 then del(.spec.quorum) else . end
    ' <<'SKELETON' | oc apply -f -
apiVersion: scale.spectrum.ibm.com/v1beta1
kind: Cluster
... (rest of your skeleton) ...
SKELETON
then
  test2Status="passed"
# ...
```

ci-operator/step-registry/interop-tests/ibm-fusion-access/create-cluster/interop-tests-ibm-fusion-access-create-cluster-commands.sh
```diff
@@ -1,163 +1,154 @@
...
+      --arg name "${FA__SCALE__CLUSTER_NAME}" \
+      --arg clientCpu "${FA__SCALE__CLIENT_CPU}" \
+      --arg clientMem "${FA__SCALE__CLIENT_MEMORY}" \
+      --arg storageCpu "${FA__SCALE__STORAGE_CPU}" \
+      --arg storageMem "${FA__SCALE__STORAGE_MEMORY}" \
+      --argjson quorum "$(( workerCount >= 3 ? 1 : 0 ))" \
+      '
+        .metadata.name = $name |
+        .metadata.namespace = $ns |
+        .spec.daemon.roles[0].resources = { cpu: $clientCpu, memory: $clientMem } |
```
sg-rh:
```suggestion
        '(.spec.daemon.roles[] | select(.name == "client")).resources = {cpu: $clientCpu, memory: $clientMem}'
```

ci-operator/step-registry/interop-tests/ibm-fusion-access/create-ebs-filesystem/interop-tests-ibm-fusion-access-create-ebs-filesystem-commands.sh
```diff
@@ -1,39 +1,60 @@
...
     role: object_r
     type: container_file_t
     user: system_u
-EOF
+YAML
+  jq -c \
+    --arg ns "${FA__SCALE__NAMESPACE}" \
+    --arg disk "${FA__LOCALDISK_NAME}" \
+    '.metadata.namespace = $ns | .spec.local.pools[0].disks[0] = $disk'
+} | oc apply -f -
```
sg-rh: The | (pipe) is inside the { ... } block but after the YAML heredoc delimiter. This would likely cause the shell to fail or jq to receive empty input because the pipe is not correctly connected to the `oc create` command.
You could use a temporary variable.

ci-operator/step-registry/interop-tests/ibm-fusion-access/create-ebs-filesystem/interop-tests-ibm-fusion-access-create-ebs-filesystem-commands.sh
```diff
@@ -1,39 +1,60 @@
...
-      - shared-ebs-disk-1
-      - shared-ebs-disk-2
+      - PLACEHOLDER_DISK
     replication: 1-way
     type: shared
   seLinuxOptions:
     level: s0
     role: object_r
     type: container_file_t
     user: system_u
```
sg-rh: Since you ares setting SELinux options explicitly, pods  from different namespaces may not be able to access this storage, if thats needed incase

ci-operator/step-registry/interop-tests/ibm-fusion-access/create-ebs-filesystem/interop-tests-ibm-fusion-access-create-ebs-filesystem-commands.sh
sg-rh: Refactor pipeline to ensure JQ correctly processes the JSON
```
jq -c \
  --arg ns "${FA__SCALE__NAMESPACE}" \
  --arg disk "${FA__LOCALDISK_NAME}" \
  '.metadata.namespace = $ns | .spec.local.pools[0].disks[0] = $disk' <<'YAML' | oc apply -f -
apiVersion: scale.spectrum.ibm.com/v1beta1
kind: Filesystem
metadata:
  name: shared-filesystem
  namespace: PLACEHOLDER_NS
spec:
  local:
    blockSize: 4M
    pools:
    - name: system
      disks:
      - PLACEHOLDER_DISK
    replication: 1-way
    type: shared
YAML
```

ci-operator/step-registry/interop-tests/ibm-fusion-access/create-fusionaccess-resource/interop-tests-ibm-fusion-access-create-fusionaccess-resource-commands.sh
```diff
@@ -1,36 +1,32 @@
...
+exists="$(oc get fusionaccess fusionaccess-object -n "${faNamespace}" --ignore-not-found -o jsonpath='{.metadata.name}')"
+if [[ -z "${exists}" ]]; then
+  {
+    oc create -f - --dry-run=client -o json --save-config |
+    jq -c \
+      --arg ns "${faNamespace}" \
+      --arg ver "${faStorageScaleVersion}" \
+      '.metadata.namespace = $ns | .spec.storageScaleVersion = $ver'
+  true
+  } 0<<'YAML' | oc apply -f -
```
sg-rh: The true command might consume the stdout of the jq command, or the shell will fail because the pipe | oc apply is outside the block.


```suggestion
if [[ -z "${exists}" ]]; then
  # Direct pipeline: Heredoc -> JQ -> OC APPLY
  jq -c \
    --arg ns "${faNamespace}" \
    --arg ver "${faStorageScaleVersion}" \
    '.metadata.namespace = $ns | .spec.storageScaleVersion = $ver' <<'YAML' | oc apply -f -
apiVersion: fusion.storage.openshift.io/v1alpha1
kind: FusionAccess
metadata:
  name: fusionaccess-object
spec:
  storageDeviceDiscovery:
    create: true
YAML

  # Wait for the object to actually exist and begin reconciling
  if ! oc wait --for=condition=Ready fusionaccess/fusionaccess-object \
    -n "${faNamespace}" --timeout=600s; then
    echo "ERROR: FusionAccess object failed to reach Ready state" >&2
    oc get fusionaccess fusionaccess-object -n "${faNamespace}" -o yaml
    exit 1
  fi
fi
```

sg-rh: Also the same case for other scripts

ci-operator/step-registry/interop-tests/ibm-fusion-access/custom-ibm-must-gather/interop-tests-ibm-fusion-access-custom-ibm-must-gather-commands.sh
```diff
@@ -1,73 +1,12 @@
...
+mkdir -p /tmp/ibm-must-gather
 
-# Set default values from environment variables
-MUST_GATHER_IMAGE="${MUST_GATHER_IMAGE:-pipeline:ibm-must-gather}"
-ARTIFACT_DIR="${ARTIFACT_DIR:-/tmp/artifacts}"
+oc adm must-gather --image="${FA__MUST_GATHER_IMAGE}" --dest-dir="/tmp/ibm-must-gather"
 
-echo "Must-gather image: ${MUST_GATHER_IMAGE}"
-echo "Artifact directory: ${ARTIFACT_DIR}"
+tar -czf "${ARTIFACT_DIR}/ibm-must-gather.tar.gz" -C /tmp ibm-must-gather
```
sg-rh:
```suggestion
# Define a safe collection point
typeset gather_dir="${ARTIFACT_DIR}/must-gather-raw"
mkdir -p "${gather_dir}"

# Run must-gather directly into the artifact location
if ! oc adm must-gather --image="${FA__MUST_GATHER_IMAGE}" --dest-dir="${gather_dir}"; then
    echo "Warning: must-gather failed or partially failed." >&2
fi

# Compress and then remove the raw files to save space
tar -czf "${ARTIFACT_DIR}/ibm-must-gather.tar.gz" -C "${ARTIFACT_DIR}" must-gather-raw
rm -rf "${gather_dir}"
```

ci-operator/step-registry/interop-tests/ibm-fusion-access/patch-buildgpl/interop-tests-ibm-fusion-access-patch-buildgpl-commands.sh
```diff
@@ -1,109 +1,50 @@
...
+
+typeset -i daemonPods=0
+daemonPods=$(oc get pods -n "${FA__SCALE__NAMESPACE}" -l app.kubernetes.io/instance=ibm-spectrum-scale,app.kubernetes.io/name=core -o jsonpath-as-json='{.items[*].metadata.name}' | jq 'length')
+
+if ((daemonPods > 0)); then
+  oc delete pods -l app.kubernetes.io/instance=ibm-spectrum-scale,app.kubernetes.io/name=core \
+    -n "${FA__SCALE__NAMESPACE}" --ignore-not-found
+
+  if ! oc wait --for=condition=Ready pods -l app.kubernetes.io/instance=ibm-spectrum-scale,app.kubernetes.io/name=core \
+    -n "${FA__SCALE__NAMESPACE}" --timeout="${FA__SCALE__CORE_PODS_READY_TIMEOUT:-120}s"; then
```
sg-rh: 120s timeout for Scale Core pods is likely too short for CI, maybe use 300s

ci-operator/step-registry/interop-tests/ibm-fusion-access/prepare-worker-nodes/interop-tests-ibm-fusion-access-prepare-worker-nodes-commands.sh
```diff
@@ -1,206 +1,151 @@
...
 
-echo ""
-echo "✅ Worker node preparation completed!"
-echo "All nodes are ready for IBM Storage Scale daemon deployment"
+  typeset testDuration=0
+  testDuration=$(($(date +%s) - testStart))
+  typeset sanitizedNode=''
+  sanitizedNode="$(printf '%s' "${node}" | tr '.-' '__')"
+  AddTestResult "test_prepare_node_${sanitizedNode}" "${testStatus}" "${testDuration}" "${testMessage}"
+done < <(printf '%s' "${nodesJson}" | jq -r '.items[].metadata.name')
```
sg-rh: it spawns too many debug pods. you can use a single call per node.
```suggestion
while IFS= read -r node; do
  typeset testStart=$(date +%s)
  typeset testStatus='failed'
  typeset testMessage=''

  # Define all directories in a single string
  typeset targetDirs="/var/lib/firmware /var/mmfs/etc /var/mmfs/tmp/traces /var/mmfs/pmcollector"

  echo "Preparing node: ${node}..."
  
  # Run all mkdir commands in ONE pod execution
  if oc debug node/"${node}" --quiet -- \
     chroot /host bash -c "mkdir -p ${targetDirs} && chmod 755 /var/mmfs"; then
    testStatus='passed'
  else
    testStatus='failed'
    testMessage="Failed to create one or more directories on node ${node}"
  fi

  typeset testDuration=$(($(date +%s) - testStart))
  typeset sanitizedNode=$(printf '%s' "${node}" | tr '.-' '__')
  AddTestResult "test_prepare_node_${sanitizedNode}" "${testStatus}" "${testDuration}" "${testMessage}"

done < <(printf '%s' "${nodesJson}" | jq -r '.items[].metadata.name')
```

ci-operator/step-registry/interop-tests/ibm-fusion-access/verify-shared-storage/interop-tests-ibm-fusion-access-verify-shared-storage-commands.sh
```diff
@@ -284,132 +246,63 @@ spec:
...
+    fi
   fi
 else
-  echo "  ❌ Failed to create IBM Fusion Access data reader pod"
-  test_message="Failed to create IBM Fusion Access data reader pod resource"
+  testMessage="Failed to create IBM Fusion Access data reader pod resource"
 fi
 
-record_test "$test_start" "test_fusion_access_data_read" "$test_status" "$test_message"
+RecordTest "${testStart}" "test_fusion_access_data_read" "${testStatus}" "${testMessage}"
```
sg-rh, line 234: Is the goal to ensure that storage provisioned for CNV is shared and accessible across different workloads (CNV-style pods vs. Fusion-style pods)? 
If `test_fusion_access_data_read `expects to see `cnv-data.txt` created by the first pod, it might fail because the two pods are looking at two different disks, even if they use the same StorageClass.

I saw you create two separate PVCs below: `cnv-shared-storage-pvc` and `fusion-shared-storage-pvc`
To test "shared data," both pods must point to the same PVC name?

ci-operator/step-registry/interop-tests/ibm-fusion-access/check-crds/interop-tests-ibm-fusion-access-check-crds-commands.sh
amp-rh: The `oc wait` is already inside an `if` guard (line 86): `if oc wait --for=condition=Established ...; then ... else ... fi`. The conditional context suppresses errexit, so when `oc wait` fails, execution continues to the `else` branch. After the `if/fi`, `AddTestResult` records the outcome (using `testStatus` which defaults to `failed` unless the `then` branch sets it to `passed`). The `trap EXIT` handler generates JUnit XML regardless of how the script exits.

Using `set +e` / `set -e` around the command would actually be less safe; it disables errexit globally rather than scoping it to a single conditional.

ci-operator/step-registry/interop-tests/ibm-fusion-access/check-nodes/interop-tests-ibm-fusion-access-check-nodes-commands.sh
amp-rh: If `oc get nodes` fails at line 94, the script exits, and the `trap EXIT` handler calls `GenerateJunitXml`, which produces a JUnit file with 0 tests and 0 failures. This is the correct behavior: if the cluster API is unreachable, there are no node tests to run. CI systems and firewatch flag 0-test results as anomalies, so it would not silently pass.

ci-operator/step-registry/interop-tests/ibm-fusion-access/configure-kmm-registry/interop-tests-ibm-fusion-access-configure-kmm-registry-commands.sh
amp-rh: This is intentional. The test name is `test_kmm_config_idempotency_check`; it verifies whether a prior run already created the ConfigMap. On first run, it correctly reports "not found," which is expected and informational. Test 2 then creates the ConfigMap. The idempotency check lets us distinguish first-run from re-run in the JUnit results.

ci-operator/step-registry/interop-tests/ibm-fusion-access/configure-kmm-registry/interop-tests-ibm-fusion-access-configure-kmm-registry-commands.sh
amp-rh: `image-registry.openshift-image-registry.svc:5000` is the standard OpenShift internal image registry service URL, present in every OCP cluster. It is safe to reference directly. With xtrace active, the value is printed when the variable is used, so no additional debug message is needed.

ci-operator/step-registry/interop-tests/ibm-fusion-access/create-cluster/interop-tests-ibm-fusion-access-create-cluster-commands.sh
amp-rh: The SKELETON is valid YAML. `name: placeholder` and `namespace: placeholder` are valid string values that `oc create -f - --dry-run=client` parses without error. The `jq` stage then replaces them with the actual values.

This is the [idempotent resource creation pattern](https://edttjredhat.github.io/RHP-Wiki/Confluence/MPEXIENG/Test+Infrastructure/Best+Practises/Section0.html) from the team's best practices. The `oc create --dry-run=client` stage provides API version normalization, schema validation, and adds the `--save-config` annotation for `oc apply`. Piping the heredoc directly to `jq` would skip all of that. You can check PR [#74892](https://github.com/openshift/release/pull/74892) (the base quality PR, already merged) which uses the same pattern.

ci-operator/step-registry/interop-tests/ibm-fusion-access/create-cluster/interop-tests-ibm-fusion-access-create-cluster-commands.sh
amp-rh: Tests 3 and 4 already use the cached `clusterJson` variable (lines 216 and 232); they don't call `oc get cluster` again. There are only two `oc get cluster` calls in the script: line 105 (idempotency check before creation) and line 206 (re-fetch after creation to populate the JSON that was empty when the cluster didn't exist). Both are necessary.

ci-operator/step-registry/interop-tests/ibm-fusion-access/create-cluster/interop-tests-ibm-fusion-access-create-cluster-commands.sh
amp-rh: Same as [above](https://github.com/openshift/release/pull/74893#discussion_r2961066839): the suggested refactor removes `oc create --dry-run=client`, which provides API version normalization, schema validation, and `--save-config` annotation. The current pattern follows the team's best practices for idempotent resource creation.

ci-operator/step-registry/interop-tests/ibm-fusion-access/create-cluster/interop-tests-ibm-fusion-access-create-cluster-commands.sh
amp-rh: We control the SKELETON template (lines 182-189), so the role order is fixed: `roles[0]` is `client`, `roles[1]` is `storage`. Positional indexing is reliable here. The `select()` alternative adds verbosity without functional benefit since the SKELETON structure is defined in the same script.

ci-operator/step-registry/interop-tests/ibm-fusion-access/create-ebs-filesystem/interop-tests-ibm-fusion-access-create-ebs-filesystem-commands.sh
amp-rh: This is valid bash syntax. The heredoc `0<<'YAML'` redirects fd 0 (stdin) to `oc create -f -`. The pipe `|` on line 15 connects `oc create`'s stdout to `jq`. The outer `} | oc apply -f -` pipes the block's combined stdout to `oc apply`. The heredoc delimiter position does not interfere with the pipe chain.

This same pattern is used across all resource creation steps in the PR (create-cluster, create-fusionaccess-resource, verify-shared-storage) and in the already-merged base PR [#74892](https://github.com/openshift/release/pull/74892).

ci-operator/step-registry/interop-tests/ibm-fusion-access/create-ebs-filesystem/interop-tests-ibm-fusion-access-create-ebs-filesystem-commands.sh
amp-rh: These SELinux settings (`level: s0, role: object_r, type: container_file_t, user: system_u`) are standard OpenShift container volume labels (`system_u:object_r:container_file_t:s0`). They enable container processes to access the GPFS filesystem. The Filesystem resource is created in the `ibm-spectrum-scale` namespace; cross-namespace access is not the intent of this step.

ci-operator/step-registry/interop-tests/ibm-fusion-access/create-ebs-filesystem/interop-tests-ibm-fusion-access-create-ebs-filesystem-commands.sh
amp-rh: Same reasoning as the [create-cluster discussion](https://github.com/openshift/release/pull/74893#discussion_r2961066839): the `oc create --dry-run=client | jq | oc apply` pattern is intentional. It provides API version normalization and `--save-config` annotation that raw `jq | oc apply` does not.

ci-operator/step-registry/interop-tests/ibm-fusion-access/create-fusionaccess-resource/interop-tests-ibm-fusion-access-create-fusionaccess-resource-commands.sh
amp-rh: `true` is a shell builtin no-op; it does not read stdin, produce stdout, or interfere with pipes. It exists at the end of the block to ensure exit code 0 per the team's [best practices](https://edttjredhat.github.io/RHP-Wiki/Confluence/MPEXIENG/Test+Infrastructure/Best+Practises/Section0.html) (preventing false-negative exits from short-circuited compound statements).

The syntax `} 0<<'YAML' | oc apply -f -` is valid: the heredoc feeds fd 0 into the block, and the block's stdout (from `oc create | jq`) pipes to `oc apply`. The pipe is correctly connected.

amp-rh: Same analysis applies across all scripts using this pattern. The `{ oc create --dry-run=client | jq; true } 0<<'HEREDOC' | oc apply` pattern is valid bash and is the team's standard for idempotent resource creation.

ci-operator/step-registry/interop-tests/ibm-fusion-access/custom-ibm-must-gather/interop-tests-ibm-fusion-access-custom-ibm-must-gather-commands.sh
amp-rh: The current code already collects to `/tmp/ibm-must-gather`, compresses to `${ARTIFACT_DIR}/ibm-must-gather.tar.gz`, and the container is ephemeral (destroyed after the step completes), so cleanup is not needed. The original script on `main` was 60+ lines with unused IBM registry auth paths; this simplified version preserves the functional behavior.

If `must-gather` fails, xtrace shows the failure, and the step exits with non-zero (since `set -e` is active). For a gather step, partial collection is better than no collection, so adding `|| true` could be considered, but that is a separate design choice.

ci-operator/step-registry/interop-tests/ibm-fusion-access/patch-buildgpl/interop-tests-ibm-fusion-access-patch-buildgpl-commands.sh
amp-rh: The timeout is configurable via environment variable: `${FA__SCALE__CORE_PODS_READY_TIMEOUT:-120}s` (line 43). The 120s default can be overridden in the step's env configuration without code changes. The original script on `main` had no timeout at all.

ci-operator/step-registry/interop-tests/ibm-fusion-access/configure-aws-security-groups/interop-tests-ibm-fusion-access-configure-aws-security-groups-commands.sh
amp-rh: With `xtrace` enabled (`set -x`), the expanded `aws ec2 describe-security-groups` command line and the subsequent `[[ -z "${workerSgId}" ]] && false` test are printed to stderr, making the failure point visible in the build log.

That said, an explicit error message is a reasonable addition for clarity. Will consider for a follow-up.

ci-operator/step-registry/interop-tests/ibm-fusion-access/prepare-worker-nodes/interop-tests-ibm-fusion-access-prepare-worker-nodes-commands.sh
amp-rh: Valid point on reducing `oc debug` calls. The current pattern runs 8 `oc debug` calls per node (4 directories x mkdir + verify). Combining into one `oc debug` per node would reduce overhead.

JUnit reporting is already per-node (one test result per node); per-directory failures are captured in the message text via `failedDirs`. A combined `oc debug` could preserve this by checking each directory individually within a single chroot session. Worth considering for a follow-up optimization.

ci-operator/step-registry/interop-tests/ibm-fusion-access/verify-shared-storage/interop-tests-ibm-fusion-access-verify-shared-storage-commands.sh
amp-rh: Both PVCs use the same StorageClass (`ibm-spectrum-scale-cnv`) with `ReadWriteOnce`. The design intent is to verify that the StorageClass provisions storage accessible from both workload types.

You raise a valid point: the `test_fusion_access_data_read` test currently checks pod readiness and collects logs, but does not assert that `cnv-data.txt` (written by the first pod) is visible from the second pod's PVC. Strengthening that assertion would make the shared-storage verification more meaningful.

Whether two separate RWO PVCs on IBM Storage Scale can see each other's data depends on the CSI driver's fileset provisioning; that is what this test is intended to explore, but the assertion needs tightening to actually validate it.

amp-rh: Restructured the test to use a single `ReadWriteMany` PVC (`shared-storage-pvc`) mounted by both pods. The CNV writer pod writes `cnv-data.txt`, and the Fusion reader pod verifies it is visible via `oc exec` on the same PVC. A bidirectional check confirms the CNV pod can also read `fusion-data.txt` written by the Fusion pod. This directly validates cross-workload shared storage on IBM Storage Scale, rather than relying on two isolated PVCs.

ci-operator/step-registry/interop-tests/ibm-fusion-access/check-nodes/interop-tests-ibm-fusion-access-check-nodes-commands.sh
sg-rh: okay thanks, just want to ensure 0 test results are handled properly

ci-operator/step-registry/interop-tests/ibm-fusion-access/configure-aws-security-groups/interop-tests-ibm-fusion-access-configure-aws-security-groups-commands.sh
sg-rh: oh right, nevertheless error message will be helpful

ci-operator/step-registry/interop-tests/ibm-fusion-access/patch-buildgpl/interop-tests-ibm-fusion-access-patch-buildgpl-commands.sh
sg-rh: okay thats fine

etirta (changes requested):
@amp-rh I can see several occurrences in this PR that don't follow our Best Practices, most of which have been discussed in https://github.com/openshift/release/pull/74892/ (some items even several times).

Please use the same principles laid out in that PR and our Best Practices doc.

This PR has 53 files; we simply cannot have several rounds of review cycles discussing the same issues over and over again.

Also, please explain how these steps actually differ from #74892. I see many steps that essentially have the same name, just duplicated in different parent directories. Why all this duplication?

Thx.

ci-operator/step-registry/interop-tests/ibm-fusion-access/configure-aws-security-groups/interop-tests-ibm-fusion-access-configure-aws-security-groups-commands.sh
etirta: test

ci-operator/step-registry/interop-tests/ibm-fusion-access/configure-aws-security-groups/interop-tests-ibm-fusion-access-configure-aws-security-groups-commands.sh
etirta: With xtrace enabled it's very clear in the CI log, there is no need to clutter CI log with extra explicit error message.

Also with errexit, the [ -z "${workerSgId}" ] is enough. or even better [ -z "$(aws ec2 describe-security-groups ...)" ], remove the need to have s1x used var.

ci-operator/step-registry/interop-tests/ibm-fusion-access/create-ebs-filesystem/interop-tests-ibm-fusion-access-create-ebs-filesystem-commands.sh
etirta: @amp-rh While this is valid bash syntax, it is not a common way of writing to put the heredoc like this.

In shell, a pipeline is a group of processes, so it's only make sense to redirect the STDIN of either the first process or the whole group. The problem with redirecting the first process, we are risking having the `heredoc` positioned in the middle of the pipeline, and disrupt the natural read, which is at the end.

A more natural read is as described in our Best Practices doc. and also as discussed in #74892.

```sh
{
  oc create -f - --dry-run=client -o json --save-config |
  jq -c \
    --arg ns "${FA__SCALE__NAMESPACE}" \
    --arg disk "${FA__LOCALDISK_NAME}" \
    '.metadata.namespace = $ns | .spec.local.pools[0].disks[0] = $disk' |
  yq -p json -o yaml eval .
} 0<<'YAML' | oc apply -f -
...yamlSkeleton...
YAML
```

On the outer pipeline There is a reason why the redirection is done on the first process (which is the whole inner pipeline group). In here, the outer pipeline actually completed on the same line, therefore the `heredoc` is still positioned at last and maintain the natural read. Doing this gives us advantage for quick troubleshooting. If we want to see the final YAML for validation without actually applying it, one can comment it out with minimal disruption to the source code:
```sh
{
  oc create -f - --dry-run=client -o json --save-config |
  jq -c \
    --arg ns "${FA__SCALE__NAMESPACE}" \
    --arg disk "${FA__LOCALDISK_NAME}" \
    '.metadata.namespace = $ns | .spec.local.pools[0].disks[0] = $disk' |
  yq -p json -o yaml eval .
} 0<<'YAML' #| oc apply -f -
...yamlSkeleton...
YAML
```
and the final constructed YAML will be printed to STDOUT instead. Once confirmed, just remove the comment and turn the last pipe back on.

This troubleshooting need is also the reason why the last `yq` converting it back to YAML is used, even though it's actually not really needed by `oc` (JSON is a subset of YAML, hence it actually can read JSON input). I personally prefer have that last conversion, but I'm not enforcing it because it's less technical and more to personal view preference, even though generally speaking YAML is much easier to read than JSON.

Please use our Best Practices model, so it doesn't introduce significant style differences and confuse the reader.

I hope this makes sense.

ci-operator/step-registry/interop-tests/ibm-fusion-access/create-fusionaccess-resource/interop-tests-ibm-fusion-access-create-fusionaccess-resource-commands.sh
etirta: @amp-rh You misunderstood one of our Best Practices. That is applicable to functions or sub-shells, and the purpose is to combat false-negative exits from short-circuited **logical** compound statements. Whereas the `{ ... }` is just for grouping, without introducing any sub-context.

In this particular case, first of all, it's not a logical compound statement; it's a pipeline process group. Second, the `{ ... }` is for grouping the inner pipeline for the purpose of STDIN redirection.

Please see below:
```sh
etirtara@3ed5c610cfbc.88 2026-03-21 w12 00:36:43 "/wrk--edttjRedHat.RHP-k8s--dev/local/tmp/ToDo"
$ bash -exc 'function foo () { false && true; }; foo; echo y'; echo $?
+ foo
+ false
1

etirtara@3ed5c610cfbc.89 2026-03-21 w12 03:26:57 "/wrk--edttjRedHat.RHP-k8s--dev/local/tmp/ToDo"
$ bash -exc '( false && true; ); echo y'; echo $?
+ false
1

etirtara@3ed5c610cfbc.90 2026-03-21 w12 03:27:03 "/wrk--edttjRedHat.RHP-k8s--dev/local/tmp/ToDo"
$ bash -exc '{ false && true; }; echo y'; echo $?
+ false
+ echo y
y
0
```

So `true` in this particular case (i.e., inside `{ ... }`) serves no function at all. Please remove.

ci-operator/step-registry/interop-tests/ibm-fusion-access/patch-buildgpl/interop-tests-ibm-fusion-access-patch-buildgpl-commands.sh
etirta: Btw, if it's configured in the Step Env. Var., then the value should be guaranteed. There is no need to put the default value here; if one is needed, put it in the Step Conf. (`-ref.yaml`) file.

ci-operator/step-registry/interop-tests/ibm-fusion-access/configure-aws-security-groups/interop-tests-ibm-fusion-access-configure-aws-security-groups-commands.sh
amp-rh: Agreed on the xtrace point; the explicit error message is unnecessary.

Regarding inlining `workerSgId`: the variable is referenced in 6 places (declaration, assignment, emptiness check, writing to `SHARED_DIR`, and two `authorize-security-group-ingress` arguments), so inlining the `aws ec2 describe-security-groups` call is not practical here.

ci-operator/step-registry/interop-tests/ibm-fusion-access/create-ebs-filesystem/interop-tests-ibm-fusion-access-create-ebs-filesystem-commands.sh
amp-rh: Makes sense. Moving the heredoc redirection to the closing `}` of the brace group keeps the YAML at the tail of the construct and maintains natural top-to-bottom reading order. The troubleshooting benefit of commenting out `| oc apply` on that same line is a nice bonus.

Will restructure all occurrences across the PR to follow the BP model.

ci-operator/step-registry/interop-tests/ibm-fusion-access/create-fusionaccess-resource/interop-tests-ibm-fusion-access-create-fusionaccess-resource-commands.sh
amp-rh: Thanks for the correction and the shell examples demonstrating the difference. I misapplied the BP; `true` guards against false-negative exits in functions and subshells (which introduce their own exit-status context), but `{ }` brace groups do not create a sub-context, so the guard serves no purpose there.

Will remove.

ci-operator/step-registry/interop-tests/ibm-fusion-access/patch-buildgpl/interop-tests-ibm-fusion-access-patch-buildgpl-commands.sh
amp-rh: Good point. Since the step env var is guaranteed to be set by CI Operator, the default value belongs in the `-ref.yaml` rather than inline in the script. Will move the `default: "120"` to the ref.yaml and drop the `:-120` fallback from the script.

ci-operator/config/RedHatQE/interop-testing/RedHatQE-interop-testing-master__ibm-fusion-access-operator-ocp-4.21-lp-interop-cr.yaml
```diff
@@ -24,47 +33,41 @@ resources:
       cpu: 200m
       memory: 400Mi
 tests:
-- as: ibm-fusion-access-operator-cr-ocp421
+- as: cr-ibm-fusion-access-operator-ocp421
```
sg-rh: Do we need the full name, just making sure it doesnt voilate long name for job

ci-operator/config/RedHatQE/interop-testing/RedHatQE-interop-testing-master__ibm-fusion-access-operator-ocp-4.22-lp-interop.yaml
```diff
@@ -33,29 +38,29 @@ tests:
...
+      FA__KMM__REGISTRY_ORG: ""
+      FA__KMM__REGISTRY_REPO: gpfs-compat-kmod
+      FA__KMM__REGISTRY_URL: ""
+      FA__NAMESPACE: ibm-fusion-access
+      FA__SCALE__CLIENT_CPU: "2"
+      FA__SCALE__CLIENT_MEMORY: 4Gi
+      FA__SCALE__CLUSTER_NAME: ibm-spectrum-scale
+      FA__SCALE__STORAGE_CPU: "2"
+      FA__SCALE__STORAGE_MEMORY: 8Gi
+      FA__STORAGE_SCALE_VERSION: v5.2.3.0.1
```
sg-rh, line 53: Since there are lot of product specific configs, what do you think about having a product specific jsons to store such parameters? there is benefit and some drawbacks of both. Maybe we can discuss this separately from PR, as I see lot of other jobs have large number of config parameters

ci-operator/config/RedHatQE/interop-testing/RedHatQE-interop-testing-master__ibm-fusion-access-operator-ocp4.20-lp-interop-cr.yaml
```diff
@@ -24,48 +28,45 @@ resources:
       cpu: 200m
       memory: 400Mi
 tests:
-- as: ibm-fusion-access-operator-cr-ocp420
-  cron: 0 3,15 * * *
+- as: cr-ibm-fusion-access-operator-ocp420
```
sg-rh: Is 4.20 job config required?

ci-operator/config/RedHatQE/interop-testing/RedHatQE-interop-testing-master__ibm-fusion-access-operator-ocp-4.21-lp-interop-cr.yaml
amp-rh: The generated periodic job name is ~130 chars, within Prow limits. Matches the naming convention used by 4.22/4.23 configs on main. CI checks passed.

ci-operator/config/RedHatQE/interop-testing/RedHatQE-interop-testing-master__ibm-fusion-access-operator-ocp-4.22-lp-interop.yaml
amp-rh: Agreed, worth discussing separately. Out of scope for this PR.

ci-operator/config/RedHatQE/interop-testing/RedHatQE-interop-testing-master__ibm-fusion-access-operator-ocp4.20-lp-interop-cr.yaml
amp-rh: These configs already exist on main. This PR only renames the env vars from `STORAGE_SCALE_*` / `KMM_*` to `FA__*` for consistency with the step changes. The cron is disabled (`0 23 31 2 *`).

ci-operator/step-registry/interop-tests/ibm-fusion-access/configure-aws-security-groups/interop-tests-ibm-fusion-access-configure-aws-security-groups-commands.sh
etirta: I see... You are right; since it is used several times, avoiding inlining and using variables is indeed the correct approach.

However, please remove the `&& false` as it's redundant and flip the test logic instead.

```suggestion
[ -n "${workerSgId}" ]
```

ci-operator/step-registry/interop-tests/ibm-fusion-access/configure-aws-security-groups/interop-tests-ibm-fusion-access-configure-aws-security-groups-commands.sh
```diff
@@ -1,149 +1,61 @@
...
+    --source-group "${workerSgId}" \
+    --group-owner "${accountId}" \
+    2>&1 1>&3; } 3>&2); then
+    if [[ "${stderr}" == *'InvalidPermission.Duplicate'* ]]; then
+      true
     else
-      echo "  ❌ Failed to add: ${description}"
+      printf '%s\n' "${stderr}" >&2
       return 1
     fi
```
etirta: Nits & Tricks:
If you switch the logic, the code will be more concise. A smaller footprint leads to higher maintainability.

```suggestion
    if [[ "${stderr}" != *'InvalidPermission.Duplicate'* ]]; then
        printf '%s\n' "${stderr}" 1>&2
        return 1
    fi
```

ci-operator/step-registry/interop-tests/ibm-fusion-access/patch-buildgpl/interop-tests-ibm-fusion-access-patch-buildgpl-commands.sh
```diff
@@ -1,109 +1,50 @@
 #!/bin/bash
-set -o nounset
-set -o errexit
-set -o pipefail
+set -eux -o pipefail; shopt -s inherit_errexit
 
-STORAGE_SCALE_NAMESPACE="${STORAGE_SCALE_NAMESPACE:-ibm-spectrum-scale}"
+typeset FA__SCALE__NAMESPACE="${FA__SCALE__NAMESPACE:-ibm-spectrum-scale}"
+typeset FA__NAMESPACE="${FA__NAMESPACE:-ibm-fusion-access}"
```
etirta: You need to improve your review handling. I can't possibly keep reviewing for similar issues over and over again. These env. var.s are defined in Step Conf. so CI Operator guarantees their settings. There is no need to re-declare and assign default value in here.
```suggestion
```
There are several occurrences of this similar, unnecessary re-declaration and default value assignment throughout multiple files touched by this PR.

ci-operator/config/RedHatQE/interop-testing/RedHatQE-interop-testing-master__ibm-fusion-access-operator-ocp-4.22-lp-interop.yaml
etirta: Since the CUT (Cluster Under Test) is on AWS, we should have `aws` as to set CR Platform correctly. The `fusion-access` string unfortunately required by current CR setting to set CR Component (LP) correctly.
```suggestion
- as: cr-fusion-access-aws
```
Please update `.config.prowgen` accordingly.

ci-operator/config/RedHatQE/interop-testing/RedHatQE-interop-testing-master__ibm-fusion-access-operator-ocp-4.21-lp-interop-cr.yaml
etirta: Why not follow the new naming in 4.22? Even though this job is disabled, I believe the reason it's not deleted is so we can turn it on if there is a need (since it's just GA-1). If we do turn it on, don't we want this to be picked up by the CR? If so, then we need to follow the agreed Test Name convention as in 4.22 onwards.

ci-operator/step-registry/interop-tests/ibm-fusion-access/create-ebs-filesystem/interop-tests-ibm-fusion-access-create-ebs-filesystem-commands.sh
```diff
@@ -1,39 +1,60 @@
...
-EOF
+YAML
 
-echo "✅ Filesystem resource created"
-oc get filesystem shared-filesystem -n ${STORAGE_SCALE_NAMESPACE}
+if ! oc wait --for=condition=Success \
+    filesystem/shared-filesystem -n "${FA__SCALE__NAMESPACE}" --timeout="${FA__FILESYSTEM_TIMEOUT}s"; then
```
etirta: In the previous PR, we discussed this issue and suggested that `FA__FILESYSTEM_TIMEOUT` should contain the time unit, so the user can actually specify, say `10m`, if desired.

This is a perfect example of inefficiency when duplicating Steps instead of reusing them.

ci-operator/step-registry/ibm-fusion-access/install-fusion-access-operator/ibm-fusion-access-install-fusion-access-operator-commands.sh
etirta, line 1: Our Best Practices were not followed. Proper data marshaling is required.

ci-operator/step-registry/interop-tests/ibm-fusion-access/check-crds/interop-tests-ibm-fusion-access-check-crds-commands.sh
```diff
@@ -1,117 +1,96 @@
...
+}
 
-echo "🔍 Waiting for IBM Storage Scale CRDs..."
+trap '{( GenerateJunitXml; true )}' EXIT
 
 # Test 1: Wait for CRDs to be established
-test_start=$SECONDS
-test_status="failed"
-test_message=""
+typeset -i testStart="${SECONDS}"
```
etirta: Avoid using a variable only once unless there is a good reason. Reducing the number of variables reduces runtime memory consumption and helps the reader.
```suggestion
typeset -i testDuration="${SECONDS}"
 ....
testDuration=$((SECONDS - testDuration))
```
There are several other occurrence throughout the PR.

ci-operator/step-registry/interop-tests/ibm-fusion-access/configure-aws-security-groups/interop-tests-ibm-fusion-access-configure-aws-security-groups-commands.sh
```diff
@@ -1,149 +1,61 @@
...
+typeset region="${LEASED_RESOURCE}"
+typeset infraId=''
 if [[ -f "${SHARED_DIR}/metadata.json" ]]; then
-  INFRA_ID=$(jq -r '.infraID' "${SHARED_DIR}/metadata.json")
-  echo "Infrastructure ID: ${INFRA_ID}"
+  infraId="$(jq -r '.infraID' "${SHARED_DIR}/metadata.json")"
 else
-  echo "ERROR: metadata.json not found"
   exit 1
 fi
```
etirta: Nits & Tricks:
Since we have `errexit` enabled, we can refactor these constructs into a more concise form.
```suggestion
[ -f "${SHARED_DIR}/metadata.json" ]
infraId="$(jq -r '.infraID' "${SHARED_DIR}/metadata.json")"
```
This `infraId` is actually used only once, so it's a good candidate for inlining. Up to you whether you want to inline this or not.

ci-operator/step-registry/interop-tests/ibm-fusion-access/create-cluster/interop-tests-ibm-fusion-access-create-cluster-commands.sh
amp-rh: @sg-rh Resolving per the explanation above (#74892 / BP). Ping if you still want a different approach.

ci-operator/step-registry/interop-tests/ibm-fusion-access/create-cluster/interop-tests-ibm-fusion-access-create-cluster-commands.sh
amp-rh: @sg-rh Same as above: no duplicate `oc get` in tests 3–4. Resolving; thanks for the careful read.

ci-operator/step-registry/interop-tests/ibm-fusion-access/create-cluster/interop-tests-ibm-fusion-access-create-cluster-commands.sh
amp-rh: @sg-rh Keeping positional `roles[0]` / `roles[1]` here since the heredoc is fixed in-repo; `select()` does not buy us anything unless we expect reordering. Resolving; thanks.

ci-operator/step-registry/interop-tests/ibm-fusion-access/create-ebs-filesystem/interop-tests-ibm-fusion-access-create-ebs-filesystem-commands.sh
amp-rh: @sg-rh @etirta The step uses the BP layout: `{ oc create --dry-run=client | jq | yq } 0<<'YAML' | oc apply -f -` with the heredoc on the closing brace line. On the current branch it lives under `ci-operator/step-registry/ibm-fusion-access/create-ebs-filesystem/ibm-fusion-access-create-ebs-filesystem-commands.sh` (the `interop-tests/.../create-ebs-filesystem` path in the review UI matches an earlier tree layout). Resolving; thanks.

ci-operator/step-registry/interop-tests/ibm-fusion-access/create-ebs-filesystem/interop-tests-ibm-fusion-access-create-ebs-filesystem-commands.sh
amp-rh: @sg-rh Thanks for flagging this. For this interop workflow we only need the shared filesystem consumed through the Fusion Access and CNV paths that are already wired in the chain, not arbitrary mounts from unrelated namespaces. The explicit `seLinuxOptions` match the IBM Storage Scale Filesystem CR sample for shared `container_file_t` content. If we later need broader cross-namespace access from workloads outside that setup, we can drop the explicit block and rely on operator defaults or adjust labels in a follow-up.

ci-operator/step-registry/interop-tests/ibm-fusion-access/configure-aws-security-groups/interop-tests-ibm-fusion-access-configure-aws-security-groups-commands.sh
amp-rh: @etirta Agreed, the inverted condition is clearer. Applying your suggestion and validating the shell.

ci-operator/step-registry/interop-tests/ibm-fusion-access/patch-buildgpl/interop-tests-ibm-fusion-access-patch-buildgpl-commands.sh
amp-rh: @etirta Agreed. I removed redundant env re-declarations in the commands and rely on the step `ref` defaults everywhere we touched for this PR. Thanks.

ci-operator/config/RedHatQE/interop-testing/RedHatQE-interop-testing-master__ibm-fusion-access-operator-ocp-4.22-lp-interop.yaml
amp-rh: @etirta Confirmed `cr-fusion-access-aws` in the 4.22 config. Prowgen is updated for the new variants and `cr-fusion-access-aws` Slack reporting; jobs regenerated with `make jobs`.

ci-operator/step-registry/interop-tests/ibm-fusion-access/create-ebs-filesystem/interop-tests-ibm-fusion-access-create-ebs-filesystem-commands.sh
amp-rh: @etirta On current head that interop file path is no longer present; we use one `ibm-fusion-access-create-ebs-filesystem` step. `FA__SCALE__FILESYSTEM_READY_TIMEOUT` is documented as a Kubernetes duration with default `1h` and is passed to `oc wait --timeout`, so `10m` and similar overrides are valid.

ci-operator/step-registry/ibm-fusion-access/install-fusion-access-operator/ibm-fusion-access-install-fusion-access-operator-commands.sh
amp-rh: @etirta The install step now uses `oc create … --dry-run=client -o json` with `jq` (`--arg` for namespace, catalog image, channel, and `storageScaleVersion`) before `oc apply` for OperatorGroup, CatalogSource, Subscription, and FusionAccess. If you had a different resource or pattern in mind, point me at it and I will align it.

ci-operator/step-registry/interop-tests/ibm-fusion-access/check-crds/interop-tests-ibm-fusion-access-check-crds-commands.sh
amp-rh: @etirta On current head the interop `check-crds` script is gone; `ibm-fusion-access-check-crds-commands.sh` already uses the `testDuration` / `SECONDS` elapsed pattern you described for the CRD wait.

ci-operator/step-registry/interop-tests/ibm-fusion-access/configure-aws-security-groups/interop-tests-ibm-fusion-access-configure-aws-security-groups-commands.sh
amp-rh: @etirta Thanks for the nit. The step now lives at `ci-operator/step-registry/ibm-fusion-access/configure-aws-security-groups/ibm-fusion-access-configure-aws-security-groups-commands.sh` (the `interop-tests/...` path in the review is from an earlier revision). I switched the metadata read to the `[ -f "${SHARED_DIR}/metadata.json" ]` guard plus `infraId="$(jq -r '.infraID' ...)"` form you suggested. I kept `infraId` as a variable since it is also used in the worker SG name filter, not only in the jq line.

amp-rh:
Closing -- Fusion Access Operator interop testing will not continue for 4.22, so this work is no longer needed.

PR #74917, opened 2026-02-16
[draft] DNM: aws gpu cluster up

swghosh:
duplicate of #71049


empovit:
/hold
I'm not sure what the purpose of this PR is, and why we need this in the NVIDIA GPU operator CI. It also seems to call a third-party website, which I find suspicions.

Also, if the purpose is to track AWS clusters, it should be done via the test platform and in the reusable cluster install/uninstall workflows.

PR #74927, opened 2026-02-16
WIP: OCPNODE-4067: Add rhcos10 + runc job

bitoku:


bitoku:
we don't need this anymore because rhcos10 doesn't support runc.

PR #74954, opened 2026-02-16
aro-hcp: add a write-config pre step

mmazur:
Puts a rendered config.yaml into $SHARED_DIR. This way every step always has access to at least some of the config values of any environment.

### DIFF
--- a/ci-operator/step-registry/aro-hcp/write-config/aro-hcp-write-config-commands.sh
+++ b/ci-operator/step-registry/aro-hcp/write-config/aro-hcp-write-config-commands.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+set -o errexit
+set -o nounset
+set -o pipefail
+
+if [[ -n "${MULTISTAGE_PARAM_OVERRIDE_LOCATION:-}" ]]; then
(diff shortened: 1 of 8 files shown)


ci-operator/step-registry/aro-hcp/write-config/aro-hcp-write-config-commands.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+set -o errexit
+set -o nounset
+set -o pipefail
+
+if [[ -n "${MULTISTAGE_PARAM_OVERRIDE_LOCATION:-}" ]]; then
+  export LOCATION="${MULTISTAGE_PARAM_OVERRIDE_LOCATION}"
+fi
```
roivaz, line 8: what's this for?

mmazur: Afaik this is for prod jobs triggered from ev2, they take their actual location from that MULTISTAGE param.

mmazur:
/pj-rehearse periodic-ci-Azure-ARO-HCP-main-periodic-stage-e2e-parallel
/pj-rehearse periodic-ci-Azure-ARO-HCP-main-periodic-integration-e2e-parallel

PR #74985, opened 2026-02-17
Fix orion UDN config mismatch: add L2/L3 dynamic selection and missing VERSION env var

mohit-sheth:
## Summary

  - Rename `openshift-qe-orion-udn-l3` step to `openshift-qe-orion-udn-density` to support both L2 and
   L3 UDN modes
  - Dynamically select Orion config based on worker node count (`large-scale`, `med-scale`,
  `small-scale`, `trt-external-payload`) and `ENABLE_LAYER_3` env var
  - Use a common `udn_ack.yaml` ACK file for both L2 and L3 modes
  - Default `ENABLE_LAYER_3` to `"false"` (L2 mode) in the udn-density-pods ref

  ## Details

  The udn-density-pods chain runs UDN density tests that can operate in L2 or L3 mode, but the Orion
  change detection step was hardcoded to `udn-l3` configs. This caused a mismatch when control-plane
  tests ran with `ENABLE_LAYER_3=false`.

  This PR fixes the mismatch by:

  1. **Renaming** the step from `udn-l3` to `udn-density` (generic, supports both modes)
  2. **Scale-aware config selection** — determines the Orion config prefix (`small-scale`,
  `med-scale`, `large-scale`, `trt-external-payload`) based on the actual worker node count at runtime
  3. **L2/L3 config selection** — appends `-udn-l2.yaml` or `-udn-l3.yaml` based on `ENABLE_LAYER_3`
  5. **Unifying the ACK file** — uses a single `udn_ack.yaml` for both L2 and L3
  
  depends on https://github.com/cloud-bulldozer/orion/pull/285 

  🤖 Generated with [Claude Code](https://claude.ai/claude-code)


ci-operator/step-registry/openshift-qe/orion/udn-density/openshift-qe-orion-udn-density-commands.sh
chentex, line 1: Can you add this to the script:
https://github.com/openshift/release/blob/main/ci-operator/step-registry/openshift-qe/orion/openshift-qe-orion-commands.sh#L97-L98
https://github.com/openshift/release/blob/main/ci-operator/step-registry/openshift-qe/orion/openshift-qe-orion-commands.sh#L63-L70

PR #75009, opened 2026-02-17
Cleaning up create-prowjob

tbuskey:
Cleanup some issues in **create-prowjob**

- `EXPECTED_OSC_VERSION` was not handled properly
  - If set, use it.  Do not set it to 0.0.0
  - If not set, derive from X.Y.Z catalog tag
  - for other tags, use the `DEFAULT_EXPECTED_OSC_VERSION` at the top of the script
- Modify the **create-prowjob** _run_ to only launch 1 cluster at a time
  - gangway has a rate limit, this will avoid hitting it
- Add directions to create-prowjob run
  - use **create-prowjob** _create_ to produce a modified yaml
  - Finding the job in spyglass

### DIFF
--- a/ci-operator/step-registry/sandboxed-containers-operator/create-prowjob/README.md
+++ b/ci-operator/step-registry/sandboxed-containers-operator/create-prowjob/README.md
@@ -66,7 +66,6 @@ PROW_API_TOKEN=your_token_here ci-operator/step-registry/sandboxed-containers-op
 | `AWS_REGION_OVERRIDE`      | `us-east-2`              | AWS region for testing                                                      | Any valid AWS region     |
 | `CUSTOM_AZURE_REGION`      | `eastus`                 | Azure region for testing                                                    | Any valid Azure region   |
 | `OSC_CATALOG_TAG`          | `latest`                 | Defaults to `:latest`. Actual tag resolved at runtime by `env-cm` step. Can override with specific version tag (e.g., `1.11.1-1766149846`) or SHA | repo tag or SHA          |
-| `EXPECTED_OSC_VERSION`     | `1.10.1`                 | Derived from X.Y.X-epoch_time catalog tag or OSC_CATALOG_TAG                | Semantic version format  |
 | `INSTALL_KATA_RPM`         | `true`                   | Whether to install Kata RPM                                                 | `true` or `false`        |
 | `KATA_RPM_VERSION`         | `3.17.0-3.rhaos4.19.el9` | Kata RPM version (when `INSTALL_KATA_RPM=true`)                             | RPM version format       |
 | `PROW_RUN_TYPE`            | `candidate`              | Prow job run type                                                           | `candidate` or `release` |
@@ -81,7 +80,6 @@ PROW_API_TOKEN=your_token_here ci-operator/step-registry/sandboxed-containers-op
   - The `env-cm` step resolves the actual latest tag (X.Y.Z-epoch_time format) at runtime
   - This ensures jobs always test against the most recent build
 - Creates `brew-catalog` source with the resolved catalog tag
-  - If catalog tag is X.Y.Z-epoch_time, the expected version of the operator is extracted
 
 #### GA (Production) Mode
 - Uses `redhat-operators` catalog source with GA images
--- a/ci-operator/step-registry/sandboxed-containers-operator/create-prowjob/sandboxed-containers-operator-create-prowjob-commands.sh
+++ b/ci-operator/step-registry/sandboxed-containers-operator/create-prowjob/sandboxed-containers-operator-create-prowjob-commands.sh
@@ -15,21 +15,7 @@ set -o pipefail
 # used to interact with Prow via REST API
 GANGWAY_API_ENDPOINT="https://gangway-ci.apps.ci.l2s4.p1.openshiftapps.com/v1/executions"
 ARO_CLUSTER_VERSION="${ARO_CLUSTER_VERSION:-4.17}"
-DEFAULT_EXPECTED_OSC_VERSION="1.11.2"
 
-get_expected_version() {
-    # Extract expected version from catalog tag
-    # If catalog tag is in X.Y.Z-epoch format (e.g. 1.10.1-1766149846), returns X.Y.Z portion
-    # If input is "latest", returns empty string (caller should use EXPECTED_OSC_VERSION)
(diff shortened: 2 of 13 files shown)


ci-operator/step-registry/sandboxed-containers-operator/create-prowjob/sandboxed-containers-operator-create-prowjob-commands.sh
```diff
@@ -15,18 +15,17 @@ set -o pipefail
 # used to interact with Prow via REST API
 GANGWAY_API_ENDPOINT="https://gangway-ci.apps.ci.l2s4.p1.openshiftapps.com/v1/executions"
 ARO_CLUSTER_VERSION="${ARO_CLUSTER_VERSION:-4.17}"
+DEFAULT_EXPECTED_OSC_VERSION="1.11.2"
```
ldoktor: Do you have any motivation for hardcoding a default one? TBH I'd rather fail if it fails to retrieve it and is not set as this version will always differ and there is no "sensible" default.

ldoktor: .. or empty value to ignore the expected osc version test...

ci-operator/step-registry/sandboxed-containers-operator/create-prowjob/README.md
```diff
@@ -170,6 +170,17 @@ The `run` command triggers a single ProwJob from a generated YAML configuration
...
+### Create Yaml for Run
+Use the **create** command to create a yaml file in the current directory.  Modify it for your case.  Follow Option A in the output directions
+
+### Viewing the Run in Spyglass
+Go to [Prow configured jobs](https://prow.ci.openshift.org/configured-jobs/)
+Scroll down to *sandboxed-containers-operator* and click on it
+Search for the prow job you specified (ex aws-ipi-peerpods) and click on _Details_
+Click on _History_
+You will be taken to a list of the **Build** numbers, etc.  Your job should be at the top.  Clicking on that will show you the Spyglass of your job with the build log, artifacts, etc.
+This URL is used by **dig&shift** for reporting and analysis
```
ldoktor: Do you think it'd make sense to put here one sample url (as users might simply get the url from the name without all the clicking?)

ci-operator/step-registry/sandboxed-containers-operator/create-prowjob/sandboxed-containers-operator-create-prowjob-commands.sh
tbuskey: It was being set in multiple places.  This puts it in one place.
The user is supposed to set EXPECTED_OSC_VERSION, but it can be derived from an X.Y.Z tag.  Otherwise, it goes to this default.  The test has caught out the code not updating the version more than once.

ci-operator/step-registry/sandboxed-containers-operator/create-prowjob/README.md
tbuskey: Done

ci-operator/step-registry/sandboxed-containers-operator/create-prowjob/sandboxed-containers-operator-create-prowjob-commands.sh
ldoktor: By set you mean in workflow scripts? If so then this commit should remove the defaults from there and always expect it to be present (otherwise we risk dead code that might hit us when we change this code in the future).

I guess the main question is, whether it always needs to be set. If not than the DEFAULT_EXPECTED_OSC_VERSION doesn't need to exist (and IMO shouldn't). If it needs to be set, then we should discuss the default value as 1.11.2 is an existing version, which might mask us not setting it when we should have set it and it might break once 1.11.2 is not default (or existing). I'd really suggest using `0.0.0` or `undefined` or whatever to emphasize it was not configured. (or raise an exception that it needs to be set)

wainersm: I understand @ldoktor 's concern. The `DEFAULT_EXPECTED_OSC_VERSION` will inevitable be outdated because this script and repository is not visible. I mean, Julien or Greg usually are in charge of preparing the sandboxed-containers-operator repository for the upcoming release, there they will update many files with the new version (e.g. 1.12.0). This repository and script is not on their radar. That's why we are passing the expected operator version and rpm from the konflux pipeline (https://github.com/openshift/sandboxed-containers-operator/blob/devel/tests/osc-test-fbc-integration.yaml#L149); Greg/Julien/Daniel will update the  osc-test-fbc-integration.yaml whenever needed.


My proposal is to move that script to sandboxed-containers-operator repo so that it gets updated alongside the other files of the operator when a version bump is required. We merge this PR and then open a new one on that repo. What do you think?

ci-operator/step-registry/sandboxed-containers-operator/create-prowjob/README.md
```diff
@@ -161,25 +161,22 @@ The `create-prowjob` script uses `:latest` as the default tag. The actual latest
 
 ## Run Command
 
-The `run` command allows you to trigger ProwJobs from a generated YAML configuration file. This command requires a valid Prow API token.
+The `run` command triggers a single ProwJob from a generated YAML configuration file. You must specify exactly one job name (the `as` value from the tests in the YAML, e.g. `azure-ipi-kata`, `aws-ipi-peerpods`). This command requires a valid Prow API token.
```
wainersm, line 164: Just a general comment about LLM generated code comments: it usually repeats parts of the prompt all over the code. For example, here in the README and the script itself you gonna see it repeating "exactly one job name" all over! When I have time and energy, I go manually editing the comments generated by AI.

ci-operator/step-registry/sandboxed-containers-operator/create-prowjob/sandboxed-containers-operator-create-prowjob-commands.sh
ldoktor: TBH I'd prefer requiring this argument rather than any kind of default, be it sensible or doomed-to-be-outdated. Adding a code to check if it was set to what it was set in the same repo seems odd to me.

PR #75020, opened 2026-02-18
destroy: skip deprovisioning verification step if metadata.json is missing

tthvo:
If `metadata.json` file is missing, the job likely failed elsewhere and the install was never ran. Thus, there would be no resources to delete or verify.


tthvo:
/pj-rehearse ack

I can't get the job to fail before install, yet. But at least, it didn't break where verification needs running.

PR #75026, opened 2026-02-18
change patch site config to merge

eifrach:


eifrach:
/retitle TELCOV10N-976: change patch site config to merge

PR #75049, opened 2026-02-18
Break the e2e-ocl tests into 2 jobs

umohnani8:
Add two new jobs for the sharded e2e-ocl test suite.

This adds jobs for https://github.com/openshift/machine-config-operator/pull/5669 that breaks up the test suite. Once this is merged and tests are stable, we will open a PR to remove the original test job and mark these two as required instead.


isabella-janssen (commented):
/lgtm

PR #75085, opened 2026-02-19
New jobs to run ARO-HCP tests against OCP candidate builds

miquelsi:
These jobs will run the ARO-HCP E2E test suite using the latest candidate build of OCP in the Y stream. Currently: 4.19, 4.20 and 4.21.

Thes job are scheduled to run once a week.

### DIFF
--- a/ci-operator/step-registry/aro-hcp/test/persistent/aro-hcp-test-persistent-commands.sh
+++ b/ci-operator/step-registry/aro-hcp/test/persistent/aro-hcp-test-persistent-commands.sh
@@ -18,4 +18,24 @@ if [[ -n "${MULTISTAGE_PARAM_OVERRIDE_LOCATION:-}" ]]; then
   export LOCATION="${MULTISTAGE_PARAM_OVERRIDE_LOCATION}"
 fi
 
+if [[ "${ARO_HCP_OPENSHIFT_CHANNEL_GROUP:-}"  =~ "candidate-" ]]; then
+  candidate_version="$(
+    curl -s "https://api.openshift.com/api/upgrades_info/v1/graph?channel=${ARO_HCP_OPENSHIFT_CHANNEL_GROUP}" \
+      | jq -r '.nodes[].version' \
+      | sort -V \
+      | tail -n1
+  )"
+
+  if [[ -z "${candidate_version}" || "${candidate_version}" == "null" ]]; then
+    echo "Failed to resolve ${ARO_HCP_OPENSHIFT_CHANNEL_GROUP} version from upgrades info graph." >&2
+    exit 1
+  fi
+
+  export ARO_HCP_OPENSHIFT_CONTROLPLANE_VERSION="${candidate_version}"
+  export ARO_HCP_OPENSHIFT_NODEPOOL_VERSION="${candidate_version}"
+  # Channel group name is just "candidate", with no version number, so we need to set it back to "candidate"
+  export ARO_HCP_OPENSHIFT_CHANNEL_GROUP="candidate"
+  export ARO_HCP_OPENSHIFT_NODEPOOL_CHANNEL_GROUP="candidate"
+fi
+
 ./test/aro-hcp-tests run-suite "${ARO_HCP_SUITE_NAME}" --junit-path="${ARTIFACT_DIR}/junit.xml" --html-path="${ARTIFACT_DIR}/extension-test-result-summary.html" --max-concurrency 100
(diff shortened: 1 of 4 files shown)


ci-operator/step-registry/aro-hcp/test/persistent/aro-hcp-test-persistent-commands.sh
```diff
@@ -16,4 +16,21 @@ if [[ -n "${MULTISTAGE_PARAM_OVERRIDE_LOCATION:-}" ]]; then
...
+      | tail -n1
+  )"
+
+  if [[ -z "${candidate_version}" || "${candidate_version}" == "null" ]]; then
+    echo "Failed to resolve candidate-4.21 version from upgrades info graph." >&2
+    exit 1
+  fi
+
+  export ARO_HCP_OPENSHIFT_CONTROLPLANE_VERSION="${candidate_version}"
+  export ARO_HCP_OPENSHIFT_NODEPOOL_VERSION="${candidate_version}"
```
mbukatov, line 35: These two values are expected to be different: control plane version is something like 4.21 while node pool version could be 4.21.11. Note that candidate build versions can look bit different (eg.  [4.22.0-ec.2](https://openshift-release.apps.ci.l2s4.p1.openshiftapps.com/releasestream/4-dev-preview/release/4.22.0-ec.2)).

ci-operator/step-registry/aro-hcp/test/persistent/aro-hcp-test-persistent-commands.sh
```diff
@@ -16,4 +16,21 @@ if [[ -n "${MULTISTAGE_PARAM_OVERRIDE_LOCATION:-}" ]]; then
   export LOCATION="${MULTISTAGE_PARAM_OVERRIDE_LOCATION}"
 fi
 
+if [[ "${CANDIDATE_TESTING:-}" == "true" ]]; then
+  candidate_version="$(
+    curl -s "https://api.openshift.com/api/upgrades_info/v1/graph?channel=candidate-4.21" \
```
mbukatov: We need to make sure that ARO HCP is fetching the candidate build version in the same way.

mbukatov (commented):
Comments about version handling.

miquelsi:
/pj-rehearse periodic-ci-Azure-ARO-HCP-main-periodic-stage-e2e-parallel-ocp-candidate

miquelsi:
/pj-rehearse periodic-ci-Azure-ARO-HCP-main-periodic-stage-e2e-parallel-ocp-candidate

ci-operator/step-registry/aro-hcp/test/persistent/aro-hcp-test-persistent-commands.sh
ahitacat: For candidate, and under the AFEC flag. Cluster version can (and should) be specified with X.Y.Z
I think we are OK to use the same version for this test, in the future it could be interesting to have different versions. i.e. testing latest X.Y for clusters and X.Y-1 and X.Y-2 latest z for nodepools

ci-operator/step-registry/aro-hcp/test/persistent/aro-hcp-test-persistent-commands.sh
```diff
@@ -16,4 +16,21 @@ if [[ -n "${MULTISTAGE_PARAM_OVERRIDE_LOCATION:-}" ]]; then
   export LOCATION="${MULTISTAGE_PARAM_OVERRIDE_LOCATION}"
 fi
 
+if [[ "${ARO_HCP_OPENSHIFT_CHANNEL_GROUP:-}" == "candidate" ]]; then
+  candidate_version="$(
+    curl -s "https://api.openshift.com/api/upgrades_info/v1/graph?channel=candidate-4.21" \
```
ahitacat: We are only testing 4.21. But, I think it could be also good to test 4.19 latest, and the near future the 4.22 (so we test -rc as soon as it arrives)
With 4.19 we will cover offering the latest for 4.19 and 4.21 in the stable channels that has been tested.
With 4.22 we will test early the next minor.

deads2k: @miquelsi suggested something neat to me.  What if we have a candidate job and the test execution for that job installs and runs the latest 4.y for all candidate channels?  That keeps the number of jobs relatively small seems straight-forward to spell in the e2e testing we already have.

miquelsi: As the original idea was to run the whole E2E test suite (not just cluster creation), we'll be adding a new job that will run E2E tests for each Y stream.

miquelsi:
/pj-rehearse periodic-ci-Azure-ARO-HCP-main-periodic-stage-e2e-parallel-ocp-candidate-4-19

ci-operator/step-registry/aro-hcp/test/persistent/aro-hcp-test-persistent-commands.sh
mbukatov: Ok, makes sense.

PR #74741, opened 2026-02-11
ARO-HCP Create e2e tools image promotion

patriksuba:
[ARO-23898](https://issues.redhat.com/browse/ARO-23898)

- PR include skip of presubmit job for baseimage-generator in case related Dockerfile is unchanged [ARO-23901](https://issues.redhat.com/browse/ARO-23901)

PR #75093, opened 2026-02-19
ARO-HCP periodic: switch to prebuilt aro-hcp-e2e-tools image

patriksuba:
[ARO-23766](https://issues.redhat.com/browse/ARO-23766)

Switching to the prebuilt image in periodic jobs after #74741 is merged.


ci-operator/config/Azure/ARO-HCP/Azure-ARO-HCP-main__periodic.yaml
```diff
@@ -1,30 +1,8 @@
-binary_build_commands: GO_COMPLIANCE_INFO=0 make build
-build_root:
-  image_stream_tag:
-    name: aro-hcp-ci-images
+base_images:
+  aro-hcp-e2e-tools:
+    name: aro-hcp-e2e-tools
```
venkateshsredhat, line 3: Didn't you change name to something else ?

patriksuba: The name stays the same as before.

venkateshsredhat: why? 

we already have this name "  to: aro-hcp-e2e-tools" for ourpresubmit images . I had this comment in first PR and you addressed it  https://github.com/openshift/release/pull/74741#discussion_r2822842279 . But I don't see it now .

venkateshsredhat: aro-hcp-e2e-tools is misleading to me ... ideally the image you are promoting is just the one with e2e binary .
Since we use the same name in presubmit it is even more confusing .

patriksuba: I would keep the name same. The content is combined at the moment with multiple tools and e2e test binary. I would like to revisit Dockerfile for e2e tests in [ARO-23919](https://issues.redhat.com/browse/ARO-23919) and separate the tools. But at the moment creating new name image with same content is more confusing.

venkateshsredhat: yes i forgot its the same content . Approving .

PR #75094, opened 2026-02-19
feat(rhdh): skip gather chain on successful PR runs via allow_skip_on_success

gustavolira:
## Summary

- Add `allow_skip_on_success: true` to the RHDH `e2e-ocp-helm` PR test steps
- The `chain: gather` refs already have `optional_on_success: true`, so this flag enables ci-operator to skip them automatically when tests pass
- **Expected savings: ~8 minutes** on successful PR runs (gather-extra ~6m + must-gather ~1m43s)

## Changes

1-line config change: `allow_skip_on_success: true` added to the `e2e-ocp-helm` test steps.

On test failure, the full gather chain continues to run as before.

## Test plan

- [ ] Rehearse job passes
- [ ] Successful PR run skips gather chain (post-phase < 1 min)
- [ ] Failed PR run still collects full gather artifacts

## Related

- Companion PR (RHDH repo): https://github.com/redhat-developer/rhdh/pull/4267
- Jira: [RHIDP-12291](https://issues.redhat.com/browse/RHIDP-12291)

### DIFF
--- a/ci-operator/config/redhat-developer/rhdh/redhat-developer-rhdh-main.yaml
+++ b/ci-operator/config/redhat-developer/rhdh/redhat-developer-rhdh-main.yaml
@@ -32,6 +32,7 @@ tests:
     version: "4.18"
   skip_if_only_changed: ^docs/|rpms\.lock\.yaml|showcase-docs/|^\.changeset/|CONTRIBUTING\.md|OWNERS|README\.md|USAGE_DATA\.md|^scripts/|^\.github/|\.threatmodel/|^ui/|\.md|\\.mdc|\\.toml$
   steps:
+    allow_skip_on_success: true
     post:
     - ref: redhat-developer-rhdh-send-data-router
     - chain: gather


ci-operator/config/redhat-developer/rhdh/redhat-developer-rhdh-main.yaml
```diff
@@ -34,7 +34,7 @@ tests:
   steps:
     post:
     - ref: redhat-developer-rhdh-send-data-router
-    - chain: gather
+    - ref: redhat-developer-rhdh-gather
```
subhashkhileri: refs in `chain: gather` have `optional_on_success: true` which means we can use `allow_skip_on_success` flag and it should skip this automatically 
they just weren't being skipped because the test config is missing allow_skip_on_success: true, could you try this flag and see if it works in rehearse?

```suggestion
allow_skip_on_success: true
post:
    - ref: redhat-developer-rhdh-send-data-router
    - chain: gather
```

gustavolira: Great catch! You're absolutely right — `allow_skip_on_success: true` is much simpler since the gather chain refs already have `optional_on_success: true`. I've simplified the PR to just add that single flag. Removed the custom gather step entirely.

The PR is now a 1-line change. Let's see how rehearse goes.

subhashkhileri: Nice, lets update the rest of the jobs as well.

PR #75103, opened 2026-02-19
Adding USE_VIRTCTL option for k8s-netperf VM run

capolrik:


ci-operator/step-registry/openshift-qe/network-perf/openshift-qe-network-perf-ref.yaml
```diff
@@ -18,6 +18,10 @@ ref:
     default: "false"
     documentation:
       To run netperf on VMs
+  - name: USE_VIRTCTL
```
josecastillolema, line 21: Are there any other users of k8s-netperf in Prow with VM mode?
if they are not, i would suggest to set this variable default true so we dont need to modify the jobs

PR #75119, opened 2026-02-19
remove 4.21 netobserv perf jobs, update 4.22 frequency

memodi:
remove 4.21 netobserv perf jobs, update 4.22 frequency


Amoghrd:
oh yeah now I see it. The diff made it look that way

PR #75154, opened 2026-02-23
Rename EXTRA_FLAGS in virt workloads

rsevilla87:
Also removing QPS, BURST and TIMEOUT variables, which are now included in the defaults.

/cc @josecastillolema @capolrik


paigerube14 (approved):
/lgtm

PR #75156, opened 2026-02-23
ovn-kubernetes: Add XXLarge Core OVN node-density-cni and cluster-density-v2 jobs.

dceara:
Adds 1500-node scale tests with the Core OVN AWS allocation.  The following jobs are added:
- cluster-density-v2 (with the same type of scaling configuration as for 750 nodes)
- node-density-cni (with a reduced pod-per-node config, 200 -> 100, to make sure the cluster doesn't fall apart)

### DIFF
--- a/ci-operator/config/openshift/ovn-kubernetes/openshift-ovn-kubernetes-master.yaml
+++ b/ci-operator/config/openshift/ovn-kubernetes/openshift-ovn-kubernetes-master.yaml
@@ -735,7 +735,8 @@ tests:
     cluster_profile: aws-ovn-perfscale
     env:
       BASE_DOMAIN: core.ovn.devcluster.openshift.com
-      CD_V2_EXTRA_FLAGS: --churn-duration=20m --service-latency --timeout=12h
+      CD_V2_EXTRA_FLAGS: --churn-duration=20m --service-latency --timeout=12h --set=metricsEndpoints.0.step=5m
+        --set=metricsEndpoints.1.step=5m
       CDV2_ITERATION_MULTIPLIER: "6"
       CLUSTER_NETWORK_CIDR: 20.0.0.0/8
       COMPUTE_NODE_TYPE: m5.2xlarge
@@ -776,7 +777,7 @@ tests:
       CONTROL_PLANE_REPLICAS: "5"
       GC: "false"
       GC_METRICS: "false"
-      ND_CNI_EXTRA_FLAGS: --timeout=12h
+      ND_CNI_EXTRA_FLAGS: --timeout=12h --set=metricsEndpoints.0.step=5m --set=metricsEndpoints.1.step=5m
       OPENSHIFT_INFRA_NODE_INSTANCE_TYPE: r5.12xlarge
       OPENSHIFT_PROMETHEUS_STORAGE_SIZE: 250Gi
       PODS_PER_NODE: "100"


dceara:
/pj-rehearse pull-ci-openshift-ovn-kubernetes-master-ovncore-perfscale-aws-ovn-xxlarge-debug

dceara:
/pj-rehearse pull-ci-openshift-ovn-kubernetes-master-ovncore-perfscale-aws-ovn-xxlarge-node-density-cni

dceara:
/pj-rehearse pull-ci-openshift-ovn-kubernetes-master-ovncore-perfscale-aws-ovn-xxlarge-node-density-cni

dceara:
/pj-rehearse pull-ci-openshift-ovn-kubernetes-master-ovncore-perfscale-aws-ovn-xxlarge-node-density-cni

ci-operator/step-registry/workers-scale/workers-scale-commands.sh
```diff
@@ -93,15 +93,36 @@ current_worker_count=$(oc get nodes --no-headers -l "node-role.kubernetes.io/wor
...
+                waitForReady $worker_count_num
+            else
+                scaleMachineSets $worker_count_num
+                waitForReady $worker_count_num
+            fi
         fi
     fi
+}
+
+first_batch_size=500
```
rsevilla87, line 111: Why do we need a first_batch_size different than batch_size?

dceara: @rsevilla87, I suspect it's increasingly costlier to add batches of nodes to a running cluster.  If we had just `batch_size` we could end up in situations like:
- `batch_size` is too small, e.g., 100 and it takes too long to bring up larger clusters => 1500 nodes would mean 15 iterations so less parallelization => longer time to scale up
- `batch_size` is too big, e.g., 500 and we fail to bring up super large clusters.  For example, assuming we want to scale up to 5000 nodes, trying to add 500 nodes during the last batch (from 4500 to 5000) might be too much for the cluster to handle at once at that scale

Internally you had suggested making these configurable through env vars.  Maybe we should do that though, to allow people to tweak these values.  I can make that change, what do you think?

jtaleric (approved):
it would be good to eventually flip the bit for pprof collection.

just would like to get better collection so we can maybe get some insights at this scale.

in general 
/lgtm

ci-operator/config/openshift/ovn-kubernetes/openshift-ovn-kubernetes-master.yaml
```diff
@@ -727,6 +727,77 @@ tests:
...
+    - chain: create-infra-move-ingress-monitoring-registry
+    test:
+    - chain: workers-scale
+    - ref: openshift-qe-cluster-density-v2
+  timeout: 16h0m0s
+- always_run: false
+  as: ovncore-perfscale-aws-ovn-xxlarge-node-density-cni
+  optional: true
+  steps:
+    allow_skip_on_success: true
```
kyrtapz, line 770: I might be missing something but don't you need to set `optional_on_success=true` in steps you want to skip on success?

dceara: Thanks @kyrtapz!

Some steps in the chains we use have `optional_on_success=true`, e.g.:
```
ref:
  as: gather-must-gather
  best_effort: true
  optional_on_success: true
```

I actually want to change that in the future, and set `allow_skip_on_success: false` for our jobs but I'm not sure the MG step won't timeout so that part needs to wait.

Please let me know if that works for you.

PR #75164, opened 2026-02-23
remove approve plugin from rhdh-* repos

kadel:
- remove `approve` plugin from rhdh repos (
We are not using it. We have just one level PR approval process, so lgtm is enough for this.)
- cleanup of plugins that we never use `goose`, `pony` 
- require e2e test for main rhdh plugins


core-services/prow/02_config/redhat-developer/rhdh-chart/_prowconfig.yaml
```diff
@@ -5,6 +5,7 @@ branch-protection:
         rhdh-chart:
           branches:
             1.2.x:
+              enforce_admins: true
```
nickboldt: this whole branch is dead, you could remove 1.2.x entirely

PR #75214, opened 2026-02-24
ART-14530: [OKD] add 5.0 nightly streams

locriandev:


core-services/release-controller/_releases/release-okd-scos-5.0-art.json
bradmwilliams, line 1: This file will need to be named: `release-okd-scos-5.0-art.json` to match the annotation logic:
https://github.com/openshift/release/blob/d4002a42c50c3024211f35b4a5c0c0a7072307d5/hack/annotate.sh#L83-L85

PR #75227, opened 2026-02-24
SREP-2757: Add ROSA HCP conformance jobs for 4.21/4.22 and fix permafailing 4.19/4.20

dustman9000:
## Summary

- Add daily `e2e-rosa-hcp-ovn` conformance jobs for OCP 4.21 and 4.22 nightly builds
- Fix permafailing 4.19/4.20 HCP conformance jobs

## Changes

1. **4.21/4.22 job configs**: Added `e2e-rosa-hcp-ovn` periodic job to `openshift-release-main__nightly-4.21.yaml` and `openshift-release-main__nightly-4.22.yaml`, matching the existing 4.19/4.20 configuration.

2. **Fix 4.19/4.20 permafail**: The existing HCP conformance jobs have been permafailing with 0% pass rate. Investigation shows the conformance tests themselves pass (~58 minutes), but the job reports FAILURE because the `aws-deprovision-stacks` step in the post phase fails when CloudFormation VPC stack deletion hits `DELETE_FAILED` status. Marked the deprovision chain as `best_effort: true` so cleanup failures do not mask passing test results.

## Root Cause Analysis (4.19/4.20 failures)

All pre steps pass (VPC, OIDC, cluster provision, IDP setup, node readiness). Conformance tests pass in 58m13s. Post cleanup fails at `aws-deprovision-stacks`:

```
Waiter StackDeleteComplete failed: Waiter encountered a terminal failure state:
For expression "Stacks[].StackStatus" we matched expected path: "DELETE_FAILED" at least once
```

## Jira

Jira: https://issues.redhat.com/browse/SREP-2757

### DIFF
--- a/ci-operator/step-registry/rosa/aws/hcp/conformance/rosa-aws-hcp-conformance-workflow.yaml
+++ b/ci-operator/step-registry/rosa/aws/hcp/conformance/rosa-aws-hcp-conformance-workflow.yaml
@@ -34,6 +34,7 @@ workflow:
     post:
       - ref: osd-gather-extra
       - chain: rosa-aws-sts-hcp-deprovision
+        best_effort: true
       - ref: send-results-to-reportportal
   documentation: |-
     This workflow installs a rosa hcp cluster and performs openshift-tests conformance testing.
--- a/ci-operator/config/openshift/release/openshift-release-main__nightly-4.21.yaml
+++ b/ci-operator/config/openshift/release/openshift-release-main__nightly-4.21.yaml
@@ -2192,6 +2192,14 @@ tests:
       enable:
       - observers-resource-watch
     workflow: baremetalds-e2e-upgrade-ovn-ipv6-runc
+- as: e2e-rosa-hcp-ovn
+  cron: '@daily'
+  steps:
+    cluster_profile: aws-perfscale-qe
+    observers:
+      enable:
+      - observers-resource-watch
+    workflow: rosa-aws-hcp-conformance
 - as: e2e-rosa-sts-ovn
   cron: '@daily'
   steps:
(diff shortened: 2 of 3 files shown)


ci-operator/step-registry/rosa/aws/hcp/conformance/rosa-aws-hcp-conformance-workflow.yaml
```diff
@@ -34,6 +34,7 @@ workflow:
     post:
       - ref: osd-gather-extra
       - chain: rosa-aws-sts-hcp-deprovision
+        best_effort: true
```
joshbranham, line 37: Do we want to be leaking resources every time? Why was this passing before and now is perma-failing on cleanup?

dustman9000: Right, we definitely don't want to be leaking resources every run.

To clarify what's actually being leaked though: the cluster itself, OIDC config, and operator roles all clean up fine. It's just the CloudFormation VPC stack that sometimes fails to delete. Looks like a race condition where CloudFormation tries to tear down the VPC before AWS finishes releasing ENIs and security groups from the cluster we just deleted.

These jobs have actually never passed, they've been permafailing since we merged them a couple weeks ago. The conformance tests pass every run (~58 min), it's just this cleanup step at the end that tanks the whole job.

The `best_effort` flag lets the actual test results drive the job status instead of cleanup failures. The Classic STS conformance jobs don't hit this because they don't use CloudFormation for VPC provisioning (no BYOVPC).

Happy to create a follow-up PR to properly fix `aws-deprovision-stacks` with retry logic or a delay before the stack delete. The leaked VPCs should get caught by the CI account resource sweeper in the meantime.

joshbranham: Yeah fine to do in a follow up, just didn't want to ignore the fact that we are choosing to leak resources.

PR #75237, opened 2026-02-24
OCPBUGS-74689: upkeep: update tuned profile used for realtime

eggfoobar:
update the tuned profile used for realtime gcp lanes to account for rhel10

### DIFF
--- a/ci-operator/step-registry/optional-operators/configure-realtime-tuned-profile/optional-operators-configure-realtime-tuned-profile-commands.sh
+++ b/ci-operator/step-registry/optional-operators/configure-realtime-tuned-profile/optional-operators-configure-realtime-tuned-profile-commands.sh
@@ -7,12 +7,8 @@ set -o pipefail
 node_role=${APPLY_NODE_ROLE:=worker}
 max_cpu=8
 isolated_cpu=${COMPUTE_NODE_ISOLATED_CPU:-4}
-sched_rt_runtime_us=-1
-stalld_service="service.stalld=start,enable"
-gcp_pattern="[n|c|m|a]{1}[1-9]{1}d?-(standard|highcpu|highmem|highgpu){1}-([0-9]+)"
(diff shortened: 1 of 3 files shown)


eggfoobar:
/pj-rehearse periodic-ci-openshift-release-main-nightly-4.22-e2e-gcp-ovn-rt-rhcos10-techpreview

eggfoobar:
/pj-rehearse periodic-ci-openshift-release-main-nightly-4.22-e2e-gcp-ovn-rt-rhcos10-techpreview

ci-operator/step-registry/optional-operators/configure-realtime-tuned-profile/optional-operators-configure-realtime-tuned-profile-commands.sh
```diff
@@ -33,6 +47,35 @@ then
   stalld_service=""
 fi
 
+# stalld may not be available on RHCOS 10; verify before enabling
```
sdodson: Should it be? Is this a bug to fix in rhcos10?

eggfoobar: I don't think we need to do anything with it. It's been a while since I poked around RT, a lot has changed, I'm making some changes with the profiles from the tuned repo. Some of these params are hold outs from when the performance profile was its own operator, I think I recall stalld was packaged differently as part of MCO/NodeTuning. I'll need to verify to be certain, but from the readme on node-tuning it should be added as part of it. Verifying changes now. 

https://github.com/openshift/cluster-node-tuning-operator/blob/343c4c770d1ed76304204c5434586359ad80cf3d/README.md?plain=1#L293-L298

eggfoobar:
Thanks @sdodson! I think we're good, the older lanes I'm not too concerned about, and the recent failure in RHCOS10 does not seem related to our change.

/unhold
/pj-rehearse ack

PR #75251, opened 2026-02-24
FLPATH-3340 Refactor cost-onprem-chart jobs to remove MinIO and add S4 storage option

testetson22:
Switch from MinIO to S4 storage for cost-onprem-chart CI testing

### DIFF
--- a/ci-operator/step-registry/insights-onprem/cost-onprem-chart/e2e/insights-onprem-cost-onprem-chart-e2e-commands.sh
+++ b/ci-operator/step-registry/insights-onprem/cost-onprem-chart/e2e/insights-onprem-cost-onprem-chart-e2e-commands.sh
@@ -147,10 +147,6 @@ fi
 
 fi  # end SKIP_COST_MGMT_INSTALL check
 
-echo "========== Configuring Helm =========="
-echo "No Helm wrapper needed for S4 storage"
-echo "Storage configuration will be handled by deploy-test-cost-onprem.sh"
-
 echo "========== Configuring OpenShift SecurityContextConstraints =========="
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/insights-onprem/cost-onprem-chart/e2e/insights-onprem-cost-onprem-chart-e2e-commands.sh
```diff
@@ -258,10 +172,16 @@ echo "========== Running E2E Tests =========="
...
+    --verbose
+)
 
 # Run the deployment script from the chart repo source
 # The step runs with from: src, so we're already in the chart repo
 # Use bash to execute since source may be read-only (can't chmod)
-bash ./scripts/deploy-test-cost-onprem.sh \
-    --namespace "${NAMESPACE}" \
-    --verbose
+bash ./scripts/deploy-test-cost-onprem.sh "${DEPLOY_ARGS[@]}"
```
masayag, line 191: The script was extended with two additional flags to instruct s4 installation. See
https://github.com/insights-onprem/cost-onprem-chart/blob/main/scripts/deploy-test-cost-onprem.sh#L23-L24
```
#   --deploy-s4               Deploy S4 (Super Simple Storage Service) for S3-compatible storage
#   --s4-namespace NAME       S4 deployment namespace (default: s4-test)
```

testetson22: Exporting the related env vars should work, they're set in the ref.yaml and exported again above this line.

PR #75261, opened 2026-02-25
lp-chaos: onboard CNV node-outage chaos test cases (CNV-11299, 11301, 11306)

Userweiwei:
This PR onboards CNV chaos test cases (CNV-11299, CNV-11301, [CNV-11306](https://issues.redhat.com/browse/CNV-11306)) into the OpenShift CI Step Registry. The implementation focuses on validating VM SSH connectivity, snapshot integrity, and VM recreation following recovery from node-outage chaos injection.


Userweiwei:
/pj-rehearse periodic-ci-redhat-chaos-lp-chaos-main-ocp4.21-nightly--cnv-4.21-stable-cnvcases--aws-lp-chaos--cnv--node-outage

Userweiwei:
/pj-rehearse periodic-ci-redhat-chaos-lp-chaos-main-ocp4.21-nightly--cnv-4.21-stable-cnvcases--aws-lp-chaos--cnv--node-outage

Userweiwei:
/pj-rehearse periodic-ci-redhat-chaos-lp-chaos-main-ocp4.21-nightly--cnv-4.21-stable-cnvcases--aws-lp-chaos--cnv--node-outage

Userweiwei:
/cc @etirta 
Could you please help to review this PR? And I try to use $0 to identify the vm creation before and after chaos, but seems it didn't work well. My suggestion is to update the default VM__PREFIX, is it OK or any other good suggestions for it? Thanks in advance~

`    - name: LPC_LP_CNV__VM__PREFIX
      default: "cnv-chaos-vm-post"`

ci-operator/config/redhat-chaos/lp-chaos/redhat-chaos-lp-chaos-main__ocp4.21-nightly--cnv-4.21-stable-cnvcases--aws.yaml
etirta, line 1: Let's have a convention on the CI Operator Variant name.

First any reason we can't use the existing `redhat-chaos-lp-chaos-main__ocp4.20-nightly--cnv-4.20-stable--aws.yaml` Job Configuration?

Second, if it's really necessary to use a different Job Configuration, let's standardize the CI Operator Variant name: `ocp<ocpVer>-nightly--<lpName>-<lpVer>-stable--<testEnvPlatform>--...`

So we have consistent something line `ocp4.20-nightly--cnv-4.21-stable--aws--` prefix.

Btw, what is the context of `stable`?

ci-operator/config/redhat-chaos/lp-chaos/redhat-chaos-lp-chaos-main__ocp4.21-nightly--cnv-4.21-stable-cnvcases--aws.yaml
```diff
@@ -0,0 +1,66 @@
...
+            microdnf install -y ${pkgs} && \
+            microdnf clean all; \
+        fi && \
+        rm -rf /var/cache/dnf/* && \
+        wget -qO - 'https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64.tar.gz' | \
+            tar zvx -C /usr/local/bin/ --transform 's|^\./yq_linux_amd64$|yq|' ./yq_linux_amd64 && \
+        # Download virtctl
+        export VERSION=$(curl -s https://storage.googleapis.com/kubevirt-prow/release/kubevirt/kubevirt/stable.txt) && \
+        curl -L -o /usr/local/bin/virtctl https://github.com/kubevirt/kubevirt/releases/download/${VERSION}/virtctl-${VERSION}-linux-amd64 && \
+        chmod +x /usr/local/bin/virtctl
```
etirta, line 30: Have you managed to investigate creating the Container Image on `redhat-chaos/lp-choas` GitHub repo then mirror it in CI Operator, instead of creating the Container Image at test run time.

ci-operator/config/redhat-chaos/lp-chaos/redhat-chaos-lp-chaos-main__ocp4.21-nightly--cnv-4.21-stable-cnvcases--aws.yaml
```diff
@@ -0,0 +1,66 @@
...
+        else \
+            microdnf install -y ${pkgs} && \
+            microdnf clean all; \
+        fi && \
+        rm -rf /var/cache/dnf/* && \
+        wget -qO - 'https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64.tar.gz' | \
+            tar zvx -C /usr/local/bin/ --transform 's|^\./yq_linux_amd64$|yq|' ./yq_linux_amd64 && \
+        # Download virtctl
+        export VERSION=$(curl -s https://storage.googleapis.com/kubevirt-prow/release/kubevirt/kubevirt/stable.txt) && \
+        curl -L -o /usr/local/bin/virtctl https://github.com/kubevirt/kubevirt/releases/download/${VERSION}/virtctl-${VERSION}-linux-amd64 && \
```
etirta: Please stick with 1 way of download. Either use `wget` or `curl`, pick one. Btw, if use `curl` please use `curl -fsSL ...`.

ci-operator/step-registry/redhat-lp-chaos/lp/cnv/vm/snapshot/redhat-lp-chaos-lp-cnv-vm-snapshot-commands.sh
```diff
@@ -0,0 +1,59 @@
...
+
+: "Reading target VM list..."
+typeset vmList
+vmList=$(cat "${SHARED_DIR}/target-vm-name.txt")
+
+function CreateAndVerifySnapshot() {
+    typeset vmName="$1"
+    typeset snapshotName="${LPC_LP_CNV__SNAPSHOT_NAME:-chaos-snapshot}-${vmName}-$(date +%s)"
+    typeset -i waitTime=0
+    typeset -i timeoutMin="${LPC_LP_CNV__SNAPSHOT_TIMEOUT_MIN:-10}"
```
etirta: The `LPC_LP_CNV__SNAPSHOT_TIMEOUT_MIN` is declared in the Step Configuration file, so it's guaranteed to have default value. Let's use single source of truth principal, so let the default value set over there, not in here. So in case of we want to change it, we only change 1 place, instead of multiple.

ci-operator/step-registry/redhat-lp-chaos/lp/cnv/vm/snapshot/redhat-lp-chaos-lp-cnv-vm-snapshot-commands.sh
```diff
@@ -0,0 +1,59 @@
...
+kind: VirtualMachineSnapshot
+metadata:
+  name: ${snapshotName}
+  namespace: ${LPC_LP_CNV__VM__NS}
+spec:
+  source:
+    apiGroup: kubevirt.io
+    kind: VirtualMachine
+    name: ${vmName}
+EOF
```
etirta: Please follow our Best Practice for Resource Creation with data marshalling.

ci-operator/step-registry/redhat-lp-chaos/lp/cnv/vm/snapshot/redhat-lp-chaos-lp-cnv-vm-snapshot-commands.sh
```diff
@@ -0,0 +1,59 @@
...
+        phase=$(oc get vmsnapshot "${snapshotName}" -n "${LPC_LP_CNV__VM__NS}" -o jsonpath='{.status.phase}' 2>/dev/null || echo "Unknown")
+        if [[ "${phase}" == "Failed" ]]; then
+            : "Error: Snapshot ${snapshotName} transitioned to Failed phase"
+            oc get vmsnapshot "${snapshotName}" -n "${LPC_LP_CNV__VM__NS}" -o yaml
+            return 1
+        fi
+
+        sleep 15
+        waitTime=$((waitTime + 15))
+    done
```
etirta: Please follow our Best Practice for waiting on Resource creation.

Not sure if we need to bother to check `.status.phase`, wouldn't what is important just whether it's ready to use or not? Unless there is specific requirement or reason that I'm not aware of?

ci-operator/step-registry/redhat-lp-chaos/lp/cnv/vm/snapshot/redhat-lp-chaos-lp-cnv-vm-snapshot-commands.sh
```diff
@@ -0,0 +1,59 @@
...
+
+    : "SUCCESS: Snapshot ${snapshotName} created for VM ${vmName}."
```
etirta: With `xtrace` active, we should try to reduce unnecessary marker. It will be very obvious in the CI log how the execution flow, so reducing the marker will reduce the amount of log lines, hence easier for post-mortem analysis.

If the idea is to give marker in the source code, use `#` instead so it doesn't pollute the CI log with unnecessary line. Apply this through out the scripts.

ci-operator/step-registry/redhat-lp-chaos/lp/cnv/vm/create/redhat-lp-chaos-lp-cnv-vm-create-commands.sh
etirta, line 18: Extra caution. Some time the namespace is no immediately ready (rare situation though), so this may be optional. To be 100% sure the created namespace is ready, do the following.
```suggestion
oc wait "Namespace/${LPC_LP_CNV__VM__NS}" \
    --for jsonpath='{.status.phase}'=Active \
    --timeout 60s 1> /dev/null
```
OR
```suggestion
oc wait "Namespace/${LPC_LP_CNV__VM__NS}" \
    --for create \
    --timeout 60s 1> /dev/null
```

The former is more technically accurate, but the latter is more expressive and for Namespace they both almost equivalent (as normally namespace creation do not have much task, different with namespace deletion). Pick one and stick with it for consistency.

Do this principal for any resource creation, check with AI whether the resource creation is blocking (synchronous) or not. If un-blocking (asynchronous) then use `oc wait` to ensure the resource is actually ready to use. Each resource may have different condition what constitute ready.

ci-operator/step-registry/redhat-lp-chaos/lp/cnv/vm/snapshot/redhat-lp-chaos-lp-cnv-vm-snapshot-commands.sh
Userweiwei: Agree. Delete the default value in the .sh file

ci-operator/config/redhat-chaos/lp-chaos/redhat-chaos-lp-chaos-main__ocp4.21-nightly--cnv-4.21-stable-cnvcases--aws.yaml
Userweiwei: redhat-chaos-lp-chaos-main__ocp4.20-nightly--cnv-4.20-stable--aws.yaml --> this job is for krkn existing scenarios 
The new added redhat-chaos-lp-chaos-main__ocp4.21-nightly--cnv-4.21-stable-cnvcases--aws.yaml is for CNV Chaos Test Cases in the following doc: https://docs.google.com/spreadsheets/d/1JEwhqPwL6jdPvCfHseCncwGjjNbDYLqzpKro8vRbj9w/edit?gid=1150998127#gid=1150998127

And 'stable' is CNV_CHANNEL and detailed info can be checked in https://github.com/Userweiwei/release/blob/main/ci-operator/step-registry/interop-tests/deploy-cnv/interop-tests-deploy-cnv-ref.yaml.

ci-operator/config/redhat-chaos/lp-chaos/redhat-chaos-lp-chaos-main__ocp4.21-nightly--cnv-4.21-stable-cnvcases--aws.yaml
Userweiwei: Thanks for the suggestion. I haven't had the chance to investigate the image mirroring approach yet as I'm focusing on the hcp bm testing. I will definitely look into how to use it into our testing as an automation improvement item.

ci-operator/config/redhat-chaos/lp-chaos/redhat-chaos-lp-chaos-main__ocp4.21-nightly--cnv-4.21-stable-cnvcases--aws.yaml
Userweiwei: Agree!

ci-operator/step-registry/redhat-lp-chaos/lp/cnv/vm/snapshot/redhat-lp-chaos-lp-cnv-vm-snapshot-commands.sh
Userweiwei: Agree.

ci-operator/step-registry/redhat-lp-chaos/lp/cnv/vm/snapshot/redhat-lp-chaos-lp-cnv-vm-snapshot-commands.sh
Userweiwei: Agree. And update to '{.status.readyToUse}'=true

ci-operator/step-registry/redhat-lp-chaos/lp/cnv/vm/snapshot/redhat-lp-chaos-lp-cnv-vm-snapshot-commands.sh
Userweiwei: Thanks～ have updated.

ci-operator/step-registry/redhat-lp-chaos/lp/cnv/vm/create/redhat-lp-chaos-lp-cnv-vm-create-commands.sh
Userweiwei: Thanks. Selected the first one and updated.

ci-operator/step-registry/redhat-lp-chaos/lp/cnv/vm/snapshot/redhat-lp-chaos-lp-cnv-vm-snapshot-commands.sh
```diff
@@ -0,0 +1,43 @@
...
+set -euxo pipefail; shopt -s inherit_errexit
+
+# Load target VM list
+typeset vmList
+vmList=$(cat "${SHARED_DIR}/target-vm-name.txt")
+
+function CreateAndVerifySnapshot() {
+    typeset vmName="$1"
+    typeset snapshotName="${LPC_LP_CNV__SNAPSHOT_NAME:-chaos-snapshot}-${vmName}-$(date +%s)"
+    typeset timeout="${LPC_LP_CNV__SNAPSHOT_TIMEOUT_MIN}"
```
etirta: There is no reason to have a mirror var., it is just a waste of heap memory. For this use case, we could have just used `LPC_LP_CNV__SNAPSHOT_TIMEOUT_MIN` directly.

ci-operator/step-registry/redhat-lp-chaos/lp/cnv/vm/snapshot/redhat-lp-chaos-lp-cnv-vm-snapshot-commands.sh
```diff
@@ -0,0 +1,43 @@
...
+
+    if ! oc wait vmsnapshot "${snapshotName}" \
+        -n "${LPC_LP_CNV__VM__NS}" \
+        --for=jsonpath='{.status.readyToUse}'=true \
+        --timeout="${timeout}"; then
+        
+        : "Error: Snapshot ${snapshotName} failed to become ready within ${timeout}"
+        oc get vmsnapshot "${snapshotName}" -n "${LPC_LP_CNV__VM__NS}" -o yaml
+        return 1
+    fi
```
etirta: Remove unnecessary markers, as the execution flow is clearly shown due to `xtrace`.

Also is there value to show the existing Snapshot when it failed (line 32)? If there is none, consider the following code, to make it more concise (by virtue of `errexit`):
```suggestion
oc wait vmsnapshot "${snapshotName}" \
    -n "${LPC_LP_CNV__VM__NS}" \
    --for=jsonpath='{.status.readyToUse}'=true \
    --timeout="${LPC_LP_CNV__SNAPSHOT_TIMEOUT_MIN}"
```

ci-operator/step-registry/redhat-lp-chaos/lp/cnv/vm/snapshot/redhat-lp-chaos-lp-cnv-vm-snapshot-commands.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+set -euxo pipefail; shopt -s inherit_errexit
+
+# Load target VM list
+typeset vmList
+vmList=$(cat "${SHARED_DIR}/target-vm-name.txt")
+
+function CreateAndVerifySnapshot() {
```
etirta, line 8: This function is only called once from a source code perspective (despite being called several times during execution if it's in a loop). Is there any reason why it's a function? calling a function is slightly more expensive (so if it's called in a loop, the extra overhead will add up). The main reason for a function is code de-duplication, for source code maintainability, but it introduces the cost of calling a function, due to context preservation of caller and context isolation of the function itself.

ci-operator/step-registry/redhat-lp-chaos/lp/cnv/vm/snapshot/redhat-lp-chaos-lp-cnv-vm-snapshot-commands.sh
```diff
@@ -0,0 +1,43 @@
...
+        --arg ns "${LPC_LP_CNV__VM__NS}" \
+        --arg vm "${vmName}" \
+        '{
+            apiVersion: "snapshot.kubevirt.io/v1alpha1",
+            kind: "VirtualMachineSnapshot",
+            metadata: { name: $name, namespace: $ns },
+            spec: { source: { apiGroup: "kubevirt.io", kind: "VirtualMachine", name: $vm } }
+        }')
+
+    echo "${snapshotJson}" | oc apply -f -
```
etirta: Our best practice for resource creation when data marshaling is required, is having the pattern of:
```shell
#   #   Data marshalling is needed.
{
    oc create -f - --dry-run=client -o json --save-config |
    jq -c \
        --argjson argName1 "...valueInValidJSONstring..." \
         ... \
        --arg argName2 "...literalValue..." \
        '
            ...updateScript...
        ' |
    yq -p json -o yaml eval .
} 0<<'ocEOF' | oc apply -f -
 ...resourceYAMLtemplate...
ocEOF
 ...
```

The reason for using YAML as a template; it's easier for humans to read and detect errors, compared to JSON.
Also the `--save-config` is crucial to create annotation to make it truly idempotent, in case a 3-way merge is performed.

Your current way is technically fine, but it will generate unnecessary warnings in case a 3-way merge is performed (due to missing annotation) and having JSON as template is simply harder to read, therefore increasing the burden of maintenance.

Please do not deviate from our Best Practices, as they are crafted for max. efficiency. This will reduce unnecessary review cycles.

ci-operator/step-registry/redhat-lp-chaos/lp/cnv/vm/snapshot/redhat-lp-chaos-lp-cnv-vm-snapshot-ref.yaml
```diff
@@ -0,0 +1,17 @@
+ref:
+  as: redhat-lp-chaos-lp-cnv-vm-snapshot
+  from: cli-plus-tools
+  commands: redhat-lp-chaos-lp-cnv-vm-snapshot-commands.sh
+  env:
+    - name: LPC_LP_CNV__SNAPSHOT_NAME
+      default: "chaos-snapshot"
+    - name: LPC_LP_CNV__VM__NS
+      default: "cnv-chaos-test-ns"
+    - name: LPC_LP_CNV__SNAPSHOT_TIMEOUT_MIN
```
etirta: Remove the `_MIN` suffix. If this is only used by `oc wait`, it's expected to have a time unit and a user may want to use a different time unit.

ci-operator/step-registry/redhat-lp-chaos/lp/cnv/vm/snapshot/redhat-lp-chaos-lp-cnv-vm-snapshot-commands.sh
Userweiwei: Agree!

ci-operator/step-registry/redhat-lp-chaos/lp/cnv/vm/snapshot/redhat-lp-chaos-lp-cnv-vm-snapshot-ref.yaml
Userweiwei: Agree and update.

ci-operator/step-registry/redhat-lp-chaos/lp/cnv/vm/snapshot/redhat-lp-chaos-lp-cnv-vm-snapshot-commands.sh
Userweiwei: Agree and update

ci-operator/step-registry/redhat-lp-chaos/lp/cnv/vm/snapshot/redhat-lp-chaos-lp-cnv-vm-snapshot-commands.sh
Userweiwei: ACK.

ci-operator/step-registry/redhat-lp-chaos/lp/cnv/vm/snapshot/redhat-lp-chaos-lp-cnv-vm-snapshot-commands.sh
Userweiwei: I agree that function calls introduce a small overhead. However, given that the bottleneck here is the OCP API calls (oc apply/wait) which take seconds/minutes, I think maybe the Bash overhead is negligible.

ci-operator/step-registry/redhat-lp-chaos/lp/cnv/node-outage/OWNERS
```diff
@@ -0,0 +1 @@
+../vm/OWNERS
```
etirta: Any reason why not link just link to parent? It's more complex to link to sibling as sibling may be deleted.

ci-operator/step-registry/redhat-lp-chaos/lp/cnv/vm/create/redhat-lp-chaos-lp-cnv-vm-create-commands.sh
etirta: Nit & Tricks.

It's better for a template to use literal heredoc (quote the heredoc delimiter). You never have to worry about potential shell expansion what so ever, which in some edge cases, it may occurs, depending on the heredoc document content.

```suggestion
  } 0<<'EOF' | oc apply -f -
```

ci-operator/step-registry/redhat-lp-chaos/lp/cnv/vm/snapshot/redhat-lp-chaos-lp-cnv-vm-snapshot-commands.sh
```diff
@@ -0,0 +1,50 @@
...
+  source:
+    apiGroup: kubevirt.io
+    kind: VirtualMachine
+    name: ""
+ocEOF
+
+    oc wait vmsnapshot "${snapshotName}" \
+        -n "${LPC_LP_CNV__VM__NS}" \
+        --for=jsonpath='{.status.readyToUse}'=true \
+        --timeout="${LPC_LP_CNV__SNAPSHOT_TIMEOUT_MIN}"
```
etirta: It still has `_MIN` suffix. The env. var. has been renamed to `LPC_LP_CNV__SNAPSHOT_TIMEOUT`.

Userweiwei: Updated.

ci-operator/step-registry/redhat-lp-chaos/lp/cnv/vm/create/redhat-lp-chaos-lp-cnv-vm-create-commands.sh
Userweiwei: Agree

ci-operator/step-registry/redhat-lp-chaos/lp/cnv/node-outage/OWNERS
Userweiwei: updated

ci-operator/step-registry/redhat-lp-chaos/lp/cnv/vm/create/redhat-lp-chaos-lp-cnv-vm-create-ref.yaml
```diff
@@ -6,7 +6,7 @@ ref:
   - name: LPC_LP_CNV__VM__NS
     default: "cnv-chaos-test-ns"
   - name: LPC_LP_CNV__VM__PREFIX
-    default: "cnv-chaos-vm"
+    default: "cnv-chaos-vm-pre"
```
etirta: It's probably not a good idea to use `pre`, becaause this is under `vm/create`, not `vm/before-chaos`. I'd say stick with original `cnv-chaos-vm` make it more generic, yet still distinguishable with the post chaos creation. Let's keep this Step as generic as possible, so it can be called stand-alone (without the `after-chaos` at the end) without having the default value creating potential confusion.

Userweiwei: Agree！

PR #75275, opened 2026-02-25
MULTIARCH-5766 : refactor heavy-build into regular build

mpkredhat:
Aim is to improve MultiArch Power CI performance.
Refactoring heavy-build jobs to run as regular jobs under multi-p-p and ppc64le arch.


ci-operator/step-registry/openshift/e2e/libvirt/conf/openshift-e2e-libvirt-conf-commands.sh
```diff
@@ -313,7 +313,7 @@ fi
 
 # Until the yellow-zone network bandwidth is upgraded, we will run the following tests in their own periodic and exclude
 # them from the conformance-parallel workflow in versions older than 4.21.
-if echo ${BRANCH} | awk -F. '{ if (($1 == 4) && ($2 <= 20)) { exit 0 } else { exit 1 } }' &&  [[ "${TEST_TYPE}" == "conformance-parallel" || "${TEST_TYPE}" == "heavy-build" ]]; then
+if echo ${BRANCH} | awk -F. '{ if (($1 == 4) && ($2 <= 20)) { exit 0 } else { exit 1 } }' &&  [[ "${TEST_TYPE}" == "conformance-parallel" || "${TEST_TYPE}" == "heavy-build" ]] && [ "${ARCH}" != "ppc64le" ]; then
```
prb112, line 316: Is there one more location to update? Let's ensure that these tests are run in the rehearsal job.

KeerthanaAP, line 316: Also, @mpkredhat , Looks like we need to adjust the condition so that the exclusion applies only to versions 4.16 and below version when ARCH is ppc64le.

KeerthanaAP (changes requested):

ci-operator/step-registry/openshift/e2e/libvirt/conf/openshift-e2e-libvirt-conf-commands.sh
prb112: > Is there one more location to update?
I was thinking https://github.com/mpkredhat/release/blob/f927eab925220d00a8c2961c92087a554a950bca/ci-operator/step-registry/openshift/e2e/libvirt/conf/openshift-e2e-libvirt-conf-commands.sh#L339 was a different conditional. I'll mark this as resolved.

prb112: @KeerthanaAP how do you like the epxression above?

KeerthanaAP: Yes Paul, As discussed, since we plan to remove the heavy-build for all versions where the architecture is ppc64le, this expression looks good.

prb112: I'll leave the LGTM to you @Keerthana

PR #75285, opened 2026-02-25
feat: update owners of console plugin

anwesha-palit-redhat:


anwesha-palit-redhat:
Reopening this PR to get rid of the previously do-not-merge labels

PR #75334, opened 2026-02-25
[tls-scanner] Provision 4.22 cluster for default-tls test step

richardsonnick:
The default-tls conformance ci step attempts to claim a 4.22 cluster, yet that version does not exist in the cluster pool atm.

This diff provisions a new 4.22 cluster for the default-tls test step, similar to the tls13-conformance test.

### DIFF
--- a/ci-operator/config/openshift/tls-scanner/openshift-tls-scanner-main.yaml
+++ b/ci-operator/config/openshift/tls-scanner/openshift-tls-scanner-main.yaml
@@ -39,18 +39,12 @@ resources:
       memory: 200Mi
 tests:
 - as: default-tls
-  cluster_claim:
-    architecture: amd64
-    cloud: aws
-    owner: openshift-ci
-    product: ocp
-    timeout: 5h0m0s
-    version: "4.22"
   optional: true
   steps:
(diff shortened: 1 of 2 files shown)


ci-operator/config/openshift/tls-scanner/openshift-tls-scanner-main.yaml
```diff
@@ -39,18 +39,12 @@ resources:
...
-    product: ocp
-    timeout: 5h0m0s
-    version: "4.22"
   optional: true
   steps:
+    cluster_profile: aws
     test:
     - ref: tls-scanner-run
-    workflow: generic-claim
+    workflow: ipi-aws
```
rhmdnd, line 47: This must be on an older copy, since we did add the periodic jobs in https://github.com/openshift/release/pull/74850

The periodic default job is still using the cluster_claim workflow, which will need to be updated.

I have that done in https://github.com/openshift/release/pull/75394 if you want to use it.

richardsonnick: Yep let's use that!

PR #75342, opened 2026-02-26
config: add ci-pull-credentials pull secret

dustymabe:
Since registry.stage.redhat.io was added in [1]
let's include it in our test jobs so we can pull
from places like registry.stage.redhat.io/rhel10/rhel-bootc:10.2

I'll admit it was really hard to understand all the config
knobs here for someone who doesn't deal with ci-operator everyday.
After some fiddling with AI I discovered the docs at [2] and
was then able to make some progress.

[1] https://github.com/openshift/release/commit/7419a1af053aae671ca9c2e0b51fc39327691494
[2] https://docs.ci.openshift.org/how-tos/adding-a-new-secret-to-ci/#use-a-secret-in-non-step-jobs

Assisted-By: <google/gemini-3.1-pro-preview>

-- 

Also a second commit to drop the 9next tests that have always been noop tests and we don't plan to use.

### DIFF
--- a/ci-operator/config/coreos/rhel-coreos-config/coreos-rhel-coreos-config-main.yaml
+++ b/ci-operator/config/coreos/rhel-coreos-config/coreos-rhel-coreos-config-main.yaml
@@ -103,30 +103,46 @@ tests:
 - as: rhcos-9-build-test-qemu
   capabilities:
   - kvm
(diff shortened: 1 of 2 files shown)


dustymabe:
/pj-rehearse pull-ci-coreos-rhel-coreos-config-main-rhcos-9-build-test-qemu

PR #75392, opened 2026-02-26
MPIIT dr-reporter: Add support for FIPS attribute, extract job related info

oharan2:
This issue is intended to prevent discrepancies between jobs, and to force users to explicitly specify this in the job config file.

### DIFF
--- a/ci-operator/step-registry/mpiit/data-router-reporter/mpiit-data-router-reporter-commands.sh
+++ b/ci-operator/step-registry/mpiit/data-router-reporter/mpiit-data-router-reporter-commands.sh
@@ -14,27 +14,38 @@ function reportToDataRouter() {
     fi
 }
 
+function extractOCPVersion() {
+    # If OCP_VERSION is not set, we don't want to fail the entire launch report process
+    # The version attribute is not essentially required, so extract it from the JOB_NAME
+    if [[ -z "${OCP_VERSION}" ]]; then
+        if [[ "${JOB_NAME}" =~ ocp-([0-9]+\.[0-9]+) ]]; then
+            export OCP_VERSION="${BASH_REMATCH[1]}"
+        else
+            export OCP_VERSION="unknown"
+        fi
+    fi
+    true
+}
+
 export DATAROUTER_RESULTS="${SHARED_DIR}/*.xml"
-export OCP_VERSION
 export REPORTPORTAL_APPLY_TFA="${APPLY_TFA}"
 export REPORTPORTAL_CMP
 export REPORTPORTAL_HOSTNAME
 export REPORTPORTAL_PROJECT="lp-interop-dr_personal"
 export REPORTPORTAL_LAUNCH_NAME="${REPORTPORTAL_CMP}"
 
-if [[ "${JOB_NAME}" == *"-fips"* ]]; then
-  export FIPS_ENABLED="true"
-else
-  export FIPS_ENABLED="false"
-fi
+extractOCPVersion
+
+typeset version="${OCP_VERSION}"
+typeset fips="${FIPS_ENABLED}"
 
 REPORTPORTAL_LAUNCH_ATTRIBUTES="$(
     jq -nc \
         --arg jobName "${JOB_NAME}" \
         --arg buildID "${BUILD_ID}" \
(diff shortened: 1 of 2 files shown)


ci-operator/step-registry/mpiit/data-router-reporter/mpiit-data-router-reporter-commands.sh
```diff
@@ -22,17 +22,25 @@ export REPORTPORTAL_HOSTNAME
 export REPORTPORTAL_PROJECT="lp-interop-dr_personal"
 export REPORTPORTAL_LAUNCH_NAME="${REPORTPORTAL_CMP}"
 
+if [[ "${JOB_NAME}" == *"-fips"* ]]; then
+  export FIPS_ENABLED="true"
+else
+  export FIPS_ENABLED="false"
+fi
```
amp-rh: **FIPS_ENABLED: inferred vs explicit**

The PR removes `OCP_VERSION`'s default to force explicit specification in job configs, but `FIPS_ENABLED` takes the opposite approach — inferring it from `JOB_NAME`. This creates an inconsistency: one value must be explicit, the other is auto-detected.

To align with the PR's intent, consider adding `FIPS_ENABLED` to `mpiit-data-router-reporter-ref.yaml` (no default) and requiring job configs to set it, same as `OCP_VERSION`.

Also, `FIPS_ENABLED` is only used in this script (passed to `jq`), so it doesn't need to be exported. Per conventions, a local `typeset fipsEnabled` would be preferred.

amp-rh (commented):
Good change forcing `OCP_VERSION` to be explicit. One concern about the `FIPS_ENABLED` approach — see inline comment.

ci-operator/step-registry/mpiit/data-router-reporter/mpiit-data-router-reporter-commands.sh
oharan2: Thanks! I have addressed by doing the followings:

1. Add `typeset` of exported vars instead of redundant exporting
2. Add `FIPS_ENABLED` as this is the 1:1 way to identify a fips enabled job anyway, and all FIPS jobs are provided with this value set to true
3. Added documentation regarding the flexibility of `OCP_VERSION` availability/extraction.

amp-rh (approved):
/lgtm

ci-operator/step-registry/mpiit/data-router-reporter/mpiit-data-router-reporter-commands.sh
```diff
@@ -14,25 +14,44 @@ function reportToDataRouter() {
...
+    true
+}
+
 export DATAROUTER_RESULTS="${SHARED_DIR}/*.xml"
-export OCP_VERSION
 export REPORTPORTAL_APPLY_TFA="${APPLY_TFA}"
 export REPORTPORTAL_CMP
 export REPORTPORTAL_HOSTNAME
 export REPORTPORTAL_PROJECT="lp-interop-dr_personal"
 export REPORTPORTAL_LAUNCH_NAME="${REPORTPORTAL_CMP}"
```
etirta, line 35: I know this review came late, so it serves mainly as a note for future consideration.

This script contains significant variable redundancy. Re-exporting existing environment variables or creating mirrors (like `REPORTPORTAL_LAUNCH_NAME`) adds unnecessary footprint. We should use these variables directly or define them as Step-level parameters (like `REPORTPORTAL_APPLY_TFA`) to keep the script lean and maintainable.

ci-operator/step-registry/mpiit/data-router-reporter/mpiit-data-router-reporter-commands.sh
```diff
@@ -14,25 +14,44 @@ function reportToDataRouter() {
...
 
+extractOCPVersion
+
+typeset version="${OCP_VERSION}"
+typeset fips="${FIPS_ENABLED}"
+
```
etirta, line 41: Redundant variable assignment. Mirroring `${VAR}` to `${localVar}` adds boilerplate without adding value. Let's reference the environment variable directly throughout the step.

PR #75401, opened 2026-02-26
baremetalds-devscripts-proxy allow http GET HEAD port 80

tdomnesc:


ci-operator/step-registry/baremetalds/devscripts/proxy/baremetalds-devscripts-proxy-commands.sh
```diff
@@ -40,9 +40,12 @@ acl cluster dstdomain .metalkube.org .test-infra-cluster.redhat.com .ocpci.eng.r
 
 acl allowed_ssl_ports port 443 5000 6443
 acl CONNECT method CONNECT
+acl http_port port 80
```
andfasano: `http_port` is also used to define the squid port, for avoiding any confusion better a slightly different name

tdomnesc: changed

PR #75409, opened 2026-02-26
CORS-4311: Adding azure ipv6 presubmit jobs

rna-afk:
Adding presubmit job for dualstack azure testing.

### DIFF
--- a/ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh
+++ b/ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh
@@ -171,6 +171,34 @@ EOF
   yq-go m -a -x -i "${CONFIG}" "${patch_user_provisioned_dns}"
 fi
 
+# Configure dual-stack networking if IP_FAMILY is set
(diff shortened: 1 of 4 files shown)


ci-operator/step-registry/ipi/conf/azure/ipi-conf-azure-commands.sh
```diff
@@ -171,6 +171,52 @@ EOF
   yq-go m -a -x -i "${CONFIG}" "${patch_user_provisioned_dns}"
 fi
 
+# Configure dual-stack networking if IP_FAMILY is set
```
patrickdillon, line 174: we also need to set installconfig.platform.azure.ipFamily

patrickdillon: In fact do we need to set anything other than `ipFamily`? If ipv6 ranges in azure can be known in advance, the installer should default them. cc @jhixson74 @sadasu

rna-afk: Yeah simplified it

PR #75419, opened 2026-02-26
kubevirt-plugin: use a metal compute node

sjd78:
https://issues.redhat.com/browse/CNV-81348

Use the ipi-aws workflow with `c5n.metal` worker nodes to provide `/dev/kvm` access for KubeVirt testing. Metal instances enable hardware virtualization without emulation.

### DIFF
--- a/ci-operator/config/kubevirt-ui/kubevirt-plugin/kubevirt-ui-kubevirt-plugin-main.yaml
+++ b/ci-operator/config/kubevirt-ui/kubevirt-plugin/kubevirt-ui-kubevirt-plugin-main.yaml
@@ -31,7 +31,9 @@ tests:
       KUBEVIRT_PLUGIN_IMG_NAME: kubevirt-plugin
     env:
       BASE_DOMAIN: cnv-ci.syseng.devcluster.openshift.com
-      SPOT_INSTANCES: "true"
+      COMPUTE_NODE_REPLICAS: "1"
+      COMPUTE_NODE_TYPE: c5n.metal
+      CONTROL_PLANE_INSTANCE_TYPE: m6i.2xlarge
     test:
     - as: tests
       cli: latest


sjd78:
# Claude comparing logs from the metal config to current production config

## Comparison: Your Metal Config vs Current Production Config

| Metric | Your PR (Metal) | Current Config | Difference |
|--------|-----------------|----------------|------------|
| **Total Time** | 1h 58m | **1h 39m** | +19 min slower |
| Cluster Install | 51m | **43m** | +8 min slower |
| Test Phase | 28m | 28m | Same |
| Deprovision | **19m** | **5m** | +14 min slower |
| Pre Phase Total | 55m | 47m | +8 min slower |
| Post Phase Total | 27m | 19m | +8 min slower |

---

## Key Findings

### 1. Metal instances add ~20 minutes overhead

| Step | Metal (c5n.metal) | Standard | Reason |
|------|-------------------|----------|--------|
| Provisioning | 51 min | 43 min | Metal takes longer to allocate |
| Deprovision | 19 min | 5 min | More resources to clean up |

### 2. Your tests run the same speed

Both configs: **~28 minutes** for the test phase. This suggests your tests may not actually benefit from `/dev/kvm` hardware acceleration.

### 3. The current config doesn't use metal

Looking at the current production config (before your changes), it uses:
- Default instance types (m6i.xlarge or similar)
- `SPOT_INSTANCES: "true"` for cost savings

---

## Recommendation

**Question: Do your KubeVirt e2e tests actually create and run VMs?**

If **NO** (tests only interact with KubeVirt API/UI without spawning VMs):
→ Revert to the original config - metal adds 20 minutes with no benefit

If **YES** (tests create VMs that need to boot and run):
→ Keep metal, but the 20 min overhead may be worth it for reliable VM testing

---

## Quick Comparison Summary

```
Current Config (production):    1h 39m  ✓ Faster, uses spot instances
Your Metal Config:              1h 58m  ✓ /dev/kvm available, but slower

Difference: +19 minutes per CI run
```

sjd78:
The rehearsal run shows a slower infrastructure time and no significant test run time improvements.  If there were further test suite changes to allow running in parallel or were more dependent on real virtualization, the extra infrastructure time could be worth it.

PR #75430, opened 2026-02-27
fix the issue when create iam cli-role

zhouying7780:


zhouying7780:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.21-amd64-nightly-4.21-upgrade-from-stable-4.21-aws-ipi-ovn-hypershift-mce-inplace-f60

ci-operator/step-registry/hypershift/mce/aws/create/hypershift-mce-aws-create-chain.yaml
```diff
@@ -23,7 +26,11 @@ chain:
             AWS_GUEST_INFRA_CREDENTIALS_FILE="/etc/hypershift-ci-jobs-awscreds/credentials"
           fi
 
+          if [[ -z "$HYPERSHIFT_AWS_REGION" ]]; then
```
gaol, line 29: it looks like the `HYPERSHIFT_AWS_REGION` has default value set to `us-east-1`, so it won't be empty, and it will be always be the default value.

PR #75432, opened 2026-02-27
Fix AWS credential update in clc-create step

tanfengshuang:
- Correct jq path from .connections to .options.connections
  - Fix field name from awsSecretAccessKey to awsSecretAccessKeyID
  - Change colon operator to echo for better logging visibility

  This fixes the issue where AWS credentials were not being properly
  updated in options.yaml, causing cluster creation to fail.

### DIFF
--- a/ci-operator/step-registry/acm/tests/clc-create/acm-tests-clc-create-commands.sh
+++ b/ci-operator/step-registry/acm/tests/clc-create/acm-tests-clc-create-commands.sh
@@ -26,15 +26,15 @@ if [[ -f "${awsCredFile}" ]]; then
 
     [ -n "${awsAccKeyID}" ] && [ -n "${awsAccKeyToken}" ]
 
-    : "Updating credentials in ${optionFile}..."
+    echo "Updating credentials in ${optionFile}..."
     yq -o json eval . "${optionFile}" |
     jq -c \
           --arg awsAccKeyID "${awsAccKeyID}" \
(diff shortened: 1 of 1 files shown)


tanfengshuang:
/pj-rehearse periodic-ci-stolostron-policy-collection-main-ocp4.20-interop-opp-aws periodic-ci-stolostron-policy-collection-main-ocp4.21-interop-opp-aws

tanfengshuang:
> @tanfengshuang Great finding. But can you please explain how come some time it works? It was working in the PR that introduce this original change. Thx.

@etirta, I was wondering why hub worked but not on managed cluster. After confiming with DPTP, I thought it should be something wrong with my side, so I ran the job again with incorrect AWS key/id in optional.yaml, and it failed, then I got the root cause.

connections:
   apiKeys:
     aws:
       name: 'opp-aws-conn-auto'
       namespace: 'default'
       provider: 'Amazon Web Services'
       baseDnsDomain: cspilp.interop.ccitredhat.com
       awsAccessKeyID: XXXX
       awsSecretAccessKeyID: XXXX

I also tried to update it on Wed, but seems I updated too late and it couldn't be loaded into CI env (This led me to rule out the problem on my end.).   

Days ago, I should update the aws info of the option.yaml to the valid key/id for the opp testing before developing this script,. I planned to change aws key/id to unvalid, but seemed I forgot it.
(I should also check options.yaml during job running to make sure the script was correct, but I didn't, I trusted the job result too much.)

PR #75435, opened 2026-02-27
Fix resources requests/limits in redhat-developer/rhdh

liangxia:
According to usage in https://github.com/openshift/release/pull/74868, the maximax usage of resources is less than 10 CPU / 5 GiB memory. And most of the time, the resources usage is low. So update the limits/requests accordingly to use the resources more efficiently.


liangxia:
> Hi, I'm okay with the changes, as the rehearsals look fine. But if we encounter scheduling issues, we'd like to discuss increasing the requested resources.
> 
> We also have plans for CPU optimizations on our side.

We made this PR because,
setting large request of CPU/memory make our nodes can only run few pods/tests, and then we need many nodes to support a batch of tests from different github org/repos, which then results short of IP addresses ( and no easy solution for this).

We can discuss further to better support your use case if large CPU/memory is required for your tests.

I'm merging this PR for now, to get us out of issue from the short of IP adddresses.

PR #75446, opened 2026-02-27
Set more generic branch protection for service mesh repos

FilipB:


core-services/prow/02_config/openshift-service-mesh/sail-operator/_prowconfig.yaml
```diff
@@ -3,34 +3,13 @@ branch-protection:
...
-              protect: true
-              required_pull_request_reviews:
-                required_approving_review_count: 1
-              unmanaged: false
+          protect: true
+          required_pull_request_reviews:
+            required_approving_review_count: 1
+          include:
+          - main
+          - lpinterop-3.2
```
ferhoyos: Maybe it is better to use same pattern as release to cover future OSSM versions: `lpinterop-*`

core-services/prow/02_config/openshift-service-mesh/istio/_prowconfig.yaml
```diff
@@ -3,32 +3,13 @@ branch-protection:
...
-            release-1.28:
-              protect: true
-              required_pull_request_reviews:
-                required_approving_review_count: 1
-              unmanaged: false
+          protect: true
+          required_pull_request_reviews:
+            required_approving_review_count: 1
+          include:
+          - master
```
ferhoyos, line 7: Do we have both master and main branches in that repo?

core-services/prow/02_config/openshift-service-mesh/ztunnel/_prowconfig.yaml
```diff
@@ -3,32 +3,13 @@ branch-protection:
...
-            release-1.28:
-              protect: true
-              required_pull_request_reviews:
-                required_approving_review_count: 1
-              unmanaged: false
+          protect: true
+          required_pull_request_reviews:
+            required_approving_review_count: 1
+          include:
+          - master
```
ferhoyos, line 7: Do we have both master and main branches?

core-services/prow/02_config/openshift-service-mesh/istio/_prowconfig.yaml
FilipB: @mkralik3 Do you want that?

core-services/prow/02_config/openshift-service-mesh/ztunnel/_prowconfig.yaml
FilipB: Just in case istio switches to main in the future...

core-services/prow/02_config/openshift-service-mesh/istio/_prowconfig.yaml
FilipB: Sorry this is a question for previous conversation about lpinterop-3.2

core-services/prow/02_config/openshift-service-mesh/sail-operator/_prowconfig.yaml
mkralik3: Sure, we can, but at the same time, only one "active" lpinterop repo exists. Once lpinterop-3.3 branch exists and lpinterop job switches to that, we can delete lpinterop-3.2 branch.

FilipB: I added regex also for interop branch. I takes up to 6 hours to apply these changes so probably when creating/removing lpinterop- branch it will be more effective to edit the rule manually directly in GH, do the work and then restore the config to be synced with what we have here.

ferhoyos: Also you can edit the name of the branch and remove it if you don't want to change the rule

PR #75448, opened 2026-02-27
[INTEROP-8906] [INTEROP-8894] Updating Openshift Pipelines operator to OCP4.22 for FIPS and NON-FIPS (CR enabled) jobs - AWS-IPI testing

Sau1506mya:
This PR introduces the following updates:

- Consolidates the FIPS and non-FIPS jobs into a single configuration, as discussed within the team.
- Updates Slack reporting in the `.config.prowgen` file to reflect the new periodic job names.
- Updates the cron schedule for the existing CR job.
- Updates the OpenShift Pipelines 4.22 test mapping identifier to `lp-interop-OpenshiftPipelines`, replacing the previous identifier. [Ref. [INTEROP-8918](https://issues.redhat.com/browse/INTEROP-8918)]

More updates:

- Applied the suggestions
          -Updated the periodic job names and accordingly `.config.prowgen` file as well for both FIPS and NON-FIPS (CR) to remove the operator name redundancy.
          -Updated the FIPS tests with Report Portal mapping.
- Updated the timeout for openshift-pipelines-tests to 3 hrs for both FIPS and NON-FIPS (CR).

Thanks! 🚀

### DIFF
--- a/ci-operator/config/openshift-pipelines/release-tests/openshift-pipelines-release-tests-release-v1.21__ocp-4.22-lp-interop.yaml
+++ b/ci-operator/config/openshift-pipelines/release-tests/openshift-pipelines-release-tests-release-v1.21__ocp-4.22-lp-interop.yaml
@@ -88,4 +88,4 @@ zz_generated_metadata:
   branch: release-v1.21
   org: openshift-pipelines
   repo: release-tests
-  variant: openshift-pipelines-ocp-4.22-lp-interop
+  variant: ocp-4.22-lp-interop
--- a/ci-operator/jobs/openshift-pipelines/release-tests/openshift-pipelines-release-tests-release-v1.21-periodics.yaml
+++ b/ci-operator/jobs/openshift-pipelines/release-tests/openshift-pipelines-release-tests-release-v1.21-periodics.yaml
@@ -1,6 +1,196 @@
 periodics:
 - agent: kubernetes
   cluster: build11
+  cron: 0 3,15 * * *
+  decorate: true
+  decoration_config:
+    skip_cloning: true
+    timeout: 3h0m0s
+  extra_refs:
+  - base_ref: release-v1.21
+    org: openshift-pipelines
+    repo: release-tests
+  labels:
+    ci-operator.openshift.io/cloud: aws
+    ci-operator.openshift.io/cloud-cluster-profile: aws-cspi-qe
+    ci-operator.openshift.io/variant: ocp-4.22-lp-interop
+    ci.openshift.io/generator: prowgen
+    job-release: "4.22"
+    pj-rehearse.openshift.io/can-be-rehearsed: "true"
+  name: periodic-ci-openshift-pipelines-release-tests-release-v1.21-ocp-4.22-lp-interop-cr-openshift-pipelines-aws
+  reporter_config:
+    slack:
+      channel: '#tektoncd-pipeline-ci'
+      job_states_to_report:
+      - success
+      - failure
+      - error
+      report_template: '{{if eq .Status.State "success"}} :slack-green: Job *{{.Spec.Job}}*
+        ended with *{{.Status.State}}*. <{{.Status.URL}}|View logs> {{else}} :failed:
+        Job *{{.Spec.Job}}* ended with *{{.Status.State}}*. <{{.Status.URL}}|View
+        logs> {{end}}'
+  spec:
+    containers:
+    - args:
+      - --gcs-upload-secret=/secrets/gcs/service-account.json
+      - --image-import-pull-secret=/etc/pull-secret/.dockerconfigjson
+      - --lease-server-credentials-file=/etc/boskos/credentials
+      - --report-credentials-file=/etc/report/credentials
+      - --secret-dir=/secrets/ci-pull-credentials
+      - --target=cr-openshift-pipelines-aws
+      - --variant=ocp-4.22-lp-interop
+      command:
+      - ci-operator
+      env:
+      - name: HTTP_SERVER_IP
+        valueFrom:
+          fieldRef:
+            fieldPath: status.podIP
+      image: quay-proxy.ci.openshift.org/openshift/ci:ci_ci-operator_latest
+      imagePullPolicy: Always
+      name: ""
+      ports:
+      - containerPort: 8080
+        name: http
+      resources:
+        requests:
+          cpu: 10m
+      volumeMounts:
+      - mountPath: /etc/boskos
+        name: boskos
+        readOnly: true
+      - mountPath: /secrets/ci-pull-credentials
+        name: ci-pull-credentials
+        readOnly: true
+      - mountPath: /secrets/gcs
+        name: gcs-credentials
+        readOnly: true
+      - mountPath: /secrets/manifest-tool
+        name: manifest-tool-local-pusher
+        readOnly: true
+      - mountPath: /etc/pull-secret
+        name: pull-secret
+        readOnly: true
+      - mountPath: /etc/report
+        name: result-aggregator
+        readOnly: true
+    serviceAccountName: ci-operator
+    volumes:
+    - name: boskos
+      secret:
+        items:
+        - key: credentials
+          path: credentials
+        secretName: boskos-credentials
+    - name: ci-pull-credentials
+      secret:
+        secretName: ci-pull-credentials
+    - name: manifest-tool-local-pusher
+      secret:
+        secretName: manifest-tool-local-pusher
+    - name: pull-secret
+      secret:
+        secretName: registry-pull-credentials
+    - name: result-aggregator
+      secret:
+        secretName: result-aggregator
+- agent: kubernetes
+  cluster: build11
+  cron: 0 23 31 2 *
+  decorate: true
+  decoration_config:
+    skip_cloning: true
+    timeout: 3h0m0s
+  extra_refs:
+  - base_ref: release-v1.21
+    org: openshift-pipelines
+    repo: release-tests
+  labels:
+    ci-operator.openshift.io/cloud: aws
+    ci-operator.openshift.io/cloud-cluster-profile: aws-cspi-qe
+    ci-operator.openshift.io/variant: ocp-4.22-lp-interop
+    ci.openshift.io/generator: prowgen
+    job-release: "4.22"
+    pj-rehearse.openshift.io/can-be-rehearsed: "true"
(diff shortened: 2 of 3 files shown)


ci-operator/config/openshift-pipelines/release-tests/openshift-pipelines-release-tests-release-v1.21__openshift-pipelines-ocp-4.22-lp-interop.yaml
```diff
@@ -0,0 +1,58 @@
...
+      version: "4.22"
+resources:
+  '*':
+    limits:
+      memory: 4Gi
+    requests:
+      cpu: 100m
+      memory: 200Mi
+tests:
+- as: openshift-pipelines-interop-aws-fips
```
etirta: The words `openshift-pipelines` and `interop` is already appear in the Job Conf. file's basename as CI Operator `variant` part. Unless there is a good reason to have this duplicated in here, please remove it.


```suggestion
- as: aws-fips
```

etirta: Btw, if the Test Name is changes (in `.tests[].as`), correct me if I'm wrong, I believe the `.config.prowgen` may also need to be updated to reflect the change.

Changing the Test Name can be optional if there is complication more than just updating the `.config.prowgen`.

Sau1506mya: Yes @etirta , you’re correct. I made the following changes in this PR:

- Combined the FIPS and non-FIPS jobs into a single configuration, as discussed within the team.
- Updated the Slack reporting in the `.config.prowgen` file with the new periodic job names.
- Updated the cron value for the old CR job as well.
- Updated the OpenShift Pipelines 4.22 test mapping identifier to `lp-interop-OpenshiftPipelines` instead of the previous one.

Thanks!

ci-operator/config/openshift-pipelines/release-tests/openshift-pipelines-release-tests-release-v1.21__openshift-pipelines-ocp-4.22-lp-interop.yaml
```diff
@@ -0,0 +1,86 @@
...
+      version: "4.22"
+resources:
+  '*':
+    limits:
+      memory: 4Gi
+    requests:
+      cpu: 100m
+      memory: 200Mi
+tests:
+- as: cr-openshift-pipelines-interop-aws
```
etirta: The current generated name includes a duplicate `interop` string. Per the [Sippy variant registry logic](https://github.com/openshift/sippy/blob/6ad73ae236bacbc841b30042ad576cc64038f4be/pkg/variantregistry/ocp.go#L1147), we only need the `-lp-interop-cr-openshift-pipelines` part for proper identification.

Technical Constraint: To stay within the Prow job name length limits and reduce redundancy, I suggest simplifying the `as` field:
```suggestion
- as: cr-openshift-pipelines-aws
```
Action Item: Please also update `.config.prowgen` to match the new naming.

For the FIPS variant, I recommend `openshift-pipelines-aws-fips` to maintain naming parity and follow the team convention) with the non-FIPS job. This is optional, please check with the rest of the team whether the prefer the convention for non-FIPS job to be followed or not.

Sau1506mya: @etirta , why are we adding `openshift-pipelines` in `.tests[].as.` ? Its already present in the periodic job name since its part of the job config name.
Isn't it creating the redundancy of the operator in the periodic job name ?

Sau1506mya: That means for FIPS jobs, the  `.tests[].as.` should be  -> `aws-fips`
And for NON-FIPS (CR enabled) jobs, the  `.tests[].as.` should be  -> `cr-aws`

Thoughts?

etirta: > The current generated name includes a duplicate interop string. Per the [Sippy variant registry logic](https://github.com/openshift/sippy/blob/6ad73ae236bacbc841b30042ad576cc64038f4be/pkg/variantregistry/ocp.go#L1147), we only need the `-lp-interop-cr-openshift-pipelines` part for proper identification.

This is because, as per the reference above, the identification logic will fail without that specific string.

ci-operator/config/openshift-pipelines/release-tests/openshift-pipelines-release-tests-release-v1.21__openshift-pipelines-ocp-4.22-lp-interop.yaml
```diff
@@ -0,0 +1,86 @@
...
+      FIREWATCH_DEFAULT_JIRA_ADDITIONAL_LABELS: '["4.22-lp","self-managed-lp","pipelines-lp","fips"]'
+      FIREWATCH_DEFAULT_JIRA_ASSIGNEE: sselvan@redhat.com
+      FIREWATCH_DEFAULT_JIRA_PROJECT: LPINTEROP
+      FIREWATCH_FAIL_WITH_TEST_FAILURES: "true"
+      USER_TAGS: |
+        scenario pipelines
+    test:
+    - ref: openshift-pipelines-install
+    - ref: openshift-pipelines-tests
+    workflow: firewatch-ipi-aws
```
oharan2:
```suggestion
    workflow: firewatch-ipi-aws-cr
```

ci-operator/config/openshift-pipelines/release-tests/openshift-pipelines-release-tests-release-v1.21__ocp-4.22-lp-interop.yaml
```diff
@@ -0,0 +1,86 @@
...
+              {"step": "openshift-pipelines-install", "failure_type": "all", "classification": "Operator Installation Failure", "group": {"name": "lp-tests", "priority": 1}, "jira_project": "SRVKP", "jira_assignee": "!default", "jira_component": ["QA"], "jira_priority": "major", "jira_additional_labels": ["!default","interop-tests"]},
+              {"step": "openshift-pipelines-tests", "failure_type": "all", "classification": "Test Failure", "group": {"name": "lp-tests", "priority": 1}, "jira_project": "SRVKP", "jira_assignee": "!default", "jira_component": ["QA"], "jira_priority": "major", "jira_additional_labels": ["!default","interop-tests"]}
+            ]
+        }
+      FIREWATCH_CONFIG_FILE_PATH: https://raw.githubusercontent.com/CSPI-QE/cspi-utils/refs/heads/main/firewatch-base-configs/aws-ipi/lp-interop.json
+      FIREWATCH_DEFAULT_JIRA_ADDITIONAL_LABELS: '["4.22-lp","self-managed-lp","pipelines-lp","fips"]'
+      FIREWATCH_DEFAULT_JIRA_ASSIGNEE: sselvan@redhat.com
+      FIREWATCH_DEFAULT_JIRA_PROJECT: LPINTEROP
+      FIREWATCH_FAIL_WITH_TEST_FAILURES: "true"
+      USER_TAGS: |
```
oharan2, line 80: In this release, Fips tests are mapped as well (reporting to RP), so please update the related vars
```suggestion
      MAP_TESTS: "true"
      OCP_VERSION: "4.22"
      REPORTPORTAL_CMP: lp-interop-OpenshiftPipelines
      USER_TAGS: |
```

ci-operator/config/openshift-pipelines/release-tests/openshift-pipelines-release-tests-release-v1.21__openshift-pipelines-ocp-4.22-lp-interop.yaml
Sau1506mya: Agreed @etirta , applied the suggestions. Please check it and review it again.
Thanks!

ci-operator/config/openshift-pipelines/release-tests/openshift-pipelines-release-tests-release-v1.21__ocp-4.22-lp-interop.yaml
Sau1506mya: Hey @oharan2 , applied with the latest changes. Please check and review it again.
Thanks!

ci-operator/config/openshift-pipelines/release-tests/openshift-pipelines-release-tests-release-v1.21__ocp-4.22-lp-interop.yaml
etirta, line 1: @Sau1506mya Apologize if I wasn't clear in my previous comment about technical constraint on Prow Job Name length limit.

The final Prow Job Name is generated using three parts, i.e. Job Type, Job Conf. File Name, and Test Name (`.as`). The Job Conf. File Name is further composed of `<org>-<repo>-<branch>__<variant>`.

Since the `org`, `repo`, and `branch`, are effectively fixed by the repository structure, and Job Type is fixed by CI Operator, we only have control over the `variant` and the Test Name.

Currently, the generated name is `periodic-ci-openshift-pipelines-release-tests-release-v1.21-openshift-pipelines-ocp-4.22-lp-interop-cr-openshift-pipelines-aws`. Note the `openshift-pipelines` appears 3x (from the `repo`, `variant`, and Test Name). We should eliminate this redundancy to ensure future Job Name stays within length limit.

While we are forced to keep the string in the `repo` and the Test Name (the latter is required for Component Readiness (CR) identification), there is no requirement for it to appear in the variant.

Proposed Change:
If we remove `openshift-pipelines` from the variant, for example, changing the Job Conf. File Name to `openshift-pipelines-release-tests-release-v1.21__ocp-4.22-lp-interop`, we can safely reduce the total Job Name length while maintaining the required CR identification string.

I hope this clarifies the necessity of the change. Thanks!

Sau1506mya: Okay, I got your point and have updated the PR by removing the redundancy where possible. Hope this looks good now.
Thanks!

PR #75460, opened 2026-02-27
Added codecov.io token secret to upload gitops unit-test coverage.

AdamSaleh:
We want to track unit-test coverage and for better or worse we run our tests in openshift-ci.

This should allow us to have the token present when running the test.

### DIFF
--- a/ci-operator/config/redhat-developer/gitops-operator/redhat-developer-gitops-operator-master.yaml
+++ b/ci-operator/config/redhat-developer/gitops-operator/redhat-developer-gitops-operator-master.yaml
@@ -12,6 +12,10 @@ tests:
     test:
     - as: unit-steps
       commands: scripts/openshiftci-presubmit-unittests.sh
+      credentials:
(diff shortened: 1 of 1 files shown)


ci-operator/config/redhat-developer/gitops-operator/redhat-developer-gitops-operator-master.yaml
```diff
@@ -12,6 +12,10 @@ tests:
     test:
     - as: unit-steps
       commands: scripts/openshiftci-presubmit-unittests.sh
+      credentials:
+      - mount_path: /var/run/codecov-token
+        name: gitops-operator-codecov-token
+        namespace: test-credentials
```
psturc, line 18: is this job going to run against PRs and master branch?

AdamSaleh: Yes, is there a problem in this config?

PR #75465, opened 2026-02-27
Adding OSSM Hypershift workflow

fjglira:
Reusing the same scripts but creating our own step to tied the creation quota to our own ossm quota

### DIFF
--- a/ci-operator/step-registry/servicemesh/hypershift/hostedcluster/create/servicemesh-hypershift-hostedcluster-create-ref.yaml
+++ b/ci-operator/step-registry/servicemesh/hypershift/hostedcluster/create/servicemesh-hypershift-hostedcluster-create-ref.yaml
@@ -9,7 +9,7 @@ ref:
     default: "3"
     documentation: "The number nodes to automatically create and join to the cluster."
   - name: BASE_DOMAIN
-    default: ""
+    default: "servicemesh.devcluster.openshift.com"
     documentation: |-
       A fully-qualified domain or subdomain name. The base domain of the cloud provider is used for setting baseDomain variable of the install configuration of the cluster.
       It will be overridden by a non-empty HYPERSHIFT_BASE_DOMAIN.


ci-operator/step-registry/servicemesh/hypershift/hostedcluster/create/servicemesh-hypershift-hostedcluster-create-ref.yaml
```diff
@@ -0,0 +1,121 @@
...
+  cli: latest
+  env:
+  - name: PLATFORM
+    default: "aws"
+    documentation: "The platform(aws/powervs) of the HostedCluster. aws by default"
+  - name: HYPERSHIFT_NODE_COUNT
+    default: "3"
+    documentation: "The number nodes to automatically create and join to the cluster."
+  - name: BASE_DOMAIN
+    default: ""
```
yxun:
```suggestion
    default: "servicemesh.devcluster.openshift.com"
```

We may set a default value for using our account base domain.

ci-operator/step-registry/servicemesh/hypershift/hostedcluster/create/servicemesh-hypershift-hostedcluster-create-ref.yaml
```diff
@@ -0,0 +1,121 @@
...
+  - name: HYPERSHIFT_CP_AVAILABILITY_POLICY
+    default: "SingleReplica"
+    documentation: "Availability policy for hosted cluster components. Supported options: SingleReplica, HighlyAvailable, default SingleReplica"
+  - name: HYPERSHIFT_INFRA_AVAILABILITY_POLICY
+    default: "SingleReplica"
+    documentation: "Availability policy for infrastructure services in guest cluster. Supported options: SingleReplica, HighlyAvailable, default SingleReplica"
+  - name: OLM_CATALOG_PLACEMENT
+    default: "management"
+    documentation: "The OLM catalog placement. Supported options: management, guest, default management"
+  - name: POWERVS_REGION
```
yxun, line 44: Do we need all those empty default value environment variables such as those Powervs ones listed in our step ref ?
Are those just place holders when we don't want to modify their provision script ?

fjglira: THose are just placeholders to avoid modify their script. I don't want to add another script to mantain in our side xD

ci-operator/step-registry/servicemesh/istio/e2e/hypershift/servicemesh-istio-e2e-hypershift-workflow.yaml
```diff
@@ -2,7 +2,7 @@ workflow:
   as: servicemesh-istio-e2e-hypershift
   steps:
     pre:
-    - chain: hypershift-hostedcluster-create
+    - chain: servicemesh-hypershift-hostedcluster-create
```
yxun, line 5: I see you are adding a new chain here. 
There is another option for replacing that new chain with all the two steps refs in that new chain here. That will reduce the number of metadata.json files in this PR. 
It's an optional change. If you think adding a new chain is more clear, then this is also fine.

fjglira: I think it is clearer. At least to me

fjglira: I'll rerun the rehearsal. THey were stopped after the commit

MaxBab:
/pj-rehearse pull-ci-openshift-service-mesh-istio-master-istio-integration-pilot

FilipB (changes requested):
I don't think this is a correct approach. What we do is avoiding the default quota defined [here](https://github.com/openshift/release/blob/main/core-services/prow/02_config/_boskos.yaml#L4587). But we are still using the same management cluster which is maintained by the platform team. So we are avoiding their quota which was set for a reason. Correct approach should be to increase the quota defined by that team so they can review it and comment on that because it will affect how many control planes will be created in the management cluster.

FilipB:
This should be closed in favor of https://github.com/openshift/release/pull/75537

PR #75512, opened 2026-03-02
add new oadp-1.6 git repos

weshayutin:


mpryc (approved):
/lgtm

The kubevirt-datamover / vmfr/ vmdp repos are correct.

We also have this https://github.com/oadp-rebasebot/oadp-rebase repository where we store rebasebot configuration, but it's outside migtools and I don't think it's required to include it in here.

PR #75516, opened 2026-03-02
stolostron/managedcluster-import-controller: scope sonar tests to pkg

xuezhaojun:
## Summary
- set `SONAR_GO_TEST_ARGS="./pkg/..."` for managedcluster-import-controller sonar jobs
- apply the same setting across backplane branches and main config
- update presubmit skip regex to include `.rhtap`-only changes

## Testing
- configuration-only change


xuezhaojun:
> @xuezhaojun can we configure this by https://github.com/stolostron/managedcluster-import-controller/blob/main/sonar-project.properties ?

Good point！

PR #75541, opened 2026-03-02
chore: fix/upgrade-osd-gcp-operator-to-ocp-4.19

Fortune-Ndlovu:
the ocp ci job is using 4.15, which is only supported by RHDH 1.6 (EOL). For 1.7 or 1.8 we should use a supported version 4.16+. ref https://access.redhat.com/support/policy/updates/developerhub  

This pr is also related to  https://github.com/redhat-developer/rhdh/pull/4531 

Fixes #? https://redhat.atlassian.net/browse/RHDHBUGS-2749

### DIFF
--- a/ci-operator/config/redhat-developer/rhdh/redhat-developer-rhdh-main.yaml
+++ b/ci-operator/config/redhat-developer/rhdh/redhat-developer-rhdh-main.yaml
@@ -293,7 +293,7 @@ tests:
   steps:
     allow_skip_on_success: true
     env:
-      OSD_VERSION: 4.20.0
+      OSD_VERSION: 4.20.17
     post:
     - ref: redhat-developer-rhdh-send-data-router
     - ref: redhat-developer-rhdh-send-alert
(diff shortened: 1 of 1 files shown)


zdrapela:
@Fortune-Ndlovu it's good to trigger rehearsal for a job change like this 😉 . You can do it by commenting:
/pj-rehearse periodic-ci-redhat-developer-rhdh-main-e2e-osd-gcp-helm-nightly
That way you can verify it will work as expected

ci-operator/config/redhat-developer/rhdh/redhat-developer-rhdh-main.yaml
zdrapela: I'm not sure if exactly `4.20.0` is available on OSD-GCP, and maybe we want a newer z-stream release, like `4.20.6` or whatever is available. cc @subhashkhileri

Fortune-Ndlovu: sounds great, while testing the following locally, are the available `4.20.x` versions for OpenShift Dedicated on GCP with marketplace subscription, therefore i will update the version to `4.20.17`:
```bash
$ command -v ocm && ocm version
/home/fndlovu/.local/bin/ocm
1.0.12
$ ocm list versions --channel-group=stable 2>&1 |
      grep "^4\.20\." | sort -V
4.20.0
4.20.1
4.20.2
4.20.3
4.20.4
4.20.5
4.20.6
4.20.8
4.20.10
4.20.11
4.20.12
4.20.13
4.20.14
4.20.15
4.20.16
4.20.17
```

PR #74082, opened 2026-01-28
Validate HostedCluster conditions for Agent

mgencur:
Add step-registry component to verify HostedCluster status conditions in Agent platform deployments. Validates conditions like Available, ValidConfiguration, and SupportedHostedCluster before tests run.                                                                                                                                                                                              
                                                                                                                                                                                                                                            
Integrated into multiple hypershift-mce-agent test chains.

This mimics the [ValidateHostedClusterConditions](https://github.com/openshift/hypershift/blob/main/test/e2e/util/util.go#L2917) from openshift/hypershift. Hovewer, it only checks conditions that are general for all platforms or specific for Agent. Checks for other platforms (AWS, Kubevirt, etc.) are excluded because this step is specific for Agent.

Covers also ControlPlaneConnectionAvailable and DataPlaneConnectionAvailable from https://issues.redhat.com/browse/CNTRLPLANE-2203

PR #75544, opened 2026-03-02
Validate PlatformCredentialsFound since 4.17

mgencur:
This condition is properly set since 4.17

Explanation in https://github.com/openshift/release/pull/74082#issuecomment-3970636099


jparrill (commented):
/approve

PR #75548, opened 2026-03-02
updating preflight prod check to always capture logs

acornett21:


ci-operator/step-registry/operator-pipelines/preflight-prod/check/operator-pipelines-preflight-prod-check-commands.sh
```diff
@@ -54,12 +54,14 @@ set -e
 
 echo "Preflight exit code: ${preflight_exit_code}"
 
-if [ "${PUBLISH_ARTIFACTS}" == "true" ]; then 
+# Always save logs to ensure we can capture logs where preflight exits on error
+cp -a preflight.log "${ARTIFACT_DIR}"/
+cp -a "${preflight_stdout_file}" "${ARTIFACT_DIR}"/
+cp -a "${preflight_stderr_file}" "${ARTIFACT_DIR}"
```
caxu-rh: This `cp` statement lost its trailing slash - not sure it matters but just a heads up.

acornett21: I noticed this too, I'll repush/retest.

PR #75562, opened 2026-03-02
maistra builder: push to correct registry

zmiklank:


ci-operator/config/maistra/test-infra/maistra-test-infra-main.yaml
```diff
@@ -823,6 +832,7 @@ tests:
       env:
       - name: MAISTRA_NAMESPACE
       - name: MAISTRA_SC_POD
+      - name: HUB
```
jwendell, line 835: You must give it a value. I think HUB is only used in postsubmit tasks, so no need to set it in other places I think. Also I'm not sure if changing other bits in this PR is the right way.

zmiklank: I defined the value at line 803. Isn't that the correct way?

>    Also I'm not sure if changing other bits in this PR is the right way.

Those are just mismatched versions of ossm.

zmiklank: >  I think HUB is only used in postsubmit tasks, so no need to set it in other places I think.

That is correct. I added the HUB env var to postsubmit task of other OSSM versions as well.

mkralik3:
lgtm

mkralik3 (approved):

PR #75611, opened 2026-03-03
OCPBUGS-77244: Make runc informing for o/k

kannon92:
runc is not default in our OCP environments anymore so moving this job to optional.


sdodson:
/retitle OCPBUGS-77244: Make runc informing for o/k

So that it's linked in the bug in hopes of that helping us remember to come back and revert this

PR #75619, opened 2026-03-03
MCO: do not shard vsphere jobs

sergiordlr:
It is not allowed to shard vsphere jobs

In this PR we remove the sharding in vsphere jobs. In order to be able to execute the existing test cases we configured the mco-e2e-test chain in those jobs, which has a longer timeout.


ci-operator/config/openshift/machine-config-operator/openshift-machine-config-operator-release-4.21__periodics.yaml
```diff
@@ -124,12 +124,13 @@ tests:
...
 - as: e2e-vsphere-mco-disruptive-techpreview
   cron: 35 21 * * 6
-  shard_count: 2
   steps:
     cluster_profile: vsphere-elastic
     env:
       FEATURE_SET: TechPreviewNoUpgrade
       TEST_SUITE: openshift/machine-config-operator/disruptive
+    test:
+    - chain: mco-e2e-test
```
pablintino: What's the reason of this change?
The original workflow does this for testing:
```yaml
    - ref: openshift-e2e-test
```
But the one you propose:
```yaml
    - ref: mco-conf-day2-add-mcoqe-robot-to-pull-secret
    - ref: mco-conf-day2-create-custom-mcp
    - ref: mco-conf-day2-kerneltype
    - ref: mco-conf-day2-pintargetrelease
    - ref: mco-conf-day2-enable-ocl
    - ref: openshift-e2e-test-longrun
```
Despite I understand we need to increase the timeout by swapping the test to `openshift-e2e-test-longrun` _I don't get why we want to add those extra steps that the rest of the platforms don't do_.

sergiordlr: Currently it is not necessary to use the full chain. I usually configure the full chain because like that the jobs have the ability to run all test cases, if we only use the test step there are tests that cannot be executed, for example the Image Mode tests, or the off cluster layering tests. Those tests are not currently run by the disruptive suite, but if we ever add one of them to the suite we will have to reconfigure the job to add the chain so that the job can run it.

The test chain gives the job the ability to use different MCO configurations too. And if we ever need to change the jobs configuration changing the chain is easier than changing every job.

Nevertheless, nothing of this is used by the current job. Do you prefer if instead of the chain we use only the step?

pablintino: Yeah, I understood that part, those extra steps would be required, but, what annoys me a bit, is that we are not consistent, I mean, if we are now adding them, why only in vsphere? I'd prefer to have test equally on all platforms, and, if required, add those extra steps to all platforms equally.

PR #75630, opened 2026-03-04
CORENET-6562: update CNCC golangci-lint version to latest

yingwang-0320:
For CNO and CNCC 1.35 kube rebase, need update the golangci-lint version to v2.8.0 to use go1.25.

### DIFF
--- a/ci-operator/config/openshift/cloud-network-config-controller/openshift-cloud-network-config-controller-main.yaml
+++ b/ci-operator/config/openshift/cloud-network-config-controller/openshift-cloud-network-config-controller-main.yaml
@@ -2,7 +2,7 @@ base_images:
   golangci-lint:
     name: golangci-lint
     namespace: ci
-    tag: v2.3.1
+    tag: latest
   hypershift-operator:
     name: hypershift-operator
     namespace: hypershift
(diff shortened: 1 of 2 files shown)


ci-operator/config/openshift/cluster-network-operator/openshift-cluster-network-operator-master.yaml
```diff
@@ -6,7 +6,7 @@ base_images:
   golangci-lint:
     name: golangci-lint
     namespace: ci
-    tag: latest
```
ricky-rav, line 9: So you changed this to latest a month ago: https://github.com/openshift/release/pull/73972
Can you elaborate why we need to hardcode the version now?

yingwang-0320: When I filed this PR, the 'latest' tag still target to v2.3.1, but as I tried today, ocp-testplatform team has sloved this issue.

yingwang-0320: Let me update to latest

PR #75636, opened 2026-03-04
RHOAIENG-49724: Add timeout limit for opendatahub-operator CI

dayakar349:
Add layered timeout configuration to prevent runaway jobs and improve fail-fast behavior for ODH E2E tests.
JIRA: https://issues.redhat.com/browse/RHOAIENG-49724

### DIFF
--- a/core-services/prow/02_config/_config.yaml
+++ b/core-services/prow/02_config/_config.yaml
@@ -403,6 +403,9 @@ plank:
       timeout: 23h0m0s
     repo: openshift/open-telemetry-opentelemetry-operator
   - config:
+      timeout: 4h0m0s
+    repo: opendatahub-io/opendatahub-operator
+  - config:
       gcs_configuration:
         bucket: qe-private-deck
         mediaTypes:
--- a/ci-operator/config/opendatahub-io/opendatahub-operator/opendatahub-io-opendatahub-operator-main.yaml
+++ b/ci-operator/config/opendatahub-io/opendatahub-operator/opendatahub-io-opendatahub-operator-main.yaml
@@ -60,7 +60,7 @@ tests:
     workflow: opendatahub-io-ci-image-mirror
 - as: opendatahub-operator-image-mirror
   postsubmit: true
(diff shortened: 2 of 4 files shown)


core-services/prow/02_config/_config.yaml
```diff
@@ -402,6 +402,9 @@ plank:
       gcs_credentials_secret: gce-sa-credentials-gcs-qe-private-deck
       timeout: 23h0m0s
     repo: openshift/open-telemetry-opentelemetry-operator
+  - config:
+      timeout: 4h0m0s
```
macgregor, line 406: This is what a global catch all timeout? So max 4 hours cumulative wall clock time?

Typical case: we run concurrently, no issues and don't have to wait for a cluster. Roughly an hour run time.

Second Worse case scenario: we're at max concurrency -1 and odh and rhoai runs have to happen effectively sequentially. First cluster comes out of the pool and tests run, about an hour, then the second, maybe with a splash of extra provisioning time. We're still at about 2 hours + wait time over head.

Worse case: completely backed up, queue time plus sequential runs. 2 hours for testing, 30 minutes times two for cluster provisioning, we're at 3 hours sequential compute with an hour of queueing allowance. 

That sound right?

4 hours is extreme most of the time, but reasonable when we get backed up. So the question becomes what other mechanisms do we have so that the worst case fail safe doesn't slow down the typical case which is he situation we are in today. I had hoped to maybe see some individual step timeouts in addition to this so that say a stuck cluster provision doesn't hold up the cluster resource for 4 hours. If this is the only timeout we get, we'll be even more reliant on the health check circuit breaker to save us.

macgregor: wait maybe i mistunderstood what this timeout is for...i was reviewing it on my phone

dayakar349: You're right about the scenarios. The 4hr timeout is the cumulative wall clock limit for entire job from start to finish. I also added step level timeouts  so issues fail faster(2hr for cluster claim and 1h30m for the E2E step). The 4 hr cap is mainly act as safety limit in worst case situations like queue delays or sequential runs, and it stays below the default 8h global timeout.

dayakar349: I'll handle e2e-gather-extra step timeout issue (55m runtime) separately since it requires step-registry changes affecting repos using the general-claim workflow. The step currently uses the default 2h timeout. I'm scrapping timings from Prow PR artifacts and will look across other repos to determine a reasonable timeout value.

PR #75449, opened 2026-02-27
AGENT-1453: Add iso-no-registry compact job

andfasano:
This patch adds a new optional job to test appliance in the OVE scenario.

Requires https://github.com/openshift-metal3/dev-scripts/pull/1855

PR #75670, opened 2026-03-04
AGENT-1454: export APPLIANCE_IMAGE config var for OVE builds in devscripts-setup step

andfasano:
This config var will be set as a dependency only in the release Appliance repo, to allow testing the current appliance image.

Required by https://github.com/openshift/release/pull/75449


andfasano:
Closing as not currently required

PR #75671, opened 2026-03-04
Add baremetal-lab-ipi-virtual-media-ipv6-static workflow to chatbot

jadhaj:


core-services/ci-chat-bot/workflows-config.yaml
```diff
@@ -284,6 +284,13 @@ workflows:
         name: "4.16"
         namespace: ocp
         tag: upi-installer
+  baremetal-lab-ipi-virtual-media-ipv6-static:
+    platform: metal-qe
+    base_images:
+      upi-installer:
+        name: "4.16"
+        namespace: ocp
+        tag: upi-installer
```
bradmwilliams: Are you sure you don't want anything more modern than `4.16`?

jadhaj: no, not sure.
can it be 4.21 or 4.22
Also what is this filed used for ?

jadhaj: @bradmwilliams  Should it be the latest released version of OCP ?

bradmwilliams: That's really a question you need to ask yourself.  This setting is telling the workflow which image it should rely on to run.  So, essentially, if you wanted to test something from say `4.23:upi-installer`, you wouldn't be able to because it will not be available in the `4.16:upi-installer` image.

jadhaj: Actually I want to be able to test 4.23 and below. That means i need change also our existing workflows

sgoveas (approved):
/lgtm

PR #75699, opened 2026-03-04
Fix udn-bgp CI metadata indexing

rajeshP524:
Udn-bgp runs via SSH on a bastion host, so Prow env vars don't propagate. This causes e2e index.sh to skip metadata indexing, breaking Orion regression analysis.


venkataanil (approved):
/lgtm

PR #75752, opened 2026-03-05
chore: cleanup manual jobs

dhaiducek:
Previously manual jobs were needed for SonarCloud, but that has since changed in favor of configs generating those jobs. In some cases, these manual jobs had not been cleaned up during the migration, so this cleans up the extra jobs that have lingered (and are often failing since the builder image specified is outdated).

This also cleans up the archived `stolostron/console-api`, unused job directories, and does the config migrations for submariner-addon and provider-credential-controller.


dhaiducek:
Thanks for the reviews, all! With the rebase, I've engaged HCM build to review to get this merged in:
- https://redhat-internal.slack.com/archives/C04L50S5XM4/p1772816887509589

PR #75758, opened 2026-03-05
Change command style for easier use & update docs

psalajova:
Update the secret-manager CLI so it's in line with the new 3-tier hierarchy (see [design doc](https://docs.google.com/document/d/16ydEULUAsMEgjLkRYpGnGicxQh6_4REFwtbS5QqxxgM/edit?tab=t.0#heading=h.kwqkj07yoz41)).

Also changed the syntax to be like:
```
./secret-manager create -c my-creds aws/password -l "value"
```
```
./secret-manager delete -c my-creds aws/subgroup/password -l "value"
```
Instead of the previous `--collection={} --secret={}` format which would have been too verbose. .

### DIFF
--- a/hack/gcp-secret-manager/tests/test_validation.py
+++ b/hack/gcp-secret-manager/tests/test_validation.py
@@ -3,6 +3,8 @@
 Test script to verify validation in GCP Secret Manager names.
 """
 
+# pylint: disable=E0401, C0413
+
 import click
 
 from ..util import validate_collection, validate_secret_name, validate_group_name


hack/gcp-secret-manager/commands/create.py
```diff
@@ -59,9 +53,18 @@
 @click.option(
     "-l", "--from-literal", default="", help="Secret data as string input.", type=str
 )
-def create(collection: str, secret: str, from_file: str, from_literal: str):
-    """Create a new secret in the specified collection."""
+def create(collection: str, secret_path: str, from_file: str, from_literal: str):
+    """Create a new secret in the specified collection.
+
+    The SECRET_PATH should be in the format 'group/field' where:
```
deepsm007, line 59: Why not `group/field ↔ group__field`

psalajova: `group/field` is easier to write + it reflects the "nestedness" of the secrets :D in GSM it is actually saved as group__field, but since there can be 1+ groups, I feel like writing "abc/subgroup/field" is easier than "abc__subgroup__field"

PR #75759, opened 2026-03-05
MCO-2114: adds io.openshift.os.streamclass to OKD OS and extensions images

cheesesashimi:
To support a singular [OS Stream](https://github.com/openshift/enhancements/pull/1874) in OKD, we need to add a label to the SCOS 10 image (`io.openshift.os.streamclass`). For RHCOS 10, this label is set within the [build-node-image.Jenkinsfile](https://github.com/coreos/fedora-coreos-pipeline/blob/main/jobs/build-node-image.Jenkinsfile). Since the value of this label depends on whether it is an RHCOS 9 or RHCOS 10 image, the value must be set at build-time instead of via the Containerfile. Since OKD will only have one OS Stream (`centos-10`), we could set it statically in the Containerfile(s) found in https://github.com/openshift/os. However, I think it makes more sense to do it at build-time until there is a better way of setting it dynamically.

With the above in mind, this PR accomplishes that by modifying the image targets to be the input for an inline Containerfile which sets this label.

### DIFF
--- a/ci-operator/config/openshift/os/openshift-os-master__okd-scos.yaml
+++ b/ci-operator/config/openshift/os/openshift-os-master__okd-scos.yaml
@@ -21,7 +21,7 @@ images:
     stream-coreos-base:
       as:
       - quay.io/openshift-release-dev/ocp-v4.0-art-dev:c9s-coreos
-  to: stream-coreos
+  to: os-image
 - build_args:
   - name: OPENSHIFT_CI
     value: "1"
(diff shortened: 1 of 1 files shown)


dustymabe:
The code in the PR seems reasonable. The actual string suggested here may be able to be improved. For rhcos 9 and rhcos 10 the string we are using for the stream class is `rhel-9` and `rhel-10`. I'm not sure if we want `scos-10` here or `centos-10`.

dustymabe:
The label part of the PR lgtm. I'm not sure about the other changes.

Should we have someone from okd review this? @Prashanth684 ?

Prashanth684:
checked the image created from the rehearsal job and it has the label set

PR #75763, opened 2026-03-05
Handle arm64 architecture in gather-must-gather step

empovit:
* Download the right mco-sanitize binary by replacing the arch string to match the download site structure
* As `camgi` is only available for x86_64, gracefully skip it on other architectures to let the script proceed


empovit:
If I can find the owners of mco-sanitize, I will ask them to consider making it part of must-gather.

empovit:
Anyways, please let me know if the fix is good as as best effort towards supporting more architectures without failing the step, or should I just close the PR and implement a workaround in my jobs.

PR #75765, opened 2026-03-05
Add promotion config for openshift/cri-o

bitoku:
Add promotion configuration to enable payload job testing. Promotes cri-o RPMs while excluding rhel-coreos-9 image from promotion to avoid overwriting the official RHCOS image.

### DIFF
--- a/ci-operator/config/openshift/cri-o/openshift-cri-o-main.yaml
+++ b/ci-operator/config/openshift/cri-o/openshift-cri-o-main.yaml
@@ -39,6 +39,14 @@ images:
       as:
       - registry.ci.openshift.org/rhcos-devel/rhel-coreos-9
   to: rhel-coreos-9
+promotion:
+  to:
+  - additional_images:
+      cri-o-rpms: rpms
+    excluded_images:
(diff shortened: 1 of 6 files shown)


bitoku:
@rphillips 
This PR is needed to verify the fix in cri-o with the openshift payload.

These presubmits were just copied over from the cri-o/cri-o repository. For the time being, I only want to run the payload command on o/cri-o; Passing the presubmits is not required. (they should pass though)

bitoku:
@rphillips
Good catch!
It's not correct, but apparently cri-o/cri-o is also incorrect. 
I don't know how long it's been like that though...

@cpmeadors 
> Are we looking for pull-ci-openshift-cri-o-main-e2e-aws-ovn pull-ci-openshift-cri-o-main-e2e-gcp-ovn to pass? Is that what proves everything built correctly and was promoted?

No, we want to run payload job command in o/cri-o, though we need to check the cri-o version is the one of the PR, as @rphillips pointed out.

I'll move it forward and see if payload job can have the correct cri-o version.

/retest-required

PR #75766, opened 2026-03-05
NO-ISSUE: run MCO's AMI updater periodic once a month

djoshy:
This changes this periodic to run once a month(instead of daily) to decrease noise.


isabella-janssen (commented):
/lgtm

PR #75781, opened 2026-03-06
fix verify images for origin rebase

kannon92:
https://github.com/openshift/origin/pull/30646

This should fix the PR failures.

I'm not sure how best to roll this out though.

The change is failing because the rebase PR added new images and that broke the job.

If I merge this then would this effectively break origin until rebase lands?

cc @jacobsee


jubittajohn:
Superseded by https://github.com/openshift/release/pull/77005
/close

PR #75787, opened 2026-03-06
run post e2e-analysis tests only when install succeed

jianlinliu:
To reduce noise from post e2e-analysis in CR, specially when install failed, skip the post e2e-analysis tests.

### DIFF
--- a/ci-operator/step-registry/gather/extra/gather-extra-commands.sh
+++ b/ci-operator/step-registry/gather/extra/gather-extra-commands.sh
@@ -741,12 +741,20 @@ wait
 mkdir -p ${ARTIFACT_DIR}/junit/
 
 if openshift-tests e2e-analysis --help &>/dev/null; then
-    echo "Post e2e-analysis check for the cluster"
-    if [[ -f "${SHARED_DIR}/install-duration.log" ]]; then
-      echo "Found install-duration.log, it will be used for collecting install durations"
-      cat "${SHARED_DIR}/install-duration.log"
+    INSTALL_EXIT_CODE=0
+    if [[ -f "${SHARED_DIR}/install-status.txt" ]]; then
+        INSTALL_EXIT_CODE=$(tail -n1 "${SHARED_DIR}/install-status.txt" | awk '{print $1}')
     fi
-    openshift-tests e2e-analysis --junit-dir "${ARTIFACT_DIR}/junit" || true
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/gather/extra/gather-extra-commands.sh
```diff
@@ -741,12 +741,20 @@ wait
...
 if openshift-tests e2e-analysis --help &>/dev/null; then
-    echo "Post e2e-analysis check for the cluster"
-    if [[ -f "${SHARED_DIR}/install-duration.log" ]]; then
-      echo "Found install-duration.log, it will be used for collecting install durations"
-      cat "${SHARED_DIR}/install-duration.log"
+    INSTALL_EXIT_CODE=0
+    if test -f "${SHARED_DIR}/install-status.txt"; then
+        INSTALL_EXIT_CODE=$(tail -n1 "${SHARED_DIR}/install-status.txt" | awk '{print $1}')
+    fi
+    if [ "$INSTALL_EXIT_CODE" ==  0  ]; then
```
petr-muller:
```suggestion
    if [[ -f "${SHARED_DIR}/install-status.txt" ]]; then
        INSTALL_EXIT_CODE=$(tail -n1 "${SHARED_DIR}/install-status.txt" | awk '{print $1}')
    fi
    if [[ $INSTALL_EXIT_CODE ==  0 ]]; then
```

That's three `test` syntaxes on five lines  (`test`, single and double square brackets), plus an int/string comparison. It'll work but it's a bit ugly.

jianlinliu: thanks for the review, updated.

ci-operator/step-registry/gather/extra/gather-extra-commands.sh
neisw, line 756: What is the expectation here regarding aggregation, if we have 5 install failures will we see aggregation failure because we only have 5 successes but no tests for the remaining 5?  It would be more work but I'm wondering if we could generate actual test skips so the junit exists but indicates they skipped due to install failure.

neisw: @jianlinliu pointed out that when installs fail we don't get any e2e tests so this would just be the same behavior.

PR #75793, opened 2026-03-06
add stable job use oc-mirror

MayXuQQ:
1. add oc-mirror in stable install for [OCP-75367 - Mirror ocp release image to quay.io and install the cluster from the quay registry](https://polarion.engineering.redhat.com/polarion/redirect/project/OSE/workitem?id=OCP-75367)

2. resolve the bootstrap fail to query the release image from the registry of registry.build*.ci.openshift.org

From @aguidirh perspective , it is not oc-mirror problem, basically if there is a signature (stable payload), we get the release image from where the signature is (ignoring what was included in the image set config), 
if there is no signature (nightly payload) so we get from the specified place.

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.22__amd64-nightly-4.22-upgrade-from-stable-4.21.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.22__amd64-nightly-4.22-upgrade-from-stable-4.21.yaml
@@ -661,12 +661,14 @@ tests:
     test:
     - chain: openshift-upgrade-qe-test
     workflow: cucushift-installer-rehearse-gcp-ipi
-- as: ibmcloud-ipi-disc-priv-f28
+- as: ibmcloud-ipi-disc-priv-oc-mirror-f28
   cron: 45 20 15 * *
   steps:
     cluster_profile: ibmcloud-qe-2
     env:
+      MIRROR_BIN: oc-mirror
       TEST_FILTERS: ~ConnectedOnly&;~HyperShiftMGMT&;~MicroShiftOnly&
+      USE_ORIGINAL_OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE: "true"
     test:
     - chain: openshift-upgrade-qe-test-disconnected
     workflow: cucushift-installer-rehearse-ibmcloud-ipi-disconnected-private
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.22__amd64-nightly-4.22-upgrade-from-stable-4.22.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.22__amd64-nightly-4.22-upgrade-from-stable-4.22.yaml
@@ -720,13 +720,15 @@ tests:
     test:
     - chain: openshift-upgrade-qe-test
     workflow: cucushift-installer-rehearse-gcp-ipi-secureboot-confidential-computing
-- as: ibmcloud-ipi-disc-priv-f60
+- as: ibmcloud-ipi-disc-priv-oc-mirror-f60
   cron: 45 13 15 1,3,5,7,9,11 *
   steps:
     cluster_profile: ibmcloud-qe-2
     env:
+      MIRROR_BIN: oc-mirror
       SERVICE_ENDPOINT_COSConfig: DEFAULT_ENDPOINT
       SERVICE_ENDPOINT_GlobalCatalog: DEFAULT_ENDPOINT
+      USE_ORIGINAL_OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE: "true"
     test:
     - chain: openshift-upgrade-qe-test-disconnected
     workflow: cucushift-installer-rehearse-ibmcloud-ipi-disconnected-private
--- a/ci-operator/jobs/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.22-periodics.yaml
+++ b/ci-operator/jobs/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.22-periodics.yaml
@@ -6662,7 +6662,7 @@ periodics:
     ci-operator.openshift.io/variant: amd64-nightly-4.22-upgrade-from-stable-4.21
     ci.openshift.io/generator: prowgen
     pj-rehearse.openshift.io/can-be-rehearsed: "true"
-  name: periodic-ci-openshift-openshift-tests-private-release-4.22-amd64-nightly-4.22-upgrade-from-stable-4.21-ibmcloud-ipi-disc-priv-f28
+  name: periodic-ci-openshift-openshift-tests-private-release-4.22-amd64-nightly-4.22-upgrade-from-stable-4.21-ibmcloud-ipi-disc-priv-oc-mirror-f28
   spec:
     containers:
     - args:
@@ -6672,7 +6672,7 @@ periodics:
       - --oauth-token-path=/usr/local/github-credentials/oauth
       - --report-credentials-file=/etc/report/credentials
       - --secret-dir=/secrets/ci-pull-credentials
-      - --target=ibmcloud-ipi-disc-priv-f28
+      - --target=ibmcloud-ipi-disc-priv-oc-mirror-f28
       - --variant=amd64-nightly-4.22-upgrade-from-stable-4.21
       command:
       - ci-operator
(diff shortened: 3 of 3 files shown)


MayXuQQ:
/pj-rehearse  periodic-ci-openshift-verification-tests-main-installer-rehearse-4.20-installer-rehearse-ibmcloud-dis

ci-operator/config/openshift/verification-tests/openshift-verification-tests-main__installer-rehearse-4.20.yaml
```diff
@@ -103,13 +103,16 @@ tests:
     allow_skip_on_success: true
     cluster_profile: ibmcloud-qe-2
     env:
+      CUSTOM_OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE: quay.io/openshift-release-dev/ocp-release:4.20.15-x86_64
```
aguidirh: @MayXuQQ I see we are hard coding `4.20.15`, how are we going to test versions > 4.20.15?

MayXuQQ: This is for a temp test.

ci-operator/step-registry/mirror-images/by-oc-mirror/mirror-images-by-oc-mirror-commands.sh
```diff
@@ -168,6 +193,49 @@ run_command "${mirrorCmd}"
 # Save output from oc-mirror
 result_folder="${oc_mirror_dir}/working-dir"
 idms_file="${result_folder}/cluster-resources/idms-oc-mirror.yaml"
+
+echo "==== Validating IDMS for CI registry source ===="
+source_registry=$(echo "${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE}" | cut -d'/' -f1)
+echo "source_registry: ${source_registry}"
+
+# Only check/fix for CI registries (registry.buildXX.ci.openshift.org)
+if [[ "${source_registry}" == *"ci.openshift.org"* ]] && [[ -s "${idms_file}" ]]; then
```
jianlinliu: To be honest, I am not a fan of hacking the IDMS output for stable payload, especially when we run nightly build ci jobs, the condition is also met, this will hide some potential issues. And the hacking way make the codes hard to read. 

The more important things is, the issue happened since 4.19, but 4.18 works well.
[4.18 oc-mirror job](https://gcsweb-qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/qe-private-deck/logs/periodic-ci-openshift-openshift-tests-private-release-4.19-amd64-stable-4.19-upgrade-from-stable-4.18-azure-ipi-disc-oidc-oc-mirror-f28-ota/2028761781651902464/artifacts/azure-ipi-disc-oidc-oc-mirror-f28-ota/mirror-images-by-oc-mirror/build-log.txt)
v.s
[4.19 oc-mirror job](https://gcsweb-qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/qe-private-deck/logs/periodic-ci-openshift-openshift-tests-private-release-4.20-amd64-nightly-4.20-upgrade-from-stable-4.19-azure-ipi-disc-oidc-oc-mirror-f28-ota/2031211327497703424/artifacts/azure-ipi-disc-oidc-oc-mirror-f28-ota/mirror-images-by-oc-mirror/build-log.txt)

I can not understand why the IDMS output from 4.19+ oc-mirror did not have the mirror config for registry.build[0-9].ci.openshift.org when users specify registry.build[0-9].ci.openshift.org payload image in ImageSetConfiguration, while 4.18 stable payload (also 4.22 nightly payload works) ci jobs works well. From my perspective, this is a regression.

If you want to run the oc-mirror job for stable payload just like what the real customers did, I think you can follow the [doc](https://docs.google.com/document/d/14h9vripsgMP2Od1Fn5-hrF_yi1bxp2eUQ1ejjZGLKyU/edit?tab=t.0) to call rest API to override `RELEASE_IMAGE_LATEST` to run the prow job to avoid the unexpected payload-to-IS transition beside the `CUSTOM_OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE` way.

MayXuQQ: How about adding the signed payload condition?
```
isSigned=check_signed "${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE}"
if [[ ! isPreVersion "4.19"]] && isSigned && [[ "${source_registry}" == *"ci.openshift.org"* ]] && [[ -s "${idms_file}" ]]; then

```
the stable build is signed, the nightly build is not signed.
After 4.18, oc-mirror uses --v2, changed much.

I prefer to cover the stable payload + oc-mirror in the periodic test.

jianlinliu: > After 4.18, oc-mirror uses --v2, changed much.

Whatever how much it changed, I think the behavior should keep consistent. Users specify a registry.build[0-9].ci.openshift.org payload image in ImageSetConfiguration, but the IDMS output does not have anything to point it.

aguidirh: Hi @jianlinliu,

In 4.18 there was not signature verification, that's why it is working there. The signature verification was not backported to 4.18.

So basically from versions >= 4.19 where the signature verification is implemented, oc-mirror v2 tries to find a signature for the specified `registry.build[0-9].ci.openshift.org` and if it finds a signature it changes the place to get the release from the same place where the signature is (usually quay). That's why you don't see `registry.build[0-9].ci.openshift.org` in the IDMS, because during the mirroring the place with the signature was used (which is safer).

Since the nightly build is not signed, this verification is not happening so it continues using the specified `registry.build[0-9].ci.openshift.org` in the image set config.

If we want to avoid the hacking way we maybe could use the quay reference in the image set config instead of the one `registry.build[0-9].ci.openshift.org` for stable payloads.

jianlinliu: @aguidirh thanks for the detailed explanation. 

> If we want to avoid the hacking way we maybe could use the quay reference in the image set config instead of the one registry.build[0-9].ci.openshift.org for stable payloads.

Actually using registry.build[0-9].ci.openshift.org image in the image set config is not what we want to do, but prow ci build farm did something like payload-to-imagestream transition which changed the payload image url. Even if we use the quay reference in the image set config, the following ipi-install-install will still looking for the registry.build[0-9].ci.openshift.org image, refer to [here](https://github.com/openshift/release/blob/main/ci-operator/step-registry/ipi/install/install/ipi-install-install-ref.yaml#L68), that will fail the installation. To make sure the installation succeeds, we have to keep consistent for the same payload image url between the oc-mirror step and ipi-install-install step.

On the other side, assuming customers do some similar payload image management in a central registry, similar to what prow ci build farm does (though maybe an edge case), I guess the same issue would happen too.

jianlinliu: Is it possible to do some change in oc-mirror to add one more mirror in the IDMS output instead of hacking the output as a post action ?

aguidirh: If we do that in oc-mirror, the customer will get the same IDMS with two mirrors pointing to two different places (one of them will be invalid because we mirrored only one of them). So doing that in oc-mirror only to satisfay our internal process would be essentialy wrong.

If the correct place to get our releases is the place where the signature is located, why `ipi-install-install` is getting `registry.build[0-9].ci.openshift.org` ? Isn't it possible to change the `ipi-install-install` to get the signed release from the place where the signature is placed?

jianlinliu: > the customer will get the same IDMS with two mirrors pointing to two different places (one of them will be invalid because we mirrored only one of them).

Assuming we are the customer in this case, not add the `registry.build[0-9].ci.openshift.org` mirror, the install would fail, so the additional mirror means something, at least, it makes the install pass. The hacking codes is just adding the "wrong" mirror to get installation passed, isn't it ? That sounds the hacking of IDMS is acceptable, but it is NOT acceptable to add it from oc-mirror.

> why ipi-install-install is getting `registry.build[0-9].ci.openshift.org`

Per my understanding, because the installing payload is using an imagestream (`release:latest`) in the build farm, which is actually pointed to the real quay image as the backend. Maybe DPTP team is the right contact to answer to it.

 > Isn't it possible to change the ipi-install-install to get the signed release from the place where the signature is placed?

The mechanism of using `release:latest` in `ipi-install-install` has been existing for 6+ years, almost all the install is following the rule, it would a big concern to break many existing install, it is very risky.

aguidirh: > The hacking codes is just adding the "wrong" mirror to get installation passed, isn't it ? That sounds the hacking of IDMS is acceptable, but it is NOT acceptable to add it from oc-mirror.

Yes, the hacking of IDMS is happening only to make the installation pass.

> Maybe DPTP team is the right contact to answer to it.

Do you know who I could talk with about this topic from DPTP team?

jianlinliu: #forum-ocp-testplatform slack channel

PR #75839, opened 2026-03-06
OCPBUGS-77944: ipi-conf-gcp: filter zones by machine type availability

stbenjam:
## Summary

- Fix GCP zone selection to filter by machine type availability, not just AI zone exclusion
- Prevents ARM64 jobs from selecting zones where `t2a-standard-4` is unavailable (e.g. `us-central1-c`)
- Select zones independently for control plane and compute nodes to support heterogeneous clusters

## Problem

The `get_zones_from_region` function in `ipi-conf-gcp-commands.sh` only filtered out AI zones but did not check whether the requested machine type was actually available in each zone. This caused ARM64 CI jobs to fail with:

```
controlPlane.platform.gcp.type: Invalid value: "t2a-standard-4":
  instance type not available in zones: [us-central1-c]
```

**Affected job:** `periodic-ci-openshift-multiarch-main-nightly-4.22-upgrade-from-stable-4.21-ocp-e2e-upgrade-gcp-ovn`
Pass rate dropped from 100% to 40%.

## Fix

Replace `get_zones_from_region` with `get_zones_for_machine_type` which queries `gcloud compute machine-types list` to find zones where the specific machine type is available, with a fallback to the previous behavior if the query returns no results.

The separate `ipi-conf-gcp-zones-commands.sh` already had a similar `get_zones_by_machine_type` function — this brings the same logic to the main `ipi-conf-gcp` chain.

## Bug

https://issues.redhat.com/browse/OCPBUGS-77944

### DIFF
--- a/ci-operator/step-registry/ipi/conf/gcp/ipi-conf-gcp-commands.sh
+++ b/ci-operator/step-registry/ipi/conf/gcp/ipi-conf-gcp-commands.sh
@@ -61,13 +61,25 @@ if [[ -z "${COMPUTE_NODE_TYPE}" ]]; then
   fi
 fi
 
-# Get standard zones from the region (excluding AI zones) and randomize selection
-# This prevents control plane nodes from being placed in AI zones when zones aren't explicitly set
-function get_zones_from_region() {
-  local zone_count=${1:-3}
-  # Get all zones from the region, filtering out AI zones and randomizing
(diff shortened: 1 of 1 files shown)


stbenjam:
Closing — this PR has gone stale and is no longer being actively pursued. (Automated cleanup by agent)

PR #75843, opened 2026-03-06
vsphere - reduce the number of worker nodes

jcpowermac:


sgaoshang:
> > @jianlinliu can you make sure the change to the cucushift config makes sense, if not I can drop it
> 
> Sounds good to me, cc @sgaoshang and @WenXinWei for double confirm.

looks good to me

PR #75852, opened 2026-03-07
fix: tls-scanner app log streams

smith-xyz:
Setting up the tls-scanner go app to stream the logs so we can inspect more at runtime during the job


rhmdnd (approved):
/lgtm

PR #75889, opened 2026-03-09
odh-model-controller: move must-gather steps from test to post phase

pierDipi:
The gather steps (testlog-gather, kserve-must-gather, openshift-must-gather) were placed in the test phase where best_effort has no effect. When e2e tests fail, the remaining test-phase steps are skipped, so must-gather data is never collected when it is needed most.

Move gather steps to the post phase and add
allow_best_effort_post_steps across all e2e tests in all odh-model-controller config files.

Similar to https://github.com/openshift/release/pull/75771

### DIFF
--- a/ci-operator/config/opendatahub-io/odh-model-controller/opendatahub-io-odh-model-controller-incubating.yaml
+++ b/ci-operator/config/opendatahub-io/odh-model-controller/opendatahub-io-odh-model-controller-incubating.yaml
@@ -51,12 +51,47 @@ tests:
   optional: true
   skip_if_only_changed: ^(?:config|\.github|hack)/|\.md$|^(?:\.*/)?(?:\.gitignore|\.pre-commit-config\.yaml|OWNERS|PROJECT|LICENSE|\.dockerignore|Makefile)$
   steps:
+    allow_best_effort_post_steps: true
     cluster_profile: aws-opendatahub
     env:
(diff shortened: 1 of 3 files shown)


ci-operator/config/opendatahub-io/odh-model-controller/opendatahub-io-odh-model-controller-incubating.yaml
```diff
@@ -51,12 +51,46 @@ tests:
...
+      resources:
+        requests:
+          cpu: 100m
+      timeout: 1m0s
+    - as: kserve-must-gather
+      best_effort: true
+      cli: latest
+      commands: |
+        oc adm must-gather \
+          --image=quay.io/modh/must-gather:rhoai-2.24 \
```
bartoszmajsak, line 77: I did a few fixes around [LLMInferenceService must-gather](https://github.com/red-hat-data-services/must-gather/pull/203), perhaps we need newer image from here https://quay.io/repository/rhoai/odh-must-gather-rhel9

pierDipi: I don't see that repository, it's probably private

bartoszmajsak: That is not helpful.

PR #75902, opened 2026-03-09
switch to 4.21

jrosental:


ci-operator/config/codeready-toolchain/host-operator/codeready-toolchain-host-operator-master.yaml
```diff
@@ -4,12 +4,12 @@ build_root:
 releases:
   initial:
     integration:
-      name: "4.20"
+      name: "4.21"
```
rsoaresd, line 7: We need to also change in https://github.com/openshift/release/blob/main/ci-operator/config/codeready-toolchain/workload-analyzer/codeready-toolchain-workload-analyzer-master.yaml

metlos: sharp eyes! 👀

PR #75909, opened 2026-03-09
trying to fix 4.21 ols periodic tests

JoaoFula:


ci-operator/config/openshift/lightspeed-service/openshift-lightspeed-service-main__4.21.yaml
```diff
@@ -11,9 +11,9 @@ build_root:
       RUN dnf install -y podman
 releases:
   latest:
-    release:
-      architecture: amd64
-      channel: candidate
```
onmete, line 16: The PR was there for some time. The nightly version was not available originally and I missed updating this. Thanks.

JoaoFula: yeah we now have a similar thing for the 4.22. We get the automatic update for the cluster pool but potentially we need to create reminders for this change... Worst case scenario I also notice it on the weekly report and that's what triggered this change.

PR #75914, opened 2026-03-09
HYPERFLEET-695 - ci: update Slack report template to multi-line format

tzhou5:


xueli181114 (commented):
Overall this looks good — improves Slack message readability. A couple of minor observations inline.

ci-operator/jobs/openshift-hyperfleet/hyperfleet-e2e/openshift-hyperfleet-hyperfleet-e2e-main-periodics.yaml
```diff
@@ -21,10 +21,11 @@ periodics:
       - success
       - failure
       - error
-      report_template: '{{if eq .Status.State "success"}}:white_check_mark:{{else}}:x:{{end}}
-        *{{.Spec.Job}}* ended with *{{.Status.State}}*. <{{.Status.URL}}|View logs>
-        · <https://prow.ci.openshift.org/job-history/gs/test-platform-results/logs/{{.Spec.Job}}|View
-        history>'
+      report_template: |
+        {{if eq .Status.State "success"}}:white_check_mark:{{else}}:x:{{end}} *HyperFleet E2E Tests* ended with *{{.Status.State}}*.
```
xueli181114: Files under `ci-operator/jobs/` are auto-generated — they should not be manually edited. Can you confirm this was produced by running `make update` after editing `.config.prowgen`? If not, the determinize tooling may reformat this differently on the next `make update` run.

ci-operator/config/openshift-hyperfleet/hyperfleet-e2e/.config.prowgen
```diff
@@ -4,6 +4,10 @@ slack_reporter:
   - success
   - failure
   - error
-  report_template: '{{if eq .Status.State "success"}}:white_check_mark:{{else}}:x:{{end}} *{{.Spec.Job}}* ended with *{{.Status.State}}*. <{{.Status.URL}}|View logs> · <https://prow.ci.openshift.org/job-history/gs/test-platform-results/logs/{{.Spec.Job}}|View history>'
+  report_template: |
+    {{if eq .Status.State "success"}}:white_check_mark:{{else}}:x:{{end}} *HyperFleet E2E Tests* ended with *{{.Status.State}}*.
```
xueli181114: Nit: the previous template used `{{.Spec.Job}}` as the title, which is dynamic. The new template hardcodes `*HyperFleet E2E Tests*`. This is fine while there is only one job in `job_names`, but if more jobs are added later they would all show the same title. Consider keeping `{{.Spec.Job}}` in the title line to stay flexible.

ci-operator/config/openshift-hyperfleet/hyperfleet-e2e/.config.prowgen
```diff
@@ -4,6 +4,10 @@ slack_reporter:
   - success
   - failure
   - error
-  report_template: '{{if eq .Status.State "success"}}:white_check_mark:{{else}}:x:{{end}} *{{.Spec.Job}}* ended with *{{.Status.State}}*. <{{.Status.URL}}|View logs> · <https://prow.ci.openshift.org/job-history/gs/test-platform-results/logs/{{.Spec.Job}}|View history>'
+  report_template: |
```
xueli181114: The YAML block scalar (`|`) preserves a trailing newline, so the Slack message will end with an extra blank line. Probably harmless, but if you want to avoid it you can use `|-` (strip trailing newline) instead of `|`.

ci-operator/config/openshift-hyperfleet/hyperfleet-e2e/.config.prowgen
tzhou5: if we switch back to {{.Spec.Job}}, the multi-line format doesn't add much readability since the first line is already very long. The only gain would be putting links on separate lines. so, if it is necessary to update the format, or keep the current single-line format as-is (with {{.Spec.Job}})?

ci-operator/config/openshift-hyperfleet/hyperfleet-e2e/.config.prowgen
tzhou5: Agreed.

ci-operator/config/openshift-hyperfleet/hyperfleet-e2e/.config.prowgen
tzhou5: I've updated the template to multi-line with {{.Spec.Job}}as suggested. Though honestly the difference from the single-line version is minimal — it just adds Build ID and puts the links on a separate line. please let me know if you think this change is still worth merging, or if we should just close this PR and keep the original single-line format.

ci-operator/jobs/openshift-hyperfleet/hyperfleet-e2e/openshift-hyperfleet-hyperfleet-e2e-main-periodics.yaml
tzhou5: this was generated by running make jobs after editing .config.prowgen. just wondering, the if our ci check the file? does the ci generated-config passed means the check passed.

xueli181114: Make sense.

PR #75916, opened 2026-03-09
OCPBUGS-77970: Added multi node-selector logic for vsphere ci jobs

vr4manta:
[OCPBUGS-77970](https://issues.redhat.com/browse/OCPBUGS-77970)

### Changes
- Added multi poolSelector support to ci script
- Fixed vsanfs test jobs to use new (second) label

### Notes
A job was failing once we added vcenter-120 to the pool since it is not enabled for the vsanfs.  The new  label has been added to vcenter-110 so that we can make sure these type jobs can find a vcenter that has their supported storage config.


gnufied:
I am kinda confused by the fix opened for this bug   . May be I am missing something.

Why do Block volumes (ext4) need vSAN fileshare which is NFS only? Is that something new? Also - since ext4 does not support fencing, it CAN NOT be used as a RWX filesystem.  Using ext4 from multiple nodes will result in filesystem corruption (even if somehow vsan fileshare supported it).

PR #75940, opened 2026-03-09
Update golangci-lint for Cluster Version Operator to v2.10.1

JoelSpeed:
This is required to support Go 1.25

### DIFF
--- a/ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-main.yaml
+++ b/ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-main.yaml
@@ -2,7 +2,7 @@ base_images:
   golangci-lint:
     name: golangci-lint
     namespace: ci
-    tag: v2.3.1
+    tag: v2.10.1
   hypershift-operator:
     name: hypershift-operator
     namespace: hypershift
--- a/ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-release-4.22.yaml
+++ b/ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-release-4.22.yaml
@@ -2,7 +2,7 @@ base_images:
   golangci-lint:
     name: golangci-lint
     namespace: ci
-    tag: v2.3.1
+    tag: v2.10.1
   hypershift-operator:
     name: hypershift-operator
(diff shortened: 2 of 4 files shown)


ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-main.yaml
```diff
@@ -2,7 +2,7 @@ base_images:
   golangci-lint:
     name: golangci-lint
     namespace: ci
-    tag: v2.3.1
+    tag: v2.11.2
```
wking: Don't we need to manually do 4.22 and later as well?  Or is there some tooling that auto-syncs them with `main`?

```console
$ git grep -A3 golangci-lint origin/main -- ci-operator/config/openshift/cluster-version-operator | grep tag | sort -V
...
origin/main:ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-release-4.21.yaml-    tag: v1.64.8
origin/main:ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-release-4.22.yaml-    tag: v2.3.1
origin/main:ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-release-4.23.yaml-    tag: v2.3.1
origin/main:ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-release-5.0.yaml-    tag: v2.3.1
```

JoelSpeed: I don't believe there's any automation, so I've updated now to include 4.22, 4.23 and 5.0 release branches

DavidHurta: I was like, "_wait a minute, I have never updated the other branches as well_"

There seems to be an automation. For example:

1. I have updated the version to `v1.64.8` in https://github.com/openshift/release/commit/623a46b708c23eba9eb948d0ee2ee4b6c5b3f74e for main on Oct 24, 2025
2. Automation updates the 4.22 and 4.21 branches the same day in https://github.com/openshift/release/commit/5e77ef7ef332c725cd741aacfb5f9f5975638def with the `v1.64.8` version

PR #75941, opened 2026-03-09
Fix iteration multiplier to support fractional values

mohit-sheth:
Bash arithmetic $((...)) only supports integers, causing failures when fractional multipliers like 0.5 are used (e.g. UDN_ITERATION_MULTIPLIER). Switch to awk for the multiplication to handle both integer and fractional values correctly.


ci-operator/step-registry/openshift-qe/udn-density-pods/openshift-qe-udn-density-pods-commands.sh
rsevilla87: Not directly related to this pr, but, can you remove `--iterations=${ITERATIONS} ` from this line to prevent adding it twice in the command line?

PR #75957, opened 2026-03-09
Adding new "nomount" flavor to virt-density

jeniferh:
Adding a new step to test virt-density w/out extra "mounts" like cloudInit. Currently replacing the existing virt-density chain because of blocking bug https://issues.redhat.com/browse/CNV-68835 , will start w/ 4.22 metal runs and may expand to 4.21 metal later. 

Also bumping to CNV 4.22 now that the operator builds are stable.

### DIFF
--- a/ci-operator/step-registry/openshift-qe/virt-density-nomount/observer/openshift-qe-virt-density-nomount-observer-observer.yaml
+++ b/ci-operator/step-registry/openshift-qe/virt-density-nomount/observer/openshift-qe-virt-density-nomount-observer-observer.yaml
@@ -4,10 +4,6 @@ observer:
   from: perfscale-observer
   # Literal image stream tag that will be used to run the observer
   env:
-  - name: EXTRA_FLAGS
-    default: ""
-    documentation: |-
-      Add args for e2e-benchmarking kube-burner-ocp cmd.
   - name: E2E_VERSION
     default: "default"
     documentation: |-
--- a/ci-operator/step-registry/openshift-qe/orion/virt-density-nomount/openshift-qe-orion-virt-density-nomount-ref.metadata.json
+++ b/ci-operator/step-registry/openshift-qe/orion/virt-density-nomount/openshift-qe-orion-virt-density-nomount-ref.metadata.json
@@ -1,5 +1,5 @@
 {
-	"path": "openshift-qe/orion/virt-density/openshift-qe-orion-virt-density-nomount-ref.yaml",
+	"path": "openshift-qe/orion/virt-density-nomount/openshift-qe-orion-virt-density-nomount-ref.yaml",
 	"owners": {
 		"approvers": [
 			"perfscale-ocp-approvers"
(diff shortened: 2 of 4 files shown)


ci-operator/step-registry/openshift-qe/virt-density-nomount/observer/openshift-qe-virt-density-nomount-observer-observer.yaml
capolrik, line 31: this var is duplicated L.7

ci-operator/config/openshift-eng/ocp-qe-perfscale-ci/openshift-eng-ocp-qe-perfscale-ci-main__metal-4.22-nightly-x86.yaml
```diff
@@ -50,8 +50,8 @@ tests:
     - ref: openshift-qe-orion-data-path
     - ref: openshift-qe-virt-udn-density
     - ref: openshift-qe-orion-virt-udn-density
-    - ref: openshift-qe-virt-density
-    - ref: openshift-qe-orion-virt-density
+    - ref: openshift-qe-virt-density-nomount
```
josecastillolema, line 53: Instead of creating a whole new step wouldn't it be more convenient just to add a `MOUNT` var to the existing one? Wdyt?

jeniferh: This would be a much simpler change, but @jtaleric was suggesting a separate step for orion tracking, unless I misunderstood and there is another way to "mark" this in orion other than a whole new step?

josecastillolema: I see, perhaps he wanted to make sure that the results are kept "separate" from an Orion point of view to avoid misleading perf improvs&regressions. Is that it @jtaleric ?

ci-operator/step-registry/openshift-qe/virt-density-nomount/observer/openshift-qe-virt-density-nomount-observer-observer.yaml
jeniferh: good catch, will need to fix in virt-density as well

ci-operator/config/openshift-eng/ocp-qe-perfscale-ci/openshift-eng-ocp-qe-perfscale-ci-main__metal-4.22-nightly-x86.yaml
jtaleric: I would prefer we call out the steps explicitly so it is obvious to the jedi this was `openshift-qe-virt-density-nomount` that had a regression or not... if we simply go by the job name, we will need to dig into each var setup and see, oh this was `openshift-qe-virt-density` but with `mount=true` for example.  

There are other examples here (UDN chain comes to mind). it is messy tbh... we have L2 and L3 UDNs in one step that is controlled by a env var... While it reduces copy-paste... it adds to complexity IMHO.

josecastillolema: Ack, let's follow this approach then, thanks!

PR #75962, opened 2026-03-10
always extract installer from cluster payload in bastion install cases

jianlinliu:
By default, `ipi-install-install-in-bastion` was using installer from upi-installer image to run the install, but in some cases, when the job is triggered by gangway API, the installer might be mismatched with the cluster payload. So enhance it as always extract installer from cluster payload in bastion install cases


jianlinliu:
cc @gpei @jinyunma @jianli-wei to review.

PR #75975, opened 2026-03-10
RRP: update creds for ci

typeid:


ci-operator/config/openshift-online/rosa-regional-platform/openshift-online-rosa-regional-platform-main.yaml
```diff
@@ -58,8 +58,23 @@ tests:
...
+      resources:
+        requests:
+          cpu: 100m
+          memory: 200Mi
+- as: nightly-integration
+  cron: 0 7 * * *
+  steps:
+    test:
+    - as: e2e
+      commands: ./ci/e2e-tests.sh
```
typeid, line 73: TBD: We'll want to pass the api gw url and use the right creds for int here.

PR #76077, opened 2026-03-11
telco: cnf-features-deploy: test tools image on PR ref

shajmakh:
ci-operator by design uses base image `build_root` for all the test lanes where relevant. However that root image is built from master's head not the PR's branch.
The problem is that the changes of the tools image will not be applied to the testing env even when it is part of the PR.

The best option we can do to at least minimize the impact of build-root limitation is building the image as part of `images` stanza and control the base (fedora) in prow rather than in the kni repo.


shajmakh:
/pj-rehearse pull-ci-openshift-kni-cnf-features-deploy-master-images

PR #76086, opened 2026-03-11
make set-cluster-version-spec-update-service works with a custom payload image

jianlinliu:
when users specify a custom payload (e.g: registry.ci.openshift.org/ocp/release:4.22.0-0.nightly-2026-03-10-100251) to run testing, `oc adm release` command in `set-cluster-version-spec-update-service` will fail to authenticate to access the payload (not a build farm build image). Fix it to use cluster profile's pull-secret to pass the authentication.

slack thread is [here](https://redhat-internal.slack.com/archives/CH76YSYSC/p1773220601118189?thread_ts=1773106016.190449&cid=CH76YSYSC).


jianlinliu:
@yunjiang29 can you help review it ?

PR #76118, opened 2026-03-12
fix: set VM_EXTRADISKS_LIST="vda vdb" to align with lvms-vg1

gaol:


gaol:
/pj-rehearse ack

the rehearsed job should fail with the monitoring cluster operator not coming up because of csrs are rejected, that revealed an known issue in the capk, which will be fixed in a newer version.

PR #76147, opened 2026-03-12
Add go-tests job to monitoring-plugin

simonpasquier:
Will need https://github.com/openshift/monitoring-plugin/pull/835 first

### DIFF
--- a/ci-operator/config/openshift/monitoring-plugin/openshift-monitoring-plugin-main.yaml
+++ b/ci-operator/config/openshift/monitoring-plugin/openshift-monitoring-plugin-main.yaml
@@ -39,6 +39,18 @@ resources:
       cpu: 1000m
       memory: 8Gi
 tests:
+- as: go-tests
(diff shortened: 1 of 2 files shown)


ci-operator/config/openshift/monitoring-plugin/openshift-monitoring-plugin-main.yaml
```diff
@@ -41,7 +41,7 @@ resources:
 tests:
 - as: lint
   commands: |
-    make lint-frontend
+    make lint-frontend lint-backend && git diff --exit-code
```
jgbernalp: lint-frontend changes the current directory. We probably need to fix this from the makefile side

simonpasquier: in fact I had to change the approach because we need a Go-based image so I'm creating a new test.

PR #76149, opened 2026-03-12
ovn-kubernetes: Remove skip_if_only_changed for on-demand Core OVN jobs.

dceara:
These jobs are supposed to never run automatically, they have always_run: false.  They're supposed to be triggered only on-demand when OVN developers want to test custom changes at very large scale.

A recent change [0] added "skip_if_only_changed" to these (and other) jobs in the hope that the jobs will not run if only unrelated changes (e.g., doc changes) are part of the PRs under test.

However, the behavior of "skip_if_only_changed" is unexpected.  From the prow docs [2]:

  "prow will consider three different types of jobs that run on pull
  requests (presubmits):

    1. jobs that run unconditionally and automatically. All jobs that set always_run: true fall into this set.
    2. jobs that run conditionally, but automatically. All jobs that set run_if_changed or skip_if_only_changed to some value fall into this set.
    3. jobs that run conditionally, but not automatically. All jobs that set always_run: false and do not set run_if_changed/skip_if_only_changed to any value fall into this set and require a human to trigger them with a command. By default, jobs fall into the third category and must have their always_run, run_if_changed, or skip_if_only_changed configured to operate differently."

Which means that now all these jobs that used to fall into the third category (and were only run on humans triggering them) will fall into the second category and be run automatically.

Revert the change for the Core OVN jobs.  All other jobs affected by [1] should also be re-assessed individually in the near future.

[1] https://github.com/openshift/release/pull/74439
[2] https://docs.prow.k8s.io/docs/jobs/

### DIFF
--- a/ci-operator/config/openshift/ovn-kubernetes/openshift-ovn-kubernetes-release-4.21.yaml
+++ b/ci-operator/config/openshift/ovn-kubernetes/openshift-ovn-kubernetes-release-4.21.yaml
@@ -652,7 +652,6 @@ tests:
 - always_run: false
   as: ovncore-perfscale-aws-ovn-large-cluster-density-v2
   optional: true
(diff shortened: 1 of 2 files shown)


dceara:
CC: @jluhrsen @jcaamano

PR #76150, opened 2026-03-12
Fix SOCKS proxy port conflict on jumphost in openshift-qe-installer-bm-deploy-commands.sh

mcornea:
- Restrict random port range to 10000-32767 to avoid overlap with
  the Linux ephemeral port range (32768-60999)
- Kill stale SOCKS proxy SSH processes before starting a new one


mcornea:
/pj-rehearse pull-ci-redhat-performance-jetlag-main-deploy-sno

PR #76157, opened 2026-03-12
Remove legacy docs/px/qe-approved label requirements for ALBO and EDO

alebedev87:
## Summary
- Remove legacy `docs-approved`, `px-approved`, and `qe-approved` tide label requirements for `aws-load-balancer-operator` and `external-dns-operator`
- Remove the redundant tide query blocks that required these labels
- Consolidate `jira/valid-reference` into the remaining `jira/valid-bug` query
- Add `verified` label requirement to all tide queries


core-services/prow/02_config/openshift/aws-load-balancer-operator/_prowconfig.yaml
Miciah, line 33: These two rules are the same except that one of them requires the "jira/valid-bug" label and the other one does not:
https://github.com/openshift/release/blob/7cbde5e84561a772065f669e70e3fddce8748219/core-services/prow/02_config/openshift/aws-load-balancer-operator/_prowconfig.yaml#L5-L19
https://github.com/openshift/release/blob/7cbde5e84561a772065f669e70e3fddce8748219/core-services/prow/02_config/openshift/aws-load-balancer-operator/_prowconfig.yaml#L20-L33
I don't see why we need both rules.  One of them should either be deleted or updated so that the label sets are disjoint.  

https://docs.ci.openshift.org/architecture/branching/ mentions "jira/valid-bug" but not "jira/valid-reference".  However, the "jira/valid-reference" label is defined here: https://github.com/openshift/release/blob/7cbde5e84561a772065f669e70e3fddce8748219/core-services/prow/02_config/_labels.yaml#L382-L387
The "jira/valid-bug" label is defined here: https://github.com/openshift/release/blob/7cbde5e84561a772065f669e70e3fddce8748219/core-services/prow/02_config/_labels.yaml#L394-L399

If I understand correctly, allowing "jira/valid-reference" would enable you to satisfy the label requirements using (for example) an "NE" ticket whereas "jira/valid-bug" requires an "OCPBUGS" ticket—right?  So I think we want one rule that requires "jira/valid-bug" but does not require "jira/valid-reference", and another rule that requires  "jira/valid-reference" but does not require "jira/valid-bug".

core-services/prow/02_config/openshift/external-dns-operator/_prowconfig.yaml
Miciah, line 80: Same as for ALBO, I think EDNSO should have one label set with "jira/valid-bug" but not "jira/valid-reference" and one label set with "jira/valid-reference" but not "jira/valid-bug".

core-services/prow/02_config/openshift/aws-load-balancer-operator/_prowconfig.yaml
alebedev87: > If I understand correctly, allowing "jira/valid-reference" would enable you to satisfy the label requirements using (for example) an "NE" ticket whereas "jira/valid-bug" requires an "OCPBUGS" ticket—right? 

Yes, that was the idea.

> So I think we want one rule that requires "jira/valid-bug" but does not require "jira/valid-reference", and another rule that requires "jira/valid-reference" but does not require "jira/valid-bug".

I read it as `OCPBUGS` should be both a valid Jira reference and a valid bug (target version, release notes, dependency). When `OCPBUGS` is added to a PR title `jira/valid-reference` label is added too. So the addition of `jira/valid-reference` to the "bug query" was to match this behavior.

Miciah: > I read it as `OCPBUGS` should be both a valid Jira reference and a valid bug (target version, release notes, dependency). When `OCPBUGS` is added to a PR title `jira/valid-reference` label is added too. So the addition of `jira/valid-reference` to the "bug query" was to match this behavior.

So if a PR has "jira/valid-reference" it also has "jira/valid-bug", right?  Is there a reason to have a rule set with both "jira/valid-bug" and "jira/valid-reference" when you also have a rule set with just "jira/valid-reference"?

alebedev87: > So if a PR has "jira/valid-reference" it also has "jira/valid-bug", right?

The other way around, if a PR has `jira/valid-bug` it also has `jira/valid-reference`.

> Is there a reason to have a rule set with both "jira/valid-bug" and "jira/valid-reference" when you also have a rule set with just "jira/valid-reference"?

Having valid-reference label for a valid bug is superfluous but it matches the behavior of the ci bot which adds both labels. This is my reasoning.

alebedev87:
https://github.com/openshift/release/pull/80239 will take over from this one.

PR #76163, opened 2026-03-12
Set e2e-hypershift as optional for cluster-dns-operator

lihongan:


rikatz:
/pj-rehearse ack

the test is permafailing, and we need to remove it from now

PR #76191, opened 2026-03-12
openperouter: Onboarding CI

zeeke:
Create base Prow configuration for openshift-kni/openperouter

### DIFF
--- a/ci-operator/config/openshift-kni/openperouter/openshift-kni-openperouter-main.yaml
+++ b/ci-operator/config/openshift-kni/openperouter/openshift-kni-openperouter-main.yaml
@@ -10,6 +10,14 @@ images:
   to: openperouter-operator
 - dockerfile_path: operator/bundle.Dockerfile.openshift
   to: openperouter-operator-bundle
+operator:
+  bundles:
+  - as: operator-bundle
+    dockerfile_path: operator/bundle.Dockerfile.openshift
+    skip_building_index: true
+  substitutions:
+  - pullspec: quay.io/redhat-user-workloads/telco-5g-tenant/openperouter-operator-[\d].*
+    with: pipeline:openperouter-operator
 promotion:
   to:
   - name: "4.22"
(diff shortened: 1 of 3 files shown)


core-services/prow/02_config/openshift-kni/openperouter/_pluginconfig.yaml
```diff
@@ -1,5 +1,5 @@
 approve:
-- commandHelpLink: https://go.k8s.io/bot-commands
+- commandHelpLink: ""
```
maiqueb, line 2: do we really need this ? why did you change it ?

zeeke: good question! it comes out when running `make new-repo` . https://docs.ci.openshift.org/ does not mention it.
I can drop it if it matters

maiqueb: I'm OK with that - just curious :)

maiqueb:
> > @zeeke what are the goals of this PR ?
> > From the outputs, I can see we are now building D/S images for openperouter - I get that part :)
> > But is there anything else ?
> 
> Not yet. This configuration add some jobs the PRs in the openshift-kni/openperouter repository. These jobs build the operator and operator-bundle images, giving some information about the project to be built correctly. As a second step, the images built in the PR can be used for some end2end tests, deploying them in an openshift cluster and seeing if they work correctly.
> 
> Please, also note that the images built here don't relate to the release process (what populates our public registry.redhat.io). That part is configured via Tekton/Konflux

OK, nice to know. Tks for explaining.
/lgtm
/hold

I'm setting the "hold" so a responsible adult other than me can vouch these are OK. To my uneducated eyes, it looks OK, and the artifacts we want built were generated. 

Tks for this PR !

ci-operator/config/openshift-kni/openperouter/openshift-kni-openperouter-main.yaml
```diff
@@ -0,0 +1,45 @@
+binary_build_commands: make build
+build_root:
+  image_stream_tag:
+    name: release
+    namespace: openshift
+    tag: rhel-9-release-golang-1.25-openshift-4.22
+canonical_go_repository: "no"
```
oribon: why is this "no"? it seems this is the only file across the repo that uses this

zeeke: good catch. It came out from the `make new-repo` command.
Changed to `github.com/openperouter/openperouter` according to the [docs](https://steps.ci.openshift.org/ci-operator-reference#line155) and what the rest of the repo is doing.

PR #76203, opened 2026-03-12
Moving 4.23 jobs to be out of the way for more 4.22 testing

tvardema:
With the limited resources in multiarch CI, we are moving the 4.23 jobs to their "forever" cadence in the short term so we can run the dual stream focused testing for 4.22 more frequently in the short term.


prb112 (approved):
/lgtm

PR #76204, opened 2026-03-12
don't use -X theirs for branch sync PRs

jluhrsen:
originally we thought branch syncs would be all we needed, but in reality we are getting manual cherry picks for different reasons and merge conflicts can arise.

this will catch it, /hold the PR and make a comment to make it clear that someone will need to manually adress the conflict which will require a new PR


jluhrsen:
@arkadeepsen , @jcaamano , what do you think on this one?

@tssurya did question if we wanted this [here](https://redhat-internal.slack.com/archives/CDCP2LA9L/p1773631462064619?thread_ts=1773510174.962759&cid=CDCP2LA9L) in slack if you want to continue that discussion?

From what I can tell, now that we are allowing expedited cherry-picks in to synced branches (from time to time) the chances of conflicts are non-zero. Using `-X theirs` doesn't always get that right, but it does stay silent about it. If we remove `-X theirs` then we will at least get a signal and can manually address it. This is exactly what happened to us recently.

jluhrsen:
/close
this never gained favor

PR #76223, opened 2026-03-13
Ensure ansible playbook error is not ignored

ngopalak-redhat:
The PR is raised due to the success reported by the job: https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-cri-o-cri-o-main-periodics-setup-fedora-periodic/2032245295688978432

In reality the logs show the failure: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/periodic-ci-cri-o-cri-o-main-periodics-setup-fedora-periodic/2032245295688978432/artifacts/setup-fedora-periodic/cri-o-setup-fedora-test/build-log.txt

Since this error was not captured by the caller, the build succeeded without the rest of the packages and caused issues while running

Test result shows that the job is failing now: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/76223/rehearse-76223-periodic-ci-cri-o-cri-o-main-periodics-setup-periodic/2032395323594248192

### DIFF
--- a/ci-operator/step-registry/cri-o/setup-fedora/test/cri-o-setup-fedora-test-commands.sh
+++ b/ci-operator/step-registry/cri-o/setup-fedora/test/cri-o-setup-fedora-test-commands.sh
@@ -14,9 +14,17 @@ timeout --kill-after 10m 400m ssh "${SSHOPTS[@]}" ${IP} -- bash - <<EOF
     SOURCE_DIR="/usr/go/src/github.com/cri-o/cri-o"
     cd "\${SOURCE_DIR}/contrib/test/ci"
     ansible-playbook setup-main.yml --connection=local -vvv
+    ANSIBLE_EXIT_CODE=\$?
     sudo rm -rf "\${SOURCE_DIR}"
+    exit \${ANSIBLE_EXIT_CODE}
 EOF
 
+if [ $? -ne 0 ]; then
+    echo "ERROR: Ansible playbook failed, not creating base image"
+    exit 1
+fi
+
+echo "Ansible playbook succeeded, creating base image..."
 currentDate=$(date +'%s')
 gcloud compute instances stop ${instance_name} --zone=${ZONE}
 disk_name=$(gcloud compute instances describe ${instance_name} --zone=${ZONE} --format='get(disks[0].source)')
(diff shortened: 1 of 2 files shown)


ci-operator/step-registry/cri-o/setup-fedora/test/cri-o-setup-fedora-test-commands.sh
```diff
@@ -14,9 +14,17 @@ timeout --kill-after 10m 400m ssh "${SSHOPTS[@]}" ${IP} -- bash - <<EOF
     SOURCE_DIR="/usr/go/src/github.com/cri-o/cri-o"
```
bitoku, line 14: Can you add `set -xeuo pipefail` here?

saschagrunert (approved):
/lgtm

ci-operator/step-registry/cri-o/setup-fedora/test/cri-o-setup-fedora-test-commands.sh
saschagrunert: The suggestion is technically correct, but adding `set -e `would skip the `sudo rm -rf` cleanup on failure. The current approach captures the exit code first, runs cleanup, then propagates the failure, which is slightly more careful. The outer if check is unreachable though (the outer set -e already aborts the script when the ssh command fails), so the error message won't appear in logs.

bitoku: The instance image won't be saved when failed, so we don't have to run `sudo rm -rf`, do we?

When we want to run some additional commands in this ssh session, we need more checks like this without `pipefail`. I just want to avoid it though I won't block this PR only by this suggestion.

ngopalak-redhat: This was already there in line 4. My understanding was that its not working and hence took this approach.

bitoku: no it's not propagated in the ssh session.
in rhel setup it's defined, but somehow fedora doesn't have it.

PR #76238, opened 2026-03-13
Acquire install lease when provisioning a cluster

danilo-gemoli:
This PR tries to mitigate the rate-limiting issue we are facing on several cloud accounts, Azure being the most affected one.
The core idea is about limiting the number of jobs, per cluster profile (hence cloud account), that are allowed to provision a cluster.

We achieve that by acquiring an `install` lease (see https://github.com/openshift/release/pull/76230) from a small pool, and holding it only for about 20m.
During the first 20m `openshift-install` makes a lot of requests to a cloud provider, therefore increasing the odds of being rate-limited, particularly during the CI rush hours.

~~There is a lot of going on in this script, but reviewing it is much easier assuming this mental model:~~
```sh
source "$LEASE_PROXY_CLIENT_SH"

function acquire_install_lease_atomic { ... }
function release_install_lease_delayed_atomic { ... }
function release_and_acquire_install_lease_atomic { ... }

trap 'release_install_lease_atomic' EXIT TERM INT

max=5
tries=1
ret=4
release_install_lease_pid=''
while [ $ret -eq 4 ] && [ $tries -le $max ]
do
  echo "Install attempt $tries of $max"

  if [ $tries -gt 1 ]; then
    if [[ -n "$release_install_lease_pid" ]] && ps -p "$release_install_lease_pid"; then
      kill "$release_install_lease_pid"
    fi
    release_install_lease_pid=''
    release_and_acquire_install_lease_atomic
  else
    acquire_install_lease_atomic
  fi

  release_install_lease_delayed_atomic &
  release_install_lease_pid=$!

  openshift-install create cluster
  ret="$?"

  echo "Installer exit with code $ret"
  tries=$((tries+1))
done
```

~~`ipi-install-install` makes several attempt to create a cluster, with regard to install lease acquisition, the execution flow performs what follows:~~

~~1. The `n`th iteration starts.~~
~~2. If a lease is being held already, then release it and acquire a new one `release_and_acquire_install_lease_atomic`.~~
  ~~2a. Otherwise acquire one `acquire_install_lease_atomic`.~~
~~3. Start a process in the background that release the just acquired lease after 20m `release_install_lease_delayed_atomic &`.~~
~~4. Create a cluster `openshift-install create cluster`.~~
~~5. Go to (1) if (4) fails.~~
~~6. Otherwise exit the loop and complete the execution.~~
~~7. Release any pending lease `trap 'release_install_lease_atomic' EXIT TERM INT`.~~

~~Since at least two processes are involved in this workflow, the functions `acquire_lease` and `release_lease` are atomic and they rely on the [flock](https://man7.org/linux/man-pages/man1/flock.1.html) synchronization primitive.~~

The lease proxy client scripts are always available `source "$LEASE_PROXY_CLIENT_SH"`, see https://github.com/openshift/ci-tools/pull/5010.
They have been defined in https://github.com/openshift/release/pull/75306.

### DIFF
--- a/ci-operator/step-registry/ipi/install/install/ipi-install-install-commands.sh
+++ b/ci-operator/step-registry/ipi/install/install/ipi-install-install-commands.sh
@@ -4,6 +4,10 @@ set -o nounset
 set -o errexit
 set -o pipefail
 
+# shellcheck disable=SC1090
+source "$LEASE_PROXY_CLIENT_SH"
+install_lease_handle=''
+
 function set-cluster-version-spec-update-service() {
     local payload_version
     local jsonpath_flag
@@ -234,6 +238,8 @@ function write_install_status() {
 }
 
 function prepare_next_steps() {
+  printf 'Releasing install lease (if any)\n'
+  lease__release --handle="$install_lease_handle" || true
   write_install_status
   set +e
   echo "Tear down the backgroup process of copying kube config"
@@ -576,6 +582,22 @@ function get_arch() {
   echo "${ARCH}"
 }
 
+function release_install_lease() {
+  lease__release --delay=20m --handle="$install_lease_handle"
+  printf 'Install lease released at %s\n' "$(date "+%F %X")"
+}
+
+function refresh_install_lease() {
+  if ! lease__install_lease_eligible; then
+      return 0
+  fi
+
+  lease__release --handle="$install_lease_handle"
+  install_lease_handle=$(lease__acquire --type="${CLUSTER_PROFILE_SET_NAME}--${CLUSTER_PROFILE_NAME}--install-quota-slice" --scope=step)
+  printf 'Install lease acquired at %s: %s\n' "$(date "+%F %X")" "$(lease__cat --handle="$install_lease_handle" --format=csv)"
+  release_install_lease &
+}
+
 trap 'CHILDREN=$(jobs -p); if test -n "${CHILDREN}"; then kill ${CHILDREN} && wait; fi' TERM
 trap 'prepare_next_steps' EXIT TERM INT
 
@@ -853,6 +875,7 @@ do
     date "+%F %X" > "${SHARED_DIR}/CLUSTER_INSTALL_START_TIME"
   fi
 
+  refresh_install_lease || true
   copy_kubeconfig_minimal "${dir}" &
   copy_kubeconfig_pid=$!
   ${INSTALLER_BINARY} --dir="${dir}" create cluster 2>&1 | grep --line-buffered -v 'password\|X-Auth-Token\|UserData:' &


ci-operator/step-registry/ipi/install/install/ipi-install-install-commands.sh
```diff
@@ -847,10 +923,13 @@ do
     cp -rfpv "$backup" "$dir"
   else
     date "+%F %X" > "${SHARED_DIR}/CLUSTER_INSTALL_START_TIME"
+    acquire_install_lease_atomic
```
jupierce:
```suggestion
    acquire_install_lease_atomic || true
```
Perhaps? I really don't care if this fails. If it works, it will have a positive impact. If it doesn't it shouldn't change the situation.

danilo-gemoli: That's fine, so this just an optimization and the whole installation process shouldn't fail if we can't acquire such a lease.

ci-operator/step-registry/ipi/install/install/ipi-install-install-commands.sh
```diff
@@ -820,15 +826,85 @@ case $CLUSTER_TYPE in
...
+}
+export -f release_install_lease_delayed_atomic
+
+function release_and_acquire_install_lease_atomic() {
+    lease_enabled || return 0
+    flock "$INSTALL_LEASE_FLOCK" -c 'release_install_lease && acquire_install_lease'
+}
+export -f release_and_acquire_install_lease_atomic
+
+trap 'release_install_lease_atomic' EXIT TERM INT
```
stbenjam: This trap overwrites previous traps set (prepare_next_steps) which I think will cause problems

Simple example

```bash
#!/bin/bash
function prepare_next_steps() {
    echo "prepare_next_steps called"
}

function release_install_lease_atomic() {
    echo "release_install_lease_atomic called"
}

trap 'prepare_next_steps' EXIT TERM INT
trap 'release_install_lease_atomic' EXIT TERM INT
echo "End, watch which trap fires"

File: ci-operator/step-registry/ipi/install/install/ipi-install-install-commands.sh [hunk 3]
```diff
@@ -820,15 +826,85 @@ case $CLUSTER_TYPE in
...
     ;;
 esac
+
+INSTALL_LEASE_ENABLED=0
+if [[ $CLUSTER_PROFILE_SET_NAME == openshift-org-* ]]; then
+    INSTALL_LEASE_ENABLED=1
+fi
+export INSTALL_LEASE_ENABLED
+
+export RELEASE_LEASE_DELAY=20m
```
stbenjam: If 50 jobs start at the exact same time (e.g. release controller) and all get a lease, the installers will likely all be doing the same things at the same time on the same cloud provider.  They'll then all release around the exact same time, and another 50 might start.

We might want to stagger lease acquisition randomly, by having a delay _before_ we acquire the lease. I tried to solve a similar problem in https://github.com/openshift/release-controller/pull/737, but it is simpler to do here.

```bash
  delay=$(( RANDOM % 901 ))
  printf 'Waiting %dm%ds before acquiring install lease\n' $(( delay / 60 )) $(( delay % 60 ))
  sleep $delay
  acquire_install_lease_atomic
```

jupierce: That could be an optimization, but the intent here is to ratchet down to a known sustainable number of concurrent installers. Both in terms of count & duration. Once we find the sweet spot, introducing jitter could allow us to increase count, but at this point, non-determinism could confuse the ratcheting process.

ci-operator/step-registry/ipi/install/install/ipi-install-install-commands.sh
jupierce: Agree. We can use trap chaining. Something like:
```
add_trap() {
    local new_cmd="$1"
    local signal="$2"
    
    # Extract the current trap command
    local existing_cmd
    existing_cmd=$(trap -p "$signal" | sed "s/trap -- '\(.*\)' $signal/\1/")
    
    if [[ -z "$existing_cmd" ]]; then
        trap "$new_cmd" "$signal"
    else
        # Prepend or append; usually appending is safer for cleanup
        trap "$existing_cmd; $new_cmd" "$signal"
    fi
}
```

ci-operator/step-registry/ipi/install/install/ipi-install-install-commands.sh
stbenjam: > non-determinism could confuse the ratcheting process

What would be confused? Jitter before install start would be more valuable than limiting concurrent installs, IMHO.  There's a huge number of aggregated jobs that hit 1-2 minute blips in build infrastructure that end up killing the entire payload because we go below the threshold we need for statistical confidence.  Not to mention the thundering herd on specific cloud resources (e.g. all RC jobs creating load balancers at the same time)

stbenjam: Without jitter, I think this PR will make the situation worse.  As it applies globally to everything installed for that cloud provider, we're going to start triggering more installs to occur in simultaneous waves.

Imagine we limit to 50 concurrent installs.

* RC triggers 50 jobs.  Over those 20 minutes, 200 more pile up. 200 jobs are now sitting in a "wait" state (instead of starting at least offset by each other a limit bit)

* The moment those first 50 leases expire (at exactly $t + 20$ minutes), another 50 will start the exact same time.

* 20 minutes later, 50 more start.

Instead of a chaotic but distributed flow -- with peaks and valleys -- you’ve created a square wave pattern. You will see 100% utilization of the lease bucket, and *always* have high numbers of installs starting at the exact same time, compounding the problem of "all the installers doing the same thing in the cloud at the same time"

PR #76239, opened 2026-03-13
Make kueue upgrade tests optional

sohankunkerkar:


ci-operator/config/openshift/kueue-operator/openshift-kueue-operator-release-1.3__upgrade-from-4.18.yaml
```diff
@@ -48,6 +48,8 @@ resources:
       memory: 4Gi
 tests:
 - as: e2e-upgrade-4-18-to-4-19-kueue-1-3
+  optional: true
+  skip_if_only_changed: ^.*$
```
kannon92: Can we just remove this field?

Let's just make the job optional entirely

PR #76243, opened 2026-03-13
add restricted label for jira/skip-dependent-bug-check

shannon:
This PR is to setup the restriction on the label recently added to the jira-lifecycle-bot. 

https://github.com/openshift-eng/jira-lifecycle-plugin/pull/154

Sustaining mostly needs this but I saw no reason not to allow the other typically included teams the same access.


shannon:
@Prucek I added it but I wasn't sure about the specifics.

PR #76245, opened 2026-03-13
Add support for payload agent to do reverts

stbenjam:
- Rename from analysis -> agent
- Optional env variable to enable revets
- Configure jira and github
- Ensure all artifacts are copied to the ARTIFACT_DIR

Example https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/76245/rehearse-76245-periodic-ci-openshift-release-main-payload-analysis-test-claude-payload-analysis-test/2032526688004345856

Note: the revert flow is turned off for now, so it works the same as it does now. I'll test it with another PR's rehearsals.


ci-operator/step-registry/openshift/release/analysis/claude-payload-agent/openshift-release-analysis-claude-payload-agent-commands.sh
```diff
@@ -4,7 +4,33 @@ set -o nounset
 set -o errexit
 set -o pipefail
 
-echo "Starting claude-payload-analysis for payload: ${PAYLOAD_TAG}"
+echo "Starting claude-payload-agent for payload: ${PAYLOAD_TAG}"
+
+# Load secrets with xtrace disabled to prevent leaking credentials in logs
```
sosiouxme, line 9: Arguably not a problem since I believe logs are scrubbed of known secrets.
But it's a good idea to prevent them being written to begin with.
To that end, this comment is slightly misleading; not sure what "loading" secrets means but with bash, any deferencing of the variables will be written in logs, so it's not just the section immediately after here that is a concern. Suggest simply:

```suggestion
# Disable xtrace to prevent leaking credentials in logs
```

ci-operator/step-registry/openshift/release/analysis/claude-payload-agent/openshift-release-analysis-claude-payload-agent-commands.sh
```diff
@@ -107,16 +133,29 @@ echo "Invoking Claude to analyze payload ${PAYLOAD_TAG}..."
...
+    fi
+}
+trap copy_reports EXIT TERM INT
+
 # Install the must-gather plugin for analyzing must-gather archives
 echo "Installing must-gather plugin..."
 claude plugin install must-gather@ai-helpers
 echo "must-gather plugin installed."
 
+ALLOWED_TOOLS="Bash Read Write Edit Grep Glob WebFetch WebSearch Task Skill"
```
sosiouxme, line 152: If it can write and execute bash then I guess there's not much restriction on what it can do...

sosiouxme (approved):
TBH the bash feels like it's gotten hairy enough that it's not really the right tool anymore. bash makes good glue code but this is getting to where there's logic and text processing... would rather see it as python or something.

However this PR need not be the occasion for such a refactor. And my other comments are just thoughts. Good to merge as-is when you're ready.

/hold

ci-operator/step-registry/openshift/release/analysis/claude-payload-agent/openshift-release-analysis-claude-payload-agent-commands.sh
stbenjam: It's in a container at least, and doesn't have r/w to anything external.  Once I've got a bigger set of runs against a diversity of payloads, I can try to lock it down more, but my initial attempts weren't that successful.

PR #76272, opened 2026-03-15
Clean: Remove unused release image

raelga:
The aro-hcp-e2e workflow does not use the release:latest image at all. Every step uses the aro-hcp-e2e-tools base image only.

This should make jobs a bit faster and more reliable, as this image pull tends to fail like in https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-Azure-ARO-HCP-main-periodic-stage-e2e-parallel/2033000265778466816


raelga:
/pj-rehearse periodic-ci-Azure-ARO-HCP-main-periodic-prod-e2e-parallel

raelga:
This doesn't work:


>   The _cluster_profile_ (e.g., `aro-hcp-dev`) provides cloud credentials and configuration, not an image. The implicit dependency on `release:latest` comes from ci-operator's internal graph resolution: When a `cluster_profile` is present, `ci-operator` assumes we might need cluster installation tooling from the release image stream.    
>                    
> Unfortunately, this is a ci-operator requirement: if `cluster_profile` is set, `release:latest` must be defined. The release image isn't "used" by our steps, but ci-operator's dependency graph requires it to be satisfiable.

PR #76295, opened 2026-03-16
CNTRLPLANE-2981: Add azure encryption key  for self-managed azure job

wewang58:


ci-operator/step-registry/hypershift/azure/run-e2e-self-managed/hypershift-azure-run-e2e-self-managed-commands.sh
bryan-cox: We should not be reusing things from the ARO HCP tests.

We need to make the new infrastructure in the self managed Azure subscription as well.

wewang58: Thanks your guide, Now created  a dynamic step 'azure-provision-vault-key-self-managed' to provision a dedicated Key Vault strictly within the CI job's temporary Resource Group,and consumed by run-e2e step

wewang58:
/pj-rehearse pull-ci-openshift-hypershift-main-e2e-azure-self-managed

PR #76296, opened 2026-03-16
skip EgressFirewall case in azure custom dns pre-submit job

jinyunma:
Test case failed in azure custom dns pre-submit job.
```
[sig-network][Feature:EgressFirewall] when using openshift ovn-kubernetes should ensure egressfirewall is created [Suite:openshift/conformance/parallel]
[sig-network][OCPFeatureGate:DNSNameResolver][Feature:EgressFirewall] when using openshift ovn-kubernetes should ensure egressfirewall with wildcard dns rules is created [Suite:openshift/conformance/parallel] 
```

Checked with ovn QE, it is known issue, and described in openshift doc, since related feature is in TP and no plan to GA, so she suggested skipping this case.
https://docs.redhat.com/en/documentation/openshift_container_platform/4.21/html/network_security/egress-firewall#domain-name-server-resolution-ovn-k_configuring-egress-firewall-ovn

> Because the egress firewall controller and pods asynchronously poll the same local name server, the pod might obtain the updated IP address before the egress controller does, which causes a race condition. Due to this current limitation, domain name usage in EgressFirewall objects is only recommended for domains with infrequent IP address changes.

### DIFF
--- a/ci-operator/config/openshift/installer/openshift-installer-main.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-main.yaml
@@ -632,6 +632,9 @@ tests:
         for a Gateway object \| The HAProxy router should pass the http2 tests \|
         The HAProxy router should pass the gRPC interoperability tests \| Services
         should fail health check node port if there are only terminating endpoints
+        \| when using openshift ovn-kubernetes should ensure egressfirewall is created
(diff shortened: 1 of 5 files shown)


jinyunma:
/pj-rehearse pull-ci-openshift-installer-main-e2e-azure-custom-dns-techpreview

PR #76321, opened 2026-03-16
OCPERT-154: Add JIRA_USERNAME env to Jira Notificator

tomasdavidorg:
https://redhat.atlassian.net/browse/OCPERT-154

With the migration to Atlassian Cloud, the authentication switches from token_auth to basic_auth (username + API token).


rioliu-rh:
observed jira api capability issue in rehearsal job
```
jira.exceptions.JIRAError: JiraError HTTP None
	text: The `search` API is deprecated in Jira Cloud. Use `enhanced_search_issues` method instead.
```
we can fix it in another PR

PR #76327, opened 2026-03-16
Adding the `driver-toolkit` as a blocking job to nightly testing.

ybettan:
Following up on 1f3172b192a3a9e2bd54575e88311a5ed3dd0db8, also making it a blocking job for 4.12 <= OCP <= 4.15.

The driver-toolkit image contain the kernel RPMs that matches the exact kernel version on the nodes for a giver OCP payload.

Having a kernel mismatch between RHCOS and DTK will break every user using DTK including the NVIDIA GPU operator.

### DIFF
--- a/core-services/release-controller/_releases/release-ocp-4.13.json
+++ b/core-services/release-controller/_releases/release-ocp-4.13.json
@@ -48,6 +48,12 @@
       "upgrade": true,
       "maxRetries": 3
     },
+    "driver-toolkit": {
+        "maxRetries": 2,
+        "prowJob": {
+            "name": "periodic-ci-openshift-release-main-nightly-4.13-e2e-aws-driver-toolkit"
+        }
+    },
     "gcp-ovn-rt-upgrade-4.13-minor": {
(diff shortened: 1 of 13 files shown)


core-services/release-controller/_releases/release-ocp-4.13.json
```diff
@@ -48,6 +48,13 @@
       "upgrade": true,
       "maxRetries": 3
     },
+    "driver-toolkit": {
+        "disabled": true,
```
sdodson: I've just noticed that we didn't actually get the jobs in the release controller in 4.16 - 4.18 like we thought, example https://amd64.ocp.releases.ci.openshift.org/releasestream/4.18.0-0.nightly/release/4.18.0-0.nightly-2026-03-16-101030

I think we need to remove these lines in 4.12-4.18. It looks like everything in priv directory is disabled so I guess we don't need those.

ybettan: Hopefully I did the right change.

PR #76359, opened 2026-03-16
OCPBUGS-78551: Temporarily account for nginx image index change in k8s 1.35

jacobsee:


jacobsee:
/retitle OCPBUGS-78551: Temporarily account for nginx image index change in k8s 1.35

PR #76376, opened 2026-03-17
adjust gcp all regions/machine types testing

jianli-wei:
- for a pariticular OverallResult "FAIL" test, only re-run every 7 days
- split the GCP all regions & machine types testing into 2 parts, one includes storage tests, another doesn't
- lower test frequency for 4.21, 4.23 and 5.0, i.e. two runs every three days; only higher frequency for 4.22, i.e. three runs every two days


jianli-wei:
@gpei @yunjiang29 please help review, thanks!

jianli-wei:
@gpei @yunjiang29 please help review, thanks!

cc @jinyunma

jianli-wei:
@liangxia could you help approve?

PR #739, opened 2018-03-29
Bring autoscaler YAML in sync with the deployed objects

stevekuznetsov:
Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

/cc @sjenning 
/assign @smarterclayton

PR #76396, opened 2026-03-17
Fix odh-model-controller E2E: use PR manifests instead of base branch

jlost:
## Summary

- For presubmit E2E tests (`e2e-odh-kserve`), fetch the PR's manifests via GitHub's `pull/<number>/head` ref instead of always using the base branch tip
- Fixes binary/manifest mismatch where the controller image is built from PR source but kustomize manifests come from the base branch, causing crashes when PRs modify webhooks, CRDs, or API versions
- Follows up on #76097 which fixed the kustomize path resolution but introduced the branch-tip-only clone

## Details

PR #76097 fixed a kustomize path issue by cloning `odh-model-controller` into a relative location. However, it hardcoded `--branch main` (or `incubating`/`stable-2.x`), so presubmit jobs always deployed manifests from the branch tip rather than the PR under test.

This caused failures like [PR #739](https://github.com/opendatahub-io/odh-model-controller/pull/739) where the controller binary expected v1alpha1 APIs but the base branch manifests registered v1alpha2 webhooks, leading to controller crashes and "no endpoints available" errors.

The fix adds a conditional fetch for presubmit jobs:
```bash
if [ -n "${PULL_NUMBER:-}" ]; then
  pushd /tmp/kserve/odh-model-controller
    git fetch --depth 1 origin "pull/${PULL_NUMBER}/head" && git checkout FETCH_HEAD
  popd
fi
```

- **Presubmit**: fetches PR head ref so manifests match the binary
- **Postsubmit/periodic**: unchanged, uses `--depth 1 --branch <name>` (branch tip IS the correct code)

`PULL_NUMBER` is a [documented Prow environment variable](https://docs.prow.k8s.io/docs/jobs/#job-environment-variables) available only in presubmit jobs.

## Test plan

- [ ] Trigger `e2e-odh-kserve` on an odh-model-controller PR that modifies manifests (e.g. webhook configs) and verify the controller starts successfully
- [ ] Verify postsubmit image-mirror jobs are unaffected

Made with [Cursor](https://cursor.com)

### DIFF
--- a/ci-operator/config/opendatahub-io/odh-model-controller/opendatahub-io-odh-model-controller-main.yaml
+++ b/ci-operator/config/opendatahub-io/odh-model-controller/opendatahub-io-odh-model-controller-main.yaml
@@ -106,6 +106,11 @@ tests:
 
           git clone --depth 1 --branch main \
             https://github.com/opendatahub-io/odh-model-controller.git /tmp/kserve/odh-model-controller
+          if [ -n "${PULL_NUMBER:-}" ]; then
+            pushd /tmp/kserve/odh-model-controller
+              git fetch --depth 1 origin "pull/${PULL_NUMBER}/head" && git checkout FETCH_HEAD
+            popd
+          fi
           export ODH_MC_MANIFEST_SOURCE=../../odh-model-controller/config/base
 
           export GITHUB_SHA="${PULL_BASE_SHA:-stable}"
--- a/ci-operator/config/opendatahub-io/odh-model-controller/opendatahub-io-odh-model-controller-incubating.yaml
+++ b/ci-operator/config/opendatahub-io/odh-model-controller/opendatahub-io-odh-model-controller-incubating.yaml
@@ -106,6 +106,11 @@ tests:
 
           git clone --depth 1 --branch incubating \
             https://github.com/opendatahub-io/odh-model-controller.git /tmp/kserve/odh-model-controller
+          if [ -n "${PULL_NUMBER:-}" ]; then
+            pushd /tmp/kserve/odh-model-controller
+              git fetch --depth 1 origin "pull/${PULL_NUMBER}/head" && git checkout FETCH_HEAD
+            popd
+          fi
           export ODH_MC_MANIFEST_SOURCE=../../odh-model-controller/config/base
 
           export GITHUB_SHA="${PULL_BASE_SHA:-stable}"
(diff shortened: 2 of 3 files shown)


ci-operator/config/opendatahub-io/odh-model-controller/opendatahub-io-odh-model-controller-main.yaml
```diff
@@ -106,6 +106,11 @@ tests:
 
           git clone --depth 1 --branch main \
             https://github.com/opendatahub-io/odh-model-controller.git /tmp/kserve/odh-model-controller
+          if [ -n "${PULL_NUMBER:-}" ]; then
+            pushd /tmp/kserve/odh-model-controller
+              git fetch --depth 1 origin "pull/${PULL_NUMBER}/head" && git checkout FETCH_HEAD
+            popd
+          fi
```
pierDipi, line 113: why are we cloning a repository from which CI has already checkout the PR code from?

pierDipi: isn't it always going to just be `export ODH_MC_MANIFEST_SOURCE=./config/base`

jlost: I looked inside the CI job container during execution, expecting to find a checkout somewhere, and wasn't able to. It's why the [original iteration](https://github.com/openshift/release/pull/75828/changes) of this change didn't work.

jlost: As to why it wasn't there, I'm not sure. It's possible I overlooked something. It might warrant deeper investigation after we've unblocked.

pierDipi: isn't the reason that we have a pushd in the Kserve checkout directory, which makes it look there is not checkout of the PR codebase?

pierDipi: I see why, we have `from: odhmc-kserve-tests`

pierDipi: can't we use `src` similar to Kserve jobs?https://github.com/openshift/release/blob/6ba70569eedc48a3d3eb1c643002ed9d81fa7ca6/ci-operator/config/opendatahub-io/kserve/opendatahub-io-kserve-master.yaml#L233

PR #76400, opened 2026-03-17
DRAFT: opendatahub-operator: consolidate presubmit jobs and mark images optional

macgregor:
## Summary

Reduce required always-run presubmits from 5 to 2. The build04 cluster is under heavy load (35.9s avg pod scheduling latency, 16% of pods waiting >60s to schedule), causing rate limiter and timeout failures across sourceStep, importReleaseStep, and projectDirectoryImageBuildStep. Each presubmit job independently rebuilds operator images from source since ci-operator provides no cross-job image sharing. Consolidating jobs reduces duplicate image builds, shortens overall CI runtime, and lowers our exposure to infrastructure failures on the shared cluster.

## Changes

- Add `OO_BUNDLE` dependency to mirror jobs so they also validate bundle builds
- Suppress auto-generated `ci-bundle-*` presubmits via `.config.prowgen`
- Mark images `optional: true` so the auto-generated `images` presubmit is non-blocking. The image build step is generated with no way to disable it like we did with the generated bundle step, so this is the best we can do. this is still a redundant build step, but marking as optional allows us to ignore it.
- Add 45m timeout to mirror jobs (p95: ~19 min, max observed: ~33 min)

Postsubmit mirrors, E2E tests, and image/bundle definitions are unchanged.

## Verification

- No `promotion:` config exists -- presubmit images are not published anywhere
- Postsubmit mirrors and Konflux pipelines are independent and unaffected
- E2E tests build their own images/bundles via existing dependencies
- Bundle validation is preserved via `OO_BUNDLE` dependency in the mirror jobs

### DIFF
--- a/ci-operator/config/opendatahub-io/opendatahub-operator/opendatahub-io-opendatahub-operator-main.yaml
+++ b/ci-operator/config/opendatahub-io/opendatahub-operator/opendatahub-io-opendatahub-operator-main.yaml
@@ -54,7 +54,6 @@ resources:
       memory: 200Mi
 tests:
 - as: opendatahub-operator-pr-image-mirror
-  timeout: 45m0s
   steps:
     dependencies:
       SOURCE_IMAGE_REF: opendatahub-operator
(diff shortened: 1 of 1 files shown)


ci-operator/config/opendatahub-io/opendatahub-operator/opendatahub-io-opendatahub-operator-main.yaml
```diff
@@ -83,7 +97,19 @@ tests:
       SOURCE_IMAGE_REF: opendatahub-operator-rhoai
     env:
       IMAGE_REPO: opendatahub-operator
+    pre:
+    - as: validate-bundle
```
davidebianchi, line 101: This need to wait until the bundle is correctly created?

PR #76403, opened 2026-03-17
opendatahub-operator: consolidate into two pipeline jobs

macgregor:
## Summary

Consolidate 7 presubmit jobs into 2 full-pipeline jobs (`e2e-odh`, `e2e-rhoai`). Each job builds images, validates the bundle, mirrors to quay.io, claims a cluster, installs, and runs e2e — reusing all artifacts within the job via ci-operator's pipeline ImageStream. Cuts total Docker builds per PR push from 27 to 13.

## Changes

- Replace standalone mirror, e2e, and bundle presubmits with `e2e-odh` and `e2e-rhoai`
- Create `opendatahub-io-ci-image-mirror-generic-claim` ref — a duplicate of `opendatahub-io-ci-image-mirror` that clears `$KUBECONFIG` before `oc registry login` so it works inside `generic-claim` workflows (where `$KUBECONFIG` points to the claimed cluster's client-cert auth, not the build cluster)
- Add the new ref as a test step in each consolidated job
- Suppress auto-generated `ci-bundle-*` presubmits via `.config.prowgen`
- Mark images `optional: true` (auto-generated `images` presubmit cannot be suppressed by prowgen)
- Postsubmit mirrors unchanged (still use the original `opendatahub-io-ci-image-mirror` workflow)

### DIFF
--- a/ci-operator/step-registry/opendatahub-io/ci/image-mirror-generic-claim/opendatahub-io-ci-image-mirror-generic-claim-ref.metadata.json
+++ b/ci-operator/step-registry/opendatahub-io/ci/image-mirror-generic-claim/opendatahub-io-ci-image-mirror-generic-claim-ref.metadata.json
@@ -2,8 +2,8 @@
 	"path": "opendatahub-io/ci/image-mirror-generic-claim/opendatahub-io-ci-image-mirror-generic-claim-ref.yaml",
 	"owners": {
 		"approvers": [
-			"zdtsw",
-			"andrewballantyne"
+			"macgregor",
+			"davidebianchi"
 		]
 	}
 }


ci-operator/step-registry/opendatahub-io/ci/image-mirror/opendatahub-io-ci-image-mirror-commands.sh
```diff
@@ -103,8 +103,11 @@ cp $REGISTRY_TOKEN_FILE $config_file || {
     exit 1
 }
 
-# Allow to pull images from internal Openshift CI registry
-oc registry login || {
+# Allow to pull images from internal Openshift CI registry.
+# When this ref runs inside a generic-claim workflow, $KUBECONFIG points to the
+# claimed cluster which uses client-cert auth (no OAuth token). Override it so
+# oc uses the build cluster's service account token for registry authentication.
+KUBECONFIG='' oc registry login || {
```
davidebianchi: This script is used also by other repositories of the org, could be a problem?

macgregor: yes, i know i thoroughly pushed on this during implementation (it was one of the reasons i originally tried to leave out the mirror step entirely :smile: ), it should be a no-op for the others using ipi claims, and if any of them switch to cluster pools with generic-claim it will work for them there as well. you can also see in the /pj-rehearse above it ran through the others, so it should work but there is risk, yes.

I tried to avoid it, its a build system quirk. Only way is to skip the mirror step or duplicate and separate it. I can make a separate one for just us if you want.

ci-operator/jobs/opendatahub-io/opendatahub-operator/opendatahub-io-opendatahub-operator-main-presubmits.yaml
```diff
@@ -1,115 +1,5 @@
 presubmits:
   opendatahub-io/opendatahub-operator:
-  - agent: kubernetes
```
macgregor, line 3: about all these removals in this file:
* ci-bundle-opendatahub-operator-bundle and ci-bundle-opendatahub-operator-rhoai-bundle -- auto-generated by prowgen to validate bundle builds.  Suppressed by the new .config.prowgen (skip_operator_presubmits). The bundles still build as part of the e2e jobs (the install step depends on OO_BUNDLE), so they're validated there.

ci-operator/step-registry/opendatahub-io/ci/image-mirror/opendatahub-io-ci-image-mirror-commands.sh
davidebianchi: If should not be a problem I'm good also with keeping only one script instead of duplicate it. Was just to confirm that.

Otherwise, if we want to be safer, what about reuse the original script and add an env var as `CLUSTER_CLAIM_MODE=true`?

macgregor: so i actually already split them apart before you made that comment. contained blast radius now to just us.

ci-operator/step-registry/opendatahub-io/ci/image-mirror-generic-claim/OWNERS
```diff
@@ -0,0 +1,3 @@
+approvers:
+- zdtsw
+- andrewballantyne
```
davidebianchi: I'd change the owners

ci-operator/step-registry/opendatahub-io/ci/image-mirror-generic-claim/opendatahub-io-ci-image-mirror-generic-claim-ref.metadata.json
```diff
@@ -0,0 +1,9 @@
+{
+	"path": "opendatahub-io/ci/image-mirror-generic-claim/opendatahub-io-ci-image-mirror-generic-claim-ref.yaml",
+	"owners": {
+		"approvers": [
+			"zdtsw",
+			"andrewballantyne"
```
davidebianchi: I think it's needed to update also there

davidebianchi (approved):

PR #76428, opened 2026-03-17
feat: add e2e conformance test step for Konflux on OpenShift

amisstea:
Add a new konflux-ci-e2e-tests step ref that runs the e2e conformance test suite after the operator is installed on OpenShift. The step sources test environment defaults, configures credentials, deploys test resources, and runs the conformance suite with ginkgo.

Wire the new step into the main, ocp420, and ocp420-arm64 test configs so it runs after konflux-ci-install-operator.

Assisted-by: Claude claude-opus-4-6

### DIFF
--- a/ci-operator/step-registry/konflux-ci/e2e-tests/konflux-ci-e2e-tests-commands.sh
+++ b/ci-operator/step-registry/konflux-ci/e2e-tests/konflux-ci-e2e-tests-commands.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+SECRETS_DIR=/usr/local/ci-secrets/konflux-ci-konflux-ci-e2e-tests-credentials
+
+# Source the e2e env template to pick up defaults
+source test/e2e/e2e.env.template
+
+# Override e2e test credentials
+export GH_TOKEN GH_ORG QUAY_DOCKERCONFIGJSON RELEASE_CATALOG_TA_QUAY_TOKEN GITHUB_TOKEN MY_GITHUB_ORG
+GH_TOKEN=$(cat "${SECRETS_DIR}/github-token")
+GH_ORG=$(cat "${SECRETS_DIR}/github-org")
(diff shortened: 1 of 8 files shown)


ci-operator/step-registry/konflux-ci/e2e-tests/konflux-ci-e2e-tests-commands.sh
```diff
@@ -0,0 +1,40 @@
...
+# Deploy test resources
+export SKIP_SAMPLE_COMPONENTS="true"
+./deploy-test-resources.sh
+
+# Run E2E conformance tests
+echo "Running E2E conformance tests..."
+cd test/go-tests
+# The CI build image sets GOFLAGS=-mod=vendor, but konflux-ci doesn't vendor test dependencies.
+unset GOFLAGS
+go test ./tests/conformance -v -timeout 30m -ginkgo.vv
```
gbenhaim: maybe wrap it in a script ? so if we need to make any change we can do it only in konflux-ci

amisstea: I created https://github.com/konflux-ci/konflux-ci/pull/6075 for this.

ci-operator/step-registry/konflux-ci/e2e-tests/konflux-ci-e2e-tests-ref.yaml
kelchen123, line 1: I recall during standup you mentioned the tests could take 1hr 30min or more; should the timeout be increased? If most of the time is for cluster provisioning it should be fine but just checking!

amisstea: That includes the cluster claim process and operator installation which have their own timeouts. The actual test suite has a 30m timeout declared within the `run-e2e.sh` script so the timeout here just needed to be a bit bigger than than. 1h is overkill but I think that's ok - the tests should fail faster.

PR #76444, opened 2026-03-18
Adding openshift-online/rosa-gap-analysis job

ravitri:
Based on [SREP-3942](https://redhat.atlassian.net/browse/SREP-3942), adding the initial CI configuration to run the Gap Analysis automation. For now, the default comparison is stable version with candidate version and runs daily (we will likely change frequency to weekly for candidate and daily for nightly). But the


ravitri:
/hold

Adding hold. Will fix some of the repo configuration first

PR #76460, opened 2026-03-18
Migrate rhoso and netobserv jobs to their own directory structure

rsevilla87:
For better manageability going forward these jobs have to live in their own organization structure

### DIFF
--- a/ci-operator/config/openshift-eng/perfscale-netobserv/openshift-eng-perfscale-netobserv-perf-tests.yaml
+++ b/ci-operator/config/openshift-eng/perfscale-netobserv/openshift-eng-perfscale-netobserv-perf-tests.yaml
@@ -48,4 +48,4 @@ tests:
 zz_generated_metadata:
   branch: perf-tests
   org: openshift-eng
-  repo: perfscale-netobserv
+  repo: ocp-qe-perfscale-ci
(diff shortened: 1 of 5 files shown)


ci-operator/config/openshift-eng/perfscale-netobserv/openshift-eng-perfscale-netobserv-perf-tests.yaml
```diff
@@ -46,6 +46,6 @@ tests:
     - ref: netobserv-perf-test-deploy
     workflow: openshift-qe-installer-aws
 zz_generated_metadata:
-  branch: netobserv-perf-tests
+  branch: perf-tests
```
memodi, line 49: I don't understand this change, it needs to be on branch `netobserv-perf-tests` since our code/configs lives here: https://github.com/openshift-eng/ocp-qe-perfscale-ci/tree/netobserv-perf-tests

rsevilla87: Correct, I'll work on a different approach then, the main idea is to improve the dir organization to prevent potential conflicts.

PR #76467, opened 2026-03-18
Update skip_if_only_changed for WA e2e tests

rajivnathan:


ci-operator/config/codeready-toolchain/workload-analyzer/codeready-toolchain-workload-analyzer-master.yaml
```diff
@@ -30,7 +30,7 @@ resources:
       memory: 250Mi
 tests:
 - as: e2e
-  skip_if_only_changed: .*pkg/rules/rulesets/.*
+  skip_if_only_changed: .*pkg/rules/rulesets/.*|^\.github/|\.[mM][dD]$|^\.gitignore$|^\.govulncheck\.yaml$|^OWNERS$|^LICENSE$
```
MatousJobanek, line 33: I guess that you need to run some makefile target to propagate the change also here: https://github.com/openshift/release/blob/2fe8c1b2bd2a7f8a851de38551f3c748355b4141/ci-operator/jobs/codeready-toolchain/workload-analyzer/codeready-toolchain-workload-analyzer-master-presubmits.yaml#L20

In addition to that, the regex seems to be a bit broader than what is in WA now - they should match otherwise the PR stays blocked because the GHA workflow won't override the status of the job

rajivnathan: Thanks!

Regarding the job, see https://github.com/openshift/release/pull/76467/commits/f8d560f9bd91fb0c4d5edf5abb00cb339e504b18 

And for the regex, yeah it's a bit broader. I planned to open a WA PR to make them match. See https://github.com/codeready-toolchain/workload-analyzer/pull/344

PR #76481, opened 2026-03-18
Add Depends-On install_yamls support to kuttl CI step

lmiccini:
The build step already supports Depends-On for install_yamls PRs, but the kuttl step clones install_yamls independently without checking for cross-repo dependencies. This adds the same Depends-On handling so that kuttl tests can use a custom install_yamls PR branch when specified in the operator PR body.


stuggi (approved):
/lgtm

PR #76484, opened 2026-03-18
Right-size opendatahub cluster pool: smaller workers, larger pool

macgregor:
Right-sizes the `opendatahub-ocp-4-19-amd64-aws` cluster pool based on 90 days of CI utilization data. Workers are heavily overprovisioned while the pool is too small, causing a 37-41% cluster claim failure rate.

| Change | Before | After | Justification |
|--------|--------|-------|---------------|
| Worker instance | m5.2xlarge | m5.xlarge | Peak utilization: 28% CPU, 29% memory. Halving raises peaks to ~57% — safe headroom. |
| Pool size | 12 | 20 | 37-41% claim failure rate (549/1492 attempts). Busiest day: 174 E2E builds, ~25-38 concurrent. |
| maxSize | 40 | 48 | Burst ceiling for release crunches. |
| runningCount | 2 | 3 | Avg claim wait is 14 min (p90: 44 min). More warm clusters reduce resume latency. |

Smaller workers partially offset the cost of the larger pool.

/cc @opendatahub

### DIFF
--- a/clusters/hosted-mgmt/hive/pools/opendatahub/install-config-aws_secret.yaml
+++ b/clusters/hosted-mgmt/hive/pools/opendatahub/install-config-aws_secret.yaml
@@ -1,6 +1,4 @@
 # Install config template for all opendatahub cluster pools.
-# Workers use m5.2xlarge (8 vCPU, 32GB) to match our current n2-standard-8
-# GCP instances used in the IPI workflow.
 apiVersion: v1
 kind: Secret
 metadata:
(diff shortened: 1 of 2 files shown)


clusters/hosted-mgmt/hive/pools/opendatahub/install-config-aws_secret.yaml
```diff
@@ -16,7 +14,7 @@ stringData:
       name: worker
       platform:
         aws:
-          type: m5.2xlarge
+          type: m5.xlarge
```
davidebianchi, line 17: Are you sure? Are we checking the total utilization, or the resource requested in the cluster? (using m5.xlarge, I hit the limit in odh-gitops: https://github.com/opendatahub-io/odh-gitops/blob/main/.tekton/helm-chart-validation-ocp-4.19.yaml#L42). It depends also on the number of nodes

PR #76496, opened 2026-03-19
NO-JIRA: update kubernetes-autoscaler test configs

elmiko:
This change updates the autoscaler configurations to use the new makefile targets.


elmiko:
this won't work until https://github.com/openshift/cluster-api-actuator-pkg/pull/471 has merged

PR #76518, opened 2026-03-19
COO-1730: Bump observability-operator default presubmit config to OCP 4.21

DavidRajnoha:
Update base_images, build_root, releases, and cluster_claim versions from 4.15 to 4.21 to align the default e2e test pipeline with current supported releases.

Made-with: Cursor


ci-operator/config/rhobs/observability-operator/rhobs-observability-operator-main.yaml
lihongyan1: why not upgrade to 4.21 also?

PR #76519, opened 2026-03-19
Fix shellcheck SC2154 warning in upi-install-heterogeneous step

lmiccini:
Variables name and mac are dynamically sourced from YAML output, which shellcheck cannot trace. Add SC2154 to the existing disable directive.


jianlinliu:
let us have a quick merge, it blocks other PRs.

/lgtm

PR #76538, opened 2026-03-19
Lease Proxy Client: Add many improvements

danilo-gemoli:
This PR has been tailored specifically to handle the use use we have in https://github.com/openshift/release/pull/76238, the script is now more reliable, robust and secure.

I believe that providing few examples of how to use the `lease__*` functions defined would help the reviewers more than trying to describe the technicalities that have been introduced.

#### Use Case 1 - Acquire and release
Script:
```sh
lease_handle="$(lease__acquire --type=openshift-org-aws --count=2)"
printf 'Acquired leases: %s\n' "$(lease__cat --handle="$lease_handle" --format=csv)"
lease__release --handle="$lease_handle"
```
Output:
```
Acquired leases: openshift-org-aws--quota-slice-00,openshift-org--quota-slice-01
openshift-org-aws--quota-slice-00 openshift-org-aws--quota-slice-01 released
```
Description:
`--scope=step` means that the leases are scoped to the current step and must be released before it completes. This is the default behavior.

`lease_handle` is an opaque file handle that can be used to:
- Print the lease names that have been acquired.
- Release them.

Under the hood is just a temporary file, but here we provide some helper functions like `lease__cat` to deal with it without being aware of the implementation details.

#### Use Case 2 - Acquire in a step and release in a different one
Script:
```sh
# Step 1
lease_handle="$(lease__acquire --type=openshift-org-aws --count=2 --scope=test)"
printf 'Acquired leases: %s\n' "$(lease__cat --handle="$lease_handle" --format=csv)"

# Step 2
lease__release --scope=test
```

Output:
```sh
Acquired leases: openshift-org-aws--quota-slice-00,openshift-org-aws--quota-slice-01
openshift-org-aws--quota-slice-00 openshift-org-aws--quota-slice-01 released
```
Description:
`--scope=test` means that the leases are scoped to the entire life-cycle of a test, therefore they can be acquired from a step (as an example: `ipi-install-install`) and released later on, possibly in post steps.

#### Use Case 3 - Deferred acquire and release
Script:
```sh
lease_handle="$(lease__acquire --type=openshift-org-aws --count=2 --jitter=15m)"
printf 'Acquired leases: %s\n' "$(lease__cat --handle="$lease_handle" --format=csv)"
lease__release --delay=20m --handle="$lease_handle"
```

Output:
```sh
# `--jitter=15m` waits up to 15m (random value in the range [0, 15m]) before acquiring the leases
Acquired leases: openshift-org-aws--quota-slice-00,openshift-org-aws--quota-slice-01

# `--delay=20m` pauses for 20m before releasing them
openshift-org-aws--quota-slice-00 openshift-org-aws--quota-slice-01 released
```
Description:
`--jitter` and `--delay` pauses the execution for the specified amount of time before acquiring and/or releasing.

#### Use Case 4 - Refresh install leases
Script:
```sh
install_lease_handle=''
trap "lease__release --handle=\"$install_lease_handle\"" EXIT TERM INT

function refresh_install_lease() {
    if ! lease__install_lease_eligible; then
        return 0
    fi

    lease__release --handle="$install_lease_handle"
    install_lease_handle=$(lease__acquire --type="openshift-org-aws" --scope=step)
    printf 'install lease acquired: %s\n' "$(lease__cat --handle="$install_lease_handle" --format=csv)"
    lease__release --delay=20m --handle="$install_lease_handle" &
}

max=3
retries=0
while [[ $retries -lt $max ]]; do
    refresh_install_lease || true

    if openshift-install create cluster; then
        echo 'The cluster has been created'
        break
    fi

    ((++$retries))
done
```

Output:
```sh
Acquired leases: openshift-org-aws--install-quota-slice-00
# After 20m
openshift-org-aws--install-quota-slice-00 released
```
Description:
`lease__install_lease_eligible` checks whether a test using a `CLUSTER_PROFILE_SET` is eligible for acquiring an install lease. So far only the cluster profile sets that match `openshift-org-*` are eligible: see https://github.com/openshift/release/pull/76068 and https://github.com/openshift/release/pull/76340.

The following lines refresh a lease:
```sh
lease__release --handle="$install_lease_handle"
install_lease_handle=$(lease__acquire --type="openshift-org-aws" --scope=step)
lease__release --delay=20m --handle="$install_lease_handle" &
```
The `lease__release` is idempotent and thread safe, therefore the following facts hold:
- Calling `lease__release --handle="$install_lease_handle"` several times is safe: leases are released only once.
- Concurrent executions of `lease__release --handle="$install_lease_handle"` are safe.

/cc @jupierce @stbenjam

### DIFF
--- a/ci-operator/lease/proxy-client.sh
+++ b/ci-operator/lease/proxy-client.sh
@@ -1,44 +1,41 @@
-# Download jq if it is not already available.
+# Internal use only: download jq if it is not already available.
 #
 # Result:
 #   Return 0, export a new PATH env variable having the jq download directory as first entry.
 #   Return 1 and print an error message otherwise.
-function lease__ensure_jq() {
+function _lease__ensure_jq() {
     if command -v jq &>/dev/null; then
         return 0
     fi
 
-    local ec=0
-    local response_body="$(mktemp --tmpdir=/tmp ensure-jq-XXXXX)" || ec=$?
-    if [ $ec -ne 0 ]; then
-        printf 'Failed to create temp file: %s\n' "$response_body"
+    local response_body="$(mktemp --tmpdir=/tmp ensure-jq-XXXXX)" || {
+        printf 'Failed to create temp file\n'
         return 1
-    fi
+    }
 
     local response="$(curl --connect-timeout 300 --max-time 600 \
+      --retry 5 --retry-delay 10 --retry-all-errors \
       --no-progress-meter -sL \
       -o "$response_body" \
       --write-out '%{response_code}' \
       -H "Accept: application/vnd.github+json" \
       -H "X-GitHub-Api-Version: 2022-11-28" \
-      https://api.github.com/repos/jqlang/jq/releases)" || ec=$? 
-    if [ $ec -ne 0 ]; then
+      https://api.github.com/repos/jqlang/jq/releases)" || {
         printf 'Failed to determine jq releases from %s\n%s\n' \
           'https://api.github.com/repos/jqlang/jq/releases' \
           "$response"
         rm -f "$response_body" &>/dev/null
         return 1
-    fi
+    }
 
-    grep -qP '2\d{2}' <<<"$response" || ec=$?
-    if [ $ec -ne 0 ]; then
+    grep -qP '2\d{2}' <<<"$response" || {
         printf 'Failed to determine jq releases from %s\n%s\n' \
           'https://api.github.com/repos/jqlang/jq/releases' \
           "$response"
         cat "$response_body"
         rm -f "$response_body" &>/dev/null
         return 1
-    fi
+    }
 
     local jq_arch=''
     case $(uname -m) in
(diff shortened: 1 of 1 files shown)


ci-operator/lease/proxy-client.sh
```diff
@@ -170,31 +170,52 @@ function lease__acquire() {
...
+        local max=$(( $jitter_value + 1 ))
+        local start_delay=$(( $RANDOM % $max ))
+        sleep "${start_delay}" || {
+            printf 'Failed to delay the execution\n'
+            return 1
+        }
     fi
 
+    local ec=0
     local response=$(curl --no-progress-meter -X POST -o "$response_body" \
```
stbenjam, line 217: Should we have a retry mechanism? If Boskos has a brief outage, I wouldn't want to skip the lease or error out

ci-operator/lease/proxy-client.sh
```diff
@@ -1,44 +1,40 @@
...
-    local response_body="$(mktemp --tmpdir=/tmp ensure-jq-XXXXX)" || ec=$?
-    if [ $ec -ne 0 ]; then
-        printf 'Failed to create temp file: %s\n' "$response_body"
+    local response_body="$(mktemp --tmpdir=/tmp ensure-jq-XXXXX)" || {
+        printf 'Failed to create temp file\n'
         return 1
-    fi
+    }
 
     local response="$(curl --connect-timeout 300 --max-time 600 \
```
stbenjam, line 16: Same comment here about a retry, although hopefully we're just using images with jq already

ci-operator/lease/proxy-client.sh
```diff
@@ -170,31 +170,52 @@ function lease__acquire() {
...
+            jitter_value="${BASH_REMATCH[1]}"
+            jitter_unit="${BASH_REMATCH[2]}"
+            if [[ "$jitter_unit" == "m" ]]; then
+                jitter_value=$(( jitter_value * 60 ))
+            fi
+        else
+            printf "jitter parameter is invalid\n"
+            return 1
+        fi
+    fi
```
stbenjam, line 188: Thank you! This is great

ci-operator/lease/proxy-client.sh
danilo-gemoli: We have `--retry 5 --retry-delay 10 --retry-all-errors` now. Transient errors like `CONN_REFUSED` or `HTTP 500` force curl to retry, whereas errors like `HTTP 404` do not. Those are treated like non-transient by curl.

ci-operator/lease/proxy-client.sh
danilo-gemoli: Same as above.

PR #76543, opened 2026-03-19
fix(dast): apply prerequisites CRDs for TALM

oblau:
TALM (topology-aware-lifecycle-manager) crashes at startup when its required ACM/MCE API groups are not present in the cluster. Add a new step `telcov10n-functional-dast-talm-prereq-crds` that applies minimal stub CRDs for those API groups before install-operators runs. Remove `skip_checking: true` workaround from the TALM operator entry in all DAST lanes (4.16–4.21) now that the CRDs are applied beforehand.

### DIFF
--- a/ci-operator/step-registry/telcov10n/functional/dast/talm-prereq-crds/telcov10n-functional-dast-talm-prereq-crds-ref.metadata.json
+++ b/ci-operator/step-registry/telcov10n/functional/dast/talm-prereq-crds/telcov10n-functional-dast-talm-prereq-crds-ref.metadata.json
@@ -1,0 +1,11 @@
+{
+	"path": "telcov10n/functional/dast/talm-prereq-crds/telcov10n-functional-dast-talm-prereq-crds-ref.yaml",
+	"owners": {
+		"approvers": [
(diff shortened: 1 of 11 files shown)


ci-operator/step-registry/telcov10n/functional/dast/talm-prereq-crds/telcov10n-functional-dast-talm-prereq-crds-ref.metadata.json
```diff
@@ -0,0 +1,13 @@
+{
+	"path": "telcov10n/functional/dast/talm-prereq-crds/telcov10n-functional-dast-talm-prereq-crds-ref.yaml",
+	"owners": {
+		"approvers": [
+			"natifridman",
```
shaior: Hi Omer, 
you may remove Nati and Ofer since they are not part of the team. 
thanks.

oblau: done

PR #76552, opened 2026-03-19
Use ironic-base for all Dockerfile stages in ironic image builds

elfosardo:
After openshift/release#76345 switched rhel9 repos from EUS to dist content, intermediate multi-stage build targets lost access to packages only available in the ironic-specific repos. Add inputs.as to replace all FROM references with the ironic-base image that has ironic repos baked in.


joepvd:
/pj-rehearse pull-ci-openshift-ironic-image-main-prevalidation-images
/lgtm

PR #76572, opened 2026-03-20
[INTEROP-8947] OCP 4.22 CR and FIPS expansions for ACS stable 4.10 interop scenario.

dfrazzette:
Also disabling ACS 4.9 stable on OCP 4.21 interop scenario by changing the cron value.


dfrazzette:
These changes have been split into [PR#77933](https://github.com/openshift/release/pull/77933)  and [PR#78151](https://github.com/openshift/release/pull/78151).

PR #76597, opened 2026-03-20
rhdh: add cluster health check and increase login timeout for hibernating clusters

gustavolira:
## Summary

- Increase `oc login` timeout from 5 to 10 minutes to handle clusters resuming from hibernation
- Increase retry interval from 20s to 30s to reduce log noise
- Add cluster health check (`oc get nodes`) after login to verify API server is fully responsive before proceeding with service account creation

## Context

The `e2e-ocp-v4-20-helm-nightly` job failed because the ephemeral OCP cluster was resuming from hibernation. The 5-minute login timeout was insufficient, and there was no validation that the API server was functional after login succeeded.

Most cluster pools (v4.16, v4.17, v4.19, v4.20, v4.21) have no `runningCount`, so clusters hibernate between jobs. Increasing `runningCount` was rejected due to AWS cost since nightly jobs only run 3x/week.

Build log: https://storage.googleapis.com/test-platform-results/logs/periodic-ci-redhat-developer-rhdh-release-1.9-e2e-ocp-v4-20-helm-nightly/2034140407427764224/build-log.txt

Ref: [RHIDP-12736](https://issues.redhat.com/browse/RHIDP-12736)

## Changes

Applied to all 3 OCP helm step scripts:
- `ocp/helm/nightly` (nightly jobs)
- `ocp/helm` (PR checks)
- `ocp/helm/upgrade/nightly` (upgrade jobs)

## Test plan

- [ ] Verify nightly jobs pass on next scheduled run
- [ ] Confirm health check logs appear in build output (`Cluster API server is ready`)

### DIFF
--- a/ci-operator/step-registry/redhat-developer/rhdh/ocp/helm/nightly/redhat-developer-rhdh-ocp-helm-nightly-commands.sh
+++ b/ci-operator/step-registry/redhat-developer/rhdh/ocp/helm/nightly/redhat-developer-rhdh-ocp-helm-nightly-commands.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set +x
 
 echo "========== Repository, Branch, and PR Variables =========="
 GITHUB_ORG_NAME="redhat-developer"
@@ -64,8 +65,6 @@ else
     exit 1
 fi
 
-# Disable tracing to protect credentials from leaking into CI logs
-set +x 2>/dev/null
 if ! timeout --foreground 10m bash <<-"EOF"; then
     while ! oc login "$OPENSHIFT_API" -u "$OPENSHIFT_USERNAME" -p "$OPENSHIFT_PASSWORD" --insecure-skip-tls-verify=true; do
             echo "Login failed, retrying in 30s..."
@@ -77,17 +76,12 @@ EOF
 fi
 
 echo "========== Cluster Health Check =========="
-echo "Verifying cluster API server is fully responsive..."
-if ! timeout --foreground 5m bash <<-"EOF"; then
-    while ! oc get nodes 2>&1; do
-        echo "API server not ready, retrying in 15s..."
-        sleep 15
-    done
-EOF
-    echo "Timed out waiting for cluster API server to become ready"
+echo "Waiting for all nodes to be ready..."
+if ! oc wait --for=condition=Ready nodes --all --timeout=300s; then
+    echo "Timed out waiting for nodes to become ready"
     exit 1
 fi
-echo "Cluster API server is ready"
+echo "All nodes are ready"
 
 echo "========== Cluster Service Account and Token Management =========="
 export K8S_CLUSTER_URL K8S_CLUSTER_TOKEN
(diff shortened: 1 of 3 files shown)


ci-operator/step-registry/redhat-developer/rhdh/ocp/helm/nightly/redhat-developer-rhdh-ocp-helm-nightly-commands.sh
```diff
@@ -64,16 +64,31 @@ else
     exit 1
 fi
 
-timeout --foreground 5m bash <<-"EOF"
+# Disable tracing to protect credentials from leaking into CI logs
+set +x 2>/dev/null
```
zdrapela: We don't use `set` anywhere in the CI scripts, so the `set +x` is present by default. If you want to configure it explicitly, it makes sense to do it at the beginning of the `commands.sh` bash script? Because this is not the only sensitive data here.

ci-operator/step-registry/redhat-developer/rhdh/ocp/helm/nightly/redhat-developer-rhdh-ocp-helm-nightly-commands.sh
```diff
@@ -64,16 +64,31 @@ else
...
+            sleep 30
     done
 EOF
-if [ $? -ne 0 ]; then
     echo "Timed out waiting for login"
     exit 1
 fi
 
+echo "========== Cluster Health Check =========="
+echo "Verifying cluster API server is fully responsive..."
```
zdrapela: This is more of a check for nodes to be ready, so why not echo what it actually is? 
Maybe also `oc wait --for=condition=Ready nodes --all --timeout=XXXs` may be handy here?

ci-operator/step-registry/redhat-developer/rhdh/ocp/helm/nightly/redhat-developer-rhdh-ocp-helm-nightly-commands.sh
gustavolira: Moved set +x to the top of the script since we don't use set -x anywhere

ci-operator/step-registry/redhat-developer/rhdh/ocp/helm/nightly/redhat-developer-rhdh-ocp-helm-nightly-commands.sh
gustavolira: Replaced the polling loop with `oc wait --for=condition=Ready nodes --all --timeout=300s` as you suggested, and updated the messages to reflect what it actually does

zdrapela: Did you push the commit?

gustavolira: Sorry I forgot to push, it's done now

PR #76664, opened 2026-03-23
USHIFT-6741: Fix bootc release periodic jobs for 4.22

vanhalenar:


ci-operator/config/openshift/microshift/openshift-microshift-release-4.22__periodics.yaml
```diff
@@ -186,24 +186,44 @@ tests:
...
       TEST_EXECUTION_TIMEOUT: 60m
     workflow: openshift-microshift-e2e-metal-tests
-- as: e2e-aws-tests-bootc-release-arm-periodic
+- as: e2e-aws-tests-bootc-release-periodic-el10
+  cron: 0 3 * * 0,2,4
+  steps:
+    cluster_profile: openshift-org-aws
+    env:
+      EC2_INSTANCE_TYPE: c5.metal
+      MICROSHIFT_OS: rhel-9.6
```
agullon, line 205: why `rhel-9.6` here? tbh, I'm not sure if this env var is used

ci-operator/config/openshift/microshift/openshift-microshift-release-4.22__periodics.yaml
```diff
@@ -186,24 +186,44 @@ tests:
...
+      SCENARIO_TYPE: bootc-releases-el10
+      TEST_EXECUTION_TIMEOUT: 60m
+    workflow: openshift-microshift-e2e-metal-tests
+- as: e2e-aws-tests-bootc-release-arm-periodic-el9
+  cron: 0 3 * * 0,2,4
+  steps:
+    cluster_profile: openshift-org-aws
+    env:
+      EC2_INSTANCE_TYPE: c7g.metal
+      MICROSHIFT_OS: rhel-9.6
```
agullon, line 215: why `rhel-9.6` here? tbh, I'm not sure if this env var is used

ci-operator/config/openshift/microshift/openshift-microshift-release-4.22__periodics.yaml
vanhalenar: It was `rhel-9.6` before, I left it like it is, it's 9.6 in all the other jobs as well. If I remember correctly that's the host OS

agullon: you are right, this is only for the EC2 machine, so `rhel-9.6` is correct here, thanks. We can merge the PR.

ci-operator/config/openshift/microshift/openshift-microshift-release-4.22__periodics.yaml
agullon: this is only for the EC2 machine, so `rhel-9.6` is correct here, thanks. We can merge the PR.

PR #76718, opened 2026-03-24
[SREP-3529] reduce the upgrade test frequency and set default value to skip must gather and destroy cluster

bmeng:
Reduce the upgrade related testing for now as they are not running properly

will need to review the tests

Also set the default value to skip the must gather and destroy cluster from the test

### DIFF
--- a/ci-operator/config/openshift/osde2e/openshift-osde2e-main.yaml
+++ b/ci-operator/config/openshift/osde2e/openshift-osde2e-main.yaml
@@ -178,7 +178,7 @@ tests:
   container:
     clone: true
     from: osde2e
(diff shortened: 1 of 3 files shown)


ci-operator/config/openshift/osde2e/openshift-osde2e-main.yaml
```diff
@@ -178,7 +178,7 @@ tests:
   container:
     clone: true
     from: osde2e
-  cron: 0 12 * * 1,5
+  cron: 0 12 * * 1
```
ritmun, line 181: we should move these periodics out of this main branch yaml so you can self service them in future

PR #76723, opened 2026-03-24
NO-JIRA: hypershift: slim down e2e footprint

sjenning:
There are two primary changes

The first is the restoration of the `hypershift-aws-e2e-nested` workflow.

The `hypershift-aws-e2e-external` workflow will henceforth be used for jobs that run on the latest promoted HO.  These will run on a shared persistent management cluster. These jobs include:
- o/h e2e periodic jobs
- external e2e presub jobs

The `hypershift-aws-e2e-nested` workflow will work as the old workflow used to work, deploying a per job management cluster, install HO from the `hypershift-operator` tagged image, and running the e2e within that cluster.  These jobs include:
- o/h e2e presub jobs
- periodic jobs that require the HO be installed with tech-preview

The result is that a great many jobs will not require a dedicated management cluster any more.

The second is the default test filter for all jobs that use the `hypershift-aws-e2e-external` flow is limited to just `TestCreateCluster` and `TestUpgradeControlPlane`.  While this reduces coverage, it would catch almost all regressions I've ever seen from an external components and greatly reduce the infra footprint of these jobs.

This PR requires changes to (if root CI) or creation of (if new cluster) the shared persistent management cluster
- Run external-dns
- Run 3 az nodepools for HA HostedClusters


ci-operator/config/openshift/hypershift/openshift-hypershift-release-4.21__periodics.yaml
```diff
@@ -140,11 +140,7 @@ tests:
   - build-tmpfs
   cron: 0 0 1 1 *
   steps:
-    cluster_profile: hypershift-aws
-    env:
-      ENABLE_HYPERSHIFT_CERT_ROTATION_SCALE: "true"
-      REQUEST_SERVING_COMPONENT_TEST: "true"
-    workflow: hypershift-aws-e2e-metrics
+    workflow: hypershift-aws-e2e-external
 - as: e2e-aws-ovn-conformance-serial
```
LiangquanLi930, line 145: `cluster_profile: hypershift-aws` has been removed.

sjenning: Good catch! I was experimenting with this job and didn't restore it.

PR #76808, opened 2026-03-25
DPTP-4267: Encrypt s3 storage used in image registry

liangxia:
Several clusters has set `encrypt: true` for s3 storage in image registry, while several of the other clusters do not.  Unify them to `encrypt: true`.
```
$ oc --context build05 get configs.imageregistry.operator.openshift.io cluster -oyaml | yq '.spec.storage.s3.encrypt'
null

$ oc --context build06 get configs.imageregistry.operator.openshift.io cluster -oyaml | yq '.spec.storage.s3.encrypt'
true
```


liangxia:
Pod scheduling timeout. 

/retest-required

PR #76832, opened 2026-03-25
Fix kueue-operator slack notifications on failures

MaysaMacedo:
Notes:
Assisted with Claude

### DIFF
--- a/ci-operator/config/openshift/kueue-operator/.config.prowgen
+++ b/ci-operator/config/openshift/kueue-operator/.config.prowgen
@@ -4,7 +4,7 @@ slack_reporter:
   job_states_to_report:
   - failure
   - error
-  report_template: ':failed: Job *{{.Spec.Job}}* ended with *{{.Status.State}}*. <{{.Status.URL}}|View logs> {{end}}'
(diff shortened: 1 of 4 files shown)


ci-operator/config/openshift/kueue-operator/.config.prowgen
```diff
@@ -4,7 +4,7 @@ slack_reporter:
   job_states_to_report:
   - failure
   - error
-  report_template: ':failed: Job *{{.Spec.Job}}* ended with *{{.Status.State}}*. <{{.Status.URL}}|View logs> {{end}}'
+  report_template: ':failed: Job *{{.Spec.Job}}* ended with *{{.Status.State}}*. <{{.Status.URL}}|View logs>'
```
anahas-redhat, line 7: Hi @MaysaMacedo. 
I was checking crio project (it posts alerts on #forum-node-jira) and, it seems the {{end}} is used. 
This is where I'm checking: https://github.com/openshift/release/blob/main/ci-operator/config/cri-o/cri-o/.config.prowgen 

Is it possible to do some kind of rehearsal.. or test.. just to see if after this change, the posts are going to work?

cpmeadors: @anahas-redhat Looks like the {{end}} is needed there because of the {{if ...}}.  Nothing to end in the kueue template.

PR #76834, opened 2026-03-25
Add image build config for  openstack-lightspeed rag-content

malingatembo:
This pr enables building of openstack-lightspeed
RAG content vector database in Prow, replacing
github actions which has a time limit of 6hours.

- add images: container file references
- configures build arguments for vector db generation
- updates test to a message indicating successful build
- Manual promotion to Quay (no auto-push yet)


ci-operator/config/openstack-lightspeed/rag-content/openstack-lightspeed-rag-content-main.yaml
```diff
@@ -24,11 +24,60 @@ resources:
...
+    echo "====================================="
+    echo "Image: rag-content-openstack"
+    echo "Index: os-docs-2025.2"
+    echo "Docs: Upstream OpenStack + Operators"
+    echo "====================================="
+    echo "Build completed successfully!"
   container:
-    from: src
+    from: rag-content-openstack
+  timeout: 8h0m0s
```
umago: 8h sounds a bit too much for this ?

ci-operator/config/openstack-lightspeed/rag-content/openstack-lightspeed-rag-content-main.yaml
```diff
@@ -24,11 +24,60 @@ resources:
     requests:
       cpu: "4"
       memory: 8Gi
+images:
+- dockerfile_path: Containerfile
+  to: rag-content-openstack
+  build_args:
+  - name: FLAVOR
+    value: cpu
```
umago: I think @lpiwowar mentioned GPU somewhere in prow, I just want to double-check this because with GPU the CI would be much much faster

ci-operator/config/openstack-lightspeed/rag-content/openstack-lightspeed-rag-content-main.yaml
```diff
@@ -24,11 +24,60 @@ resources:
...
+  - name: NUM_WORKERS
+    value: "4"
+  - name: OS_VERSION
+    value: "2025.2"
+  - name: BUILD_UPSTREAM_DOCS
+    value: "true"
+  - name: BUILD_OPERATORS_DOCS
+    value: "true"
```
umago: I think we will need to remove this for the upstream RAG, the behavior for it will change to match RHOSO and OCP docs ingestion (using the pre-rendered repo) and it won't be available to upstream RAG for now.

ci-operator/config/openstack-lightspeed/rag-content/openstack-lightspeed-rag-content-main.yaml
malingatembo: shall we drop it to 5 hours or even less  if GPUs will be used...

ci-operator/config/openstack-lightspeed/rag-content/openstack-lightspeed-rag-content-main.yaml
malingatembo: I'll wait to have this verified. thanks for pointing it out

ci-operator/config/openstack-lightspeed/rag-content/openstack-lightspeed-rag-content-main.yaml
umago: I think so, but let's clarify the GPU use first. Cause it will affect this timeout for sure.

ci-operator/config/openstack-lightspeed/rag-content/openstack-lightspeed-rag-content-main.yaml
malingatembo: thanks @umago . i have updated the config based on your feedback. I have set BUILD_OPERATORS_DOCS and BUILD_UPSTREAM_DOCS to false. 

quick one: what is the timeline into prerendered state?

umago: The last piece is this one https://github.com/openstack-lightspeed/rag-content/pull/100. Should be soon

ci-operator/config/openstack-lightspeed/rag-content/openstack-lightspeed-rag-content-main.yaml
malingatembo: Alright, i will wait for clarification from @lpiwowar , when he gets the time to take a look at this

ci-operator/config/openstack-lightspeed/rag-content/openstack-lightspeed-rag-content-main.yaml
lpiwowar: I do not remember exactly where I saw the configuration for GPU the last time when I was working in this repo. But I believe this should be the way how to get access to GPU (just a quick search):

https://github.com/openshift/release/blob/b63bcb762cb3db4c29a3b0ea0fab39e1455c59da/ci-operator/config/opendatahub-io/vllm/opendatahub-io-vllm-main.yaml#L72

https://github.com/openshift/release/blob/b63bcb762cb3db4c29a3b0ea0fab39e1455c59da/ci-operator/config/opendatahub-io/vllm/opendatahub-io-vllm-main.yaml#L50

I'm not sure whether it is really simple like this, but I believe it is worth taking a look.

umago: Nice! If turns out it's not that simple, let's go ahead with CPU @malingatembo. I don't think we need to block this patch GPU vs CPU.

If anything, we can improve it later. Regarding the timeout, perhaps keep it 8h as well and we see how long it takes to build with CPU. And we can adjust it later too.

malingatembo: @umago and @lpiwowar thank you for the responses concerning the use of the GPUs. I will go ahead and propose use of the CPU in the pr... Evenually, like we discussed, in person, we will try for the GPU ( when all is working as expected )

ci-operator/config/openstack-lightspeed/rag-content/openstack-lightspeed-rag-content-main.yaml
```diff
@@ -8,6 +8,42 @@ build_root:
...
+images:
+- build_args:
+  - name: FLAVOR
+    value: cpu
+  - name: NUM_WORKERS
+    value: "4"
+  - name: OS_VERSION
+    value: "2025.2"
+  - name: BUILD_UPSTREAM_DOCS
+    value: "false"
```
umago: BUILD_UPSTREAM_DOCS should be "true" no ? Otherwise what we will be building here ?

ci-operator/config/openstack-lightspeed/rag-content/openstack-lightspeed-rag-content-main.yaml
```diff
@@ -25,10 +61,19 @@ resources:
...
+- as: build-rag-content-image
+  commands: |
+    echo "====================================="
+    echo "RAG Content Vector DB Image Build"
+    echo "====================================="
+    echo "Image: rag-content-openstack"
+    echo "Index: os-docs-2025.2"
+    echo "Docs: Minimal build (waiting for pre-rendered repos)"
+    echo "====================================="
+    echo "Build completed successfully!"
```
umago: Sorry, but why we want to print these ? It's under "tests" so I don't know if we want to hardcode some of the parameters (like Index:)  or not.

ci-operator/config/openstack-lightspeed/rag-content/openstack-lightspeed-rag-content-main.yaml
umago, line 117: I was also wondering, as part of this job, do we also want to push the final image to Quay. io ? Like we current do with GH Workflows [0]

[0] https://github.com/openstack-lightspeed/rag-content/blob/82efb82d7e7a736c1f56a513f1cec3044be59e91/.github/workflows/build-and-push-rag-content.yaml#L99-L116

umago (changes requested):

ci-operator/config/openstack-lightspeed/rag-content/openstack-lightspeed-rag-content-main.yaml
lpiwowar: Yes, I think we want this 💯 . Also, we should ensure that there are two types of jobs:

1. Runs on every change in a PR. It runs only partial build of the image.
2. Runs after a PR gets merged and periodically every day at midnight. This one is a full build of the image and as part of this job we want to push the image to quay.io.

If it turns out that it is too much of a work then we can IMO start with no. 1. (this does not require the pushing of the image). And then another PR for no. 2.

malingatembo: @umago , @lpiwowar , thank you for the clarification. I will start looking into it  ( push the final image to Quay. iob)

PR #76864, opened 2026-03-26
Unblock build02

liangxia:
build02 upgrade is done.


core-services/sanitize-prow-jobs/_clusters.yaml
deepsm007, line 12: unblock this as well

liangxia: build03 not that healthy

PR #64410, opened 2025-04-30
max-finding for setting PPN

mehabhalodiya:
Related to https://github.com/openshift/release/pull/63477 
As the set-max-pods functionality worked in the previous PR, this PR is for finding the max PPN for the desired threshold and instance size.

(Ref. ticket https://issues.redhat.com/browse/OCPQE-28897)

PR #76889, opened 2026-03-26
add large ppn node-density-heavy job

qiliRedHat:
Add 1500 pod per node node-density-heavy tests to 4.21 and 4.22
https://redhat.atlassian.net/browse/OCPNODE-3202

### DIFF
--- a/ci-operator/config/openshift-eng/ocp-qe-perfscale-ci/openshift-eng-ocp-qe-perfscale-ci-main__4.21-nightly-node-density-heavy.yaml
+++ b/ci-operator/config/openshift-eng/ocp-qe-perfscale-ci/openshift-eng-ocp-qe-perfscale-ci-main__4.21-nightly-node-density-heavy.yaml
@@ -152,6 +152,41 @@ tests:
     test:
     - chain: openshift-qe-node-density-heavy
     workflow: baremetal-lab-upi
+- always_run: false
+  as: aws-3nodes-1500ppn
+  steps:
+    allow_skip_on_success: true
+    cluster_profile: aws-perfscale-qe
+    env:
+      BASE_DOMAIN: qe.devcluster.openshift.com
+      BURST: "10"
+      CLUSTER_NETWORK_CIDR: 10.128.0.0/14
+      CLUSTER_NETWORK_HOST_PREFIX: "20"
+      COMPUTE_NODE_TYPE: c6a.16xlarge
+      CONTROL_PLANE_INSTANCE_TYPE: m6a.4xlarge
+      ITERATIONS_PER_NAMESPACE: "100"
+      MAX_PODS: "1550"
+      NDH_EXTRA_FLAGS: --pod-ready-threshold=500s
+      OPENSHIFT_INFRA_NODE_INSTANCE_TYPE: r5.4xlarge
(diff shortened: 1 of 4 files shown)


ci-operator/step-registry/openshift-qe/node-density-heavy/openshift-qe-node-density-heavy-ref.yaml
```diff
@@ -19,10 +19,6 @@ ref:
     default: "default"
     documentation: |-
       Override the e2e version
-  - name: ADDITIONAL_WORKER_NODES
```
qiliRedHat, line 22: This was not used in this ref anywhere, could be added by mistake in https://github.com/openshift/release/pull/64410.

ci-operator/config/openshift-eng/ocp-qe-perfscale-ci/openshift-eng-ocp-qe-perfscale-ci-main__4.21-nightly-node-density-heavy.yaml
```diff
@@ -152,6 +152,41 @@ tests:
...
+    cluster_profile: aws-perfscale-qe
+    env:
+      BASE_DOMAIN: qe.devcluster.openshift.com
+      BURST: "10"
+      CLUSTER_NETWORK_CIDR: 10.128.0.0/14
+      CLUSTER_NETWORK_HOST_PREFIX: "20"
+      COMPUTE_NODE_TYPE: c6a.16xlarge
+      CONTROL_PLANE_INSTANCE_TYPE: m6a.4xlarge
+      ITERATIONS_PER_NAMESPACE: "100"
+      MAX_PODS: "2750"
```
afcollins: IF we are only sending 1500 pods per node, I think having max pods of 2750 could allow nodes to be unbalanced, which could create stability issues.

I MAX_PODS should be equal to or only slightly above PODS_PER_NODE. So 1500, or 1550

ci-operator/step-registry/openshift-qe/node-density-heavy/openshift-qe-node-density-heavy-ref.yaml
afcollins: It is used by [scale-workers step](https://github.com/openshift/release/blob/main/ci-operator/step-registry/openshift-qe/workers-scale/openshift-qe-workers-scale-ref.yaml#L35-L38). I see some node-density-heavy jobs using it, and other jobs use it as well, but I am not sure whether it needs to be declared in this -ref or not.

ci-operator/config/openshift-eng/ocp-qe-perfscale-ci/openshift-eng-ocp-qe-perfscale-ci-main__4.21-nightly-node-density-heavy.yaml
qiliRedHat: Updated to 1550

mehabhalodiya (approved):
I checked with my max pods PR https://github.com/openshift/release/pull/64410/changes , and the parameters looks same (like what worked for me earlier), so yeah, these changes looks good to me!

/lgtm

PR #76926, opened 2026-03-26
OSDOCS-18881: Add MCP Gateway docs TP CI configuration

ShaunaDiaz:
Adding Prow jobs to the `mcp-gateway-docs-tp` feature branch being used to build MCP Gateway docs for Summit.

/pj-rehearse


core-services/prow/02_config/openshift/controller-runtime-common/_prowconfig.yaml
```diff
@@ -5,6 +5,7 @@ tide:
     labels:
     - approved
     - lgtm
+    - verified
```
kalexand-rh: I wouldn't change these last three files - the first two are the only ones we need for our jobs.

PR #76928, opened 2026-03-26
Acquire install lease when deprovisioning a cluster

danilo-gemoli:
Counterpart of https://github.com/openshift/release/pull/76238.
`openshift-install destroy cluster` seems to be another API calls intensive operation like `openshift-install create cluster`.


petr-muller (approved):
LGTM

PR #76951, opened 2026-03-27
Unblocking ote ci jobs for SVMO

gangwgr:
Unblocking ote ci jobs for SVMO

### DIFF
--- a/ci-operator/config/openshift/cluster-kube-storage-version-migrator-operator/openshift-cluster-kube-storage-version-migrator-operator-release-4.22.yaml
+++ b/ci-operator/config/openshift/cluster-kube-storage-version-migrator-operator/openshift-cluster-kube-storage-version-migrator-operator-release-4.22.yaml
@@ -58,9 +58,7 @@ tests:
   steps:
     cluster_profile: openshift-org-aws
     workflow: openshift-e2e-aws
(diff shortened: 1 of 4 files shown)


ci-operator/config/openshift/cluster-kube-storage-version-migrator-operator/openshift-cluster-kube-storage-version-migrator-operator-release-4.22.yaml
```diff
@@ -58,19 +58,15 @@ tests:
   steps:
     cluster_profile: openshift-org-aws
     workflow: openshift-e2e-aws
-- always_run: false
-  as: e2e-gcp-operator-disruptive
-  optional: true
```
wangke19, line 63: are you sure this? all your ci job has been migratied? If not, when you are debugging cases, that will block your merge.

gangwgr: @wangke19 it is for network policy cases which we have already tested and merged, And in svmo only 2 cases are there, which are only operator health check. I will migrate those after this to ote

PR #76964, opened 2026-03-27
rhdh: replace QUAY_REPO override with IMAGE_REPO and IMAGE_REGISTRY

zdrapela:
## Summary
- Replace `MULTISTAGE_PARAM_OVERRIDE_QUAY_REPO` with `MULTISTAGE_PARAM_OVERRIDE_IMAGE_REPO` and add `MULTISTAGE_PARAM_OVERRIDE_IMAGE_REGISTRY` in all RHDH step-registry ref YAMLs and command scripts
- Use `IMAGE_REPO` internally in all command scripts, setting `QUAY_REPO` from it for backward compatibility with older release branches (1.8, 1.9) that still read `QUAY_REPO`
- Aligns with [redhat-developer/rhdh#4463](https://github.com/redhat-developer/rhdh/pull/4463) which makes the image registry configurable via `IMAGE_REGISTRY` and `IMAGE_REPO` env vars
https://redhat.atlassian.net/browse/RHIDP-12951


ci-operator/step-registry/redhat-developer/rhdh/aks/helm/nightly/redhat-developer-rhdh-aks-helm-nightly-commands.sh
```diff
@@ -204,46 +210,57 @@ elif [[ "$ONLY_IN_DIRS" == "true" && "$JOB_TYPE" == "presubmit" ]];then
...
 
-    ELAPSED_TIME=0
+        ELAPSED_TIME=0
 
-    while true; do
-        # Check image availability
-        response=$(curl -s "https://quay.io/api/v1/repository/${QUAY_REPO}/tag/?specificTag=$TAG_NAME")
+        while true; do
+            # Check image availability
+            response=$(curl -s "https://quay.io/api/v1/repository/${QUAY_REPO}/tag/?specificTag=$TAG_NAME")
```
gustavolira, line 227: This curl still uses `${QUAY_REPO}` but at this point `QUAY_REPO` is still `""` (initialized at the top of the script). The sync line `QUAY_REPO="${IMAGE_REPO}"` only runs after the entire `if/fi` block.

The URL becomes `https://quay.io/api/v1/repository//tag/...` -- empty repo path, no tags ever match, so the polling loop runs until timeout (60 min) and fails the job.

Should be `${IMAGE_REPO}` here -- the `osd-gcp` script got it right.

Same bug in: `eks-helm-nightly`, `gke-helm-nightly`, `ocp-helm-nightly`, `ocp-helm` (PR checks), `ocp-helm-upgrade-nightly`.

ci-operator/step-registry/redhat-developer/rhdh/send/data-router/redhat-developer-rhdh-send-data-router-ref.yaml
```diff
@@ -6,9 +6,12 @@ ref:
     name: rhdh-send-data-router
     namespace: test-credentials
   env:
-    - name: MULTISTAGE_PARAM_OVERRIDE_QUAY_REPO
+    - name: MULTISTAGE_PARAM_OVERRIDE_IMAGE_REPO
       default: ""
```
gustavolira, line 10: The ref YAML was updated to declare `MULTISTAGE_PARAM_OVERRIDE_IMAGE_REPO`, but the command script (`redhat-developer-rhdh-send-data-router-commands.sh:8`) still checks `${MULTISTAGE_PARAM_OVERRIDE_QUAY_REPO}` in the override detection array. Since this env var is no longer declared here, CI operator won't inject it, so the data-router won't detect Gangway overrides and will report results when it should skip.

Should also add `MULTISTAGE_PARAM_OVERRIDE_IMAGE_REGISTRY` to the array.

ci-operator/step-registry/redhat-developer/rhdh/aks/helm/nightly/redhat-developer-rhdh-aks-helm-nightly-commands.sh
```diff
@@ -204,46 +210,57 @@ elif [[ "$ONLY_IN_DIRS" == "true" && "$JOB_TYPE" == "presubmit" ]];then
...
+IMAGE_REGISTRY="${IMAGE_REGISTRY:-quay.io}"
+QUAY_REPO="${IMAGE_REPO}" # Keep QUAY_REPO in sync for backward compatibility
+export IMAGE_REPO IMAGE_REGISTRY QUAY_REPO
 
 echo "========== Current branch =========="
 echo "Current branch: $(git branch --show-current)"
-IMAGE_SHA=$(curl -s "https://quay.io/api/v1/repository/${QUAY_REPO}/tag/?specificTag=${TAG_NAME}" | jq -r '.tags[0].manifest_digest')
-echo "Using image: ${QUAY_REPO}:${TAG_NAME}, with digest: ${IMAGE_SHA}"
+IMAGE_SHA=$(curl -s "https://quay.io/api/v1/repository/${IMAGE_REPO}/tag/?specificTag=${TAG_NAME}" | jq -r '.tags[0].manifest_digest')
+if [[ "${IMAGE_REGISTRY}" == "quay.io" ]]; then
```
gustavolira, line 258: nit: this curl unconditionally hits quay.io even when `IMAGE_REGISTRY` is not quay.io. The conditional block below correctly skips the digest display, but the curl still runs and sets `IMAGE_SHA` to null. Could be wrapped in the same `if [[ "${IMAGE_REGISTRY}" == "quay.io" ]]` check.

zdrapela: Nice catch!

ci-operator/step-registry/redhat-developer/rhdh/aks/helm/nightly/redhat-developer-rhdh-aks-helm-nightly-commands.sh
zdrapela: True, thanks!

PR #76974, opened 2026-03-27
NO-JIRA: fix: resolve errors in sno 4vcpu lanes

eggfoobar:
fix issue with missing ocp version for dev-scripts, this is required now for 4vcpu runs 
add timeout for tests in serial

### DIFF
--- a/ci-operator/config/openshift/release/openshift-release-main__nightly-4.22.yaml
+++ b/ci-operator/config/openshift/release/openshift-release-main__nightly-4.22.yaml
@@ -504,6 +504,7 @@ tests:
     cluster_profile: openshift-org-aws
     env:
       SINGLE_NODE_AWS_INSTANCE_TYPE: r6i.xlarge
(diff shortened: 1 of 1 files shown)


ci-operator/config/openshift/release/openshift-release-main__nightly-4.22.yaml
```diff
@@ -514,6 +515,7 @@ tests:
     env:
       DEVSCRIPTS_CONFIG: |
         AGENT_E2E_TEST_SCENARIO=SNOMIN_IPV4
+        OPENSHIFT_VERSION=4.22
```
jaypoulz, line 518: do we have to update this every release? This feels not-ideal to me.

eggfoobar: This momentary for now, it just needs to be 4.22+ the script will be updated to handle this better, I just need some more time to dig into it.

PR #76986, opened 2026-03-27
mco: use pipeline in lgtm mode 

djoshy:
This updates the all active branches(>= 4.16) to use the pipeline controller.

### DIFF
--- a/ci-operator/config/openshift/machine-config-operator/openshift-machine-config-operator-main.yaml
+++ b/ci-operator/config/openshift/machine-config-operator/openshift-machine-config-operator-main.yaml
@@ -154,8 +154,9 @@ tests:
     test:
     - ref: openshift-e2e-test
     workflow: ipi-aws
-- as: e2e-aws-ovn
-  skip_if_only_changed: ^docs/|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|OWNERS_ALIASES|LICENSE)$
+- always_run: false
+  as: e2e-aws-ovn
+  pipeline_skip_if_only_changed: ^docs/|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|OWNERS_ALIASES|LICENSE)$
   steps:
     cluster_profile: openshift-org-aws
     workflow: openshift-e2e-aws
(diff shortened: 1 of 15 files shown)


isabella-janssen (commented):
/lgtm

ci-operator/config/openshift/machine-config-operator/openshift-machine-config-operator-main.yaml
```diff
@@ -269,9 +272,9 @@ tests:
       timeout: 1h40m0s
     workflow: ipi-gcp
   timeout: 4h0m0s
-- as: e2e-gcp-op-ocl
-  optional: true
```
yuqi-zhang, line 274: Wouldn't we want to keep these optional still? From the docs this seems that they will mark these as required.

ci-operator/config/openshift/machine-config-operator/openshift-machine-config-operator-main.yaml
```diff
@@ -299,9 +302,9 @@ tests:
           cpu: 100m
       timeout: 2h30m0s
     workflow: ipi-gcp
-- as: e2e-gcp-op-ocl-part2
-  optional: true
```
yuqi-zhang, line 304: same as above

core-services/pipeline-controller/lgtm-config.yaml
```diff
@@ -6,3 +6,4 @@ orgs:
       - name: hypershift
         branches:
           - main
+      - machine-config-operator
```
yuqi-zhang, line 9: we should probably specify main and release-4.21 for now? Since that's the ones you changed. I'd imagine the older branches have slightly different configuration and tests.

ci-operator/config/openshift/machine-config-operator/openshift-machine-config-operator-main.yaml
jmguzik: optional jobs are not part of pipelines anyway, so you need to keep it as it is

ci-operator/config/openshift/machine-config-operator/openshift-machine-config-operator-main.yaml
jmguzik: if optional, it has to stay as it is

core-services/pipeline-controller/lgtm-config.yaml
jmguzik: I advise you to change configs for all maintained older branches as well (to 4.16?) and just to fully onboard. Otherwise you would every release need to update this config.

ci-operator/config/openshift/machine-config-operator/openshift-machine-config-operator-main.yaml
djoshy: missed this - will revert back!

core-services/pipeline-controller/lgtm-config.yaml
djoshy: ack, will add this back to 4.16

yuqi-zhang (approved):
/lgtm

lgtm from the MCO perspective. I think technically older releases still can get changes? But should be fine to deal with on a per-case basis.

jmguzik (approved):

PR #76993, opened 2026-03-27
fix(hypershift/gcp): correct DNS zone name and surface cleanup errors

cristianoveiga:
## Summary
- The `e2e-gke` workflow had `HYPERSHIFT_GCP_CI_DNS_ZONE` set to `hypershift-ci-zone` but the actual zone in `gcp-hcp-hypershift-ci` is `hypershift-ci-gcp-hcp-openshiftapps-com`. This caused the deprovision step's DNS cleanup to silently fail, leaving orphaned DNS records.
- The `gcloud dns record-sets list` command had `2>/dev/null || true` which swallowed permission errors (403 Forbidden), making it appear that no DNS records existed. Replaced with explicit error handling that logs failures.
- **Note:** DNS cleanup also requires `roles/dns.admin` for the `hypershift-ci` service account — see openshift-online/gcp-hcp-infra#429.

## Test plan
- [ ] Verify `gcloud dns record-sets list --zone=hypershift-ci-gcp-hcp-openshiftapps-com --project=gcp-hcp-hypershift-ci` works with the corrected zone name
- [ ] After gcp-hcp-infra#429 is applied, re-rehearse and confirm DNS cleanup succeeds in deprovision logs

### DIFF
--- a/ci-operator/step-registry/hypershift/gcp/gke/deprovision/hypershift-gcp-gke-deprovision-commands.sh
+++ b/ci-operator/step-registry/hypershift/gcp/gke/deprovision/hypershift-gcp-gke-deprovision-commands.sh
@@ -98,6 +98,7 @@ gcloud projects delete "${CP_PROJECT_ID}" --quiet || true
 EXTERNAL_DNS_GSA="external-dns@${HYPERSHIFT_GCP_CI_PROJECT}.iam.gserviceaccount.com"
 
 # Clean up DNS records from the CI zone (DNS records use the hosted cluster name)
+DNS_CLEANUP_FAILED=false
 if [[ -n "${HC_CLUSTER_NAME}" ]]; then
   echo "Cleaning up DNS records for hosted cluster ${HC_CLUSTER_NAME}..."
   DNS_SUFFIX="in.${HC_CLUSTER_NAME}.${HYPERSHIFT_GCP_CI_DNS_DOMAIN}."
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/hypershift/gcp/gke/deprovision/hypershift-gcp-gke-deprovision-commands.sh
```diff
@@ -101,20 +101,25 @@ EXTERNAL_DNS_GSA="external-dns@${HYPERSHIFT_GCP_CI_PROJECT}.iam.gserviceaccount.
...
       [[ -z "${name}" ]] && continue
       echo "Deleting DNS record: ${name} ${type}"
-      gcloud dns record-sets delete "${name}" \
+      if ! gcloud dns record-sets delete "${name}" \
         --type="${type}" \
         --zone="${HYPERSHIFT_GCP_CI_DNS_ZONE}" \
-        --project="${HYPERSHIFT_GCP_CI_PROJECT}" --quiet || true
+        --project="${HYPERSHIFT_GCP_CI_PROJECT}" --quiet; then
+        echo "WARNING: Failed to delete DNS record ${name} ${type}"
+      fi
```
jimdaga, line 125: One concern;

The error handling change is better than the old `2>/dev/null || true` but there is still a chance we could end up with orphaned DNS records. If the list or delete fails, we log a WARNING and keep going, the step exits 0, and nobody notices until we've got a pile of leaked records. Should we consider failing the step so we notice?

cristianoveiga: Fair enough - this still wouldn't fail the job (because of  `best_effort: true`), but if the step fails, we would have a higher chance of someone taking a look at the logs and fix this (avoiding the orphan DNS records).

Addressed.

jimdaga: Good point, thanks!

PR #77023, opened 2026-03-29
Add native OLM e2e test for SBR CI using optional-operators workflow

ugreener:
Adds a new `openshift-e2e-olm` presubmit test that validates Storage Based Remediation (SBR) operator installation through the standard customer deployment path (CatalogSource + Subscription via the `optional-operators-ci-aws` workflow), targeting the `openshift-workload-availability` namespace.

**Changes:**
- Added `sbd-operator-bundle` bundle entry (without `skip_building_index`) to enable CI index image generation
- Added `openshift-e2e-olm` test with four steps: namespace PSA configuration, CI agent image injection into the CSV, ODF storage setup, and e2e test execution
- Uses active polling for OLM CSV propagation instead of static sleep
- Generated corresponding Prow presubmit jobs via `make update`
- The existing `openshift-e2e` test remains completely unchanged

### DIFF
--- a/ci-operator/config/medik8s/storage-based-remediation/medik8s-storage-based-remediation-main__4.20.yaml
+++ b/ci-operator/config/medik8s/storage-based-remediation/medik8s-storage-based-remediation-main__4.20.yaml
@@ -17,6 +17,8 @@ operator:
   - as: my-bundle
     dockerfile_path: bundle.Dockerfile
     skip_building_index: true
+  - as: sbd-operator-bundle
+    dockerfile_path: bundle.Dockerfile
   substitutions:
   - pullspec: quay.io/medik8s/sbd-operator:.*
     with: pipeline:storage-based-remediation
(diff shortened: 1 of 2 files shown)


ci-operator/config/medik8s/storage-based-remediation/medik8s-storage-based-remediation-main__4.20.yaml
```diff
@@ -36,67 +35,62 @@ tests:
...
-
-        # workaround for OLM pod not running with restricted PSA
-        oc label --overwrite ns "$OPERATOR_NAMESPACE" security.openshift.io/scc.podSecurityLabelSync=false
-        oc label --overwrite ns "$OPERATOR_NAMESPACE" pod-security.kubernetes.io/enforce=privileged
-
-        # OPERATOR_NAMESPACE variable is used by bundle-run and bundle-run-update
-        BUNDLE_IMG=quay.io/medik8s/sbd-operator-bundle:v"$OPERATOR_RELEASED_VERSION" make bundle-run
-      env:
-      - name: OPERATOR_NAMESPACE
-      - name: OPERATOR_RELEASED_VERSION
```
mshitrit, line 58: Doesn't that mean that we completely drop the upgrade test ?

ugreener: The upgrade test is dropped from this job, but it can be added back as a dedicated job in a follow-up PR using the `optional-operators-ci-aws-upgrade` workflow. This workflow is the standard Prow pattern for OLM upgrade testing -- it installs an initial version via `optional-operators-subscribe`, then upgrades to the CI-built version via `optional-operators-ci-upgrade`. MetalLB and File Integrity Operator already use it.

The dedicated upgrade job would use `INITIAL_CSV` and `INITIAL_CHANNEL` env vars to specify the starting version (e.g., v0.2.0), and the workflow handles the upgrade to the PR's bundle automatically.

Separating install and upgrade into two jobs is the idiomatic CI pattern -- it keeps each job focused and makes failures easier to diagnose.

mshitrit: >but it can be added back as a dedicated job in a follow-up PR

I'm not 100% clear on the suggestion of how to re-implement the test, but I think it doesn't make sense to drop a valid working test for a potential future implementation.
I don't mind a different implantation to the upgrade test, but I think it needs to be part of this PR.

ugreener: To add to the above -- the old upgrade test using `bundle-run-update` was not a real OLM upgrade. It sideloaded the bundle via operator-sdk, bypassing the CatalogSource/Subscription flow entirely. So this PR is not removing a properly functioning OLM upgrade test.

To add a proper OLM upgrade test (using the standard `optional-operators-ci-aws-upgrade` workflow), the SBR bundle first needs a `replaces: sbd-operator.v0.2.0` field in the CSV and a version bump above 0.2.0. Currently the CSV on main is `sbd-operator.v0.0.1` with no `replaces` field, so OLM has no upgrade path to follow.

Proposed path forward:
1. Merge this PR as-is (install-only, standard OLM flow)
2. Add `replaces` metadata to the SBR bundle in the SBR repo (one-line CSV change)
3. Follow-up PR here to add the upgrade job using `optional-operators-ci-aws-upgrade`

Step 3 would give us a real OLM upgrade test that validates the actual customer upgrade path -- something the old `bundle-run-update` approach never did.

razo7: >To add a proper OLM upgrade test (using the standard optional-operators-ci-aws-upgrade workflow), the SBR bundle first needs a replaces: sbd-operator.v0.2.0 field in the CSV and a version bump above 0.2.0. Currently the CSV on main is sbd-operator.v0.0.1 with no replaces field, so OLM has no upgrade path to follow.

Interesting point, even though the current upgrade test was a nice way to mimic the real way

ci-operator/config/medik8s/storage-based-remediation/medik8s-storage-based-remediation-main__4.20.yaml
```diff
@@ -107,23 +101,20 @@ tests:
...
-        export OPERATOR_NS="$OPERATOR_NAMESPACE"
+        export OPERATOR_NS="${OO_INSTALL_NAMESPACE}"
         # Use cluster profile AWS credentials so e2e tests can make EC2 API calls (e.g. ec2:DescribeInstances, ec2:RebootInstances)
         # The node's IAM role alone doesn't have these permissions
         export AWS_SHARED_CREDENTIALS_FILE="${CLUSTER_PROFILE_DIR}/.awscred"
-        make test-e2e && TEST_EXIT_CODE=0 || TEST_EXIT_CODE=$? # TEST_EXIT_CODE=$? always returns true, preventing set -e from aborting the script, while still capturing the real exit code
+        make test-e2e && TEST_EXIT_CODE=0 || TEST_EXIT_CODE=$?
         # Copy test artifacts (logs, junit report) to CI artifact directory regardless of test outcome
-        cp -r .tests/* "$ARTIFACT_DIR/" 2>/dev/null || echo "WARNING: cp failed (no artifacts to copy?)"
-        # Agents logs expected at <test-number>/artifacts/openshift-e2e/test-command/artifacts/<unique identifier>/
```
razo7, line 117: Any good reason to remove the comment?

ci-operator/config/medik8s/storage-based-remediation/medik8s-storage-based-remediation-main__4.20.yaml
```diff
@@ -36,67 +35,62 @@ tests:
...
-          -n "$OPERATOR_NAMESPACE" --timeout=120s
-      dependencies:
-      - env: OO_BUNDLE
-        name: my-bundle
-      - env: SBR_AGENT_IMAGE
-        name: sbd-agent
-      env:
-      - name: OPERATOR_NAMESPACE
+        make setup-odf-storage
+        make run-setup-odf-storage-retry
```
razo7, line 192: Do we already have these targets in upstream? If they are good, could you help me understand why we haven't used them before?

I like makefile targets as it is easier to test and verify the target result

razo7 (changes requested):
Can we avoid making changes to a running (and working) e2e test workflow?
Given that this PR and motivation are experimental (we don't know if it will work or meet our new needs/goals), I prefer not to break what is working for now. We can achieve that in two ways:

- Create an e2e test for main on OCP 4.21/4.22
- Add a new test (after  `as: openshift-e2e`) with all your new changes under OCP 4.20

Don't forget to run `make update` to create a job and run a rehearsal of the test you affected

ci-operator/config/medik8s/storage-based-remediation/medik8s-storage-based-remediation-main__4.20.yaml
mshitrit: We do and we have, the confusion is mostly from how the diff is shown (i.e deleted and added)

ci-operator/config/medik8s/storage-based-remediation/medik8s-storage-based-remediation-main__4.20.yaml
ugreener: Two comments were removed:

1. The inline comment on `TEST_EXIT_CODE` explained a standard `set -e` bash workaround. The code is self-explanatory, making the comment unnecessary.
2. The comment regarding the agent logs path referenced the old artifact directory structure. Because the script now uses `cp -a .tests/. "$ARTIFACT_DIR/"` to copy all artifacts recursively for automatic Prow indexing, the old path description is no longer accurate.

ci-operator/config/medik8s/storage-based-remediation/medik8s-storage-based-remediation-main__4.20.yaml
ugreener: The steps that have upstream Makefile targets (`setup-odf-storage`, `run-setup-odf-storage-retry`, `test-e2e`) are called via `make` in this PR.

The remaining inline commands (namespace PSA labeling and CSV agent image injection) are CI-specific scaffolding. Prow's `optional-operators-ci-aws` workflow installs the operator via OLM before the test steps run, but it does not handle post-install customizations like applying PSA labels or injecting CI-built images into the CSV. Because these steps are only required for the CI environment and are not part of the operator's normal lifecycle, they do not belong in the upstream Makefile.

This follows the pattern used by other operators in openshift/release (e.g., SR-IOV, PTP): CI-specific post-install steps remain inline in the Prow config, while operator logic remains in the Makefile.

ci-operator/config/medik8s/storage-based-remediation/medik8s-storage-based-remediation-main__4.20.yaml
```diff
@@ -124,6 +126,90 @@ tests:
...
+        oc label --overwrite ns "${OO_INSTALL_NAMESPACE}" pod-security.kubernetes.io/enforce=privileged
+      from: src
+      resources:
+        requests:
+          cpu: 100m
+          memory: 200Mi
+    - as: inject-ci-agent-image
+      cli: latest
+      commands: |
+        # Inject CI-built agent image into the CSV so the operator uses it for DaemonSet pods
```
slintes, line 159: Why do we need this...? Is this something customers need to do as well? Why doesn't the bundle step set the right image? Do we actually really use 2 separate images? So many questions.... ;)

slintes (changes requested):
Why do we target OCP 4.20? SBR main (=0.3) is for 4.22, not?
I'd suggest to make a copy of the old config for the SBR 0.2 branch (only update OCP to 4.21), and an updated config with your new deployment version only for SBR main and OCP 4.22

ci-operator/config/medik8s/storage-based-remediation/medik8s-storage-based-remediation-main__4.20.yaml
mshitrit: I'll try to answer.
- Unlike SNR SBR uses 2 separate images (operator and agent).
- When config is created SBR creates the DS dynamically with the agent image.
- SBR figures out what the agent image name would be based on the operator image.
- In our CI the "figuring out" the agent name based on the operator name isn't working due to how the images are named, so we inject the image name of the agent so it can be picked up.

So to sum it up:
>Why do we need this...?

In order for the operator to get the agent image

> Is this something customers need to do as well? 

No

>Why doesn't the bundle step set the right image?

It does, the problem is that the agent image is consumed in runtime and SBR can't figure out what it is.

>Do we actually really use 2 separate images? 

Yes

slintes: this sounds overly complex... and makes this test far away from what customers will do, while the intention was the opposite :/

slintes:
Ignore my earlier comments. This is moving CI config from a supported way of deploying the operator to a deprecated way: https://docs.ci.openshift.org/how-tos/testing-operator-sdk-operators/#building-an-index-deprecated

IMHO there also is little value in adding the extra step of building and using an index image in upstream CI. However, if QE insists on using an index image, we need a supported way of doing so.

/hold

ugreener:
The index image approach is deprecated — closing the PR.  
https://docs.ci.openshift.org/how-tos/testing-operator-sdk-operators/#building-an-index-deprecated

PR #73270, opened 2026-01-08
CORS-4294: Adding dualstack presubmit job

rna-afk:
Adding two presubmit jobs for the dualstack work in aws to support ipv6 along with ipv4. One for ipv4 and one for ipv6 primary.

Code generated by claude.

PR #77043, opened 2026-03-30
Add CI tests for ConfigurablePKI installer feature

hasbro17:
This provides the presubmits for the installer support of the PKI config https://github.com/openshift/installer/pull/10396

### Summary

The `ConfigurablePKI` feature gate changes installer-generated signer certificates from `RSA-2048` to `ECDSA P-384` by default. This PR adds CI coverage to validate both the default ECDSA behavior and explicit
`RSA-4096` override, using `TechPreviewNoUpgrade`.

**New step-registry component:**

**`openshift-installer-pki-verify`** (ref) — Checks 7 cluster-accessible signer CA secrets and the PKI CR post-install. Runs before the e2e suite to fail fast on mismatches. Outputs a pass/fail summary table
and writes full certificate details to the artifact directory.

**`ipi-conf-aws` modifications:**

Adds `PKI_ALGORITHM`, `PKI_RSA_KEY_SIZE`, `PKI_ECDSA_CURVE` env vars with empty defaults (no impact on existing jobs). When `PKI_ALGORITHM` is set, injects the `pki` section into `install-config.yaml`.

**New CI jobs (`openshift/installer`, `main` + release branches 4.22, 4.23, 5.0):**

**Optional presubmits:**
- `e2e-aws-ovn-pki-default-techpreview` — Feature gate ON, no explicit config → expects `ECDSA P-384` signers
- `e2e-aws-ovn-pki-rsa-techpreview` — Feature gate ON, explicit `RSA-4096` → expects `RSA-4096` signers

### DIFF
--- a/ci-operator/config/openshift/installer/openshift-installer-main.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-main.yaml
@@ -272,6 +272,34 @@ tests:
         keyC valueC
     workflow: openshift-e2e-aws
 - always_run: false
+  as: e2e-aws-ovn-pki-default-techpreview
+  optional: true
+  steps:
+    cluster_profile: openshift-org-aws
+    env:
+      EXPECTED_ALGORITHM: ECDSA
+      EXPECTED_KEY_PARAM: secp384r1
+      FEATURE_SET: TechPreviewNoUpgrade
+    test:
+    - ref: openshift-installer-pki-verify
+    - ref: openshift-e2e-test
+    workflow: openshift-e2e-aws
+- always_run: false
+  as: e2e-aws-ovn-pki-rsa-techpreview
+  optional: true
+  steps:
+    cluster_profile: openshift-org-aws
+    env:
+      EXPECTED_ALGORITHM: RSA
+      EXPECTED_KEY_PARAM: "4096"
+      FEATURE_SET: TechPreviewNoUpgrade
+      PKI_ALGORITHM: RSA
+      PKI_RSA_KEY_SIZE: "4096"
+    test:
+    - ref: openshift-installer-pki-verify
+    - ref: openshift-e2e-test
+    workflow: openshift-e2e-aws
+- always_run: false
   as: e2e-aws-ovn-dualstack-ipv4-primary-techpreview
   optional: true
   steps:
--- a/ci-operator/step-registry/openshift/installer/pki/OWNERS
+++ b/ci-operator/step-registry/openshift/installer/pki/OWNERS
@@ -1,0 +1,10 @@
+approvers:
+- hasbro17
+- kaleemsiddiqu
+- rh-roman
+- sanchezl
+reviewers:
+- hasbro17
+- kaleemsiddiqu
+- rh-roman
+- sanchezl
--- a/ci-operator/config/openshift/installer/openshift-installer-release-4.22.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-release-4.22.yaml
@@ -273,6 +273,34 @@ tests:
         keyC valueC
     workflow: openshift-e2e-aws
 - always_run: false
+  as: e2e-aws-ovn-pki-default-techpreview
+  optional: true
+  steps:
+    cluster_profile: openshift-org-aws
+    env:
+      EXPECTED_ALGORITHM: ECDSA
+      EXPECTED_KEY_PARAM: secp384r1
+      FEATURE_SET: TechPreviewNoUpgrade
+    test:
+    - ref: openshift-installer-pki-verify
+    - ref: openshift-e2e-test
+    workflow: openshift-e2e-aws
+- always_run: false
+  as: e2e-aws-ovn-pki-rsa-techpreview
+  optional: true
+  steps:
+    cluster_profile: openshift-org-aws
+    env:
+      EXPECTED_ALGORITHM: RSA
+      EXPECTED_KEY_PARAM: "4096"
+      FEATURE_SET: TechPreviewNoUpgrade
+      PKI_ALGORITHM: RSA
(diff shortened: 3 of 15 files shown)


ci-operator/config/openshift/installer/openshift-installer-main.yaml
```diff
@@ -255,6 +255,116 @@ tests:
...
+    - ref: ipi-conf-telemetry
+    - ref: ipi-conf-aws
+    - ref: ipi-conf-aws-byo-ipv4-pool-public
+    - ref: ipi-install-monitoringpvc
+    - chain: aws-provision-iam-user-minimal-permission
+    - ref: rhcos-conf-osstream
+    - chain: ipi-install
+    test:
+    - ref: openshift-installer-pki-verify
+    - ref: openshift-e2e-test
```
patrickdillon: You  should use the existing openshift-e2e-aws workflow and edit ipi-conf-aws to read your environment variable and set the install config fields that way. it will be simpler.

here's an example pr: https://github.com/openshift/release/pull/73270

ci-operator/config/openshift/installer/openshift-installer-main.yaml
```diff
@@ -255,6 +255,116 @@ tests:
...
+  optional: true
+  steps:
+    allow_best_effort_post_steps: true
+    cluster_profile: openshift-org-aws
+    env:
+      AWS_INSTALL_USE_MINIMAL_PERMISSIONS: "yes"
+      EXPECTED_ALGORITHM: ECDSA
+      EXPECTED_KEY_PARAM: secp384r1
+      FEATURE_GATES: '["ConfigurablePKI=true"]'
+      FEATURE_SET: CustomNoUpgrade
```
patrickdillon: The feature is in the TechPreviewNoUpgrade feature set, so you can just set the feature set to TechPreviewNoUpgrade

ci-operator/config/openshift/installer/openshift-installer-main.yaml
```diff
@@ -255,6 +255,116 @@ tests:
         keyB valueB
         keyC valueC
     workflow: openshift-e2e-aws
+- always_run: false
+  as: e2e-aws-ovn-pki-default
```
patrickdillon: I *think* `-ovn` is supposed to be at the end of the job name. not sure if that matters anymore.

ci-operator/config/openshift/installer/openshift-installer-main.yaml
hasbro17: That looks much simpler, thanks for the tip.

ci-operator/config/openshift/installer/openshift-installer-main.yaml
hasbro17: I wanted to very specifically test only the `ConfigurablePKI ` feature but I guess it would be good to see that this plays nicely with other techpreview features turned on as well.

ci-operator/config/openshift/installer/openshift-installer-main.yaml
hasbro17: I see many other presubmits/jobs using it right after the platform. E.g: 
- e2e-aws-ovn-dualstack-ipv6-primary-techpreview
- e2e-aws-ovn-proxy

Honestly not sure it's even needed but I just followed the existing convention.

yunjiang29: @hasbro17 Per the convention, `-techpreview` postfix is supposed to be at the end of the job name for the jobs enabled TP. 
And should these jobs to be added to `openshift-installer-release-4.22.yaml`/`openshift-installer-release-4.23.yaml`/`openshift-installer-release-5.0.yaml` as well?

hasbro17: @yunjiang29 Updated the job names. I figured the jobs would auto sync from main to 4.22/4.23/5.0 but I've added them nevertheless.

ci-operator/config/openshift/installer/openshift-installer-main.yaml
```diff
@@ -271,13 +271,66 @@ tests:
...
+    env:
+      EXPECTED_ALGORITHM: RSA
+      EXPECTED_KEY_PARAM: "4096"
+      FEATURE_SET: TechPreviewNoUpgrade
+      PKI_ALGORITHM: RSA
+      PKI_RSA_KEY_SIZE: "4096"
+    test:
+    - ref: openshift-installer-pki-verify
+    - ref: openshift-e2e-test
+    workflow: openshift-e2e-aws
```
patrickdillon: I don't think we need periodics, particularly at this stage in the development cycle?

hasbro17: Yeah it's a bit early. This can wait until we need pass rates for graduating to GA later and have a more comprehensive test. Will remove them.

ci-operator/step-registry/openshift/installer/pki/OWNERS
patrickdillon: The OWNERS file should be whoever is responsible for maintaining these steps. We can certainly remove @mtulio 

I would suggest all of the existing OWNERS are removed, and replaced with you and perhaps any QE you may be working with this on the feature.

@yunjiang29 wdyt?

hasbro17: I'll add myself and the rest of the folks from the PKI feature plus QE team.

yunjiang29: > I would suggest all of the existing OWNERS are removed, and replaced with you and perhaps any QE you may be working with this on the feature.

+1

hasbro17: Updated

PR #76905, opened 2026-03-26
hypershift: add AI-driven e2e failure analysis post-step

enxebre:
## Summary
- Add a best-effort post-step (`hypershift-analyze-e2e-failure`) that uses Claude to automatically analyze test failures in e2e jobs
- Clones the ai-helpers repo and loads the `/ci:prow-job-analyze-test-failure` skill to perform root cause analysis against local job artifacts
- Posts a PR comment with a link to the HTML analysis report (presubmit jobs only)
- Added to `hypershift-aws-e2e-external` and `hypershift-azure-aks-e2e` workflows (`/test e2e-aws` and `/test e2e-aks`)
- Exits early with no cost when tests pass

> **Note:** Includes a temporary commit to skip failure detection and hardcode PR openshift/hypershift#8076 for rehearsal validation. Will be reverted after testing.

## Test plan
- [ ] Rehearsal job runs the analysis post-step successfully
- [ ] PR comment is posted to openshift/hypershift#8076
- [ ] HTML report is accessible in job artifacts
- [ ] Revert temporary commit after validation

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #77047, opened 2026-03-30
hypershift: re-add e2e failure analysis with correct post chain handling

enxebre:
## Summary
- Re-adds the `hypershift-analyze-e2e-failure` post-step (reverted in #77014 due to missing cleanup chains)
- Fixes the post chain override: config now specifies the **full post chain** (analysis ref + original cleanup) so workflow cleanup is preserved
- Fixes failure detection: checks `finished.json` from sibling test steps via GCS instead of local `ARTIFACT_DIR` (which only contains the current step's artifacts)
- Only applied to `e2e-aws`, `e2e-aks`, and `e2e-azure-self-managed` jobs

### What went wrong in #76905
Config-level `post:` **overrides** the workflow's `post:` entirely ([docs](https://docs.ci.openshift.org/architecture/step-registry/#ci-operator-test-configuration)). The original PR only listed the analysis ref, dropping the cleanup chains (e.g., `hypershift-destroy-nested-management-cluster`).

### Fix
Each job's config now lists the full post chain:
```yaml
post:
- ref: hypershift-analyze-e2e-failure    # analysis (best_effort, exits early on pass)
- chain: hypershift-destroy-nested-management-cluster  # original cleanup
```

## Test plan
- [ ] Rehearsal jobs run with cleanup chains intact
- [ ] Analysis step detects failures via GCS `finished.json`
- [ ] Analysis step skips when tests pass

### DIFF
--- a/ci-operator/step-registry/hypershift/analyze-e2e-failure/hypershift-analyze-e2e-failure-commands.sh
+++ b/ci-operator/step-registry/hypershift/analyze-e2e-failure/hypershift-analyze-e2e-failure-commands.sh
@@ -1,0 +1,375 @@
+#!/bin/bash
+set -euo pipefail
+
+echo "=== HyperShift E2E Failure Analyzer ==="
+
+# ---------------------------------------------------------------------------
+# 1. Construct GCS base path and detect test failures
+# ---------------------------------------------------------------------------
+JOB_NAME="${JOB_NAME:-unknown}"
+BUILD_ID="${BUILD_ID:-unknown}"
+JOB_TYPE="${JOB_TYPE:-}"
+PULL_NUMBER="${PULL_NUMBER:-}"
+REPO_OWNER="${REPO_OWNER:-}"
+REPO_NAME="${REPO_NAME:-}"
+
+# Build GCS bucket path from CI env vars
+if [[ "$JOB_TYPE" == "presubmit" ]] && [[ -n "$PULL_NUMBER" ]]; then
+  GCS_BUCKET_PATH="pr-logs/pull/${REPO_OWNER}_${REPO_NAME}/${PULL_NUMBER}/${JOB_NAME}/${BUILD_ID}"
+else
+  GCS_BUCKET_PATH="logs/${JOB_NAME}/${BUILD_ID}"
+fi
+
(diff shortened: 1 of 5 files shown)


ci-operator/step-registry/hypershift/analyze-e2e-failure/hypershift-analyze-e2e-failure-commands.sh
```diff
@@ -0,0 +1,375 @@
...
+echo ""
+echo "Running Claude with /ci:prow-job-analyze-test-failure skill..."
+echo "Primary test: $FIRST_FAILED_TEST"
+echo ""
+
+set +e
+claude -p "$USER_PROMPT" \
+  --system-prompt "$SYSTEM_PROMPT" \
+  --allowedTools "Bash Read Write Grep Glob WebFetch" \
+  --max-turns 30 \
```
bryan-cox: Did you find 30 turns enough for triaging e2e issues? Especially if multiple tests were broken.

ci-operator/step-registry/hypershift/analyze-e2e-failure/hypershift-analyze-e2e-failure-commands.sh
```diff
@@ -0,0 +1,375 @@
...
+${SKILL_CONTENT}
+
+DEPENDENT SKILL - Fetch ProwJob JSON:
+${FETCH_PROWJOB_CONTENT}"
+
+# Build the user prompt with the job URL and test name
+USER_PROMPT="${PROW_JOB_URL} ${FIRST_FAILED_TEST} --fast"
+
+# If multiple tests failed, add them as additional context
+ADDITIONAL_TESTS=$(echo "$FAILED_TESTS" | tail -n +2 | head -9)
```
bryan-cox, line 162: Can we add some prompt here or somewhere else to use subagents to triage failed tests in parallel?

enxebre: I wanted to start simple until this works. I would rather leave it or even don't pass test as argument at all and let the skill work generically

PR #77064, opened 2026-03-30
MULTIARCH-5144: HCP with type Kubevirt platform conformance test cases analysis

jpattara:
**Summary**

As part of the conformance testing for HCP with the KubeVirt platform on IBM Z, we have observed a set of test case failures during the CI runs. After initial analysis, these failures appear to be either environment-specific, known limitations with the KubeVirt-based setup, or not directly impacting the core functionality being validated for this release.

Based on this assessment, these specific failing test cases are currently being skipped to allow the conformance runs to proceed without blocking overall validation.


orenc1:
/lgtm
so the job will pass.
but I recommend to track down the exact reason these tests are failing specifically on s390x / IBM Z arch

LiangquanLi930:
@jpattara I didn't see any failed jobs. Could you please share the failed job link?

PR #77079, opened 2026-03-30
Allow group propagates to dex

bear-redhat:


clusters/app.ci/dex/manifests.yaml
```diff
@@ -56,8 +56,9 @@ spec:
           connectors:
           - config:
               claimMapping:
-                groups: groups
                 preferred_username: preferred_username
+                groups: groups
```
deepsm007, line 59: can you only add "insecureEnableGroups: true" instead of rearranging? breaking-changes will be complaining?

PR #77091, opened 2026-03-30
fix default catalog tests for 5.0

grokspawn:
5.0 versions of the default catalog consistency tests periodic are not using the right builder base


rashmigottipati (approved):
/lgtm

PR #77093, opened 2026-03-30
OTA-1896: Add a configuration for the openshift/cluster-update-console-plugin repository

DavidHurta:
Add a Prow configuration for the new [openshift/cluster-update-console-plugin](https://github.com/openshift/cluster-update-console-plugin) repository.

Onboarding a New Component for Testing and Merge Automation Documentation: https://docs.ci.openshift.org/how-tos/onboarding-a-new-component/

### DIFF
--- a/ci-operator/config/openshift/cluster-update-console-plugin/OWNERS
+++ b/ci-operator/config/openshift/cluster-update-console-plugin/OWNERS
@@ -1,0 +1,12 @@
+approvers:
+- davidhurta
+- fao89
+- hongkailiu
+- pratikmahajan
+- wking
+options: {}
+reviewers:
+- davidhurta
+- fao89
+- hongkailiu
+- wking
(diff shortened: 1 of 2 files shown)


ci-operator/config/openshift/cluster-update-console-plugin/OWNERS
```diff
@@ -0,0 +1,12 @@
+approvers:
+- davidhurta
+- fao89
+- hongkailiu
+- pratikmahajan
+- wking
```
wking, line 6: nit: Instead of hard-coding again locally here, maybe we can update [the `cincinnati-approvers` alias][1] and then use it in all of these locations?

[1]: https://github.com/openshift/release/blob/abc602b94843a385747fb3e99d4e2cd9a954f669/OWNERS_ALIASES#L22-L28

wking, line 6: Nevermind, I see from your b952f94a4ce8a commit message that we don't actually care what goes in as seed data here.

PR #77095, opened 2026-03-30
Add skip check for docs/CI only PRs | Disable e2e tests

mynhardtburger:
1. Disable `e2e-test-maas` job for `rhaoi` and `stable` branches.
When `always_run: false` and no `skip_if_only_changed` / `run_if_changed` is set, the job is only triggerable manually.
https://docs.prow.k8s.io/docs/jobs/#triggering-jobs

2. Add skip check to the `ci/prow/images` pipeline to avoid running for docs/CI only PRs.
This is the same skip check used for the other pipelines in this file.
This is done directly in the generated `jobs/*` yaml as allowed per the documentation -> https://docs.ci.openshift.org/how-tos/contributing-openshift-release/#tolerated-changes-to-generated-jobs


mynhardtburger:
Closing. Superceded by https://github.com/openshift/release/pull/77280

PR #77098, opened 2026-03-30
Fix issue DPTP-4738 Forward logs to thanos

bear-redhat:


clusters/_cluster-install/build02.yaml
```diff
@@ -6,6 +6,14 @@ onboard:
   hosted: false
   unmanaged: false
   useTokenFileInKubeconfig: true
+  multiarchBuilderController:
+    skip: true
+  quayioPullThroughCache:
```
deepsm007: We dont need this

clusters/_cluster-install/build08.yaml
```diff
@@ -0,0 +1,173 @@
...
+provision:
+  gcp: {}
+onboard:
+  osd: false
+  hosted: false
+  unmanaged: false
+  useTokenFileInKubeconfig: true
+  multiarchBuilderController:
+    skip: true
+  quayioPullThroughCache:
```
deepsm007: Same as above

clusters/_cluster-install/build01.yaml
```diff
@@ -121,3 +133,5 @@ onboard:
               - key: node-role.kubernetes.io/infra
                 operator: Exists
                 effect: NoExecute
+  nestedPodman:
```
liangxia: We don't need this now.  We have nested-podman on all build farm clusters via https://github.com/openshift/release/pull/77036

clusters/_cluster-install/build02.yaml
```diff
@@ -36,13 +38,12 @@ onboard:
...
   ciSchedulingWebhook:
     skip: true
-  multiarchBuilderController:
-    skip: true
   machineSet:
     skip: true
   passthrough:
     exclude:
     - "**/supplemental-ci-images/**"
+    - "**/upgrader-sa.yaml"
```
liangxia: We don't need this, too.
The periodic upgrade jobs are using `config-updater` and no longer use the `upgrader-sa`

PR #77100, opened 2026-03-30
use build_root from_repository on periodics

jluhrsen:
this has been resolved for our presubmits already but I missed the periodics. fix it

### DIFF
--- a/ci-operator/jobs/openshift/ovn-kubernetes/openshift-ovn-kubernetes-release-4.21-periodics.yaml
+++ b/ci-operator/jobs/openshift/ovn-kubernetes/openshift-ovn-kubernetes-release-4.21-periodics.yaml
@@ -237,8 +237,6 @@ periodics:
   cluster: build11
   cron: 0 * * * *
   decorate: true
-  decoration_config:
-    skip_cloning: true
   extra_refs:
   - base_ref: release-4.21
     org: openshift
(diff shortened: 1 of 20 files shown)


ci-operator/jobs/openshift/ovn-kubernetes/openshift-ovn-kubernetes-release-4.21-periodics.yaml
```diff
@@ -644,8 +634,6 @@ periodics:
   cluster: build07
   cron: 0 4 * * *
   decorate: true
-  decoration_config:
```
tssurya, line 647: what does this decoration_config do really?

jluhrsen: this just meant to not clone the repo immediately and to use the specified image in the config for the root builder. now, we need to clone the repo first so we can see what that .ci-operator.yaml has for the right root builder image.

can't remember the details (it's out there in the ether somewhere in slack or email), but some change in the past
couple of months or so started having some issues with the builder images and now especially since we are on
a new go version that breaks with the older go version.

In the end, our jobs using the specific images was some kind of stop gap we did years ago (before me I think) and I think the intention was to get back to `from_repository: true` but we just never did. so all those ART bumps in .ci-operator.yaml were pointless and we kept having to manually bump them in release. this should help going forward.

PR #77138, opened 2026-03-31
Feat: add janitor for dev environment

typeid:
This PR adds a janitor job for the ephemeral-dev environment that runs every Saturday. 

I have also adjusted the janitor for the ci environment to run every Saturday as well.


typeid:
/hold

Waiting until after demo week

PR #77140, opened 2026-03-31
OpenStack: fix openstack-installer image building

mandre:
After the openstack repos were removed from the base image via https://github.com/openshift/release/pull/76590...

### DIFF
--- a/ci-operator/config/openshift/installer/openshift-installer-main.yaml
+++ b/ci-operator/config/openshift/installer/openshift-installer-main.yaml
@@ -39,6 +39,10 @@ base_images:
     name: "4.22"
     namespace: ocp
     tag: cli
+  ocp_4.22_base-rhel9:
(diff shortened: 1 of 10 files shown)


ci-operator/config/openshift/installer/openshift-installer-main.yaml
```diff
@@ -66,6 +70,11 @@ images:
 - dockerfile_path: images/installer/Dockerfile.upi.ci
   to: upi-installer
 - dockerfile_path: images/openstack/Dockerfile.ci
+  from: openstack-base
+  inputs:
+    openstack-base:
+      as:
+      - registry.ci.openshift.org/ocp/4.17:base-rhel9
```
mandre, line 77: The 4.17 here should be updated, but this is what we currently have in installer main.

PR #77149, opened 2026-03-31
Fix machine-network CIDR issue in AWS DualStack cluster

yunjiang29:
AWS DualStack BYO-VPC cluster required IPv6 machine CIDR in the config file.


tthvo (commented):
/lgtm
/approve

PR #77164, opened 2026-03-31
CNTRLPLANE-3278: Add Prow presubmit to enforce GitHub Actions workflow consistency for HyperShift

bryan-cox:
## Summary

Adds a `verify-workflows` first-stage Prow presubmit to the HyperShift ci-operator config that ensures PR branches have up-to-date GitHub Actions workflow files, and enables `/retest` to re-trigger failed GitHub Actions workflows.

## Problem

GitHub Actions workflows only run if the workflow YAML file exists on the PR's head branch. PRs branched before a workflow was introduced or updated will never trigger those jobs, allowing them to merge without running the latest CI checks. Additionally, when GHA checks flake, Prow's `/retest` command does not re-trigger them by default.

### 1. `verify-workflows` presubmit

A lightweight `container` test (`from: src`) that runs on every PR:

1. Fetches upstream `main` and computes the merge-base
2. Iterates each workflow file on `main`, comparing per-file blob hashes
3. For each file: if missing on the PR branch or outdated (matches merge-base but not main), the check fails
4. PRs that intentionally modify workflow files are allowed (hash differs from both main and merge-base)

Per-file comparison prevents bypassing the check by touching an unrelated workflow file.

### 2. `trigger_github_workflows` enablement

Enables the Prow trigger plugin's `trigger_github_workflows` flag for `openshift/hypershift`. This allows `/retest` and `/test all` comments to re-trigger failed GitHub Actions workflows alongside Prow jobs.

## Test plan

- [x] Tested locally: up-to-date branch passes
- [x] Tested locally: branch predating workflow files fails with clear per-file output
- [ ] Verify presubmit runs on a real HyperShift PR after merge
- [ ] Verify `/retest` re-triggers failed GHA workflows after merge

### DIFF
--- a/ci-operator/config/openshift/hypershift/openshift-hypershift-main.yaml
+++ b/ci-operator/config/openshift/hypershift/openshift-hypershift-main.yaml
@@ -468,6 +468,40 @@ tests:
       CHECK_MOD_LIST: "false"
     test:
     - ref: go-verify-deps
+- as: verify-workflows
+  commands: |
+    set -euo pipefail
+    git fetch --unshallow 2>/dev/null || true
+    if ! git fetch https://github.com/openshift/hypershift.git main; then
+      echo "ERROR: Could not fetch main from GitHub."
+      exit 1
+    fi
+    MAIN_REF="FETCH_HEAD"
+    BASE=$(git merge-base HEAD "$MAIN_REF")
+    FAILED=0
+    while IFS= read -r file; do
+      [ -z "$file" ] && continue
+      main_hash=$(git rev-parse "${MAIN_REF}:${file}" 2>/dev/null) || continue
+      head_hash=$(git rev-parse "HEAD:${file}" 2>/dev/null) || head_hash=""
+      if [ -z "$head_hash" ]; then
+        echo "MISSING: $file exists on main but not on this branch."
+        FAILED=1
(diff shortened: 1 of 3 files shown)


ci-operator/config/openshift/hypershift/openshift-hypershift-main.yaml
```diff
@@ -468,6 +468,40 @@ tests:
       CHECK_MOD_LIST: "false"
     test:
     - ref: go-verify-deps
+- as: verify-workflows
+  commands: |
+    set -euo pipefail
+    git fetch --unshallow 2>/dev/null || true
```
sjenning, line 474: what is this line doing?

bryan-cox: From Claude:
```
⏺ git fetch --unshallow 2>/dev/null || true — it converts a shallow clone (depth=1) into a full clone so that git merge-base can find the common ancestor
  between HEAD and main. CI's src image is typically a shallow clone, and without full history merge-base would fail.

  The || true is there because if the repo is already a full clone (e.g., running locally), git fetch --unshallow errors out — and we don't want that to fail
  the script.
```

celebdor:
/lgtm

We should consider having a follow-up that checks for the changes in the files that affect each of the github actions

PR #77181, opened 2026-03-31
Add Testing Repository 'autoscale-tests' So that 

prozehna:
CC: @maxcao13 @jkyros for visibility.

Pasting some of the important-seeming 'make new-repo' interactive command choices for future-self, just in case.

```
What version of Go does the repository build with? [default: 1.13] 1.25
*skipped* [OPTIONAL] Enter the Go import path for the repository if it uses a vanity URL (e.g. "k8s.io/my-repo"): 
*skipped* [OPTIONAL] What commands are used to build binaries in the repository? (e.g. "go install ./cmd/...") 
*skipped* [OPTIONAL] What commands are used to build test binaries? (e.g. "go install -race ./cmd/..." or "go test -c ./test/...") 

Are there any test scripts to configure?  [default: no] yes
What is the name of this test (e.g. "unit")?  vpa-unit-1
What commands in the repository run the test (e.g. "make test-unit")?  make test-vpa-unit-1
Are there any more test scripts to configure?  [default: no] no

Are there any end-to-end test scripts to configure?  [default: no] yes
What is the name of this test (e.g. "e2e-operator")?  vpa-e2e-1
Which specific cloud provider does the test require, if any?  [default: aws] 
What commands in the repository run the test (e.g. "make test-e2e")?  make test-vpa-e2e-1
Does your test require the OpenShift client (oc)?  [default: no] yes
```

### DIFF
--- a/ci-operator/config/openshift-kni/eco-ci-cd/openshift-kni-eco-ci-cd-main__cnf-compute-nrop-compact-stage.yaml
+++ b/ci-operator/config/openshift-kni/eco-ci-cd/openshift-kni-eco-ci-cd-main__cnf-compute-nrop-compact-stage.yaml
@@ -19,7 +19,8 @@ tests:
   steps:
     env:
       CLUSTER_NAME: kni-qe-52
-      OCP_VERSION_RELEASE_TYPE: nightly
+      OCP_VERSION_RELEASE_AGE_MAX_DAYS: "7"
+      OCP_VERSION_RELEASE_TYPE: release-candidate
       OPERATORS: |
         [
           { "name": "numaresources-operator", "catalog": "numaresources-operator-fbc", "og_name": "numaresources-operator", "nsname": "openshift-numaresources", "deploy_default_config": false, "fbc_iib_repo": "latest", "channel": "5.0", "ocp_operator_mirror_fbc_image_base": "quay.io/redhat-user-workloads/telco-5g-tenant/numaresources-operator-fbc-5-0" }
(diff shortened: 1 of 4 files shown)


ci-operator/config/openshift/autoscale-tests/openshift-autoscale-tests-main.yaml
```diff
@@ -0,0 +1,39 @@
...
+tests:
+- as: vpa-unit-1
+  commands: make test-vpa-unit-1
+  container:
+    from: src
+- as: vpa-e2e-1
+  steps:
+    cluster_profile: aws
+    test:
+    - as: vpa-e2e-1
```
maxcao13: Is the plan that we will followup after this merges in openshift/autoscale-tests repo so that these tests do stuff? Just want to make sure we are doing what we want to 😛

prozehna: Exactly.
IMO consensus right now is to try to migrate tests, or at least some of them, and if we can't migrate them, then write new tests, first goal being one smoke test per operator.
The name vpa-unit-1 is just a placeholder.

ci-operator/config/openshift/autoscale-tests/openshift-autoscale-tests-main.yaml
```diff
@@ -0,0 +1,39 @@
+build_root:
+  image_stream_tag:
+    name: release
+    namespace: openshift
+    tag: golang-1.25
```
jkyros: was this what the `new repo` command gave you? Very strange ( if so ) that it wouldn't be valid.

jkyros: huh, yeah that's totally what mine generates too

prozehna: I'll try again, maybe it was a flake.
@jkyros  If it wasn't, I'll decrement to 1.24 and retest, unless you think that's grody.

PR #77186, opened 2026-03-31
Fix issue DPTP-4738 Params tweaks

bear-redhat:


deepsm007 (approved):

clusters/gitops/app-000-starter.yaml
liangxia, line 25: For temp debug or persistent change ?

PR #77189, opened 2026-03-31
NO-ISSUE: chore: use node module stream

jbpratt:
seems we are installing node 16! that's wrong


shaonrh (approved):
LGTM

PR #77197, opened 2026-03-31
AGENT-1486: Create generic ISO no registry workflow

pawanpinjarkar:


ci-operator/step-registry/agent/e2e/generic/conformance/iso-no-registry/agent-e2e-generic-conformance-iso-no-registry-workflow.yaml
rwsu, line 7: I think we should add these common iso-no-registry dev-scripts env vars to the workflow definition:
```suggestion
      DEVSCRIPTS_TARGET: agent
      DEVSCRIPTS_CONFIG: |
        AGENT_E2E_TEST_BOOT_MODE=ISO_NO_REGISTRY
        AGENT_CLEANUP_ISO_BUILDER_CACHE_LOCAL_DEV=true
```
Would it be helpful to leave AGENT_CLEANUP_ISO_BUILDER_CACHE_LOCAL_DEV unset and default to false to aid in debugging?

andfasano: Since they are going to be replaced in every usage, better to leave it empty

ci-operator/step-registry/agent/test/conformance/iso-no-registry/agent-test-conformance-iso-no-registry-chain.yaml
andfasano, line 4: Shouldn't point to @rwsu  simpler test step?

PR #77082, opened 2026-03-30
AGENT-1429: Add a periodic job for agent installer iso-no-registry

bfournie:
Add a periodic job to run tests on and agent-based installer cluster created using iso-no-registry which uses the openshift/appliance to create a disconnected cluster with installed operators. The deployed cluster must be in TechPreview in 4.22.

As a follow on we will add tests to exercise the NoRegistryClusterInstall.

## Summary by CodeRabbit

* **Tests**
  * Added a new E2E conformance test and workflow for the agent-installer ISO "no-registry" (TechPreview, COMPACT_IPV4, ISO_NO_REGISTRY); scheduled every 12 hours on equinix-ocp-metal with intranet capability (applies to main and nightly pipelines).
  * Added a test runner/filter script to run conformance tests matching NoRegistryClusterInstall and capture logs/JUnit outputs; step resource/time limits configured.

* **Chores**
  * Added assisted-component and dev-scripts base image references to CI configs.
  * Added approver/ownership metadata for the new workflows and test steps.

PR #77201, opened 2026-03-31
AGENT-1487: add baremetalds-e2e-test-iso-no-registry ref

rwsu:
Add a new ref for the ISO NO REGISTRY scenario that runs all origin tests without image mirroring. Unlike the existing baremetalds-e2e-test ref, the commands.sh simply exports the kubeconfig and runs openshift-tests run all, optionally filtering via TEST_SKIPS.

### DIFF
--- a/ci-operator/step-registry/baremetalds/e2e/test/iso-no-registry/baremetalds-e2e-test-iso-no-registry-commands.sh
+++ b/ci-operator/step-registry/baremetalds/e2e/test/iso-no-registry/baremetalds-e2e-test-iso-no-registry-commands.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+set -o nounset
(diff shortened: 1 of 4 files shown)


ci-operator/step-registry/baremetalds/e2e/test/iso-no-registry/baremetalds-e2e-test-iso-no-registry-commands.sh
pawanpinjarkar, line 22: Only running `/openshift-tests run all --run ".*NoRegistryClusterInstall.*"` should be good for now

pawanpinjarkar (changes requested):

rwsu:
Closing, replaced by https://github.com/openshift/release/pull/77082

PR #77248, opened 2026-04-01
OCPBUGS-80930 | Fix UltraSSD StorageClass to not override cluster default

radeore:
## Summary

- Remove `is-default-class` annotation from the `managed-csi-ultrassd` StorageClass to avoid breaking other cluster components that rely on the default StorageClass

- The UltraSSD StorageClass is now created as an additional non-default StorageClass, referenced explicitly by the CSI driver test manifest (`ultrassd-manifest.yaml`)

Depends on: https://github.com/openshift/csi-operator/pull/538


ci-operator/step-registry/storage/conf/storageclass/pre-set-default-azure-disk-ultrassd/storage-conf-storageclass-pre-set-default-azure-disk-ultrassd-commands.sh
```diff
@@ -8,24 +8,10 @@ apiVersion: storage.k8s.io/v1
...
-    storageclass.kubernetes.io/is-default-class: "true"
 provisioner: disk.csi.azure.com
 parameters:
   skuName: UltraSSD_LRS
 volumeBindingMode: WaitForFirstConsumer
 allowVolumeExpansion: true
 reclaimPolicy: Delete
 EOF
-
-cat << EOF > ${SHARED_DIR}/manifest_cluster_csi_driver.yaml
```
Phaow, line 21: If we only create the `managed-csi-ultrassd` storageclass consume by the vac e2e, we'd better create it for post action(day 1 is not required) which mean the step better to add as `ref: storage-conf-storageclass-add-azure-disk-ultrassd-storageclass`, current step name also mislead `set default`

radeore: Good catch, thanks!

Phaow:
I could see the install succeed, let's go ahead merge this and rehearse in https://github.com/openshift/csi-operator/pull/538 .

PR #77270, opened 2026-04-01
Refactor cost-onprem-chart job config to avoid long running tests on PRs

testetson22:
- Add optional on-demand IQE test jobs (e2e-iqe-extended, e2e-iqe-stable) for deeper test coverage
- Jobs configured with always_run: false to prevent automatic triggering on PRs
- Trigger manually with /test e2e-iqe-extended or /test e2e-iqe-stable
- Main e2e job unchanged - continues running IQE smoke tests on code changes


jordigilh:
### Review feedback

**1. Commit message is misleading**

The commit message says:

> "Removed `skip_if_only_changed` condition from presubmit jobs to ensure all changes trigger the tests."

This describes the opposite of what the PR does. With `always_run: false` and no `skip_if_only_changed`, the tests are *never* automatically triggered — they only run via explicit `/test` commands. Suggest amending to something like:

> "Removed `skip_if_only_changed` condition from presubmit jobs; combined with `always_run: false`, these tests now only run when manually triggered via `/test`."

**2. When should these be manually triggered?**

After this change, extended/stable/full only run if someone remembers to `/test` them. Is there guidance (e.g., a PR template checklist, or a wiki page) on when contributors should trigger these? For example: before release cuts, after infra changes, etc. Without that, these tests could go un-triggered for long stretches.

**3. Periodic/postsubmit coverage?**

With presubmits now manual-only, is there a periodic or postsubmit job that runs these profiles against `main`? If not, a regression could land without anyone noticing until the next manual run.

PR #77273, opened 2026-04-01
MCO-2199: add AWS-SNO job variants for feature promotion

yuqi-zhang:
Our original periodics did not account for SNO, which is now added as a feature promotion requirement. Adding these for 4.22+.

Assisted-by: Claude Opus 4.6


yuqi-zhang:
/close

in favour of https://github.com/openshift/release/pull/79344

PR #77284, opened 2026-04-02
feat: add ability to deploy plugin integration pieces alongside rhdh

PatAKnight:
## Description 

Adds support for the `--plugins` flag being introduced in the companion [PR #30](https://github.com/redhat-developer/rhdh-test-instance/pull/30), which deploys optional plugin integrations alongside an RHDH test instance. This allows users to deploy an RHDH instance with whichever plugins they want to test against.

Also included is a `redeploy` action, which allows users to switch to a new set of plugins without manually tearing down the cluster first. Under the hood, it looks up the previously deployed plugins, calls `teardown.sh` to cleanly remove those plugin integrations and the existing RHDH instance, and then redeploys fresh with the new plugin configuration.

Made with [Cursor](https://cursor.com/)

/cc @subhashkhileri 

## Summary

This PR adds CI infrastructure support for deploying optional plugin integrations alongside RHDH (Red Hat Developer Hub) test instances in the OpenShift CI pipeline.

### Changes

The PR modifies the RHDH test instance CI step registry to enhance the deployment command parser and execution flow:

**Enhanced Deployment Command Recognition**
- The test deployment trigger command now accepts both `/test deploy` and `/test redeploy` formats (previously only `/test deploy`)
- Supports configurable deployment duration (defaults to 3 hours if not specified)
- Recognizes a new optional `--plugins` argument that accepts a comma-separated list of plugin names to deploy alongside the RHDH instance

**Redeploy Capability**
- Enables users to redeploy RHDH instances with different plugin configurations without manual cluster teardown
- The redeploy action internally manages plugin cleanup and redeployment by invoking teardown and redeploy workflows

**Updated Messaging**
- Error messages and inline documentation updated to reflect the new `redeploy` and `--plugins` command syntax
- Success comment generation now includes a "Plugins" section when plugins are specified in the deployment

### Impact

This enhances the RHDH test instance CI configuration to provide more flexible testing scenarios, allowing developers to easily test RHDH with different plugin combinations through simple comment-based CI triggers on pull requests.

### DIFF
--- a/ci-operator/step-registry/redhat-developer/rhdh-test-instance/redhat-developer-rhdh-test-instance-commands.sh
+++ b/ci-operator/step-registry/redhat-developer/rhdh-test-instance/redhat-developer-rhdh-test-instance-commands.sh
@@ -118,25 +118,64 @@ echo "Found comment: $comment_body"
 
 if [[ -n "$comment_body" && "$comment_body" != "null" ]]; then
     read -r -a comment_parts <<< "$comment_body"
-    
-    if [ ${#comment_parts[@]} -ge 4 ] && [[ "${comment_parts[2]}" == "helm" || "${comment_parts[2]}" == "operator" ]]; then
-        # Extract install.sh arguments (skip /pj-rehearse or /test and job_name)
+
+    action="${comment_parts[1]}"
+
+    if [ ${#comment_parts[@]} -ge 4 ] && [[ "${comment_parts[2]}" == "helm" || "${comment_parts[2]}" == "operator" ]] && [[ "$action" == "deploy" || "$action" == "redeploy" ]]; then
+        # Extract deploy arguments (skip /pj-rehearse or /test and action)
         install_type="${comment_parts[2]}"
         rhdh_version="${comment_parts[3]}"
-        
-        # Check if duration is provided (5th argument), otherwise default to 3h
-        if [ ${#comment_parts[@]} -ge 5 ]; then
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/redhat-developer/rhdh-test-instance/redhat-developer-rhdh-test-instance-commands.sh
```diff
@@ -118,25 +118,64 @@ echo "Found comment: $comment_body"
...
+            fi
+        done
+
         echo "Parsed arguments: $install_type $rhdh_version"
         echo "Time duration: $time"
-        
-        source ./deploy.sh "$install_type" "$rhdh_version"
+        [[ ${#plugins_args[@]} -gt 0 ]] && echo "Plugins: ${plugins_args[*]}"
+
+        if [[ "$action" == "redeploy" ]]; then
```
subhashkhileri: Hey, since these CI jobs spin up a fresh ephemeral cluster each time, the teardown-before-redeploy path (looking up the previous deploy comment, parsing old plugins, etc.) never has anything to actually clean up. Can we keep it simple and just have `redeploy` call `deploy.sh` directly, same as `deploy`?

PatAKnight: Done

PR #77298, opened 2026-04-02
[o/cert-manager-operator] CM-871: Updates e2e-operator-tech-preview label filter to exclude certain cases

bharath-b-rh:
The PR is for updating the gingko label filter in `e2e-operator-tech-preview` job config, which is required to exclude the negative cases of a TechPreview feature, which are for testing on TechPreview disabled cluster and shouldn't be executed on TechPreview clusters.


chiragkyal (approved):
/lgtm

PR #77304, opened 2026-04-02
OLS-2830 - fix ols disconnected tests

JoaoFula:


ci-operator/step-registry/openshift/lightspeed-service/install/disconnected/openshift-lightspeed-service-install-disconnected-commands.sh
xrajesh, line 146: Do we want to run the next block if we set SSL_CERT_FILE already ? or thats a fallback ?

JoaoFula: it;s a fallback. I tried just getting the TLS from the registry-config and ignore the user-ca-bundle but we were still hitting the TLS errors so I'm fairly sure we need it.

xrajesh: I think the fallback will execute always - do we want to skip if the first one passes ?

JoaoFula: nope, again, only got the tests to pass consistently after mounting the user-ca-bundle certificates. Perhaps we could go without the first but I don't see a strong reason not to have it either.

PR #77305, opened 2026-04-02
cluster-api-actuator-pkg-ote-extension

huali9:
Add non-payload cluster-api-actuator-pkg-ote-extension binary


huali9:
/hold

this won't work until https://github.com/openshift/cluster-api-actuator-pkg/pull/423 has merged

PR #77345, opened 2026-04-02
rhcos-10: restrict aws to us-east-1

patrickdillon:
Currently the installer only has a an ami for us-east-1, so we are depending on the installer to copy a new ami for each install.

I think it would be better to localize the testing to one region until we expand to publish images in more regions.

These tests are causing machine-api operator to be degraded in component readiness with https://redhat.atlassian.net/browse/OCPBUGS-81552


patrickdillon:
/pj-rehearse periodic-ci-openshift-release-main-nightly-4.22-e2e-aws-ovn-single-node-rhcos10-techpreview-serial

PR #77352, opened 2026-04-02
AUTOSCALE-594: Add keda + cma e2e tests for GCP

maxcao13:
https://redhat.atlassian.net/browse/AUTOSCALE-594 where we need to enable gcp tests for keda. This PR does not enable the GCP WIF tests. That will come later when we make a new test config file specifically for gcp tests in our keda fork.

This PR also refactors the test steps for keda and cma into re-usable chains and steps in the step-registry, so we don't have to repeat config through the aws and gcp workflows (and across the operator and operand repos).


maxcao13:
/pj-rehearse pull-ci-openshift-kedacore-keda-main-keda-e2e-gcp-ovn

maxcao13:
/pj-rehearse

I added a override to the keda cma tests for the config file. Let's see if it works -- it should fail since the file doesn't exist yet.

PR #77355, opened 2026-04-03
stolostron: Update search-e2e target release to 2.17

jlpadilla:
/assign @smcavey


ci-operator/jobs/opendatahub-io/models-as-a-service/opendatahub-io-models-as-a-service-main-postsubmits.yaml
```diff
@@ -1,7 +1,7 @@
 postsubmits:
   opendatahub-io/models-as-a-service:
   - agent: kubernetes
-    always_run: false
```
smcavey: I don't think you meant to modify this file

jlpadilla: Yes, the generated-config job was failing. Then, I ran `make update` and it had the same change.

smcavey: Is this the real Jorge or an agent acting upon Jorge's behalf?

jlpadilla: lol, that was the real me. 👨

PR #77376, opened 2026-04-03
Mirror more apollo-ci images

porridge:
I missed these in https://github.com/openshift/release/pull/77311 thinking just stackrox-test would be enough for my needs.
Turns out we seem to be using the ui-test image even for non-ui tests these days 🤷🏻 
/cc @tommartensen @davdhacs


porridge:
@danilo-gemoli do you need to /approve as well? 🤔

PR #77408, opened 2026-04-06
in nvidia-ci use promoted base image instead of rebuilding in variant configs

TomerNewman:
/cc @empovit


TomerNewman:
closing because it seems the problem is gone

PR #77416, opened 2026-04-06
hypershift: fix e2e failure analyzer to poll GCS for sidecar artifacts

enxebre:
## Summary
- Fix the analyze-e2e-failure post-step which was silently skipping analysis on all test failures
- Root cause: `grep -oP` (PCRE) not portable, and hardcoded step name fallback list was incomplete (missing `hypershift-aws-run-e2e-nested`)
- The test step's sidecar uploads artifacts to GCS asynchronously — artifacts may not be in GCS yet when this post-step starts
- Now polls known test step `finished.json` files in GCS every 15s (up to 5 min) until they appear, then passes the Prow job URL to Claude for full artifact-based analysis
- On passing jobs, exits immediately when `passed: true` is found
- Removed test name extraction logic — Claude discovers failures itself from the Prow job URL

## Test plan
- [ ] Trigger a presubmit e2e-aws job on openshift/hypershift with a known test failure and verify the analyzer detects it and produces a report
- [ ] Trigger a passing e2e-aws job and verify the analyzer exits quickly without running Claude

### DIFF
--- a/ci-operator/step-registry/hypershift/analyze-e2e-failure/hypershift-analyze-e2e-failure-commands.sh
+++ b/ci-operator/step-registry/hypershift/analyze-e2e-failure/hypershift-analyze-e2e-failure-commands.sh
@@ -4,7 +4,7 @@ set -euo pipefail
 echo "=== HyperShift E2E Failure Analyzer ==="
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


bryan-cox (approved):
/lgtm

PR #77419, opened 2026-04-06
NO-JIRA: [cert-manager-operator] Add CI config for cert-manager-operator-1.19

chiragkyal:
Add CI config for cert-manager-operator-1.19

/cc @bharath-b-rh


chiragkyal:
Already covered through another PR. 

/close

PR #77426, opened 2026-04-06
SREP-4345: Add VPC cleanup step to rosa-e2e deprovision

dustman9000:
## Summary

- Add `rosa-e2e-cleanup-vpc` step that cleans up leftover AWS resources (VPC endpoints, load balancers, ENIs, security groups) before CloudFormation stack deletion
- Break out the deprovision chain in the rosa-e2e workflow to insert the cleanup step before `aws-deprovision-stacks`
- Fixes the VPC CloudFormation DELETE_FAILED issue caused by AVO cleanup race condition during cluster deprovisioning

The rosa-e2e nightly job passes all tests but consistently fails in the post (cleanup) phase because the VPC stack can't be deleted while AVO-managed VPC endpoints still hold references to subnets.

Jira: https://redhat.atlassian.net/browse/SREP-4345

### DIFF
--- a/ci-operator/step-registry/rosa/e2e/cleanup-vpc/rosa-e2e-cleanup-vpc-ref.yaml
+++ b/ci-operator/step-registry/rosa/e2e/cleanup-vpc/rosa-e2e-cleanup-vpc-ref.yaml
@@ -1,0 +1,14 @@
+ref:
+  as: rosa-e2e-cleanup-vpc
+  from: rosa-aws-cli
+  grace_period: 5m
+  commands: rosa-e2e-cleanup-vpc-commands.sh
+  resources:
+    requests:
+      cpu: 100m
+      memory: 300Mi
+  documentation: |-
+    Cleans up leftover AWS resources (ENIs, security groups, load balancers)
+    in the VPC before CloudFormation stack deletion. Works around the AVO
+    cleanup race condition where VPC endpoints hold references that prevent
+    stack deletion.
--- a/ci-operator/step-registry/rosa/e2e/cleanup-vpc/OWNERS
+++ b/ci-operator/step-registry/rosa/e2e/cleanup-vpc/OWNERS
@@ -1,0 +1,10 @@
+approvers:
+- tiwillia
+- dustman9000
+- bmeng
+- ravitri
+reviewers:
+- tiwillia
+- dustman9000
+- bmeng
+- ravitri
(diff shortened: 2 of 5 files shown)


ci-operator/step-registry/rosa/e2e/cleanup-vpc/rosa-e2e-cleanup-vpc-ref.yaml
```diff
@@ -0,0 +1,14 @@
...
+  from: rosa-aws-cli
+  grace_period: 5m
+  commands: rosa-e2e-cleanup-vpc-commands.sh
+  resources:
+    requests:
+      cpu: 100m
+      memory: 300Mi
+  documentation: |-
+    Cleans up leftover AWS resources (ENIs, security groups, load balancers)
+    in the VPC before CloudFormation stack deletion. Works around the AVO
```
joshbranham, line 12: Is there a way to force AVO to reconcile "faster" or shorten retries so we can rely on it to clean up properly?

dustman9000: Good question. There are actually two separate issues causing orphaned VPC resources during deprovisioning:

1. **HyperShift CPO** (ROSA HCP): The control-plane-operator creates VPC endpoint services and PrivateLink connections in the customer AWS account. During deprovision, the hosted control plane namespace gets deleted but the VPCE in osd-privatelink isn't cleaned up. VPCE holds VPC Endpoint Service -> NLB -> ENIs -> VPC subnets -> CFN stack can't delete.

2. **AVO race condition** (SCs/MCs): AVO has a 30s terminationGracePeriodSeconds. If AWS doesn't finish deleting VPC endpoint resources in time, Hive/SyncSet removal proceeds and effectively uninstalls AVO before it can finish. Orphaned VPC endpoints remain.

Neither can be fixed from the step registry -- the HyperShift fix needs to happen in CPO's cleanup logic, and the AVO fix needs better coordination between OSDFM, Hive, and AVO during teardown. Both are being discussed in #forum-rosa-eng.

This cleanup step is a pragmatic workaround so the nightly job reports clean results while the upstream fixes are worked on. It's also defensive -- even after the fixes land, leftover resources from edge cases won't block deprovision.

Tracking: SREP-4357 (HyperShift/CPO), SREP-4358 (AVO/Hive)

PR #77434, opened 2026-04-06
platform-externa/aws/opct: add "None" workflow to cover VCSP

mtulio:
Introducing the job validate OCP install with configuration Platform Type `External` without CCM on AWS provider, running VCSP conformance workflow deployed by OPCT


mtulio:
Closing as obsolete. OPCT workflows will validate only paltform external jobs which is declared as supported.

PR #77464, opened 2026-04-07
NO-JIRA: Skip DRA ResourceSlice Controller creates slices test

AnushaIH11:
Skipping DRA testcase:

[sig-node] [DRA] ResourceSlice Controller creates slices

Observed failures in Libvirt job

This change adds the test to excluded_tests for affected jobs.

This is a temporary skip and will be re-enabled once the issue is resolved.


ci-operator/step-registry/openshift/e2e/libvirt/conf/openshift-e2e-libvirt-conf-commands.sh
KeerthanaAP, line 279: @AnushaIH11  Could you add a comment here, stating that "Skipping the DRA test case due to its current failure in libvirt CI. It will be re-enabled once the issue is resolved."

AnushaIH11: Sure, I’ll add the comment and update the PR.

PR #77498, opened 2026-04-07
vsphere VCM: use shared ipi-ci-clusters pool for UPI jobs

jcpowermac:
Align UPI with IPI on vsphere-elastic: failureDomains and govc env now target the pre-provisioned ipi-ci-clusters resource pool. Emit vsphere_install_resource_pool_name in variables.ps1 for PowerCLI.

Remove unused IPI flag now that UPI/IPI pool paths match.

Made-with: Cursor


jcpowermac:
related: https://github.com/openshift/installer/pull/10473

PR #77517, opened 2026-04-08
remove 4.23 periodic tests

jianli-wei:
- remove 4.23 tests under openshift-tests-private and verification-tests, as they are scheduled the same time as 4.22 ones, which are likely causing conflicts on acquiring quota slices
- we'll batch adding 4.22 tests to 4.23 later


jianlinliu:
@jianli-wei let us remove the whole 4.23 ci yaml files, per the OCP roadmap, 5.0 will be the next release, 4.23 will be released after 5.0, we can add 4.23 later when we are close to 5.0 GA.

PR #77544, opened 2026-04-08
MPIIT: Add support for skipJobTriggers switch for triggered job lists

oharan2:
This PR adds an optional top-level boolean `skipJobTriggers` on each job list Json object.

When set to true, Gangway-API job triggering is skipped for that object’s entire `jobList`; post tasks still run. 
Omitted or false preserves existing behavior.

`trigCondFlgs` masks only control which trigger-condition steps run; they do not turn off the whole block. 
This gives an explicit vault-driven switch for “disable all jobs in this group” without misusing trigCond or toggling every active field.

Will be verified using the following `jobList`:
```
{"jobType": "periodic", "jobName": "periodic-ci-CSPI-QE-MSI-openshift-ci-trigger-poc-test-fail-setup", "active": true}
```
This job is failing on setup, so if triggered, nothing happens.

### DIFF
--- a/ci-operator/step-registry/trigger-jobs/README.md
+++ b/ci-operator/step-registry/trigger-jobs/README.md
@@ -19,6 +19,7 @@ The two sources of data are:
       ```json
       [
         {
+          "skipJobTriggers": false,
           "trigCond": [
             {
               "trigCondFlgs": 1,
@@ -285,7 +286,8 @@ This Step will do the following:
     the bitwise mask `(trigCondFlgs & JT__TRIG_COND_EXEC_FLGS)` evaluates to non-0. If any checked trigger condition is not met, the rest
     of the processing is skipped.
  4. **(If applicable)** Loop through the jobs-to-trigger JSON and trigger each Job that has `active: true` and matches the set trigger condition, if any.
-    This Step is skipped if `JT__SKIP_TRIG_MAIN_JOBS` is non-zero.
+    For each JSON list object, its `jobList` block is skipped when `skipJobTriggers` key is set to true.
+    If the step env `JT__SKIP_TRIG_MAIN_JOBS` is non-zero (set in the `ci-operator` config), every `jobList` item is skipped the same way.
  5. **(If applicable)** Set execution markers so the downstream processes can determine whether to execute post-Steps. Markers are only set when
     the bitwise mask `(postTaskFlgs & JT__POST_TASK_EXEC_FLGS)` evaluates to non-0.
  6. Continue processing until all Jobs have been processed.
--- a/ci-operator/step-registry/trigger-jobs/trigger-jobs-commands.sh
+++ b/ci-operator/step-registry/trigger-jobs/trigger-jobs-commands.sh
@@ -8,6 +8,7 @@ typeset jobList='' jobType='' jobName='' stepName=''
 typeset postTask='' postTaskStep='' postTaskPars=''
 typeset rsp=''
 typeset -i trigCondFlgs=0 postTaskFlgs=0 jobExecType=0
+typeset -i skipJobTriggers=0
 typeset -i dryRun=0 tryLeft=0 retryWait=60
 typeset -a failedJobs=()
 #   https://github.com/kubernetes-sigs/prow/blob/95b2a34128de51a4f618c8d6bb9d0c6b587fd29c/pkg/gangway/gangway.proto#L108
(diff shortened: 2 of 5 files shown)


amp-rh (approved):

ci-operator/step-registry/trigger-jobs/trigger-jobs-commands.sh
```diff
@@ -243,7 +251,12 @@ done 0< <(jq -cr '
         else
             .
         end
-    ) | [(.trigCond//[] | @json), (.jobList//[] | @json), (.postTask//[] | @json)] | @tsv
+    ) | [
+        (.trigCond//[] | @json),
+        (.jobList//[] | @json),
+        (.postTask//[] | @json),
+        (if (.skipJobTriggers // false) then 1 else 0 end | tostring)
+    ] | @tsv
```
etirta, line 259: If a master switch is desired, then it should be treated as if the whole entry is undefined or excluded:
```suggestion
    ) | select(.disabled != true) |
    [(.trigCond//[] | @json), (.jobList//[] | @json), (.postTask//[] | @json)] | @tsv
```
Remove all other changes.

ci-operator/step-registry/trigger-jobs/trigger-jobs-ref.yaml
```diff
@@ -21,6 +21,8 @@ ref:
       ```json
       [
         {
+          "skipJobTriggers": <true|false>,                                      # Optional. Master switch: if true,
+                                                                                # skip all `jobList` triggers; `postTask` still runs.
```
etirta, line 25: First, the master switch Key name could be confused with just skipping the `trigCond`. Following an idiomatic approach and borrowing from other frameworks such as Systemd Units and many CI Configs, a simple Key name like `disabled` can be used without any ambiguity.

Second, use `[...]` to indicate the default value.

```suggestion
          "disabled": <true|[false]>,                                           # Optional. Set to `true` to skip this entry entirely.
```

ci-operator/step-registry/trigger-jobs/README.md
```diff
@@ -285,7 +286,8 @@ This Step will do the following:
     the bitwise mask `(trigCondFlgs & JT__TRIG_COND_EXEC_FLGS)` evaluates to non-0. If any checked trigger condition is not met, the rest
     of the processing is skipped.
  4. **(If applicable)** Loop through the jobs-to-trigger JSON and trigger each Job that has `active: true` and matches the set trigger condition, if any.
-    This Step is skipped if `JT__SKIP_TRIG_MAIN_JOBS` is non-zero.
+    For each JSON list object, its `jobList` block is skipped when `skipJobTriggers` key is set to true.
+    If the step env `JT__SKIP_TRIG_MAIN_JOBS` is non-zero (set in the `ci-operator` config), every `jobList` item is skipped the same way.
  5. **(If applicable)** Set execution markers so the downstream processes can determine whether to execute post-Steps. Markers are only set when
     the bitwise mask `(postTaskFlgs & JT__POST_TASK_EXEC_FLGS)` evaluates to non-0.
  6. Continue processing until all Jobs have been processed.
```
etirta, line 293:
```suggestion
 3. Loop through `JT__TRIG_JOB_LIST`, skipping entries with `disabled: true`.
 4. **(If applicable)** Check trigger conditions written by pre-Steps (e.g., resource ownership). Trigger conditions are only checked when
    the bitwise mask `(trigCondFlgs & JT__TRIG_COND_EXEC_FLGS)` evaluates to non-0. If any checked trigger condition is not met, the rest
    of the processing is skipped.
 5. **(If applicable)** Loop through the jobs-to-trigger JSON and trigger each Job that has `active: true` and matches the set trigger condition, if any.
    This Step is skipped if `JT__SKIP_TRIG_MAIN_JOBS` is non-zero.
 6. **(If applicable)** Set execution markers so the downstream processes can determine whether to execute post-Steps. Markers are only set when
    the bitwise mask `(postTaskFlgs & JT__POST_TASK_EXEC_FLGS)` evaluates to non-0.
 7. Continue processing until all Jobs have been processed.
```

etirta (changes requested):
I am okay with this more streamlined implementation.

I would like to see a test using the CI Operator Emulator for a `JT__TRIG_JOB_LIST` that contains two entries: one with `disabled: true` and the other without that key.

PR #77600, opened 2026-04-09
CNTRLPLANE-3197 (hypershift): fix image refs in 4.22 presubmits

clebs:
Correct all the back-release refs to be appropriate for 4.22.

### DIFF
--- a/ci-operator/config/openshift/hypershift/openshift-hypershift-release-4.22.yaml
+++ b/ci-operator/config/openshift/hypershift/openshift-hypershift-release-4.22.yaml
@@ -35,11 +35,7 @@ images:
     to: hypershift-tests
 promotion:
   to:
-  - disabled: true
-    excluded_images:
-    - hypershift-operator
-    - hypershift-tests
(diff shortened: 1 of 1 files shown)


ci-operator/config/openshift/hypershift/openshift-hypershift-release-4.22.yaml
```diff
@@ -63,431 +59,431 @@ releases:
...
-      version: "4.22"
+      version: "4.19"
   n4minor:
     candidate:
       product: ocp
       stream: ci
-      version: "4.22"
+      version: "4.18"
 resources:
-  '*':
```
bryan-cox, line 108: Why the change to double quote?

clebs: I think that was my formatter. Same wth all the spacing and so on.

sjenning:
superseded by https://github.com/openshift/release/pull/77738

PR #77601, opened 2026-04-09
Remove old kas gcp ci

gangwgr:
Remove old kas gcp ci as case migrated to ote ci job


ci-operator/config/openshift/cluster-kube-apiserver-operator/openshift-cluster-kube-apiserver-operator-release-4.22.yaml
```diff
@@ -85,20 +85,6 @@ tests:
   steps:
     cluster_profile: openshift-org-aws
     workflow: openshift-e2e-aws
-- as: e2e-gcp-operator
```
p0lyn0mial, line 88: could we just remove it from the main branch ?

p0lyn0mial: we made the -ote ci job required only on the master.

PR #77649, opened 2026-04-10
gather-ibmcloud-resource: skip delete_pending LBs

Pacho20:
Some test cases create Load balancers. The cleanup part of the tests trigger deletion but that takes time on IBM Cloud and the gather-ibmcloud-resource starts by then. These Load balancers are in "delete_pending" state. Load balancers in "delete_pending" state can be fully removed between the initial listing and the sub-resource queries (lb-ls, lb-ps, lb-pms), causing "load_balancer_not_found" errors that fail the gather step under set -o errexit.

Replace the awk-based LB listing with a jq filter on JSON output that excludes delete_pending LBs while preserving all other states.

## Summary by CodeRabbit

Updates OpenShift CI’s IBM Cloud resource-gathering step (`ci-operator/step-registry/gather/ibmcloud-resource/gather-ibmcloud-resource-commands.sh`) to make collection more resilient and avoid transient failures:

- When collecting IBM Cloud load balancers, it now lists LB IDs via `ibmcloud is lbs --output JSON` and filters with `jq` to exclude load balancers in `delete_pending` provisioning state, preventing `load_balancer_not_found` errors when those LBs are removed between listing and sub-resource queries.
- During the generic per-resource “list → per-item detail query” phase, it now treats per-item query failures as non-fatal by logging a warning to stderr instead of failing the gather step—handling cases where resources are deleted mid-collection.

### DIFF
--- a/ci-operator/step-registry/gather/ibmcloud-resource/gather-ibmcloud-resource-commands.sh
+++ b/ci-operator/step-registry/gather/ibmcloud-resource/gather-ibmcloud-resource-commands.sh
@@ -84,7 +84,28 @@ function ibmcloud_login {
 
 # Gather load-balancer resources
 function gather_lb_resources {
-    mapfile -t LBS < <("${IBMCLOUD_CLI}" is lbs -q | awk -v filter="${CLUSTER_FILTER}" '$0 ~ filter {print $1}')
+    local lb_ids
+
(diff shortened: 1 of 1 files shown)


Pacho20:
/pj-rehearse periodic-ci-openshift-verification-tests-main-installer-rehearse-4.22-installer-rehearse-ibmcloud

PR #75653, opened 2026-03-04
fix baseui image built failure

kuiwang02:
/cc @Xia-Zhao-rh 

the images-images job on main branch is failing, for example https://qe-private-deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/qe-private-deck/pr-logs/pull/openshift_openshift-tests-private/29307/pull-ci-openshift-openshift-tests-private-main-images-images/2029010682741002240

Assisted-By: Claude code

PR #77657, opened 2026-04-10
Update openshift-tests-private release-4.21 images variant to Go 1.25

Amoghrd:
## Summary

This PR updates the `__images.yaml` variant config that was missed in PR #77609.

The `images-images` presubmit check uses this variant configuration, which is why it continued to fail on https://github.com/openshift/openshift-tests-private/pull/29722 even after the main configs were updated to Go 1.25.

## Changes

- `ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.21__images.yaml`:
  - Update `build_root` to use `rhel-9-golang-1.25-openshift-4.21` (was `rhel-9-golang-1.24-openshift-4.21`)

## Context

Related to:
- #77609 (updated main configs but missed variant config)
- https://github.com/openshift/openshift-tests-private/pull/29722 (backport that requires Go 1.25)

### DIFF
--- a/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.21__images.yaml
+++ b/ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.21__images.yaml
@@ -20,7 +20,7 @@ build_root:
   image_stream_tag:
     name: builder
     namespace: ocp
-    tag: rhel-9-golang-1.24-openshift-4.21
+    tag: rhel-9-golang-1.25-openshift-4.21
 images:
   items:
   - dockerfile_literal: |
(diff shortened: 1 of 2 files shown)


memodi:
/pj-rehearse pull-ci-openshift-openshift-tests-private-release-4.21-images-images

Amoghrd:
## Updated PR to include Dockerfile fix

I've updated this PR to also backport the Dockerfile fix from #75653 (which fixed main and 4.22 but wasn't backported to 4.21).

### Why the rehearse job was failing:

The Dockerfile in the `__images.yaml` file used:
```bash
yum config-manager --disable ubi-9-appstream ubi-9-baseos ubi-9-codeready-builder
```

But the base image uses `microdnf` which doesn't support the `config-manager` subcommand, causing:
```
error: Unknown command: 'config-manager'
```

### What I fixed:

1. **Go 1.25 upgrade** (original intent):
   - Updated `build_root` to use `rhel-9-golang-1.25-openshift-4.21`

2. **Dockerfile fix** (backport from #75653):
   - Replaced `yum config-manager --disable ...` with `--disablerepo=` flags
   - Updated package removal/reinstall logic for subscription-manager and sos
   - Made `tests-private-baseui` image optional

This should now pass the rehearse check.

Amoghrd:
/pj-rehearse pull-ci-openshift-openshift-tests-private-release-4.21-images-images

kapjain-rh:
looks good to me but one point if we are putting `optional : true` means are we skipping these jobs?

Amoghrd:
seems like images-images check is optional on main as well. I pulled all the changes made while upgrading main to go1.25, seems like that was added as well

ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.21__images.yaml
```diff
@@ -98,6 +98,7 @@ images:
           yum install -y ./google-chrome-stable_current_*.rpm && rm -f ./google-chrome-stable_current_*.rpm && \
           yum clean all && rm -rf /var/cache/yum/* /tmp/*
     from: tools
+    optional: true
```
memodi, line 101: why are we setting it as optional? this should be required test so that images in CI are in working order?

Amoghrd: In this PR we are making the changes made in PR https://github.com/openshift/release/pull/75653 to fix the image checks. That set the `optional: true` so I am bringing the same changes to 4.21 as well

PR #77684, opened 2026-04-12
OAPE-693:adds coverage test for the external-secrets-operator using Codecov

siddhibhor-56:
Adds the e2e code coverage collector for the External-secrets operator.

### DIFF
--- a/ci-operator/config/openshift/external-secrets-operator/openshift-external-secrets-operator-main.yaml
+++ b/ci-operator/config/openshift/external-secrets-operator/openshift-external-secrets-operator-main.yaml
@@ -15,6 +15,8 @@ images:
   - dockerfile_path: images/ci/Dockerfile
     from: base-rhel9
     to: external-secrets-operator
+  - dockerfile_path: images/ci/Dockerfile.coverage
+    to: external-secrets-operator-coverage
   - dockerfile_path: images/ci/operand.Dockerfile
     from: base-rhel9
     to: external-secrets
@@ -73,6 +75,21 @@ tests:
     version: "4.21"
   skip_if_only_changed: ^(docs/|bundle/|tools/)|\.md$|^(?:.*/)?(?:\.gitignore|.dockerignore|.golangci.yml|Makefile|OWNERS|PROJECT|LICENSE)$
   steps:
+    post:
+    - as: collect-coverage
+      best_effort: true
+      cli: latest
+      commands: hack/e2e-coverage.sh collect
+      credentials:
+      - mount_path: /var/run/secrets/codecov
+        name: external-secrets-operator-codecov-token
+        namespace: test-credentials
+      from: src
+      resources:
+        requests:
+          cpu: 100m
+      timeout: 15m0s
+    - chain: gather
     test:
     - as: install
       cli: latest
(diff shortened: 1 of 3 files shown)


rausingh-rh:
/pj-rehearse pull-ci-openshift-external-secrets-operator-main-e2e-operator

ci-operator/config/openshift/external-secrets-operator/openshift-external-secrets-operator-main.yaml
```diff
@@ -87,6 +104,74 @@ tests:
...
+      - env: COVERAGE_IMAGE
+        name: external-secrets-operator-coverage
+      from: src
+      resources:
+        requests:
+          cpu: 100m
+    - as: test
+      cli: latest
+      commands: |
+        E2E_GINKGO_LABEL_FILTER='"Platform: isSubsetOf {AWS} && !TechPreview"' make test-e2e
```
bharath-b-rh, line 120: Do we need to execute the suite twice? One here and another in `publish-e2e-coverage`. Can't we merge all into e2e-operator job?

ci-operator/config/openshift/external-secrets-operator/openshift-external-secrets-operator-main.yaml
bharath-b-rh, line 99: Can we move this to Makefile with elaborate check like [here](https://github.com/openshift/cert-manager-operator/blob/master/Makefile#L288-L307)

ci-operator/config/openshift/external-secrets-operator/openshift-external-secrets-operator-main.yaml
siddhibhor-56: In the `publish-e2e-coverage` we are not executing the e2e tests we are just publishing the coverage result to the codecov dashboard

bharath-b-rh: Maybe I am missing something, [[1]](https://github.com/siddhibhor-56/release/blob/f730d380040ff4c260f35856e91b0b1a3c208352/ci-operator/config/openshift/external-secrets-operator/openshift-external-secrets-operator-main.yaml#L154-L157) and [[2]](https://github.com/siddhibhor-56/release/blob/f730d380040ff4c260f35856e91b0b1a3c208352/ci-operator/config/openshift/external-secrets-operator/openshift-external-secrets-operator-main.yaml#L175-L182) are not part of `publish-e2e-coverage` flow?

ci-operator/config/openshift/external-secrets-operator/openshift-external-secrets-operator-main.yaml
siddhibhor-56: Good point, but the install step uses `from: operator-sdk` (for the operator-sdk binary), while the Makefile is only available in `from: src`. They're different container images, so we can't call a Makefile target from the install step. Moving this would require pre-installing operator-sdk in the src image which is a broader change right.

ci-operator/config/openshift/external-secrets-operator/openshift-external-secrets-operator-main.yaml
siddhibhor-56: Sorry for the confusion. You are right, the postsubmit does re-run E2E tests.
Each CI job gets its own fresh cluster claim. When the presubmit (PR) job finishes, its cluster is destroyed along with all coverage data stored in an emptyDir. There's no way to carry data from a presubmit cluster to a postsubmit cluster.
The postsubmit runs after merge to main and uploads coverage tagged as the "main branch baseline." Codecov uses this baseline to show deltas on future PRs, like- PR #xyz increased coverage by +3% vs main. Without it, Codecov can only show absolute numbers ("coverage is 65%") with no comparison.

bharath-b-rh:
/pj-rehearse pull-ci-openshift-external-secrets-operator-main-publish-e2e-coverage

PR #77736, opened 2026-04-13
Prevent latest apollo-ci tag on stackrox release branch configs

davdhacs:
## Summary
- Add validation to `config.py` that rejects release branch configs using the `latest` floating tag for the build root image (apollo-ci)
- The `latest` tag is a moving target intended only for validation before promoting to `stable`
- This check runs as the existing `stackrox-stackrox-check` presubmit on openshift/release PRs that modify stackrox configs

## Summary by CodeRabbit

* **StackRox CI validation**
  * Validates the `apollo-ci` build root tag in StackRox configurations.
  * Rejects release branch configurations that use `latest` or `stable`.
  * Reports `latest` tags in master configurations for correction.
  * Enforces these rules through the existing `stackrox-stackrox-check` presubmit.

### DIFF
--- a/ci-operator/config/stackrox/stackrox/config.py
+++ b/ci-operator/config/stackrox/stackrox/config.py
@@ -776,13 +776,13 @@ def check_configs(data):
         #   allow -stable, not -latest, for master
         #   no -stable or -latest tags for release branches
         # > latest is OK for in-flight PRs, but the check should still fail, to prevent merging such PR
+        tag = config.build_root_tag
         if config.branch == 'master':
             if 'latest' in tag:
                 check_error(f"{config.short_filename}: master branch '{config.branch}' "
                             f"must not use 'latest' build root tag '{tag}'. "
                             f"Pin to 'stable' or a specific version (e.g. stackrox-ui-test-0.5.7).")
         if config.branch.startswith('release-'):
-            tag = config.build_root_tag
             if not tag:
                 break
             if 'latest' in tag:


porridge:
@davdhacs I think we want:
- only allow `-X.Y.Z` tags for release branches
- only allow `-stable` or `-X.Y.Z` for `master`
- `-latest` is OK for in-flight PRs, but the check should still fail, to prevent merging such PR

ci-operator/config/stackrox/stackrox/config.py
```diff
@@ -761,12 +761,41 @@ def check_configs(data):
...
 
     for config in data.configs:
+        # 1. Check each job configuration
         # check config for master and the release template - could also cover
         # release branches if this check proves useful.
         if config.branch == 'master' or config.branch == 'release-x.y':
             if not check_config(config, data):
                 configs_ok = False
 
+        # 1. Check the rox-ci-image tag
```
porridge, line 771: Why add a section here instead of in `check_config`?

ci-operator/config/stackrox/stackrox/config.py
```diff
@@ -761,12 +761,41 @@ def check_configs(data):
...
+        # Check master/release branch configs do not use the "latest" floating
+        # tag for the rox-ci-image (apollo-ci). The "latest" tag is a moving
+        # target and should only be used for validation, not for CI jobs.
+        # Rules:
+        #   allow -stable, not -latest, for master
+        #   no -stable or -latest tags for release branches
+        # > latest is OK for in-flight PRs, but the check should still fail, to prevent merging such PR
+        tag = config.build_root_tag
+        if config.branch == 'master':
+            if 'latest' in tag:
```
porridge, line 781: This seems to look for a substring which is a bit sloppy, I think it should instead use `endswith`?
Similar below.

ci-operator/config/stackrox/stackrox/config.py
```diff
@@ -761,12 +761,41 @@ def check_configs(data):
...
+                check_error(f"{config.short_filename}: master branch '{config.branch}' "
```
porridge, line 782: So this reports an error, but still passes? 🤔

ci-operator/config/stackrox/stackrox/config.py
porridge, line 898: Maybe mention the image versioning rules briefly here too and advise looking above for actual error?

PR #77807, opened 2026-04-14
Output version and md5sum of downloaded rpm

tbuskey:
After files is downloaded and checked for bad URL, output its:
    ls -lh
    rpm -q
    md5sum

For tracking tests, its important to track the rpm version

This will output the rpm version that is put on the nodes into the artifacts
If the RPM isn't supposed to be uploaded, the log will have
`INSTALL_KATA_RPM=***** Do not install the Kata RPM`

dig will have to be updated to track this fully

## Summary by CodeRabbit

* **Chores**
  * Enhanced CI/CD logging output with improved message formatting and additional information display in the Kata RPM build step.
  * Added visibility into RPM version details and checksum values during the build process.

### DIFF
--- a/ci-operator/step-registry/sandboxed-containers-operator/get-kata-rpm/sandboxed-containers-operator-get-kata-rpm-commands.sh
+++ b/ci-operator/step-registry/sandboxed-containers-operator/get-kata-rpm/sandboxed-containers-operator-get-kata-rpm-commands.sh
@@ -44,7 +44,7 @@ if [ $err -ne 0 ]; then
     exit 2
 fi
 
-ls -lh kata-containers.rpm
+echo "upload RPM to workers: $(ls -lh kata-containers.rpm)"
 
 # checks for a bad URL
 if grep -q 'title.*404 Not Found' kata-containers.rpm && \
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/sandboxed-containers-operator/get-kata-rpm/sandboxed-containers-operator-get-kata-rpm-commands.sh
```diff
@@ -55,6 +55,9 @@ if grep -q 'title.*404 Not Found' kata-containers.rpm && \
 fi
 
 kata_rpm_md5sum=$(md5sum kata-containers.rpm | cut -d' ' -f1)
+# output the rpm version
+echo "RPM version: $(rpm -q ./kata-containers.rpm)"
+echo "md5sum: ${kata_rpm_md5sum}"
```
ldoktor: It might be useful to mention this is a local downloaded size/version/md5sum. Additionally it might be more consistent to log a header for the `ls` as well, what do you think? Eg:

```
downloaded RPM: $ls
downloaded RPM version: $rpm
downloaded md5sum: $rpm
```
or `using` or something better (you are native, not me, I'd like to use worker but since we then push it to OCP workers that word might be misleading)

ldoktor: (on the other hand we already have the header, so take it for your consideration, I'm fine either way but I'd prefer to change it as it might be easier to read as a newbie or non-regular user)

tbuskey: Good idea. Changed.
Also makes it easier to pull out for analysis

ldoktor (approved):
thanks, could you please squash them into one commit? (since it's one change)

ldoktor (approved):
Thanks, lgtm, shellcheck is also happy so let's get this in

PR #77947, opened 2026-04-17
Migrate image push and cspr cd to prow

inbharajmani:
https://redhat.atlassian.net/browse/AROSLSRE-523

Migrates aro hcp service ci image builds, image push and cspr cd from github actions to prow

**Why `from: src` was removed:** 
The `from` field in ci-operator overrides the last `FROM` directive in a Dockerfile, which for our multi-stage service builds (backend, frontend, admin, sessiongate) was replacing the minimal `distroless/base:3.0` runtime image with the full `src` build image containing Go toolchain, source code, and build tools. Removing it lets the Dockerfiles use their intended multi-stage pattern — building in the golang stage and producing lean distroless production images.

## Summary by CodeRabbit

* **New Features**
  * Added container image builds for aro-hcp-oc-mirror and aro-hcp-exporter.
  * Added CI stages to publish CI-built images to registry and to deploy CSPR infrastructure.
  * New workflows to deploy services to SVC and MGMT clusters using CI-built images.

* **Chores**
  * Added ownership/approval reviewers for the new CI/CD steps and workflows.
* **CI**
  * New postsubmit jobs to run image-publish and CSPR deployment pipelines.


inbharajmani:
/pj-rehearse pull-ci-Azure-ARO-HCP-main-images-push pull-ci-Azure-ARO-HCP-main-cspr pull-ci-Azure-ARO-HCP-main-e2e-parallel

ci-operator/config/Azure/ARO-HCP/Azure-ARO-HCP-main.yaml
```diff
@@ -466,6 +471,25 @@ tests:
       resource_type: aro-hcp-dev-global-pipeline-quota-slice
     test:
     - ref: aro-hcp-provision-global-pipeline
+- as: images-push
+  postsubmit: true
+  steps:
+    leases:
+    - count: 1
+      env: LEASED_IMAGE_PUSH_RESOURCE
+      resource_type: aro-hcp-dev-image-push-quota-slice
```
roivaz: Any reason two image build jobs can't run at the same time?

inbharajmani: It is to limit the jobs from pushing concurrently to acr

inbharajmani: Ah, I think this might limit our ability to push the latest image to ACR if there are multiple commits merged to main in quick succession - is that the concern?
I checked the Prow documentation and didn’t find anything about automatic cancellation of older post-submit jobs during commit bursts. One option could be to align the lease count with the Tide batch merge limit, which is currently 5 for ARO-HCP

Or completely remove the resource leasing. Any thoughts on this :)

roivaz (changes requested):

ci-operator/config/Azure/ARO-HCP/Azure-ARO-HCP-main.yaml
roivaz: Just remove it. If several jobs run at the same time because 2 PRs got merged at aproximately the same time, that's fine. A job will trigger per ARO-HCP commit pushed to main so an image will be built and pushed per commit - adding the lease just makes that process sequential.

inbharajmani: removed

PR #77979, opened 2026-04-17
Add europe-west4 Region in the openshift-org-gcp profile set

danilo-gemoli:
We have been using only the `us-central1` region so far and, recently, we are noticing the `ZONE_RESOURCE_POOL_EXHAUSTED` in the GCP audit logs.
Add another region to mitigate this kind of quota limit issue. 

## Summary by CodeRabbit

* **Chores**
  * Rebalanced GCP quota allocations by introducing capacity in europe-west4 and reducing corresponding us-central1 allocations, creating a two-region split for several quota slices.
  * Updated public-facing quota identifiers to reflect the europe-west4 region for affected slices, improving regional distribution and availability.

### DIFF
--- a/core-services/prow/02_config/_boskos.yaml
+++ b/core-services/prow/02_config/_boskos.yaml
@@ -4046,6 +4046,21 @@ resources:
   state: free
   type: fleet-manager-qe-quota-slice
 - names:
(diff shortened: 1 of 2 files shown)


danilo-gemoli:
/hold
Next Monday

PR #77996, opened 2026-04-18
fleetshift: add CI config for image build and mirror

mshort55:
Add initial image mirroring CI for the following repos:
https://github.com/fleetshift/fleetshift-poc
https://github.com/fleetshift/fleetshift-user-interface

## Summary by CodeRabbit

* **Chores**
  * Added CI configurations to build multiple container images and set default resource requests.
  * Added automated image-mirroring workflows for PRs and post-merge (latest) pushes to the registry.
  * Added presubmit/postsubmit job definitions and plugin trigger configs for CI automation.
  * Added ownership metadata for the new CI configuration files.

PR #78073, opened 2026-04-21
fleetshift - adding multi arch support for image mirroring

mshort55:
Add multi arch support for fleetshift repo image builds/mirroring. Continuation of #77996 and #78066


mshort55:
@alechenninger can I get your lgtm for this? Thanks!

PR #78235, opened 2026-04-23
fix rosa list version in account role creation step

bmeng:
`rosa list versions` does not have --classic parameter, so set it only when hcp is true

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved account role creation fallback logic to more reliably select versions based on cluster configuration during creation failures.


bmeng:
/pj-rehearse periodic-ci-openshift-release-main-nightly-4.22-e2e-rosa-sts-ovn

PR #78242, opened 2026-04-23
HIVE:3148: [WIP]hive extension binary test

miyadav:
/hold 

depends on - https://github.com/openshift/origin/pull/31059

## Summary by CodeRabbit

* **Tests**
  * Added OTE end-to-end presubmit workflows for AWS, GCP, Azure, and vSphere with provider-specific gating, environment setup, and on-demand PR triggers.
* **Chores**
  * Added a dedicated CI test image to standardize OTE test execution.
  * Removed empty namespace declarations from credential entries to clean up mounts.

### DIFF
--- a/ci-operator/config/openshift/hive/openshift-hive-master.yaml
+++ b/ci-operator/config/openshift/hive/openshift-hive-master.yaml
@@ -26,6 +26,18 @@ images:
     dockerfile_path: Dockerfile
     from: ubi9-minimal-entitled
     to: hive
+  - dockerfile_literal: |
+      FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.25-openshift-4.22 as builder
+      RUN mkdir -p /go/src/github.com/openshift/hive
+      WORKDIR /go/src/github.com/openshift/hive
+      RUN git clone --branch fix-ote-init --depth 1 https://github.com/miyadav/hive.git .
+      RUN make -C test/ote build
+      FROM registry.access.redhat.com/ubi9/ubi-minimal:latest
+      RUN microdnf install -y gzip jq && microdnf clean all
+      COPY --from=builder /go/src/github.com/openshift/hive/test/ote/bin/hive /usr/bin/openshift-tests-extension
+      RUN gzip -f -k /usr/bin/openshift-tests-extension
+      WORKDIR /tmp
+    to: hive-tests
   skip_if_only_changed: ^(?:docs|\.tekton)/|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|PROJECT|LICENSE)$
 promotion:
   to:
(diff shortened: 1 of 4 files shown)


ci-operator/config/openshift/hive/openshift-hive-master.yaml
```diff
@@ -289,6 +291,146 @@ tests:
...
+        name: hive
+      - env: RELEASE_IMAGE
+        name: release:latest
+      from: hive-tests
+      grace_period: 15m0s
+      resources:
+        requests:
+          cpu: 100m
+    workflow: generic-claim
+- as: e2e-pool-ote
```
2uasimojo: As written, this is not different than the above `e2e-ote`. The presubmits for e2e and e2e-pool actually run different test scripts. I don't think such a split is necessary or even possible here unless the OTE suite supports some kind of differentiator input.

ci-operator/config/openshift/hive/openshift-hive-master.yaml
```diff
@@ -289,6 +291,146 @@ tests:
...
+        name: hive
+      - env: RELEASE_IMAGE
+        name: release:latest
+      from: hive-tests
+      grace_period: 15m0s
+      resources:
+        requests:
+          cpu: 100m
+    workflow: generic-claim
+- as: e2e-gcp-ote
```
2uasimojo, line 346: As written, this and e2e-azure-ote and e2e-vsphere-ote appear to differ from e2e-ote only in the platform of the hub cluster.

In general for hive we care much less about the platform of the hub -- which technically is allowed to be any k8s and runs almost identically -- and more about that of the spoke, as that informs:
- A completely different provisioning code path
- Platform-specific knobs in hive CRDs (primarily ClusterDeployment/ClusterPool & MachinePool) and corresponding behavior.

Is the test suite somehow detecting the cloud provider of the hub and creating spokes accordingly? And also filtering platform-specific test cases?

2uasimojo (commented):
As discussed [here](https://github.com/openshift/hive/pull/2894#issuecomment-4315273733), we're going to need to figure out whether OTE suites should be defined as presubmits, periodics, or both. Summarizing, we should consider our ability to trigger these tests:
- On demand on a big/risky PR for more aggressive regression coverage.
- On demand on the tip of `master` e.g. as pre-release/pre-promotion validation.
- Periodically/automatically for ongoing regression peace-of-mind.

Also as indicated inline, I need to understand how platform- (or other-characteristic-) specific test behavior is triggered. Even if a single invocation *can* exercise all platforms, I reckon we have a requirement to at least be able to filter down to a subset. This could be where @newtonheath's tagging suggestion comes into play. For example, "This PR adds a knob to GCP MachinePools, so run all tests with `platform=gcp` and `scope=machinepool`."

miyadav:
/close 
for - https://github.com/openshift/release/pull/78554

PR #78275, opened 2026-04-23
opendatahub: Set runningCount equal to size to disable hibernation

dayakar349:
Set runningCount: 12 (equal to size: 12) on the 4.20 pool so all clusters stay running and never hibernate.

By keeping all clusters running, we eliminate the hibernation resume path entirely until the upstream bug is fixed.

## Summary by CodeRabbit

* **Chores**
  * Updated ClusterPool configuration to increase the number of concurrently running instances, enhancing overall system capacity.

### DIFF
--- a/clusters/hosted-mgmt/hive/pools/opendatahub/opendatahub-ocp-4-20-amd64-aws_clusterpool.yaml
+++ b/clusters/hosted-mgmt/hive/pools/opendatahub/opendatahub-ocp-4-20-amd64-aws_clusterpool.yaml
@@ -34,7 +34,7 @@ spec:
       region: us-east-1
   pullSecretRef:
     name: pull-secret
-  runningCount: 6
+  runningCount: 12
   size: 12
   skipMachinePools: true
 status:


clusters/hosted-mgmt/hive/pools/opendatahub/opendatahub-ocp-4-20-amd64-aws_clusterpool.yaml
macgregor, line 28: whats max concurrent, should match size/runningCount?

2uasimojo: `maxConcurrent` is the maximum number of clusters whose infras we will provision + deprovision at any one time. One would use this setting to throttle cloud account activity e.g. to mitigate rate limiting errors. It doesn't really have a direct correlation to size/runningCount.

dayakar349: It will limit the simultaneous installations but setting it to 6 protects the AWs quota on sudden increase in  load.

dayakar349: without the maxConcurrent there will be no limit on concurrent provisioning.

PR #78380, opened 2026-04-27
changing aro-hcp cleanup-sweeper to run every 4 hours to try to fix throttling issues

sclarkso:
Cleanup job is currently running hourly.  We're seeing a significant amount of 429's on all network operations in the e2e sub, and hoping reducing the frequency of this job helps reduce pressure.

This changes the jobs to run every 4 hours instead.

## Summary by CodeRabbit

* **Chores**
  * Adjusted periodic cleanup job schedules to run every 4 hours instead of hourly, with timing offsets at minutes 5 and 35 respectively.


roivaz:
@sclarkso add this envvar as well `CLEANUP_SWEEPER_PARALLELISM: "1"` to the sweeper so it runs with just a single worker. 4 is the default, but we don't really care if the job just takes longer.

roivaz:
Superseded by https://github.com/openshift/release/pull/78320

PR #78391, opened 2026-04-27
CNTRLPLANE-3258: Gate jira-agent on ready-to-solve label

enxebre:
## Summary

- Adds `labels = ready-to-solve` to the jira-agent JQL query
- Issues must now have the `ready-to-solve` label (in addition to `issue-for-agent`) before the jira-agent will process them
- This ensures issues have been validated for adequate grooming (via `/jira:ready-to-solve`) before automated solving is attempted

Goes with https://github.com/openshift/release/pull/81553

## Test plan

- [x] Verify JQL query returns only issues with both `issue-for-agent` and `ready-to-solve` labels
- [ ] Run jira-agent periodic job and confirm it skips issues without `ready-to-solve`

/cc @bryan-cox

## Summary by CodeRabbit

* **Chores**
  * Updated the Jira agent issue selection criteria to require an additional label status before issues are automatically processed, refining the workflow prioritization.


bryan-cox:
/hold

Do we have an existing job running that adds this label?

bryan-cox:
/hold cancel

We have a periodic job that is assigning the label now.

PR #78416, opened 2026-04-27
MPIIT: Use ExitTrap--PostProcessPrep for updated test mapping

oharan2:
Main changes:

- Instead of handling the installation of yq, TS mapping and original XML archive, use CleanupCollect as a trap function and import the [RedHatQE/OpenShift-LP-QE--Tools](https://github.com/RedHatQE/OpenShift-LP-QE--Tools) `ExitTrap--PostProcessPrep` mechanism
- Update file structure to use best practice principles
- Update CR and DR identifiers to `lp-ocp-compat--OpenshiftPipelines`. Following #78819, update env var to be `DR__RP__CR_COMP_NAME`.

Verified run: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/78416/rehearse-78416-periodic-ci-openshift-pipelines-release-tests-release-v1.21-ocp-4.22-lp-interop-cr-openshift-pipelines-aws/2052313354864693248

- [x] Mapped the XML with the desired prefix used in TS name
- [x] Original results are compressed
- [x] Successfully reporting results to DR

`

## Summary by CodeRabbit

* **Chores**
  * Centralized test-result post-processing into an exit-triggered cleanup flow, standardizing artifact handling and strengthening script error controls while temporarily suppressing command tracing during credential/login steps.
  * Hardened runtime handling (safer quoting and options) and defaulted the operator install channel to "latest" for more robust installs.
  * Added a 10m grace period to the install and test step references.

### DIFF
--- a/ci-operator/config/openshift-pipelines/release-tests/openshift-pipelines-release-tests-release-v1.21__ocp-4.22-lp-interop.yaml
+++ b/ci-operator/config/openshift-pipelines/release-tests/openshift-pipelines-release-tests-release-v1.21__ocp-4.22-lp-interop.yaml
@@ -48,7 +48,7 @@ tests:
       FIREWATCH_FAIL_WITH_TEST_FAILURES: "true"
       MAP_TESTS: "true"
       OCP_VERSION: "4.22"
-      REPORTPORTAL_CMP: lp-interop-OpenshiftPipelines
+      REPORTPORTAL_CMP: lp-ocp-compat--OpenshiftPipelines
       USER_TAGS: |
         scenario pipelines
     test:
@@ -77,7 +77,7 @@ tests:
       FIREWATCH_FAIL_WITH_TEST_FAILURES: "true"
       MAP_TESTS: "true"
       OCP_VERSION: "4.22"
-      REPORTPORTAL_CMP: lp-interop-OpenshiftPipelines
+      REPORTPORTAL_CMP: lp-ocp-compat--OpenshiftPipelines
       USER_TAGS: |
         scenario pipelines
     test:
--- a/ci-operator/config/openshift-pipelines/release-tests/openshift-pipelines-release-tests-release-v1.21__ocp-4.22-lp-interop.yaml
+++ b/ci-operator/config/openshift-pipelines/release-tests/openshift-pipelines-release-tests-release-v1.21__ocp-4.22-lp-interop.yaml
@@ -48,7 +48,7 @@ tests:
       FIREWATCH_FAIL_WITH_TEST_FAILURES: "true"
       MAP_TESTS: "true"
       OCP_VERSION: "4.22"
-      REPORTPORTAL_CMP: lp-ocp-compat--OpenshiftPipelines
+      DR__RP__CR_COMP_NAME: lp-ocp-compat--OpenshiftPipelines
       USER_TAGS: |
         scenario pipelines
     test:
@@ -77,7 +77,7 @@ tests:
       FIREWATCH_FAIL_WITH_TEST_FAILURES: "true"
       MAP_TESTS: "true"
       OCP_VERSION: "4.22"
-      REPORTPORTAL_CMP: lp-ocp-compat--OpenshiftPipelines
+      DR__RP__CR_COMP_NAME: lp-ocp-compat--OpenshiftPipelines
       USER_TAGS: |
         scenario pipelines
     test:
--- a/ci-operator/config/openshift-pipelines/release-tests/openshift-pipelines-release-tests-release-v1.21__ocp-4.22-lp-interop.yaml
+++ b/ci-operator/config/openshift-pipelines/release-tests/openshift-pipelines-release-tests-release-v1.21__ocp-4.22-lp-interop.yaml
@@ -33,6 +33,7 @@ tests:
     cluster_profile: aws-cspi-qe
     env:
       BASE_DOMAIN: cspilp.interop.ccitredhat.com
+      DR__RP__CR_COMP_NAME: lp-ocp-compat--OpenshiftPipelines
       FIREWATCH_CONFIG: |
         {
           "failure_rules":
@@ -48,7 +49,6 @@ tests:
       FIREWATCH_FAIL_WITH_TEST_FAILURES: "true"
       MAP_TESTS: "true"
       OCP_VERSION: "4.22"
-      DR__RP__CR_COMP_NAME: lp-ocp-compat--OpenshiftPipelines
       USER_TAGS: |
         scenario pipelines
     test:
@@ -62,6 +62,7 @@ tests:
     cluster_profile: aws-cspi-qe
     env:
       BASE_DOMAIN: cspilp.interop.ccitredhat.com
+      DR__RP__CR_COMP_NAME: lp-ocp-compat--OpenshiftPipelines
       FIREWATCH_CONFIG: |
         {
           "failure_rules":
@@ -77,7 +78,6 @@ tests:
       FIREWATCH_FAIL_WITH_TEST_FAILURES: "true"
       MAP_TESTS: "true"
       OCP_VERSION: "4.22"
-      DR__RP__CR_COMP_NAME: lp-ocp-compat--OpenshiftPipelines
       USER_TAGS: |
         scenario pipelines
     test:
--- a/ci-operator/step-registry/openshift-pipelines/install/openshift-pipelines-install-commands.sh
+++ b/ci-operator/step-registry/openshift-pipelines/install/openshift-pipelines-install-commands.sh
@@ -1,85 +1,36 @@
 #!/bin/bash
+set -euxo pipefail; shopt -s inherit_errexit
 
-set -o nounset
-set -o errexit
-set -o pipefail
-
-function install_yq_if_not_exists() {
-    # Install yq manually if not found in image
-    echo "Checking if yq exists"
-    cmd_yq="$(yq --version 2>/dev/null || true)"
-    if [ -n "$cmd_yq" ]; then
-        echo "yq version: $cmd_yq"
-    else
-        echo "Installing yq"
-        mkdir -p /tmp/bin
-        export PATH=$PATH:/tmp/bin/
-        curl -L "https://github.com/mikefarah/yq/releases/latest/download/yq_linux_$(uname -m | sed 's/aarch64/arm64/;s/x86_64/amd64/')" \
-         -o /tmp/bin/yq && chmod +x /tmp/bin/yq
-    fi
-}
-
-function mapTestsForComponentReadiness() {
-    if [[ $MAP_TESTS == "true" ]]; then
-        results_file="${1}"
-        echo "Patching Tests Result File: ${results_file}"
-        if [ -f "${results_file}" ]; then
-            export cmp="${REPORTPORTAL_CMP}--"
-            
-            echo "Mapping Test Suite Name To: ${REPORTPORTAL_CMP}"
-            yq eval -px -ox -iI0 '.testsuites.testsuite.+@name |= sub("^(.*)$", env(cmp) + "${1}")' $results_file || echo "Warning: yq failed for ${results_file}, debug manually" >&2
-        fi
-    fi
-}
-
-# Archive results function
-function cleanup-collect() {
-    if [[ $MAP_TESTS == "true" ]]; then
-      install_yq_if_not_exists
-      original_results="${ARTIFACT_DIR}/original_results/"
-      mkdir "${original_results}" || true
-      echo "Collecting original results in ${original_results}"
-
-      # Keep a copy of all the original Junit files before modifying them
-      cp -r "${ARTIFACT_DIR}"/xml-report/ "${original_results}" || echo "Warning: couldn't copy original files" >&2
-
-      # Remove timestamped dir to avoid spacing in filename
-      mv "${ARTIFACT_DIR}"/xml-report/*/result.xml "${ARTIFACT_DIR}/xml-report/" || echo "Warning: couldn't move file to top level dir" >&2
-
-      result_file="${ARTIFACT_DIR}/xml-report/result.xml"
-      # Map tests if needed for related use cases
-      mapTestsForComponentReadiness $result_file
-
-      # Send modified file to shared dir for Data Router Reporter step
-      cp "${result_file}" "${SHARED_DIR}" || echo "Warning: couldn't copy files to SHARED_DIR" >&2
-    fi
-}
-
-CONSOLE_URL=$(cat $SHARED_DIR/console.url)
-API_URL="https://api.${CONSOLE_URL#"https://console-openshift-console.apps."}:6443"
-export CONSOLE_URL
-export API_URL
-export gauge_reports_dir=${ARTIFACT_DIR}
-export overwrite_reports=false
-export KUBECONFIG=$SHARED_DIR/kubeconfig
-export GOPROXY="https://proxy.golang.org/"
+# Map results by setting identifier prefix in tests suites names for reporting tools
+# Merge original results into a single file and compress
+# Send modified file to shared dir for Data Router Reporter step
+if [ "${MAP_TESTS}" = "true" ]; then
+    eval "$(
+        curl -fsSL \
+https://raw.githubusercontent.com/RedHatQE/OpenShift-LP-QE--Tools/refs/heads/main/libs/bash/ci-operator/interop/common/ExitTrap--PostProcessPrep.sh
+    )"; trap '
+        LP_IO__ET_PPP__NEW_TS_NAME="${REPORTPORTAL_CMP}--%s" \
+            ExitTrap--PostProcessPrep junit--openshift-pipelines__install__openshift-pipelines-install.xml
+    ' EXIT
+fi
 
 # Add timeout to ignore runner connection error
 gauge config runner_connection_timeout 600000 && gauge config runner_request_timeout 300000
 
 # login for interop
-if test -f ${SHARED_DIR}/kubeadmin-password
-then
-  OCP_CRED_USR="kubeadmin"
-  export OCP_CRED_USR
-  OCP_CRED_PSW="$(cat ${SHARED_DIR}/kubeadmin-password)"
-  export OCP_CRED_PSW
-  oc login -u kubeadmin -p "$(cat $SHARED_DIR/kubeadmin-password)" "${API_URL}" --insecure-skip-tls-verify=true
+if [ -s "${KUBECONFIG}" ]; then
+    oc whoami
 else #login for ROSA & Hypershift platforms
-  eval "$(cat "${SHARED_DIR}/api.login")"
+    (set +x; eval "$(cat "${SHARED_DIR}/api.login")")
 fi
 
-echo "Running olm.spec to install operator"
-CATALOG_SOURCE=redhat-operators CHANNEL=${OLM_CHANNEL} gauge run --log-level=debug --verbose --tags install specs/olm.spec || true
+# Install openshift-pipelines operator (olm.spec)
+CONSOLE_URL="$(oc whoami --show-console)" \
+    API_URL="$(oc whoami --show-server)" \
+    CATALOG_SOURCE=redhat-operators \
+    CHANNEL="${OLM_CHANNEL:-latest}" \
+    gauge_reports_dir="${ARTIFACT_DIR}" \
+    overwrite_reports=false \
+    gauge run --log-level=debug --verbose --tags install specs/olm.spec
 
-cleanup-collect
+true
--- a/ci-operator/step-registry/openshift-pipelines/install/openshift-pipelines-install-commands.sh
+++ b/ci-operator/step-registry/openshift-pipelines/install/openshift-pipelines-install-commands.sh
@@ -9,7 +9,7 @@ if [ "${MAP_TESTS}" = "true" ]; then
         curl -fsSL \
 https://raw.githubusercontent.com/RedHatQE/OpenShift-LP-QE--Tools/refs/heads/main/libs/bash/ci-operator/interop/common/ExitTrap--PostProcessPrep.sh
     )"; trap '
-        LP_IO__ET_PPP__NEW_TS_NAME="${REPORTPORTAL_CMP}--%s" \
+        LP_IO__ET_PPP__NEW_TS_NAME="${DR__RP__CR_COMP_NAME}--%s" \
             ExitTrap--PostProcessPrep junit--openshift-pipelines__install__openshift-pipelines-install.xml
     ' EXIT
 fi
--- a/ci-operator/step-registry/openshift-pipelines/tests/openshift-pipelines-tests-commands.sh
+++ b/ci-operator/step-registry/openshift-pipelines/tests/openshift-pipelines-tests-commands.sh
@@ -12,7 +12,7 @@ if [ "${MAP_TESTS}" = "true" ]; then
         curl -fsSL \
 https://raw.githubusercontent.com/RedHatQE/OpenShift-LP-QE--Tools/refs/heads/main/libs/bash/ci-operator/interop/common/ExitTrap--PostProcessPrep.sh
     )"; trap '
-        LP_IO__ET_PPP__NEW_TS_NAME="${REPORTPORTAL_CMP}--%s" \
+        LP_IO__ET_PPP__NEW_TS_NAME="${DR__RP__CR_COMP_NAME}--%s" \
             ExitTrap--PostProcessPrep junit--openshift-pipelines__tests__openshift-pipelines-tests.xml
     ' EXIT
 fi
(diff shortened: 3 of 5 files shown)


ci-operator/step-registry/openshift-pipelines/install/openshift-pipelines-install-commands.sh
```diff
@@ -1,60 +1,24 @@
...
-      mapTestsForComponentReadiness $result_file
-
+      # Map results by setting identifier prefix in tests suites names for reporting tools
+      # Merge original results into a single file and compress
       # Send modified file to shared dir for Data Router Reporter step
-      cp "${result_file}" "${SHARED_DIR}" || echo "Warning: couldn't copy files to SHARED_DIR" >&2
+      export LP_IO__ET_PPP__NEW_TS_NAME="${REPORTPORTAL_CMP}--%s"
+      eval "$(
+          curl -fsSL \
+      https://raw.githubusercontent.com/RedHatQE/OpenShift-LP-QE--Tools/refs/heads/main/libs/bash/ci-operator/interop/common/ExitTrap--PostProcessPrep.sh
```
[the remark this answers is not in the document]
etirta: @coderabbitai No, it will create a maintenance nightmare. An update on the library repo would require tons of PRs against this repo.

The nature of the test is not reproducible anyway, because we don't peg the OCP version (it's using the head of the channel when the test is run).

ci-operator/step-registry/openshift-pipelines/install/openshift-pipelines-install-commands.sh
```diff
@@ -1,60 +1,23 @@
...
+      # Map results by setting identifier prefix in tests suites names for reporting tools
+      # Merge original results into a single file and compress
       # Send modified file to shared dir for Data Router Reporter step
-      cp "${result_file}" "${SHARED_DIR}" || echo "Warning: couldn't copy files to SHARED_DIR" >&2
-    fi
+      export LP_IO__ET_PPP__NEW_TS_NAME="${REPORTPORTAL_CMP}--%s"
+      eval "$(
+          curl -fsSL \
+      https://raw.githubusercontent.com/RedHatQE/OpenShift-LP-QE--Tools/refs/heads/main/libs/bash/ci-operator/interop/common/ExitTrap--PostProcessPrep.sh
+      )"
```
etirta: The `EXIT` trap is essentially an interrupt handler. As a rule of thumb, an interrupt handler should only perform minimal activity. It is not a good idea to fetch an external library inside an interrupt handler. Please move the actual fetching to outside the trap handler.

In fact, it's not a good idea to use another function as a thin wrapper, as every function call introduces runtime context overhead. This approach is better:

```shell
if [ "${MAP_TESTS}" = "true" ]; then
    eval "$(
        curl -fsSL \
https://raw.githubusercontent.com/RedHatQE/OpenShift-LP-QE--Tools/refs/heads/main/libs/bash/ci-operator/interop/common/ExitTrap--PostProcessPrep.sh
    )"; trap '
        LP_IO__ET_PPP__NEW_TS_NAME="${REPORTPORTAL_CMP}--%s" \
            ExitTrap--PostProcessPrep junit--openshift-pipelines__tests__openshift-pipelines-tests.xml
    ' EXIT
fi
```

ci-operator/step-registry/openshift-pipelines/install/openshift-pipelines-install-commands.sh
etirta: Any reason why we are using a login with a username and password instead of `KUBECONFIG`? The `kubeadmin` user is actually meant for Web Console login. The `oc` command is meant to use `KUBECONFIG`.

Please find out why we need to define the Env. Vars `OCP_CRED_USR` and `OCP_CRED_PSW`. Is it used by Gauge?

There is a better way to protect secrets than the `set +x; ... ; set -x` block, but it depends on exactly how it's meant to be used. Please figure it out so I can suggest the best secret handling method for this case.

Also let me know what is the content of `${SHARED_DIR}/api.login`, so I can suggest accordingly.

ci-operator/step-registry/openshift-pipelines/install/openshift-pipelines-install-commands.sh
```diff
@@ -1,60 +1,23 @@
...
+      ExitTrap--PostProcessPrep junit--openshift-pipelines__tests__openshift-pipelines-tests.xml
```
etirta: Please confirm that the junit XML file name is `junit--openshift-pipelines__tests__openshift-pipelines-tests.xml`. Why is it using `tests`? We are in the `install` Step.

If a single XML junit file is desired, why not use a simple name like `junit--openshift-pipelines.xml`?

ci-operator/step-registry/openshift-pipelines/install/openshift-pipelines-install-commands.sh
etirta: The `KUBECONFIG` environment variable should be controlled by the CI Operator. Users should not change it unless there is a specific reason. For example, if you need to temporarily communicate with another cluster, using the `--kubeconfig` CLI option is preferred.

etirta (changes requested):
The files still have violations of our Best Practices:
- Using `echo` while `xtrace` is active.
- Missing `true` at the end of functions and scripts.
- Some variables are not pre-declared.
- etc.

This suggests the AI Agent may not be configured according to our Best Practices.

More details are in the reviews below. Bear in mind that I only flag one occurrence; it does not mean there are no other occurrences of similar issues. Please tell the AI Agent to scan all affected files for similar issues.

ci-operator/step-registry/openshift-pipelines/install/openshift-pipelines-install-commands.sh
oharan2: Pending PQE's answer, thanks.

ci-operator/step-registry/openshift-pipelines/tests/openshift-pipelines-tests-commands.sh
```diff
@@ -1,99 +1,60 @@
...
 gauge run --log-level=debug --verbose --tags sanity specs/operator/rbac.spec || true
 
-echo "Rename xml files to junit_test_*.xml"
-readarray -t path <<< "$(find ${ARTIFACT_DIR}/xml-report/ -name '*.xml')"
+# Rename xml-report outputs to junit_test_*.xml for collectors
+readarray -t path <<< "$(find "${ARTIFACT_DIR}/xml-report" -name '*.xml')"
 for index in "${!path[@]}"; do
-  mv "${path[index]}" ${ARTIFACT_DIR}/junit_test_result$[index+1].xml
+  mv "${path[index]}" "${ARTIFACT_DIR}/junit_test_result$((index + 1)).xml"
 done
```
etirta: Wouldn't this cause `${ARTIFACT_DIR}/xml-report/` to be an empty dir.? It's probably better to rename only and not move to different directory.
```suggestion
typeset f=
typeset -i i=0
while IFS= read -r f; do
    mv "${f}" "${f%/*}/junit_test_result$((++i)).xml"
done < <(find "${ARTIFACT_DIR}/xml-report" -name '*.xml')
```

etirta (changes requested):

ci-operator/step-registry/openshift-pipelines/install/openshift-pipelines-install-commands.sh
srivickynesh: this was followed across most services in step registry. feel free to change it. gauge don't use OCP_CRED_USR and OCP_CRED_PSW env vars in tests.

ci-operator/step-registry/openshift-pipelines/install/openshift-pipelines-install-commands.sh
```diff
@@ -1,85 +1,46 @@
...
-CONSOLE_URL=$(cat $SHARED_DIR/console.url)
+CONSOLE_URL=$(cat "${SHARED_DIR}/console.url")
 API_URL="https://api.${CONSOLE_URL#"https://console-openshift-console.apps."}:6443"
 export CONSOLE_URL
 export API_URL
-export gauge_reports_dir=${ARTIFACT_DIR}
+export gauge_reports_dir="${ARTIFACT_DIR}"
 export overwrite_reports=false
-export KUBECONFIG=$SHARED_DIR/kubeconfig
+export KUBECONFIG="${SHARED_DIR}/kubeconfig"
```
etirta: Since the Cluster Under Test (CUT) has been deployed in this step, the `KUBECONFIG` should have already been set by the CI operator.
```suggestion
```

ci-operator/step-registry/openshift-pipelines/install/openshift-pipelines-install-commands.sh
```diff
@@ -1,85 +1,46 @@
...
-      mapTestsForComponentReadiness $result_file
-
-      # Send modified file to shared dir for Data Router Reporter step
-      cp "${result_file}" "${SHARED_DIR}" || echo "Warning: couldn't copy files to SHARED_DIR" >&2
-    fi
-}
-
-CONSOLE_URL=$(cat $SHARED_DIR/console.url)
+CONSOLE_URL=$(cat "${SHARED_DIR}/console.url")
 API_URL="https://api.${CONSOLE_URL#"https://console-openshift-console.apps."}:6443"
```
etirta: Since `KUBECONFIG` points to the CUT, if the `oc` binary is present in this run-time env., we can simply run:
```suggestion
CONSOLE_URL="$(oc whoami --show-console)"
API_URL="$(oc whoami --show-server)"
```
This ensures 100% accuracy and avoids the risk of manual parsing failures if the API URL structure changes in future versions.

ci-operator/step-registry/openshift-pipelines/install/openshift-pipelines-install-commands.sh
```diff
@@ -1,85 +1,46 @@
...
   export OCP_CRED_USR
-  OCP_CRED_PSW="$(cat ${SHARED_DIR}/kubeadmin-password)"
+  OCP_CRED_PSW="$(cat "${SHARED_DIR}/kubeadmin-password")"
   export OCP_CRED_PSW
-  oc login -u kubeadmin -p "$(cat $SHARED_DIR/kubeadmin-password)" "${API_URL}" --insecure-skip-tls-verify=true
+  oc login -u kubeadmin -p "$(cat "${SHARED_DIR}/kubeadmin-password")" "${API_URL}" --insecure-skip-tls-verify=true
 else #login for ROSA & Hypershift platforms
   eval "$(cat "${SHARED_DIR}/api.login")"
 fi
+set -x
```
etirta: Let's use the `KUBECONFIG` and verify the login status instead.
```suggestion
if [ -s "${KUBECONFIG}" ]; then
    oc whoami
else #login for ROSA & Hypershift platforms
    eval "$(cat "${SHARED_DIR}/api.login")"
fi
```
Please investigate the contents of `${SHARED_DIR}/api.login` so we can determine whether we need to protect secrets. If secret protection is necessary and the login script does not set variables for the global context, change the `eval` to:
```shell
    (set +x; eval "$(cat "${SHARED_DIR}/api.login")")
```
If the login script does set variables for the global context, use:
```shell
    set +x; eval "$(cat "${SHARED_DIR}/api.login")"; set -x
```

ci-operator/step-registry/openshift-pipelines/install/openshift-pipelines-install-commands.sh
etirta: Since the Cluster Under Test (CUT) has been deployed in this step, the `KUBECONFIG` should have already been set by the CI operator. And if the `oc` binary is present in this run-time env., we should use it to obtain the Console & API Server URL to ensures 100% accuracy and avoids the risk of manual parsing failures if those URL structure changes in future versions:
```suggestion
```
Let's do inline export when `gauge` is called for precise export env. as per our Best Practices.

ci-operator/step-registry/openshift-pipelines/install/openshift-pipelines-install-commands.sh
```diff
@@ -1,85 +1,46 @@
...
 
-echo "Running olm.spec to install operator"
-CATALOG_SOURCE=redhat-operators CHANNEL=${OLM_CHANNEL} gauge run --log-level=debug --verbose --tags install specs/olm.spec || true
+# Install openshift-pipelines operator (olm.spec)
+CATALOG_SOURCE=redhat-operators CHANNEL="${OLM_CHANNEL:-latest}" gauge run --log-level=debug --verbose --tags install specs/olm.spec || true
```
etirta:
```suggestion
CONSOLE_URL="$(oc whoami --show-console)" \
    API_URL="$(oc whoami --show-server)" \
    CATALOG_SOURCE=redhat-operators \
    CHANNEL="${OLM_CHANNEL:-latest}" \
    gauge_reports_dir="${ARTIFACT_DIR}" \
    overwrite_reports=false \
    gauge run --log-level=debug --verbose --tags install specs/olm.spec || true
```

Please investigate whether `|| true` is actually necessary. We should consider whether we should allow the step to error out instead to avoid masking legitimate failures.

For example, it is acceptable if we rely solely on JUnit test case statuses for reporting failures while wanting the overall test run step to return a successful exit code.

ci-operator/step-registry/openshift-pipelines/install/openshift-pipelines-install-commands.sh
etirta: Thx @srivickynesh for the info, now I can suggest accordingly.

etirta (changes requested):
Some more suggestions based on @srivickynesh feedback.

ci-operator/step-registry/openshift-pipelines/install/openshift-pipelines-install-commands.sh
oharan2: `${SHARED_DIR}/api.log` in this file has the whole login command like 
`oc login -u rosa-admin -p <password> api_url`.
Updating script.

etirta: Ok, that command do not change anything in the global context. So we do do:
```shell
else
    (set +x; eval "$(cat "${SHARED_DIR}/api.login")")
fi
```

ci-operator/step-registry/openshift-pipelines/tests/openshift-pipelines-tests-commands.sh
oharan2: We don't need that, this loop only iterates on XMLs and:
1. Rename (done by the exittrap which produces a ` junit--..` file)
2. Bump them up one level from subdir to `ARTIFACT_DIR` (the exittrap finds all XMLs in original location)

I verified that the original location would be still achievable for exit trap by stimulating the fetch from subdir. See ref log: [pipelines_map.txt](https://github.com/user-attachments/files/27215417/pipelines_map.txt)
- `ARTIFACT_DIR` is `openshift-pipelines-tests/dir`
- But XMLs are in subdir `openshift-pipelines-tests/dir/artifacts`

Removed this whole section as it's redundant.
In any way, will verify all recent changes in rehearsal.

oharan2:
/pj-rehearse periodic-ci-openshift-pipelines-release-tests-release-v1.21-ocp-4.22-lp-interop-cr-openshift-pipelines-aws

ci-operator/config/openshift-pipelines/release-tests/openshift-pipelines-release-tests-release-v1.21__ocp-4.22-lp-interop.yaml
```diff
@@ -48,7 +48,7 @@ tests:
       FIREWATCH_FAIL_WITH_TEST_FAILURES: "true"
       MAP_TESTS: "true"
       OCP_VERSION: "4.22"
-      REPORTPORTAL_CMP: lp-interop-OpenshiftPipelines
+      REPORTPORTAL_CMP: lp-ocp-compat--OpenshiftPipelines
```
etirta: You have already defined this environment variable with a default value in the corresponding step configuration. There is no need to set it here with the exact same value. We should only include it in the CI configuration if we need to override the default with a different value.

ci-operator/step-registry/openshift-pipelines/install/openshift-pipelines-install-commands.sh
etirta: Please see https://github.com/openshift/release/pull/78416#discussion_r3161984863 and https://github.com/openshift/release/pull/78416#discussion_r3163643830.
```suggestion
if [ -s "${KUBECONFIG}" ]; then
    oc whoami
else #login for ROSA & Hypershift platforms
    (set +x; eval "$(cat "${SHARED_DIR}/api.login")")
fi
```

ci-operator/step-registry/openshift-pipelines/tests/openshift-pipelines-tests-commands.sh
etirta, line 34: Following our Best Practices, please be disciplined in pre-declaring all variables used in the script.
```shell
 ...
typeset -a specs=()
 ...
```
Pre-declaring all variables at the beginning of the script is a good practice, especially when `set -u` (`nounset`) is active. In this particular case, the script might work without it, but tracking which variables need pre-declaration and which do not creates a maintenance burden. Always pre-declaring everything removes that burden entirely.

etirta (changes requested):
@oharan2 Look much better. A few more comments then I think it's a go.

ci-operator/config/openshift-pipelines/release-tests/openshift-pipelines-release-tests-release-v1.21__ocp-4.22-lp-interop.yaml
oharan2: This is not applicative, since the `mpiit-data-router-reporter` step has to have the product specific identifier (no default). 
I understand that it duplicates, but I don't see any better alternative.
Do you want me to set the test&install steps default value (for mapping) with `""`?

ci-operator/step-registry/openshift-pipelines/install/openshift-pipelines-install-commands.sh
oharan2: Waiting for PQEs response

ci-operator/config/openshift-pipelines/release-tests/openshift-pipelines-release-tests-release-v1.21__ocp-4.22-lp-interop.yaml
etirta: Ahh... thx for clarification. In this case, you are right, this is needed due to dependency with from `mpiit-data-router-reporter` inside `firewatch-ipi-aws-cr` Workflow.

ci-operator/step-registry/openshift-pipelines/tests/openshift-pipelines-tests-commands.sh
oharan2: Thanks, updated.

oharan2:
/pj-rehearse periodic-ci-openshift-pipelines-release-tests-release-v1.21-ocp-4.22-lp-interop-cr-openshift-pipelines-aws

oharan2:
/pj-rehearse periodic-ci-openshift-pipelines-release-tests-release-v1.21-ocp-4.22-lp-interop-cr-openshift-pipelines-aws

ci-operator/step-registry/openshift-pipelines/install/openshift-pipelines-install-commands.sh
srivickynesh: exitcode is used to differentiate failures, won't require "|| true."

ci-operator/step-registry/openshift-pipelines/install/openshift-pipelines-install-commands.sh
```diff
@@ -1,6 +1,13 @@
 #!/bin/bash
 set -euxo pipefail; shopt -s inherit_errexit
 
+typeset CONSOLE_URL=''
+typeset API_URL=''
+typeset -x gauge_reports_dir="${ARTIFACT_DIR}"
+typeset -x CHANNEL="${OLM_CHANNEL:-latest}"
+typeset -x overwrite_reports='false'
+typeset -x CATALOG_SOURCE='redhat-operators'
```
etirta: You misunderstood my previous point. When I mentioned always declaring variables, I was referring to the variable name `specs`, as suggested in this [comment](https://github.com/openshift/release/pull/78416#discussion_r3169000482).
```suggestion
```

None of the identifiers here (such as `CONSOLE_URL` through `CATALOG_SOURCE`) need to be pre-defined. When you run `FOO=bar cmdToExecute`, the shell does not define `FOO` as a variable in the caller's context. Instead, it creates an on-the-fly environment variable within the child process context.

Furthermore, using on-the-fly environment variables is recommended so that only the affected child process receives the settings. This is especially important if the environment variable contains a secret; you won't need to worry about cleaning it up to avoid unintentional leaks to other child processes.

See:
```shell
etirtara@14c692f90aab.11 2026-05-04 w19 23:16:53 "/wrk--edttjRedHat.RHP-k8s--dev/local/repo/openshift/release/sg-rh~abi"
$ typeset -p FOO
bash: typeset: FOO: not found

etirtara@14c692f90aab.12 2026-05-04 w19 23:16:57 "/wrk--edttjRedHat.RHP-k8s--dev/local/repo/openshift/release/sg-rh~abi"
$ FOO=bar bash -c 'typeset -p FOO'
declare -x FOO="bar"

etirtara@14c692f90aab.13 2026-05-04 w19 23:17:17 "/wrk--edttjRedHat.RHP-k8s--dev/local/repo/openshift/release/sg-rh~abi"
$ typeset -p FOO
bash: typeset: FOO: not found
```

ci-operator/step-registry/openshift-pipelines/tests/openshift-pipelines-tests-commands.sh
```diff
@@ -1,6 +1,15 @@
 #!/bin/bash
 set -euxo pipefail; shopt -s inherit_errexit
 
+typeset -a specs=()
+typeset spec=''
+typeset CONSOLE_URL=''
+typeset API_URL=''
+typeset -x GOPROXY='https://proxy.golang.org/'
+typeset -x gauge_reports_dir="${ARTIFACT_DIR}"
+typeset -x overwrite_reports='false'
```
etirta:
```suggestion
```

ci-operator/step-registry/openshift-pipelines/install/openshift-pipelines-install-commands.sh
```diff
@@ -27,10 +34,6 @@ fi
 # Install openshift-pipelines operator (olm.spec)
 CONSOLE_URL="$(oc whoami --show-console)" \
     API_URL="$(oc whoami --show-server)" \
-    CATALOG_SOURCE=redhat-operators \
-    CHANNEL="${OLM_CHANNEL:-latest}" \
-    gauge_reports_dir="${ARTIFACT_DIR}" \
-    overwrite_reports=false \
```
etirta:
```suggestion
    CATALOG_SOURCE=redhat-operators \
    CHANNEL="${OLM_CHANNEL:-latest}" \
    gauge_reports_dir="${ARTIFACT_DIR}" \
    overwrite_reports=false \
```

ci-operator/step-registry/openshift-pipelines/tests/openshift-pipelines-tests-commands.sh
```diff
@@ -35,9 +44,6 @@ done
 
 CONSOLE_URL="$(oc whoami --show-console)" \
     API_URL="$(oc whoami --show-server)" \
-    GOPROXY="https://proxy.golang.org/" \
-    gauge_reports_dir="${ARTIFACT_DIR}" \
-    overwrite_reports=false \
```
etirta:
```suggestion
    GOPROXY="https://proxy.golang.org/" \
    gauge_reports_dir="${ARTIFACT_DIR}" \
    overwrite_reports=false \
```

etirta (changes requested):
This should clarify the misunderstanding regarding my previous point.

Please handle the refactoring of the Data Router step and update this accordingly. Run it for a few days while updating the AI documentation. Afterwards, run it against the remaining Launch Pads (LPs).

ci-operator/step-registry/openshift-pipelines/install/openshift-pipelines-install-commands.sh
oharan2: Thank you for the details. Addressed.

oharan2:
/pj-rehearse periodic-ci-openshift-pipelines-release-tests-release-v1.21-ocp-4.22-lp-interop-cr-openshift-pipelines-aws

oharan2:
/pj-rehearse periodic-ci-openshift-pipelines-release-tests-release-v1.21-ocp-4.22-lp-interop-cr-openshift-pipelines-aws

oharan2:
Waiting for the merge of #78819:
/hold

srivickynesh (approved):
Overall LGTM. Could you start the CI Jobs for any breaking changes?

oharan2:
> Overall LGTM. Could you start the CI Jobs for any breaking changes?

Thanks @srivickynesh.
Since the last run that was acknowledged, I only updated the following env var **_name_**:
`REPORTPORTAL_CMP` -> `DR__RP__CR_COMP_NAME`
(To be compatible with changes implemented by another PR that is merging soon)

Therefore, I don't think it requires further verification. 
Hope this is acceptable with you.
Thanks!

PR #78474, opened 2026-04-28
Add CSR approval to ODH preflight health check

dayakar349:
Approve pending CSRs before the ServiceAccount controller check to work around MCO CSR renewal timing gap on on long-idle clusters. 

## Summary by CodeRabbit

* **New Features**
  * Automatically approves pending certificate signing requests during preflight health checks
  * Includes CSR state and approval duration in preflight diagnostics and summary metrics
  * Improves failure messaging to indicate CSR approval was attempted as part of recovery efforts

### DIFF
--- a/ci-operator/step-registry/opendatahub-operator-e2e/cluster-preflight-health-check/opendatahub-operator-e2e-cluster-preflight-health-check-commands.sh
+++ b/ci-operator/step-registry/opendatahub-operator-e2e/cluster-preflight-health-check/opendatahub-operator-e2e-cluster-preflight-health-check-commands.sh
@@ -10,22 +10,36 @@ NODE_WAIT_START=$SECONDS
 oc wait --for=condition=Ready nodes --all --timeout=4m
 NODE_WAIT_DURATION=$((SECONDS - NODE_WAIT_START))
 
-# ServiceAccount controller health check
-# After Hive hibernation resume, kube-controller-manager needs time to sync
-# informer caches before the SA controller can process namespace events
+# ServiceAccount controller health check with continuous CSR approval
+# Workaround for MCO CSR renewal timing gap on long-idle clusters
+# Kube-controller-manager needs time to sync informer caches before
+# the SA controller can process namespace events
 SA_CHECK_START=$SECONDS
 SA_TIMEOUT=600  # 10 minutes per analysis of 14-day CI data
 
 oc create namespace ci-preflight
 
+echo "Waiting for ServiceAccount controller (with continuous CSR approval)..."
 # Wait for default ServiceAccount to be auto-created by SA controller
-if ! timeout "${SA_TIMEOUT}" bash -c 'until oc get serviceaccount default -n ci-preflight &>/dev/null; do sleep 5; done'; then
+# Approve any pending CSRs during the wait to unblock kubelet communication
+if ! timeout "${SA_TIMEOUT}" bash -c '
+  until oc get serviceaccount default -n ci-preflight &>/dev/null; do
+    # Approve any pending CSRs while waiting
+    PENDING=$(oc get csr --no-headers 2>/dev/null | awk '\''$6=="Pending" {print $1}'\'' || true)
+    if [[ -n "$PENDING" ]]; then
+      echo "Approving $(echo "$PENDING" | wc -l) pending CSR(s)..."
+      echo "$PENDING" | xargs oc adm certificate approve 2>/dev/null || true
+    fi
+    sleep 5
+  done
+'; then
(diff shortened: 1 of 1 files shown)


ci-operator/step-registry/opendatahub-operator-e2e/cluster-preflight-health-check/opendatahub-operator-e2e-cluster-preflight-health-check-commands.sh
```diff
@@ -10,6 +10,29 @@ NODE_WAIT_START=$SECONDS
...
+# Approve pending CSRs — workaround for MCO CSR renewal timing gap on
+# long-idle/hibernated clusters where kubelet certs expire (OCPBUGS-61847)
+CSR_APPROVE_START=$SECONDS
+echo "CSR state before approval:"
+oc get csr --no-headers 2>/dev/null | head -20 || true
+
+PENDING_CSRS=$(oc get csr --no-headers | awk '$6=="Pending" {print $1}' || true)
+if [[ -n "$PENDING_CSRS" ]]; then
+    echo "Found $(echo "$PENDING_CSRS" | wc -l) pending CSR(s), approving..."
+    echo "$PENDING_CSRS" | xargs oc adm certificate approve
```
macgregor: did you verify the CI environment has oc adm?

ci-operator/step-registry/opendatahub-operator-e2e/cluster-preflight-health-check/opendatahub-operator-e2e-cluster-preflight-health-check-commands.sh
```diff
@@ -10,6 +10,29 @@ NODE_WAIT_START=$SECONDS
...
+    sleep 10
+    # Second pass: approving kubelet client CSR triggers a serving CSR
```
macgregor: hmmm. I wonder if it makes more sense to just put the approval inside the loop below. so each time the loop wakes up we approve any pending csrs while we wait for the cluster to become healthy. should help with racy stuff. 

```
until oc get serviceaccount default -n ci-preflight &>/dev/null; do sleep 5; done
```
replace the `do sleep 5` with approve and then sleep. might need to refactor that line a bit into a proper bash loop...

macgregor (changes requested):

ci-operator/step-registry/opendatahub-operator-e2e/cluster-preflight-health-check/opendatahub-operator-e2e-cluster-preflight-health-check-commands.sh
dayakar349: Yes, this step uses `from: cli` which is the full openshift/cli image with all oc adm subcommands.
```
  as: opendatahub-operator-e2e-cluster-preflight-health-check
  from: cli
  commands: opendatahub-operator-e2e-cluster-preflight-health-check-commands.sh
```

ci-operator/step-registry/opendatahub-operator-e2e/cluster-preflight-health-check/opendatahub-operator-e2e-cluster-preflight-health-check-commands.sh
dayakar349: That makes sense, moved CSR approval into the SA wait loop for continuous approval every 5s. Thanks for the suggestion!

ci-operator/step-registry/opendatahub-operator-e2e/cluster-preflight-health-check/opendatahub-operator-e2e-cluster-preflight-health-check-commands.sh
```diff
@@ -10,6 +10,29 @@ NODE_WAIT_START=$SECONDS
...
+    PENDING_CSRS=$(oc get csr --no-headers | awk '$6=="Pending" {print $1}' || true)
+    if [[ -n "$PENDING_CSRS" ]]; then
+        echo "Found $(echo "$PENDING_CSRS" | wc -l) additional pending CSR(s), approving..."
+        echo "$PENDING_CSRS" | xargs oc adm certificate approve
+    fi
```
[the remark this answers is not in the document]
macgregor: dont think we care about this because these are ephemeral test clusters. if the cluster is somehow compromised in isolation and we grant a "malicious" CSR, theres not really much damage it could do before the cluster is disposed of. and if the CI process is compromised, it already has full admin access (and we have much bigger problems to worry about). I guess if youre being super paranoid scoping to expected kubelet signers makes sense, but i think its overkill.

PR #78484, opened 2026-04-28
Move the CLI tool to ci-tools repo and add dockerfile

psalajova:
## Summary by CodeRabbit

* **Refactor**
  * secret-manager migrated to run as a containerized application; workflow now uses a prebuilt image.
  * Local Python CLI and supporting local packages/tests removed; invocation now runs the tool inside the container.

* **Chores**
  * Launcher script updated to use container image, manage cloud auth cache, and pull the image before execution.


Prucek (approved):

hack/gcp-secret-manager/.gitkeep
```diff
@@ -0,0 +1,3 @@
+# This directory contains runtime state for the secret-manager tool.
```
danilo-gemoli, line 1: What is this file?

psalajova: It's so that the directory, where the virtual environment created by the script will be housed, won't be removed/will be tracked by github.
> It is an empty file (or sometimes a file with a single line of content) placed in an otherwise empty directory to force Git to track the directory

Technically it isn't necessary, but I liked this better than removing the directory when the script has not run.

Prucek (approved):

PR #78539, opened 2026-04-29
Don't recreate override config in aro hcp deprov step

mmazur:
## Summary by CodeRabbit

* **Chores**
  * Provisioning workflow now uses fixed configuration file locations from shared resources instead of generating timestamped files during each run
  * Deprovisioning workflow simplified to reference pre-existing configuration override files instead of creating new dynamic overrides
  * Reduces infrastructure complexity and file system operations during deployment cycles


mmazur:
/pj-rehearse pull-ci-Azure-ARO-HCP-main-e2e-parallel-infra-shard1

PR #78542, opened 2026-04-29
Add env var to skip chain trigger in curl triggers

shaior:
## Summary by CodeRabbit

* **New Features**
  * Added support for skipping subsequent job triggers via a new configuration option, allowing users to control whether chained jobs execute in their CI/CD pipeline.


ci-operator/step-registry/telcov10n/functional/cnf-network/trigger-job/telcov10n-functional-cnf-network-trigger-job-ref.yaml
```diff
@@ -20,3 +20,6 @@ ref:
     - name: JOB_TYPE
       default: ""
       documentation: The type of job
+    - name: SKIP_CHAIN_TRIGGER
+      default: ""
```
rdiscala: Is it better to use "false" or `false` here, instead of the empty string, since the value type is implied to be a boolean?

PR #78550, opened 2026-04-29
Disabling threshold for virt runs

josecastillolema:
## Summary by CodeRabbit

* **Chores**
  * Updated default configuration for virtual machine instance ready threshold across test step registries. A threshold value of 0 is now ignored by default.


capolrik (approved):

rsevilla87 (commented):
I wonder if it's better to set 0 as default in kube-burner-ocp, so we don't need to pass this flag anymore. wdyt @josecastillolema

